PR #182, opened 2015-11-19
introducing the kit

cvan:


cvan:
unnecessary per https://github.com/aframevr/aframe-site

PR #228, opened 2015-11-25
Is a root node really necessary?

dmarcos:
This is not meant to be merged. It's just me trying to get rid of the root node. @cvan Do you think we really need it? Is there a way we can get rid of it? This PR seems to work but probably my approach is simplistic


ngokevin:
I think we can close this PR. We plan on moving to use primitives, which extend entities rather than template entities.

cvan:
There are some good points raised in here. Primitives are for wrapping and extending single entities. Templates serve a different use case: grouping multiple elements (and potentially some more advanced templating logic beyond the current variable string interpolation). Templates ought to continue to exist, but perf-wise there are some good ideas @dmarcos had that we ought to implement. In due time, yeah. I think they'll need to happen outside of this PR though.

PR #290, opened 2015-12-17
remove commented-out block from A-Frame logo animation scene

cvan:
r? @jcarpenter 

just to make the source shorter and less scary. if you need that markup later, you can refer back to this PR.


cvan:
r+ looks good

Issue #282, opened 2015-12-11
Add some mechanism for using `build/aframe.js` when running dev server


cvan:
Perhaps the act of running `npm run dev` (instead of `npm start` or `npm run server`) should rewrite the HTML files in the `examples` to use `build/aframe.js` instead of `dist/aframe.js`.

ngokevin:
nice job budo

PR #294, opened 2015-12-17
Fix Budo live-reload on mobile

mattdesl:
This PR introduces these changes:

- Removes the `host` field from budo options. This will make budo resolve to the internal IP, and log that instead of `0.0.0.0`. Now it looks like this:  
screen shot 2015-12-17 at 11 06 48 am
- The above change will fix LiveReload when you hit the IP on mobile devices. This seems like a bug with budo, I opened an issue here: https://github.com/mattdesl/budo/issues/118
- I have un-commented the `live()` feature since it now works on mobile. I'm not sure whether you had this commented intentionally (some devs prefer not to use it; my scripts sometimes include a `--live` option to toggle this).

Some other small things that could change with the build script:

- `browserify-css` should now emit file events (https://github.com/cheton/browserify-css/issues/20), so it should be picked up by watchify and trigger a hard live-reload. However it seems like all the browserify-css stuff is legacy, I am not seeing much CSS in this repo.
- The HTML inside `examples/` are not live-reloading
- The `browserifyArgs` option can be replaced by a `browserify: { ... }` options object for code clarity

### DIFF
--- a/scripts/budo.js
+++ b/scripts/budo.js
@@ -25,6 +25,7 @@ app
 .live()
 .on('watch', function (eventType, fn) {
   if (eventType !== 'change' && eventType !== 'add') { return; }
+
   if (path.extname(fn) === '.css') {
     // We want to trigger a reload of the entire document, since
     // browserify-css injects CSS into the page.
--- a/scripts/budo.js
+++ b/scripts/budo.js
@@ -16,17 +16,15 @@ var app = budo('./index.js:build/aframe.js', {
   debug: process.env.NODE_ENVIRONMENT !== 'production',
   verbose: true,
   stream: process.stdout,  // log to stdout
-  host: '0.0.0.0',
   port: 9000,
   browserifyArgs: '-s AFRAME'.split(' ')
 });
 
 app
 .watch('**/*.{css,js,html}')
-// .live()
+.live()
 .on('watch', function (eventType, fn) {
   if (eventType !== 'change' && eventType !== 'add') { return; }
-
   if (path.extname(fn) === '.css') {
     // We want to trigger a reload of the entire document, since
     // browserify-css injects CSS into the page.


cvan:
sweeet - thanks for the PR!

> - `browserify-css` should now emit file events (https://github.com/cheton/browserify-css/issues/20), so it should be picked up by watchify and trigger a hard live-reload. However it seems like all the browserify-css stuff is legacy, I am not seeing much CSS in this repo.

the CSS has since been moved to [`aframe-core`](https://github.com/aframevr/aframe-core/blob/master/style/aframe-core.css), so, yeah, can remove.

> - The HTML inside `examples/` are not live-reloading

this seemed to work with [my original patch](https://github.com/cvan/budo/commit/81c9e336). I gave up on using my fork because while I could get one example to work, loading multiple documents proved problematic (which I filed as mattdesl/budo#79). also, we used to use HTML imports in all the examples, which we removed but it's still a supported use case for folks who want to define templates in external HTML imports (I filed an issue against budo to address/discuss: mattdesl/budo#119).

now if I point to the latest budo, and enable the `.live()` line in `scripts/budo.js` (or using your branch here, with the `host` changes too), I no longer can get the examples to live reload.

> - The `browserifyArgs` option can be replaced by a `browserify: { ... }` options object for code clarity

:+1:

mattdesl:
The `examples/` HTML files get picked up by the watch task, but nothing is telling budo to `reload()` when they change. I'm not exactly sure how the templates are being handled here (haven't looked too closely) but it seems like getting rid of the `.live()`, `.watch()` and `.on('watch')` and just replacing it with `live: true` in options should mostly work (defaults to watching JS/CSS/HTML and triggering reloads).

Will test further...

cvan:
gotcha, thanks.

also, related, which you probably already noticed: #282

cvan:
thanks!

I figured out the other issues - PR coming :smile:

PR #310, opened 2015-12-19
serve build vs. dist with server (fixes #282); fix Live Reloading (fixes #309, #109)

cvan:
r? @mattdesl @ngokevin

PR #320, opened 2015-12-21
Serve build when deving

mattdesl:
I branched this off @cvan's PR, #310. Features changed:

- Move `opt.live` to `consts.LIVE` so that it doesn't get sent as an option to budo (don't want to duplicate reloads or anything)
- Remove the build step in `npm run dev` since it's now superfluous. Even if a local `build` directory doesn't exist, budo will still serve our watchified (incremental) `src/index.js` bundle during development


scripts/budo.js
```diff
@@ -12,29 +12,51 @@ function execCmd (cmd) {
...
+  port: process.env.PORT || consts.PORT,
+  browserifyArgs: ['-s', consts.NAME],
+  middleware: function (req, res, next) {
+    // Route `dist/aframe.js` to `build/aframe.js` so we can
+    // dev against the examples :)
+    var path = urlParse(req.url).pathname;
+    if (path.indexOf('/' + consts.DIST) !== -1) {
+      req.url = req.url.replace('/dist/', '/build/');
+    }
+    // TODO: Consider adding middleware that targets specific directories
```
cvan: do you have ideas of how to cleanly do this?

scripts/budo.js
```diff
@@ -12,29 +12,51 @@ function execCmd (cmd) {
...
+};
+
+var app = budo(consts.ENTRY + ':' + consts.BUILD, opts);
+
+app.watch(consts.WATCH)
+.on('update', function () {
   execCmd('semistandard -v $(git ls-files "*.js") | snazzy');
 });
+
+if (consts.LIVE) {
```
cvan: ideally, it'd be nice if budo supported a non-hardcoded value for `watch`, so that we could just pass `opts.watch` and `opts.live` and be good to go and not have to recreate this ourselves. thoughts?

scripts/budo.js
mattdesl: I think this is only possibly by extending `livereload-js`. Calling `budo.reload()` just ends up telling the live reload server to trigger a [full page reload](https://github.com/livereload/livereload-js/blob/e1d943628005ad8d18a50ee1e8c29858ca748d10/src/reloader.coffee#L95), so it ends up affecting all open tabs on the same host + port.

Maybe the `livereload-js` server could have a `reloadHref(regexToMatch)` function which only triggers a page reload when the client's `location.href` matches the regex.

cvan:
seems to work well - thanks for cleaning things up :+1:

scripts/budo.js
cvan: ah, gotcha, that's what I thought. I assume this is a valid use case others care about, but definitely not urgent IMO.

cvan:
let me know about the above harcoded `watch` above and we can merge (we can do address that as an issue against `budo` later as well)

scripts/budo.js
mattdesl: How about a `watchGlob` option that overrides the default? This way it doesn't introduce any side-effects (like turning on chokidar) and won't cause confusion with the way the `watch()` function works.

cvan: :+1: I like that

mattdesl:
`watchGlob` added to `budo@7.1.0`, small example here:
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#http-server

I updated this PR.

Things to note:

- I made sure the `open` and `live` options being passed are booleans rather than string/number
-  `watchGlob` only needs to specify files that won't be picked up by browserify's `require()` algorithm. Because `polymerize` emits its HTML dependencies, watchify/budo will trigger an `'update'` and LiveReload when one of the templates changes.
- Now the `watchGlob` is just looking at the `examples/` folder

The only file that isn't triggering a live-reload now is `VRLoader.js`, but I can't find any require references to it.

cvan:
> `watchGlob` added to `budo@7.1.0`, small example here:
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#http-server

heh, that was quick. amazing. :tada: 

> The only file that isn't triggering a live-reload now is `VRLoader.js`, but I can't find any require references to it.

yeah, we aren't using it anymore. will be revisiting the loader likely with aframevr/aframe-core#16.

cvan:
can you pull `dev` and rebase when you get a chance?

mattdesl:
K, should be fixed with `dev` now.

scripts/budo.js
```diff
@@ -12,29 +12,39 @@ function execCmd (cmd) {
   return p;
 }
 
-var app = budo('./index.js:build/aframe.js', {
+var consts = {
+  NAME: 'AFRAME',
+  ENTRY: './src/index.js',
+  DIST: 'dist/aframe.js',
+  BUILD: 'build/aframe.js',
+  WATCH: 'examples/**/*', // Additional files to watch for LiveReload
```
cvan: I typically put two spaces for inline comments, but if `semistandard` didn't complain, neither can I :smirk:

scripts/budo.js
```diff
@@ -12,29 +12,39 @@ function execCmd (cmd) {
...
-  if (path.extname(fn) === '.css') {
-    // We want to trigger a reload of the entire document, since
-    // browserify-css injects CSS into the page.
-    app.reload();
-  } else if (path.extname(fn) === '.js') {
-    app.reload(fn);
-  } else if (fn.indexOf('elements/templates') === 0) {
-    app.reload();
+  open: Boolean(process.env.OPEN),  // To enable: `OPEN=1 npm start`
+  live: Boolean(process.env.LIVE),  // To enable: `LIVE=1 npm start`
```
cvan: do you think live should be the default? I'm thinking yes, but I want to make it easily configurable for folks to disable live reloading + auto-opening of the server URL. thoughts here? the env vars are making me a bit uncomfortable, though I initially thought it was a decent solution.

mattdesl: I'm a fan of LiveReload so I'd say yeah, but I know a lot of devs who get annoyed with it. :smile:

scripts/budo.js
cvan: good call with the simplified glob pattern, since the other things will be picked up by browserify

scripts/budo.js
cvan: hate to nitpick, but you think we could have enable Live Reload by default (and adjust the `README` accordingly):

```js
'LIVE' in process.env ? Booelan(process.env.LIVE) : true
```

and keep `OPEN` disabled by default probably.

cvan:
thanks, this works great.

p.s. I like keeping everything in this script, but it's interesting to me that now actually most of the stuff we're doing in this budo script can be done via the budo CLI except the middleware. :+1:

mattdesl:
Actually, since they are (always?) converted to strings, it has to be something like this:

```js
  live: 'LIVE' in process.env  // To disable: `LIVE=0 npm start`
    ? String(process.env.LIVE) === '1'
    : true,
  ...
```

The env vars are a bit clunky. Maybe just enable live reload by default and move the live/open flags to a new PR. I think `--` is a cleaner approach and also easier to type.

cvan:
> The env vars are a bit clunky. Maybe just enable live reload by default and move the live/open flags to a new PR. I think `--` is a cleaner approach and also easier to type.

:+1: let's do that. thanks.

cvan:
:shipit: awesome, everything looks + works well

cvan:
thanks, @mattdesl!

Issue #157, opened 2015-11-13
Create a `ring` template


cvan:
https://github.com/MozVR/aframe-core/blob/93f1bdc/src/components/geometry.js#L80-L83

cvan:
done in #336 :smile:

Issue #241, opened 2015-11-30
Flesh out the Contributing Guidelines


cvan:
See [our `CONTRIBUTING.md`](https://github.com/MozVR/aframe/blob/master/CONTRIBUTING.md).

For inspiration, [three.js'](https://github.com/mrdoob/three.js/blob/master/CONTRIBUTING.md) is pretty good.

cvan:
We should also describe `dev` (bleeding edge) vs. `master` (stable) à la https://github.com/aframevr/aframe-core/issues/702#issue-122857808

cemkod:
Is the new CONTRIBUTING.md good enough? Any other concerns?

ngokevin:
Good enough to close. We can modify if we want later.

PR #322, opened 2015-12-21
Added ring template (Fixes issue #157)

cemkod:


cvan:
first off, woo thanks!

we haven't mentioned it in the Contributing doc yet (#241), but no need to run `dist` when pushing changes. also, see #320 that uses the `build/` when serving the local server (instead of `dist/`).

also, looks like `a-ring.html` is missing in this branch

cemkod:
I seem to have forgot to add the files. I'm still learning git. Sorry for that.

So is doing **npm start** after changing the files enough to test the changes I made?

cvan:
> So is doing npm start after changing the files enough to test the changes I made?

until #320 lands, the best way to test is to

1. run `npm start`
2. change `../../dist/aframe.js` to `../../build/aframe.js` in `examples/template-ring/index.html`
3. make changes, load the page, etc.

it'd probably be easier to do a new branch to remove the `dist/` changes. also, something we (I) haven't done a good job of communicating yet (again: apologies, will address in #241) is that PRs should be made against `dev` (unstable, under "development") instead of `master` (stable).

so you'll likely want to close this PR, commit only `examples/` and `elements/` files, and open a new one PR against `dev` instead of `master`.

thanks for your understanding - and thanks for contributing!

cemkod:
I'm afk now. No worries, I understand. Will do once I get back.

PR #326, opened 2015-12-22
rename to "Spheres & Fog" (to be consistent with "Cursor & Hover")

cvan:
r? @jcarpenter 

related: aframevr/aframe-site#107


jcarpenter:
Makes sense to me :+1:

Issue #347, opened 2015-12-30
Problems including aframe in browserify


mattpolicastro:
Apologies if this is missing the obvious, but I'm trying to add aframe as a project dependency based on the instructions listed in the README:
````
$ npm install aframe
````
And inside `index.js`:
````
require('aframe')
````
However, when I try to run browserify (installed globally), I get errors related to dependencies:
````
$ browserify index.js
Error: Cannot find module 'polymerize' from '/test-aframe/node_modules/aframe'
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)
````
Again, apologies if this is barking up the wrong tree but I'm out of ideas.

mareknowakowski1:
Hi,

It's because you don't have it in your node_modules directory I suppose. Go to you aforce node_module and then try npm install. This should install all dependencies from aforce directory. good luck!

mattpolicastro:
Thanks for the pointer. I suppose I'm still wrestling with this: if someone adds Aframe to their project purely as a dependency, shouldn't Aframe install its own dependencies as well? I might be stretching here, but I think that's a reasonable expectation when adding a utility to a project.

mareknowakowski1:
No problem :) Right it should be. I had the same problem. Unfortunately if you are adding it to your own dependencies then it doesn't run all other stuff - just downloading aforce. Thats it. 

Did you manage to handle that error ? :)

mattpolicastro:
Yes, the tip you provided worked. However, I believe my point still stands: Aframe should install dependencies necessary to its functioning as part of the install process.

cvan:
Very sorry about this; we'll be addressing this very soon. Thanks for filing!

mattpolicastro:
[...]

PR #357, opened 2016-01-05
Move browserify dependencies in package.json

mattpolicastro:
As discussed in #347, I moved some of the browserify-specific dependencies so they're properly installed during a bundling process.


ngokevin:
Thanks!

PR #775, opened 2016-01-13
Creates an objects3D variable where the components will store their c…

dmarcos:
…reated meshes, lights, cameras, and sounds.

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -16,8 +16,14 @@ module.exports.Component = registerComponent('camera', {
   init: function () {
     var camera = this.camera = new THREE.PerspectiveCamera();
     var el = this.el;
-    camera.el = el;
-    el.object3D.add(camera);
+    el.setObject3D('camera', camera);
+  },
+
+  /**
+   * Remove camera on remove (callback).
+   */
+  remove: function () {
+    this.el.setObject3D('camera', null);
   },
 
   /**
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -70,7 +70,7 @@ module.exports.Component = registerComponent('sound', {
   },
 
   remove: function () {
-    this.el.object3D.remove(this.sound);
+    this.el.setObject3D('sound', null);
   },
 
   /**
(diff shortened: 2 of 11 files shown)


src/components/camera.js
```diff
@@ -16,8 +16,14 @@ module.exports.Component = registerComponent('camera', {
...
-    camera.el = el;
-    el.object3D.add(camera);
+    el.setObject3D('camera', camera);
+  },
+
+  /**
+   * Remove camera on remove (callback).
+   */
+  remove: function () {
+    this.el.setObject3D('camera');
```
ngokevin: can we get a separate `removeObject3D`?

src/core/a-entity.js
```diff
@@ -44,7 +44,9 @@ var proto = Object.create(ANode.prototype, {
       this.states = [];
       this.components = {};
       this.paused = true;
-      this.object3D = new THREE.Mesh();
+      this.object3D = new THREE.Group();
+      this.object3D.el = this;
+      this.objects3D = {};
```
ngokevin: correct plural `object3Ds`

src/core/a-entity.js
```diff
@@ -127,6 +129,40 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  getObject3D: {
+    value: function (type, Constructor) {
```
ngokevin: this is the same as `getOrCreate`

src/core/a-entity.js
dmarcos: object3Ds is kind of weird

ngokevin: `object3DMap`?

src/components/sound.js
```diff
@@ -70,7 +70,7 @@ module.exports.Component = registerComponent('sound', {
   },
 
   remove: function () {
-    this.el.object3D.remove(this.sound);
+    this.el.setObject3D('sound', null);
```
ngokevin: kinda feel like a `removeObject3D` would be more explicit and clear in the long run

src/core/a-entity.js
ngokevin: somewhat prefer `object3DMap`, `objects3D` implies an array. Or it implies "`objects`, but 3D"

dmarcos:
Change variable name from objects3D to objects3DMap

ngokevin:
can we do `object3DMap`?

Issue #759, opened 2016-01-13
Add events to know when VR mode has been entered


ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Thursday Dec 17, 2015 at 21:27 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/698_

----

It's useful to know when VR mode has been entered. With the current API that piggybacks off of `requestFullscreen`, you can listen for the `fullscreenchange` event, but it's kind of a hack because you don't know necessarily if it's a `vrDisplay`.

Related: see aframevr/aframe#286 and https://github.com/aframevr/aframe/pull/286#issuecomment-164705052

and also `match-media` proposal here: https://github.com/aframevr/aframe-core/pull/518/files#diff-68b5544f3cdb6b74d38ac2e61a2e2f74R37

ngokevin:
https://github.com/aframevr/aframe/pull/776

`enter-vr` and `exit-vr` as well as states on the scene.

PR #776, opened 2016-01-13
move scene logic to components (fixes #239, #759, #886)

### LINKED ISSUE #239: Add a way of hiding the "Enter VR" button via query-string param
`http://localhost:9000/examples/cubes/?ui=0` or `http://localhost:9000/examples/cubes/?hide_ui=1` or something

Useful for when you want to `<iframe>` an A-Frame scene and perhaps want to enable some other method of entering VR - or you may not not even want the user to view the scene in VR.

ngokevin:
Scene components:

- **canvas** - Controls creation or selection of a canvas render target.
- **stats** - Handles RStats.
- **keyboard-shortcuts**
- **vr-mode-ui** - Handles VR UI (button/modal).

Scene modules:

- **fullscreen** - Handles fullscreen events.
- **iframe** - all the postmesage stuff
- **meta-tags** - Handles injecting of iOS metatags.
- **wakelock**

Additional changes:

- **AEntity** - `HTMLElement.setAttribute` to DOM for default components for scene.
- **canvas** - Height/width configurable, triggers a `render-target-loaded` event on scene for scene to set up renderer.
- **look-controls** - Point canvas to scene canvas.
- **stats** - Put the UI within the scene.
- Remove ?vr URL parameter, not sure if it is being used or actually needed in the core.
- **vr-mode-ui** - Can be disabled. Put the UI within the scene.

Tested:

- Some unit tests for new components
- Macbook Pro on Firefox/Chrome. Enter/exit VR work for Firefox, modal is thrown on Chrome as before
- On iPhone 6, enter/exit VR and orientation modal work as before


src/core/a-scene.js
```diff
@@ -93,24 +89,12 @@ var AScene = module.exports = registerElement('a-scene', {
         var resizeCanvas = this.resizeCanvas.bind(this);
         this.setupKeyboardShortcuts();
         this.attachFullscreenListeners();
```
caseyyee: Fullscreen events should be ignored when vr-mode is forced to VR?    Stereo rendering resets to mono when exiting fullscreen.

ngokevin: I pushed my updated changes to this PR. Moved fullscreen stuff to https://github.com/aframevr/aframe/pull/776/files#diff-344cbe83c2f901ba50d1f8824d89ccfeR18

So you want the use case to always force stereo and completely ignore fullscreen events?

caseyyee: Not sure, debatable I suppose.  my thought was that if you are going to manage your own viewport rendering, then you would want to also manually handle fullscreen events as well.

ngokevin: I'm a bit confused. Are you suggesting that *if vr-mode is forced to VR*, then ignore fullscreen events? And how should we allow vr-mode being forced to VR?

ngokevin:
Okay after what must be a good week of working this, this is ready.

src/core/a-entity.js
```diff
@@ -29,12 +29,12 @@ var AEntity;
...
  */
 var proto = Object.create(ANode.prototype, {
-  defaults: {
+  defaultComponents: {
     value: {
-      position: '',
-      rotation: '',
-      scale: '',
-      visible: ''
+      position: '0 0 0',
```
dmarcos: Can we set those to empty string so they're not defined for every entity?

tests/components/scene/canvas.test.js
```diff
@@ -0,0 +1,27 @@
+ /* global assert, process, suite, test */
+'use strict';
+
+suite('cavnas', function () {
```
dmarcos: Typo ```cavnas```

src/components/scene/meta-tags.js
```diff
@@ -0,0 +1,58 @@
+var register = require('../../core/component').registerComponent;
```
dmarcos: I'm not sure if this should be a component since responsiveness is part of the core value proposition. Should we make it configurable? I 100% agree it should be a separate module to clean a-scene but not sure about it being a component. This could cause fragmentation of experiences if people remove the component of their scenes by accident or misunderstanding.

src/components/scene/wakelock.js
```diff
@@ -0,0 +1,15 @@
+var register = require('../../core/component').registerComponent;
+var Wakelock = require('../../../vendor/wakelock/wakelock');
```
dmarcos: Same for this. Not sure it should be a component. This should not be optional or configurable.

src/components/scene/meta-tags.js
ngokevin: What if I just remove the `remove` handler so that it cannot be removed?

src/components/scene/wakelock.js
ngokevin: There's not much way to remove this component, just as position is a non-optional component.

dmarcos: Why do you want to make it a component then. If it's not removable or not configurable? It's stuff that leaks out of the scene that it's an internal implementation detail.

dmarcos: Position are at least configurable.

ngokevin: It seemed to make sense with the system. Components modify behaviors of whatever they attach to. There doesn't seem to be a rule about components that they have to be configurable or optional.

dmarcos: when you attach a component to an entity it becomes part of its public API. Things that we don't want to expose to the outside world should stay as internal modules instead of being a component.

ngokevin: Okay, I will move wakelock + metatags into modules rather than components.

ngokevin:
Updated. Moved fullscreen/metatags/wakelock to modules. Added keyboard-shortcuts scene component.

src/core/a-entity.js
```diff
@@ -29,7 +29,7 @@ var AEntity;
  * @member {boolean} paused - true if dynamic behavior of the entity is paused
  */
 var proto = Object.create(ANode.prototype, {
-  defaults: {
+  defaultComponents: {
```
dmarcos: Are these changes related to the intent of the PR?

ngokevin: Yeah wanted to expose the scene components in the DOM

src/components/scene/keyboard-shortcuts.js
```diff
@@ -0,0 +1,35 @@
+var registerComponent = require('../../core/component').registerComponent;
+var THREE = require('../../lib/three');
```
dmarcos: Do we want to make this non configurable as well? It would be good for to keep these basic keyboard shortcuts consistent across all aframe experiences.

src/components/scene/stats.js
```diff
@@ -0,0 +1,55 @@
+var registerComponent = require('../../core/component').registerComponent;
+var RStats = require('../../../vendor/rStats');
```
dmarcos: Stats are a a dev tool so it's nice to make them optional/configurable

src/components/scene/keyboard-shortcuts.js
ngokevin: I was trying to make a text editor, and I couldn't input the letter "f" so I really wanted a way to turn that off.

src/components/scene/vr-mode-ui.js
```diff
@@ -0,0 +1,188 @@
+var registerComponent = require('../../core/component').registerComponent;
```
dmarcos: Devs will want to provide their own customized UIs so it's good to make this optional/configurable

src/components/scene/stats.js
ngokevin: yeah it's optional. if you want it, attach it to the scene. else, detach it.

src/components/scene/vr-mode-ui.js
ngokevin: yeah it can be disabled if they want to provide their own ui

PR #816, opened 2016-01-13
pass time to scene behavior callbacks

ngokevin:


dmarcos:
Instead of landing this I would prefer an implicit API that checks if the tick method is defined on the component. If it's defined the component is automatically registered as a behavior. There's nothing that the component developer has to manage besides defining (or not) the ```tick``` method.

ngokevin:
What's the ETA on `tick`? I'd like it in the meantime so that maybe we'd be able to publish a `beta` so I can somewhat release physics.

dmarcos:
I'm not working on the ```tick``` API at the moment. You can either wait or implement the new API as part of this PR.

dmarcos:
The new API would be just a logic that checks if the ```tick``` method is defined on the component at initialization. If it is the component is registered as behavior. We have to automatically remove the behavior when the component is removed and make sure that the ```tick``` methods are not called when the scene us paused.

PR #821, opened 2016-01-13
add core examples back to index.html

ngokevin:


examples/index.html
```diff
@@ -174,17 +164,30 @@
...
     </ul>
 
-    <h2>Components</h2>
-
-    <ul class="links">
-      <li><a href="entitycomponent-basics/">Basics</a></li>
-    </ul>
-
-
     <h2>Tests</h2>
```
cvan: is "Tests" the correct heading?

examples/index.html
```diff
@@ -174,17 +164,30 @@
...
     <ul class="links">
-      <li><a href="animation-template-test/">Animations + Templates</a></li>
+      <li><a href="animation/">Animation</a></li>
+      <li><a href="controls/">Controls</a></li>
+      <li><a href="cursor/">Cursor</a></li>
+      <li><a href="cube/">Cube</a></li>
+      <li><a href="cylinders/">Cylinders</a></li>
+      <li><a href="fog/">Fog</a></li>
+      <li><a href="geometries/">Geometries</a></li>
+      <li><a href="geometry-gallery">Geometry Gallery</a></li>
```
cvan: `/`

examples/index.html
ngokevin: It's sufficient I think. We use it to test so they're test examples.

cvan: okay, thought that was an accident. that's fine, yeah

dmarcos:
We have to change all the paths of the examples. They still point to ```<script src="../../dist/aframe-core.js"></script>``

ngokevin:
Ah, weird they were still working. Must be Budo magic.

PR #823, opened 2016-01-13
component tick

ngokevin:


ngokevin:
r? @dmarcos 

Good call, a lot easier than manually registering/removing behaviors to the scene.

src/core/component.js
```diff
@@ -57,35 +89,42 @@ Component.prototype = {
    * Called during component initialization and is only run once.
    * Components can use this to set initial state.
    */
-  init: function () { /* no-op */ },
+  init: noopFn,
```
dmarcos: That's kind of obscure. I liked better before.

src/core/component.js
```diff
@@ -57,35 +89,42 @@ Component.prototype = {
...
-   * Called to start any dynamic behavior
-   * like animations, AI, physics.
+   * Tick handler.
+   * Called on each tick of the scene render loop.
+   * Affected by play and pause.
+   *
+   * @param {number} t - Scene tick time.
    */
-  play: function () { /* no-op */ },
+  tick: noopFn,
```
dmarcos: You can just asigg to ```undefined``` for documentation purposes

src/core/component.js
ngokevin: lets me check if the user has defined something else.

src/core/component.js
```diff
@@ -38,11 +40,41 @@ var components = module.exports.components = {};  // Keep track of registered co
...
+  var self = this;
+
   this.el = el;
   this.data = {};
   this.buildData(this.parse(rawData));
   this.init();
   this.update();
+
+  // Set up tick behavior.
+  if (this.tick !== noopFn) {
```
dmarcos: If tick === undefined on the component prototype you can just do:
````javascript
if (this.tick) { ... }
````

File: src/core/component.js (hunk 2 above)
Comment (ngokevin) [inline src/core/component.js]: hm ok

File: src/core/component.js [hunk 4]
```diff
@@ -38,11 +40,41 @@ var components = module.exports.components = {};  // Keep track of registered co
...
+    tickFn = function (t) {
+      // Don't run tick if paused.
+      if (el.paused) { return; }
```
dmarcos: This logic should be in the render loop. we should check:
````javascript
if (!this.paused) {
  this.behaviours.forEach(function(behaviour) {
     if (behaviour.el.paused) { return; }
     behavior.tick(t);
  });
}
````

File: src/core/a-scene.js [hunk 5]
```diff
@@ -730,8 +730,8 @@ var AScene = module.exports = registerElement('a-scene', {
           stats('FPS').frame();
         }
         TWEEN.update(t);
-        this.behaviors.forEach(function (behavior) {
-          behavior.update();
+        this.behaviors.forEach(function (behaviorFn) {
```
dmarcos: this should be:
`````javascript
This logic should be in the render loop. we should check:
````javascript
if (!this.paused) {
  this.behaviours.forEach(function(behaviour) {
     if (behaviour.el.paused) { return; }
     behavior.tick(t);
  });
}
````

File: src/core/component.js [hunk 6]
```diff
@@ -38,11 +40,41 @@ var components = module.exports.components = {};  // Keep track of registered co
...
+    };
+
+    if (this.isScene) {
+      this.el.addBehavior(tickFn);
+    } else {
+      this.el.sceneEl.addBehavior(tickFn);
+    }
+
+    // Wrap remove to deregister tick on component detach.
+    var removeFn = this.remove;
```
dmarcos: Instead of this I would create a removeComponent method on the a-entity and call it wherever we remove a component. I think it's here:

https://github.com/aframevr/aframe/blob/dev/src/core/a-entity.js#L280

and here:

https://github.com/aframevr/aframe/blob/dev/src/core/a-entity.js#L321

ngokevin:
Updated

dmarcos:
It looks good but travis is complaining

ngokevin:
Restarted the build. Will have to see how often it flakes.

ngokevin:
Thanks!

PR #839, opened 2016-01-14
pass timeDelta into tick

ngokevin:
Would find it useful for physics and other components that do interpolation.

### DIFF
--- a/src/core/a-scene.js
+++ b/src/core/a-scene.js
@@ -41,6 +41,7 @@ var isMobile = utils.isMobile();
  * @member {bool} renderStarted
  * @member {object} stats
  * @member {object} stereoRenderer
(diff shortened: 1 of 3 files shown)


src/core/a-scene.js
```diff
@@ -721,27 +723,31 @@ var AScene = module.exports = registerElement('a-scene', {
...
         var stats = this.stats;
+        var timeDelta = time - this.time;
 
         if (stats) {
           stats('rAF').tick();
           stats('FPS').frame();
         }
-        TWEEN.update(t);
+
+        TWEEN.update(time);
```
dmarcos: Tweens should not run if the scene is paused. Can we put it inside the if statement below?

PR #842, opened 2016-01-15
defer loading of web font in case fails

cvan:
so when the CDN site (my site) is down the Examples index page still loads :smile:

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -5,9 +5,7 @@
     <title>A-Frame Examples</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width">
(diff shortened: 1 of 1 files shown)


examples/index.html
```diff
@@ -191,5 +189,23 @@
     </ul>
 
     <img src="_images/logo-mozilla-white.svg" id="logo-moz" alt="Mozilla">
+
+    <script>
+      (function () {
+        window.requestAnimationFrame(function () {
+          // Defer loading of web fonts.
+          var fonts = [
```
ngokevin: maybe let's simplify this block and just hardcode one webfont

ngokevin:
r+

PR #843, opened 2016-01-15
fix data building for single-prop components, test Component.buildData (fixes #838)

### LINKED ISSUE #838: single-property schema default value overwritten if no value is passed to component (attribute)

ngokevin:
- Fixes mixins for single-prop components.
- Fixes default values for single-prop components that don't define data.
- Isolate and unit test component data building logic.


ngokevin:
thanks, this will unblock me on some components

PR #844, opened 2016-01-15
emit event on child attach

ngokevin:
Need this for keyboard + layout component so that the layout component can reflow whenever a new text node is added.

Event is called `child-attached`, any better ideas for name?


ngokevin:
thanks!

PR #845, opened 2016-01-15
remae "loader" component name to "model"

ngokevin:
`loader` will still work for one more version, with a warning.


src/components/model.js
```diff
@@ -90,4 +99,7 @@ module.exports.Component = registerComponent('loader', {
       el.setObject3D('mesh', collada.scene);
     });
   }
-});
+};
+
+module.exports.Component = registerComponent('loader', ModelComponent);
+module.exports.Component = registerComponent('model', ModelComponent);
```
dmarcos: We should remove the loader one

ngokevin: i'd like a version of deprecation so that people have time to adjust their scenes. we can remove it the release after next?

dmarcos: this things will pollute the code and will won't remember to remove it. I think there are not that many examples out there so I'm not concerns with backwards compatibility

ngokevin: hm? most examples out there use models. we will remember to remove it, it's in the changelog.

dmarcos: I really don't want to deal with this right now. Old scenes will depend on the old version of aframe so they will be fine.

dmarcos:
I'm not sure that model is a good name either because the collada loader does actually load a complete subtree of Objec3D with bones, textures etc.

src/components/model.js
ngokevin: Okay, I think it was sensible to deprecate, but I changed it to your preference.

ngokevin:
I think internally, it might not be 100% accurate, but conceptually the end result is a 3D model, right?

ngokevin:
Just got a bit more worried about this since most people are pointing their `index.html`s to A-Frame **latest** on the CDN. We really should have released a versioned CDN link.

cvan:
> We really should have released a versioned CDN link.

Yeah, I agree. I thought we'd need to push out a bunch of hotfixes but we didn't. Major regrets.

ngokevin:
Will split to `collada-model`, `obj-model`, and deprecate `loader`.

PR #847, opened 2016-01-15
Replaces setOBject3D(type, null) with removeObject3D(type)

dmarcos:


ngokevin:
r+, maybe a test

dmarcos:
removeObject3D is just an alias of setObject3D(type, null) and setObject3D is already tested

PR #854, opened 2016-01-19
show Enter VR modal when goggles icon hovered only (fixes #846)

### LINKED ISSUE #846: Enter VR modal should not trigger on hover when invisible
Only hovering over the actual button should trigger the Enter VR modal.

cvan:


ngokevin:
r+ thanks, this was annoying and didn't get around to ti

PR #864, opened 2016-01-20
don't play animation if begin is set to event name (fixes #863)

### LINKED ISSUE #863: <a-animation> is playing even when begin="evtName" is passed
Regression introduced with the play/pause feature.

ngokevin:
@andrenatal ran into this when trying to build a speech recognition component.


dmarcos:
Can we write a test to cover for this?

Issue #191, opened 2015-11-20
Port primitive templates to be in JS instead of templates


cvan:
It's faster and let's us do cool defaults/heuristics without having to resort to template logic.

I have a branch of this working for cubes. I'd like to finish porting the other primitives to use it, and I'll open a PR then.

cvan:
Where we left off: https://github.com/jcarpenter/aframe/commits/primitives

cvan:
see PR #883

ngokevin:
https://github.com/aframevr/aframe/pull/883

Issue #828, opened 2016-01-14
Examples using templates are busted (especially in Chrome)


cvan:
All examples that use templates now have camera issues; I see this error in Chrome but not in Firefox:

> THREE.WebGLRenderer.render: camera is not an instance of THREE.Camera.

[Anime UI](http://localhost:9000/examples/showcase-anime-UI/) in Chrome vs Firefox:

![screenshot 2016-01-13 17 01 26](https://cloud.githubusercontent.com/assets/203725/12312812/497ef5ca-ba17-11e5-8f46-5c6ee511a03e.png)
<img width="1494" alt="screenshot 2016-01-13 17 02 54" src="https://cloud.githubusercontent.com/assets/203725/12312830/7e164a22-ba17-11e5-932a-ca78202ad053.png">

The [Spheres & Fog example](http://localhost:9000/examples/showcase-spheres-and-fog/) doesn't have any camera errors but also appears quite busted in Chrome:

![screenshot 2016-01-13 16 55 54](https://cloud.githubusercontent.com/assets/203725/12312783/020ba76a-ba17-11e5-9d4c-f3bfb163d786.png)

<hr>

This must be related to a recent aframe-core change and now since the merge it's become apparent. Perhaps related to #794?

cvan:
This is due to a change with the loading process that was upstreamed in the aframe-core -> aframe merge. @ngokevin is revisiting the Primitives work I started by extending `<a-entity>` so we don't have to worry about load order. This will still pose a problem for folks who ever use templates (or even just injecting inert objects into the scene that aren't registered custom elements of `<a-entity>` yet)  - so we'll need to fix the core of the issue later.

cvan:
So note to everyone: just use entities everywhere for the time being.

cvan:
this has been magically fixed by #776. so if we actually wanted to, we could keep templates in their current form. but I'm comfortable with removing them until we've confident we need them (see issue #930).

ngokevin:
Examples using templates mostly still not working.

ngokevin:
https://github.com/aframevr/aframe/pull/883

Issue #827, opened 2016-01-14
Video example busted in Firefox


cvan:
<img width="1494" alt="screenshot 2016-01-13 16 58 14" src="https://cloud.githubusercontent.com/assets/203725/12312759/d5ee209a-ba16-11e5-8a3a-7e9ac727db0b.png">

cvan:
Possibly related to #828

ngokevin:
https://github.com/aframevr/aframe/pull/883

Issue #261, opened 2015-12-06
Primitive instances inheriting attributes of other primitive instances in scene


jcarpenter:
Any primitive instance that does not manually declare it’s own attribute value (eg `radius="2"`) will automatically inherit the value of that attribute if it is declared on another primitive instance in the scene.

For example:

``` html
<!-- Orange -->
<a-sphere position="-2 4 -5" radius="0.5" color="#F16745"></a-sphere>

<!-- Green -->
<a-sphere position="2 4 -5" color="#7BC8A4"></a-sphere>
```
Expected result is the green sphere renders with `radius="1"`, which is default (as specified in the primitive definition). Instead it inherits the `0.5` value of the orange sphere.

![screen shot 2015-12-05 at 5 09 07 pm](https://cloud.githubusercontent.com/assets/1425497/11611152/25bdde16-9b73-11e5-8c59-afa99dc4fb6a.png)

cvan:
@ngokevin: can you confirm this is fixed in PR #883?

ngokevin:
This is fixed in the PR from the initial implementation on josh's branch, but I'm closing this issue since the new primitives never landed anywhere.

PR #883, opened 2016-01-22
extras/declarative-events and extras/primitives (fixes #122, #159, #191, #200, #261, #827, #828)

### LINKED ISSUE #122: Add `color` attribute to images template
Want to be able to specify just a color for an image template. Handy for stubbing layouts out quickly, without having to apply an image. 

Challenge is that specifying a `src` on a material (even if `src` is empty), seems to negate the color. So we'll either need to update core to still show a color if `src` is empty, or include JS inside the template to only add `src` to material if the template `src` attribute contains a value.

ngokevin:
- Initial implementation of primitives by @cvan which currently replaces templates until we have a better solution.
- Moved events + primitives to `extras/` which acts like Django's `contrib/`. They are standalone tools/libraries/helpers that are not necessarily part of the core, are on their own layer. They ship with the A-Frame distribution to satisfy the **batteries included** philosophy.
- Removed templates/HTMLImports.

**Events:**

```html
<a-event name="mouseenter" target="#cube" property="material.color" value="red"></a-event>
```

- Scope reduction, promote `<a-event>` that takes any event name.
- Deprecate `<a-EVENTNAME>` a-event wrappers
- Deprecate `<a-event type>` to be `<a-event name>`
- Remove code about state, now only concerned with listening to an event, and running `setAttribute`s.
- Make it a lot simpler by not worrying about delegated events.

**Primitives:**

```html
<a-box color="red" width="5" material="shader: flat"></a-box>
```

- Named entities that map HTML attributes to component properties while still being able to accept components. Syntactic sugar and can be a little smart.
- Deprecate `<a-cube>` and `<a-model>`
- Was very DRY by making a meshl mapping mixin since most primitives were using some common props
- Unit tests


cvan:
woo!

src/elements/a-event/index.js
```diff
@@ -1,10 +0,0 @@
-module.exports = {
-  'a-click': require('./a-click'),
-  'a-event': require('./a-event'),
-  'a-hover': require('./a-hover'),
-  'a-load': require('./a-load'),
-  'a-mousedown': require('./a-mousedown'),
```
cvan: can't remove these; lots of folks use them. and they're not templates anyway. so keep them. this whole file in fact.

ngokevin: I was going to start recommending `<a-event name="click"...>` instead since that can support every event possible. We can put it in the `CHANGELOG` as a breaking change, and I think it's somewhat okay to break as it was undocumented. I will update examples to all use `<a-event>`.

cvan: that breaks _a lot_ of people's sites. can we have breaking changes happen in `1.0.0` or the next release?

at least make a deprecation if you intend to.

cvan: and if we're already supporting primitives, what's the harm in supporting a few nice helpers out of the box (e.g., `<a-click>`, etc.)?

ngokevin: I had edited my comment to say I'll add a layer of deprecation.

The helpers add nothing on top of `<a-event>`. They can actually limit developers to think they can only handle a limited number of events, when actually they are able to handle any event. Plus, some of them are misleading because they don't map to the actual event name.

ngokevin: If we release physics and entities emit a collide event, people might try to do `<a-collide>` when that doesn't exist. Better to accustom people to handle any event via `<a-event name="collide">`.

cvan: > I had edited my comment to say I'll add a layer of deprecation.

cool, thanks

> The helpers add nothing on top of `<a-event>`.

it might be slight, but being able to visually scan and see the event name helps a ton with readability, not to mention a couple fewer characters. think of it like `<div>` versus `<span style="display: block">`.

> They can actually limit developers to think they can only handle a limited number of events, when actually they are able to handle any event. 

> Plus, some of them are misleading because they don't map to the actual event name.

just `<a-hover>`; you can remove that one if you want w/o deprecation - haven't seen anyone use that one (but if you want to be safe keep it).

ngokevin: Yeah, saves a few characters, but it's not worth losing the clarity of `<a-event name="XXX">` (tells you that it is an event handler, and gives the exact event name). Anyways, we can remove next release.

ngokevin:
@cvan @dmarcos 

This is roughly ready for review. I'll need to a take a couple more passes and add in a couple more primitives, but it's mostly there. All the examples work including new primitives and declarative events.

@cvan I took the liberty of making the declarative events code a lot simpler, decreasing the file to about a third. It doesn't do delegated events because I feel the people that will use this are those that are not using JS and don't have to worry about elements being injected later. Will give us a lot less code to maintain.

Rest of notes are in the description.

package.json
```diff
@@ -25,7 +25,9 @@
   "dependencies": {
     "browserify-css": "^0.8.2",
     "debug": "^2.2.0",
+    "deep-assign": "^2.0.0",
     "document-register-element": "^0.5.2",
+    "dot-prop": "^2.2.0",
```
ngokevin: not used

ngokevin:
I am thinking of getting rid of the transform logic to simplify things. The only one we have is for `src` to wrap `url()`. After doing some work on assets, I am not sure if we want that because:

- We want to encourage devs to utilize `<a-assets>` for better performance. That means being able to pass in a selector *or* a URL.
- Makes everything a lot simpler and predictable by having the mapping be one-to-one; whatever is set as a primitive attribute is passed directly to the component property.
- For the editor, we'd be able to two-way bind the primitive attribute to the component property, or update the primitive attribute if the component property changes. Currently, we cannot do that because of the transform function only goes one way.
- Least importantly, less code to maintain.

cvan:
> I am thinking of getting rid of the transform logic to simplify things. The only one we have is for `src` to wrap `url()`. After doing some work on assets, I am not sure if we want that because:

can you open a separate issue for that? it'd probably be best to not change too much surface area in this switch-a-roo.

> - We want to encourage devs to utilize `<a-assets>` for better performance.

I like what we're doing with improving asset loading. but it should go without saying that loading everything ahead of time does not necessarily mean _better_ performance. yes, the user experience is most likely.

the more I think about it, the more I think perhaps I think we should not be reinventing browsers' [Resource Hints](https://www.w3.org/TR/resource-hints/): resource prefetching, preloading, and prerendering.

not to mention we now have [Service Workers](https://serviceworke.rs/) which can give developers precise control over when resources (assets) should be loaded, cached, invalidated, fetched from cache or refetched, etc.

Firefox's `<link rel="prefetch">` is unique in that it uses [idle time to prefetch resources](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ).

I like [this Service Worker example](https://serviceworke.rs/fetching/).

``, `<audio>`, and `<image>` elements aside, you could imagine some declarative way of defining which assets should be preloaded and which assets should be cached.

for preloading, if you don't want to have developers write their own Service Workers (or have A-Frame provide some way of declaratively registering SWs), you could use a `Promise` chain of `fetch`s, which unlike XHR lets you fetch a cross-origin resource and get back an "opaque" response, which means you can't introspect it but it's been fetched.

I don't think we can require `<a-assets>` to always block. blocking the initial scene loading and rendering on assets causes bandwidth, time, causes redundant requests - all of which directly degrade the user experience every time the scene is loaded for the first time.

there are also the cases of ordering, timeouts, sync, async, defer - these are all things that ought to be considered and delegated control of to developers.

https://github.com/youmustfight/aframe-asset-lazy-load does some of these things, fwiw.

> That means being able to pass in a selector *or* a URL.

I'd say that's what explicit preloading is for.

dmarcos:
@cvan do you think this is ready to go?

cvan:
@ngokevin can you rebase and I'll test locally?

ngokevin:
Rebased

dmarcos:
@cvan if this is ok to you we're ready to merge.

cvan:
hey, sorry, folks - didn't get a chance to review this yesterday. next on my todo list today. thx!

cvan:
thanks for doing this. I noticed several issues with most of the examples though. could you take a look?

<hr>

this example is still broken:

[http://localhost:9000/examples/showcase-shopping/](http://localhost:9000/examples/showcase-shopping/)

<hr>

[http://localhost:9000/examples/primitives-video/](http://localhost:9000/examples/primitives-video/) does not load (but does on `dev`)

[http://localhost:9000/examples/animation-aframe-logo/](http://localhost:9000/examples/animation-aframe-logo/) does not load (but does on `dev`, albeit with the camera messed up; there's a JS error)

[http://localhost:9000/examples/animation-arms/](http://localhost:9000/examples/animation-arms/)
[http://localhost:9000/examples/animation-plane-reveals/](http://localhost:9000/examples/animation-plane-reveals/) (animation does not work; JS errors)
[http://localhost:9000/examples/animation-generic-logo/](http://localhost:9000/examples/animation-generic-logo/) (does not load at all; JS error)
[http://localhost:9000/examples/animation-translates/](http://localhost:9000/examples/animation-translates/) (does not load at all; JS error)
[http://localhost:9000/examples/animation-unfold/](http://localhost:9000/examples/animation-unfold/) (animation is wrong; JS errors)
[http://localhost:9000/examples/animation-warps/](http://localhost:9000/examples/animation-warps/) (animation is wrong; JS errors)
[http://localhost:9000/examples/interaction-cursor/](http://localhost:9000/examples/interaction-cursor/) (does not load at all; JS error)
[http://localhost:9000/examples/interaction-lookat/](http://localhost:9000/examples/interaction-lookat/) (does not load at all; JS error)
[http://localhost:9000/examples/interaction-hovers/](http://localhost:9000/examples/interaction-hovers/) (does not load at all; JS error)

[http://localhost:9000/examples/primitives-defaults/](http://localhost:9000/examples/primitives-defaults/) (because the `src` is not blank, the entities appear fine; see related issue though: #944)
[http://localhost:9000/examples/primitives-images/](http://localhost:9000/examples/primitives-images/) (controls do not work; JS errors)
[http://localhost:9000/examples/primitives-lookat/](http://localhost:9000/examples/primitives-lookat/) (does not load at all; multiple JS errors)
[http://localhost:9000/examples/primitives-models/](http://localhost:9000/examples/primitives-models/) (no controls work; multiple JS errors)
[http://localhost:9000/examples/primitives-planes/](http://localhost:9000/examples/primitives-planes/) (look controls do not work; multiple JS errors)
[http://localhost:9000/examples/primitives-sky/](http://localhost:9000/examples/primitives-sky/) (look controls do not work; JS error)
[http://localhost:9000/examples/primitives-video/](http://localhost:9000/examples/primitives-video/) (does not load at all; JS errors)
[http://localhost:9000/examples/primitives-videosphere/](http://localhost:9000/examples/primitives-videosphere/) (look controls do not work; JS error)
[http://localhost:9000/examples/declarative-events-cursor/](http://localhost:9000/examples/declarative-events-cursor/) (does not load at all; JS error)
[http://localhost:9000/examples/declarative-events-hovers/](http://localhost:9000/examples/declarative-events-hovers/) (does not load at all; JS error)

<hr>

[http://localhost:9000/examples/primitives-sky/](http://localhost:9000/examples/primitives-sky/) is inverted:

## `dev`
screenshot 2016-02-05 02 30 20

## `declarative-events`
screenshot 2016-02-05 02 29 45

cvan:
could we keep the event delegation? I think that's important.

cvan:
If this were a major release, we could remove (i.e., break) functionality, but it's not. So I really don't think we can or should. Especially without warning.

> - Deprecate `<a-EVENTNAME>` a-event wrappers

this breaks a lot of scenes

> - Deprecate `<a-event type>` to be `<a-event name>`

same

> - Make it a lot simpler by not worrying about delegated events.

I *really* think we should. it's the right thing to do. and it's less expensive.

> - Deprecate `<a-cube>` to be `<a-box>`

this is a huge breaking change

> - Various attributes have been renamed to match the component property name.

again, breaking

ngokevin:
Can you make sure your server is serving up-to-date files? The examples are working:

screen shot 2016-02-05 at 10 22 18 am

screen shot 2016-02-05 at 10 21 26 am

ngokevin:
- I think we've talked about this. We're not going to follow semver. We're going to do what many big projects do and just use the "minor" version as the "major" version for simplicity. 0.x.0 will be a "major" version. It's coming after 3 months of rapid major changes, and many things are already expected to break.

- Deprecate means it will still work but with a warning. It's not going to break this version.

- Planning on keeping event delegation removed. It's a lot of code that targets no one and would break nothing if removed. People that are building static scenes don't have to worry about event delegation as they don't have JS inserting elements. People that are using JS would just use `addEventListener` as they insert children.

- I will address the component prop renames and the inverted X sky.

cvan:
if you can add "(fixes #191)" to one of these commit messages, that'd be great

package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "browserify-css": "^0.8.2",
     "debug": "^2.2.0",
+    "deep-assign": "^2.0.0",
```
cvan: if we're using this, want to remove the `object-assign` dependency and all its usage?

cvan:
I blew out my `node_modules` and reinstalled everything. The scenes seem to be better now heh.

For some reason, the [Shopping example](http://localhost:9000/examples/showcase-shopping/) is super scaled down in Chrome but appears fine in Firefox:

screenshot 2016-02-06 04 57 11

screenshot 2016-02-06 04 56 54

cvan:
additionally, most of the examples that I was previously having issues with are okay such as [http://localhost:9000/examples/primitives-video/](http://localhost:9000/examples/primitives-video/)

these ones I noticed still have issues (and there may be others too I missed):

[http://localhost:9000/examples/animation-plane-reveals/](http://localhost:9000/examples/animation-plane-reveals/)
animation of blue panel is wrong

[http://localhost:9000/examples/animation-translates/](http://localhost:9000/examples/animation-translates/)
animations completely busted

[http://localhost:9000/examples/template-sky/](http://localhost:9000/examples/template-sky/)
still busted and backwards (fwiw, fine on `dev`)

[http://localhost:9000/examples/model/](http://localhost:9000/examples/model/)
still very busted (though not caused by this PR; filed #942)

cvan:
DrawVR's [Equalizer scene](http://drawvr.com/equalizer/) does not load in Firefox/Chrome. Firefox doesn't throw any errors, but Chrome does:

screenshot 2016-02-06 05 20 10

Source:

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>Equalizer Ride WebVR Virtual Reality Experience A-Frame</title>
<meta name="author" content="Donovan Kraeker">
<meta name="description" content="Equalizer Ride WebVR virtual reality example website experience using Aframe by DrawVR">
<script src="http://localhost:9000/dist/aframe.js"></script>
<!-- <script src="http://drawvr.com/equalizer/aframe.min.js"></script> -->
</head>
<body>
 <a-assets>
  <a-mixin id="floor" geometry="primitive: plane; height: 10000; width: 10000"  position="0 3 -5" rotation="-90 0 0" material="shader: flat; src: url(images/cubes.jpg); repeat: 200 200"></a-mixin>
  <a-mixin id="mix" material="shader: flat; side: double; src: url(images/mixer.jpg); repeat: 8 1" geometry="primitive: cylinder; height: 400" position="0 198 -5"></a-mixin>
  <a-mixin id="music" sound="src: sounds/marble.mp3; autoplay: true; loop: true"></a-mixin>
  <a-mixin id="curve"radius="3000" position="0 180 -10" theta-length="360" segments-radial="360" height="4" rotation="90 0 0" opacity="0"></a-mixin>
 </a-assets>

 <a-scene>
 <!-- Hang On! -->
 <a-entity look-controls geometry="primitive: plane; height: 10; width: 50" position="3000 40 -50" material="shader: standard; transparent: true; side: both; src: url(images/logo.png)">
  <a-animation attribute="position" from"3000 40 -40" to="0 40 0" repeat="indefinite" ease="linear" begin="10000" dur="1"></a-animation>
 </a-entity>

<a-curvedimage mixin="curve">
 <a-animation attribute="rotation" from="0 0 0" to="0 359 0" repeat="indefinite" ease="linear" begin="3000" dur="90000"></a-animation>
  <a-entity camera position="3000 -150 0" look-controls="enabled: true" rotation="0 0 0" cursor-visible="enabled: false" wasd-controls="enabled: true; acceleration: 1000;">
  </a-entity>
</a-curvedimage>

<a-entity mixin="floor"></a-entity>
<a-entity mixin="music"></a-entity>
<a-entity mixin="mix" geometry="radius: 2900"></a-entity>
<a-entity mixin="mix" geometry="radius: 3100"></a-entity>

  <!-- Lighting and Background -->
  <a-sky src="images/sky.jpg" radius="5000"></a-sky>
 </a-scene>
<script>

var scene = document.querySelector('a-scene');

if (scene.hasLoaded) {
  load();
} else {
  scene.addEventListener('load', load);
}

function load () {
  document.querySelector('#sbutton').addEventListener('click',
  function () {
  document.querySelector('#ride').emit('start-button');
  setTimeout(function(){
  document.querySelector('#cursor').setAttribute('visible', 'false');
  }, 500);
  });
}

</script>
<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-56782327-24', 'auto');
  ga('send', 'pageview');

</script>
 </body>
</html>
```

src/extras/primitives/primitives/a-video.js
```diff
@@ -0,0 +1,23 @@
...
+    },
+    material: {
+      color: '#FFF',
+      shader: 'flat',
+      side: 'double',
+      transparent: true
+    }
+  },
+
+  mappings: {
```
cvan: why's `src` missing?

cvan: I assume you want to force people to create `<video>` elements in `<a-assets>`. but keep in mind that removing `src` is a breaking change. and not to mention that makes most of these primitives pretty useless.

ngokevin: No. It's part of the material mixin.

ngokevin:
The translates example works for me. Still polishing up this PR.

ngokevin:
re-r? @cvan

examples/showcase-anime-UI/index.html
```diff
@@ -8,18 +8,19 @@
   </head>
   <body>
     <a-scene>
```
cvan: [http://localhost:9000/examples/showcase-anime-UI/](http://localhost:9000/examples/showcase-anime-UI/) loads in Firefox but doesn't load in Chrome (does load on `dev`):

screenshot 2016-02-17 02 57 39
screenshot 2016-02-17 02 57 41
screenshot 2016-02-17 03 00 21

examples/showcase-curved-mockups/index.html
```diff
@@ -8,18 +8,19 @@
   </head>
   <body>
     <a-scene>
```
cvan: [http://localhost:9000/examples/showcase-curved-mockups/](http://localhost:9000/examples/showcase-curved-mockups/)

camera regressed and cursor missing:

screenshot 2016-02-17 03 18 01

screenshot 2016-02-17 03 08 28

examples/showcase-curved-mockups/index.html
```diff
@@ -8,18 +8,19 @@
...
+      <a-curvedimage src="../_images/ui/3.png" radius="5.7" theta-length="180" height="9"
+                     rotation="0 90 0" scale="1.2 1.2 1.2"></a-curvedimage>
+      <a-curvedimage src="../_images/mozvr.png" radius="5.7" theta-length="17" height=".36"
+                     opacity="0.2" rotation="0 250 0" position="0 2 0"></a-curvedimage>
+      <a-curvedimage src="../_images/ui/1.png" radius="5.7" theta-length="72" height="2.6"
+                     rotation="0 80 0" position="0 0.7 0" scale="0.6 0.6 0.6"></a-curvedimage>
+      <a-curvedimage src="../_images/ui/2.png" radius="5.7" theta-length="70" height="3.02"
+                     rotation="0 -130 0" scale="0.8 0.8 0.8"></a-curvedimage>
 
-      <a-camera fov="80" position="0 1.8 0" near="0.2" cursor-color="#33CCFF"></a-camera>
```
cvan: [http://localhost:9000/examples/showcase-curved-mockups/](http://localhost:9000/examples/showcase-curved-mockups/)

camera regressed and cursor missing:

screenshot 2016-02-17 03 18 01

screenshot 2016-02-17 03 08 28

examples/animation-generic-logo/index.html
```diff
@@ -19,7 +19,7 @@
           <a-animation attribute="scale" from="0 0 0" to="1 1 1" begin="500" dur="500" fill="both" easing="ease-out"></a-animation>
           <a-animation attribute="position" from="0 -0.5 0" to="0 0 0" begin="500" dur="500" fill="both" easing="ease-out"></a-animation>
           <a-animation attribute="rotation" from="0 -300 0" to="30 54 36" begin="500" dur="1200" fill="both" easing="ease-out"></a-animation>
-          <a-cube width="0.25" depth="0.25" height="0.25" color="#EF2D5E"></a-cube>
+          <a-box width="0.25" depth="0.25" height="0.25" color="#EF2D5E"></a-box>
```
cvan: [http://localhost:9000/examples/animation-generic-logo/](http://localhost:9000/examples/animation-generic-logo/)

Firefox vs. Chrome:

screenshot 2016-02-17 03 12 20
screenshot 2016-02-17 03 12 25

examples/animation-aframe-logo/index.html
```diff
@@ -8,106 +8,133 @@
...
+        <a-asset-item id="tree" src="../_models/tree1/tree1.dae"></a-asset-item>
+      </a-assets>
 
       <!-- Quasi-isometric camera -->
       <a-entity>
-        <a-animation attribute="position" from="0 5 0" to="0 0 0" dur="4000" easing="ease-out"></a-animation>
-        <a-camera id="camera" near="1000" far="4000" position="0 880 1290" fov="2.2" cursor-visible="false" rotation="-34 0 0" wasd- controls-enabled="false" look-controls-enabled="false"></a-camera>
+        <a-animation attribute="position" from="0 5 0" to="0 0 0" dur="4000"
+                     easing="ease-out"></a-animation>
+        <a-camera id="camera" near="1000" far="4000" position="0 880 1290" fov="2.2"
```
cvan: [http://localhost:9000/examples/animation-aframe-logo/](http://localhost:9000/examples/animation-aframe-logo/)

this was already busted on `dev` in both Chrome and Firefox:

screenshot 2016-02-17 03 22 48
screenshot 2016-02-17 03 22 32

and in this branch the Chrome one still is broken, but the Firefox one looks good (nice work):

screenshot 2016-02-17 03 11 46
screenshot 2016-02-17 03 11 44

examples/animation-translates/index.html
```diff
@@ -8,13 +8,13 @@
   </head>
   <body>
     <a-scene>
-
-      <a-camera position="0 0 8" cursor-visible="false"></a-camera>
+      <a-entity position="0 0 8">
```
cvan: [http://localhost:9000/examples/animation-translates/](http://localhost:9000/examples/animation-translates/)

this was already busted on `dev` in both Chrome and Firefox:

screenshot 2016-02-17 03 25 38
screenshot 2016-02-17 03 25 44

and in this branch the Chrome one still is broken, but the Firefox one looks good (nice work):

screenshot 2016-02-17 03 13 00
screenshot 2016-02-17 03 13 04

examples/animation-unfold/index.html
```diff
@@ -18,26 +18,26 @@
...
+                  <a-box position="0 -1 0" width="6" height="1" depth=".05" translate="0 -0.5 0" color="#93648D">
                     <a-animation attribute="rotation" from="-180 0 0" to="0 0 0" dur="1000" begin="1000" fill="both"></a-animation>
 
-                  </a-cube>
-                </a-cube>
-            </a-cube>
-          </a-cube>
-        </a-cube>
+                  </a-box>
+                </a-box>
```
cvan: indentation?

examples/showcase-shopping/index.html
```diff
@@ -8,50 +8,47 @@
   </head>
   <body>
     <a-scene>
```
cvan: [http://localhost:9000/examples/showcase-shopping/](http://localhost:9000/examples/showcase-shopping/)

this scene loads _much_ faster now. great work :+1:

<hr>

in Firefox, however, the sizes of the images appear off:

### [http://localhost:9000/examples/showcase-shopping/](http://localhost:9000/examples/showcase-shopping/)
screenshot 2016-02-17 03 34 40

### https://aframe.io/aframe/examples/showcase-shopping/
showcase_curved_mockups

<hr>

and also in this branch compare in Chrome vs. Firefox:

screenshot 2016-02-17 03 32 51
screenshot 2016-02-17 03 33 02

cvan:
this is looking quite fantastic. let me know when you've figured out the issues above.

ngokevin:
Okay, thanks. Fixed camera issues in Chrome. A bit awkward to fix without making primitives a first-class citizen.

I'd ignore small changes in placement of stuff in the shopping + mockups example. The cursor removal was intentional as well.

re-r?

ngokevin:
Wouldn't mind getting this out and doing everything else as follow-up. One-month monster PR :)

cvan:
> One-month monster PR :)

that's fine; if you don't want to address something now, please file issues. otherwise, corners are going to get cut and things will be forgotten.

ngokevin:
ofc, let's see what else we need to address

ngokevin:
btw, have no problem addressing big issues. just getting antsy to get this PR off my chest and into the wild

examples/primitives-boxes/index.html
```diff
@@ -0,0 +1,44 @@
...
+      <a-entity position="0 0 -7" rotation="0 45 0">
+
+        <a-box src="http://i.imgur.com/6qHKYGf.png" position="-5 0 -2" rotation="0 0 0" width="1" color="#F16745"></a-box>
+        <a-box position="0 0 0" rotation="0 0 0" height="2" color="#FFC65D"></a-box>
+        <a-box position="5 0 2" rotation="0 0 0" depth="5" color="#7BC8A4"></a-box>
+        <a-box position="5 5 -5" rotation="0 0 0" width="1" height="2" depth="22" color="#4CC3D9"></a-box>
+        <a-box position="1 2 1" rotation="0 0 0" width="1" height="1" depth="1" color="#93648D"></a-box>
+        <a-box position="-3 0 0" rotation="0 0 0" width="0.5" height="10" depth="0.5" color="#404040"></a-box>
+
+        <a-frame position="-2 5 4" scale="0.15 0.15 0.15" rotation="0 0 0" thickness="0.1" color="#999"></a-frame>
```
cvan: feel free to remove this and [that one](#diff-fd6c52500482671540e1f39d3870e749R34) if we're removing the `<a-frame>` template

examples/animation-generic-logo/index.html
```diff
@@ -19,7 +19,7 @@
...
           <a-frame scale="0.04 0.04 0.04"></a-frame>
```
cvan: screenshot 2016-02-18 11 48 56
screenshot 2016-02-18 11 48 43



since the `<a-frame>` template definition is removed, feel free to replace with this:

```html
<!-- an actual frame … yes, an a-frame ;-) -->
<a-entity scale="0.04 0.04 0.04">
  <a-box position="5 5 0" rotation="0 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="5 -5 0" rotation="0 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="-5 5 0" rotation="0 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="-5 -5 0" rotation="0 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>

  <a-box position="0 5 5" rotation="0 90 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="0 5 -5" rotation="0 90 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="0 -5 5" rotation="0 90 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="0 -5 -5" rotation="0 90 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>

  <a-box position="5 0 5" rotation="90 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="5 0 -5" rotation="90 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="-5 0 5" rotation="90 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
  <a-box position="-5 0 -5" rotation="90 0 0" width="0.1" height="0.1" depth="10" color="#404040"></a-box>
</a-entity>
```

cvan:
* [http://localhost:9000/examples/showcase-anime-UI](http://localhost:9000/examples/showcase-anime-UI) loads in Chrome now :+1: 
* [http://localhost:9000/examples/animation-aframe-logo/](http://localhost:9000/examples/animation-aframe-logo/) looks correct in both browsers :+1: 
* [http://localhost:9000/examples/animation-generic-logo/](http://localhost:9000/examples/animation-generic-logo/) looks correct in both browsers :+1:
* [http://localhost:9000/examples/animation-translates/](http://localhost:9000/examples/animation-translates/)looks correct in both browsers :+1:
* [http://localhost:9000/examples/showcase-curved-mockups](http://localhost:9000/examples/showcase-curved-mockups) looks correct in both browsers :+1: 
* [http://localhost:9000/examples/showcase-shopping/](http://localhost:9000/examples/showcase-shopping/)looks correct in both browsers :+1: 
* http://drawvr.com/tron/ loads with new code :+1:
* http://drawvr.com/equalizer/ loads with new code :+1:
* [http://localhost:9000/examples/animation-generic-logo/](http://localhost:9000/examples/animation-generic-logo/) is missing the frame (replace with [this code](https://github.com/aframevr/aframe/pull/883/files#r53371969)) :name_badge:

ngokevin:
ok, fixed the a-frames. i think we can add features to `<a-mixin>` to make making those frames easier

examples/animation-generic-logo/index.html
```diff
@@ -8,19 +8,35 @@
...
-
         <a-entity id="box">
           <a-animation attribute="scale" from="0 0 0" to="1 1 1" begin="500" dur="500" fill="both" easing="ease-out"></a-animation>
           <a-animation attribute="position" from="0 -0.5 0" to="0 0 0" begin="500" dur="500" fill="both" easing="ease-out"></a-animation>
           <a-animation attribute="rotation" from="0 -300 0" to="30 54 36" begin="500" dur="1200" fill="both" easing="ease-out"></a-animation>
-          <a-cube width="0.25" depth="0.25" height="0.25" color="#EF2D5E"></a-cube>
-          <a-frame scale="0.04 0.04 0.04"></a-frame>
+          <a-box width="0.25" depth="0.25" height="0.25" color="#EF2D5E"></a-box>
+
+          <a-entity scale="0.04 0.04 0.04">
```
cvan: lol you removed my comment

cvan:
# R+

everything looks great and works well.

thanks so much for jumping on this! this merge deserves some celebration! :tada: :balloon:

examples/animation-generic-logo/index.html
ngokevin: the "omg it's a-frame in a-frame!" comment? doesn't make much sense now does it :)

ngokevin:
@dmarcos i may be merging some code in your absence, we can do more code review afterwards.

ngokevin:
thanks @cvan! i feel much better now.

PR #884, opened 2016-01-22
fix wakelock on Android

cvan:
Wakelock worked fine in WebKit on iOS because we never entered fullscreen (because WebKit doesn't support the Fullscreen API, heh).

Wakelock never worked on Android because we would (1) enable Wakelock, (2) enter fullscreen, (3) fullscreen event listener gets called, and (4) Wakelock is accidentally disabled.

### DIFF
--- a/src/core/a-scene.js
+++ b/src/core/a-scene.js
@@ -168,8 +168,8 @@ var AScene = module.exports = registerElement('a-scene', {
           if (!fsElement) {
             this.showUI();
             this.setMonoRenderer();
+            if (this.wakelock) { this.wakelock.release(); }
           }
-          if (this.wakelock) { this.wakelock.release(); }
         }
         document.addEventListener('mozfullscreenchange',
                                   fullscreenChange.bind(this));


vendor/wakelock/wakelock.js
```diff
@@ -29,8 +29,8 @@ function AndroidWakeLock() {
...
+      // Base64 version of videos_src/no-sleep-120s.mp4.
```
cvan: not sure why, but why isn't just a 1-second video used and looped? do you know, @caseyyee?

caseyyee: Not sure.  Probably just because it's all in one place?    If we can require/package it, it will fix the issue with not working in FF as well.

caseyyee: Because there isn't any video data, i don't think that it really makes a difference if it's 1s or 60s.   I think 60s is good though -- prevents any hiccups from restarting/looping the video.

dmarcos:
Have we confirmed that this fixes the problem on Android?

cvan:
@dmarcos: I spent most of yesterday discovering and fixing mobile issues. and, yes, see the comment above where I discovered we were exiting wakelock immediately after enabling it. I have two other branches that (1) fix orientation messages for Android and (2) restructure things so we can emit events when VR mode is entered and exited - even on mobile. will open those later today.

I've been testing on various Android devices and my iPhone 6. I could use an iPhone 5 if anyone has one.

I'll remove the commit to bump wakelock; I'll handle in another PR since it's unrelated to the original issue of wakelock not working on Android.

ngokevin:
FYI https://github.com/aframevr/aframe/pull/776 introduces VR events.

cvan:
rr? @dmarcos 

-

@ngokevin: lol gd, looks like most of refactoring I did yesterday is moot with your changes. happy that you did it, but sad that I didn't know you were working on that. I thought you were working on a mechanism of simply hiding the UI. is there an issue on file for the huge entirety of PR #776? that's why we have issues on file.

ngokevin:
Sorry, I should've went into more detail and asked what you were working on to make sure we didn't collide. I added events because splitting out into components required some form of communication across the components.

dmarcos:
Tests not passing

cvan:
> Tests not passing

[Looks like the build was terminated?](https://travis-ci.org/aframevr/aframe/jobs/104209510)

ngokevin:
You can restart the build in Travis as I did this morning if the failures look unrelated or intermittent.

cvan:
> You can restart the build in Travis as I did this morning if the failures look unrelated or intermittent.

yep, I did 15 min ago and it's passing now

PR #885, opened 2016-01-22
fix animation regression, check if begin is a delay

ngokevin:


src/core/a-animation.js
```diff
@@ -285,7 +285,7 @@ module.exports.AAnimation = registerElement('a-animation', {
         utils.splitString(evts).forEach(function (evt) {
           el.addEventListener(evt, self.start);
         });
-        if (!evts) { el.addEventListener('play', this.start); }
+        if (!isNaN(evts)) { el.addEventListener('play', this.start); }
```
dmarcos: A comment here?

PR #891, opened 2016-01-24
When removing a component that is a behavior the component has to be …

dmarcos:
…removed from the list of behaviors instead of the tick function


ngokevin:
r+

PR #893, opened 2016-01-24
camera color on mobile browser.

sergioroja:


dmarcos:
Thanks for the PR! It's weird that the cursor doesn't work on mobile with a flat shading.If we change the shader to standard the cursor then will be affected by the lights in the scene and this is something you don't normally want. Let's investigate why the flat shading is not working for you.  Do you have a screenshot of the problem?

sergioroja:
cur

ngokevin:
This might be due to the Android bug we fixed by bumping three.js. The standard shader had some bugs on Android. It might work on the latest branch now.

sergioroja:
The bug is not fix

dmarcos:
@sergioroja Have you tried if the bug is fixed on the dev branch?

PR #894, opened 2016-01-25
Fixes issue when parsing/stringify component attributes of the follow…

### LINKED ISSUE #889: setAttribute not working for vec3 types
For the following schema on ```my-component```:
{ position: { type: 'vec3', default: '0 0 0' } }

The following does not work as expected
````javascript
el.setAttribute('my-component', 'position', { x: 0, y: 1, z: 0 });
````
What I see on the DOM is 
````javascript
<a-entity my-component="position: [object Object]"></a-entity>
```` 
Something in the parsing/serialization logic of the component is not working properly.

### PR DESCRIPTION
**Pull request by dmarcos**
…ing form position: '0 1 0'. It should return

position: { x: 0, y: 1, z: 0 } if the property is the type vec3 (fixes #889)

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -113,6 +113,8 @@ Component.prototype = {
   parse: function (value) {
     var typeName;
     var type;
+    var schema = this.schema;
+    var properties;
 
     if (isSingleProp(this.schema)) {
       typeName = this.schema.type;
@@ -121,7 +123,17 @@ Component.prototype = {
       return error(typeName + ' is not a valid type.');
     }
 
-    return objectParse(value);
+    properties = objectParse(value);
+
+    if (properties === null || typeof properties !== 'object') { return properties; }
+
+    Object.keys(properties).forEach(parseProperty);
+    return properties;
+
+    function parseProperty (key) {
+      if (!schema[key]) { return; }
+      properties[key] = schema[key].parse(properties[key]);
+    }
   },
 
   /**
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/core/component.js [hunk 1]
```diff
@@ -121,7 +122,17 @@ Component.prototype = {
       return error(typeName + ' is not a valid type.');
     }
 
-    return objectParse(value);
+    value = objectParse(value);
```
ngokevin: We can name this variable `properties`.

src/core/component.js
```diff
@@ -121,7 +122,17 @@ Component.prototype = {
...
+
+    if (value === null || typeof value !== 'object') { return value; }
+
+    Object.keys(value).forEach(parseProperty)
+    return value;
+
+    function parseProperty (key) {
+      if (!schema[key] || !schema[key].parse) { return; }
```
ngokevin: If the key is there, then it should always have a `parse` method as given by the schema processor.

src/core/component.js
```diff
@@ -121,7 +122,17 @@ Component.prototype = {
...
 
-    return objectParse(value);
+    value = objectParse(value);
+
+    if (value === null || typeof value !== 'object') { return value; }
+
+    Object.keys(value).forEach(parseProperty)
+    return value;
+
+    function parseProperty (key) {
```
ngokevin: `propertyName` fits well.

src/core/component.js
```diff
@@ -134,15 +145,28 @@ Component.prototype = {
   stringify: function (data) {
     var typeName;
     var type;
+    var dataResult;
```
ngokevin: maybe `processedData`

src/core/component.js
```diff
@@ -134,15 +145,28 @@ Component.prototype = {
...
+    if (typeof data !== 'object') { return data; }
+
+    dataResult = utils.extend({}, data);
+
+    Object.keys(dataResult).forEach(stringifyProperty)
+
+    return objectStringify(dataResult);
+
+    function stringifyProperty (key) {
+      if (!schema[key] || !schema[key].stringify) { return; }
```
ngokevin: `stringify` should always exist too

ngokevin:
thanks! r+wc+lint

src/core/component.js
```diff
@@ -110,18 +110,29 @@ Component.prototype = {
    * @param {string} value - HTML attribute value.
    * @returns {object} Component data.
    */
-  parse: function (value) {
+  parse: function (properties) {
```
ngokevin: Well, the name `value` is fine here since it is just a string.

src/core/component.js
```diff
@@ -110,18 +110,29 @@ Component.prototype = {
...
     if (isSingleProp(this.schema)) {
       typeName = this.schema.type;
       type = propertyTypes[typeName];
-      if (type) { return type.parse.call(this, value); }
+      if (type) { return type.parse.call(this, properties); }
       return error(typeName + ' is not a valid type.');
     }
 
-    return objectParse(value);
+    properties = objectParse(properties);
```
ngokevin: But once it's parsed, it turns into an object of properties.

dmarcos:
Shipping!

PR #901, opened 2016-01-26
expose registerPropertyTypes/propertyTypes (fixes #881)

### LINKED ISSUE #881: Expose `registerPropertyType`
Forgot to expose it in index.js

ngokevin:


dmarcos:
Why do we need to expose propertyTypes?

ngokevin:
Don't know, devs might want to register their own types for reusability. Reason not to?

dmarcos:
I'm worried about opening up the types to 3rd parties. Imagine that a library o component registers a type. Users don't have a way to differentiate if a type is a-frame native or defined by a 3rd party component. You might remove a component and suddenly things stop working because the type is not defined anymore.

ngokevin:
OK

Issue #779, opened 2016-01-13
Support loading screen while assets download and entities load


ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Friday Dec 18, 2015 at 22:45 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/718_

----

Per https://twitter.com/footedesign/status/677967278890221570

ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ngokevin](https://github.com/ngokevin)**
_Friday Dec 18, 2015 at 23:54 GMT_

----

depends on adding asset blocking to `<a-assets>` before it fires loaded.

then users than listen on `loaded` and manage loading screens however they want.

ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cvan](https://github.com/cvan)**
_Friday Dec 18, 2015 at 23:59 GMT_

----

@ngokevin but in the case of external URLs that get loaded from `material` component and then from `THREE.js`'s loaders, we can handle those today without changing making `<a-assets>` inert (which we could do, but then we'd have to enable some like `preload` boolean).

ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ngokevin](https://github.com/ngokevin)**
_Saturday Dec 19, 2015 at 00:00 GMT_

----

i think we intend if you don't put it in assets there is no preloading/blocking logic.

right now a-assets does nothing. we should enforce through a-assets because that is canonical, anyone can make a material component.

ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cvan](https://github.com/cvan)**
_Saturday Dec 19, 2015 at 00:02 GMT_

----

> i think we intend if you don't put it in assets there is no preloading/blocking logic.

[...]

Issue #367, opened 2016-01-11
Add events to know when assets (e.g., `<video>` and models) have loaded


cvan:
There are quite a few events: https://www.w3.org/TR/html5/embedded-content-0.html#media-controllers

We could let folks listen for the events on the `<video>` elements themselves. If the developer has created a `<video>` asset on the page, they could listen to that. But if we had to create the `<video>` ourselves, or if they want a more convenient way of querying the video, we could attach the `video` as a property to the `<a-entity>`. Developers could then do this, for example:

```js
document.querySelector('a-entity#frogs_video').video.onplaying = function () { … };
```

cvan:
Also worth referencing #40

cvan:
I know things have changed a lot, but as I last understood it, we block `<a-scene>` on rendering until everything in `<a-assets>` has loaded.

Templates in their current form (e.g., `<a-video>`) are exempt from this blocking.

@dmarcos, @ngokevin, do you have ideas on how/whether to continue to block on `<a-assets>`? and which types of events we may want to emit to give developers finer-grain controls over things such as videos?

ngokevin:
I don't think `<a-assets>` does anything special at the moment. The most reliable way would be having the developer create the video tag themselves.

There aren't any guarantees on adding events or states on the entities because that is up the component developer.

ngokevin:
Okay, we now have the `loaded` and `timeout` events on the assets element. We can explore providing more granular  events later on

PR #906, opened 2016-01-27
block render loop on assets (fixes #873, #367, #779)

### LINKED ISSUE #873: a-assets based preload mechanism
We can use Three.js XHR loader to prefetch anything included in the ```a-assets``` tag. This will block the render loop until everything has loaded.

ngokevin:
- Currently waits for img (or anything with `src`) to load, and video/audio to load if they had preload or autoplay. 
- Configurable timeout, defaults to 3 seconds. After timeout, assets will no longer block
- https://github.com/aframevr/aframe/pull/913 will follow to get blocking on models since we need to allow models to specify selector to asset.

### DIFF
--- a/examples/model/index.html
+++ b/examples/model/index.html
@@ -8,16 +8,21 @@
   </head>
   <body>
     <a-scene stats="true">
-      <a-entity position="0 0 15">
-        <a-entity camera look-controls wasd-controls></a-entity>
-      </a-entity>
+      <a-assets>
+        <a-asset-item id="crate-obj" src="models/crate/crate.obj"></a-asset-item>
+        <a-asset-item id="crate-mtl" src="models/crate/crate.mtl"></a-asset-item>
+        <a-asset-item id="monster" src="models/monster.dae"></a-asset-item>
+      </a-assets>
 
       <a-entity collada-model="url(models/monster.dae)"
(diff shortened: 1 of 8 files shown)


src/components/loader.js
```diff
@@ -24,11 +21,11 @@ module.exports.Component = registerComponent('loader', {
...
     var mtlUrl = parseUrl(data.mtl);
     var format = data.format;
+
     if (model) { el.removeObject3D('mesh'); }
-    if (!url) {
-      warn('Model URL not provided');
-      return;
-    }
+
+    if (!url) { return console.warn('Model `src` was not specified'); }
```
dmarcos: What is warn returning? I would keep the ```return;``` we had before.

dmarcos: Why are using anymore the debug library here?

ngokevin: I'm just going to use `console` for critical warnings/errors for now so that dev/users can actually see them without doing `localStorage.logs = 1` which isn't documented very well. Have an issue to expose warnings/errors but hide log messages.

ngokevin: I can put this in another PR

dmarcos: We should use ```debug.js``` and enable debug logs by default in a different PR. It will be annoying to chase down the console.log calls later on.

Issue #908, opened 2016-01-27
Move <a-assets> into <a-scene>


ngokevin:
For asset preloading and future mixin work, we need the assets to have a reference to the scene and not pollute other scenes.

ngokevin:
If we want shared assets between multiple scenes, which would be a lesser priority use case, we can think about that down the road.

PR #910, opened 2016-01-28
move <a-assets> within <a-scene>

ngokevin:
Set up work for preloading and later on mixin features.

### DIFF
--- a/examples/animation/index.html
+++ b/examples/animation/index.html
@@ -7,16 +7,16 @@
     <script src="../../dist/aframe.js"></script>
   </head>
   <body>
-    <a-assets>
-      <a-mixin id="planet" geometry="primitive: sphere; radius: 0.5">
-      </a-mixin>
-      <a-mixin id="orbit" attribute="rotation" to="0 360 0"
-                repeat="indefinite" easing="linear" dur="3000"></a-mixin>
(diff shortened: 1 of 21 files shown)


dmarcos:
With multiple scenes we might want to share assets between them. Imagine you divide multiple levels in multiple separate scenes. How are we going to handle that?

ngokevin:
I pre-commented on that in the issue: https://github.com/aframevr/aframe/issues/908#issuecomment-175910999

Some reasons why I would move `<a-assets>` into the `<a-scene>` in the meantime:

- Focus on getting good single scenes first, then maybe think about multiple scenes.
- Enables work on preloading assets and mixin stylesheets
- That's with the assumption that multiple levels would utilize multiple scenes/canvases. I don't think multiple scenes are necessary with that case.
- Global assets means asset/mixin scope pollution between scenes.
- Global assets raises questions on whether unrelated assets from one scene block another scene.
- Raises question on whether the scene blocks on the global assets or just the local one.
- More code to handle global assets for multiple scenes.
- Flexible choice as IIRC we currently don't have anyone using a single asset for multiple scenes, and if we wanted to down the road, we could always layer on extra logic for shared/global assets.

PR #914, opened 2016-01-29
fix component double-parsing (fixes #912)

### LINKED ISSUE #912: component properties sometimes getting parsed twice
We call `Component.parse` (which calls `schema.parsePropert(y/ies)`) before passing into `buildData` (which also calls `schema.parsePropert(y/ies)`). For property types that don't provide the same result when calling the parser two times in a row, this would break.

ngokevin:
This is blocking adding a property type for assets and the model work:

- Need to change Component to not call `parse` before doing `buildData` since that already does schema parsing. We used to call it because we needed the `style-attr`-parsing. Since we only need that, I changed `buildData` to do `style-attr` parsing for multi-prop schemas.
- Simplified Component.parse/Component.stringify to use the schema functions since it was doing the same thing the schema already handles.
- Bind property parse/stringify to component prop definition to give vec3 parser access to default value

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -1,14 +1,15 @@
 /* global HTMLElement */
 var debug = require('../utils/debug');
-var propertyTypes = require('./propertyTypes').propertyTypes;
 var schema = require('./schema');
 var styleParser = require('style-attr');
 var utils = require('../utils/');
 
 var parseProperties = schema.parseProperties;
 var parseProperty = schema.parseProperty;
-var isSingleProp = schema.isSingleProperty;
 var processSchema = schema.process;
+var isSingleProp = schema.isSingleProperty;
+var stringifyProperties = schema.stringifyProperties;
+var stringifyProperty = schema.stringifyProperty;
 var error = debug('core:register-component:error');
 
 var components = module.exports.components = {};  // Keep track of registered components.
@@ -42,7 +43,7 @@ var Component = module.exports.Component = function (el) {
   var scene;
 
   this.el = el;
-  this.data = buildData(el, name, this.schema, this.parse(elData), elData);
+  this.data = buildData(el, name, this.schema, elData);
   this.init();
   this.update();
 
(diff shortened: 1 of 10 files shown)


src/core/component.js
```diff
@@ -111,63 +112,27 @@ Component.prototype = {
...
-
-    Object.keys(processedData).forEach(stringifyProperty);
-
-    return objectStringify(processedData);
-
-    function stringifyProperty (key) {
-      if (!schema[key]) { return; }
-      processedData[key] = schema[key].stringify(processedData[key]);
-    }
+    if (isSingleProp(schema)) { return stringifyProperty(data, schema); }
```
dmarcos: There was an error returned here if the type was not a vaid one. If I do a type on a type name does it now fail silently?

dmarcos: I think if a schema specifies a ```type``` and the type is not a valid one we should return a loud error somewhere. I see people declaring a type as ```v3``` or ```vector3``` and not knowing how it's not working.

ngokevin: the schema module does property type lookup and warnings during the processing stage.

ngokevin: that already happens at https://github.com/aframevr/aframe/blob/dev/src/core/schema.js#L63

src/core/schema.js
```diff
@@ -59,43 +60,59 @@ function processPropertyDefinition (propDefinition) {
   }
 
   propType = propertyTypes[typeName];
-
   if (!propType) {
     warn('Unknown property type: ' + typeName);
```
dmarcos: This should be an error instead of a warning.

src/core/schema.js
```diff
@@ -59,43 +60,59 @@ function processPropertyDefinition (propDefinition) {
   }
 
   propType = propertyTypes[typeName];
-
   if (!propType) {
-    warn('Unknown property type: ' + typeName);
+    error('Unknown property type: ' + typeName);
     return propDefinition;
```
dmarcos: Do we still need the return if we trigger an error?

ngokevin: ok changed to throw rather than log

dmarcos:
travis is complaining

dmarcos:
This one is ready to go if we just fix the tests

ngokevin:
yeah trying to figure out a failure

ngokevin:
Okay, updated. The component attribute animation was running into `undefined` for the component property value. Also removed a test involving `removeAttribute(component, propName, propValue)` since that doesn't seem necessary.

PR #924, opened 2016-02-02
Preliminary system API

dmarcos:


ngokevin:
Fitting the classic definition of ECS, Should system type to component type be one-to-one? A system might want to register which component(s) it wants to manage. And it seems like the logic should be top down, from system to component, rather than component to system.

```js
registerSystem('basic-mesh-system', function () {
  components: ['geometry', 'material'],

  entities: [ // List of all entities that have both a geometry + materal. ],

  update: function () {},

  tick: function () {}
});
```

ngokevin:
With that way, the ```material``` system wouldn't have material components register themselves, they would listen for all entities that have a material component, and manage the logic. Something extremely roughly like:

```
registerSystem('basic-mesh-system', function () {
  components: ['geometry', 'material'],

  entities: [ // List of all entities that have both a geometry + materal. ],

  // System does the registering. Component does not need to talk up to the system.
  update: function (el, elComponents) {
    this.materials.push(elComponents.material);
  },

  tick: function () {}
});
```

dmarcos:
I expect each component to ship with its system if it has any. Tying a system with multiple components creates strange coupling between them. If two components require one specific system is an indication that that system should probably be an aframe core feature.

dmarcos:
A system provides services common to a specific type of component

ngokevin:
How much have we researched the *systems* part of ECS pattern? I think the way this is implemented, it's not true ECS. But it's more just EC with systems being something else.

I imagine we could register our own systems that manage inter-communication between entities and batched updates of groups of entities, but grouping entities of a single component might be very limiting. 

Also have to think about where we want the logic to be. In this PR and in EC, a lot of the behavior is put in the component (e.g., the material component has to tell the system to register itself), but in ECS, the logic is placed on the system.

ngokevin:
Just reading around. Systems would be a good place to be able to put all game/global logic.

http://www.gamedev.net/page/resources/_/technical/game-programming/understanding-component-entity-systems-r3013

dmarcos:
I don't care that much about what literature says or about theoretical purity. I just care about the work to be done. For certain components you want a centralize logic and state common to all of them. For physics engine for instance or caching systems for a variety of things. This is what I'm trying to achieve in this PR.

src/components/scene/fog.js
```diff
@@ -30,7 +30,7 @@ module.exports.Component = register('fog', {
     // (Re)create fog if fog doesn't exist or fog type changed.
     if (!fog || data.type !== fog.name) {
       el.object3D.fog = getFog(data);
-      el.updateMaterials();
+      el.systems.material.updateMaterials();
```
ngokevin: system or systems?

src/core/component.js
```diff
@@ -9,7 +9,7 @@ var processSchema = schema.process;
 var isSingleProp = schema.isSingleProperty;
 var stringifyProperties = schema.stringifyProperties;
 var stringifyProperty = schema.stringifyProperty;
-
+var systems = require('./system');
```
ngokevin: should go with rest of requires

src/core/system.js
```diff
@@ -0,0 +1,87 @@
+var components = require('./component');
+var systems = module.exports.systems = {};  // Keep track of registered components.
+
+/**
+ * System class definition.
+ *
+ * Systems are pieces of data and logic that are associated to one particular
+ * component. They're global to all the instantiated components of that type
+ * and provide an API for them to use. An example would be a physics component
+ * that would contain a physic system that oversees all the entities and
```
ngokevin: physics

src/core/system.js
```diff
@@ -0,0 +1,87 @@
...
+ * System class definition.
+ *
+ * Systems are pieces of data and logic that are associated to one particular
+ * component. They're global to all the instantiated components of that type
+ * and provide an API for them to use. An example would be a physics component
+ * that would contain a physic system that oversees all the entities and
+ * calculate the necessary updates based on the entities interactions.
+ *
+ * @namespace System
+ * @member {Element} sceneEl - handle to the scene element where the system applies to
```
ngokevin: capitalize/period

src/core/system.js
```diff
@@ -0,0 +1,87 @@
...
+   */
+  play: function () { /* no-op */ },
+
+  /**
+   * Called to stop any dynamic behavior (e.g., animation, AI, events, physics).
+   */
+  pause: function () { /* no-op */ },
+
+  /**
+   * Remove handler. Shutsdown the system
```
ngokevin: ```Shuts down the system.```

src/core/system.js
```diff
@@ -0,0 +1,87 @@
...
+  // Format definition object to prototype object.
+  Object.keys(definition).forEach(function (key) {
+    proto[key] = {
+      value: definition[key],
+      writable: true
+    };
+  });
+
+  if (systems[name]) {
+    throw new Error('The system "' + name + '" has been already registered');
```
ngokevin: maybe be more consistent with component error:

```
    throw new Error('The system `' + name + '` has been already registered. ' +
                    'Check that you are not loading two versions of the same system ' +
                    'or two different systems of the same name.');
```

ngokevin:
Are you going to do a light system and camera system?

ngokevin:
r+wc if you gonna do the light/camera systems later

dmarcos:
Created issues to add a light and camera systems:

https://github.com/aframevr/aframe/issues/1079

https://github.com/aframevr/aframe/issues/1080

PR #936, opened 2016-02-05
remove unused VRLoader (fixes #935)

### LINKED ISSUE #935: Remove unused VRLoader
https://github.com/aframevr/aframe/blob/dev/src/lib/VRLoader.js

cvan:
it was awesome, but we're not using it


ngokevin:
lgtm

cvan:
actually there's postMessage code I can remove in the scene code

cvan:
@dmarcos @ngokevin: there's actually code [here](https://github.com/aframevr/aframe/blob/dev/src/core/scene/iframe.js) that handles receiving messages from the parent window and sending messages too.

should I remove that, or bring back VRLoader? it's a good hack, but it's for Firefox only. I'd rather folks wait for true link traversal in new API. so can we agree to remove `iframe.js`?

PR #950, opened 2016-02-09
import 0.1.0 docs from A-Frame site (fixes #948)

### LINKED ISSUE #948: Import 0.1.0 docs from aframe.io site
See aframevr/aframe-site#121 and aframevr/aframe-site#123

cvan:
Related: aframevr/aframe-site#121 and aframevr/aframe-site#123


dmarcos:
Are we going to have a different directory for each version of the documentation?

ngokevin:
That's the plan

dmarcos:
It seems and anachronistic to use directories for different version in the age of version control software. There's not a cleaner and more elegant way to handle this? I don't want to see a forever growing directory with tons of redundant information.

ngokevin:
hm, that is a good point. we could just use github tags

cvan:
> hm, that is a good point. we could just use github tags

I thought about that, but then I realized we constantly make changes to the docs: typos, corrections, more examples.

what would you do then? delete the tag and create a new one with the same name?

ngokevin:
maybe. a bit of friction but i imagine the majority of changes will be on the latest version of the docs. don't imagine we're going to prioritize the old version of docs

ngokevin:
thinking about it more, directories seems like a simpler solution.

cvan:
though it sounds ideal, I really don't think tags are going to work out. we've already seen too many typos and corrections and additions that had to be made for `0.1.0` docs.

ngokevin:
Going to wait for @dmarcos to come back, but I'm onboard with each folder having its own docs, and having a `current` folder for current version docs.

cvan:
can we merge this? r? @dmarcos + @ngokevin?

dmarcos:
I don't think having a directory for version is a good idea. What happens if you find a typo? You have to go every single directory to fix it? We have to find a better way to handle this

ngokevin:
Then you ignore it, it's just a typo. The alternative is to go into every tag, commit, delete the tag, push the tag.

ngokevin:
related discussion https://github.com/aframevr/aframe/pull/988#issuecomment-185047446

ngokevin:
I'm also okay with just using git tagging, I'm not very concerned with maintaining old versions closely. Optimize for just showing newest docs on the site and in the repo. Could go either way, though it looks cleaner using git.

dmarcos:
Tagging is the way to go IMO. As I said, using a directory per version is like pre CVS times when people had copies of files with the version appended to it. myFile_ver2, myFile_ver3. We will end up with a lot of redundant information in the repo that will lead to confusion.

ngokevin:
@cvan I'm fine with tagging. I don't expect too much going back in time. You cool with that?

cvan:
> Tagging is the way to go IMO. As I said, using a directory per version is like pre CVS times when people had copies of files with the version appended to it. myFile_ver2, myFile_ver3. We will end up with a lot of redundant information in the repo that will lead to confusion.

yep, tags are fine. manually, it's a bit of a pain. but it'll keep things clean when viewing a single version/checkout at a time.

fwiw, folks using RTD seem to use tags:

http://docs.wagtail.io/en/latest/
http://docs.wagtail.io/en/v1.3.1/
http://docs.wagtail.io/en/v1.2/
https://github.com/torchbox/wagtail/tree/v1.2/docs

if a typo appears in our first version (tag `v0.1.0`), we have to update every single tag (`v0.1.0`, `v0.1.1`, `v0.1.2`, `v0.1.3`, `v0.2.0`) to make sure it's fixed in all them. that's undeniably way more work than editing the files in one directory and not having to retag things.

let's go with tags though. I'm cool with that. but just making folks aware it's more work to maintain for sure.

ngokevin:
Not worth for a typo, but let's see how often we have critical changes to docs that we need to retroactively patch. Yeah, aware that in the (I expect) rare case we do need to, that it will be a bit more commands to run. If it becomes a nightmare, we can re-eval.

PR #961, opened 2016-02-11
fix obj-model component for OBJ-only, test obj-model (fixes #959)

### LINKED ISSUE #959: obj-model component errors when loading just an obj

ngokevin:
- Was accidentally using `this` in obj load callback.
- obj-model-loaded event

Related necessary changes:

- For now, don't run attributeChangedCallback setEntityAttribute if newVal is an object.
- Fix component dependency initialization.


src/core/a-entity.js
```diff
@@ -461,22 +467,23 @@ var proto = Object.create(ANode.prototype, {
...
       var oldValue;
+      var serializedValue = value;
 
       if (component) {
         if (typeof value === 'string' && componentPropValue !== undefined) {
           // Update currently-defined component data with the new property value.
           partialComponentData = self.getAttribute(attr) || {};
           partialComponentData[value] = componentPropValue;
-          value = partialComponentData;
+          deserializedValue = partialComponentData;
```
dmarcos: Just reading the code I feel friction trying to remember: Is serialized a string and deseralized an object or the other way around? Can we call them valueObj / valueStr ?

ngokevin: **s**erialized a **s**tring, deserialized is an obj. i think it's a common term in APIs

 i changed it because `valueStr` told me it was a string, but knowing the type wasn't telling me what it was supposed to be used for or what it represented.

ngokevin: Could do `serializedValueStr` and `deserializedValueObj`?

dmarcos: It's a common term but doesn't quickly parse in your brain. You have to think and it can be error prone if you are not 100% focused. Would you prefer JSON.serialize / JSON.deseralize instead of JSON.stringify / JSON.parse? What about componentString / componentObj or dataStr or dataObj?

ngokevin: don't mind either actually, they're both descriptive. not sure about component/data since it's a bit vague and overloading.

ngokevin: just want to convey that one value is parsed, and the other value is for serializing to the DOM

ngokevin: not ideal for me, but i can just keep the var name to what you prefer, and then i'll add comments for myself

tests/helpers.js
```diff
@@ -6,12 +6,21 @@
...
   var scene = document.createElement('a-scene');
   var assets = document.createElement('a-assets');
   var entity = document.createElement('a-entity');
   scene.appendChild(assets);
   scene.appendChild(entity);
+
+  opts = opts || {};
+
+  if (opts.assets) {
+    opts.assets.forEach(function (asset) {
```
dmarcos: Can you do ```opts.assets.forEach(assets.appendChild)```?

ngokevin:
forEach + assets.appendChild doesn't work for the tests. Probably because it passes the loop index as 
the second arg.

@dmarcos re-r?

tests/helpers.js
dmarcos: we can still make this simpler as per my previous comment

ngokevin: we cannot: https://github.com/aframevr/aframe/pull/961#issuecomment-187446864

ngokevin:
re-r? ^^

src/core/a-entity.js
```diff
@@ -264,15 +270,15 @@ var proto = Object.create(ANode.prototype, {
...
-      } else {
-        if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
-        }
-        this.components[name] = new components[name].Component(this);
-        if (!this.paused) { this.components[name].play(); }
       }
 
+      if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
```
dmarcos: Why is not part of the else clause anymore?

ngokevin: it was a bug in the component dependency system. the material dep of the obj-model wasn't getting init before obj-model

ngokevin: You could see that if it was a dependency, the component instantiation never happened.

dmarcos: This should init the material dependency: 
https://github.com/aframevr/aframe/pull/961#discussion-diff-53723787L262
What does this line change exactly?

ngokevin: I don't know the exact cause, but the `setAttribute(name, '')` wasn't working as expected. The `obj-model` component started initializing before its `material` component dependency. After this change, the load order was correct.

dmarcos: We need to understand why or it will cause problems down the road. Programming by coincidence is bad.

dmarcos: the ```initComponentDependencies``` call from above should properly initialize the material. If that's not the case there's a bug in the depenency initialization logic.

ngokevin: ok i'll check it out

ngokevin:
@dmarcos I've investigated and found that initializing dependencies using `setAttribute` didn't seem to work because in `setEntityAttribute`, we don't initialize the component if the entity hasn't loaded? Will look more, but at the moment, the current solution of not initializing dependency component directly rather than roundabout is easier to reason.

ngokevin:
OK, I think we can't initialize dependency components using `setAttribute`. If an entity has not yet loaded (e.g., not appended to scene), `setEntityAttribute` will block calls by checking `!this.hasLoaded`. So we need to initialize component dependencies directly in order to handle the case that the entity has not loaded:

```
el.setAttribute('obj-model', '');  // Material dependency wouldn't get initialized because entity wasn't loaded.
el.addEventListener('loaded', function () {
  // ...
});
```

I will try to add a direct unit test for this.

src/core/a-entity.js
```diff
@@ -50,8 +49,14 @@ var proto = Object.create(ANode.prototype, {
...
   },
 
+  /**
+   * Handle changes coming from the browser DOM inspector.
+   */
   attributeChangedCallback: {
     value: function (attr, oldVal, newVal) {
+      // TODO: Remove this return once we bump DRE polyfill. The DRE polyfill is currently
+      // calling our wrapped AEntity.getAttribute which sometimes returns a parsed object.
+      if (typeof newVal === 'object') { return; }
```
dmarcos: The shaders PR has landed and we now use a fork of the registerElement polyfill that makes this change unnecessary

src/core/a-entity.js
```diff
@@ -174,7 +179,7 @@ var proto = Object.create(ANode.prototype, {
   add: {
     value: function (el) {
       if (!el.object3D) {
-        error("Trying to add an object3D that doesn't exist");
+        throw Error("Trying to add an element that doesn't have an `object3D`");
```
dmarcos: this has to be ```throw new Error("Trying to add an element that doesn't have an `object3D`");```

src/core/a-entity.js
```diff
@@ -262,18 +267,17 @@ var proto = Object.create(ANode.prototype, {
...
-      } else {
-        if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
-        }
-        this.components[name] = new components[name].Component(this);
-        if (!this.paused) { this.components[name].play(); }
+        // Expose dependency component in the DOM.
+        HTMLElement.prototype.setAttribute.call(this, name, '');
+      }
```
dmarcos: Why is the else statement gone? Is there a case where we want to call ```setAttribute``` twice?

ngokevin: Did a bunch of debugging, the indirect use setAttribute to set up dependency components doesn't work. https://github.com/aframevr/aframe/pull/961#issuecomment-190437392

PR #984, opened 2016-02-16
fixes look controls in chrome (Fixes #983)

### LINKED ISSUE #983: Look controls not working in Chrome (dev branch)
Cannot read property 'addEventListener' of undefined error using Chrome (dev)

image

caseyyee:
Fixes https://github.com/aframevr/aframe/issues/983

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -69,7 +69,14 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   addEventListeners: function () {
-    var canvasEl = this.el.sceneEl.canvas;
+    var sceneEl = this.el.sceneEl;
+    var canvasEl = sceneEl.canvas;
+
+    // listen for canvas to load.
+    if (!canvasEl) {
+      sceneEl.addEventListener('render-target-loaded', this.addEventListeners.bind(this));
+      return;
+    }
 
     // Mouse Events
     canvasEl.addEventListener('mousedown', this.onMouseDown, false);
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -375,9 +375,11 @@ var AScene = module.exports = registerElement('a-scene', {
           this.resize();
 
           // Kick off render loop.
-          this.render();
-          this.renderStarted = true;
-          this.emit('renderstart');
+          if (this.renderer) {
+            this.render();
+            this.renderStarted = true;
+            this.emit('renderstart');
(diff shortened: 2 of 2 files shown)


src/components/look-controls.js
```diff
@@ -28,7 +28,11 @@ module.exports.Component = registerComponent('look-controls', {
 
   play: function () {
     this.previousPosition.set(0, 0, 0);
-    this.addEventListeners();
+
+    // defer adding listeners until everything has run.
```
ngokevin: what are we waiting on to run specifically? maybe we should use an event?

caseyyee: `this.el.sceneEl.canvas` is undefined.    The scene canvas isn't setup before the event listener runs.

I tried using `loaded` event, but for some reason it runs multiple times on scene.  Not sure why this is.

ngokevin: The `canvas` component will trigger `render-target-loaded` event on the scene. Want to try that?

caseyyee: Same issue with the `Cannot read property 'render' of undefined` error in the console in the same issue.   

If I checked `scene.hasLoaded` it's ` false`.   Resolves to `true` in FF.   Just a difference in the order in which the methods are called In Chrome?

caseyyee: I'll give that a shot.

caseyyee:
@ngokevin ?

ngokevin:
Okay, looks good. Gonna do some spot checks and wait for Travis.

src/core/scene/a-scene.js
```diff
@@ -450,7 +450,9 @@ var AScene = module.exports = registerElement('a-scene', {
           });
         }
 
-        this.renderer.render(this.object3D, camera);
+        if (this.renderer) {
```
ngokevin: is there a place we can do this check earlier?

caseyyee: lemme take a look.

src/core/scene/a-scene.js
```diff
@@ -375,9 +375,11 @@ var AScene = module.exports = registerElement('a-scene', {
           this.resize();
 
           // Kick off render loop.
-          this.render();
-          this.renderStarted = true;
-          this.emit('renderstart');
+          if (this.renderer) {
```
ngokevin: Might be worth filing an issue to see why renderer is not there at this point.

PR #992, opened 2016-02-17
waits for canvas to be loaded

caseyyee:
same as commit eb12979ce9ddb291aeb9e617b2121b949d6060b6, but for cursor.


caseyyee:
It seems this is the only place that this is necessary.   Missed it last time.

PR #1010, opened 2016-02-22
release AudioContext and remove listener, sound from scene when remove sound component

meta-meta:
`this.listener.context.close()` is all that's needed to free up the Audio Context and allow more sounds to be loaded when loading and removing sounds dynamically.

Seems like we want to remove the sound and listener from the scene as well to free resources. Does this happen automagically?


ngokevin:
lgtm

PR #1013, opened 2016-02-22
entity play on attach (fixes #1011)

### LINKED ISSUE #1011: injected entities (with animations) don't play
I'm injecting this with a templating engine:

```html
<a-mixin id="wing" geometry="primitive: box; depth: .005; height: 1;
                             translate: -0.25 0 0; width: 0.5"
                   material="color: red" rotation="0 0 0"
                   scale="1 1.25 1"></a-mixin>
<a-mixin id="flap" attribute="rotation" direction="alternate" fill="forwards"
                   repeat="indefinite"></a-mixin>

<a-entity position="0 2.5 -5">
  <a-animation attribute="rotation" to="0 360 0" fill="forwards" dur="3000"
               easing="linear" repeat="indefinite"></a-animation>
  <a-entity rotation="0 90 0" position="2.5 0 0">
    <a-entity rotation="-45 0 90">
      <a-entity mixin="wing">
        <a-animation mixin="flap" to="0 120 0"></a-animation>
      </a-entity>
    </a-entity>
    <a-entity rotation="-45 0 90">
      <a-entity mixin="wing">
        <a-animation mixin="flap" to="0 -120 0"></a-animation>
      </a-entity>
    </a-entity>
  </a-entity>
</a-entity>
```

the animation is not running. if i do entity.play(), it runs

ngokevin:
At the moment, entities only play since the scene calls play. We need to call play when an entity is attached after the scene is already running.

- [x] tests

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -61,6 +61,7 @@ var proto = Object.create(ANode.prototype, {
       this.addToParent();
       if (!this.isScene) {
         this.load();
+        if (!this.parentNode.paused) { this.play(); }
       }
     }
   },
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -69,16 +69,29 @@ suite('a-entity', function () {
     });
 
     test('waits for children to load', function (done) {
+      var scene = document.createElement('a-scene');
       var entity = document.createElement('a-entity');
       var entityChild1 = document.createElement('a-entity');
       var entityChild2 = document.createElement('a-entity');
       entity.appendChild(entityChild1);
       entity.appendChild(entityChild2);
-      document.body.appendChild(entity);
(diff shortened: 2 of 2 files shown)


dmarcos:
Travis is complaining

ngokevin:
updated

ngokevin:
easy r? @dmarcos I got some CodePen scenes and examples that depend on this fix

src/core/a-entity.js
```diff
@@ -61,6 +61,9 @@ var proto = Object.create(ANode.prototype, {
       this.addToParent();
       if (!this.isScene) {
         this.load();
+        if (!this.parentNode.paused) {
```
dmarcos: one liner

ngokevin:
updated

ngokevin:
thanks

PR #1016, opened 2016-02-23
Add modelloaded event.

donmccurdy:
Helpful to know when model assets have finished loading.

For example, to get @ngokevin's shadow component working with .OBJ models:

```javascript

module.exports = {
  schema: {
    cast:     { default: false },
    receive:  { default: false }
  },

  init: function () {
    this.update();
    this.el.addEventListener('modelloaded', () => this.update());
  },

  update: function (previousData) {
    var data = this.data;
    // Applied recursively to support imported models.
    this.el.object3D.traverse(function(node) {
      if (node instanceof THREE.Mesh) {
        node.castShadow = data.cast;
        node.receiveShadow = data.receive;
      }
    });
  },

  tick: function () {}
}
```

### DIFF
--- a/src/components/collada-model.js
+++ b/src/components/collada-model.js
@@ -25,6 +25,7 @@ module.exports.Component = registerComponent('collada-model', {
     loader.load(src, function (colladaModel) {
       self.model = colladaModel.scene;
       el.setObject3D('mesh', self.model);
+      el.emit('model-loaded', {format: 'collada', model: self.model});
     });
   },
 


donmccurdy:
Oh, looks like there's a very similar change in #961 — no particular preference about the event name.

ngokevin:
Yeah, might be organized to prepend the component name to the event.

I should probably make it past tense like yours though to be consistent with the rest of the codebase. Ideally, it would be present tense like how the rest of the web does it, but a bit too late now.

donmccurdy:
(was originally requested by https://github.com/donmccurdy/aframe-obj-loader-component/pull/2)

'\*load' or '\*loaded' both seem fine.

As far as prepending "obj" and "collada", the only hassle would be needing to listen for unique events from each different model format. Could also do `emit('modelloaded', {format: 'obj', model: object})` though.

ngokevin:
Hm, that works

ngokevin:
Want to change this PR to handle for collada? I can handle for the OBJ since I have that already

donmccurdy:
:+1:

ngokevin:
lgtm

dmarcos:
Thanks!

Issue #796, opened 2016-01-13
JS errors have a big delay from when they are thrown to when they are surfaced, or being eaten


ngokevin:
<a href="https://github.com/mattdesl"><img src="https://avatars.githubusercontent.com/u/1383811?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mattdesl](https://github.com/mattdesl)**
_Thursday Dec 24, 2015 at 18:36 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/750_

----

I'm noticing that when building a component, if I throw an error from the `init()` function, it will get swallowed and execution will stop.

ngokevin:
<a href="https://github.com/dmarcos"><img src="https://avatars.githubusercontent.com/u/39342?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dmarcos](https://github.com/dmarcos)**
_Friday Jan 08, 2016 at 14:48 GMT_

----

@cvan @ngokevin ideas?

ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ngokevin](https://github.com/ngokevin)**
_Friday Jan 08, 2016 at 14:58 GMT_

----

It doesn't get completely swallowed, but there is a long delay from when the error is thrown until the error displays in the console.

ngokevin:
<a href="https://github.com/dmarcos"><img src="https://avatars.githubusercontent.com/u/39342?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dmarcos](https://github.com/dmarcos)**
_Friday Jan 08, 2016 at 15:22 GMT_

----

@ngokevin What is the recommended path of action? Are promises slow? Is it the polyfill problem?

ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ngokevin](https://github.com/ngokevin)**
_Saturday Jan 09, 2016 at 02:36 GMT_

----

I think the issue title should be `Errors are being swallowed`? It doesn't seem to have anything to do with Promises.

Needs investigation on how to surface errors sooner.

ngokevin:
Actually, I am noticing errors within promises are completely swallowed.

PR #1026, opened 2016-02-25
Switched out Promise polyfill due to old one polyfilling chrome

JoshGalvin:
^ Both FF and Chrome will throw uncaught exceptions within promises. Unfortunately the old polyfill library was polyfilling Chrome even though it had native Promise support.


dmarcos:
Can you remove the dist file from the PR?

JoshGalvin:
Done.

JoshGalvin:
Fixes #796 btw.

package.json
```diff
@@ -27,7 +27,7 @@
     "debug": "^2.2.0",
     "deep-assign": "^2.0.0",
     "document-register-element": "^0.5.2",
-    "es6-promise": "^3.0.2",
+    "promise-polyfill": "^3.1.0",
```
dmarcos: What's the difference between the polyfills?

src/index.js
```diff
@@ -1,4 +1,9 @@
-require('es6-promise').polyfill();  // Polyfill `Promise`.
+// Polyfill `Promise`.
+var promise = require('promise-polyfill');
+if (!global.Promise) {
```
ngokevin: maybe ```global.Promise = global.Promise || require('promise-polyfill');```

src/index.js
```diff
@@ -1,4 +1,9 @@
-require('es6-promise').polyfill();  // Polyfill `Promise`.
+// Polyfill `Promise`.
+var promise = require('promise-polyfill');
```
dmarcos: Should we do 
````javascript
window.Promise = window.Promise || require('promise-polyfill');
````

Comment (ngokevin) [inline src/index.js]: first! https://github.com/aframevr/aframe/pull/1026#discussion_r54186419

File: src/index.js (hunk 2 above)
Comment (dmarcos) [inline src/index.js]: I don't think that it's even necessary. 
```javascript
require('promise-polyfill');
````
it should be enough. It seems the polyfill already attaches itself to the global scope if Promise is not defined:

https://github.com/taylorhakes/promise-polyfill/blob/master/Promise.js#L196

**Comment by JoshGalvin** [comment]
I prefer using global within the browserify environment as it maps to window and would still work in a node environment.

File: src/index.js [hunk 4]
```diff
@@ -1,4 +1,6 @@
-require('es6-promise').polyfill();  // Polyfill `Promise`.
+// Polyfill `Promise`.
+global.Promise = global.Promise || require('promise-polyfill');
```
dmarcos: We should use window instead of global for consistency as we do below in this file.

JoshGalvin: Roger, its not like this could ever really run in a node environment anyhow :gun:

JoshGalvin:
@dmarcos the other polyfill was checking "Object.prototype.toString.call(P.resolve()) === '[object Promise]'", which came out as [object Object] in Chrome. Technically you could still use the other library as long as we did the polyfill check ourselves.

dmarcos:
Ok. I had no idea what the difference was.

dmarcos:
Waiting for green!

JoshGalvin:
Its kinda funny because we only need this polyfill to support IE http://caniuse.com/#feat=promises

dmarcos:
AFrame works on IE and almost on windows phone. There's a problem with the values we read from the device orientation API that we will fix on next release.

dmarcos:
Thanks! Congratulations! You're the 8th contributor. I guess we have to order t-shirts soon to give away :). Ideas for a awesome designs are more than welcome.

package.json
cvan: I'm also curious. as long as it's A+ compliant, that's legit.

src/index.js
```diff
@@ -1,4 +1,6 @@
-require('es6-promise').polyfill();  // Polyfill `Promise`.
+// Polyfill `Promise`.
+window.Promise = window.Promise || require('promise-polyfill');
```
cvan: personally, I check for `'Promise' in window` instead of always assigning

PR #1028, opened 2016-02-26
add "material-texture-loaded" and "video-ended" events to material component

nylki:
I am using the `texturePromises` array to store a promise that resolves to the texture once the image is loaded by three.js.
To emit the `textureLoaded` event, a `.then` is added to the `texturePromises[src]`.


ngokevin:
`material-texture-loaded` would be a good event name. It prefixes the component name to the event name, and we haven't been using camelCase for our event names.

examples/primitives-images/index.html
```diff
@@ -14,14 +14,17 @@
         <img id="office3" src="mozvr-office-3.jpg">
       </a-assets>
 
-      <a-image position="0 1 -8" width="16" height="4.5" src="#office1"></a-image>
```
ngokevin: we should keep these images around

examples/primitives-images/index.html
```diff
@@ -14,14 +14,17 @@
...
-      <a-image position="5 -2 -3" width="12" height="8" scale="0.3 0.3 0.3" src="#office3"></a-image>
+      <a-image id="eventImage" position="0 1 -8" width="16" height="4.5" src="https://upload.wikimedia.org/wikipedia/commons/5/5e/Big_Wood,_N2.JPG">
+
+      </a-image>
 
       <a-sky color="#ECECEC"></a-sky>
     </a-scene>
+
+    <script type="text/javascript">
+      document.getElementById('eventImage').addEventListener('textureLoaded', function (e) {
```
dmarcos: Is a message on console a good way to illustrate the feature? What about doing something visual when the textures load?

src/components/material.js
```diff
@@ -201,37 +209,39 @@ module.exports.Component = registerComponent('material', {
...
-  } else {
-    TextureLoader.load(src, onLoad, onProgress, onError);
-  }
+    if (isEl) {
+      createTexture(src);
+    } else {
+      TextureLoader.load(src, onLoad, onProgress, onError);
+    }
 
-  function createTexture (texture) {
```
dmarcos: I think this lines should not change. Wrong indentation

nylki:
I made changes according to your comments. @dmarcos the indentation in the line is correct (because of the Promise resolver function).

examples/primitives-video/index.html
```diff
@@ -27,7 +27,10 @@
       </a-assets>
 
       <a-video position="0 60 -150" src="https://ucarecdn.com/4130c599-445f-4e76-9052-49978e972641/"
-               width="160" height="90"></a-video>
+               width="160" height="90">
+<a-animation begin="material-texture-loaded" attribute="rotation" from="180 0 0" to="0 0 0" dur="2000"> </a-animation>
```
ngokevin: Not a big deal, but this HTML should be indented.

src/components/material.js
```diff
@@ -176,20 +177,38 @@ module.exports.Component = registerComponent('material', {
...
       material.needsUpdate = true;
     }
-    function loadImage (src) { loadImageTexture(material, src, data.repeat); }
-    function loadVideo (src) { loadVideoTexture(material, src, data.width, data.height); }
+
+    function loadImage (src) {
+      texturePromises[src] = loadImageTexture(material, src, data.repeat);
+      texturePromises[src].then(function (texture) {
+        self.el.emit('material-texture-loaded');
+      }); }
```
ngokevin: Can you move this brace down

src/components/material.js
```diff
@@ -176,20 +177,38 @@ module.exports.Component = registerComponent('material', {
...
+    function loadVideo (src) {
+      texturePromises[src] = loadVideoTexture(material, src, data.width, data.height);
+      texturePromises[src].then(function (videoEl) {
+        videoEl.addEventListener('loadeddata', function (e) {
+          self.el.emit('material-texture-loaded', videoEl);
+        });
+        videoEl.addEventListener('ended', function (e) {
+          // works only for non-loop videos
+          self.el.emit('video-ended');
```
ngokevin: Let's do `material-video-ended`

dmarcos: Should we create an emit method on the component that appends the name automatically?

ngokevin:
Once we rename to `material-video-ended`, this looks good to me. The nitpicks are optional, I can clean up later.

ngokevin:
Adding a test would be a bonus, but I could do that if you want.

dmarcos:
Tests and nitpicks are important. It costs nothing to do it right away. We will never do it later.

src/components/material.js
ngokevin: A bit too magical. I'd be worried people making components will try to emit an event but wonder why their event is not firing because they didn't prepend the event name in the listener.

dmarcos: Nobody has to prepend anything to the method. It would be a method in the component API. You would do:

````javascript
this.emit('texture-loaded')
````
instead of 
````javascript
self.el.emit('material-texture-loaded')
`````

**Comment by ngokevin** [comment]
Do we need our own style guide on top of https://github.com/Flet/semistandard? The idea of that was to spend less time nitpicking, but it seems we need more rules?

File: src/components/material.js (hunk 6 above)
Comment (ngokevin) [inline src/components/material.js]: Not the method, the handler. Let's say I write a component that does `this.emit('jump')`. Then in another snippet I try to do `el.addEventListener('jump');`. I'll spend a lot of time what the heck is going on. It will be confusing unless I get lucky enough to read the line in the documentation that will mention that.

**Comment by dmarcos** [comment]
It would be nice If there's a way to automate those nits.

File: src/components/material.js (hunk 6 above)
Comment (dmarcos) [inline src/components/material.js]: I see. Is there any value on knowing when an event is triggered from a component? Maybe the emit method doesn't change the event name but adds additional info / console messages that indicates what component is triggered from.

Comment (dmarcos) [inline src/components/material.js]: What about wrapping ```el.addEventListener``` so you can do ```el.addEventListener('componentName', 'eventName')```? Just throwing some ideas.

**Comment by ngokevin** [comment]
I looked and we can configure our own with http://eslint.org/docs/rules/ , there are a lot of rules there that we are using that aren't configured in semistandard (e.g., force vars on top, no else after return)

File: src/components/material.js (hunk 6 above)
Comment (ngokevin) [inline src/components/material.js]: Not sure. It sounds ideal that we can automate prepending the component name, but the drawback is that it increases API surface area. Maybe we can just keep it as a best practice guideline.

**Comment by nylki** [comment]
So many comments!
I renamed the event (and moved the brace).
Regarding tests:
I'd be happy to include a test, but I am not sure how you are doing it, I have only had some minimal experience with *mocha*. Also I am not sure what should be tested here (testing if event is actually triggered?).

**Comment by ngokevin** [comment]
Yeah, sorry, we hijacked your PR to talk about stuff :)

You can just add a test in the material component to check the event is triggered, yeah.

File: src/components/material.js [hunk 7]
```diff
@@ -176,20 +177,40 @@ module.exports.Component = registerComponent('material', {
...
       // Texture removed.
       material.map = null;
       material.needsUpdate = true;
     }
-    function loadImage (src) { loadImageTexture(material, src, data.repeat); }
-    function loadVideo (src) { loadVideoTexture(material, src, data.width, data.height); }
+
+    function loadImage (src) {
+      texturePromises[src] = loadImageTexture(material, src, data.repeat);
+      texturePromises[src].then(function (texture) {
```
dmarcos: Can we do
````javascript
texturePromises[src].then(emitEvent);
function emitEvent() { self.el.emit('material-texture-loaded'); }
````
````

**Comment by nylki** [comment]
Hey again,
@dmarcos The emitted event now yields an object as detail: `{src, element}`. For image sources there would be only a `src`, for video sources the event returns `src` and `element`. I figured for consistency reasons it would be better to use an object here and always include `src`, although you could get it via the video element. This way you can simply do stuff with `event.detail.src` without worrying if it is a video or image.

@ngokevin I added a small test and However I am not able to run the unit tests here (`Error: Cannot find module 'lolex'
` on `npm test`). I added a test.jpg to the test directory to have a valid `src`, hope that's ok.

File: tests/components/material.test.js [hunk 8]
```diff
@@ -68,6 +68,15 @@ suite('material', function () {
       assert.notOk(el.sceneEl.materials[oldMaterial.uuid]);
       assert.equal(el.sceneEl.materials[newMaterial.uuid], newMaterial);
     });
+
+    test('texture emits an event on finished loading', function (done) {
+      var el = this.el;
+      el.setAttribute('material', 'src:"test.jpg"');
```
ngokevin: this should be `url(test.jpg)`

src/components/look-at.js
```diff
@@ -45,6 +45,7 @@ module.exports.Component = registerComponent('look-at', {
   },
 
   init: function () {
+    this.tickCount = 0;
```
ngokevin: some leftover debug stuff?

tests/components/material.test.js
```diff
@@ -68,6 +68,15 @@ suite('material', function () {
       assert.notOk(el.sceneEl.materials[oldMaterial.uuid]);
       assert.equal(el.sceneEl.materials[newMaterial.uuid], newMaterial);
     });
+
+    test('texture emits an event on finished loading', function (done) {
```
ngokevin: : `emits event when loading texture`

tests/components/material.test.js
```diff
@@ -68,6 +68,15 @@ suite('material', function () {
...
+      var el = this.el;
+      el.setAttribute('material', 'src:"test.jpg"');
+      el.addEventListener('material-texture-loaded', function (e) {
```
ngokevin: can we do s/e/evt/g ?

src/components/material.js
```diff
@@ -278,6 +301,10 @@ function loadVideoTexture (material, src, height, width) {
   texture.needsUpdate = true;
   material.map = texture;
   material.needsUpdate = true;
+
+  // the promise is not necessary here, as the function is synchronous
```
ngokevin: Can we capitalize beginning of comments + add period at end?

tests/components/material.test.js
ngokevin: adding a true image won't work unless you define the path in the karma config to serve it. we have an example there of loading a true sound file

src/components/look-at.js
nylki: oh damn, `look-at` should not have been commited

ngokevin:
do you have your devdeps installed for testing?

nylki:
> do you have your devdeps installed for testing?

Should have. But `npm install` yields some warning: 
```
npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.8
npm WARN optional Skipping failed optional dependency /babel/chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@0.3.8
npm WARN chai-shallow-deep-equal@1.4.0 requires a peer of chai@>= 1.9.0 but none was installed.
npm WARN karma-chai-shallow-deep-equal@0.0.4 requires a peer of chai@* but none was installed.
npm WARN karma-sinon-chai@1.2.0 requires a peer of chai@^3.2.0 but none was installed.
npm WARN karma-sinon-chai@1.2.0 requires a peer of lolex@^1.4.0 but none was installed.
npm WARN karma-sinon-chai@1.2.0 requires a peer of sinon@^1.17.2 but none was installed.
npm WARN karma-sinon-chai@1.2.0 requires a peer of sinon-chai@^2.8.0 but none was installed.
```

dmarcos:
Tests don't pass

nylki:
@dmarcos yep its the new test i added that fails, but unfortunatelay I have problems with the test environment. Maybe someone of you can take a look at the test.

nylki:
@dmarcos passing now.

ngokevin:
PR needs rebase

tests/components/material.test.js
```diff
@@ -68,6 +68,15 @@ suite('material', function () {
       assert.notOk(el.sceneEl.materials[oldMaterial.uuid]);
       assert.equal(el.sceneEl.materials[newMaterial.uuid], newMaterial);
     });
+
+    test('emits event when loading texture', function (done) {
+      var el = this.el;
+      el.setAttribute('material', 'src: url(http://localhost:9876/base/examples/_images/mozvr.png)');
```
ngokevin: don't think we have to hardcode the domain, just start with ```/base/```

tests/components/material.test.js
```diff
@@ -68,6 +68,15 @@ suite('material', function () {
...
+      el.addEventListener('material-texture-loaded', function (evt) {
+        assert(evt.detail.src === 'http://localhost:9876/base/examples/_images/mozvr.png', 'src in event detail is the same as the url set to be the image texture src.');
```
ngokevin: Can store the URL in a var and use assert.equal, and don't need the failure message

```assert.equal(evt.detail.src, imageUrl)```

**Comment by ngokevin** [comment]
OK looks mostly good :)

@dmarcos we should get this in before your custom shaders so you can handle the future merge conflict

File: src/components/material.js [hunk 15]
```diff
@@ -176,20 +177,40 @@ module.exports.Component = registerComponent('material', {
...
+        // Material is first to load this texture. Load and resolve texture.
       }
     } else {
       // Texture removed.
       material.map = null;
       material.needsUpdate = true;
     }
-    function loadImage (src) { loadImageTexture(material, src, data.repeat); }
-    function loadVideo (src) { loadVideoTexture(material, src, data.width, data.height); }
+
```
dmarcos: The logic to handle textures has moved to the texture module. It should be pretty much copying and pasting the code into https://github.com/aframevr/aframe/blob/dev/src/utils/texture.js

nylki:
@dmarcos phew, should have created this PR after your changes :)
Just took a look at the changes. There are now flat/standard shaders, which both use `loadImageTexture()` via their own `loadImage()` function that are identical. This is no problem, as your loadImage() functions are very concise, but if I would copy paste my loadImage() functions to both files, it's a bit too much redundancy for my taste. But you may decide.

Alternatively, I thought about defining `loadImage()` in texture.js and export it to be used as `utils.loadImage()` in both shader files, which would be problematic because of argument handling.

nylki:
@dmarcos phew, should have created this PR after your changes :)
Just took a look at the changes. There are now flat/standard shaders, which both use `loadImageTexture()` via their own `loadImage()` function that are identical. This is no problem, as your loadImage() functions are very concise, but if I would copy paste my loadImage() functions to both files, it's a bit too much redundancy for my taste. But you may decide.

Alternatively, I thought about defining `loadImage()` in texture.js and export it to be used as `utils.loadImage()` in both shader files, which would be problematic because of argument handling.

ngokevin:
@dmarcos looks good to me

dmarcos:
Cool stuff! Thanks!

PR #1029, opened 2016-02-26
Make look-controls passive when not in use

### LINKED ISSUE #975: Extensible controls (Trouble using multiple controls, refactor controls)
There's a change on the dev branch that breaks one of my examples. Code:

``` html
<a-entity camera
          rotation="0 0 0"
          look-controls
          gamepad-controls>
</a-entity>
```

(actual scene i'm testing is [here](https://github.com/donmccurdy/sandbox-aframe/blob/master/examples/platforms/index.html), but has `look-controls` removed temporarily)

In that example there are three ways of rotating the camera: HMD (w/ `look-controls`), mouse (w/ `look-controls`), and gamepad right joystick (w/ `gamepad-controls`). In v0.1.\* this just worked, because my update() function in `gamepad-controls` was always called last, and I was careful to only update rotation when the gamepad was in use. On _dev_ that seems to have changed – `gamepad-controls.tick()` is called first, and `look-controls.tick()` overwrites my rotation regardless of whether any HMD or mouse events are happening. I've [tried to work around this in my code](https://github.com/donmccurdy/aframe-gamepad-controls/blob/feat-lock-rotation/gamepad-controls.js#L191), but it gets very messy quickly and I don't see a good solution without asking users of `gamepad-controls` to _also_ use some new custom version of `look-controls`.

I don't care much about the order in which component `tick()` calls should be made (that seems too complicated), but it would be great to make it easy for demos to support multiple input devices – I don't want to target a demo _only_ at users with an HMD and a gamepad, for example, and ideally the default controls with A-Frame would be passive when they're not in use.

Can think of a couple ways to get around this:
1. Controls should always export their state to position/rotation components, and not hide state internally. `wasd-controls` already does this, but it might be harder for `look-controls` because we want that dolly pitch/roll/yaw abstraction.
2. Controls should only update state if they're in use. E.g. if there's an HMD `look-controls` should be permanently in use, but between mouse events it should avoid overwriting `rotation`.

Would you take a PR for one of these, or can you think of another way to have controls play nicely together?

donmccurdy:
Fixes #975.

More comments there. Issue is that `gamepad-controls` and `look-controls` are incompatible on the dev branch, and it's not something I can work around within my component.

I could also package my own fork of `look-controls` *into* `gamepad-controls`, and that would be fine, or not sure if you all want to embed gamepad support officially into A-Frame?

In any case though, it would take some change like this (where both controls are passive when inactive), or else the two controls need to be aware of one another, which gets messier.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -111,6 +111,9 @@ module.exports.Component = registerComponent('look-controls', {
     var hmdEuler = new THREE.Euler();
     hmdEuler.order = 'YXZ';
     return function () {
+      if (!this.mouseDown && this.dolly.quaternion.equals(this.zeroQuaternion)) {
+        return;
+      }
       var pitchObject = this.pitchObject;
       var yawObject = this.yawObject;
       var hmdQuaternion = this.calculateHMDQuaternion();


ngokevin:
@dmarcos will look at this once he finishes up his current work

ngokevin:
@dmarcos r?

dmarcos:
Does not nesting solve the problem? 
````html
<a-entity gamepad-controls>
   <a-entity camera rotation="0 0 0" look-controls></a-entity>
</a-entity>
````

donmccurdy:
sort of, but with [a few issues](https://github.com/aframevr/aframe/issues/975#issuecomment-189099197).

ngokevin:
@donmccurdy Should we investigate your universal controls solution rather than pulling in this temporary fix?

donmccurdy:
Cool! I guess I assumed that would be too much of a refactor to consider right now, but I can clean up that solution and get a PR to review/consider later today or tomorrow.

ngokevin:
@dmarcos hasn't looked at it yet, but it looks promising to me. I'd poke him.

ngokevin:
Going to close for now to look at a bigger solution. If we don't arrive to that, we can reopen this.

PR #1039, opened 2016-02-27
Adds the array property type and allows for default: [] to auto type to array

### LINKED ISSUE #1038: add array property type
It would be nice to be able to register custom property types. This would enable things like 

```javascript
  schema: {
      enabled: { default: true },
      templates: { type: 'array', default: [] },
      timing: { type: 'range', default: [0, 50] }
    },
```

JoshGalvin:
Fixes #1038

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -41,11 +41,11 @@ module.exports.registerPropertyType = registerPropertyType;
 
 function arrayParse (value) {
   if (Array.isArray(value)) { return value; }
-  return value === null || value.length === 0 ? [] : value.split(' ');
+  return value === null || value.length === 0 ? [] : value.split(',');
 }
 
 function arrayStringify (value) {
-  return value.join(' ');
+  return value.join(',');
 }
 
 function defaultParse (value) {
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -13,6 +13,7 @@ registerPropertyType('selector', '', selectorParse, selectorStringify);
 registerPropertyType('src', '', srcParse);
 registerPropertyType('string', '', defaultParse, defaultStringify);
 registerPropertyType('vec3', { x: 0, y: 0, z: 0 }, vec3Parse, coordinates.stringify);
+registerPropertyType('array', [], arrayParse, arrayStringify);
 
 /**
  * Register a parser for re-use such that when someone uses `type` in the schema,
@@ -38,8 +39,18 @@ function registerPropertyType (type, defaultValue, parse, stringify) {
 }
 module.exports.registerPropertyType = registerPropertyType;
 
+function arrayParse (value) {
+  return value.length === 0 ?
+    [] :
+    value.split(' ');
+}
+
 function defaultParse (value) {
   return value;
+}
+
+function arrayStringify (value) {
+  return value.join(' ');
 }
 
 function defaultStringify (value) {
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -41,10 +41,9 @@ module.exports.registerPropertyType = registerPropertyType;
 
 function arrayParse (value) {
   if (Array.isArray(value)) { return value; }
-  if (value === null || value.length === 0) { return []; }
-  return value.split(',').map(function (str) {
-    return str.trim();
-  });
+  if (!value || typeof value !== 'string') { return []; }
+  return value.split(',').map(trim);
+  function trim (str) { return str.trim(); }
 }
 
 function arrayStringify (value) {
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -41,11 +41,10 @@ module.exports.registerPropertyType = registerPropertyType;
 
 function arrayParse (value) {
   if (Array.isArray(value)) { return value; }
-  return value === null || value.length === 0
-    ? []
-    : value.split(',').map(function (str) {
-      return str.trim();
-    });
+  if (value === null || value.length === 0) { return []; }
+  return value.split(',').map(function (str) {
+    return str.trim();
+  });
 }
 
 function arrayStringify (value) {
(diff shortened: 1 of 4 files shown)


src/core/propertyTypes.js
```diff
@@ -38,10 +39,20 @@ function registerPropertyType (type, defaultValue, parse, stringify) {
...
+  return value.length === 0 ?
+    [] :
+    value.split(' ');
+}
+
 function defaultParse (value) {
   return value;
 }
 
+function arrayStringify (value) {
```
ngokevin: nitpick but this can sit next to the `arrayParse`, it's a bit alphabetized right now

src/core/propertyTypes.js
```diff
@@ -38,10 +39,20 @@ function registerPropertyType (type, defaultValue, parse, stringify) {
 }
 module.exports.registerPropertyType = registerPropertyType;
 
+function arrayParse (value) {
+  return value.length === 0 ?
```
ngokevin: this can be a one-liner

ngokevin: you should check if the value is already an array to see if you need to parse at all

tests/core/propertyTypes.test.js
```diff
@@ -62,4 +62,19 @@ suite('propertyTypes', function () {
...
     });
   });
+
+  suite('array', function() {
+    var parse = propertyTypes.array.parse;
+    var stringify = propertyTypes.array.stringify;
+
+    test('parses array', function () {
+      assert.deepEqual(parse(''), []);
+      assert.deepEqual(parse('5 test'), ['5', 'test']);
```
ngokevin: try passing in some arrays here

tests/core/propertyTypes.test.js
```diff
@@ -62,4 +62,19 @@ suite('propertyTypes', function () {
...
+    });
+
+    test('stringifies array', function () {
+      assert.equal(stringify([]), '');
+      assert.equal(stringify([5, 10]), '5 10');
```
ngokevin: it should probably handle strings too, just in case

src/core/propertyTypes.js
```diff
@@ -13,6 +13,7 @@ registerPropertyType('selector', '', selectorParse, selectorStringify);
 registerPropertyType('src', '', srcParse);
 registerPropertyType('string', '', defaultParse, defaultStringify);
 registerPropertyType('vec3', { x: 0, y: 0, z: 0 }, vec3Parse, coordinates.stringify);
+registerPropertyType('array', [], arrayParse, arrayStringify);
```
ngokevin: nitpick: these are alphabetized

ngokevin:
Looks good after those comments! Thanks!

tests/core/propertyTypes.test.js
JoshGalvin: How do you think we should denote strings with spaces in them? Or is that just disallowed?

ngokevin: Oh, I think we should do comma-delimited. Let's do that.

ngokevin: just want to handle the error case we don't try to do `<string>.join(',')`

JoshGalvin: What about strings with commas in them?

JoshGalvin: When would the stringify ever receive a string?

ngokevin: Hm, you're right. I guess it wouldn't. `setAttribute` should pass it through.

ngokevin: If someone wants strings with commas, they can define their own parse.

JoshGalvin: :hocho: or :gun: for the duel concerning comma or space delimitation?

tests/core/propertyTypes.test.js
JoshGalvin: I did but it didn't autoclose your comment :(

tests/core/propertyTypes.test.js
ngokevin: I'm ok with just :facepunch: 

Hm, spaces seem cleaner, but then we wouldn't be able to have spaces in our strings. Not sure. @dmarcos?

tests/core/propertyTypes.test.js
ngokevin: - [x] donzo

tests/core/propertyTypes.test.js
dmarcos: We already have a vector type: for instance ```lookAt: 0 0 0``` I would use coma to differentiate between vector and arrays. Also we might want to have arrays of vectors ```0 1 0, 0 3 0, -1 0 0, ...```

src/core/propertyTypes.js
```diff
@@ -38,6 +39,19 @@ function registerPropertyType (type, defaultValue, parse, stringify) {
 }
 module.exports.registerPropertyType = registerPropertyType;
 
+function arrayParse (value) {
+  if (Array.isArray(value)) { return value; }
+  return value === null || value.length === 0 ?
```
dmarcos: For better readability I would do:
````javascript
if (!value || typeof value !== 'string') { return []; }
return value.split(',').map(trim);
function trim (str) { return str.trim(); }
````
return

File: src/core/propertyTypes.js [hunk 7]
```diff
@@ -38,6 +39,18 @@ function registerPropertyType (type, defaultValue, parse, stringify) {
 }
 module.exports.registerPropertyType = registerPropertyType;
 
+function arrayParse (value) {
+  if (Array.isArray(value)) { return value; }
```
dmarcos: For better readibility. Can we do?
````javascript
if (Array.isArray(value)) { return value; }
if (!value || typeof value !== 'string') { return []; }
return value.split(',').map(trim);
function trim (str) { return str.trim(); }
````

dmarcos:
Thanks!

PR #1046, opened 2016-02-29
Add polygon primitive to geometry component

meta-meta:
How do you guys feel about this? I know it still needs some work. I'm not super happy about passing a string of "x1 y1 z1 x2 y2 z2 x3 y3 z3" and it could be a bit more robust considering all the possible errors parsing input.

Does this make sense for core or should it be a separate a-frame component?

### DIFF
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -52,7 +52,7 @@ module.exports.Component = registerComponent('geometry', {
     p: { default: 2, if: { primitive: ['torusKnot'] }, type: 'int' },
     primitive: {
       default: '',
-      oneOf: ['', 'box', 'circle', 'cylinder', 'plane',
+      oneOf: ['', 'box', 'circle', 'cylinder', 'plane', 'polygon',
               'ring', 'sphere', 'torus', 'torusKnot'] },
     q: { default: 3, if: { primitive: ['torusKnot'] }, type: 'int' },
     phiLength: { default: 360, if: { primitive: ['sphere'] } },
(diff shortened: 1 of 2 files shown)


src/components/geometry.js
```diff
@@ -143,6 +144,33 @@ function getGeometry (data, schema) {
     case 'plane': {
       return new THREE.PlaneBufferGeometry(data.width, data.height);
     }
+    case 'polygon': {
+      var verts = data.vertices.split(' ');
```
ngokevin: we should have a `vec3Array` property type that handles this out-of-the-box where it looks like ```1 2 3, 2 4 5, 1 6 2```

src/components/geometry.js
```diff
@@ -143,6 +144,33 @@ function getGeometry (data, schema) {
     case 'plane': {
       return new THREE.PlaneBufferGeometry(data.width, data.height);
     }
+    case 'polygon': {
```
ngokevin: maybe if we're passing in vertices, we don't need to specify a primitive?

```
geometry="vertices: 1 2 3, 4 5 6"
```

src/components/geometry.js
ngokevin: similar to https://github.com/aframevr/aframe/blob/dev/src/core/propertyTypes.js#L42

ngokevin:
I think this sort of makes sense for stdlib. Later on, we have APIs to be able to dynamically swap the schema based on conditions, which will allow us to internally split up the geometry component for cleaner code.

ngokevin:
We could possibly have ```registerGeometry```. Dunno.

ngokevin:
Need a lot of input from @dmarcos since the potential API for geometry component might change

Thanks for the PR! I was thinking of being able to pass vertices.

dmarcos:
A registerGeometry API makes more sense to me than passing vertices on the component. I need to think a bit more about this.

ngokevin:
Yeah, at the very least it can be a separate component for now. I feel like we might the geometry component API more generic and extensible later.

ngokevin:
I think we will close this in favor of a future `registerGeometry` where you can define a name + Geometry object + options and share with other people.

```js
AFRAME.registerGeometry('triangle', {
  vertices: [ // ... ],
  faces: [ // ...]
});
```

ngokevin:
But it can be separate component until then!

meta-meta:
Sounds good to me!

PR #1054, opened 2016-03-01
rename entity.paused to entity.isPlaying (fixes #1044)

### LINKED ISSUE #1044: [code quality] rename entity `paused` to entity `isPlaying`
`paused` is a negative/off state, `isPlaying` would be more appropriate to keep track of

ngokevin:
Want to do this before 0.2.0 because this is public API which will be hard to undo

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -13,20 +13,16 @@ var registerElement = re.registerElement;
 var AEntity;
 
 /**
- * Entity element definition.
- * Entities represent all elements that are part of the scene, and always have
- * a position, rotation, and scale.
- * In the entity-component system, entities are just a container of components.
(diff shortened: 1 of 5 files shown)


src/core/a-entity.js
```diff
@@ -14,9 +14,6 @@ var AEntity;
 
 /**
  * Entity element definition.
- * Entities represent all elements that are part of the scene, and always have
```
dmarcos: Why is this comment gone?

ngokevin: not necessary to have tutorial material here since it's explained in the docs

ngokevin: i can put it back though, this one isn't too bad

ngokevin: ok i put it back, but tightened it a little

PR #1056, opened 2016-03-01
Add pointerlock support.

### LINKED ISSUE #848: Change default desktop look controls to not be mouse drag
Controls should be non-opinionated and should be free to swap out, but for an out-of-the-box experience, we should use some form of pointer lock rather than click/drag (on desktop, maybe VR mode only?).

bnolan:
Fixes #848.


ngokevin:
Thanks! We'll take a look at this

src/components/look-controls.js
```diff
@@ -192,12 +207,18 @@ module.exports.Component = registerComponent('look-controls', {
     var yawObject = this.yawObject;
     var previousMouseEvent = this.previousMouseEvent;
 
-    if (!this.mouseDown || !this.data.enabled) { return; }
+    if (!this.data.enabled) {
```
cvan: good catch. just realised most of these methods need to check this too.

cvan: we need to add `if (!this.data.enabled) { return; }` checks to

* `onMouseDown`
* `onMouseMove`
* `releaseMouse`
* `releaseMouse`
* `onTouchStart`
* `onTouchMove`
* `onTouchEnd`
* `onPointerLockChange`
* `onPointerLockError`
* `onPointerLockChange`
* `onPointerLockError`

cvan:
this works very nicely. thanks for doing this :+1:.

perhaps we want some way of toggling between mouse-drag controls and Pointer Lock (which IMO is much easier and more intuitive when you actually have a headset on).

src/components/look-controls.js
cvan: on second thought, most of these are probably harmless

ngokevin:
Is there an argument to keeping mouse-drag? Either mouse-drag or pointer lock, you have to click to discover.

cvan:
though we can't assume that the user will use the mouse/pointer for look controls, it might be nice to call `requestPointerLock` if it hasn't been already. maybe not. it could be too presumptuous + annoying.

ngokevin:
This would be assuming if the user is using the mouse. If not, then it doesn't matter mouse-drag vs. pointer lock. But let's assume the user is using the mouse. When to mouse-drag (if at all) vs pointer lock?

cvan:
> This would be assuming if the user is using the mouse.

Precisely what I said above, yeah.

> If not, then it doesn't matter mouse-drag vs. pointer lock. But let's assume the user is using the mouse. When to mouse-drag (if at all) vs pointer lock?

Pointer Lock requires you to keep still. Mouse drag allows you to move and sit still until you explicitly click + drag. Two different types of look controls with some different and some overlapping use cases.

cvan:
again, we don't know if the user is going to use the mouse. I meant to mention above it would be nice to request pointer lock upon entering VR (if we deem the user wants pointer lock). this way you can avoid the Pointer Lock permission prompt in Firefox. [see this example.](http://dontcallmedom.github.io/web-permissions-req/tests/pointerlock-fullscreen-combined.html)

src/components/look-controls.js
```diff
@@ -211,6 +232,17 @@ module.exports.Component = registerComponent('look-controls', {
   onMouseDown: function (event) {
     this.mouseDown = true;
     this.previousMouseEvent = event;
+
+    var sceneEl = this.el.sceneEl;
+    var canvasEl = sceneEl.canvas;
+
+    if (!this.pointerLocked) {
```
cvan: if I decline the Pointer Lock permission prompt, every subsequent mouse click shouldn't try to request Pointer Lock again. if I enter Pointer Lock, press ESC to exit Pointer Lock, then it's fine to try to request Pointer Lock again.

I couldn't figure out a reliable way of figuring out if the user declined Pointer Lock permission. `{moz}pointerlockerror` I couldn't get to fire.

ideally, this API would be covered by the Permissions API, but it's not (yet).

bnolan: Yeah, I've had problems with picking up pointerlockerror and using that detect that the user doesn't want to lock their mouse. I can see that in a situation where you have multiple a-frame components, it might be nicer to use drag to look instead of grabbing pointer lock. Maybe we could make it a parameter on look-controls where pointer lock can be enabled or disabled?

bnolan: (And have it on by default).

dmarcos:
I think pointer lock should belong to a different component. ```fps-controls```?

cvan:
> I think pointer lock should belong to a different component. ```fps-controls```?

I like that :+1:

I fancy myself FPS in (Web)VR, so I'd use that component _a lot_.

bnolan:
I think another component makes sense - but my vote would be for pointer lock to be the default interaction component on desktop. The drag-to-look is an inferior user experience imo.

ngokevin:
Sorry @bnolan for hijacking the PR for discussion. Let's figure how this is going to work. Are we thinking something like:

```first-person-controls```: WASD + mouse (pointer lock by default, I agree, even with the prompt)
```look-controls```: WASD + VR + WebVR polyfill

How does responsiveness work? What if we want first-person-controls for a mono experience, but look-controls for a stereo experience? Is that done by the developer?

**Comment by bnolan** [comment]
I think first-person-controls should be WASD + mouselook on desktop, cardboard stereoscopic on mobile. Do you have a locomotion solution on mobile?

**Comment by ngokevin** [comment]
Mobile would be handled by look-controls that has the WebVR polyfill. Maybe first-person isn't the best name, maybe fps-controls might have been the better name even though it's not always a shooter.

We don't have solutions out of the box for mobile locomotion yet. Right now you can do gaze-based teleport navigation manually. Later will probably use Cardboard V2 mouse/touchdown to move forward

**Comment by cvan** [comment]
> I think another component makes sense - but my vote would be for pointer lock to be the default interaction component on desktop.

I 100% think A-Frame should ship with a Pointer Lock component. For non-VR content on desktop, mouselook can sometimes be more desirable.

Because we didn't have a Pointer Lock-like component, we've seen folks create custom components to support the mouse-based FPS controls that Pointer Lock makes real simple.

Here's [`no-click-look-controls`](https://github.com/alexrkass/aframe-no-click-look-controls). And an [example](http://alexrkass.com/no-click-example/). And an [issue I filed about using Pointer Lock](https://github.com/alexrkass/aframe-no-click-look-controls/issues/1).

Besides my [old closed PR](https://github.com/aframevr/aframe-core/pull/121), I guess I forgot someone made a component for Pointer Lock: [`fps-look-component`](https://github.com/cemkod/aframe-fps-look-component/blob/master/index.js).

Have you compared this last one with yours? I haven't looked closely, but I'm curious if there's anything better or overlooked in his approach. Personally, I like the `sensitivity` property - that's super useful (even to make it configurable for mouse-look controls too).

> The drag-to-look is an inferior user experience imo.

For VR mode, I 100% agree.

My opinion: I love Pointer Lock for FPS (whether in VR or just plain mono content on desktop) and whenever I have a VR headset on and can't see my trackpad/keyboard. I like mouse-look controls for when I want to be careful about where/when/how I want to pan.

My conclusion: If I had to pick one, I'd go with Pointer Lock (`fps-controls`) be the default with a configurable `sensitivity` property.

**Comment by cvan** [comment]
We could make this real simple and just always enable Pointer Lock when entering VR mode. Can everyone agree that'd be the best user experience?

**Comment by ngokevin** [comment]
That's the hard part. Encapsulating WASD+mouse control schemes vs. HMD-oriented control schemes.

Pointer Lock for mono desktop + mono full screen is great. But it doesn't pair well with an HMD.

**Comment by ngokevin** [comment]
- **Desktop, headset, not fullscreen**: WASD + pointer lock
- **Desktop, headset, fullscreen**: WASD + HMD
- ** Desktop, no headset**: WASD + pointerlock
- **Mobile**: Touch to move + WebVR polyfill

File: src/components/look-controls.js [hunk 3]
```diff
@@ -88,6 +91,12 @@ module.exports.Component = registerComponent('look-controls', {
     canvasEl.addEventListener('touchstart', this.onTouchStart);
     canvasEl.addEventListener('touchmove', this.onTouchMove);
     canvasEl.addEventListener('touchend', this.onTouchEnd);
+
+    // Pointer lock
+    document.addEventListener('pointerlockchange', this.onPointerLockChange, false);
+    document.addEventListener('pointerlockerror', this.onPointerLockError, false);
+    document.addEventListener('mozpointerlockchange', this.onPointerLockChange, false);
```
cvan: FYI, I just checkend and Firefox seems to fire both `mozpointerlockchange` and `pointerlockchange` (ugh).

src/components/look-controls.js
```diff
@@ -242,5 +274,13 @@ module.exports.Component = registerComponent('look-controls', {
 
   onTouchEnd: function () {
     this.touchStarted = false;
+  },
+
+  onPointerLockChange: function () {
+    this.pointerLocked = !!(document.pointerLockElement || document.mozPointerLockElement);
```
cvan: shouldn't we check that `document.pointerLockElement`/`document.mozPointerLockElement` is `this.el.sceneEl`?

dmarcos:
Also be aware that aframe scenes when in mono mode might not be alone on the page. Look at code pen or aframe.io. It doesn't seem right to me to capture the pointer by default. For next version we will also allow multiple aframe scenes on a single page.

ashconnell:
re: @ngokevin (https://github.com/aframevr/aframe/issues/1060#issuecomment-191054349)

When you're in VR you probably want pointerlock to only rotate the Y axis to allow sitting down experiences. Outside of VR you want pointerlock to be the replacement of the HMD so full axis rotation.

ngokevin:
re: @dmarcos multiple scenes on a page, haven't seen anyone try request that use case yet so i'm wary about going out of our way for it. Even then I think pointer lock is better, you click on one of the scenes to select it and lock into it. The experience of mouse-drag is poor enough to warrant that.

re: @ashconnell Yeah, I noted in that comment to restrict it to an axis. But yeah, WASD + HMD + restricted-axis-mouse sounds good.

dmarcos:
I want people to easily integrate vr content on 2d pages. aframe scenes have to easily cohexist with other elements in the page. When a scene goes fullscren pointer lock would be acceptable.

dmarcos:
Aframe scene hijacking or hiding the pointer does not seem a good experience to me

src/components/look-controls.js
cvan: if the user rejects the permission prompt, then for the rest of the page session, we probably shouldn't ever try to request Pointer Lock again.

src/components/look-controls.js
```diff
@@ -242,5 +274,13 @@ module.exports.Component = registerComponent('look-controls', {
...
+  },
+
+  onPointerLockError: function () {
+    this.pointerLocked = false;
```
cvan: should probably also add

```js
  this.pointerLockedDenied = true;
```

and guard against that for the rest of the page session (until a browser reload).

I did [a bunch of tests](https://people.mozilla.org/~cwiemeersch/pointerlockerror-bug.html).

I discovered that Chrome actually fires `fullscreenerror` if you explicitly deny the Pointer Lock permission prompt.
Unfortunately, Firefox does not. (I filed a [platform bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1252749) with the test case.) A hacky workaround I can think of is to add some `setTimeout` to wait a few seconds and then assume the user rejected it.

I discovered that the `fullscreenerror` event _does_ get fired in Firefox - but only when trying to call `requestPointerLock` outside a user-initiated event.

cvan: should probably also add

```js
  this.pointerLockedDenied = true;
```

and guard against that for the rest of the page session (until a browser reload).

I did [a bunch of tests](https://people.mozilla.org/~cwiemeersch/pointerlockerror-bug.html).

I discovered that Chrome actually fires `fullscreenerror` if you explicitly deny the Pointer Lock permission prompt.
Unfortunately, Firefox does not. (I filed a [platform bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1252749) with the test case.) A hacky workaround I can think of is to add some `setTimeout` to wait a few seconds and then assume the user rejected it.

I discovered that the `fullscreenerror` event _does_ get fired in Firefox - but only when trying to call `requestPointerLock` outside a user-initiated event.

src/components/look-controls.js
cvan: see [this comment below](https://github.com/aframevr/aframe/pull/1056/files#r54680634)

ngokevin:
> I want people to easily integrate vr content on 2d pages.

That's extra work and consideration for something I haven't seen anyone try. We should optimize for full-page 3D + VR.

> Aframe scene hijacking or hiding the pointer does not seem a good experience to me

Mousedrag is definitively worse though. Hiding the pointer (via a browser API) is not that bad if it's activated with focus/click. In fact, it's expected for a first-person experience.

cvan:
> I want people to easily integrate vr content on 2d pages.

I don't understand why Pointer Lock makes it more difficult to integrate VR content. The browser handles the pointer capturing/releasing. You can have a button over your VR content that says "play"/"enter." For obvious performance reasons and presentation reasons (not wanting audio/video to play in the background), it's unlikely you're going to be playing more than one 3D/VR experience at a time.

I think it'd be great to have a newspaper article with three embedded 3D/VR A-Frame scenes. But those scenes IMO shouldn't just automatically load and have to rely on tedious mousedrag look controls.

You can add a Play/Start button. That's how YouTube/Vimeo embeds work, and it's how JSFiddle snippet embeds used to work as well.

> Mousedrag is definitively worse though. Hiding the pointer (via a browser API) is not that bad if it's activated with focus/click. In fact, it's expected for a first-person experience.

:+1: definitely agree here.

dmarcos:
you haven't seen it because there's no production sites out there yet. Do you think people are going to go all out and convert their sites to aframe? I doubt it. I want to see the new york times including pieces of VR content or Wikipedia enhancing articles with VR media. aframe.io or codepen already follow that model. When an aframe scene is on 2D / no fullscreen mode I consider that the experience is on preview mode. Drag to look is ok. As I said pointer lock is acceptable when going fullscreen. What we should do is enaging fullscreen + mono when the user clicks on the goggles icon if an HMD is not present. We can then enable pointer lock.

dmarcos:
As cvan said the play button would be the google icon. It conveys that the user wants to engage with the experience. Otherwise pointerlock should stay out of the way

ngokevin:
That sounds like an improvement (given that desktop users are allowed to enter fullscreen without a headset) and a step in the right direction.

Though it's not very optimal to consider a full-page A-Frame site "preview mode". Mousedrag still sucks in that case. Allowing non-VR mono fullscreen experiences will help though.

bnolan:
I disagree with @dmarcos, I think pointerlock is suitable all the time, not just fullscreen, but I'm happy to make a seperate component for proper pointerlock + wasd + arrow keys (for the germans) and we can abandon this PR.

cvan:
> a seperate component for proper pointerlock + wasd + arrow keys (for the germans) and we can abandon this PR.

I think that could be done in this PR

bnolan:
Ok cool, let me know what the plan is :D

ngokevin:
@dmarcos @cvan We talked about this earlier. Did we decide on:

- Non-fullscreen
  - Pointer-lock by default. Can be configured to mousedrag (e.g., ```look-controls="nonFullscreenMouseDrag: true") to satisfy embedded scenes
  - Or vice-versa? Mousedrag by default. Can be configured to pointer lock (e.g., ```look-controls="nonFullScreenPointerLock: true"```) to satisfy full A-Frame sites

- Fullscreen
  - Pointer-lock always
  - Pointer-lock restricted axis if headset is plugged in

ngokevin:
Bump for discussion.

dmarcos:
When it's not fullscreen pointer lock is disabled by default to prevent aframe hijacking the page where it's embedded.

ngokevin:
In the proposal, it'd be up to the developers who embed A-Frame to specify mousedrag in non-fullscreen mode. I think once we validate the embedded use case, we can re-evaluate. But in the meantime, it seems less confusing for A-Frame sites, most of which are non-embedded, to default to pointer lock. I think it's not optimal to start with mouse drag and then go to pointer lock (remember not everyone will think to full screen).

dmarcos:
As I said what we're seeing now is people getting their feet wet with aframe and not creating production content. Websites won't be completely rebuilt in aframe. It just does not make sense at this point. Hijacking the pointer by default is too prescriptive and gets in the way of the experience a website wants to provide. The user should decide when to fully engage with the experience by clicking fullscreen. We should consider no-fullscreen as a **preview mode** by default and drag to look around seems appropriate.

I think that our codepen demos would be immediately a worse experience if pointerlock is the default behavior in non fullscreen mode.

Most of [playcanvas] (https://playcanvas.com/explore/featured) and [turbulenz] (https://ga.me/) content doesn't use pointer lock and the demos that do I find it irritating. Look at tanks for instance (http://tanx.playcanvas.com/) and try to interact with your browser (switch tabs for instance). It's just not a good experience.

(Clara.io)[https://clara.io/library] and (sketchfab) [https://sketchfab.com/models/staffpicks] previewers don't lock the pointer either I think the experience is what it should be.

dmarcos:
Referencing the author of the PR. Visit the gallery of sceneVR:  https://www.scenevr.com/scenes
With AFrame those previews could be interactive with drag to look. Selecting one is for me the equivalent of clicking play/fullscreen. The user wants to engage with a particular experience and pointer lock then makes sense.

ngokevin:
OK I have reservations about that:

- Mousedrag is not great for embedded scenes with preview mode. You have to click the scene to play it (assuming not all scenes will be rendering at the same time), and then click again to drag it.
- I think hijacking is okay given that the user has already expressed interest in interacting with the scene by clicking it + giving the browser permission.

But let's move forward and re-evaluate later. We can make these agreed improvements today:

- Pointer lock upon fullscreen (axis-restricted if HMD is plugged in). This can be done in this PR
- Option for pointer lock on non-fullscreen (something like ```look-controls="pointerlock-always: true"```)

bnolan:
I've been playing with the pointerlock in the scenevr editor, and there's a few things that need a change:

* It shouldn't grab mouse on right click
* It shouldn't fire the click event on objects when clicking to get focus
* If you grant 'always allows' permission in firefox, you have to left click then hit escape to get pointerlock (I dunno if theres a way around this, it is fine in chrome)

Shall I make pointerlock a setting on look-controls? 

`look-controls="pointer-lock: true"`

I think it's a shame if most people don't enable it by default, but we can only recommend people enable it in their scenes.

ngokevin:
@bnolan I think we should handle our extensible controls story first (https://github.com/aframevr/aframe/issues/975#issuecomment-196093953) because every change to controls is like a temporary fix at the moment. If we go ahead with Don's proposal, all the controls API will change.

ngokevin:
Going to close for now to checkout https://github.com/aframevr/aframe/issues/975 first, and since this PR has gone a bit off-track. But we still do want pointer lock as an option, just not sure if it will be part of a look-controls component.

bnolan:
Sweet sounds good, I'll try to remake the pointer lock controls as a
component.

On Thursday, 24 March 2016, Kevin Ngo <notifications@github.com> wrote:

> Going to close for now to checkout #975
> <https://github.com/aframevr/aframe/issues/975> first, and since this PR
> has gone a bit off-track. But we still do want pointer lock as an option,
> just not sure if it will be part of a look-controls component.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/aframevr/aframe/pull/1056#issuecomment-200536432>
>


-- 
Regards,
Ben Nolan

http://bennolan.com/

skype: bennolan
cell: +64 21 770 662

ngokevin:
Related: https://github.com/cemkod/aframe-fps-look-component

donmccurdy:
@bnolan if we go in the direction of #975, I've merged this PR into that refactor [here](https://github.com/donmccurdy/sandbox-aframe/blob/master/src/controls/mouse-controls.js).

PR #1061, opened 2016-03-02
fix default scene component initializing twice (fixes #1059) + fix sound tests

### LINKED ISSUE #1059: Scene components listed in `defaultComponents` get initialised twice in Chrome (or with native Custom Elements support?)
When a component is defined as a key in `defaultComponents`, the component's `init` method gets called twice.

Reproduced with `<a-scene>` for `canvas`, `keyboard-shortcuts`, and `vr-mode-ui`.

ngokevin:
- Sound tests broken in https://github.com/aframevr/aframe/pull/1050 , make sure Travis passes before merging

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -266,10 +266,6 @@ var proto = Object.create(ANode.prototype, {
         // Add component if it is a dependency and not yet defined.
         this.setAttribute(name, '');
       } else {
-        if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
-        }
         this.components[name] = new components[name].Component(this);
         if (!this.paused) { this.components[name].play(); }
       }
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -296,8 +296,12 @@ var AScene = module.exports = registerElement('a-scene', {
 
         // setTimeout in case the camera is being set dynamically with a setAttribute.
         setTimeout(function checkForCamera () {
-          var sceneCameras = self.querySelectorAll('[camera]');
-          if (sceneCameras.length !== 0) { return; }
+          var camera = self.querySelector('[camera]');
+
+          if (camera) {
+            self.emit('camera-ready', { camera: camera });
+            return;
+          }
 
           // DOM calls to create camera.
           cameraWrapperEl = document.createElement('a-entity');
@@ -309,6 +313,7 @@ var AScene = module.exports = registerElement('a-scene', {
           defaultCamera.setAttribute('look-controls');
           cameraWrapperEl.appendChild(defaultCamera);
           self.appendChild(cameraWrapperEl);
+          self.emit('camera-ready', { camera: defaultCamera });
         });
       }
     },
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -92,9 +92,17 @@ module.exports.Component = registerComponent('sound', {
     }
 
     listener = this.listener = new THREE.AudioListener();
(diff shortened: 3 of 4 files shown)


src/core/scene/a-scene.js
```diff
@@ -296,8 +296,12 @@ var AScene = module.exports = registerElement('a-scene', {
 
         // setTimeout in case the camera is being set dynamically with a setAttribute.
         setTimeout(function checkForCamera () {
-          var sceneCameras = self.querySelectorAll('[camera]');
-          if (sceneCameras.length !== 0) { return; }
+          var camera = self.querySelector('[camera]');
+
+          if (camera) {
+            self.emit('camera-ready', { camera: camera });
```
cvan: turning into @dmarcos

src/core/scene/a-scene.js
```diff
@@ -296,8 +296,12 @@ var AScene = module.exports = registerElement('a-scene', {
 
         // setTimeout in case the camera is being set dynamically with a setAttribute.
         setTimeout(function checkForCamera () {
-          var sceneCameras = self.querySelectorAll('[camera]');
-          if (sceneCameras.length !== 0) { return; }
+          var camera = self.querySelector('[camera]');
```
cvan: I know it's unlikely, but let's hope nobody else has a `camera` attribute on any other custom elements

src/core/scene/a-scene.js
ngokevin: i slightly prefer no spaces around braces for objects, but no biggie

src/core/scene/a-scene.js
ngokevin: Could also check isEntity for good measure

src/core/scene/a-scene.js
```diff
@@ -296,8 +296,12 @@ var AScene = module.exports = registerElement('a-scene', {
...
+
+          if (camera) {
```
cvan: could check `if (camera instanceof AEntity) {`

src/core/scene/a-scene.js
cvan: or `self.querySelector('[camera][rotation][position]')` or [this](https://github.com/aframevr/aframe/pull/1061/files#r54686055)

src/core/a-entity.js
```diff
@@ -266,10 +266,6 @@ var proto = Object.create(ANode.prototype, {
         // Add component if it is a dependency and not yet defined.
         this.setAttribute(name, '');
       } else {
-        if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
```
cvan: ah!

cvan: hmm, wish we could do this, but just not initialise the component lifecycle stuff. (it turns out `enterVR` among other things was getting called twice.)

ngokevin: yeh, it's because we have an attributeChangedCallback, and its sole reason of being there is to listen for DOM inspector changes

src/core/scene/a-scene.js
ngokevin: `isEntity` is probably simplest, instanceof AEntity would work too

querySelector with rotation + position not as reliable because can't guarantee those will be there

ngokevin: I can change this in a follow-up commit

src/core/a-entity.js
cvan: yeah, I follow. it'd be hacky, but we could like ignore reinitialising those components twice upon those attribute updates for those particular components on the scene. it'd be a mess but yeah.

ngokevin: Better solution would be to ensure components don't get init twice

cvan: yup

PR #1066, opened 2016-03-02
Properly detach animations by removing event listeners

masom:
Detaching an animation from the DOM kept the event handlers registered, resulting in the animation still being triggered even if it had no DOM parent.

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -102,6 +102,7 @@ module.exports.AAnimation = registerElement('a-animation', {
 
     detachedCallback: {
       value: function () {
+        this.removeEventListeners(this.evt);
         if (!this.isRunning) { return; }
         this.stop();
       }


src/core/a-animation.js
```diff
@@ -102,6 +102,7 @@ module.exports.AAnimation = registerElement('a-animation', {
 
     detachedCallback: {
       value: function () {
+        this.removeEventListeners(this.evt);
```
dmarcos: I think we should make ```this.removeEventListeners``` to use this.evt internally instead of being passed as a parameter. @ngokevin what do you think?

ngokevin: sounds good, @masom want to do that?

ngokevin:
Can you add a test to `tests/core/a-animation.test.js` in `suite('detachedCallback')`:

```js
    test('stops animation', function (done) {
      setupAnimation({begin: 'test-event'}, function (el, animationEl) {
        animationEl.parentNode.removeChild(animationEl);
        // Check that it doesn't have event listener.
      });
    });
```

masom:
@ngokevin Yep, I'm trying to setup the test environment on my machine and writing some tests for that behaviour.

src/core/a-animation.js
masom: :+1:

masom:
The issue is a little bit deeper / more complex than anticipated by this hastly PR.

PR #1076, opened 2016-03-03
fix scene default component double init (fixes #1059)

### LINKED ISSUE #1059: Scene components listed in `defaultComponents` get initialised twice in Chrome (or with native Custom Elements support?)

ngokevin:
Had to move the component-already-initialized check closer to the component initialization. I think because several components are trying to initialize the same component (on top of the scene setAttribute call) at the same time.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -64,8 +64,7 @@ var proto = Object.create(ANode.prototype, {
 
   /**
    * Tell parent to remove this element's object3D from its object3D.
-   * Do not call on scene element because that will cause a call to
-   * document.body.remove().
+   * Do not call on scene element because that will cause a call to document.body.remove().
    */
   detachedCallback: {
     value: function () {
@@ -151,10 +150,10 @@ var proto = Object.create(ANode.prototype, {
   },
 
   /**
-   * Returns an object3D of a given type or creates it if it doesn't exist and
-   * a Constructor is passed as an argument
-   * @param {string} type - Type of the object3D .
-   * @param {string} name - Component name.
+   * Gets or creates an object3D of a given type.
+
+   * @param {string} type - Type of the object3D.
+   * @param {string} Constructor - Constructor to use if need to create the object3D.
    * @type {Object}
    */
   getOrCreateObject3D: {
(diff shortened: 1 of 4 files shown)


src/core/component.js
```diff
@@ -212,7 +215,7 @@ module.exports.registerComponent = function (name, definition) {
   });
 
   if (components[name]) {
-    throw new Error('The component ' + name + ' has been already registered');
+    warn('The component ' + name + ' has been already registered');
```
dmarcos: I think this makes more sense to be an error than a warning. If a component is registered twice is a big problem.

ngokevin: OK. It's usually a benign operation and a warning usually suffice. Halting everything is pretty harsh. But I'll just leave it as an error until it becomes a problem.

dmarcos: A component registered twice it means that there are two versions of the same components or two different components with the same name colliding. That's a good reason to stop everything loudly and force the user to reevaluate.

src/core/a-entity.js
```diff
@@ -266,12 +264,17 @@ var proto = Object.create(ANode.prototype, {
           // For scene default components, expose them in the DOM.
           HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
         }
+
+        // Check if component already defined.
+        if (name in this.components) { return; }
```
dmarcos: Why is this necessary? isComponentDefined should catch if the component is already defined.

ngokevin: I changed some comments.

`isComponentDefined` checks if the component *should be defined* for the entity. What we want to check here if the component *has already been initialized* for the entity.

src/core/a-entity.js
```diff
@@ -551,17 +551,20 @@ var proto = Object.create(ANode.prototype, {
...
 function checkComponentDefined (el, name) {
   // Check if default components contain the component.
-  var inDefaults = el.defaultComponents[name];
+  if (el.defaultComponents[name] !== undefined) { return true; }
+
   // Check if element contains the component.
-  var inAttribute = el.hasAttribute(name);
-  if (inDefaults !== undefined || inAttribute) { return true; }
+  if (el.hasAttribute(name)) { return true; }
+
```
dmarcos: Why is this if statement gone?
The comment of the function is obsolete.

ngokevin: Just wanted to reduce one LOC

src/core/component.js
```diff
@@ -212,7 +213,7 @@ module.exports.registerComponent = function (name, definition) {
   });
 
   if (components[name]) {
-    throw new Error('The component ' + name + ' has been already registered');
+    throw new Error('The component `' + name + '` has been already registered');
```
dmarcos: Can we make this error message a bit better. What about? 
```The component `' + name + '` has been already registered. You might be loading two version of the same component or two different components with the same name are colliding.```

ngokevin: :+1:

src/core/a-entity.js
dmarcos: nevermind.

PR #1078, opened 2016-03-04
Sound single listener

meta-meta:


src/components/sound.js
```diff
@@ -83,27 +81,32 @@ module.exports.Component = registerComponent('sound', {
...
+
     if (sound) {
       this.stop();
       el.removeObject3D('sound');
     }
 
-    listener = this.listener = new THREE.AudioListener();
-    sound = this.sound = new THREE.Audio(listener);
-    el.setObject3D('sound', sound);
+    function addListenerToCamera() {
```
dmarcos: Can you define this after the if statement?

src/components/sound.js
```diff
@@ -83,27 +81,32 @@ module.exports.Component = registerComponent('sound', {
...
+      sceneEl.camera.add(listener);
```
dmarcos: The active camera can change. Do we have to change the listener when it happens?

dmarcos:
Nice I love to see you tinkering with the soud. 

Travis is complaining. One linter error:

https://travis-ci.org/aframevr/aframe/jobs/113555636#L503

And the sound tests fail as well.

src/components/sound.js
meta-meta: sure

src/components/sound.js
meta-meta: we'll have to move the listener to the active camera when it changes

meta-meta:
I can run the lint task by itself but it doesn't output anything. I had to commit with --no-verify. Not sure if it's a Windows thing. Will investigate.

src/components/camera.js
```diff
@@ -43,6 +43,7 @@ module.exports.Component = registerComponent('camera', {
     if (!oldData || oldData.active !== data.active) {
       if (data.active) {
         sceneEl.setActiveCamera(camera);
+        sceneEl.emit('active-camera-set', { camera: camera });
```
ngokevin: - Can we do `camera-set-active`? Want to prefix our event names with the component emitting it.
- Can we pass in the camera element rather than the camera object to give more info?

src/components/sound.js
```diff
@@ -92,18 +89,22 @@ module.exports.Component = registerComponent('sound', {
       el.removeObject3D('sound');
     }
 
-    listener = this.listener = new THREE.AudioListener();
-    sound = this.sound = new THREE.Audio(listener);
-    el.setObject3D('sound', sound);
+    // we want to make sure we have exactly one AudioListener. Let's stick it somewhere
```
ngokevin: Can tighten up the comment: ```// Only want one AudioListener. Cache it on the scene.```

src/components/sound.js
```diff
@@ -92,18 +89,22 @@ module.exports.Component = registerComponent('sound', {
...
 
     if (sceneEl.camera) {
       sceneEl.camera.add(listener);
-    } else {
-      sceneEl.addEventListener('camera-ready', function addAudioListener (evt) {
-        evt.detail.cameraEl.object3D.add(listener);
-      });
     }
 
+    // if we don't have a camera yet or our active camera gets changed
```
ngokevin: Sorry, I like small comments :: ```// Wait for camera if necessary.```

src/components/camera.js
meta-meta: Hmm we have `camera` here. `camera.el.components` is an empty object at this point so cameraEl would actually give us less info

ngokevin: can't we do ```cameraEl.getObject3D('camera')```? because with the el, we can tell exactly which camera it is

meta-meta: ahhh!

dmarcos:
I have the same problem with the linter. On Windows it doesn't output anything. I filed https://github.com/aframevr/aframe/issues/1084

dmarcos:
Thanks! Sound really needs love!

ngokevin:
@dmarcos SQUASH!

dmarcos:
I always forget :S

PR #1081, opened 2016-03-04
Pass the target element to the events emitted by the cursor (fixes #1…

### LINKED ISSUE #1: Add ZIP

dmarcos:
…067)

### DIFF
--- a/tests/components/raycaster.test.js
+++ b/tests/components/raycaster.test.js
@@ -1,0 +1,18 @@
+/* global assert, process, setup, suite, test */
+var entityFactory = require('../helpers').entityFactory;
+
+suite('raycaster', function () {
+  setup(function (done) {
+    var el = this.el = entityFactory();
+    el.setAttribute('raycaster', '');
+    el.addEventListener('loaded', function () {
+      done();
+    });
+  });
+
+  suite('init', function () {
+    test('the raycaster component is initalized', function () {
+      assert.isOk(this.el.components.raycaster);
+    });
+  });
+});
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -50,7 +50,7 @@ module.exports.Component = registerComponent('cursor', {
 
   emit: function (evt) {
     var intersectedEl = this.intersectedEl;
-    this.el.emit(evt);
+    this.el.emit(evt, { target: this.intersectedEl });
     if (intersectedEl) { intersectedEl.emit(evt); }
   },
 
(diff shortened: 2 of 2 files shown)


dmarcos:
As requested by @mayognaise. Is this what you need?

ngokevin:
looks good. initial raycaster tests would be great

mayognaise:
Hi! Is it possible to put this into `0.1.3`?

dmarcos:
@ngokevin Can we cherry pick this commit and make a 0.1.3 release?

ngokevin:
@mayognaise I can provide you a custom build of 0.1.2 with this cherry-picked. We might be able to release next week, and we might clear the blockers preventing you to upgrade as well.

tests/components/raycaster.test.js
```diff
@@ -0,0 +1,24 @@
+/* global assert, process, setup, suite, test */
+var entityFactory = require('../helpers').entityFactory;
+
+suite('raycaster', function () {
+  'use strict';
```
ngokevin: don't think we need the strict since we have a linter

tests/components/raycaster.test.js
```diff
@@ -0,0 +1,24 @@
...
+var entityFactory = require('../helpers').entityFactory;
+
+suite('raycaster', function () {
+  'use strict';
+
+  setup(function (done) {
+    var el = this.el = entityFactory();
+    var sceneEl = document.querySelector('a-scene');
+    // Mocks canvas
+    sceneEl.canvas = document.createElement('div');
```
ngokevin: why do we need to mock the canvas?

tests/components/raycaster.test.js
```diff
@@ -0,0 +1,24 @@
...
+    el.setAttribute('raycaster', '');
+    if (el.hasLoaded) { done(); }
```
ngokevin: don't think we need this check, none of the other tests have it

tests/components/raycaster.test.js
```diff
@@ -0,0 +1,24 @@
...
+    el.addEventListener('loaded', function () {
+      done();
+    });
+  });
+
+  suite('init', function () {
+    test('the raycaster component is initalized', function () {
```
ngokevin: maybe just ```initializes raycaster```

ngokevin:
r+wc we'll do stronger tests later when we improve the raycaster component

tests/components/raycaster.test.js
dmarcos: I like the test to express the expected result

tests/components/raycaster.test.js
dmarcos: that was copied from the cursor test

dmarcos:
Notice that the raycaster test is a bonus. This patch doesn't touch the raycaster but the cursor

tests/components/raycaster.test.js
ngokevin: Yeah, don't have to prepend every test with "the raycaster component" in the context of the raycaster suite though. I like to think of it reading:

```suite('raycaster') suite('init') test('initializes raycaster')``` -> `raycaster.init initializes raycaster`

ngokevin:
looks good

ngokevin:
Thanks for the bonus.

PR #1082, opened 2016-03-04
add onended event to sound component

meta-meta:


meta-meta:
Travis is complaining about an unnecessary call() but I think it's the proper thing to do. Need to call the function with a this ref. Seems more unnecessary to say `sound.unended.bind(sound)()`

ngokevin:
Should probably do the ```bind``` method since that is the intent, and just to please the standard linter.

src/components/sound.js
```diff
@@ -104,6 +104,12 @@ module.exports.Component = registerComponent('sound', {
       });
     }
 
+    sound.source.onended = function () {
+      // we've "overridden" sound's onended binding so call it here
```
ngokevin: If we use ```.bind```, does this comment supply us with any info?

meta-meta:
maybe I'm crazy. I could swear `this` evalutated to `window` when not binding. object.fn.call(object) is indeed equivalent to object.fn()

ngokevin:
Yeah, I was wondering why we couldn't just do ```sound.onEnded```

src/components/sound.js
```diff
@@ -104,6 +104,12 @@ module.exports.Component = registerComponent('sound', {
...
+      sound.onEnded();
+      el.emit('onended');
```
ngokevin: Let's emit ```sound-ended```

src/components/sound.js
ngokevin: Don't think we need this comment then

src/components/sound.js
ngokevin: Can also add this event to the sound docs https://github.com/aframevr/aframe/blob/dev/docs/components/sound.md#properties

Similar to https://github.com/aframevr/aframe/blob/dev/docs/components/material.md#events

dmarcos:
There are conflicts. Needs rebase. We also moved development to master.  @meta-meta Would you mind to reopen the PR against master? Sorry for the inconvenience.

dmarcos:
These changes landed on https://github.com/aframevr/aframe/pull/1106

PR #1090, opened 2016-03-04
We don't want to emit warnings when parsing the data before updating …

dmarcos:
…the schema

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -41,7 +41,10 @@ var Component = module.exports.Component = function (el) {
   var scene;
 
   this.el = el;
-  this.updateSchema(buildData(el, name, this.schema, elData));
+  // The last parameter of builData suppresses the warnings
+  // We don't want to display warning messages when parsing the data
+  // before updating the schema
+  this.updateSchema(buildData(el, name, this.schema, elData, true));
   this.data = buildData(el, name, this.schema, elData);
   this.init();
   this.update();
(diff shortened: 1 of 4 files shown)


src/core/component.js
```diff
@@ -110,12 +110,13 @@ Component.prototype = {
    * If component is single-property, then parses the single property value.
    *
    * @param {string} value - HTML attribute value.
+   * @param {boolean} silent - Suppress warning messages.
```
ngokevin: --> ```Surpress```

src/core/component.js
```diff
@@ -248,8 +249,15 @@ module.exports.registerComponent = function (name, definition) {
  * 3. Attribute data.
  *
  * Finally coerce the data to the types of the defaults.
+ *
+ * @param {object} el - Element we're build data from.
```
ngokevin: --> ```Element to build data from.```

src/core/component.js
dmarcos: Did you mean `suppress`? http://dictionary.reference.com/browse/suppress

ngokevin: I think here would be a good place to explain why we have this flag.

```Generally want to use while swapping the schema like with material shaders.```

ngokevin:
r+wc wfm!

src/core/component.js
ngokevin: Haha, yeah, suppress, oops.

Issue #1097, opened 2016-03-05
add more explicit dependencies for karma-sinon-chai to fix lolex dep issue on tests


legendtang:
When I run `npm test`, here is the output: 

````
> aframe@0.2.0 test E:\github\aframe
> karma start ./tests/karma.conf.js

06 03 2016 01:57:17.760:INFO [preprocessor.env]: Publishing variables:  [ 'TEST_ENV' ]
module.js:341
    throw err;
    ^

Error: Cannot find module 'lolex'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.require.resolve (internal/module.js:23:19)
    at framework (E:\github\aframe\node_modules\karma-sinon-chai\index.js:26:40)
    at Array.invoke (E:\github\aframe\node_modules\di\lib\injector.js:75:15)
    at get (E:\github\aframe\node_modules\di\lib\injector.js:48:43)
    at E:\github\aframe\node_modules\karma\lib\server.js:138:20
    at Array.forEach (native)
    at Server._start (E:\github\aframe\node_modules\karma\lib\server.js:137:21)
    at invoke (E:\github\aframe\node_modules\di\lib\injector.js:75:15)
    at Server.start (E:\github\aframe\node_modules\karma\lib\server.js:102:18)
    at Object.exports.run (E:\github\aframe\node_modules\karma\lib\cli.js:243:26)
    at Object.<anonymous> (E:\github\aframe\node_modules\karma\bin\karma:3:23)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
npm ERR! Test failed.  See above for more details.
````

I noticed it was reported in https://github.com/kmees/karma-sinon-chai/issues/26 and the latest version has fixed the issue. So it is time to bump this dependency version?

nylki:
In the meantime, doing a manual `npm install lolex chai sinon sinon-chai` worked for me to start `karma`/`npm test`.

ngokevin:
@legendtang want to bump karma-sinon-chai?

legendtang:
@ngokevin From what I've seen, the v1.2.0 of karma-sinon-chai doesn't fix that issue since they use `peerDependencies` instead of `dependencies`. As @nylki said, we need to add them as more explicit dependencies.

PR #1099, opened 2016-03-06
add `selectorAll` property type, fix #1049

### LINKED ISSUE #1049: add `selectorAll` property type
Similar to `selector` property type.

Add the convenience of converting the NodeList to an array

legendtang:
I can't do local ci testing using latest node.js and npm. It's reported in https://github.com/aframevr/aframe/issues/1097 and might have been fixed in https://github.com/aframevr/aframe/pull/1098.

So I switched to node v0.12 as online travis does then everything goes fine.


legendtang:
Redundant nodes cleared :)

nylki:
> I can't do local ci testing using latest node.js and npm.

Doing `npm install lolex chai sinon sinon-chai` fixed it for me. Might have to be added to `package.json`.

dmarcos:
@nylki Are you on Windows/OSX? Can you file an issue?

nylki:
@dmarcos Fedora 23. Will file an issue.

dmarcos:
Can you describe a use case for this? Should the ```selector``` type behave like ```selectorAll``` instead of having two separate types? @ngokevin What do you think?

ngokevin:
It's useful when you want a list of entities (class) rather than a single entity (id). But both have distinct cases:

- selectorAll use case was to be able to pass in a class name to whitelist raycaster targets.
- selector use might be to select a single target like ```look-at```.

ngokevin:
This looks good to me.

dmarcos:
A single selector type that returns all the matches  can cover both cases described above: raycaster and look-at. In the case of look-at and multiple matches the component would just pick the first one.

tests/core/propertyTypes.test.js
```diff
@@ -63,6 +63,46 @@ suite('propertyTypes', function () {
     });
   });
 
+  suite('selectorAll', function () {
+    var parse = propertyTypes.selectorAll.parse;
```
dmarcos: Thanks for the tests

ngokevin:
It *would* work, it's better to explicitly define in the schema whether a property is intended to return a single match or multiple matches. Doesn't make sense (and is inconvenient) to have it return an array when you only want a singular match; it'd be akin to defining an int array when you intend to define an int.

src/core/propertyTypes.js
```diff
@@ -88,6 +95,17 @@ function selectorStringify (value) {
   return defaultStringify(value);
 }
 
+function selectorAllStringify (value) {
+  if (value.item) {
+    var els = '';
+    for (var i = 0; i < value.length; ++i) {
```
dmarcos: We can make this loop a bit cleaner:

````javascript
var i;
for (i = 0; i < value.length; ++i) { 
  els +=  '#' + value[i].getAttribute('id');
  if (i !== value.length - 1) { els += ','; }
}

dmarcos:
On more style change and this is ready to go

src/core/propertyTypes.js
ngokevin: Filed https://github.com/aframevr/aframe/issues/1113 for later

legendtang:
Here we go for this! We need a better way to stringify those types indeed.

legendtang:
Close due to dev deprecated.

PR #1110, opened 2016-03-08
Add fit-texture component

nylki:
I propose adding a `fit-texture` component, that automatically scales planes with textures (eg. `a-image`) to fit the image texture without distortions.

Right now, the user has to manually set the width and height if they don't want the image to appear stretched. This is not ergonomic or intuitive, as newcomers to aframe would possibly expect the behaviour of standard html `img` elements, that display images undistorted. Even with the knowledge, it's not really efficient to set dimensions by hand.

The `fit-texture` should be applied to `a-image` by default and can optionally set to other `a-entitites` and may be disabled on `a-image`'s by setting`fit-texture="false"`.
It is possible to set only one side (width or height), and the component will scale the other one accordingly.
If both sides are set, the component will use `width` as base side and emit a warning that it doesn't make sense to set width and height, unless you set `fit-texture="false"`.

I kept thinking wether it is necessary to merge this component into master or not. But I think it adds much needed comfort to use image textures in aframe. Take a look at `examples/primitives/images`!
In future, we might want to approach this differently, but at the moment I thought I'd leave this here for you to discuss/decide :)

### DIFF
--- a/examples/primitives-images/index.html
+++ b/examples/primitives-images/index.html
@@ -14,19 +14,27 @@
         <img id="office3" src="mozvr-office-3.jpg">
       </a-assets>
 
-      <a-image position="-4 -7 -20" width="16" height="4.5" src="#office1"></a-image>
-      <a-image position="10 10 -20" width="14" height="9" scale="0.4 0.4 0.4" src="#office2"></a-image>
-      <a-image position="-6 6 -4" width="14" height="9" scale="0.4 0.4 0.4" src="#office2"></a-image>
-      <a-image position="-2 -5 -10" width="14" height="9" scale="0.4 0.4 0.4" src="#office2"></a-image>
-      <a-image position="-15 -1 -1" width="12" height="8" scale="0.5 0.5 0.5" src="#office3"></a-image>
-      <a-image position="5 -2 -3" width="12" height="8" scale="0.3 0.3 0.3" src="#office3"></a-image>
+      <!--You can define the width (in meters). The height will be set automatically to fit the image properly.  -->
+      <a-image position="-4 -7 -20" width="16" src="#office1"></a-image>
+
+      <!-- Same applies to height: -->
+      <a-image position="10 10 -20" height="6" src="#office2"></a-image>
+
+      <!-- You don't have to use width/height, the image will use default dimensions, you can later scale it via `scale` if you like: -->
+      <a-image position="-6 6 -4" src="#office2"></a-image>
+      <a-image position="-2 -5 -10" src="#office2" scale="8 8 8"></a-image>
+      <a-image position="-15 -1 -1" src="#office3" scale="8 8 8"></a-image>
+
+
+      <!-- If you want to have complete control over the a-image's dimensions you can set `fit-texture="false", like so:` -->
+      <a-image position="5 -2 -3" width="0.5" height="5" fit-texture="false" src="#office3"></a-image>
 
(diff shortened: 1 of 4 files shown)


dmarcos:
@ngokevin Should fit-texture be a property of the flat/standard components instead of a separate one?

ngokevin:
It's very edge case because it only works well for the plane geometry, and is mostly for ```<a-video>```. I don't think this crosses the line for being a standard component, but we make this a third-party component and document this harder.

ngokevin:
Then maybe once we have lazy-loading curated components, this won't even need to be a standard component. Does that sound alright?

nylki:
@ngokevin 
> Then maybe once we have lazy-loading curated components, this won't even need to be a standard component. Does that sound alright?

The lazy-loading for curated components sounds really interesting!
But somehow I still feel images and videos should default to being un-stretched/un-distorted, how ever this may be accomplished in the end (this component or else). But I also understand that you want to keep the core functionality to a certain limit and avoid potential clutter.

PS: With some minor adjustments the component should work for video textures too.

ngokevin:
Yeah, thinking about it again, images/videos on planes will be very common. Gonna rethink this.

dmarcos:
This component seems to only works with primitives that expose a height and width. Two problems:
1. Primitives other than a-image or a-plane might expose height and width that have different semantics.
2. We shouldn't couple components to primitve APIs. Primitive are just aliases for an entity and a set of components.

If we are going to implement a fit-texture it should be implemented at the material component level

nylki:
@dmarcos You are right, it's more problematic with non-rectangular geometries.
> If we are going to implement a fit-texture it should be implemented at the material component level`

I was wondering about that. The *fit-texture* functionality needs to change the geometry (width/height eg), so that the texture fits nicely. Now, If implementing *fit-texture* into *material*,  *geometry* would need to know, wether there is a texture loaded, and if thats the case, use its properties to apply changes. Unless we want to change geometries from inside *material*, which I am not sure about.

Would it therefore not better to make it a part of the `geometry` component, instead of the `material` component? like so:

- if *material* loads a texture, save it (or it's dimensions) somewhere.
- if *geometry* updates at some point, it takes a look if it has a *fit-texture* property/attribute (or registers a listener for texture changes?)
- if *geometry* has *fit-texture*: use the saved texture, to apply transformations, based on the specified geometry (plane, sphere, torus)

What do you think?

ngokevin:
Let me know if you disagree, but I think this is better as a third-party component.

In some cases, you might not want to keep it to proportion. And usually if a component is for convenience and could be done thru other means, it doesn't need to belong in the core.

nylki:
@ngokevin personally I disagree, but that's totally ok :)
Had a lot to do recently and got distracted with other projects and haven't had this PR on my mind anymore. Will do some more fun stuff with aframe again in the future though!

ngokevin:
It's a fine line to decide between usefulness and inclusion to core code. We're thinking more on the stricter side. Gives us less code to maintain, allows module ownership, and with the component system, this can be used by anyone that would like it.

Just a matter of teaching and showcasing to devs about these cool components.

ngokevin:
Would be awesome if you can publish it in a repo or code snippet!

nylki:
@ngokevin 
I understand your reasoning, and in your situation I possibly would have decided the same way.

> Would be awesome if you can publish it in a repo or code snippet!

Sure thing, I put it on my todo list!

nylki:
And here is the NPM package: https://www.npmjs.com/package/aframe-fit-texture-component

PR #1115, opened 2016-03-09
removeState looks up for the state index to be removed

dmarcos:


tests/core/a-entity.test.js
```diff
@@ -98,6 +98,48 @@ suite('a-entity', function () {
...
+    test('it does not add an existing state', function () {
+      var el = this.el;
+      el.states = ['happy'];
+      el.addState('happy');
+      assert.ok(el.states.length === 1);
+    });
+  });
+
+  suite('removeState', function () {
+    test('removes existing state', function () {
```
ngokevin: Can add cases for:

- Removing non-existent state.
- Removing one state when entity has multiple states.

tests/core/a-entity.test.js
```diff
@@ -98,6 +98,48 @@ suite('a-entity', function () {
...
 
+  suite('addState', function () {
+    test('add state', function () {
+      var el = this.el;
+      el.states = [];
+      el.addState('happy');
+      assert.ok(el.states[0] === 'happy');
+    });
+
+    test('it does not add an existing state', function () {
```
ngokevin: Don't need ```it```

tests/core/a-entity.test.js
```diff
@@ -98,6 +98,48 @@ suite('a-entity', function () {
     });
   });
 
+  suite('addState', function () {
+    test('add state', function () {
```
ngokevin: Nit pick ```adds```

src/core/a-entity.js
```diff
@@ -527,8 +527,11 @@ var proto = Object.create(ANode.prototype, {
 
   removeState: {
     value: function (state) {
-      var stateIndex = this.is(state);
-      if (stateIndex === false) { return; }
+      var stateIndex;
+      if (!this.is(state)) { return; }
+      this.states.forEach(function (elState, index) {
```
ngokevin: Can't we just do indexOf?

ngokevin:
r+wc

ngokevin:
Linting issue.

PR #1123, opened 2016-03-09
don't use current already-parsed data when updating component property (fixes #1128)

### LINKED ISSUE #959: obj-model component errors when loading just an obj

### LINKED ISSUE #1128: Issues with setAttribute updating component property with asymmetrical property type.
Consider ```<a-entity obj-model="obj: #obj">```.

The obj-model component data will be after running through schema: ```{ obj: 'somemodel.obj' }```.

Then if you try to do ```entity.setAttribute('obj-model', 'mtl', '#mtl')```, this will try to update the component. This is done internally by grabbing the current data, and attaching the new data.

However, we are using ```getAttribute``` which will return parsed data. So it will try to update with:

```{ obj: 'somemodel.obj', mtl: '#mtl' }``` and the `src` property type will error thinking `somemodel.obj` is a selector. What it should be trying to update with is ```obj: #obj, mtl: #mtl```

### PR DESCRIPTION
**Pull request by ngokevin**
- Add tests for obj-model, which has surfaced some bugs
- setAttribute for single component property didn't work if property type produced asymmetrical value

### CODE REVIEW

**Comment by ngokevin** [comment]
Sorry, I moved some tests around because they were in the wrong suite.

File: src/components/obj-model.js [hunk 1]
```diff
@@ -36,7 +37,7 @@ module.exports.Component = registerComponent('obj-model', {
 
     if (mtlUrl) {
       // .OBJ with an .MTL.
-      if (el.hasAttribute('material')) {
+      if (HTMLElement.prototype.getAttribute.call(el, 'material')) {
```
dmarcos: Why this change?

ngokevin: Because the dependency is ```material=""```

src/core/a-entity.js
```diff
@@ -465,22 +466,23 @@ var proto = Object.create(ANode.prototype, {
...
+      var componentObj = value;  // Deserialized value to send to the component.
+      var componentStr = value;  // Serialized value to send to the DOM.
       var oldValue;
 
       if (component) {
         if (typeof value === 'string' && componentPropValue !== undefined) {
           // Update currently-defined component data with the new property value.
-          partialComponentData = self.getAttribute(attr) || {};
+          partialComponentData = styleParser.parse(
+            HTMLElement.prototype.getAttribute.call(this, attr)) || {};
```
dmarcos: is the ```or``` ( || {} ) condition needed?

ngokevin: I think so, I was getting null

PR #1134, opened 2016-03-10
mixin component props merge with (not replace) entity attribute component props (fixes #999)

### LINKED ISSUE #999: mixin properties should be merged not replaced
I saw these `Primitive not supported: ` errors when I was testing [DrawVR's Tron scene](http://drawvr.com/tron/) against `dev`:

image

Source:

```html
<a-scene>

  <a-assets>
    <a-mixin id="build" geometry="primitive: box" material="shader: standard; metalness: 0.6"></a-mixin>
    <!-- … -->
  </a-assets>

    <!-- … -->
  <a-entity mixin="build" geometry="depth: 150; width: 150; height: 140;" position="110 69 1500" material="src: url(images/building2.png)"></a-entity>
    <!-- … -->

</a-scene>
```

Let's document this caveat in the [Mixin docs](https://aframe.io/docs/core/mixin.html).

ngokevin:
- Introduce a flag whether to set or merge old and new properties.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -77,14 +77,19 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  /**
+   * Implement <a-node>'s applyMixin method.
+   * Set component properties if component not yet attached.
+   * Merge component properties if attached.
+   */
   applyMixin: {
     value: function (attr) {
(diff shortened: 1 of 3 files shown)


ngokevin:
Already fixed. @dmarcos will pull in the tests

PR #1137, opened 2016-03-10
Add .npmignore to stop npm install from downloading tests and examples (fixes #250)

### LINKED ISSUE #250: Consider excluding `examples`, etc. directories in `package.json`
When folks `npm install aframe`, by default, since we haven't defined any specific `directories` to only publish, the entire project (except the `.git/` and `.gitignore`'d files will be picked up + the `devDependencies` won't get installed neither). For folks wanting embedding the `aframe` build stuff, I'd be willing to just not support that case and have them either **clone this repo** and/or use the **forkable boilerplate projects**.

larouxn:
Added a .npmignore that stops `npm install` from downloading the tests and examples folders. Do we still want tests cloned via npm? Do most users embedding A-Frame in their project use the tests? Is there anything besides examples that we want ignored? I suppose the ignoring may be able to be done within the package.json instead.

Note: this does not stop `git clone` from grabbing tests and examples.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+tests
+examples


ngokevin:
r+

dmarcos:
Thanks!

PR #1140, opened 2016-03-10
removed unnecessary parameter from vec2 primitave

roninb:
vec2 only showed up in 4 places in the repo, two were commented out examples and the other two were the switch where Vector2 is created with threejs in src/core/shader and the other is where vec2 is initialized in src/core/propertyType.

I installed and ran aframe after I made changes and opened up a few of the boiler plates and showcase (figuring lightroom would have plenty of shaders in it).

@larouxn also tested saying: "Cloned it myself. Ran the npm test tests, started the server, and checked out a few examples. No errors anywhere as far as I can tell."


ngokevin:
Can you rebase, get rid of those misc commits.

roninb:
Hahahah, looks like I almost got it -_-

dmarcos:
Almost. You need to squash and remove this comment:
https://github.com/aframevr/aframe/pull/1140/files#diff-04c6e90faac2675aa89e2176d2eec7d8R3

ngokevin:
Probably easiest to just start over with a fresh branch since you got a merge commit in there. Cherry-pick the single commit onto a fresh branch and re-PR.

ngokevin:
https://github.com/aframevr/aframe/pull/1143

PR #1142, opened 2016-03-10
fix up .npmignore

cvan:
- `tests` should be `test`
- and added a few more

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,5 @@
-tests
+.DS_Store
+.npmignore
 examples
+node_modules
+test


.npmignore
```diff
@@ -1,2 +1,5 @@
-tests
+.DS_Store
+.npmignore
 examples
+node_modules
+test
```
ngokevin: The folder is called test

cvan: oh weird, I must've been in the wrong project

larouxn:
I would argue .DS_Store and node_modules are unnecessary additions as they're already in the .gitignore and thus can never be committed in the first place. However, .npmignore is a good addition.

PR #1148, opened 2016-03-11
fix release version embedded in Builds snippet

cvan:
(related to #1147)

because of how the `hexo-marked-render` + template variable string interpolation work, the version isn't getting correctly replaced:

screenshot 2016-03-11 02 35 17

<hr><hr>

I tried a [couple hacky attempts](https://gist.github.com/cvan/a97f05260444c5c8d352):

screenshot 2016-03-11 02 29 27

<hr><hr>

while still a hack, this is the one I'm the least disgusted by:

screenshot 2016-03-11 02 36 30

there's probably some proper more clever fix, but I couldn't easily figure one out yet.

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -28,7 +28,7 @@ There are several ways to get started with A-Frame:
 
 The boilerplate contains:
 
-- A simple HTML file that links to the [current version of A-Frame][current] of A-Frame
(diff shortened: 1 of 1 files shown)


cvan:
potentially could be fixed by https://hexo.io/api/rendering.html#after-render-Filters

PR #1150, opened 2016-03-11
Setting on look-controls to allow vertical touch movement

RSpace:
A client recently complained that you could only move horizontally, not vertically, with touch. I see a comment about explicitly only allowing horizontal touch movement, but I don't understand why.

Changes proposed:
- Add a property called allowVerticalTouchMovement to the look-controls component.
- allowVerticalTouchMovement defaults to false.
- When explicitly set to true, vertical touch movement will also work.

I'd be happy to submit another PR where vertical movement is always allowed, but I didn't want to break the defaults when there might be a good reason for it being as it is.

I was unable to find any tests that covers the look-controls component, so I have not included any tests for my change.

### DIFF
--- a/docs/components/look-controls.md
+++ b/docs/components/look-controls.md
@@ -22,9 +22,10 @@ The look-controls component is usually used alongside the [camera component][com
 
 ## Properties
 
-| Property  | Description                        | Default Value |
-|-----------|-----------------------------------------------------
-| enabled   | Whether look controls are enabled. | true          |
+| Property                    | Description                                 | Default Value |
+|-----------------------------|--------------------------------------------------------------
+| enabled                     | Whether look controls are enabled.          | true          |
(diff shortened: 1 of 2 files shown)


dmarcos:
We don't want to enable vertical panning by default on mobile. It's disorienting when combined with the orientation given by the phone sensors. You quickly lose track of what's up or down and rotating the phone becomes frustrating. It's a better experience to constraint panning to the y axis

RSpace:
Fair enough, that's why the PR is adding a setting where it's disabled by
default. Let's at least make it an option, I think it works fine on vrlab.dk
?

ngokevin:
I don't think it will scale to enable/maintain every controls use case under the sun. For now, I recommend forking the controls like you have. And hopefully https://github.com/aframevr/aframe/issues/975 will make it easier to extend the controls.

PR #1151, opened 2016-03-11
Settings on look-controls to configure mouse movement

RSpace:
A client recently complained that mouse movement in A-Frame felt wrong compared to ie. mouse movement in YouTube's and Facebook's 360 videos. He want to "drag" the canvas around with the mouse, rather than to "push" it. He also wanted a slightly faster movement for the mouse, as he felt the experience was a bit "sluggish" with the default setting.

Changes proposed:
- Add a property called reverseMouseMovement to the look-controls component with a default of false. When set to true, mouse movement will work reversed.
- Add a property called mouseMovementSpeed to the look-controls component with a default of 0.002 - the current hardcoded setting.

I was unable to find any tests that covers the look-controls component, so I have not included any tests for my change.

### DIFF
--- a/docs/components/look-controls.md
+++ b/docs/components/look-controls.md
@@ -22,9 +22,11 @@ The look-controls component is usually used alongside the [camera component][com
 
 ## Properties
 
-| Property  | Description                        | Default Value |
-|-----------|-----------------------------------------------------
-| enabled   | Whether look controls are enabled. | true          |
+| Property              | Description                                 | Default Value |
+|-----------------------|--------------------------------------------------------------
+| enabled               | Whether look controls are enabled.          | true          |
+| reverseMouseMovement  | Enable for a "drag feel" to mouse movement. | false         |
+| mouseMovementSpeed    | Speed when looking with mouse (careful!)    | 0.002         |
 
 ## Caveats
 
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -8,7 +8,9 @@ module.exports.Component = registerComponent('look-controls', {
   dependencies: ['position', 'rotation'],
 
   schema: {
-    enabled: { default: true }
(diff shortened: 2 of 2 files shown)


dmarcos:
I made similar changes for the [a-invaders demo](http://swimminglessonsformodernlife.com/a-invaders/) but never submitted them as a PR. For wasd-controls I see use cases where you want to constraint / invert movement on a specific axis or assign WS and AD keys to a different axis:

https://github.com/dmarcos/a-invaders/blob/master/js/components/wasd-controls.js#L20

@RSpace What do you think?

ngokevin:
BTW, while the review process goes down, for clients, feel free to copy-and-paste the look-controls, modify it, register it under a different name, and use it.

RSpace:
@dmarcos I think it would be very useful to have configuration for wasd-controls in. 

@ngokevin Sure, I already did, the solution with both modifications is live at vrlab.dk, and my custom look controls look like this: https://github.com/RSpace/vr-lab/blob/master/src/js/utils/custom-look-controls.js

But I'd prefer not to have duplicate 99% of look-controls next time :)

ngokevin:
Nice. I think the idea of forking/remixing of components is interesting. I can imagine a site where you can write components with a live editor, share them, and remix them. Like CodePen for components.

ngokevin:
I'm going to tag this as needs-dependencies-fixed for #975

RSpace:
@ngokevin What does that mean? What should I do?

ngokevin:
We may be looking to overhaul the controls API, meaning this patch will have to change. Going to keep it open for now in case we want it later.

ngokevin:
I am going to close this for now. With Don's patch, all the code will completely change. Once that lands, we can evaluate adding this. Thanks!

PR #1153, opened 2016-03-11
Fix a small typo in the fading cubes example

K4orta:
Changes proposed:
- Fix a small typo in the Animations docs


dmarcos:
Good catch! Thanks!

PR #1157, opened 2016-03-12
Removes unnecessary rotation applied to the camera absolute position …

### LINKED ISSUE #802: Positional tracking is not properly tracked to orientation of camera.
<a href="https://github.com/caseyyee"></img></a> **Issue by [caseyyee](https://github.com/caseyyee)**
_Friday Jan 08, 2016 at 00:07 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/772_

----

If I rotate the camera 90deg and move forward/back, I am moved left/right.

The position needs to be multiplied by the camera rotation.

dmarcos:
Changes proposed:
-
-
-

…returned by the WebVR APIs. The camera position reported from the HMD is in absolute / world coordinates and can be applied directly to the aframe camera as an offset of the current position (fixes #802)

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -136,23 +136,16 @@ module.exports.Component = registerComponent('look-controls', {
     };
   })(),
 
-  updatePosition: (function () {
-    var position = new THREE.Vector3();
-    var quaternion = new THREE.Quaternion();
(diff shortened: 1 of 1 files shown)


ngokevin:
r+

PR #1145, opened 2016-03-11
Prevents running the mixin attributes through the schema when buildin…

dmarcos:

PR #1159, opened 2016-03-12
Passes the parsed new value to the updateSchema call in updatePropert…

### LINKED ISSUE #1158: When calling updateSchema we are passing the current attribute value and this can be null
We are also parsing the attribute value through the styleParser. The ```updateSchema``` will probably expect the data parsed through the previous schema.

dmarcos:
Changes proposed:
-
-
-

…ies (fixes #1158)

### DIFF
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -216,13 +216,54 @@ suite('Component', function () {
 
     test('extends the schema', function () {
       var TestComponent = registerComponent('dummy', {
-        schema: { color: { default: 'red' } }
+        schema: {color: { default: 'red' }}
       });
       var el = document.createElement('a-entity');
       var component = new TestComponent(el);
       component.extendSchema({ size: { default: 5 } });
       assert.ok(component.schema.size);
       assert.ok(component.schema.color);
+    });
+  });
+
+  suite('updateProperties', function () {
+    setup(function () {
+      components.dummy = undefined;
+    });
+
+    test('updates the schema of a component', function () {
+      var TestComponent = registerComponent('dummy', {
+        schema: {color: { default: 'red' }},
+        updateSchema: function (data) {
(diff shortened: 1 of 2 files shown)


dmarcos:
Fixes the pano-explorer example

tests/core/component.test.js
```diff
@@ -225,4 +225,30 @@ suite('Component', function () {
...
   });
+
+  suite('updateProperties', function () {
+    setup(function () {
+      components.dummy = undefined;
+    });
+
+    test('updates the schema with the new value', function () {
+      var TestComponent = registerComponent('dummy', {
+        schema: { color: { default: 'red' } }
```
ngokevin: shouldn't we test with a component that defines `updateSchema`?

tests/core/component.test.js
```diff
@@ -225,4 +225,30 @@ suite('Component', function () {
...
+      });
+      var el = document.createElement('a-entity');
+      var component = new TestComponent(el);
+      component.updateProperties(null);
+      assert.equal(component.data.color, 'red');
+    });
+
+    test('updates the schema with a null value', function () {
+      var TestComponent = registerComponent('dummy', {
+        schema: { color: { default: 'red' } }
```
ngokevin: going forward, should we stick with spaces-in-objects style? can visually confuse it with block braces, not consistent with arrays, less compact

ngokevin:
I am getting an error with https://github.com/aframevr/aframe/pull/1145 and this PR doesn't fix it.

Mixins + Material = shader not defined: http://codepen.io/team/mozvr/pen/jqMGEq?editors=100

ngokevin:
screen shot 2016-03-13 at 11 40 56 am

ngokevin:
I'll work off of this

PR #1164, opened 2016-03-14
expose registered shaders on AFRAME

ngokevin:
Changes proposed:
- expose AFRAME.shaders


dmarcos:
Why do you want to do this?

ngokevin:
To inspect what shaders are registered or are available, to see if my shader got registered correctly. We are exposing systems and components already.

Issue #692, opened 2016-01-13
Imported models do not work with cursor


ngokevin:
<a href="https://github.com/caseyyee"><img src="https://avatars.githubusercontent.com/u/1560884?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [caseyyee](https://github.com/caseyyee)**
_Sunday Nov 29, 2015 at 20:56 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/554_

----

Imported models don't work with the cursor and no intersections are returned from the raycaster.

ngokevin:
<a href="https://github.com/jcarpenter"><img src="https://avatars.githubusercontent.com/u/1425497?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jcarpenter](https://github.com/jcarpenter)**
_Sunday Nov 29, 2015 at 21:07 GMT_

----

Good catch

ngokevin:
<a href="https://github.com/caseyyee"><img src="https://avatars.githubusercontent.com/u/1560884?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [caseyyee](https://github.com/caseyyee)**
_Monday Nov 30, 2015 at 19:51 GMT_

----

Looks like the cursor returns the element as undefined.

I also found that it depends on the model.   If the model has a grouped geometry in it, you need to cursively search for intersections by passing `true` on recursive argument on `Raycaster.intersectObjects()` otherwise it will only test for intersections on the top-most model.

ngokevin:
workaround is just to create an transparent bounding box and listen for click events on that

PR #1166, opened 2016-03-14
Fix for #692. Iterate up to a parent when matching subobject.

bnolan:
Changes proposed:
- Iterate up object tree to the element that is attached to the dom.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -66,11 +66,26 @@ module.exports.Component = registerComponent('raycaster', {
     var intersectedObjs = this.intersect(scene.children);
     for (var i = 0; i < intersectedObjs.length; ++i) {
       intersectedObj = intersectedObjs[i];
+
+      while (
+        (intersectedObj.object.parent) &&
+        (intersectedObj.object.el === undefined)
+      ) {
+        intersectedObj.object = intersectedObj.object.parent;
+      }
+
       // If the intersected object is the cursor itself
       // or the object is further than the max distance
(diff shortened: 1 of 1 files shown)


src/components/raycaster.js
```diff
@@ -66,11 +66,26 @@ module.exports.Component = registerComponent('raycaster', {
...
+      }
+
       // If the intersected object is the cursor itself
       // or the object is further than the max distance
-      if (intersectedObj.object.el === undefined) { continue; }
-      if (intersectedObj.object.el === cursorEl) { continue; }
-      if (!intersectedObj.object.visible) { continue; }
+
+      if (intersectedObj.object.el === undefined) {
+        continue;
```
dmarcos: Let's keep these as one liners

src/components/raycaster.js
```diff
@@ -66,11 +66,26 @@ module.exports.Component = registerComponent('raycaster', {
     var intersectedObjs = this.intersect(scene.children);
     for (var i = 0; i < intersectedObjs.length; ++i) {
       intersectedObj = intersectedObjs[i];
+
+      while (
+        (intersectedObj.object.parent) &&
```
ngokevin: we can put this if statement on one line without the parens

and comment the while loop ```//  Iterate up to parent when matching subobject.```

dmarcos:
This is almost ready to go. @bnolan a couple of quick changes and we can merge

bnolan:
Ok sweet, made those changes! :+1:

dmarcos:
Thanks!

PR #1173, opened 2016-03-15
trim jsdocs that belong to and are in the docs

ngokevin:
Changes proposed:
- Remove comments that are "user"-facing rather than maintainer-facing.
- User-facing docs exist in one place, in docs/, as source of truth.
- Consistent indent on a-mixin prototype
- No actual code changes
- Don't use "ATnamespace" since we don't do auto-generation of docs

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -17,37 +17,9 @@ var isCoordinate = coordinates.isCoordinate;
  * Animation element that applies Tween animation to parent element (entity).
  * Takes after the Web Animations spec.
  *
- * @namespace <a-animation>
- * @param {string} attribute -
- *   Entity attribute to animate. Can be a component name (e.g., `position`) if the component
- *   is set via a single value. Or can be a dot-separated componentName.componentProp to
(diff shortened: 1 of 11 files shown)


src/core/a-animation.js
```diff
@@ -17,37 +17,9 @@ var isCoordinate = coordinates.isCoordinate;
...
- * @param {number|string} repeat -
- *   How the animation should repeat (e.g., a number or `indefinite`).
- * @param {number} to - End value.
- * @member {number} count -
- *   Decrementing counter for how many cycles of animations left to run.
- * @member {Element} el - Entity which the animation is modifying to.
+ * @member {number} count - Decrementing counter for how many cycles of animations left to
+ *         run.
+ * @member {Element} el - Entity which the animation is modifying.
  * @member initialValue - Value before animation started. Used to restore state.
```
dmarcos: What about these? What's the criteria? Which ones are leaving / staying?

ngokevin: I'm keeping internal properties of the class since that's useful to maintainers to understand the state and such.

I'm removing outward-facing public API documentation.

PR #1174, opened 2016-03-15
Fix iOS viewport issue (Fixes #1122)

### LINKED ISSUE #1122: Fix viewport issue for iOS appview
img_5596

K4orta:
Changes proposed:
- Add maximum-scale=1 to fix iOS viewport scaling (Fixes #1122)

Before:
before fix

After:
after fix

I tested this in the Xcode iOS simulator and an iPhone 6, did not get a chance to test a 6S or newer devices.

### DIFF
--- a/src/core/scene/metaTags.js
+++ b/src/core/scene/metaTags.js
@@ -25,7 +25,7 @@ function injectMetaTags () {
   meta = document.createElement('meta');
   meta.name = 'viewport';
   meta.content =
-    'width=device-width,initial-scale=1,shrink-to-fit=no,user-scalable=no';
+    'width=device-width,initial-scale=1,shrink-to-fit=no,user-scalable=no,maximum-scale=1';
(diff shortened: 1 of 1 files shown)


dmarcos:
Nice! Does this also solve when an a-frame scene is open within the twitter app using its webview?

K4orta:
I'm not sure. I didn't know that was an issue, but I can try to setup a test. Do you know if it is only a problem with Webkit based web views on iOS or if it also affects Twitter's Android app?

dmarcos:
@K4orta I think it's specific to the Webkit based views. But It seems to work in the Facebook app web view.

dmarcos:
I confirm that this fixes the issue on the twitter WebView. Thanks a lot. This bug is a tricky one to figure out unless you're well seasoned in WebKit matters.

ngokevin:
Thanks! Saved me couple hours of my life.

PR #1179, opened 2016-03-16
support inline (and autoplay) videos on iOS when added to Homescreen

cvan:


src/core/scene/a-scene.js
```diff
@@ -147,6 +150,22 @@ var AScene = module.exports = registerElement('a-scene', {
       }
     },
 
+    promptForIOS: {
```
cvan: I'm going to move this logic to its own third-party component because I want to add an "Add to Homescreen" toast message. and, it's way too hairy to add all that logic to A-Frame core IMO.

cvan:
this change enables inline videos to be played _and_ automatically (when `autoplay` is set) when an A-Frame site is added to the homescreen from Safari for iOS. this change also enables any other webviews that want to allow inline playback. do note that inline playback will work only if `webkit-playsinline` is set on the `<video>` elements that are to played.

src/utils/texture.js
```diff
@@ -257,6 +258,12 @@ function fixVideoAttributes (videoEl) {
   if (videoEl.getAttribute('preload') === 'false') {
     videoEl.preload = 'none';
   }
+  if (!videoEl.hasAttribute('crossorigin')) {
+    videoEl.crossOrigin = true;
+  }
+  if (!videoEl.hasAttribute('webkit-playsinline')) {
```
dmarcos: Do we need to check if the attribute is already there? Can we just set it?

cvan: we could

Issue #714, opened 2016-01-13
Add depthTest property in material


ngokevin:
<a href="https://github.com/fernandojsg"><img src="https://avatars.githubusercontent.com/u/782511?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [fernandojsg](https://github.com/fernandojsg)**
_Saturday Dec 05, 2015 at 17:52 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/605_

----

It could be useful when you want to keep things stay on top of everything, for example:
https://github.com/aframevr/aframe-editor/issues/40

Issue #666, opened 2016-01-13
Resolve alpha depth issues automatically


ngokevin:
<a href="https://github.com/jcarpenter"><img src="https://avatars.githubusercontent.com/u/1425497?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jcarpenter](https://github.com/jcarpenter)**
_Friday Nov 20, 2015 at 23:32 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/498_

---

It will be common to superimpose images w/ alpha channels. There is a currently a footgun where order of declaration in the DOM must be carefully managed, or transparent elements will cull each other.

For example, in the AnimeUI test scene in A-Frame, I have a series of transparent images arranged on the z-axis. I must declare them in order of depth, furthest from camera to closest, or I get blending problems.

``` html
<vr-image src="images/schematic1.png" width="7" height="7" position="0 0 -6"></vr-image>
<vr-image src="images/ring2.png" width="3" height="3" position="0 0 -2"></vr-image>
<vr-image src="images/ring1.png" width="0.5" height="0.5" position="0 0 0"></vr-image>
```

This is what it looks like when the user takes the correct steps:

![anime-ui](https://cloud.githubusercontent.com/assets/1425497/11314580/b3768a20-8f9b-11e5-97d4-a49b65d7f3ca.png)

This is problematic because:
- The failure state is obvious and ugly
- The cause will not be obvious to users
- The resolution is sometimes not feasible. The point of A-Frame is grouping elements to fit your scene layout and mental model. The current resolution forces you to group things by z-axis.

We should solve this automatically for users.

ngokevin:
<a href="https://github.com/jcarpenter"><img src="https://avatars.githubusercontent.com/u/1425497?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jcarpenter](https://github.com/jcarpenter)**
_Friday Nov 20, 2015 at 23:34 GMT_

----

I believe Jaume refers to this issue in a recent tweet re: OIT (order independent transparency):

https://twitter.com/mrdoob/status/667835175108112384

Paging @caseyyee who's dealt with this before. Perhaps @fernandojsg also has some suggestions.

ngokevin:
[...]

PR #1185, opened 2016-03-17
Allow setting depthTest to false

masom:
**Description:**
Allow setting the `depthTest` property of the material component.

Fixes https://github.com/aframevr/aframe/issues/714
Relates to https://github.com/aframevr/aframe/issues/666

**Changes proposed:**
- Add `depthTest` in the list of parameters allowed with `material`.
- Apply the `depthTest` value to the material object.

### DIFF
--- a/tests/components/material.test.js
+++ b/tests/components/material.test.js
@@ -272,4 +272,13 @@ suite('material', function () {
       assert.ok(el.getObject3D('mesh').material.transparent);
     });
   });
+
+  suite('depthTest', function () {
+    test('can be set to false', function () {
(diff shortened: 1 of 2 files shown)


tests/components/material.test.js
```diff
@@ -272,4 +272,13 @@ suite('material', function () {
       assert.ok(el.getObject3D('mesh').material.transparent);
     });
   });
+
+  suite('depthTest', function () {
```
dmarcos: thanks for remembering the tests

dmarcos:
it looks great! Thanks!

PR #1191, opened 2016-03-18
Add point to click event

bnolan:
**Description:**

Add uv attribute to detail in click event.

**Changes proposed:**
- Add uv attribute to detail in click event.
- Don't propagate uv attribute up to parent events


bnolan:
I think I got this pull request wrong,

PR #1192, opened 2016-03-18
Add UV attribute to click event.

bnolan:
**Description:**

Add uv attribute to detail in click event.

**Changes proposed:**
- Add uv attribute to detail in click event.
- Don't propagate uv attribute up to parent elements

### DIFF
--- a/examples/test-cursor/index.html
+++ b/examples/test-cursor/index.html
@@ -61,7 +61,8 @@
         var clickedEl = null;
         var els = document.querySelectorAll('a-entity');
         Array.prototype.forEach.call(els, function (el) {
-          el.addEventListener('click', function () {
+          el.addEventListener('click', function (e) {
+            console.log('uv', e.detail.uv);
             if (clickedEl && clickedEl !== el) {
               clickedEl.removeState('selected');
             }
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -70,12 +70,12 @@ module.exports.Component = registerComponent('cursor', {
     var data = this.data;
     var el = evt.detail.el;
     var distance = evt.detail.distance;
-    if (this.intersectedEl === el) { return; }
-    if (distance >= this.data.maxDistance) { return; }
-    this.intersectedEl = el;
     this.intersectionEvent = {
       uv: evt.detail.uv
     };
+    if (this.intersectedEl === el) { return; }
+    if (distance >= this.data.maxDistance) { return; }
+    this.intersectedEl = el;
     el.addState('hovered');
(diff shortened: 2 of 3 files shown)


examples/test-cursor/index.html
```diff
@@ -61,7 +61,8 @@
         var clickedEl = null;
         var els = document.querySelectorAll('a-entity');
         Array.prototype.forEach.call(els, function (el) {
-          el.addEventListener('click', function () {
+          el.addEventListener('click', function (e) {
+            console.log('uv', e.detail.uv);
```
dmarcos: Remove log

src/components/cursor.js
```diff
@@ -50,8 +50,13 @@ module.exports.Component = registerComponent('cursor', {
 
   emit: function (evt) {
     var intersectedEl = this.intersectedEl;
-    this.el.emit(evt, { target: this.intersectedEl });
-    if (intersectedEl) { intersectedEl.emit(evt); }
+    var dispatchEvent = Object.assign({}, this.intersectionEvent, {
```
dmarcos: Why do we need to copy this.intersectionEvent?

src/components/cursor.js
```diff
@@ -65,6 +70,9 @@ module.exports.Component = registerComponent('cursor', {
     var data = this.data;
     var el = evt.detail.el;
     var distance = evt.detail.distance;
+    this.intersectionEvent = {
```
dmarcos: Keep one liner

dmarcos:
Thanks! A few changes + squash and we ship this :)

src/components/raycaster.js
```diff
@@ -85,9 +85,9 @@ module.exports.Component = registerComponent('raycaster', {
   /**
    * Remembers the last intersected element
    */
-  setExistingIntersection: function (el, distance) {
+  setExistingIntersection: function (el, distance, uv) {
     this.intersectedEl = el;
-    this.el.emit('intersection', { el: el, distance: distance });
+    this.el.emit('intersection', { el: el, distance: distance, uv: uv });
```
ngokevin: I would just get rid of distance/uv and return the whole intersection object.

src/components/cursor.js
cvan: we have an npm package that polyfills `Object.assign` where it's not natively supported. to use it, call `utils.extend`.

ngokevin:
I plan on handling everything at https://github.com/aframevr/aframe/pull/1196, such as returning full intersection data. Thanks for the PR though! Sorry for your time.

PR #1196, opened 2016-03-19
improve raycaster component (fixes #829)

### LINKED ISSUE #829: Investigate slow raycasting
From @brianchirls on [WebVR Slack](https://webvr.slack.com/messages/help-and-bugs/):

> __brianchirls [5:53 AM]__
> @joshcarpenter et al: I ran a quick CPU profile of a-frame, and it looks like raycasting is the slowest thing. I’ll be working on my own optimized solution for that today and will let you know what I find out. But…a good way to speed it up is to turn off raycasting for photo spheres, since they’ll ​*always*​ show up wherever you look. Or even exclude any objects that don’t have an event listener or something on them.

ngokevin:
**Description:** Improve overall cursor + raycaster components with respect to performance, configuration, and code quality. This is a stepping stone for more complex cursor/interactions later on.

**[DEMO](http://ngokevin.github.io/aframe/examples/test-raycaster/)**

**Changes proposed:**
- Don't use requestInterval which adds an additional requestAnimationFrame call. Work off of the existing tick handler.
- Generalize raycaster component for general use rather than just the cursor. Allow specifying the near, far, recursive. Thus decoupling raycaster and cursor, and having cursor be an abstraction of the raycaster.
- Raycaster origin/direction can be set with position/rotation of entities
- Allow passing in a list of entities (via selector) that the raycaster should intersect rather than intersecting everything
- Allow passing in a list of entities to ignore
- Return full intersection data in events
- Move intersectioncleared/mouseleave logic to the cursor component to remove cursor/raycaster coupling
- Be able to update the raycaster 
- Prefix event/state names with respective component (```cursor-```/```raycaster-```)
- Removed `cursor.maxDistance` as it was redundant with `raycaster.far` 
- Fully unit test the raycaster and cursor components


amberroy:
Cool, I was also thinking it would be useful to get the full intersection data in events. I got it working in my cursor-hitpoint branch, with an example: https://github.com/amberroy/aframe/commit/593ab4c

ngokevin:
Yeah, returning full intersection data is part of the changes in that huge list.

examples/test-cursor/index.html
```diff
@@ -10,58 +10,61 @@
     <a-scene stats="true">
       <a-assets>
         <a-mixin id="cube" geometry="primitive: box"></a-mixin>
-        <a-mixin id="cube-hovered" material="color: magenta"></a-mixin>
-        <a-mixin id="cube-selected" material="color: cyan"></a-mixin>
-        <a-mixin id="red" material="color: red"></a-mixin>
-        <a-mixin id="green" material="color: green"></a-mixin>
-        <a-mixin id="blue" material="color: blue"></a-mixin>
-        <a-mixin id="yellow" material="color: yellow"></a-mixin>
+        <a-mixin id="cube-hovered" material="color: #CC435F"></a-mixin>
```
dmarcos: Why are you replacing color names with hex values?

src/components/cursor.js
```diff
@@ -1,94 +1,162 @@
 var registerComponent = require('../core/component').registerComponent;
+var THREE = require('../lib/three');
 var utils = require('../utils/');
 
+var EVENTS = {
+  CLICK: 'cursor-click',
```
dmarcos: Why renaming to cursor click?

ngokevin: So you can tell which event came from which component. It emphasizes that these aren't special core events, just events provided by a component.

examples/test-cursor/index.html
ngokevin: Color palette.

src/components/cursor.js
dmarcos: it should be automated. People will forget to prepend the component name to the emitted event

ngokevin: Possibly. It may be unexpected for component developers to emit an event, try to add a listener, and not knowing that it's prefixing under-the-hood.

ngokevin:
@dmarcos r?

- [Example](http://ngokevin.github.io/aframe/examples/test-raycaster/)
- [Example Code](https://github.com/aframevr/aframe/pull/1196/files#diff-37c2905af9feb5afcd95d3cdca3cf5d1R1)

examples/boilerplate-panoexplorer/index.html
```diff
@@ -24,21 +24,19 @@
...
                  fill="backwards" from="0.1 0.1 0.1" to="1 1 1" dur="150"></a-animation>
-        <a-mixin id="fuse-animation" begin="fusing" easing="ease-in" attribute="scale"
-                 fill="backwards" from="1 1 1" to="0.1 0.1 0.1" dur="1500"></a-mixin>
+        <a-mixin id="fuse-animation" begin="cursor-fusing" easing="ease-in" attribute="scale"
+                 fill="forwards" from="1 1 1" to="0.1 0.1 0.1" dur="1500"></a-mixin>
       </a-assets>
 
       <!-- Camera + Cursor. -->
       <a-entity camera look-controls>
-        <a-entity mixin="cursor" position="0 0 -3">
```
dmarcos: we should not be using primitives here.

src/components/raycaster.js
```diff
@@ -1,112 +1,138 @@
...
+  updateOriginDirection: function () {
+    var el = this.el;
+    var object3D = el.object3D;
 
-  handleIntersection: function (obj) {
-    var el = obj.object.el;
+    // Update matrix world.
+    object3D.updateMatrixWorld();
+    // Grab the position and rotation.
+    object3D.matrixWorld.decompose(this.originVec3, this.directionHelper, scaleDummy);
```
dmarcos: the direction helper does not need to be part. We have the following pattern for these cases:
https://github.com/aframevr/aframe/blob/master/src/components/look-controls.js#L111

dmarcos: the same applies for ```originVec3```

src/components/raycaster.js
```diff
@@ -1,112 +1,138 @@
...
+    recursive: {default: true}
+  },
+
   init: function () {
+    this.direction = new THREE.Vector3();
+    this.directionHelper = new THREE.Quaternion();
+    this.intersectedEls = [];
+    this.objects = null;
+    this.originVec3 = new THREE.Vector3();
+    this.prevCheckTime = null;
```
dmarcos: prevCheckTime is a number not an object. we should not use ```null``` as initial value.

src/components/cursor.js
```diff
@@ -1,94 +1,138 @@
...
-    this.schema.fuse.default = utils.isMobile();
-    this.attachEventListeners();
-  },
-
-  attachEventListeners: function () {
-    var el = this.el;
-    var canvas = el.sceneEl.canvas;
+    var cursorEl = this.el;
+    var canvas = cursorEl.sceneEl.canvas;
+    this.fuseTimeout = null;
```
dmarcos: fuseTimeout is not an object. Use ```undefined``` as initial value

src/core/scene/a-scene.js
```diff
@@ -281,6 +281,30 @@ var AScene = module.exports = registerElement('a-scene', {
     },
 
     /**
+     * Behavior-updater meant to be called from scene render.
+     * Abstracted to a different function to facilitate unit testing (`scene.tick()`) without
+     * needing to render.
+     */
+    tick: {
```
dmarcos: that's good

examples/boilerplate-panoexplorer/index.html
ngokevin: This is gonna be the boilerplate. `<a-cursor>` is one of the times i prefer the primitive because manually configuring the inner/outer radius of the ring is hard.

src/components/raycaster.js
ngokevin: You could have many raycasters in the scene, and there's a likely chance they'll be overwriting each other's vectors and cause issues. `look-controls` is fine since you usually only have one of them.

src/components/raycaster.js
ngokevin: `null` means it's never been checked. `0` would work, but it'd be incorrect.

dmarcos: it should be ```undefined``` then. null is used when a variable that should point to an object points to nothing. In this case prevCheckTime is a number not an object so the absence of a value is ```undefined```

src/components/raycaster.js
dmarcos: It doesn't matter that directionHelper is used by multiple raycaster because you only use the variable during the duration of the function call. There won't be multiple instances of the same function running at the same time and they cannot interrupt each other.

src/components/raycaster.js
```diff
@@ -1,112 +1,138 @@
...
+  updateOriginDirection: function () {
+    var el = this.el;
+    var object3D = el.object3D;
 
-  handleIntersection: function (obj) {
-    var el = obj.object.el;
+    // Update matrix world.
+    object3D.updateMatrixWorld();
+    // Grab the position and rotation.
+    object3D.matrixWorld.decompose(originVec3, directionHelper, scaleDummy);
```
dmarcos: I would still make the variable private to this function so no other method relies on it to store any state. Like in this example:
https://github.com/aframevr/aframe/blob/master/src/components/look-controls.js#L111

dmarcos: We should wrap directionHelper, originVec3 and scaleDummy in a closure of the method so they're not allocated on every frame. Like the example I had pointed out: 
https://github.com/aframevr/aframe/blob/master/src/components/look-controls.js#L111

src/components/raycaster.js
```diff
@@ -1,112 +1,142 @@
...
-    // and now needs a mouseleave event.
-    if (this.intersectedEl !== el) {
-      this.clearExistingIntersection();
-    }
-    this.setExistingIntersection(el, obj.distance);
-  }
+      // Update matrix world.
+      object3D.updateMatrixWorld();
+      // Grab the position and rotation.
+      object3D.matrixWorld.decompose(originVec3, directionHelper, scaleDummy);
```
dmarcos: You can add scaleDummy to the closure as well.

ngokevin: I'm not actually using `scaleDummy`. I just pass it because `decompose` will expect it.

PR #1205, opened 2016-03-20
pause sound only if sound is playing

meta-meta:
**Description:**
When destroying a sound component, we pause the sound which throws an error when the sound is not playing.
**Changes proposed:**
-Check if sound is playing before attempting to pause.

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -119,7 +119,7 @@ module.exports.Component = registerComponent('sound', {
   },
 
   pause: function () {
-    if (!this.sound.source.buffer) { return; }
+    if (!this.sound.source.buffer || !this.sound.isPlaying) { return; }
     this.sound.pause();
   }
 });


src/components/sound.js
```diff
@@ -120,6 +120,8 @@ module.exports.Component = registerComponent('sound', {
 
   pause: function () {
     if (!this.sound.source.buffer) { return; }
```
dmarcos: we can do 
````javascript
if (!this.sound.source.buffer || !this.sound.isPlaying) { return; }
this.sound.pasue();
````

**Comment by ngokevin** [comment]
Similarly, might also want to check if sound is already playing before playing

**Comment by meta-meta** [comment]
@ngokevin good call, but THREE already handles play:
```
THREE.Audio.prototype.play = function () {

	if ( this.isPlaying === true ) {

		console.warn( 'THREE.Audio: Audio is already playing.' );
		return;

	}
```

**Comment by ngokevin** [comment]
In `tests/components/sound.test.js`, can we add a small test?:

```js
suite('pause', function () {
  test('checks if playing before pausing', function () {
    var el = this.el;
    el.pause();
    el.pause();
  });
});
```

ngokevin:
I can add the test.

dmarcos:
@ngokevin You should let contributors handle that. PRs have to come with tests or they won't merge.

PR #1206, opened 2016-03-20
Fix cursor example

amberroy:
- Fix query selector to match all cubes, by using wildcard (*)
- Also upgrade to using a-camera and a-cursor tags, instead of depreciated attributes

### DIFF
--- a/examples/test-cursor/index.html
+++ b/examples/test-cursor/index.html
@@ -73,7 +73,7 @@
 
         // If you want to respond to mouse events, this is how. Otherwise,
         // if you just want a working cursor, you don't need any of this.
-        var cubes = document.querySelectorAll('a-entity[mixin=cube]');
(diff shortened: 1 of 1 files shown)


ngokevin:
Probably keep using a-entity. Those aren't deprecated. a-camera and a-cursor are helpers and we'd rather have these examples not be testing that layer

ngokevin:
Thanks!

dmarcos:
a-camera and a-cursor are just aliases of a-entities + components. These test examples illustrate how to configure a camera and a cursor using components. If we use primitives / aliases we lose the learning value.

amberroy:
Understood. I reverted back to a-entity, now it's just a one-liner fixing the query selector (which wasn't matching anything before).

dmarcos:
@amberroy nice! thanks! can you squash your commits?

amberroy:
squashed.

PR #1208, opened 2016-03-20
Windows cmd.exe does not have command substitution and standard-reporter

dmarcos:
**Description:**

Fix linter on Windows

**Changes proposed:**
Removes the explicit list of files passed to the linter via command substitution. The feature is not available on cmd.exe. Non git tracked files will be parsed by the linter. We need to find an alternative to reduce noise but in the short term is better to have a linter working on Windows.
It also replaces standard-reporter with snazzy. standard-reporter fails silently and snazzy that was broken on Windows in the past seems to be working now.

@cvan

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "dist": "mkdirp dist/ && npm run browserify -s -- --debug | exorcist dist/aframe.js.map > dist/aframe.js && uglifyjs dist/aframe.js -c warnings=false -m -o dist/aframe.min.js",
     "gh-pages": "npm run ghpages",
     "ghpages": "node ./scripts/gh-pages",
-    "lint": "semistandard -v $(git ls-files '*.js') | standard-reporter --stylish",
+    "lint": "semistandard -v | snazzy",
     "precommit": "npm run lint",
     "preghpages": "npm run dist && rimraf gh-pages && mkdirp gh-pages && cp -r {.nojekyll,dist,lib,examples,index.html,style} gh-pages/. 2>/dev/null || : && git checkout dist/ && replace 'build/aframe.js' 'dist/aframe.min.js' gh-pages/ -r --silent",
     "release:bump": "npm run dist && git commit -am 'bump dist' && npm version patch --preminor",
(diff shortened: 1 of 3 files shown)


ngokevin:
The linter is not excluding unrelated files correctly.

package.json
```diff
@@ -64,7 +64,7 @@
     "semistandard": "^7.0.2",
     "sinon": "^1.17.3",
     "sinon-chai": "^2.8.0",
-    "standard-reporter": "^1.0.5",
+    "snazzy" : "3.0.0",
```
cvan: don't you want `^3.0.0`?

package.json
```diff
@@ -10,7 +10,7 @@
     "dist": "mkdirp dist/ && npm run browserify -s -- --debug | exorcist dist/aframe.js.map > dist/aframe.js && uglifyjs dist/aframe.js -c warnings=false -m -o dist/aframe.min.js",
     "gh-pages": "npm run ghpages",
     "ghpages": "node ./scripts/gh-pages",
-    "lint": "semistandard -v $(git ls-files '*.js') | standard-reporter --stylish",
+    "lint": "semistandard -v | snazzy",
```
cvan: yeah, this should be fine, since we're already excluding quite a few files for `semistandard`

dmarcos:
@ngokevin what do you mean by non related files? I mentioned in the comment that with this change the linter will parse the files not tracked by git. we have to find away to exclude those files that is not bash specific like the current command substition.

ngokevin:
I mean Travis is broken.

cvan:
> @ngokevin what do you mean by non related files? I mentioned in the comment that with this change the linter will parse the files not tracked by git. we have to find away to exclude those files that is not bash specific like the current command substition.

you can simply add [the files not tracked by git](https://github.com/aframevr/aframe/blob/master/.gitignore) to the [`semistandard .ignore` rules in the `package.json`](https://github.com/aframevr/aframe/blob/383f20b/package.json#L82-L90).

and, fwiw, [the `node_modules` directory](https://github.com/Flet/semistandard#ignoring-files) is already ignored. so only a few more to add.

PR #1064, opened 2016-03-02
fix unhandled edge cases for keyboard shortcuts (fixes #927, fixes #932, fixes #1063)

cvan:

PR #1211, opened 2016-03-21
fix unhandled edge cases for keyboard shortcuts (fixes #927, fixes #932, fixes #1063)

### LINKED ISSUE #927: WASD controls should stop on document blur
I noticed this a while back but kept forgetting to file.

STR:

1. Load https://aframe.io/aframe/examples/showcase-spheres-and-fog/ and stay in mono view
2. Hold down the `A` key
3. Whilst holding the `A` key, move your cursor outside of your browser window
4. Notice you are still moving left

### LINKED ISSUE #932: WASD controls should not be engaged w/ shift/control/alt/meta keys
(Similar but different from #927.)

STR:

1. Load https://aframe.io/aframe/examples/showcase-spheres-and-fog/
2. Hold down the `Command` + `A` keys
3. Release the keys
4. Notice you keep moving left

For WASD controls, we should abort if `shiftKey`, `ctrlKey`, `altKey`, or `metaKey` are also being held.

### LINKED ISSUE #1063: Keypresses in form fields should not be considered keypresses for keyboard shortcuts + WASD controls
1. Place a form field somewhere on the page (outside of the `<a-scene>` of course): `<input style="position:absolute;z-index:1">`
2. Start typing these keys: `w`, `a`, `s`, `d`, `f`, `z`

cvan:


src/utils/index.js
```diff
@@ -131,14 +131,25 @@ var isIOS = module.exports.isIOS = function () {
...
  */
 module.exports.isLandscape = function () {
   return window.orientation === 90 || window.orientation === -90;
 };
 
 /**
+ * Returns whether we can safely listen to this keyboard event.
+ * @returns {Boolean} True if we can.
+ */
+module.exports.shouldCaptureKeyEvent = function (event) {
```
cvan: renamed per @dmarcos' suggestion in the previous PR: #1064

src/components/wasd-controls.js
```diff
@@ -98,21 +100,29 @@ module.exports.Component = registerComponent('wasd-controls', {
...
 
   removeEventListeners: function () {
     // Keyboard events
+    window.removeEventListener('blur', this.onBlur);
     window.removeEventListener('keydown', this.onKeyDown);
     window.removeEventListener('keyup', this.onKeyUp);
   },
 
+  onBlur: function (event) {
+    this.keys = {};
```
cvan: calling `this.pause()` does not work. must do this.

cvan: I will change the `pause` method to clear `this.keys`.

src/components/wasd-controls.js
```diff
@@ -98,21 +101,29 @@ module.exports.Component = registerComponent('wasd-controls', {
...
   },
 
   removeEventListeners: function () {
     // Keyboard events
+    window.removeEventListener('blur', this.onBlur);
     window.removeEventListener('keydown', this.onKeyDown);
     window.removeEventListener('keyup', this.onKeyUp);
   },
 
+  onBlur: function () {
```
dmarcos: Should we move this to the component module and always call ```pause``` on blur for every component?

cvan: interesting idea; probably actually, yeah. you think that's good?

dmarcos: When is the blur event exactly triggered?

src/utils/index.js
```diff
@@ -131,14 +131,25 @@ var isIOS = module.exports.isIOS = function () {
...
- * @return {Boolean} True if landscape orientation
+ * Checks mobile device orientation.
+ * @return {Boolean} True if landscape orientation.
  */
 module.exports.isLandscape = function () {
   return window.orientation === 90 || window.orientation === -90;
 };
 
 /**
+ * Returns whether we can safely listen to this keyboard event.
```
ngokevin: Could expand on what is "safely".

src/utils/index.js
```diff
@@ -131,14 +131,25 @@ var isIOS = module.exports.isIOS = function () {
...
+ * @returns {Boolean} True if we can.
```
ngokevin: s/True/true

src/components/wasd-controls.js
cvan: when the current tab is no longer the focused tab - or if the window loses focus

src/utils/index.js
cvan: it's capitalised because a sentence - and we do elsewhere. I would say `true`. will change.

src/components/wasd-controls.js
ngokevin: Hard to say whether to want to force pause on blur. Seems too prescriptive at the moment.

src/utils/index.js
ngokevin: just brought it up because `true` is a value

dmarcos:
@cvan What's the state of this. Is this ready to go? Let's use the ```needs review``` label otherwise it's hard to tell when PRs are ready.

src/components/wasd-controls.js
```diff
@@ -98,21 +101,29 @@ module.exports.Component = registerComponent('wasd-controls', {
 
   attachEventListeners: function () {
     // Keyboard events
-    window.addEventListener('keydown', this.onKeyDown, false);
-    window.addEventListener('keyup', this.onKeyUp, false);
+    window.addEventListener('blur', this.onBlur);
```
dmarcos: Should we also have a listener for the symmetric event.  ```window.addEventListener("focus", this.onFocus);```?

cvan: oh, very good call.

hmm, the more I think about it we probably shouldn't call `pause` on `blur` because that [removes the event listeners](https://github.com/aframevr/aframe/pull/1211/files#diff-6231789e7e7fac0fbea47834d09b3bdcR91). and then we'd have to reattach them. seems like maybe a waste. thoughts?

cvan:
@dmarcos updated to handle blur/focus/visibilitychange pause/play correctly. good catch, thanks

PR #1220, opened 2016-03-22
fix obj model component for multiple models (fixes #1194)

### LINKED ISSUE #1194: OBJ loader concurrency problem due to shared object
**Description:**
The OBJ loader uses a shared `THREE.OBJLoader` and `THREE.MTLLoader` in `obj-model.js`.

```
    var objLoader = new THREE.OBJLoader();
    var mtlLoader = new THREE.MTLLoader(objLoader.manager);
```

Those two objects are not safe to use concurrently and result in unexpected behaviours.

Moving the initialization to the `loadOBj` method solves the shared state issue.

- A-Frame Version: https://github.com/aframevr/aframe/commit/c4e2d794c901ed9aa3a349184e23cb661191d278 / Master and bellow.

- Platform/Device: N/A

ngokevin:
**Description:** Fix having multiple OBJ models in a scene.

screen shot 2016-03-22 at 1 16 24 pm

**Changes proposed:**
- Make objloader/mtlloader not behave as singletons


tests/components/obj-model.test.js
```diff
@@ -32,12 +28,37 @@ suite('obj-model', function () {
...
+  test('can load .OBJ + .MTL', function (done) {
     var el = this.el;
+    var handled = false; // Event listener is not getting torn down for some reason.
     el.addEventListener('model-loaded', function () {
+      if (handled) { return; }
+      handled = true;
       assert.ok(el.components['obj-model'].model);
       done();
     });
-    el.setAttribute('obj-model', { mtl: '#mtl', obj: '#obj' });
```
cvan: :+1:

PR #1228, opened 2016-03-22
Added type=color to schemas

fernandojsg:
**Description:** Explicitly added ```type='color'``` for every color parameters in the components' schemas.

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -12,8 +12,8 @@ var warn = debug('components:light:warn');
 module.exports.Component = registerComponent('light', {
   schema: {
     angle: { default: 60, if: { type: ['spot'] } },
-    color: { default: '#FFF' },
-    groundColor: { default: '#FFF', if: { type: ['hemisphere'] } },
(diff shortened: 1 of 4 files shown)


ngokevin:
Want to make the default value of the `color` property type, `#FFF`? Then we can simply do `{type: 'color'}`.

fernandojsg:
Yes, I think ```#FFF``` could be nice as default.

dmarcos:
is white a nice default or better a color that pops out so it's useful for debugging?

ngokevin:
If you have a different default, then every mesh with a texture you have to define white, else the color will blend with the texture.

fernandojsg:
Yep, and also in the lights you will need to change to white too

ngokevin:
We can change here, and then that will be the default default: https://github.com/aframevr/aframe/blob/master/src/core/propertyTypes.js#L11

dmarcos:
let's do white as default then.

PR #1236, opened 2016-03-23
fix modal for changing orientation when in portrait in VR mode (fixes #1213)

### LINKED ISSUE #1213: Instructions for cardboard/landscape are missing when entering VR
**Description:**
Instructions on rotating the device into landscape / into cardboard are now missing.

- A-Frame Version: master
- Platform/Device: Android, iOS

cvan:


src/components/scene/vr-mode-ui.js
```diff
@@ -81,12 +57,74 @@ module.exports.Component = registerComponent('vr-mode-ui', {
...
+    this.toggleEnterVRButtonIfNeeded();
+    this.toggleOrientationModalIfNeeded();
+  },
+
+  onExitVR: function () {
+    this.toggleEnterVRButtonIfNeeded();
+    this.hideOrientationModal();
+  },
+
+  hideEnterVRButton: function () {
```
cvan: do you think we need these abstractions, or should I just move these two lines to `toggleEnterVRButtonIfNeeded`?

src/components/scene/vr-mode-ui.js
```diff
@@ -81,12 +57,74 @@ module.exports.Component = registerComponent('vr-mode-ui', {
...
     this.enterVREl.classList.add(HIDDEN_CLASS);
   },
 
-  show: function () {
+  showEnterVRButton: function () {
+    if (!this.enterVREl) { return; }
     this.enterVREl.classList.remove(HIDDEN_CLASS);
+  },
+
+  toggleEnterVRButtonIfNeeded: function () {
```
cvan: taking better name recommendations heh

src/components/scene/vr-mode-ui.js
```diff
@@ -81,12 +57,74 @@ module.exports.Component = registerComponent('vr-mode-ui', {
...
+  },
+
+  onEnterVR: function () {
+    this.toggleEnterVRButtonIfNeeded();
+    this.toggleOrientationModalIfNeeded();
+  },
+
+  onExitVR: function () {
+    this.toggleEnterVRButtonIfNeeded();
+    this.hideOrientationModal();
```
cvan: I could call `this.toggleOrientationModalIfNeeded();` but figured it's more explicit

src/components/scene/vr-mode-ui.js
```diff
@@ -81,12 +57,74 @@ module.exports.Component = registerComponent('vr-mode-ui', {
...
+    window.addEventListener('orientationchange', this.toggleOrientationModalIfNeeded.bind(this));
+    this.toggleOrientationModalIfNeeded();
+  },
+
+  onEnterVR: function () {
+    this.toggleEnterVRButtonIfNeeded();
+    this.toggleOrientationModalIfNeeded();
+  },
+
+  onExitVR: function () {
```
cvan: I could combine this with `onEnterVR` to be `onVRModeChange` or something

src/components/scene/vr-mode-ui.js
```diff
@@ -176,11 +214,8 @@ function createOrientationModal (exitVRHandler) {
   var exit = document.createElement('button');
   exit.innerHTML = 'Exit VR';
 
-  // Hide modal and exit VR on close.
-  exit.addEventListener('click', function () {
-    exitVRHandler();
-    modal.classList.add(HIDDEN_CLASS);
```
cvan: this gets handled by `onExitVR` now

src/components/scene/vr-mode-ui.js
```diff
@@ -71,6 +45,8 @@ module.exports.Component = registerComponent('vr-mode-ui', {
 
     this.orientationModalEl = createOrientationModal(this.exitVR);
     this.el.appendChild(this.orientationModalEl);
+
+    this.attachEventListeners();
```
cvan: this has to happen here because we want `this.enterVREl` and `this.orientationModalEl` to exist

src/components/scene/vr-mode-ui.js
dmarcos: i like small methods. easier to understand and test

src/components/scene/vr-mode-ui.js
dmarcos: I usually don't have issues with verbose names. Less space for missinterpretation and autocompletion in your editor helps with the typing

src/components/scene/vr-mode-ui.js
ngokevin: I would move the creation (without appending) of the ```Els``` into ```init``` so that we can get the event listeners attached in ```init```. The danger of this is that if a property changes, then we will end up registering duplicate listeners.

ngokevin: Then we don't need a separate method that only gets called once and pollutes the prototype/public API

src/components/scene/vr-mode-ui.js
ngokevin: On the other hand, we only want to split out to other methods if we re-use them a lot, plan to unit test them, or if we want it to be public API. Else we create a lot of indirection in the module, and it's hard to keep track of the data flow.

src/components/scene/vr-mode-ui.js
ngokevin: Yeah, I think while `toggle` would be less code, it's easier to manage explicitly and possibly less prone to bugs.

ngokevin:
Skimming through, I think we should get rid of the redundant/explicit ```hide/show``` methods since we already added toggle methods, and those methods abstract only one LOC. Just encapsulate them inside the toggle.

src/components/scene/vr-mode-ui.js
cvan: I need to move the creation and and insertion before toggleOrientationModalIfNeeded gets called. I could listen for some event from when we insert in the `update` method. Why's it done in that method though? I'm curious.

Also, I've seen attachEventListeners methods in other prototypes. I know you try to minimize public API surface area, which is good.

ngokevin: Can't do:

```
init
  Set up variables
  Create elements
  Attach listeners
update
  Insert elements
```

dmarcos: the public API is what is documented. I'm fine with having methods that are not "public" in the prototype. we can flag it as so. and if it's really something we don't want to be called we can make private functions. we already have a bunch of those. Small, easy to understand and testable methods that encourage good hygiene is the priority.

ngokevin: Not the only reason though. We don't want to create unnecessary abstractions:

> we only want to split out to other methods if we re-use them a lot, plan to unit test them, or if we want it to be public API. Else we create a lot of indirection in the module, and it's hard to keep track of the data flow.

dmarcos: it's not hard to keep track of the data flow if you name the methods appropriately.  Having all the code exposed with no encapsulation is what makes hard the code to understand because there are more things you have to keep in your short term memory. local variables, different paths in the control flow

cvan: @ngokevin: https://github.com/aframevr/aframe/pull/1236/files#diff-2e99cd43fe5dd47f035f54a4fcc06f44R77 is the line I care about. I want to call that _after_ update. I'm moving things around to make this happen.

cvan:
thanks for feedback! updated

src/components/scene/vr-mode-ui.js
```diff
@@ -71,6 +60,8 @@ module.exports.Component = registerComponent('vr-mode-ui', {
 
     this.orientationModalEl = createOrientationModal(this.exitVR);
     this.el.appendChild(this.orientationModalEl);
+
+    this.onVRModeChange();
```
dmarcos: What about calling this method something like 'updateEnterVRInterface' instead of being bound to the event name is attached to?

cvan: I like that; done

PR #1241, opened 2016-03-23
increased default segments for circle primitives

roninb:
Changed from 8 to 32. Also changed the documentation for geometry to reflect changes in src.


dmarcos:
Thanks!

ngokevin:
@roninb As a follow up, can you also increase the other default segments?

Issue #1216, opened 2016-03-21
increase default segments in geometry component


ngokevin:
**Description:** The low number of default segments causes lots of jagged/low-poly shapes, like octagons instead of circles. Which is fine to allow, but by default a circle primitive should look like an actual circle. The default should be at least 25-36 rather than 8.

- A-Frame Version: master
- Platform/Device: all

dmarcos:
Fixed on https://github.com/aframevr/aframe/pull/1241

dmarcos:
Kept open for other shapes

PR #1246, opened 2016-03-24
increased default segments for other round primitives

roninb:
Should close up issue #1216. Increased default segmentsTheta and segmentsTubular to make round primitives more round (or smooth).


ngokevin:
Thanks! Can you do segmentsPhi as well?

roninb:
segmentsPhi is actually how many triangles are rendered per SegmentsTheta (how many segments built for each trapezoid (?) that makes up the ring) so it doesn't make the circle more round, at least according to the docs.

ngokevin:
Aye, good catch.

ngokevin:
Future ref, you can append commit message with ```(fixes #ISSUENUMBER)``` so that issue auto closes on merge

cvan:
@ngokevin as I discovered, it'll auto-close the issue only if the author is added as a ["Collaborator"](https://github.com/aframevr/aframe/settings/collaboration) to the project. I wish GitHub did that automatically.

ngokevin:
As long as it's in the PR message when one of us merges it, it'll close.

roninb:
Thanks for the heads up, I'll make sure to remember that next time!

Issue #543, opened 2016-01-13
Divide components controls into body, head and cursor.


ngokevin:
<a href="https://github.com/jcarpenter"><img src="https://avatars.githubusercontent.com/u/1425497?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jcarpenter](https://github.com/jcarpenter)**
_Wednesday Oct 14, 2015 at 04:42 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/256_

----

Discussed control scheme with @caseyyee, per action item from weekly sync. Tried to come up with a proposal that captures immediate and future needs.

# Proposal for A-Frame controls

## Priorities

* v1.0 hard requirement: basic control scheme for "look and click" experiences, that is automatically responsive across desktop mono, desktop VR and mobile scenarios:
    * Issues including but not limited to:
    * #5 Add look-based cursor
    * #129 Add fuse-based cursor trigger for mobile devices w/o inputs
* Pave way for potential future features
    * Devs can modify control attributes (in v1 we probably don't expose them)
    * Support is added for new input schemes (eg motion controls, pointer lock)

## Proposal

Naming convention proposed in #241 is:

* `keyboard-controls`
* `mouse-controls`

We should instead group controls into "types", within which multiple input devices are possible. For example:

* `body-controls`
* `cursor-controls`

For additional separation of concerns, we should also move head orientation logic into it's own `head-controls` component.

#### `body-controls`

Contains logic for user position in space.

Support for:

* Positional camera
* WASD (future)
* Gamepad (future)
* Vive (future)

#### `cursor-controls`

Contains logic for cursor.

Support for:

* Mouse clicks
* Gamepad button presses (future)
* Keyboard key presses (future)
* Fuse timed-events

#### `head-controls`

Contains logic for head orientation.

Support for:

* Mouse drag
* HMD
* Pointer lock (future)
* Mobile device orientation

## Responsiveness

* In v1, we automatically toggle responsiveness (eg enabling fuse cursor on mobile devices) in the three control components. The defaults are not user editable.
* In v2, we can imagine giving the dev more control. 

For example:

**v1**: 

* For desktop browsers without webvr support (or HMD not connected):
    * `body-control` default is keyboard (option to disabled)
    * `cursor-control` default is mouse click
    * `head-control` default is mouse drag

* For mobile browsers:
    * `body-control` default is no support (devs can always implement own solutions with inline JS).
[...]

PR #1248, opened 2016-03-24
Controls refactor + extensibility.

### LINKED ISSUE #333: Add ZQSD for controller
A simple request from France, if it would be possible to have the Z key mapped to the W action, and same thing with Q for the A, it would allow a seamless controller for AZERTY keyboards

### LINKED ISSUE #975: Extensible controls (Trouble using multiple controls, refactor controls)

donmccurdy:
## Description:

Fixes #975, #848, #543. ~~Includes @bnolan's PR #1056 for pointerlock.~~

Proposed in #975, this attempts to address a couple issues:
1. Integrating third-party controls is a bit problematic.
2. Adding pointerlock as an option (default or no) would be awesome.
3. If a physics implementation is added, collision detection on the player would be a nice-to-have. The abstraction makes this much easier.
## Changes proposed:
- Add `controls` component to fetch user input from input device components and apply position/rotation to the object.
- Add device-specific components `wasd-controls`, ~~`gamepad-controls`~~, ~~`touch-controls` (touch-to-move)~~, `hmd-controls`, and `mouse-controls`. Probably not all of these should be enabled by default, but they could be.
- ~~Adds `velocity` component to update position over time. (Not necessary to the PR, but makes physics integration easier).~~
### Usage:

**Default:**

``` html
<a-entity camera controls></a-entity>
```

**Custom:**

``` html
<a-entity camera controls="rotation: custom-controls hmd-controls gamepad-controls mouse-controls"></a-entity>
```
### Controls Component

No exposed interface, but the basic implementation is:
1. Iterate over rotation controls, by order of precedence, finding first active control and setting rotation. 
2. Iterate over movement controls, by order of precedence, and get velocity relative to heading.
3. Apply rotation to velocity, to get absolute velocity.
4. Set velocity.

Developers can modify the `movement` and `rotation` properties to enable/disable input devices, or change the order of precedence.
### Movement Controls
- **isVelocityActive()** – _(required)_ Returns true if this control is actively in use. For example, a connected gamepad should return false unless the left joystick is tilted.
- **getVelocityDelta(dt)** – Returns an incremental THREE.Vector3 velocity change, with values [-1,1]. In general, this will affect the X and Z axes.
- **getVelocity()** – Returns a self-sufficient THREE.Vector3 velocity. May not be necessary, I can't think of a need.
### Rotation Controls
- **isRotationActive()** – _(required)_ Returns true if the control is actively in use. For example, mouse controls would only return true if the mouse is currently moving. HMD controls can probably return true any time an HMD is detected.
- **getRotationDelta(dt)** – Returns an incremental THREE.Vector2 rotation change, with X and Y rotation values. To be roughly calibrated, these should be near the range [-1,1].
- **getRotation()** – Returns a self-sufficient THREE.Vector3 (XYZ euler) world orientation, in degrees.
## Pending:
- [x] I'm assuming `controls` should inject the other components if they aren't present, rather than specifying all 5-6 in HTML.
- [x] Need to polyfill KeyboardEvent.code, which tangentially happens to fix #333 (ZQSD controls) in Chrome.
- :heavy_multiplication_x:  ~~Maybe `mouse-controls` should be split into mousedrag and pointerlock components?~~
- [x] Inject velocity component automatically
- [x] Deprecate look-controls and wasd-controls.
- [x] Rename movement -> position
- [x] Update all examples to use the new controls.
- [x] Update the default camera configuration in `systems/camera.js`.
- [x] Documentation
- [x] Unit tests
- [ ] Resolve conflicts with #1474 
- [ ] Consider adding `touch-controls` back in.
## Demos:
- Within A-Frame, updated the "Controls" demo.
- [Wall collision detection](https://sandbox.donmccurdy.com/vr/walls/) with gamepad/wasd/touch movement.
- [Motion sickness-inducing physics demo](https://sandbox.donmccurdy.com/vr/playground/) with ramps and moving platforms, because hey why not.

Side note, on physics integration, the wrapper around input controls is actually less critical than the `velocity` component. An example implementation (2nd and 3rd demos above) is [kinematic-body.js](https://github.com/donmccurdy/sandbox-aframe/blob/master/src/physics/kinematic-body.js), and the key need is having access to rotation/position/velocity data.


src/components/controls/mouse-controls.js
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Mouse-controls component.
+ *
+ * Updates rotation with mousedrag or pointerlock input.
+ */
+module.exports.Component = registerComponent('mouse-controls', {
+  schema: {
+    enabled: { default: true },
+    pointerlockEnabled: { default: true },
```
ngokevin: @dmarcos will veto this :)

ngokevin:
Gonna consolidate my ideas here:

- Maybe controls components need their own registration API and prototype? ```AFRAME.registerControlsComponent```
- Could probably rename ```universal-controls``` to just ```controls```. 
- Perhaps all registered controls include universal-controls as a dependency so it doesn't need to be explicitly attached.

ngokevin:
> ```<a-entity camera universal-controls="rotationControls: custom hmd gamepad mouse"></a-entity>```

How would you configure individual controls?

src/components/controls/universal-controls.js
```diff
@@ -0,0 +1,140 @@
...
+
+    flyingEnabled: { default: false },
+
+    movementEnabled: { default: true },
+    movementControls: { default: ['gamepad', 'wasd', 'touch'] },
+    movementEasing: { default: 15 }, // m/s2
+    movementAcceleration: { default: 65 }, // m/s2
+
+    rotationEnabled: { default: true },
+    rotationControls: { default: ['hmd', 'gamepad', 'mouse'] },
```
dmarcos: On aframe we currently limit rotation to the y-axis on touch and let it unconstrained for mouse. Have you thought on how we should handle individual configuration of controls?

donmccurdy: I'm guessing most configuration (invert Y axis, acceleration, ...) should go at the top level, but it would make sense to expose some on the individual controls:

```
<a-entity universal-controls
          mouse-controls="pointerlockEnabled: false; constrainAxis: y;">
</a-entity>
```

Also, in this PR I've actually removed touch rotation and replaced that with touch-to-move in `touch-controls`.. can certainly revert that to include the original touch rotation if there's a preference.

src/components/controls/mouse-controls.js
donmccurdy: Fair enough lol, i'll disable that.

donmccurdy:
- A registration API might make sense. I was hoping I could make `universal-controls` automatically add any components that were missing, so users don't need to specify all 5-6 components. It seems like adding components during init() might not work? Is that something the Systems API might help with?
- `controls` is good. :+1: 
- re: "universal-controls as a dependency" – I was hoping to do the opposite of that, and have `universal-controls` append the others if they're missing. But I don't think I can do that through the dependencies, since we'll only know which controls are specified as properties at init().
- `default-control-scheme`, I don't think I understand how that would work? But yes it would be nice for them to be added automatically.

src/core/controls.js
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Enables the given component names, and sets the given order as their precedence.
+   * @param {Array<string>} controls
+   */
+  enableRotationControls: function (controls) {
+    this.enabledRotationControls = controls;
+  },
+
+  /**
```
ngokevin: Hm, it does seem like these could be done within the `controls` component, right? I feel like a system could provide something, trying to think on what that could be.

ngokevin: Note, multiple entities could have controls (imagine hand-tracked controllers, or a third-person platforming game).

ngokevin: I think what a system could provide for now, is just a list of registered rotation + movement controls. So that controls component can check for validity.

donmccurdy: Yeah, makes sense. It would be nice to support multiple gamepads for example.

I've moved this logic back into the controls component, so in theory we can support several of them. Also used the system's registry to detect invalid controls and inject missing ones on load.

src/components/controls/controls.js
```diff
@@ -0,0 +1,222 @@
...
+
+  dependencies: ['velocity', 'rotation'],
+
+  schema: {
+    enabled: { default: true },
+
+    flyingEnabled: { default: false },
+
+    movement: { default: ['gamepad-controls', 'wasd-controls', 'touch-controls'] },
+    movementEnabled: { default: true },
```
ngokevin: When would we need to disable these?

src/components/index.js
```diff
@@ -4,7 +4,6 @@ require('./cursor');
 require('./geometry');
 require('./light');
 require('./look-at');
-require('./look-controls');
```
ngokevin: Might want to deprecate these for one version.

src/core/controls.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+var ControlsComponent = module.exports.ControlsComponent = function () {};
+
+ControlsComponent.prototype = {
+
+  /**
+   * Init handler. Called during scene initialization and is only run once.
+   * Controls can use this to set initial state.
+   */
+  init: function () { /* no-op */ },
```
ngokevin: Need an update handler?

ngokevin:
I think `position` makes sense over `movement` to keep the mapping. Just as `rotation` modifies `rotation`, `position` modifies `position`. It also maps well to terms like "positional tracking".

ngokevin:
I love that this controls system will automatically disable pointerlock/mousedrag/gamepad when an HMD is connected.

ngokevin:
I would keep `mouse-controls` together for now and re-evaluate later. It is nicer for code to split them up, though if someone wants to use pointer-lock over mousedrag, they would have to redefine the whole controls rotation order, rather than just flipping a flag.

ngokevin:
Nice, how much do you got left before you think it's ready for full line-by-line review phase?

src/components/controls/controls.js
donmccurdy: I think being able to disable movement seems important. Lots of cardboard demos out now that want to use their one-and-only button for something other than movement, or even for real VR you might want to constrain the camera inside a vehicle, or write your own teleportey-style movement.

ngokevin: I see. I was thinking you would just exclude those controls from `movement`, but this allows syntax that is nicer and more consistent.

src/core/controls.js
donmccurdy: :+1:

src/components/index.js
donmccurdy: Done – this does leave a conflict on the `wasd-controls`name. Use `keyboard-controls` for the new component instead, on the assumption we might include arrow key support and/or custom bindings?

donmccurdy:
Updated the "pending" list in original post, I think those two things are all I have in mind.

src/components/index.js
ngokevin: Perfect, that sounds great.

ngokevin:
There's also docs if you want to handle that as well, else I could pitch in.

donmccurdy:
Oh, yes docs would be key. I'll at least start them in this CL then.

ngokevin:
- Also need to update all examples to use the new controls.
- Need to update the default camera configuration in `systems/camera.js`.

ngokevin:
Found an edge case where updating `mouse-controls.pointerlockEnabled` to `true` in the DOM inspector instantly crashes Firefox

examples/test-controls/index.html
```diff
@@ -11,7 +11,7 @@
     <a-scene stats="true">
       <!-- We wrap the camera in an <a-entity> so we can position it without colliding with the HMD updates -->
       <a-entity>
-        <a-entity position="0 1.8 4" camera look-controls wasd-controls></a-entity>
+        <a-entity position="0 1.8 4" camera controls></a-entity>
```
dmarcos: controls is very generic you cannot tell what this will do without looking at the implementation

ngokevin: Documentation will help. If `controls` is used without anything else, it will use the default controls scheme. It'll also inject the other components.

People are still free to do ```controls="position: hmd-controls keyboard-controls gamepad-controls; rotation: hmd-controls touch-controls```, but it's convenient to abstract that if they want to use it out of the box.

src/components/controls/controls.js
```diff
@@ -0,0 +1,222 @@
...
+    flyingEnabled: { default: false },
+
+    position: { default: ['gamepad-controls', 'keyboard-controls', 'touch-controls'] },
+    positionEnabled: { default: true },
+    positionEasing: { default: 15 }, // m/s2
+    positionAcceleration: { default: 65 }, // m/s2
+
+    rotation: { default: ['hmd-controls', 'gamepad-controls', 'mouse-controls'] },
+    rotationEnabled: { default: true },
+    rotationSensitivity: { default: 0.05 } // radians/frame, ish
```
dmarcos: Are sensitivity and easing values that can be applied regardless of the input method? You probably want more rotation sensitivity with a mouse than a gamepad.

examples/test-controls/index.html
donmccurdy: is `controls` the wrong word, do you think, or would you want to have more components explicitly declared in the markup?

Other ideas: `navigation`, `user-controls`, `camera-controls` .. those second two may be too specific, since this could presumably be moving something other than the camera.

src/components/controls/controls.js
donmccurdy: Maybe. In this case there's an overall sensitivity, and mouse-controls has a property to override it. Could move sensitivity directly into `gamepad-controls` and `mouse-controls` since it doesn't apply to `hmd-controls` anyway.

Easing/acceleration make more sense at the top level to me, since movement speed should probably be consistent.

examples/test-controls/index.html
ngokevin: I think it is good to leave to `controls` to be agnostic enough to be used for anything that takes input, and then modifies position/rotation of any entity. This example just displays using the out of the box default control scheme, so it won't show much info (until it injects all of its components).

donmccurdy:
Quick update, I've patched `<a-camera/>` and all of the examples. Remaining TODOs are updating Documentation and writing unit tests. Is it worth it to start on those now, or want to iterate on the API more or do a more line-by-line review first?

examples/test-controls/index.html
dmarcos: controls does not mean much in a general way. It's application dependent. FPS vs Panorama Viewer vs Platform Game vs Flight simulator. It's not the same controlling a camera, than a character in 3rd person view. I suggest them giving a prefix that convey what kind of controls we're talking about. Relying on documentation I think it's just lazy thinking on naming things. People don't read docs.

dmarcos: I suggest ```pov-controls```, ```first-person-controls```, ```fps-controls```

src/components/controls/controls.js
dmarcos: Can a user of controls specify sensitivity separately for mouse / gamepad?

examples/test-controls/index.html
dmarcos: Disregard the above I'm fine with the name controls

dmarcos:
@donmccurdy go ahead with the docs and the unit tests

src/extras/primitives/primitives/a-camera.js
```diff
@@ -3,17 +3,16 @@ var registerPrimitive = require('../registerPrimitive');
...
   },
 
   mappings: {
     active: 'camera.active',
     far: 'camera.far',
     fov: 'camera.fov',
-    'look-controls-enabled': 'look-controls.enabled',
     near: 'camera.near',
-    'wasd-controls-enabled': 'wasd-controls.enabled'
+    'position-controls-enabled': 'controls.positionEnabled',
```
ngokevin: I know it's a WIP, but caught that this isn't consistent with the docs and the line below.

I might keep it as `position-enabled` + `rotation-enabled` to be consistent with the prop name.

src/components/velocity.js
```diff
@@ -0,0 +1,25 @@
+var registerComponent = require('../core/component').registerComponent;
+
+/**
+ * Velocity component.
+ *
+ * Updates object's position at each tick.
+ */
+module.exports = registerComponent('velocity', {
+  dependencies: ['position'],
```
ngokevin: I don't think we need to specify this since it is a default component on all entities

ngokevin: rotation as well/g

donmccurdy: ah! ok thanks. 

qq on that – the first time I call `el.getAttribute('position')` I'm getting `undefined`, even though it's a default dependency. But `el.getComputedAttribute('position')` returns `{x:0,y:0,z:0}`.  After the position has been set once, either of those works fine. Is that a bug? Or is `getComputedAttribute` just the thing I should use?

ngokevin: Yeah. getAttribute returns defined values, getComputedAttribute returns all data including mixins, defaults, etc. I wonder if that's intuitive

donmccurdy: Oh, now that you put it that way it does make sense. I guess `getComputedAttribute` is generally what I want (to avoid initial `undefined`) then.

src/extras/primitives/primitives/a-camera.js
donmccurdy: eek, thanks yes

src/components/controls/controls.js
donmccurdy: Done: there was already `mouse-controls="sensitivity: ..."` and I added `gamepad-controls="sensitivity: ..." to match. Removed the option from the top-level `controls` component (probably less necessary there?).

src/components/controls/controls.js
```diff
@@ -0,0 +1,218 @@
...
+ */
+module.exports.Component = registerControls('controls', {
+  dependencies: ['velocity'],
+
+  schema: {
+    enabled: { default: true },
+
+    flyingEnabled: { default: false },
+
+    position: { default: ['gamepad-controls', 'keyboard-controls', 'touch-controls'] },
```
ngokevin: What about HMD controls for positional tracking?

donmccurdy: soon6

This CV1 shipping delay is rough though. :(

Anyway I would expect to do this with a `getPosition()` method on `hmd-controls` maybe, as opposed to the current `getVelocityDelta()`?

ngokevin: Do you have a DK2? We might have extras if you need it.

donmccurdy: I just have a couple cardboards here. I'm theoretically in the March 28th shipping group, but got delayed until April 12th-ish. Thanks, might be helpful if one of you wants to take a look at attaching `hmd-controls::getVelocity()`, otherwise I hopefully don't have too much longer to wait.

ngokevin: Think you could copy over the look-controls code to hmd-controls, we could test it? If not, yeah I can help.

donmccurdy: I'll give it a shot! Maybe hold into something or sit down when you test it ;)

donmccurdy:
First pass at docs (for the component, not the controls system) up now. Didn't seem like enough content to merit separate pages or a new section, but open to ideas.

ngokevin:
Nice on the docs. Yeah, good to keep it simple and in one page for now.

We're probably going to split up the geometry component documentation later, so we might do it with the controls to be consistent and give it its own little section. But later.

src/components/controls/controls.js
ngokevin: No worries, I got an iron stomach.

docs/components/controls.md
```diff
@@ -0,0 +1,122 @@
...
+
+The following input components are provided by A-Frame, in addition to any custom controls created by the user:
+
+* [gamepad-controls](#gamepad-controls) *(position + rotation)*
+* [hmd-controls](#head-mounted-display-controls) *(position + rotation)*
+* [keyboard-controls](#keyboard-controls) *(position)*
+* [mouse-controls](#mouse-controls) *(rotation)*
+* [touch-controls ](#touch-controls) *(position)*
+
+[//]: # (HMD-controls link is spelled out, to prevent Hexo from replacing "md" with ".html".)
```
ngokevin: lol

ngokevin: is that actually happening? https://github.com/aframevr/aframe-site/blob/master/scripts/helpers.js#L46

donmccurdy: hah yeah, is `data.path` the current page, or the target? Maybe just need to escape the "." in the line 49 regexp.

cvan: it's the document being processed. so in this case, it'd be `docs/components/controls.md`.

cvan: thanks, @donmccurdy! (fixed in #210 and #211)

donmccurdy:
Some unit tests added now, but still need to get the HMD position tracking in there.

donmccurdy:
I think I'm caught up. The core functionality all has unit tests, and `hmd-controls` has positional tracking (which I haven't manually tested, but THREE.VRControls makes it seem pretty simple).

Maybe still some API questions to resolve:

1. The parent `controls` will reorient `getVelocityDelta` to the rotation heading, but not `getVelocity` (used only by `hmd-controls`). That's possibly unclear from the names, but makes sense in context.
2. Use of "position" (e.g. `controls="positionEnabled: true"`) and "velocity" (e.g. `component.isVelocityActive()`) feels inconsistent. Not sure what names to use that would be both meaningful and consistent, though.

docs/components/controls.md
```diff
@@ -0,0 +1,120 @@
...
+|----------|---------------------------------|---------------|
+| enabled  | Whether touch input is enabled. | true          |
+
+## Custom Controls
+
+To create your own component for position or rotation controls, use the `AFRAME.registerControls()` API.
+
+Example:
+
+```html
```
ngokevin: js

ngokevin:
Thanks, will test positional tracking tomorrow.

1. What's the context?
2. That seems okay to me. We can always change it later if we come up with API verbiage.

donmccurdy:
**re: getVelocity[Delta] and heading** So far, four components update position: hmd, gamepad, keyboard, and touch. The last three (using `getVelocityDelta`) all are relative to heading – pressing <kbd>W</kbd> or <kbd>DPAD_UP</kbd> should move in the direction you're looking, not just `z=-1`. But HMD position changes (using `getVelocity`) are absolute. 

But now that I think about it maybe that's fine. The word "delta" implies that it's a change from some original velocity, whereas `getVelocity` does not.

ngokevin:
Should we insert `enabled: {default: true}` into the schema automatically and have it work out of the box? Then you wrap every `is*Active` and automatically return depending on `data.enabled`.

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+  init: function () {
+    // Movement
+    this.velocity = new THREE.Vector3();
+
+    // Rotation
+    this.pitch = new THREE.Object3D();
+    this.yaw = new THREE.Object3D();
+    this.yaw.position.y = 10;
+    this.yaw.add(this.pitch);
+    this.heading = new THREE.Euler(0, 0, 0, 'YXZ');
```
ngokevin: Can you teach me why we wouldn't use a quaternion?

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+      this.updateVelocity(dt);
+    }
+  },
+
+  /**
+   * Updates rotation based on input from the active rotation component, if any.
+   * @param  {number} dt
+   */
+  updateRotation: function (dt) {
+    var control, dRotation;
```
ngokevin: let's do one-per-line

```
var control = ...;
var dRotation;
```

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+    if (this.el.sceneEl.hasLoaded) {
+      this.validateControls();
+    } else {
+      this.el.sceneEl.addEventListener('loaded', this.validateControls.bind(this));
+    }
+  },
+
+  update: function () {
+    // Re-validate if scene has loaded. If not, listener was already bound in init().
+    if (this.el.sceneEl.hasLoaded) {
```
ngokevin: we use `return`s a lot to decrease indents

```
if (...) { return; }
this.validate();
```

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+   * @param  {number} dt
+   */
+  updateRotation: function (dt) {
+    var control, dRotation;
+
+    control = this.getActiveRotationControls();
+    if (control && control.getRotationDelta) {
+      dRotation = control.getRotationDelta(dt);
+      this.yaw.rotation.y -= dRotation.x;
+      this.pitch.rotation.x -= dRotation.y;
```
ngokevin: we like to use intermediate vars if they're reused

```
var el = this.el
var pitch = this.pitch;
```

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Updates velocity based on input from the active position component, if any.
+   * @param {number} dt
+   */
+  updateVelocity: function (dt) {
+    var control;
+    var data = this.data;
+
+    if (this.el.hasAttribute('velocity')) {
+      this.velocity.copy(this.el.getComputedAttribute('velocity'));
```
ngokevin: I think you can use `getAttribute` if you already checked the attribute is there.

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+ */
+module.exports.Component = registerControls('controls', {
+  dependencies: ['velocity'],
+
+  schema: {
+    enabled: { default: true },
+
+    flyingEnabled: { default: false },
+
+    position: {
```
ngokevin: Maybe `positionInputs` and `rotationInputs`?

ngokevin: Hmm...well maybe not since they're components.

src/components/controls/deprecated/look-controls.js
```diff
@@ -1,9 +1,12 @@
-var registerComponent = require('../core/component').registerComponent;
-var THREE = require('../lib/three');
+var registerComponent = require('../../../core/component').registerComponent;
+var THREE = require('../../../lib/three');
+var debug = require('../../../utils/debug');
```
ngokevin: Can just import `utils` and do `utils.debug`

src/components/controls/gamepad-controls.js
```diff
@@ -0,0 +1,110 @@
+var registerControls = require('../../core/controls').registerControls;
+var THREE = require('../../lib/three');
+var Gamepad = require('../../constants').gamepad;
```
ngokevin: Maybe cap it ```GAMEPAD```

src/components/controls/gamepad-controls.js
```diff
@@ -0,0 +1,110 @@
...
+
+var JOYSTICK_EPS = 0.2;
+
+/**
+ * Gamepad-controls component.
+ *
+ * Supports movement with the left joystick or dpad, and rotation with the right joystick.
+ *
+ * For more information about the Gamepad API, see:
+ * https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API
```
ngokevin: Can remove the `en-US`

src/components/controls/gamepad-controls.js
```diff
@@ -0,0 +1,110 @@
...
+  },
+
+  getVelocityDelta: function () {
+    var dpad = this.getDpad();
+    var joystick0 = this.getJoystick(0);
+    var inputX = dpad.x || joystick0.x;
+    var inputY = dpad.y || joystick0.y;
+    var dVelocity = new THREE.Vector3();
+    if (Math.abs(inputX) > JOYSTICK_EPS) {
+      dVelocity.x += inputX;
```
ngokevin: can do one liners like below

src/components/controls/hmd-controls.js
```diff
@@ -0,0 +1,77 @@
...
+  schema: {
+    enabled: { default: true }
+  },
+
+  init: function () {
+    this.dolly = new THREE.Object3D();
+    this.hmdEuler = new THREE.Euler();
+    this.hmdEuler.order = 'YXZ';
+    this.hmdQuaternion = new THREE.Quaternion();
+    this.controls = new THREE.VRControls(this.dolly);
```
ngokevin: `vrControls`?

src/core/controls.js
```diff
@@ -0,0 +1,136 @@
...
+// Keep track of registered controls.
+var controls = module.exports.controls = {
+  positionControls: {},
+  rotationControls: {}
+};
+
+/**
+ * Initializes the A-Frame controls system.
+ */
+systems.registerSystem('controls', {
```
ngokevin: Maybe define this in `systems/`?

ngokevin:
**Testing:**

## keyboard + mousedrag

keyboard-mousedrag

## keyboard + hmd

docs/components/controls.md
```diff
@@ -0,0 +1,120 @@
...
+To customize which input methods are enabled, specify the `position` or `rotation` properties:
+
+```html
+<a-entity camera controls="position: custom-controls gamepad-controls; rotation: hmd-controls;"></a-entity>
+```
+
+To configure one of the input methods, include its component and set the properties you want to change:
+
+```html
+<a-entity camera controls mouse-controls="pointerlockEnabled: true"></a-entity>
```
ngokevin: Pointer Lock as two words might be more accurate `pointerLockEnabled`

src/components/controls/mouse-controls.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Mouse-controls component.
+ *
+ * Updates rotation with mousedrag or pointerlock input.
+ */
+module.exports.Component = registerControls('mouse-controls', {
+  schema: {
+    enabled: { default: true },
+    pointerlockEnabled: { default: false },
```
ngokevin: I'm starting to think Pointer Lock controls should be a separate third-party controls component. We're getting more serious about actual VR, and rotating the camera with Pointer Lock is not something you'd do in VR.

The downside is responsiveness with flat screens. But I think we'll see responsiveness in VR means that it works on multiple headsets/platforms, not that it works on everything. Though if someone wants to use A-Frame for flat experiences, they are free to pull/implement a custom third-party Pointer Lock controls.

So we should maybe just get rid of Pointer Lock from the core repo and publish it as an easy-to-consume third-party repo. It increases testing overhead and code maintenance by a fair amount. @dmarcos Agree?

ngokevin: Though we'd keep mouse drag as a bare minimum so people can at least preview experiences. Mouse drag is a lot easier to test, more predictable, and Pointer Lock has too many steps (browser modals) if you just want to preview around.

examples/animation-aframe-logo/index.html
```diff
@@ -17,8 +17,8 @@
         <a-animation attribute="position" from="0 5 0" to="0 0 0" dur="4000"
                      easing="ease-out"></a-animation>
         <a-camera id="camera" near="1000" far="4000" position="0 880 1290" fov="2.2"
-                  rotation="-34 0 0" wasd-controls-enabled="false"
-                  look-controls-enabled="false"></a-camera>
+                  rotation="-34 0 0" position-enabled="false"
```
dmarcos: position-enabled and rotation-enabled they're misleading names. Position and rotation are not really disabled. They can still be changed without controls.

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Updates internal velocity. Takes a velocity delta (relative to current heading), rotates it
+   * to the heading, and calculates velocity in world space.
+   * @param {number} dt
+   * @param {THREE.Vector3} dVelocity
+   */
+  applyVelocityDelta: function (dt, dVelocity) {
+    var data = this.data;
+    var rotation = this.el.getAttribute('rotation');
```
dmarcos: Do you want `getComputedAttribute` in case the rotation is defined via mixins?

src/components/controls/mouse-controls.js
dmarcos: I'm always happy to get rid of things from the core :). The things that makes me happiest in the world is to remove code.

ngokevin: Hehe, well it isn't removing code. We're just talking about not adding code.

@donmccurdy Let's publish PL outside core. Once we have lazy-loading components, it will be a better story for consumption.

examples/animation-aframe-logo/index.html
ngokevin: Agreed, `position-controls-enabled` or something.

src/components/controls/controls.js
ngokevin: It's starting to feel more intuitive to me that `getAttribute` should return all data since that is the most common case, and then we create a special method for `getDefinedAttribute` or something. Not sure the `computedStyle` analog is worth it.

src/components/controls/controls.js
donmccurdy: if we're changing `positionEnabled` to `positionControlsEnabled`, I might be in favor of `positionControls` here.

ngokevin: Maybe. With the attribute on the camera, there's no context of what it refers to. Though within the controls component, it refers to controls. Don't have a strong opinion.

src/components/controls/keyboard-controls.js
```diff
@@ -0,0 +1,94 @@
+var registerControls = require('../../core/controls').registerControls;
+var THREE = require('../../lib/three');
+
+// Tiny KeyboardEvent.code polyfill.
+var KEYCODE_TO_CODE = {
```
ngokevin: maybe useful in constants?

src/components/controls/gamepad-controls.js
```diff
@@ -0,0 +1,110 @@
+var registerControls = require('../../core/controls').registerControls;
```
ngokevin: I'm unsure about gamepad controls in the core. Maybe too prescriptive? Trying to do a use case where I want the right joystick to snap the rotation in like 30-degree increments, so I feel like immediately gamepad controls should be left to the developer because each app is so specific. Like developer might not want DPad to do velocity as well.

But we'd definitely publish it separately if we excluded it here.

@donmccurdy @dmarcos

ngokevin: Though for my case, it's nice that I am able to override either the position or rotation pieces of the gamepad controls separately.

```
<a-entity controls="position: your-gamepad-controls; rotation: my-gamepad-controls">
```

ngokevin: I think for locomotion, we should focus around room-scale long-term.

src/components/controls/gamepad-controls.js
```diff
@@ -0,0 +1,110 @@
...
+    if (Math.abs(lookVector.x) <= JOYSTICK_EPS) { lookVector.x = 0; }
+    if (Math.abs(lookVector.y) <= JOYSTICK_EPS) { lookVector.y = 0; }
+    return lookVector.multiplyScalar(data.sensitivity);
+  },
+
+  /**
+   * Returns the Gamepad instance attached to the component.
+   * @return {Gamepad}
+   */
+  getGamepad: function () {
```
ngokevin: rather than calling this, you could set `this.gamepad` on `update()`

src/components/controls/keyboard-controls.js
```diff
@@ -0,0 +1,94 @@
+var registerControls = require('../../core/controls').registerControls;
```
ngokevin: In the future, I might try to make the tough call to extract out keyboard controls. It's not really great to have for first-person VR and we should encourage discovering better locomotion mechanisms. WASD has been a bit of a crutch.

For third-person VR is fine, but then WASD + third-person is too specific of a use-case to include in core.

@dmarcos Should we just move it out now?

ngokevin: We'd still have it around in a separate repo, and then we could try UX patterns like ghosting:

http://uploadvr.com/convrge-developers-introduce-improved-ghosting-a-method-for-vr-locomotion/

src/components/controls/controls.js
```diff
@@ -0,0 +1,220 @@
...
+    control = this.getActiveRotationControls();
+    if (control && control.getRotationDelta) {
+      dRotation = control.getRotationDelta(dt);
+      this.yaw.rotation.y -= dRotation.x;
+      this.pitch.rotation.x -= dRotation.y;
+      this.pitch.rotation.x = Math.max(-PI_2, Math.min(PI_2, this.pitch.rotation.x));
+      this.el.setAttribute('rotation', {
+        x: THREE.Math.radToDeg(this.pitch.rotation.x),
+        y: THREE.Math.radToDeg(this.yaw.rotation.y),
+        z: 0
```
ngokevin: Should we allow rotation about the Z-axis? Not good for the first-person camera, but say I want to use the cursor to pick something up and want to spin it around?

donmccurdy: We could add `this.roll` but I think in that case that you might want `getRotation` instead of `getRotationDelta`.

ngokevin: For like Vive controllers, I see you'd use `getRotation`.

But just want to make it flexible in case you want to, say, click an object to select it, and then scroll the mousewheel to roll it.

src/components/controls/keyboard-controls.js
donmccurdy: This is vaguely my feeling on Pointer Lock as well, but I feel a little stronger about it in the keyboard case – I'd vote to aim for whatever we think the VR-equivalent of progressive enhancement is going to be. Lots of users won't have HMDs or even Cardboard around all the time, they're just clicking a fun-looking link they found on Twitter. If we remove keyboard support and require drag-mouse-to-look-around, a lot of otherwise interesting VR scenes are going to be pretty unusable a desktop PC. Lot of cases where WebVR experiences should support non-VR users if they can – something casual like Convrge, where I might want to leave the window open while I'm chatting with friends in the scene, but checking my email at the same time. Obviously a separate `keyboard-controls` repo would let devs work around that easily, but defaults are important.

On the flip-side of "defaults are important", I agree that we should strongly encourage better locomotion options. I just feel bad nuking the desktop experience to do that. Any way to keep a decent fallback experience while still discouraging WASD as the primary locomotion? If there's a middle ground then great, if not then maybe you're right and we need Tough Choices.

donmccurdy: Also, slight risk that by removing `keyboard-controls` here we'll encourage more people to keep using the now-deprecated `wasd-controls` + `look-controls` combination. Maybe that's avoidable with strong warnings.

We could also get going on a quick follow-up PR for `hotspot` and `hotspot-controls` components.

ngokevin: Yeah, I get what you mean. I really want to commit to this Tough Choice though.

The future I imagine, I don't think VR can bring along flat experiences; it's a different medium. If a developer wants their experience to be accessible from a non-VR experience, then they can put in and extra effort and design it to also work on flat. But for most VR experiences that are designed only for VR (room-scale, Touch/Vive controllers), which is what A-Frame is focused on, non-VR users will not be able to experience them in any meaningful way. 

Flat experiences, almost like the current form of Cardboard, are sort of distractions for true WebVR. Defaults are important is almost a case that we *should* nuke default flat experiences. Because with WASD controls available, most developers are content with that form of locomotion and don't optimize the UX for VR. I think the middle ground is having it available in a separate repo if a developer wants it.

If we go this route, we'd probably just nuke default WASD controls for camera as well, and then really prototype better locomotion. I think this would possibly hurt A-Frame's popularity since a lot of people are using it for just flat experiences, but it's better for the long-term. Flat web games never really took off anyways, so might be a good idea to drop that baggage.

ngokevin: I wonder if hotspot needs to be a control rather than a standard component? All it needs to do is tween the camera (wrapper) to the position of the hotspot. If the camera "collides" with something on the way (which it shouldn't with well-placed hotspots), then you just let it clip.

src/components/controls/mouse-controls.js
```diff
@@ -7,30 +7,20 @@ var warn = debug('components:mouse-controls:warn');
 /**
  * Mouse-controls component.
  *
- * Updates rotation with mousedrag or pointerlock input.
+ * Updates rotation when dragging the mouse.
  */
 module.exports.Component = registerControls('mouse-controls', {
   schema: {
     enabled: { default: true },
-    pointerlockEnabled: { default: false },
```
ngokevin: Sorry for the extra work. It is 100% salvageable as 3rd-party at least

src/components/controls/keyboard-controls.js
ngokevin: It's interesting that I think a lot of people say that the Web was meant to work everywhere. I think for WebVR, responsiveness means that it works on multiple HMDs and platforms, not that it works across different mediums.

src/components/controls/mouse-controls.js
donmccurdy: np, better now than after merge.

src/components/controls/controls.js
```diff
@@ -0,0 +1,225 @@
...
+                        'registered with AFRAME.registerControls().');
+      } else if (!el.components[name]) {
+        missingControls.push(name);
+      }
+    }
+
+    for (i = 0; i < data.rotation.length; i++) {
+      name = data.rotation[i];
+      if (!system.rotationControls[name]) {
+        throw new Error('Component `' + name + '` must be ' +
```
ngokevin: Need to get the specific control

```Component `comfort-turn-keyboard-controls mouse-controls` must be registered with AFRAME.registerControls().```

File: src/components/controls/controls.js [hunk 37]
```diff
@@ -0,0 +1,225 @@
...
+    var name;
+    var el = this.el;
+    var data = this.data;
+    var system = this.system;
+    var missingControls = [];
+
+    for (i = 0; i < data.position.length; i++) {
+      name = data.position[i];
+      if (!system.positionControls[name]) {
+        throw new Error('Component `' + name + '` must be ' +
```
ngokevin: + here

src/components/controls/controls.js
donmccurdy: ?

ngokevin: You are logging the whole list of controls rather than the offending unregistered controls

ngokevin:
I get an error when I try my custom component alongside a standard one:

```html
<a-entity camera
               controls="position: hmd-controls; rotation: comfort-turn-keyboard-controls mouse-controls"></a-entity>
```

src/components/controls/controls.js
donmccurdy: Ah ok I see it now. I get the same error with just this: `controls="rotation: mouse-controls hmd-controls"`

It looks like `data.rotation` is `["mouse-controls hmd-controls"]`. Is that a regression somewhere or was I making too many assumptions about array support?

ngokevin:
I'm trying it out here: https://github.com/ngokevin/aframe-comfort-turn-controls/blob/master/index.js#L123

Couldn't get the `getRotationDelta` to get called. Will try more tmr.

src/components/controls/keyboard-controls.js
ngokevin: @dmarcos Thoughts?

donmccurdy: ^Assuming Diego agrees this sounds fine to me.

src/components/controls/controls.js
ngokevin: Yeah, there's no array support like that. I thought it was working magically somehow.

I think you can specify `type: "array"`, and then you deliminate with commas.

ngokevin: I think that's why my controls aren't working https://github.com/ngokevin/aframe-comfort-turn-controls

Because `data.rotation` will be a string, and it will loop over each letter of 'comfort-turn-keyboard-controls`?

donmccurdy: So that's just a bug in my docs, it sounds like – should be `<a-entity controls="rotation: custom-controls1, custom-controls2"/>` with a comma, not just a space. Comma-delimited is working fine for me; fixed the docs.

ngokevin: ok sweet

src/components/controls/gamepad-controls.js
donmccurdy: Either way on this –

Pro of removing:

- Gamepad buttons are important, but probably too open-ended to live in core, and not related to position/rotation. My current solution (not in this PR) is firing two events: `gamepadbuttondown` and `gamepadbuttondown:BUTTON_NAME`. So that's an opinionated choice.

Cons of removing:

- Every Oculus owner has a gamepad.

src/components/controls/deprecated/look-controls.js
donmccurdy: hm? this looks the same as is done everywhere else in the codebase

ngokevin: Slowly moving towards that pattern, since `AFRAME.utils` is public API, I wanna mimic that.

src/components/controls/gamepad-controls.js
ngokevin: The con is true, but like you said the gamepad wouldn't necessarily be used for first-person movement/rotation. Even in third-person games, like a platform, the gamepad behaves differently (left stick controls both rotation + velocity). So it might be another tough choice to leave it non-prescriptive. But I think the gamepad could be used for lots and lots of other things.

Well, then we only ship HMD + mousedrag, which are about the minimum + non-prescriptive we can get I think.

ngokevin:
@dmarcos A couple of questions about strong changes to the direction the project we'd like your input on.

src/components/controls/gamepad-controls.js
donmccurdy: ship `touch-controls` or no? I don't think we discussed that one

ngokevin: Hmm, I imagine proper Cardboard experiences would use hotspot navigation and/or reserve the touch controls for clicking on stuff.

Though you could say something similar for mousedrag, but that's sort of the bare minimum so people can preview scenes without an HMD. Scenes that actually intend to use the mouse alongside an HMD though, those scenes won't work well on flat experiences. Tough compromises.

ngokevin: In other words, I think we could also publish that separately as well because UX-wise, I'm not sure if I want to encourage that form of locomotion

donmccurdy: It would admittedly be weird to have locomotion by default on mobile but not on desktop, and yeah most of the time you probably want the cardboard button for interacting with things, not moving. So i guess we're not including any default locomotion then – maybe hotspot navigation later on.

ngokevin:
@dmarcos and I talked.

Going to keep `keyboard-controls` around as a placeholder to ease into new locomotion. But the compromise for VR is that we remove it from the default camera, and have it go in "steps" rather than continuous motion. We can add the step/teleportation in later, but I imagine it wouldn't use the velocity API.

src/components/controls/controls.js
```diff
@@ -0,0 +1,225 @@
...
+      yaw.rotation.y -= dRotation.x;
+      pitch.rotation.x -= dRotation.y;
+      pitch.rotation.x = Math.max(-PI_2, Math.min(PI_2, pitch.rotation.x));
+      el.setAttribute('rotation', {
+        x: THREE.Math.radToDeg(pitch.rotation.x),
+        y: THREE.Math.radToDeg(yaw.rotation.y),
+        z: 0
+      });
+    } else if (control) {
+      el.setAttribute('rotation', control.getRotation());
```
ngokevin: You should store the new pitch/yaw/roll so it doesn't get reset by `getRotationDelta`. I tried doing comfort snaps, but they were overwritten by the HMD.

src/components/controls/hmd-controls.js
```diff
@@ -0,0 +1,81 @@
...
+    return this.data.enabled && !this.position.equals(this.dolly.position);
+  },
+
+  getVelocity: function () {
+    var dolly = this.dolly;
+    var position = this.position;
+    var velocity = this.velocity;
+    velocity.copy(dolly.position).sub(position);
+    position.copy(dolly.position);
+    return velocity;
```
ngokevin: Was testing this out. Wasn't moving as much as it should have (barely noticable). Should we be using velocity for HMD?

donmccurdy: Oh right. Try again now – was calculating velocity as `position - lastPosition`, which would be distance over 8-16ms, but then inside `velocity.js` treats m/s. Needed to scale by 1000 / dt inside hmd-controls. (But maybe a separate `getPosition` function would have uses too..)

donmccurdy:
Ok, not sure what you mean about steps (like, simulated bipedal lurching?) - i assume that's to eliminate any misconceptions that WASD is acceptable for VR lol? Anyway, pushed `keyboard-controls` back up and removed it from the default set.

src/components/controls/controls.js
donmccurdy: Pitch/yaw do get stored, I think maybe what you're seeing is that we don't work backwards to compute pitch and yaw from HMD `getRotation` updates?

Off hand I would think we need:

```javascript
yaw.rotation.y = THREE.Math.degToRad(rotation.y);
pitch.rotation.x = THREE.Math.degToRad(rotation.x);
pitch.rotation.x = Math.max(-PI_2, Math.min(PI_2, pitch.rotation.x));
```

dmarcos:
By steps kevin meant that instead of continuous movement like wasd-controls do now we would move on discrete steps. You teleport 1 unit forward each time you press 'W'. The step should be configurable. There's no acceration or easing that can cause motion sickness

ngokevin:
This might complicate things, but perhaps we need `getPosition`. For HMD positional tracking, I don't think we want velocity controlling the position because the position should always absolute. When I move the camera up above the tracking volume, there's some leftover velocity that keeps moving the camera up and up even though the HMD is no longer moving.

ngokevin:
OK, I found my comfort turn snap controls aren't working because both my controls and the HMD controls use `getRotation` so the HMD controls are overwriting changes from my controls.

- Comfort Turn changes rotation by 15 degrees from the current rotation using `getRotation` on keyboard press.
- HMD is active again and sets its absolute rotation using `getRotation`, overwriting comfort turn.

donmccurdy:
re: `getPosition` -  yeah i'm ok with that, just makes it a bit trickier to detect imminent collisions in physics land later on. we could always compute a velocity based on that `getPosition()` result, so that the component author doesn't need to worry about scaling by `1000/dt` (to counter the scaling in `velocity`) themselves.

re: comfort snap controls – cool. Ideally that would use `getRotationDelta()` I think - maybe we shouldn't scale the delta by `dt / 1000` so it's absolute? Or maybe you just account for that in your component. This all sounds like it can fit in the controls framework OK though.

ngokevin:
OK. Let's try `getPosition` that calculates velocity under the hood. And make rotation absolute (though if you didn't want it absolute, how would you do it?).

The main use cases for the controls component I think is:

- HMD - has to be absolute and exact. any controls changing the rotation would probably not want easing
- Vive controllers - has to be absolute and exact. but will still want to calculate velocity and angular velocity data and case you want to hit stuff
- But will still have to support flat experience controls? (drag-look, wasd for now, which we may want to snap by default as well)

Do we replace velocity with position, or do we add getPosition? That sort of solves the issue from earlier with rotation mapping to rotation but position mapping to velocity.

donmccurdy:
Guess I was wrong about rotation – it's already absolute. `getRotation()` does update yaw/pitch now, so if your comfort controls are using `getRotationDelta` I think you should be good.

I'd agree that HMD + Vive controllers are the main use cases, but it would be nice to be able to support third-party flat experience controls. Added `getPosition`, and `hmd-controls` are now using that. It converts position to velocity under the hood. Focus being on HMD controls, this is hitting the limit of what I can actually test with just a Cardboard. Hoping to have my HMD in a couple weeks.

ngokevin:
OK, comfort turn controls work for me now.

Let's publish touch controls separately as well for reasons discussed above (recommendations on movement in VR, reserve for interaction, less code to maintain/test for us).

I'll have to wait until next week to get access to a headset, I'm abroad atm :(

donmccurdy:
Ok that sounds good, done.

ngokevin:
VR controls stopped working (no positional/rotation tracking or connection).

ngokevin:
error: `Component `touch-controls` must be registered with AFRAME.registerControls().`

ngokevin:
Disregard my deleted comments, I think I needed to restart my server.

ngokevin:
Positional tracking is off. When I move my head forward and back, I see the camera move left and right.

src/systems/camera.js
```diff
@@ -37,8 +37,7 @@ module.exports.System = registerSystem('camera', {
       cameraWrapperEl.setAttribute(DEFAULT_CAMERA_ATTR, '');
       defaultCameraEl = document.createElement('a-entity');
       defaultCameraEl.setAttribute('camera', {'active': true});
-      defaultCameraEl.setAttribute('wasd-controls', '');
-      defaultCameraEl.setAttribute('look-controls', '');
+      defaultCameraEl.setAttribute('controls');
```
ngokevin: Needs a second argument.

ngokevin:
OK, I think most stuff works. We're going to land the new WebVR API soon so the hmd-controls might need changing in a bit. @dmarcos will update the current look-controls so we'd port those changes over.

donmccurdy:
Great! I haven't had the chance to read through the WebVR 1.0 changes but will do this soon. I got my Rift yesterday ( 🎉 ) so I can test the positional tracking as well now.

ngokevin:
Got your rift? WOOT!

ngokevin:
@dmarcos just landed the WebVR 1.0 API changes. Just need to port his changes from the old look-controls.

ngokevin:
Changes to look-controls at https://github.com/aframevr/aframe/pull/1423/files#diff-6285039eee70ed830f70f7d2f7b7dfa5R33 

Gonna test this out and see what needs changing.

donmccurdy:
Rebased to to the WebVR 1.0 changes – starting on updating `hmd-controls` now.

Also, somehow we ended up with `rotation-enabled` and `position-controls-enabled` .. those should both be `*.-controls-enabled` i think right?

ngokevin:
I think in the context of the controls component, makes sense to be just `positionEnabled` and `rotationEnabled`. Only for `<a-camera>` does it need to be explicit.

ngokevin:
OK, I tested this with Vive + Oculus. Mostly works, but compared to `master`, it is jankier. Any ideas why? Maybe something around the velocity stuff.

I see that we have a `getPositionDelta` now. For controllers + HMD, should we just make `getPosition`? Those don't really care about the velocity, we just want to set absolute pose, right?

Right now, what's the flow for updating the position? Is it:

- Get new position
- Get delta using old position
- Use delta to calculate velocity
- Have velocity set position

donmccurdy:
Will profile on performance a bit, my first thought would be maybe `velocity` is updating before the controls component, or after `position`, and so the HMD is delayed by 1 frame? Is there logic somewhere that makes sure position/rotation have updates from all the components before they're applied? Maybe we need that for velocity.

Reasoning on `getPositionDelta` is that the player might not start at 0,0,0 in the scene, so `hmd-controls` can't necessarily return an absolute position. Unless we just want it to initialize with `getAttribute('position')` i guess. The flow now:

1. `hmd-controls` gets position from the HMD.
2. `hmd-controls` compares position to last position, and returns the delta.
3. `controls` computes velocity from the delta (`velocity = delta * 1000 / dt`).
4. `velocity` updates `position`.

donmccurdy:
Ah this gets back to what you said about not wanting any easing with the HMD though. Hm. Maybe we should just turn off easing by default anyway? Or I'm fine w/ going back to `getPosition`, but `hmd-controls` would need to be re-zeroed if you teleport or something.

ngokevin:
Not sure about having velocity update the position of the HMD. If you lose tracking, the velocity will just keep going, right? The tracking needs to be on point, velocity introduces some unpredictability with that.

Hmm, I'll have to think about the absolute vs. relative position. 

Yeah, no easing for tracked inputs. I'm not sure how that will play for flat experiences. Will need to think more on that also. For teleporting, you can use a wrapper entity.

donmccurdy:
Velocity won't keep going if HMD loses tracking - easing is very slight, so you'd stop almost instantly. I think removing easing (or just default=none) would be no problem; might as well let the input component do that if it's needed.

Jank-wise, it does look like the velocity component is consistently `tick()`'ing before controls. So the HMD is a frame behind. Is there anything in place to force `position` to update later, and can we do the same for `velocity`? Maybe position/rotation/velocity should be part of a system or something.

ngokevin:
The post-processing PR adds support for tick/tock. So you could put controls before position. I'll try to see if that helps.

donmccurdy:
Sounds good. I had to solve a similar problem with my [physics system](https://github.com/donmccurdy/aframe-extras/blob/master/src/physics/system/physics.js#L66-L79) using an `addBehavior(<component>, <phase>)` API, but that might be overkill here.

ngokevin:
I think a phase or order is starting to sound good. That's what the PP PR did at first... Will think about that.

ngokevin:
Would the next step for this be to decouple the controls from the velocity, and for physics we can have velocity be inferred based on changes to position?

donmccurdy:
That sounds fine, I'll pull out velocity.

You mentioned getting some jank from hmd-controls – is that rotation, position, or both? I'm not noticing anything definite, but if it's rotation then there's an issue unrelated to velocity we'll want to track down...

ngokevin:
Not sure. I tested it by turning my head around quickly a lot, but maybe caused by subtle changes in position?

Are there any disadvantages of pulling out velocity?

donmccurdy:
Removed velocity, that seemed pretty harmless. I just turned easing off by default, as well – maybe that was the issue with the jank feeling? I'll test that out in a bit.

ngokevin:
Looks like velocity wasn't the issue, still getting some lag. Have you tried testing the PR vs. master? I usually test on the geometry gallery scene.

On the Chromium builds (WebVR API 1.0), master is smooth, and the PR has some slight jank when turning head.

On Firefox Nightly (old WebVR API), master is smooth, and the PR stutters a lot when turning head.

donmccurdy:
Found the latency problem-- `hmd-controls::tick()` was called *after* `controls::tick()`, so HMD updates were out of sync. Might want a different solution once #1437 is resolved, but for now I've moved the main controls update into the controls system, so that it's guaranteed to happen after `hmd-controls` has had a chance to update its `THREE.VRControls` instance.

ngokevin:
Awesome, everything looks good so far. Looks amazingly smooth on CV1. The Vive looks good so far, I'll have to side-by-side compare it to master more.

In the component property, I would rename rotationControlsEnabled/positionControlsEnabled to just rotationEnabled/positionEnabled. To be consistent and less redundant within the context.

ngokevin:
@dmarcos This seems ready for review after passing manual regression testing.

Note that we disabled `keyboard-controls` by default. Perhaps let's enable `keyboard-controls` by default again, but perhaps put it before `hmd-controls` in terms of precedence. This allows flat experiences to move around, but dissuades against using it for VR. Either way, this will be a change, what do you prefer?

donmccurdy:
Updated w/ ____ControlsEnabled → _____Enabled renaming.

One other change, at least in the current state of things — `hmd-controls` and `mouse-controls` do not support touch events, whereas `look-controls` does. I removed that while the touch-to-move component was in the PR, which has since been removed. Do we want touch events back?

src/components/controls/mouse-controls.js
```diff
@@ -0,0 +1,107 @@
...
+    var previousMouseEvent = this.previousMouseEvent;
+
+    if (!this.data.enabled || !this.mouseDown) { return; }
+
+    lookVector.x += event.screenX - previousMouseEvent.screenX;
+    lookVector.y += event.screenY - previousMouseEvent.screenY;
+    this.previousMouseEvent = event;
+  },
+
+  onMouseDown: function (event) {
```
ngokevin: While we're here, can we use the `grab`/`grabbing` cursors if this component is in effect?

src/components/controls/mouse-controls.js
```diff
@@ -0,0 +1,107 @@
...
+var utils = require('../../utils');
+
+var warn = utils.debug('components:mouse-controls:warn');
+
+/**
+ * Mouse-controls component.
+ *
+ * Updates rotation when dragging the mouse.
+ */
+module.exports.Component = registerControls('mouse-controls', {
```
ngokevin: `look-controls` is being updated to listen to drag on window rather than canvas: https://github.com/aframevr/aframe/pull/1474

src/components/controls/controls.js
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Controls component.
+ *
+ * Receives input events from device-specific control components, and applies movement and rotation
+ * to the element accordingly.
+ */
+module.exports.Component = registerControls('controls', {
+  schema: {
+    enabled: { default: true },
```
ngokevin: let's start changing these to w/o the space `{default: true}`

ngokevin:
Yeah, I think we can have touch-drag back, but only if it doesn't affect tapping.

src/components/controls/mouse-controls.js
donmccurdy: That's just CSS right? Would rather do that in a separate commit if so.

donmccurdy:
@ngokevin about adding `keyboard-controls` before `hmd-controls`, what's the effect? Keyboard controls aren't active unless a key is down, so I don't follow how/if that would dissuade their use in VR (although I agree with the goal).

donmccurdy:
I've added TODOs in the original comment for those last two items (resolving conflicts with #1474, and adding `touch-controls`).

Before I do that though, @dmarcos, do you have concerns with merging this into core, or is it just a matter of doing a full review? I'm really just playing catch-up-and-rebase at this point, and I see you have incoming changes to `look-controls`.. Fine with me if this should be separate, or put into an `aframevr/aframe-controls` repo or something, but let's make that decision with v0.3.0 coming up. Thanks!

dmarcos:
Let's look at this first thing post 0.3.0. This is a big change and I'm worried about rushing it.

ngokevin:
I'm currently writing integration unit tests for controls to cover all the cases, protect from regressions, and since there's currently no coverage. Combined with your unit tests, that should make getting this large change through easier once we get to it.

donmccurdy:
@ngokevin sounds good! I'm traveling until Monday so I'll be slow on replies, but will try to take a look at those integration tests.

ngokevin:
I think we'll close this for now due to age and amount of work. What A-Frame has now can be considered a sensible default geared for WebVR, and other control schemes are welcome as community plugins. Both thanks and sorry for the amount of work on this. Could perhaps revisit later on?

Though I think perhaps we could take the Pointer Lock pieces and hook it up to full screen mode for A-Frame if you don't have a headset.

donmccurdy:
Makes sense to me, no problem. There are also enough open questions about encapsulating a 'player' and keeping arbitrary controller components aligned with teleporting, that maybe it's better to come at any controls changes from the direction of roomscale, rather than touch/gamepad/keyboard as done here.

donmccurdy:
(Referring to #2265)

Issue #1168, opened 2016-03-14
animation.stop() should consider fill value


IagoLast:
With the current code when an animation is removed from a component the animated property is set to the initial value.

This behavior could not be the desired one in some cases, i.e: when you want to stop an animation before it ends keeping the property value. Maybe I´m wrong and there is a better way for doing this, let me know in that case.

See https://github.com/IagoLast/totem for a working example when this feature could be used.
(Note that totem uses a modified a-frame version where the `setInitialValue` line was commented.)

dmarcos:
@IagoLast  Does setting fill="backwards" / fill="forwards" do what you need? 

https://aframe.io/docs/core/animation.html#Fill

ngokevin:
That won't do. We have code that always resets to initial value on ```.stop()```.

dmarcos:
Maybe ```stop``` should consider the fill value.

IagoLast:
@dmarcos yep, for me this would be the expected behavior

PR #1252, opened 2016-03-24
Animation.stop() should consider fill value (#1168)

IagoLast:
#### When animation.stop() it´s called, the animation value considers fill value

- **backwards**: : When animation is stopped set the starting value to the initial value.
- **both**: Animation value is saved when stopped.
- **forwards**: Animation value is saved when stopped.
- **none**: When animation is stopped set the starting value to the initial value.

### DIFF
--- a/tests/core/a-animation.test.js
+++ b/tests/core/a-animation.test.js
@@ -362,8 +362,7 @@ suite('a-animation', function () {
     });
 
     test('gets correct values coordinate component with no `from`', function () {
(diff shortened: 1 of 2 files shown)


tests/core/a-animation.test.js
```diff
@@ -399,6 +398,100 @@ suite('a-animation', function () {
         done();
       });
     });
+
+    suite('considers fill value', function () {
```
ngokevin: Wow, nice tests.

ngokevin:
Thanks!

PR #1256, opened 2016-03-24
Fixed error of function handleVideoTextureLoaded @ texture.js:103

mayognaise:
There was an error in handleVideoTextureLoaded if no video element exist.

error
code

Now it only emits/adds event lister if videoEl exits.

This still works with multiple videos and entities.
[Here is a reference video (please use webkit based browser)](http://cl.meowyo.com/fVRi)

### DIFF
--- a/src/utils/texture.js
+++ b/src/utils/texture.js
@@ -64,6 +64,7 @@ function loadVideo (material, data, src) {
   var hash;
   var texture;
   var videoEl;
+  var videoTextureResult;
 
   if (typeof src !== 'string') {
     // Check cache before creating texture.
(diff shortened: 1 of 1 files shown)


ngokevin:
Do you know in what case would there be no videoEl?

mayognaise:
Hi @ngokevin it is a really good point.
Actually the issue was the `videoEl` has overwritten by arguments.
As the `videoEl` is  in the same scope, it doesn't have to be passed.

So I simple removed the `videoEl` in the 2nd argument and it works properly :smile:

src/utils/texture.js
```diff
@@ -96,7 +96,7 @@ function loadVideo (material, data, src) {
   textureCache[calculateVideoCacheHash(videoEl)] = Promise.resolve(texture, videoEl);
   handleVideoTextureLoaded(texture, videoEl);
 
-  function handleVideoTextureLoaded (texture, videoEl) {
+  function handleVideoTextureLoaded (texture) {
```
ngokevin: I don't think this will work. Other textures waiting on the promise need to know about the `videoEl`. Maybe we should try renaming this parameter? `resolvedVideoEl`.

src/utils/texture.js
```diff
@@ -94,9 +94,9 @@ function loadVideo (material, data, src) {
 
   // Cache as promise to be consistent with image texture caching.
   textureCache[calculateVideoCacheHash(videoEl)] = Promise.resolve(texture, videoEl);
-  handleVideoTextureLoaded(texture, videoEl);
+  handleVideoTextureLoaded(texture);
 
-  function handleVideoTextureLoaded (texture, videoEl) {
```
ngokevin: I think we need this parameter for when the promise resolves. It might just be a matter of renaming this var.

mayognaise:
Found `Promise.resolve` only can pass one argument so made it as array

src/utils/texture.js
```diff
@@ -93,10 +93,12 @@ function loadVideo (material, data, src) {
   texture.minFilter = THREE.LinearFilter;
 
   // Cache as promise to be consistent with image texture caching.
-  textureCache[calculateVideoCacheHash(videoEl)] = Promise.resolve(texture, videoEl);
-  handleVideoTextureLoaded(texture, videoEl);
+  textureCache[hash] = Promise.resolve([texture, videoEl]);
```
ngokevin: Ah, I see! Can you create an object instead?

```js
var videoTextureResult;  
// ...
videoTextureResult = {
  texture: texture,
  videoEl: videoEl
};
// ... resolve(videoTextureResult)
```

mayognaise: sure!

ngokevin:
Thanks!

ngokevin:
Oops, pulled a @dmarcos and forgot to squash.

PR #1257, opened 2016-03-24
API to merge entity meshes to reduce the number of draw calls

dmarcos:
It is destructive / no revertible. Once an entity is merged into another one the entity is removed from the scene. Making this dynamic would add a lot of overhead (we would have to cache all the meshes in case the operation wants to be reverted). It would also lead to tons of edge cases that will affect performance in ways that are difficult to understand. For instance if you change position or material one of the entities the whole merge group would have to be regenerated. On the provided example we reduce the number of calls from 1988 to a single one when merging the entities.

### DIFF
--- a/examples/test-merge-geometry/index.html
+++ b/examples/test-merge-geometry/index.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Merge Geometries Test</title>
+    <meta name="description" content="Merge Geometry - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+    <style>
+      body { background-color: black; }
+    </style>
+  </head>
+  <body>
+    <div class="stats"></div>
+    <a-scene stats="true">
+      <a-assets>
+        <a-mixin id="cube"
+                 geometry="primitive: box; height: 16; width: 16; depth: 16; mergeTo: #boxes; skipCache: true; buffer: false;"
+                 material="color: #167341; roughness: 1.0; metalness: 0.2;"></a-mixin>
+      </a-assets>
+      <a-entity position="0 20 0">
+        <a-entity camera wasd-controls look-controls></a-entity>
+      </a-entity>
+      <a-entity id="boxes" material="color: #167341; roughness: 1.0; metalness: 0.2;"></a-entity>
+    </a-scene>
+    <script>
+      var cubeEl;
+      var sceneEl = document.querySelector('a-scene');
+      var xStartPos = -90
+      var zStartPos = -10;
+      var xPos;
+      var zPos;
+      var xMargin = 20;
+      var zMargin = -20;
+      var cubesNum = 200;
+      var rootEl;
+      var x;
+      zPos = zStartPos;
+      for (var i = 0; i < cubesNum; i += 1) {
+        xPos = xStartPos;
+        for (var j = 0; j < 10; j += 1) {
+          cubeEl = document.createElement('a-entity');
+          cubeEl.setAttribute('mixin', 'cube');
+          cubeEl.setAttribute('position', { x: xPos, y: 0, z: zPos});
+          xPos += xMargin;
+          sceneEl.appendChild(cubeEl);
+        }
+        zPos += zMargin;
+      }
+    </script>
+  </body>
+</html>
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -1,8 +1,10 @@
+var debug = require('../utils/debug');
 var geometries = require('../core/geometry').geometries;
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 
 var dummyGeometry = new THREE.Geometry();
+var warn = debug('components:geometry:warn');
 
 /**
  * Geometry component. Combined with material component to make a mesh in 3D object.
@@ -10,9 +12,10 @@ var dummyGeometry = new THREE.Geometry();
  */
 module.exports.Component = registerComponent('geometry', {
   schema: {
-    buffer: {default: true},
-    primitive: {default: ''},
-    skipCache: {default: false}
+    buffer: { default: true },
+    mergeTo: { type: 'selector' },
+    primitive: { default: '' },
+    skipCache: { default: false }
   },
 
   init: function () {
@@ -35,6 +38,52 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+    if (data.mergeTo) {
+      this.mergeTo(data.mergeTo);
+    }
+  },
+
+  /**
+   * It merges the geometry into another entity's geometry. The original
+   * entity is removed from the scene. This is a not reversible operation
+   *
+   * @param {object} toEl - The entity where the geometry will be merged to
+   */
+  mergeTo: function (toEl) {
+    var el = this.el;
+    var mesh = el.getObject3D('mesh');
+    var toMesh;
+    if (!toEl) {
+      warn('There is not a valid entity to merge the geometry to');
+      return;
+    }
+    if (toEl === el) {
+      warn('Source and target geometries cannot be the same for merge.');
+      return;
+    }
+    toMesh = toEl.getObject3D('mesh');
+    // If the to entity does not have a mesh we create one
+    if (!toMesh) {
+      toMesh = toEl.getOrCreateObject3D('mesh', THREE.Mesh);
+      toEl.setAttribute('material', el.getComputedAttribute('material'));
+      return;
+    }
+
+    if (toMesh.geometry instanceof THREE.Geometry === false ||
+        mesh.geometry instanceof THREE.Geometry === false) {
+      warn('Geometry merge is only available for `THREE.Geometry` types');
+      return;
+    }
+
+    if (this.data.skipCache === false) {
+      warn('Cached geometries are not allowed to merge. Set `skipCache` to true');
+      return;
+    }
+
+    mesh.parent.updateMatrixWorld();
+    toMesh.geometry.merge(mesh.geometry, mesh.matrixWorld);
+    el.emit('geometry-merged', { mergeTarget: toEl });
+    el.sceneEl.removeChild(el);
   },
 
   /**
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -267,10 +267,10 @@ module.exports.registerComponent = function (name, definition) {
   components[name] = {
     Component: NewComponent,
     dependencies: NewComponent.prototype.dependencies,
-    parse: NewComponent.prototype.parse.bind(NewComponent.prototype),
-    parseAttrValueForCache: NewComponent.prototype.parseAttrValueForCache.bind(NewComponent.prototype),
+    parse: NewComponent.prototype.parse,
+    parseAttrValueForCache: NewComponent.prototype.parseAttrValueForCache,
     schema: utils.extend(processSchema(NewComponent.prototype.schema)),
-    stringify: NewComponent.prototype.stringify.bind(NewComponent.prototype),
+    stringify: NewComponent.prototype.stringify,
     type: NewComponent.prototype.type
   };
   return NewComponent;
--- a/src/systems/geometry.js
+++ b/src/systems/geometry.js
@@ -59,9 +59,13 @@ module.exports.System = registerSystem('geometry', {
     var cache = this.cache;
     var cacheCount = this.cacheCount;
     var geometry;
-    var hash = this.hash(data);
+    var hash;
 
-    if (!cache[hash] || data.skipCache) { return; }
+    if (data.skipCache) { return; }
+
+    hash = this.hash(data);
+
+    if (!cache[hash]) { return; }
 
     decrementCacheCount(cacheCount, hash);
 
--- a/examples/index.html
+++ b/examples/index.html
@@ -152,7 +152,6 @@
       <li><a href="test-mixin/">Mixin</a></li>
       <li><a href="test-model/">Model</a></li>
       <li><a href="test-opacity/">Opacity</a></li>
-      <li><a href="test-panoexplorer/">Pano Explorer</a></li>
       <li><a href="test-physical/">Physically-Based Materials</a></li>
       <li><a href="test-pivot/">Pivot</a></li>
       <li><a href="test-raycaster/">Raycaster</a></li>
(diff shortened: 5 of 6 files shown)


src/systems/geometry.js
```diff
@@ -59,9 +59,13 @@ module.exports.System = registerSystem('geometry', {
     var cache = this.cache;
     var cacheCount = this.cacheCount;
     var geometry;
-    var hash = this.hash(data);
+    var hash;
+
+    if (data.skipCache || data.mergeTo) { return; }
+
+    hash = this.hash(data);
```
dmarcos: This is still not very clean. It's to avoid the error ```TypeError: cyclic object value``` when sending a selector type property to the JSON.parse function

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Cube</title>
```
ngokevin: Fix title

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Cube</title>
+    <meta name="description" content="Merge Geometry - A-Frame">
```
ngokevin: I'm proposing in another PR to namespace these tests under `performance`

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Cube</title>
+    <meta name="description" content="Merge Geometry - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+    <style>
+      body { background-color: black; }
```
ngokevin: needed?

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,64 @@
...
+  <body>
+    <div class="stats"></div>
+    <a-scene stats="true">
+      <a-assets>
+        <a-mixin id="cube"
+                 geometry="primitive: box; height: 16; width: 16; depth: 16; mergeTo: #boxes; skipCache: true; buffer: false"
+                 material="color: #167341; roughness: 1.0; metalness: 0.2;"></a-mixin>
+      </a-assets>
+      <a-entity position="0 20 0">
+        <a-entity camera="active:true" wasd-controls="" look-controls="" rotation="0 0 0" position="0 0 0"></a-entity>
```
ngokevin: `active: true`, `rotation`, and `position`, and the `=""` aren't needed

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,64 @@
...
+      </a-entity>
+      <a-entity id="boxes" mixin="cube" geometry="height: 0"></a-entity>
```
ngokevin: `height: 0` is a bit awkward if you want to merge to a container.

we should probably just allow not specifying a "primitive" to just instantiate a blank `new THREE.Geometry()`.

src/components/geometry.js
```diff
@@ -24,6 +29,7 @@ module.exports.Component = registerComponent('geometry', {
    */
   update: function (previousData) {
     var data = this.data;
+    var diffData = diff(previousData || {}, data);
```
ngokevin: Try not to use `diff` when possible, it's more expensive than just doing an equality check

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+    if (diffData.mergeTo) {
```
ngokevin: Why not just `if (data.mergeTo)`

src/core/component.js
```diff
@@ -266,10 +266,10 @@ module.exports.registerComponent = function (name, definition) {
   components[name] = {
     Component: NewComponent,
     dependencies: NewComponent.prototype.dependencies,
-    parse: NewComponent.prototype.parse.bind(NewComponent.prototype),
```
ngokevin: Nice

src/systems/geometry.js
```diff
@@ -59,9 +59,13 @@ module.exports.System = registerSystem('geometry', {
     var cache = this.cache;
     var cacheCount = this.cacheCount;
     var geometry;
-    var hash = this.hash(data);
+    var hash;
+
+    if (data.skipCache || data.mergeTo) { return; }
```
ngokevin: Doesn't seem like you need this because you already check for `skipCache` in `geometry.mergeTo`

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+      if (!data.mergeTo) { return; }
+      this.mergeTo(data.mergeTo);
+    }
+  },
+
+  mergeTo: function (toEl) {
+    var el = this.el;
+    var mesh = this.el.getObject3D('mesh');
+    var toMesh = toEl.getObject3D('mesh');
+    if (!toEl) {
```
ngokevin: You'd probably check this before doing `getObject3D`

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+      warn('There is not a valid entity to merge the geometry to');
+      return;
+    }
+    if (toEl === el) { return; }
```
ngokevin: Warn here

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+    if (!toMesh) {
+      warn('Merge is not posible. The entity ' + toEl.id + ' doesn\'t have a mesh');
+      return;
+    }
+
+    if (toMesh.geometry instanceof THREE.Geometry === false) {
+      warn('Geometry merge is only available for THREE.Geometry types');
```
ngokevin: Is it valid to have a  `mesh.geometry` that isn't an instance of THREE.Geometry?

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+    var el = this.el;
+    var mesh = this.el.getObject3D('mesh');
+    var toMesh = toEl.getObject3D('mesh');
+    if (!toEl) {
+      warn('There is not a valid entity to merge the geometry to');
+      return;
+    }
+    if (toEl === el) { return; }
+    if (!toMesh) {
+      warn('Merge is not posible. The entity ' + toEl.id + ' doesn\'t have a mesh');
```
ngokevin: possible

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+      return;
+    }
+
+    if (toMesh.geometry instanceof THREE.Geometry === false) {
+      warn('Geometry merge is only available for THREE.Geometry types');
+      return;
+    }
+
+    if (this.data.skipCache === false) {
+      warn('Cached geometries are not allowed to merge. Set skipCache to true');
```
ngokevin: backtick `skipCache` and `THREE.Geometry`

src/components/geometry.js
```diff
@@ -9,10 +12,12 @@ var dummyGeometry = new THREE.Geometry();
  * Extended with registered geometries.
  */
 module.exports.Component = registerComponent('geometry', {
+  dependencies: ['position', 'rotation'],
```
ngokevin: newline after this.

is this needed? position/rotation are default components. are they being initialized after?

src/components/geometry.js
```diff
@@ -1,6 +1,9 @@
+var debug = require('../utils/debug');
+var diff = require('../utils').diff;
 var geometries = require('../core/geometry').geometries;
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var warn = debug('components:geometry:warn');
```
ngokevin: nit: can move this under `var dummyGeometry` to separate from requires

examples/test-merge-geometry/index.html
ngokevin: Alternatively, `mergeTo` could handle creating a `THREE.Geometry` on the mesh. Then we won't have to involve the geometry component at all on the target entity.

Either way, would like to not have to do `height: 0` when treating the target entity as a container for a mergeTo.

src/components/geometry.js
```diff
@@ -35,6 +41,39 @@ module.exports.Component = registerComponent('geometry', {
...
+      return;
+    }
+
+    if (this.data.skipCache === false) {
+      warn('Cached geometries are not allowed to merge. Set skipCache to true');
+      return;
+    }
+
+    mesh.parent.updateMatrixWorld();
+    toMesh.geometry.merge(mesh.geometry, mesh.matrixWorld);
```
ngokevin: Might be nice to emit an event on the target entity (e.g, `geometry-merged`). I might want to switch the final geometry to a buffer geometry after everything is merged, would be nice to have a declarative way of doing so

examples/test-merge-geometry/index.html
dmarcos: Yup. Otherwise the background is white.

src/components/geometry.js
```diff
@@ -35,6 +38,46 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+    if (data.mergeTo) {
+      if (!data.mergeTo) { return; }
```
ngokevin: Don't need this line.

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
...
+      </a-assets>
+      <a-entity position="0 20 0">
+        <a-entity camera="active:true" wasd-controls look-controls></a-entity>
+      </a-entity>
+      <a-entity id="boxes"></a-entity>
+    </a-scene>
+    <script>
+      var cubeEl;
+      var sceneEl = document.querySelector('a-scene');
+      sceneEl.addEventListener('renderstart', function() {
```
ngokevin: Space after function

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
...
+  <body>
+    <div class="stats"></div>
+    <a-scene stats="true">
+      <a-assets>
+        <a-mixin id="cube"
+                 geometry="primitive: box; height: 16; width: 16; depth: 16; mergeTo: #boxes; skipCache: true; buffer: false;"
+                 material="color: #167341; roughness: 1.0; metalness: 0.2;"></a-mixin>
+      </a-assets>
+      <a-entity position="0 20 0">
+        <a-entity camera="active:true" wasd-controls look-controls></a-entity>
```
ngokevin: Don't need `active: true`, right?

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
...
+      var sceneEl = document.querySelector('a-scene');
+      sceneEl.addEventListener('renderstart', function() {
+        var statsEl = document.querySelector('.stats');
+        statsEl.innerHTML = cubesNum + " entities rendered</br> The scene took " + window.performance.getEntriesByName('render-started')[0].startTime.toFixed(2) + " ms to load";
+      });
+      var x = -90;
+      var z = -10;
+      var cubesNum = 0;
+      var rootEl;
+      for (var i = 0; i < 200; i+=1) {
```
ngokevin: Could use less magic numbers for these loops to be more readable. Like:

```js
var xMargin = 20;
var xPosStart = -90;
var zPosStart = -20;
var zMargin = -20;
var cubesNum = 200;  // Don't need to do cubeNums += 1 then.
```

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
...
+          cubeEl.setAttribute('mixin', 'cube');
+          cubeEl.setAttribute('position', { x: x, y: 0, z: z});
+          x += 20;
+          cubesNum += 1;
+          sceneEl.appendChild(cubeEl);
+        }
+        x = -90;
+        z -= 20;
+      }
+      document.querySelector('#boxes').setAttribute(
```
ngokevin: Just define this on the HTML.

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
...
+      sceneEl.addEventListener('renderstart', function() {
+        var statsEl = document.querySelector('.stats');
+        statsEl.innerHTML = cubesNum + " entities rendered</br> The scene took " + window.performance.getEntriesByName('render-started')[0].startTime.toFixed(2) + " ms to load";
+      });
+      var x = -90;
+      var z = -10;
+      var cubesNum = 0;
+      var rootEl;
+      for (var i = 0; i < 200; i+=1) {
+        for (var j = 0; j < 10; j+=1) {
```
ngokevin: Space before/after `+=`

src/components/geometry.js
```diff
@@ -35,6 +38,46 @@ module.exports.Component = registerComponent('geometry', {
...
+      this.mergeTo(data.mergeTo);
+    }
+  },
+
+  mergeTo: function (toEl) {
+    var el = this.el;
+    var mesh = this.el.getObject3D('mesh');
```
ngokevin: Use `el`

src/components/geometry.js
```diff
@@ -35,6 +38,46 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+    if (data.mergeTo) {
+      if (!data.mergeTo) { return; }
+      this.mergeTo(data.mergeTo);
+    }
+  },
+
+  mergeTo: function (toEl) {
```
ngokevin: docstring

src/components/geometry.js
```diff
@@ -35,6 +38,46 @@ module.exports.Component = registerComponent('geometry', {
...
+    var toMesh;
+    if (!toEl) {
+      warn('There is not a valid entity to merge the geometry to');
+      return;
+    }
+    if (toEl === el) {
+      warn('Source and target geometries cannot be the same for merge.');
+      return;
+    }
+    toMesh = toEl.getObject3D('mesh');
```
ngokevin: nit: Probably okay to just do this in the declaration. It's an object access.

src/components/geometry.js
```diff
@@ -35,6 +38,46 @@ module.exports.Component = registerComponent('geometry', {
...
+    if (!toMesh) {
```
ngokevin: I'd comment what this block does

examples/test-merge-geometry/index.html
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
```
ngokevin: Add to listing index.html

ngokevin:
r+wc

PR #1259, opened 2016-03-24
This avoids an extra builData call for components with static schema …

dmarcos:
We are introducing an extra call to buildData for all of the components regardless if they have dynamic squema or not. This commit removes the overhead for all the component with static schemas


ngokevin:
Nice. This is testable by `sinon.spy`ing the number of calls to buildData.

dmarcos:
The problem is that the function is going to be called if defined. If you mock it you define it and if it's not defined there's nothing to spy.

dmarcos:
Well yes. we can count the calls to buildData

ngokevin:
r+

PR #1263, opened 2016-03-24
Adds aframe specific numbers to rStats

dmarcos:
like number of entities and scene load time

rStats

### DIFF
--- a/vendor/rStats.extras.js
+++ b/vendor/rStats.extras.js
@@ -129,10 +129,10 @@ window.threeStats = function ( renderer ) {
     };
 
     var _groups = [ {
-        caption: 'three.js - Memory',
+        caption: 'Three.js - memory',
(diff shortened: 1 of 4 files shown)


vendor/rStats.extras.js
```diff
@@ -1,3 +1,55 @@
+window.aframeStats = function (scene) {
```
ngokevin: Can this go into ```lib/```?

ngokevin:
r+wc

vendor/rStats.extras.js
```diff
@@ -256,6 +256,7 @@ window.BrowserStats = function () {
 
 if (typeof module === 'object') {
   module.exports = {
+    aframeStats: window.aframeStats,
```
ngokevin: Does rStats extras need to know about this?

PR #1270, opened 2016-03-25
Add intensity to ambient light (fixes #1201)

### LINKED ISSUE #1201: add intensity property to ambient light
**Description:** three.js r74 introduced intensity property on ambient light https://github.com/mrdoob/three.js/pull/7621 . Add that to the light component.

- A-Frame Version: master 
- Platform/Device: all

MichaelRupertDev:
**Description:**
three.js r74 introduced intensity property on ambient light mrdoob/three.js#7621 . Add that to the light component.
**Changes proposed:**
- updated component
- added test

### DIFF
--- a/tests/components/light.test.js
+++ b/tests/components/light.test.js
@@ -36,6 +36,12 @@ suite('light', function () {
       el.setAttribute('light', 'type: ambient');
       assert.equal(el.object3D.children[0].type, 'AmbientLight');
     });
+
+    test('can update parameters on updated light', function () {
+      var el = this.el;
+      el.setAttribute('light', 'intensity: 5');
+      assert.equal(el.object3D.children[0].intensity, 5);
+    });
   });
 
   suite('getLight', function () {
--- a/docs/components/light.md
+++ b/docs/components/light.md
@@ -25,7 +25,7 @@ We will go through the different types of lights and their respective properties
 
 ### Ambient
 
-Ambient lights are applied to all entities in the scene globally. They are defined only by the `color` property. And, `position`, `rotation`, and `scale` have no effect on ambient lights.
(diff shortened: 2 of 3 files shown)


tests/components/light.test.js
```diff
@@ -38,6 +38,12 @@ suite('light', function () {
     });
   });
 
+  test('can update parameters on updated light', function () {
```
dmarcos: i would just update the test ```updatesLight``` above to also check for the intensity value

dmarcos:
Almost there!

tests/components/light.test.js
MichaelRupertDev: The only reason I didn't do that is because at that time, it's a directional light, rather than an ambient light.

dmarcos: This test should be anyways in the update suite and not hanging by itself

michaelrupert-wf:
@dmarcos Updated.

tests/components/light.test.js
```diff
@@ -36,6 +36,12 @@ suite('light', function () {
       el.setAttribute('light', 'type: ambient');
       assert.equal(el.object3D.children[0].type, 'AmbientLight');
     });
+
+    test('can update parameters on updated light', function () {
+      var el = this.el;
+      el.setAttribute('light', 'intensity: 5');
+      assert.shallowDeepEqual(el.object3D.children[0].intensity, 5);
```
ngokevin: A simple `equal` would do

dmarcos: Let's replace this with an equal and squash the commits.

ngokevin:
Can you update the docs as well?

tests/components/light.test.js
dmarcos: If we can replace ```shallowDeepEqual``` with ```equal``` I think we can merge

MichaelRupertDev:
@dmarcos @ngokevin Finally fixed the messed up rebase, should be good to go

dmarcos:
Nice! Thank you for keeping at it

PR #1273, opened 2016-03-25
Three and aframe rStats introduced a lot of overhead and are too big …

dmarcos:
…for a small screen. This patch disables the rStats plugins in mobile


src/components/scene/stats.js
```diff
@@ -48,6 +49,7 @@ module.exports.Component = registerComponent('stats', {
 function createStats (scene) {
   var threeStats = new ThreeStats(scene.renderer);
   var aframeStats = new AFrameStats(scene);
+  var plugins = isMobile ? [] : [threeStats, aframeStats];
```
ngokevin: you can use ```sceneEl.isMobile```

ngokevin:
r+wc

PR #1281, opened 2016-03-28
Update building-an-advanced-scene.md

captDaylight:
Superfluous variable, changed in doc, needs to be changed in the codepen but I imagine you'd want to do that through your account.


cvan:
very good catch. I updated the CodePen pen. thanks!

PR #1287, opened 2016-03-28
dispose geometries + materials when no longer in use (fixes #1215)

### LINKED ISSUE #1215: deallocate geometries and materials when no longer in use
**Description:** Possibly on entity detach and we detect that its geometry, material, and/or texture are no longer in use. iOS crashing sometimes due to running out of memory. One test case is to have one photosphere with six panos, and switching back and forth them a lot.

http://mrdoob.github.io/three.js/examples/webgl_test_memory.html

- A-Frame Version: master
- Platform/Device: all

ngokevin:
**Description:** Reduce memory leaks.

Before, lots of geometries being created on every tick for the torus knot:

screen shot 2016-03-28 at 2 55 32 pm

After, the geometry count stays at 3.

screen shot 2016-03-28 at 2 55 21 pm

**Changes proposed:**
- Dispose geometry on type update or remove.
- Dispose material on shader update or remove.

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -105,12 +105,14 @@ module.exports.Component = registerComponent('material', {
 
   /**
    * Remove material on remove (callback).
+   * Dispose of it from memory and unsubscribe from scene updates.
    */
   remove: function () {
     var defaultMaterial = new THREE.MeshBasicMaterial();
+    var material = this.material;
     var object3D = this.el.getObject3D('mesh');
     if (object3D) { object3D.material = defaultMaterial; }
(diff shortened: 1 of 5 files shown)


src/components/material.js
```diff
@@ -151,3 +153,11 @@ function parseSide (side) {
     }
   }
 }
+
+/**
+ * Dispose of material from memory and unsubscribe material from scene updates like fog.
+ */
+function disposeMaterial (material, system) {
+  material.dispose();
```
dmarcos: is this order you want? first dispose and later unregister?

PR #1292, opened 2016-03-29
Mobile support message

dmarcos:
Message to manage expectations about smartphone support. I'm not sure about the wording and where this should go. This is just a starting point for a conversation.

Smartphone support has turned into a beast on itself. While handy in the short term to gain adoption is going to get less and less important over time. Support and maintenance across different hardware, OSs and browser is adding complexity to the code base that will make harder to implement higher value features. We all agree that cardboard is not proper VR right? :). 

@ngokevin @cvan What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,3 +105,7 @@ This program is free software and is distributed under an [MIT License](LICENSE)
 [downloads-url]: https://npmjs.org/package/aframe
 [semistandard-image]: https://img.shields.io/badge/code%20style-semistandard-brightgreen.svg?style=flat-square
 [semistandard-url]: https://github.com/Flet/semistandard
+
+## Mobile Support
+
+A-Frame smartphone support is limited and not the core focus of the project. We have some compatibility (thanks to webvr-polyfill) for testing purposes and as a convenient way to quickly check scenes without having a headset at hand. We're a small team and we don't have the resources to support all the permutations of browsers and hardware. We test on iPhone6 and Samsung Galaxy S6 which are the phones we personally own. If you have a problem specific to your device we won't be able to investigate it but we're glad to review patches that fix those particular problems. Thanks for the understanding.


ngokevin:
I might put it in the FAQ in response to one of the questions relating to mobile support. I do like that, something along the lines of there is some support, but we're not going to focus on it.

README.md
```diff
@@ -105,3 +105,7 @@ This program is free software and is distributed under an [MIT License](LICENSE)
 [downloads-url]: https://npmjs.org/package/aframe
 [semistandard-image]: https://img.shields.io/badge/code%20style-semistandard-brightgreen.svg?style=flat-square
 [semistandard-url]: https://github.com/Flet/semistandard
+
+## Mobile Support
+
+A-Frame smartphone support is limited and not the core focus of the project. We have some compatibility (thanks to webvr-polyfill) for testing purposes and as a convenient way to quickly check scenes without having a headset at hand. We're a small team and we don't have the resources to support all the permutations of browsers and hardware. We test on iPhone6 and Samsung Galaxy S6 which are the phones we personally own. If you have a problem specific to your device we won't be able to investigate it but we're glad to review patches that fix those particular problems. Thanks for the understanding.
```
cvan: `iPhone 6`

cvan: comma before `but we're glad`

cvan: perhaps also link to FAQ for some of the mobile issues.

I like that we're saying this. thanks, @dmarcos

README.md
```diff
@@ -105,3 +105,7 @@ This program is free software and is distributed under an [MIT License](LICENSE)
 [downloads-url]: https://npmjs.org/package/aframe
 [semistandard-image]: https://img.shields.io/badge/code%20style-semistandard-brightgreen.svg?style=flat-square
 [semistandard-url]: https://github.com/Flet/semistandard
+
+## Mobile Support
```
cvan: `Caveats about Mobile Support` or `The State of Mobile Support`?

README.md
ngokevin: i would put the whole thing the FAQ. the README is a bit of an awkward place to put an important message.

cvan: yeah, I think you're right. just hope nobody misses it.

we should also have somewhere prevalent a table of browse support

ngokevin:
Going to elect to move to FAQ. Once we get a browser support page on the site, we can move it there as well. I like this messaging.

ngokevin:
Though note that the wording made several people think A-Frame was dropping Cardboard completely.

PR #1300, opened 2016-03-30
Added cone to the list of supported primitives in geometry schema

fernandojsg:
**Description:** Cone primitive was missing on the list of supported primitives for the attribute ```primitive```.


dmarcos:
We need to update the docs too

fernandojsg:
@dmarcos the documents already has the correct documentation for `cone` and `a-cone`

Issue #733, opened 2016-01-13
Support animating colors


ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ngokevin](https://github.com/ngokevin)**
_Wednesday Dec 09, 2015 at 23:09 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/640_

----

ngokevin:
<a href="https://github.com/dmarcos"><img src="https://avatars.githubusercontent.com/u/39342?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dmarcos](https://github.com/dmarcos)**
_Wednesday Dec 09, 2015 at 23:14 GMT_

----

If we move to three.js animation system we'll get color interpolation:
https://github.com/mrdoob/three.js/blob/master/src/animation/tracks/ColorKeyframeTrack.js#L3

ngokevin:
- Need to detect whether a property/attribute is a color value, somehow. Maybe if it starts with a hash and has six hexadecimal characters.
- Need to tangle thru the a-animation code to add logic for color.

ngokevin:
https://github.com/aframevr/aframe/pull/1302

PR #1302, opened 2016-03-30
adds the feature to animate a color property on the a-animation compo…

JohnRodney:
**Description:**
  Adds the support for using `a-animation` components in order to animate colors. #733
**Changes proposed:**
-  Adds the detection of a color as an attribute
-  Adds a partialSetAttribute for color
-  Adds a conversion from rgb object to hex (maybe these belong in utils?)
-  Adds two unit tests for start and finish to ensure the color is properly tweened

…nent


src/utils/coordinates.js
```diff
@@ -48,7 +48,7 @@ module.exports.stringify = stringify;
  * @returns {bool}
  */
 module.exports.isCoordinate = function (value) {
-  return regex.test(value);
+  return value[0] !== '#' && regex.test(value);
```
ngokevin: Color hexes were passing the regex test?

tests/core/a-animation.test.js
```diff
@@ -118,6 +118,34 @@ suite('a-animation', function () {
     });
   });
 
+  suite('component color animation', function () {
```
ngokevin: Would be nice to have tests for:

- Shorthand `#fff`
- Capital `#FFF`
- Interpolated value. Check that between the animation, the color is somewhere between the start and end.

src/utils/coordinates.js
JohnRodney: Yes

ngokevin: So coordinates will always have whitespaces in between the numbers. Colors won't. I see that the whitespace characters uses star (`*`) rather than (`+`) which will match even 0 whitespace characters. Need to change those to `+` to force at least 1 whitespace character.

try `/\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*/;`

tests/core/a-animation.test.js
JohnRodney: These test cases have been added.

src/core/a-animation.js
```diff
@@ -463,3 +479,35 @@ function strToBool (str) {
...
   return bool ? 1 : 0;
 }
+
+/**
+ * Converts a number 0-255 to hex
+ * @param {color}
+ * @returns {string}
+ */
+function componentToHex (color) {
+  var hex = color.toString(16);
```
dmarcos: Should we check if color < 0 or > 255 and clamp it to 0-255?

tests/core/a-animation.test.js
```diff
@@ -118,6 +118,78 @@ suite('a-animation', function () {
...
+
+    test('start value', function () {
+      assert.equal(this.el.getComputedAttribute('color'), '#ffffff');
+    });
+
+    test('between value', function () {
+      var color;
+      this.animationEl.tween.update(this.startTime + 500);
+      color = this.el.getComputedAttribute('color');
+      assert.isAbove(color, '#000000');
```
ngokevin: Very nice that you can assert like this.

tests/core/a-animation.test.js
ngokevin: Thanks!

src/core/a-animation.js
```diff
@@ -5,6 +5,7 @@ var parseProperty = require('./schema').parseProperty;
 var registerElement = require('./a-register-element').registerElement;
 var TWEEN = require('tween.js');
 var utils = require('../utils/');
+var THREE = require('three');
```
ngokevin: do ```../lib/three```

src/utils/coordinates.js
JohnRodney: This has been updated as well thanks for the help on the regex.

src/core/a-animation.js
JohnRodney: @dmarcos @ngokevin Hey this could be clamped in the function that calls this by doing a:
`  return Math.floor(Match.min(Math.abs(color), 255) * 255);`
inside of the convertToIntegerColor() function.  Would this be acceptable?  Or would you prefer a separate clamp function?

src/core/a-animation.js
```diff
@@ -363,6 +364,8 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
     getForCoordinateComponent();
   } else if (['true', 'false'].indexOf(dataTo) !== -1) {
     getForBoolean();
+  } else if (dataTo && dataTo[0] === '#') {
```
dmarcos: We should also use the schema to infer the color type and not just based on the '#' prefix. We could do:
````html
<a-animation attribute="material.color" from="blue" to="red" dur="4000" easing="ease-out"></a-animation>
````

File: src/core/a-animation.js (hunk 3 above)
Comment (JohnRodney) [inline src/core/a-animation.js]: This is also possibly unneeded as the value that is passed comes from a THREE.Color which is already tested to be within 0 and 1 from a hex value correct?

File: src/core/a-animation.js (hunk 6 above)
Comment (JohnRodney) [inline src/core/a-animation.js]: Good idea.  Can you point me to the right direction on this?  Is there already a currently existing way to check if these are valid colors or would this be something I need to add?

Comment (ngokevin) [inline src/core/a-animation.js]: I think ```#``` is more flexible. Using the schema would not work in cases like:

```html
<a-box color="#CCC">
  <a-animation attribute="color" to="#FFF"></a-animation>
</a-box>
```

dmarcos: The THREE.Color class does it for you.
http://threejs.org/docs/#Reference/Math/Color
You can do ```new THREE.Color('red').getHexString()````

dmarcos: API consistency is important. Let's do it now.

JohnRodney: Alright I'm about to do my hour drive home and then I'll try to parse through the schema code to make these adjustments.  May not get it done until tomorrow.  Thanks for all the feedback. @dmarcos @ngokevin

dmarcos: @JohnRodney Thanks for the effort. This is good stuff.

JohnRodney: @dmarcos @ngokevin.  Alright on the drive home I was thinking the catching else statement does an automatic type conversion to number which will turn out to be a NaN property on the object.  So instead of checking prop or color I figured we can just check for `isNaN(dataTo)` and then feed it directly to the Three.js conversions.  This will always get a color (white) if an invalid color is passed and will still throw a warning from three.js when the color is not valid.

This gets around checking specific color formats and allows Three.js to handle improper input. Is there any reason you would think this would fail?

JohnRodney:
I think all feedback is resolved and I went through a refactor phase.  It should be ready to squash the commits if everything else is approved.

ngokevin:
@dmarcos This looks pretty good. He made it not explicitly check for color, just let it fall through the if/else statements, saving about probably 50 LOC. I think even if we add more interpolation types, this should still hold.

src/core/a-animation.js
```diff
@@ -363,6 +364,8 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
     getForCoordinateComponent();
   } else if (['true', 'false'].indexOf(dataTo) !== -1) {
     getForBoolean();
+  } else if (isNaN(dataTo)) {
```
dmarcos: Does this work for the following case? Can we add it to the tests?
````html
<a-animation attribute="material.color" from="red" to="blue"></a-animation>
````

Comment (JohnRodney) [inline src/core/a-animation.js]: Good catch. I'll look at it in the morning.

File: src/core/a-animation.js [hunk 8]
```diff
@@ -391,7 +398,14 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
     }
     from[attribute] = parseProperty(from[attribute], schema[componentPropName]);
     to[attribute] = parseProperty(dataTo, schema[componentPropName]);
```
ngokevin: If our `color` property type parser converted to `THREE.Color` I think everything would work out of the box. But I'm not sure if that would break things.

src/core/a-animation.js
```diff
@@ -391,7 +398,14 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
...
+    if (componentPropName === 'color') {
+      from[attribute] = new THREE.Color(from[attribute]);
+      to[attribute] = new THREE.Color(to[attribute]);
+    }
     partialSetAttribute = function (value) {
+      if (attribute.indexOf('color') !== -1) {
```
ngokevin: You could do ```schema[componentPropName].type === 'color'```, but I'm still looking at the whole thing.

src/core/a-animation.js
```diff
@@ -138,7 +139,11 @@ module.exports.AAnimation = registerElement('a-animation', {
         }
 
         // Create Tween.
-        return new TWEEN.Tween(cloneValue(from))
+        if (from['material.color']) {
```
ngokevin: How come we need this?

JohnRodney: I'm not sure.  I am sure it was added for a reason because of some case to lose an object reference since all object assignments are references in JS.  I haven't dug in to see if this is a shallow clone or a deep clone.  

I mainly made this commit to share how I was able to 'fix' the issue, but I wasn't completely sure why these changes fixed it. I know its just an object reference somewhere.

JohnRodney:
@ngokevin @dmarcos Updated with a new approach to support both `color` and `material.color`.  I feel better about this approach.  Will add in some unit tests to ensure no regressions are introduced around the 'material.color' property.

src/core/a-animation.js
```diff
@@ -356,13 +357,14 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
   var from = {};
   var partialSetAttribute;
   var to = {};
-
-  if (attributeSplit.length === 2) {
+  if (attributeSplit.length === 2 && attributeSplit.indexOf('color') === -1) {
```
ngokevin: Note that `material.color` is only a single test case. Properties can be named anything and still be of type color. I'll try to help find a solution.

JohnRodney: are you saying something with a color property could be passed that did not contain the word color in it?

ngokevin: Yup. Anyone can create components with whatever schema they want. So I could do:

```
schema: {
  body: {type: 'color'}
}
```

JohnRodney: Ok so exact matching the scheme type should be safe here and then the rest should still work as far as the tweets are concerned.  I'm using the passed attributed to set when there are more than one in attributeSplit

ngokevin: Ah, that's interesting. Seems like that might work. It's not ideal we need to special-case colors, but I'm feel okay with this.

JohnRodney: Elaborate on special case colors?  I don't mind getting the solution all the way done rather than leaving something for someone else.

ngokevin: As in we have to a special check for colors. Though I don't have a better simple solution right now.

dmarcos: We need to check the schema. If we have for example a phong shader you will have three properties: diffuse, ambient, specular that are all colors.

JohnRodney: Ok I have something like: 

```
function isColor() {
  if (el.mappings && attributeSplit.length === 1 && !el.components[attribute]){
    return el.components[el.mappings[attribute].split('.')[0]].schema[attribute].type === 'color';
  } else if(attributeSplit.length > 1 && el.components[attributeSplit[0]]) {
    return el.components[attributeSplit[0]].schema[attributeSplit[1]].type === 'color';
  }
  return false;
}
```

There has to be an easier way to access the schema?
The reason I had to do this was to access the correct schema on a single attribute 'color' I had to access the mappings because there was no color schema it was under 'material.color' as that is the default map.

ngokevin: If the attribute can be split, then you can use `el.components[attrSplit[0]].schema[attrSplit[1]].type === 'color'`.

Else you can let it fall thru the if/else statements with the `isNaN`.

JohnRodney: @ngokevin In this case `attribute = 'color'` so `attrSplit[1]  = undefined`.
The `mapping[attribute] = 'material.color'` so `mapping[attribute].split('.')[0]] = 'material'` aka the component name.

ngokevin: If the attribute is `color`, then the first clause will fail because there's nothing to split on. Thus you just move on to the `isNaN(dataTo)` and do color interpolation there.

If the attribute is `material.color`, then you know you can check the schema, right?

JohnRodney: You are right that will support both.  And make all the test cases for color animation currently commited still work.  

Then we would just have to figure out how to add `el.components.material.schema.color.type = 'color'` to the generated entity to test the cases of `material.color` or `material.specular`

ngokevin: You don't have to add anything. Just set the material component onto the entity.

```el.setAttribute('material', 'shader: flat; color: #FFF');```

Comment (JohnRodney) [inline src/core/a-animation.js]: Now that I look at it I had tried that method earlier but it breaks 4 tests because I'm guessing in certain test cases the mapping doesn't exist so its not safe to check for it as the tests will try to call a property on an undefined object.

Comment (ngokevin) [inline src/core/a-animation.js]: I'd ignore everything about mappings. That's part of what is currently not core code (e.g., `<a-box>`). Though it should be able to interpolate attributes that it detects are colors (e.g., `color="#FFFFFF"`, but that doesn't require looking up schema.

Comment (JohnRodney) [inline src/core/a-animation.js]: Alright took your suggestion and pushed another version up with the fallback to isNaN and a isColor function that checks to make sure schemas are defined so that older tests don't fail with this introduction.

File: src/core/a-animation.js [hunk 12]
```diff
@@ -373,6 +375,17 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
   };
 
   /**
+   * Match the sceme type to color
+   * @return {bool} if the schema is of type color
+   */
+  function isColor () {
+    if (el.components[attributeSplit[0]] && el.components[attributeSplit[0]].schema) {
```
dmarcos: You can clean this up a little bit:
````javascript
function isColor () {
  var componentName = attributeSplit[0];
  var propertyName = attributeSplit[1];
  var component = el.components[componentName];
  var schema = component && component.schema;
  if (schema) { return schema[propertyName].type === 'color' }
  return true;
}

File: src/core/a-animation.js [hunk 13]
```diff
@@ -373,6 +375,17 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
   };
 
   /**
+   * Match the sceme type to color
```
dmarcos: typo sceme -> schema

src/core/a-animation.js
```diff
@@ -356,13 +357,14 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
   var from = {};
   var partialSetAttribute;
   var to = {};
-
-  if (attributeSplit.length === 2) {
+  if (attributeSplit.length === 2 && !isColor()) {
```
dmarcos: It's better to be explicit that let it pass through:
````javascript
if (attributeSplit.length === 2) {
  if (isColor()) { getForColorComponent(); }
  else {  getForComponentAttribute() }
}
````

File: tests/core/a-animation.test.js [hunk 15]
```diff
@@ -40,6 +40,51 @@ function setupAnimation (animationAttrs, cb, elAttrs) {
 }
 
 /**
+ * This will generate a color animation test using passed format
+ *   of to and from in order to test rgb, hsl, and nouns.
+ *  @param {description} description of the test
+ *  @param {from} expects a color format of white
+ *  @param {to} expects a color format of black
+ */
+function generateColorAnimationTest (description, from, to) {
```
dmarcos: We need a test for the case:
````html
<a-animation attribute="material.color" from="red" to="blue"></a-animation>
````

**Comment by JohnRodney** [comment]
Updated with refactors as suggested and added the red to blue test case.  Should probably go through another refactor phase and then add an example and update the docs?

Will probably update a bit more tomorrow.

File: src/core/a-animation.js [hunk 16]
```diff
@@ -373,6 +379,19 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
...
   /**
+   * Match the scema type to color
+   * @return {bool} if the schema is of type color
+   */
+  function isColor () {
+    var componentName = attributeSplit[0];
+    var propertyName = attributeSplit[1];
+    var component = el.components[componentName];
+    var schema = component && component.schema;
+    if (schema) { return schema[propertyName].type === 'color'; }
```
dmarcos: we can simplify the return statment:
````javascript
return schema && schema[propertyName].type === 'color';
````

Comment (JohnRodney) [inline src/core/a-animation.js]: This throws an error when schema[propertyName] is undefined so it would have to be:

```
  return schema && schema[propertyName] && schema[propertyName].type === 'color';
```

Would protect from all. Not sure how often the schema is undefined I think its only when given bad input, but probably better than letting an error get thrown?

Comment (dmarcos) [inline src/core/a-animation.js]: Now that I think about this. We could animate a component that it's not defined on the element yet (before the animation runs like)
````html
<a-entity geometry="..."><a-animation attribute="material.color" from="red" to="blue"></a-animation></a-entity>
````
Is this case covered? I think you would not be able to check ```el.components``` to retrieve the schema.

Comment (ngokevin) [inline src/core/a-animation.js]: Worth testing, though I suspect it will set the `from` before starting?

Comment (JohnRodney) [inline src/core/a-animation.js]: As far as I have seen yes it sets the from as the default on the image even if a color hasn't been set.

Comment (JohnRodney) [inline src/core/a-animation.js]: @ngokevin Did you ever get a chance to check out the branch?  Also was wondering if going in this direction for the animation should maybe be rethought?  Having to add special cases for each animation type seems unintuitive.  Couldn't we do a more effective remapping of primitives to their object by passing a reference?

Just worried this branching if of detecting type is going to get uglier as we try to support different interpolation types.  I built out something similar in a repo where I was trying to essentially build out an aframe-like framework (before I heard of aframe) in meteor.  

Through this I had build out a transitioner object that would take a payload and tween any object with numeric values.  As long as you passes a string describing it such as `mesh.material.specular` and a ending object `{ r: 1, g: 0, b: .4 }`.

Here is a reference to the code and how I approached it.  Not sure if it would help https://github.com/poetic/poetic-vr-boilerplate/blob/master/utils-animate.js.

Comment (dmarcos) [inline src/core/a-animation.js]: @JohnRodney Yes, I think this PR is very ad-hoc for the color type. The animation code is one of the most confusing areas in aframe and any ideas to improve it are more than welcome. In your approach, Would you pass the variable you want interpolated and the end value? Do you infer the type from the end value?

Comment (ngokevin) [inline src/core/a-animation.js]: - tween.js can already tween any object with numeric values.
- Might want to refactor if it gets any more complicated than this, will re-evaluate. Are there any other special cases we need to consider? Colors might be the last big special case.

Code looks OK for this scope. Minimal changes, tests pass, things work :white_check_mark:

File: src/core/a-animation.js [hunk 17]
```diff
@@ -373,6 +379,19 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
   };
 
   /**
+   * Match the scema type to color
```
ngokevin: typo schema

src/core/a-animation.js
```diff
@@ -463,3 +498,34 @@ function strToBool (str) {
...
+ * @param {color}
+ * @returns {number}
+ */
+function convertToIntegerColor (color) {
+  return Math.floor(Math.min(Math.abs(color), 1) * 255);
+}
+
+/**
+ * Converts a rgb object into a hex string
+ * @param {color} object { r: 1, g: 1, b: 1 }
```
ngokevin: reversed jsdoc `@param {object} color - {r: 1, g: 1, b: 1}`

src/core/a-animation.js
JohnRodney: lol I've mistyped that one twice now.  Updating the commit.

src/core/a-animation.js
JohnRodney: @dmarcos Yea I approached it from the idea that it would tween values from your passed object and whatever it's value already was.  As for type checking it was a pure number tween so it would just diff the values calculate delta from last from and decide the amount to tween based on the current framerate.

@ngokevin Understood on tween.js.  I ended up writing that because it didn't take much code to accomplish tween js w/o the overhead of another library.  

I am still not positive this would even work on 'material.specular' as we discussed earlier.  I need to dig in a bit more to how to setup a color value for specular and then try to see if we can animate that as well.

src/core/a-animation.js
JohnRodney: resolved

ngokevin:
Tested, works on both ECS box and primitive-style box:

box

cube

(In that, I discovered an issue that is not related to this PR: https://github.com/aframevr/aframe/issues/1334)

ngokevin:
I think we're ready to squash, and @dmarcos to sign off.

JohnRodney:
Excellent did we want to update the docs in animation.md to explain color animation support with a short snippet?

ngokevin:
Yeah that'd be great. An `test-animations-color` example would be cool too.

JohnRodney:
I am guessing a `test-animations-color` is different from an animation example?  It looks like `test` example namespace folders are all ones that provide stats and seem to be attempting to put some burden on the system?

ngokevin:
We need to organize/rename the examples. But `test-` examples, to me, are examples that you check to make sure a particular feature works. They all have stats enabled on the scene at the moment so we can check mobile stats, but we could re-eval that.

JohnRodney:
Sounds good I'll follow the examples of the other test-examples and put some color on it.  I had been playing around with some pulse animations that domino event emissions. Just to see if too many animations on color would cause a problem.  Never really saw an issue, but I didn't add stats to the scene. 

out

cvan:
woo! this is awesome! :tada:

JohnRodney:
@ngokevin Sorry about the delay I got really sick last week.  I added the readme entry and a test-animate-color.  Once this is ready I can add the test we talked about to the other PR that ensures a from color is correctly retrieved when a animating a color and the from is not defined in the a-animation.

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,25 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Animate Color</title>
+    <meta name="description" content="Animate Color - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
+      <a-sky color="#afafaf"></a-sky>
+      <a-box color="white" position="2 1 -10">
```
ngokevin: Can you make a few of these `<a-entity>` syntax?

tests/core/a-animation.test.js
```diff
@@ -550,4 +606,55 @@ suite('a-animation', function () {
...
+  generateColorAnimationTest('accepts hex shorthand', '#fff', '#000');
+  generateColorAnimationTest('accepts nouns and rgb', 'rgb(255, 255, 255)', 'black');
+  generateColorAnimationTest('accepts hsl', 'hsl(1, 100%, 100%)', 'hsl(0, 0%, 0%)');
+  generateColorAnimationTest('accepts dot notation', 'white', 'black', 'material.color');
+
+  suite('component color animation: accepts dot notation', function () {
+    var attribute = 'material.color';
+    setup(function (done) {
+      var self = this;
+      var elAttrs;
```
ngokevin: Can you build this object in just one line?

```js
var elAttrs = { 
  material: {}
};
```

ngokevin:
Thanks, hope you're better!

dmarcos:
@JohnRodney Is there any work left here?

JohnRodney:
@dmarcos yea I just need to alter @ngokevin suggestions.  I'm digging into now.  Not sure if you guys seen the news but weather has been pretty crazy in Houston the last few days.  I'll be pushing an update up shortly.

dmarcos:
@JohnRodney Thanks! I hope the sun shines soon over there.

JohnRodney:
@dmarcos We all do here too haha.

@ngokevin I adjusted the requested changes.  The only thing that may be missing I can think of now is a link to the example on the index page that is on localhost:9000 when you start the server.  I'm guessing I should probably add this yes?

docs/core/animations.md
```diff
@@ -73,6 +73,16 @@ We can animate numeric attributes as well. For example, we can animate the inten
 </a-light>
 ```
 
+### Color Attributes
+
+We can animate color attributes as well. For example, we can animate a box from white to red.
+
+```html
+<a-entity id="blushing-cube" geometry="primitive: box">
+  <a-animation attribute="color" begin="white" to="red" dur="1000"></a-animation>
```
ngokevin: Should be `attribute="material.color"`

`begin="white"` should be `from="white"`

docs/core/animations.md
```diff
@@ -73,6 +73,16 @@ We can animate numeric attributes as well. For example, we can animate the inten
 </a-light>
 ```
 
+### Color Attributes
+
+We can animate color attributes as well. For example, we can animate a box from white to red.
```
ngokevin: I'd add `We can animate any component property that has a `color` type as well.`

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,28 @@
...
```
ngokevin: Is it possible to use the example you posted?

ngokevin: This one is really cool

src/core/a-animation.js
ngokevin: It works.

ngokevin:
OK, that should be the last of it!

examples/test-animate-color/index.html
JohnRodney: @ngokevin I used setTimeouts in that to stagger the animation emission of events.  If you guys are ok with that.  Some repo's are real against timeouts so I wasn't sure if I should use it as an example.

JohnRodney: If that is cool I will recreate that example again.

ngokevin: You can use the animation 'begin' attribute to delay animation starts

JohnRodney: Alright I'll look into that and get the working example to run around that method.  Getting caught up at work today too so probably look for this adjustment tomorrow.

JohnRodney:
@ngokevin alright I updated the demo to a little more fancy of a demo.  I also opened an issue because you can't use begin delays to stagger because they calculate the begin delay on a repeat meaning the "stagger" animation eventually gets out of sync.  

This can be updated to remove the intervals once the bug is fixed there.

ngokevin:
Awesome! I'll merge this once I get to the office.

JohnRodney:
Cool once you give it one more look over and are ready let me know and I'll squash the commits

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,153 @@
...
+    <meta name="description" content="Animate Color - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
+      <a-sky color="#000000">
+        <a-animation attribute="color" dur="10000" from="#afafaf" to="#000000" repeat="indefinite">
+      </a-sky>
+
+      <a-entity position="4.5 1 -3" class="pulse" geometry="primitive: box; depth: 0.5; height: 1; width: 0.5" material="color: #5698Af">
```
ngokevin: You can use mixins to clean up.

```
html
<a-assets>
  <a-mixin id="box" geometry="primitive: box..." material="..."></a-mixin>
  <a-mixin id="pulse" attribute="material.color" from="#fff" to="#5698Af" dur="1000" begin="pulse"></a-mixin>
</a-assets>

<a-entity mixin="box"><a-animation mixin="pulse"></a-animation></a-entity>
<a-entity mixin="box"><a-animation mixin="pulse"></a-animation></a-entity>
```

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,153 @@
+<!DOCTYPE html>
```
ngokevin: We organized the examples, so you have to move this to ```examples/test/animate-color```

And change to ```../../../dist/aframe.js```

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,153 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Animate Color</title>
+    <meta name="description" content="Animate Color - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
```
ngokevin: Can just do `<a-scene stats>`

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,153 @@
...
+      <a-entity position="-4.5 1 -3.5" class="pulse" geometry="primitive: box; depth: 0.5; height: 1; width: 0.5" material="color: #5698Af">
+        <a-animation attribute="material.color" from="#fff" to="#5698Af" dur="1000" begin="pulse"></a-animation>
+      </a-entity>
+
+      <a-plane rotation="-90 0 0" width="40" height="40" color="#7BC8A4"></a-plane>
+    </a-scene>
+  </body>
+
+  <script>
+    window.onload = function() {
```
ngokevin: Our linter isn't running on examples yet, but:

- Space after `function` (`function ()`)
- Space around operators (`length / 2`)

Might be better to wait for `scene` `loaded` event too.

examples/test-animate-color/index.html
```diff
@@ -0,0 +1,153 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Animate Color</title>
+    <meta name="description" content="Animate Color - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
+      <a-sky color="#000000">
```
ngokevin: It is `#000` here, but the `from` is different.

JohnRodney:
@ngokevin adjusted to mixins, new file structure and stats attribute.

examples/test/animate-color/index.html
```diff
@@ -0,0 +1,79 @@
...
+      <a-assets>
+        <a-mixin id="box" class="pulse" geometry="primitive: box; depth: 0.5; height: 1; width: 0.5" material="color: #5698Af">
+        <a-mixin id="pulse" attribute="material.color" from="#fff" to="#5698Af" dur="1000" begin="pulse"></a-mixin>
+      </a-assets>
+
+      <a-sky color="#000000">
+        <a-animation attribute="color" dur="10000" from="#afafaf" to="#000000" repeat="indefinite">
+      </a-sky>
+
+      <a-entity position="4.5 1 -3" class="pulse" mixin="box"><a-animation mixin="pulse"></a-animation></a-entity>
```
ngokevin: You don't need the class anymore because your `selectorAll` can be `[mixin="box"]`

JohnRodney: done

ngokevin:
Thanks so much!!! That was a long process, but everything works, and your example is pretty cool and running at 60fps  :)

PR #1309, opened 2016-03-31
stop karma tests from running twice on every change (fixes #1253)

### LINKED ISSUE #1253: Test suite runs twice on every change
**Description:** On every change (`npm run test`), the Karma runner is running twice.

- A-Frame Version: master
- Platform/Device: OS X Macbook Pro

ngokevin:
**Description:** https://github.com/nikku/karma-browserify/issues/67

**Changes proposed:**
- Wrap pattern in object.

### DIFF
--- a/tests/karma.conf.js
+++ b/tests/karma.conf.js
@@ -22,9 +22,9 @@ module.exports = function (config) {
     ],
     files: [
       // Define test files.
-      'tests/**/*.test.js',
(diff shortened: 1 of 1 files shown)


dmarcos:
how does this change the behavior?

ngokevin:
It's on a bug on karma-browserify with OS X.  Before, tests were running twice for me on every single change. Now they only run once as they should.

PR #1313, opened 2016-03-31
add material updateTick tests, some material code cleanup

ngokevin:
**Description:** Originally had timeDelta property type code, but removed to make the PR simpler.

**Changes proposed:**
- Add tests for material updateTick, shader registration
- Fix sqema/schema typo
- Some minor variable renames
- Inline functions for `forEach` and `map`

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -10,6 +10,7 @@ var diff = utils.diff;
 var registerComponent = component.registerComponent;
 var shaders = shader.shaders;
 var shaderNames = shader.shaderNames;
+var TYPE_TIME = 'time';
 
 /**
  * Material component.
@@ -27,6 +28,9 @@ module.exports.Component = registerComponent('material', {
     depthTest: { default: true }
   },
 
+  /**
+   * Init handler that can be overridden to provide a custom three.js material.
+   */
   init: function () {
     this.material = null;
   },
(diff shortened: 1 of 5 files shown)


src/core/shader.js
```diff
@@ -60,9 +60,7 @@ Shader.prototype = {
     var self = this;
     var variables = {};
     var schema = this.schema;
-    var squemaKeys = Object.keys(schema);
-    squemaKeys.forEach(processSquema);
-    function processSquema (key) {
```
dmarcos: I like having independent functions defined at the bottom. When you have multiple instructions makes the code more legible rather that having a foreign function embedded in the main code flow.

````javascript
var squemaKeys = Object.keys(schema);
squemaKeys.forEach(processSquema);
doThis();
doThat();
doMoreStuff();
function processSquema() { ... }
````

Comment (ngokevin) [inline src/core/shader.js]: IMO, adds indirection making me jump up and down to read the code

`s/sqema/schema`

Comment (dmarcos) [inline src/core/shader.js]: That indirection is good because usually you don't want to know the details. You just want to read first the function and get a general idea of what it does. If you want to know what processSquema does you can go and read it but otherwise is noise to get a general understanding.

Comment (ngokevin) [inline src/core/shader.js]: indirection != abstraction. in these cases we're not trying to abstract, the functions are part of the code.

File: src/components/material.js [hunk 2]
```diff
@@ -55,31 +62,46 @@ module.exports.Component = registerComponent('material', {
...
-    this.updateBehavior();
+    this.updateTick();
   },
 
-  updateBehavior: function () {
+  /**
+   * Update tick handler that continuously updates certain shader uniforms if the shader
+   * specifies certain uniform types like `time`.
+   */
+  updateTick: function () {
```
dmarcos: I have a different way to handle this on my shader toy PR. Instead of being hard coded the schema of the property defines an update function if it's meant to be updated on every frame.

dmarcos: I will handle this on that PR

Issue #633, opened 2016-01-13
Support buffer geometries


ngokevin:
<a href="https://github.com/ngokevin"><img src="https://avatars.githubusercontent.com/u/674727?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ngokevin](https://github.com/ngokevin)**
_Monday Nov 16, 2015 at 01:44 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/441_

----

buffer geometries are more efficient at the cost of being harder to modify.

```html
<vr-object geometry="primitive: sphere; buffer: true"></vr-object>
```

results in using THREE.SphereBufferGeometry. or just default to always use buffer geometries for now as we have no feature to modify vertices

ngokevin:
<a href="https://github.com/jcarpenter"><img src="https://avatars.githubusercontent.com/u/1425497?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jcarpenter](https://github.com/jcarpenter)**
_Wednesday Nov 18, 2015 at 10:53 GMT_

----

Yeah will be interesting to see if the perf wins are appreciable. I'm sure as we create stress test scenes (per #464) that this will be important. Can also get Jaume's input when he's in office this week.

ngokevin:
Default to buffer for perf. Allow:

`<a-entity geometry="buffer: false">`

PR #1317, opened 2016-03-31
registerGeometry API (fixes #1162)

### LINKED ISSUE #1162: registerGeometry API
Description: Be able to register new geometries into the standard geometry component.

Context: I wanted to be able to use a plane geometry instead of the plane buffer geometry that A-Frame uses with the plane primitive.

```js
AFRAME.registerGeometry('plane', {
  schema: {
    height: { default: 1 },
    width: { default: 1 }
  },

  update: function () {
    this.geometry = new THREE.PlaneGeometry(data.width, data.height);
  }
}
```

```html
geometry="type: plane; height: 1; width: 1"
```

ngokevin:
**Description:** `registerGeometry`, similar to `registerShader`. Cleans up the geometry component, allows us to define a different schema per geometry type, allows us to set different defaults per geometry type, opens up another door for extensibility and sharing.

Prerequisite to geometry caching system and automatically using buffer geometries.

- [ ] docs

**Changes proposed:**
- registerGeometry API
- move geometries to individual files
- normalize all geometries to non-buffer to prepare for https://github.com/aframevr/aframe/issues/633

### DIFF
--- a/src/core/geometry.js
+++ b/src/core/geometry.js
@@ -1,0 +1,72 @@
+var schema = require('./schema');
+
+var processSchema = schema.process;
+var geometries = module.exports.geometries = {};  // Registered geometries.
+var geometryNames = module.exports.geometryNames = [];  // Names of registered geometries.
+var THREE = require('../lib/three');
+
+/**
+ * Geometry class definition.
+ *
+ * Geometries extend the geometry component API to create and register geometry types.
+ */
+var Geometry = module.exports.Geometry = function () {};
+
+Geometry.prototype = {
+  /**
+   * Contains the type schema and defaults for the data values.
+   * Data is coerced into the types of the values of the defaults.
+   */
+  schema: {},
+
+  /**
+   * Init handler. Similar to attachedCallback.
+   * Called during shader initialization and is only run once.
+   */
+  init: function (data) {
+    this.geometry = new THREE.Geometry();
+    return this.geometry;
+  },
+
+  /**
+   * Update handler. Similar to attributeChangedCallback.
+   * Called whenever the associated geometry data changes.
+   *
+   * @param {object} data - New geometry data.
+   */
+  update: function (data) { /* no-op */ }
+};
+
+/**
+ * Registers a geometry to A-Frame.
+ *
+ * @param {string} name - Geometry name.
+ * @param {object} definition - Geometry property and methods.
+ * @returns {object} Geometry.
+ */
+module.exports.registerGeometry = function (name, definition) {
+  var NewGeometry;
+  var proto = {};
+
+  // Format definition object to prototype object.
+  Object.keys(definition).forEach(function expandDefinition (key) {
+    proto[key] = {
+      value: definition[key],
+      writable: true
+    };
+  });
+
+  if (geometries[name]) {
+    throw new Error('The geometry `' + name + '` has been already registered');
+  }
+  NewGeometry = function () { Geometry.call(this); };
+  NewGeometry.prototype = Object.create(Geometry.prototype, proto);
+  NewGeometry.prototype.name = name;
+  NewGeometry.prototype.constructor = NewGeometry;
+  geometries[name] = {
+    Geometry: NewGeometry,
+    schema: processSchema(NewGeometry.prototype.schema)
+  };
+  geometryNames.push(name);
+  return NewGeometry;
+};
--- a/src/geometries/circle.js
+++ b/src/geometries/circle.js
@@ -1,0 +1,18 @@
+var registerGeometry = require('../core/geometry').registerGeometry;
+var THREE = require('../lib/three');
+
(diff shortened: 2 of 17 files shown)


ngokevin:
ready for r?, going to be documenting in the meantime.

src/geometries/cone.js
```diff
@@ -0,0 +1,24 @@
...
+registerGeometry('cone', {
+  schema: {
+    height: {default: 1, min: 0},
+    openEnded: {default: false},
+    radiusBottom: {default: 1, min: 0},
+    radiusTop: {default: 0.8, min: 0},
+    segmentsHeight: {default: 18, min: 0, type: 'int'},
+    segmentsRadial: {default: 36, min: 0, type: 'int'},
+    thetaLength: {default: 360, min: 0},
+    thetaStart: {default: 0}
```
dmarcos: missing ```min: 0```?

src/geometries/circle.js
```diff
@@ -0,0 +1,18 @@
...
+var THREE = require('../lib/three');
+
+var degToRad = THREE.Math.degToRad;
+
+registerGeometry('circle', {
+  schema: {
+    radius: {default: 1, min: 0},
+    segments: {default: 32, min: 0, type: 'int'},
+    thetaLength: {default: 360, min: 0},
+    thetaStart: {default: 0}
```
dmarcos: missing ```min: 0```?

src/geometries/cylinder.js
```diff
@@ -0,0 +1,22 @@
...
+
+registerGeometry('cylinder', {
+  schema: {
+    height: {default: 1, min: 0},
+    openEnded: {default: false},
+    radius: {default: 1, min: 0},
+    segmentsHeight: {default: 18, min: 0, type: 'int'},
+    segmentsRadial: {default: 36, min: 0, type: 'int'},
+    thetaLength: {default: 360, min: 0},
+    thetaStart: {default: 0}
```
dmarcos: missing ```min: 0```

src/geometries/sphere.js
```diff
@@ -0,0 +1,23 @@
+var registerGeometry = require('../core/geometry').registerGeometry;
+var THREE = require('../lib/three');
+
+var degToRad = THREE.Math.degToRad;
+
+registerGeometry('sphere', {
+  schema: {
+    radius: {default: 1, min: 0},
+    segments: {default: 32, min: 0, type: 'int'},
+    phiLength: {default: 360},
```
dmarcos: missing ```min: 0```?

src/geometries/cone.js
ngokevin: I think negative thetaStart is valid.

src/geometries/torusKnot.js
```diff
@@ -0,0 +1,19 @@
+var registerGeometry = require('../core/geometry').registerGeometry;
+var THREE = require('../lib/three');
+
+registerGeometry('torusKnot', {
+  schema: {
+    p: {default: 2, type: 'int'},
```
dmarcos: Can p, q be negative?

src/core/geometry.js
```diff
@@ -0,0 +1,72 @@
...
+ * @param {string} name - Geometry name.
+ * @param {object} definition - Geometry property and methods.
+ * @returns {object} Geometry.
+ */
+module.exports.registerGeometry = function (name, definition) {
+  var NewGeometry;
+  var proto = {};
+
+  // Format definition object to prototype object.
+  Object.keys(definition).forEach(function (key) {
```
dmarcos: name function

src/geometries/torusKnot.js
ngokevin: Yeah, works for me.

src/geometries/plane.js
```diff
@@ -0,0 +1,13 @@
...
+var THREE = require('../lib/three');
+
+registerGeometry('plane', {
+  schema: {
+    height: {default: 1, min: 0},
+    width: {default: 1, min: 0}
+  },
+
+  init: function (data) {
+    this.geometry = new THREE.PlaneGeometry(data.width, data.height);
```
dmarcos: Why are we dropping buffer geometries?

http://threejs.org/docs/#Reference/Core/BufferGeometry

ngokevin: I am going to make every geometry a buffer geometry by default, then you specify `buffer: false` if you want otherwise. But that's done with `THREE.BufferGeometry().fromGeometry( geometry, settings );` so we can't hardcode PlaneBufferGeometry here.

ngokevin: It will be temporary: https://github.com/aframevr/aframe/issues/633

I can limit these changes to another PR if you want.

dmarcos: let's make buffer as a default and add the flag to control it in a different PR

ngokevin:
updated

Issue #1310, opened 2016-03-31
Reduce `<a-sky>` default radius


ngokevin:
**Description:** The default radius is currently 5000, which is huge (5000 meters to scale). A single sky alone creates 8K faces and 24K vertices. Maybe a better default would be 50 to 100. 

- A-Frame Version: master

JohnRodney:
This is also a good opportunity to reduce the number of faces and vertices by reducing the number of height segments.  The lower the number of height segments the less distortion you get on an equatorial projected image as the src attribute.

64 Height segments: 
<img width="725" alt="screen shot 2016-03-31 at 12 26 42 am" src="https://cloud.githubusercontent.com/assets/11075982/14166243/5e2975de-f6d7-11e5-8c91-249ada2ee9cf.png">

10 Height segments:
<img width="593" alt="screen shot 2016-03-31 at 12 24 34 am" src="https://cloud.githubusercontent.com/assets/11075982/14166251/7109e198-f6d7-11e5-96d7-1a12fdeb3954.png">

PR #1319, opened 2016-04-01
reduces the radius of sky component's default in order to increase pe…

JohnRodney:
**Description:**
  Reduces the number of default vertices used in the a-sky component. In reference to issue #1310 
**Changes proposed:**
-  Reduce the radius 
-  Reduce the heightSegments for equatorial projection distortion reduction.


ngokevin:
Thanks!

Issue #1311, opened 2016-03-31
remove stringifying to DOM in most cases, have getAttribute access a cache rather than from DOM


ngokevin:
**Description:** Stringification to DOM, and using the DOM to fetch data is one of the bottlenecks in perf.

We only want to stringify to DOM if the DOM inspector is open (still want to debounce that) or a flag is set. This may affect integration with other libraries, in which case the flag can be turned on.

bnolan:
How are we testing performance gains here? Just against the performance-test scene? Be interesting to get some before / after metrics. Also - please make sure the flag can be turned on, because the editor needs it. ;)

dmarcos:
@bnolan look at https://github.com/aframevr/aframe/pull/1323

Issue #1163, opened 2016-03-13
mixins + materials with different shader do not work


ngokevin:
Description: Complains about shader being undefined. http://codepen.io/team/mozvr/pen/jqMGEq?editors=100 or if I check out the most recent dist, there is no error, but it doesn't work

- A-Frame Version: master
- Platform/Device: All

dmarcos:
This PR fixes the issue https://github.com/aframevr/aframe/pull/1159

ngokevin:
This is with your PR, I checked it out.

ngokevin:
One issue is single-property shader schemas. Need to:

- Allow schemaless shader (e.g., ```schema: {}```) if it doesn't take any properties. It is being treated as single-prop schema.

ngokevin:
If shader does not have schema ({} or empty), it will be treated as single-prop schema. Then I was having issues where `material="shader: flat"` was getting parsed to a string rather than object. Then it was unable to find the shader.

PR #1323, opened 2016-04-03
Refactors component updates to minimize the number of string parsings…

dmarcos:
Attributes updates is the most common operation in a-frame scenes. I've been looking for performance bottlenecks on setAttribute/getAttribute calls. Using the DOM to store the current state of the component incurs in a lot of component data stringifications (to feed the DOM attribute via HTMLElement.prototype.setAttribute) and parse calls to recalculate the component data from the attribute string or when getAttribute is called by the application logic. By caching the parsed attribute corresponding to the component we can save a lot of stringify/parse calls. The overhead is specially noticeable on scenes with a lot of dynamic behavior. I've added a stress test to illustrate the problem and show the performance gains. In the ```stress-animation``` demo I create a lot of elements with animations that will call ```setAttribute``` continuously. On my machine and without the patch the example runs in the lows-mids 20s FPS. With this patch applied it goes up to the lows-mids 30s FPS. That's approximately a 50% perf gain.

### DIFF
--- a/examples/stress-animation/index.html
+++ b/examples/stress-animation/index.html
@@ -1,0 +1,77 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Animation</title>
+    <meta name="description" content="Animation - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
+      <a-assets>
+        <a-mixin id="ball" geometry="primitive: sphere; radius:1"  material="shader: flat; color: red">
+        </a-mixin>
+        <a-mixin id="opacity" attribute="material.opacity" dur="2000" direction="alternate" fill="forwards"
+                  ease="linear" repeat="indefinite" from="0.3" to="1.0"><a-mixin>
+        <a-mixin id="spin" attribute="rotation" dur="4000"
+                 repeat="indefinite" easing="linear" dur="3000" to="0 0 360"><a-mixin>
+      </a-assets>
+
+      <a-entity position="0 0 80">
+        <a-entity camera look-controls wasd-controls></a-entity>
+      </a-entity>
+
+      <a-entity id="fans"></a-entity>
+      
+    </a-scene>
+  </body>
+  <script>
+    var i;
+    var fansEl = document.querySelector('#fans');
+    var numRows = 4;
+    var numFans = 3
+    var numFanRows = 3;
+    var numBalls = 11;
+    var ball =
+      '<a-entity>' +
+        '<a-entity mixin="ball">' +
+          '<a-animation mixin="opacity"></a-animation>' +
+        '</a-entity>' +
+      '</a-entity>';
+    var ballEl;
+    var rowEl;
+    var ballInitX = -15;
+    var fanInitX = -60;
+    var fanInitY = -40;
+    var offset;
+
+    for (var n = 0; n < numFanRows; ++n) {
+      var fanRow = document.createElement('a-entity');
+      fanRow.setAttribute('position', { x: 0, y: fanInitY - (n * fanInitY), z: 0 });
+      for (var k = 0; k < numFans; ++k) {
+        var fanEl = document.createElement('a-entity');
+        fanEl.setAttribute('position', { x: fanInitX - (k * fanInitX), y: 0, z: 0});
+        var animationEl = document.createElement('a-animation');
+        animationEl.setAttribute('mixin', 'spin');
+        fanEl.appendChild(animationEl);
+        for (var i = 0; i < numRows; ++i) {
+          var rowEl = document.createElement('a-entity');
+          rowEl.setAttribute('rotation', { x: 0, y: 0, z: i * (180 / numRows) });
+          for (var j = 0; j < numBalls; ++j) {
+            offset = ballInitX + 3 * j;
+            if (offset === 0) { continue; }
+            var ballEl = document.createElement('a-entity');
+            ballEl.setAttribute('position', { x: ballInitX + 3 * j, y: 0, z: 0 });
+            ballEl.innerHTML = ball;
+            animationEl = ballEl.querySelector('a-animation');
+            animationEl.setAttribute('begin', Math.random() * 2000);
+            rowEl.appendChild(ballEl)
+          }
+          fanEl.appendChild(rowEl);
+        }
+        fanRow.appendChild(fanEl);
+      }
+      fansEl.appendChild(fanRow);
+    }
+  </script>
+</html>
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -51,7 +51,7 @@ module.exports.Component = registerComponent('camera', {
     // If `active` property changes, or first update, handle active camera with system.
     if (data.active && system.activeCameraEl !== this.el) {
       // Camera enabled. Set camera to this camera.
-      system.setActiveCamera(el, camera);
+      system.setActiveCamera(el);
     } else if (!data.active && system.activeCameraEl === this.el) {
       // Camera disabled. Set camera to another camera.
       system.disableActiveCamera();
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -9,7 +9,6 @@ var AEntity;
 var isNode = re.isNode;
 var debug = utils.debug('core:a-entity:debug');
 var registerElement = re.registerElement;
-var styleParser = utils.styleParser;
 
 /**
  * Entity is a container object that components are plugged into to comprise everything in
@@ -49,6 +48,13 @@ var proto = Object.create(ANode.prototype, {
    */
   attributeChangedCallback: {
     value: function (attr, oldVal, newVal) {
+      var component = this.components[attr];
+      // If the empty string is passed by the component initialization
+      // logic we ignore the component update.
+      if (component && component.justInitialized && newVal === '') {
+        delete component.justInitialized;
+        return;
+      }
       this.setEntityAttribute(attr, oldVal, newVal);
     }
   },
@@ -58,7 +64,7 @@ var proto = Object.create(ANode.prototype, {
       this.addToParent();
       if (!this.isScene) {
         this.load();
-        if (!this.parentNode.paused) { this.play(); }
+        if (this.parentEl.isPlaying) { this.play(); }
       }
     }
   },
@@ -78,11 +84,14 @@ var proto = Object.create(ANode.prototype, {
 
   applyMixin: {
     value: function (attr) {
-      var attrValue = this.getAttribute(attr);
+      var attrValue;
       if (!attr) {
         this.updateComponents();
         return;
       }
+      attrValue = this.getAttribute(attr);
+      // Make absence of attribute for getAttribute return undefined rather than null
+      attrValue = attrValue === null ? undefined : attrValue;
       this.updateComponent(attr, attrValue);
     }
   },
@@ -203,10 +212,8 @@ var proto = Object.create(ANode.prototype, {
   load: {
     value: function () {
       if (this.hasLoaded) { return; }
-
       // Attach to parent object3D.
       this.addToParent();
-
       if (this.isScene) {
         ANode.prototype.load.call(this, this.updateComponents.bind(this));
       } else {
@@ -244,13 +251,17 @@ var proto = Object.create(ANode.prototype, {
 
   /**
    * Initialize component.
+   *
+   * @param {string} name - Component name.
+   * @param {object} data - Component data
+   * @param {boolean} isDependency - True if the component is a dependency.
    */
   initComponent: {
-    value: function (name, isDependency) {
+    value: function (name, data, isDependency) {
       var component;
-      var isComponentDefined = checkComponentDefined(this, name);
+      var isComponentDefined = checkComponentDefined(this, name) || data !== undefined;
 
-      // Check if component is registered and whether component should be iniitalized.
+      // Check if component is registered and whether component should be initialized.
       if (!components[name] || (!isComponentDefined && !isDependency)) {
         return;
       }
@@ -258,21 +269,20 @@ var proto = Object.create(ANode.prototype, {
       // Initialize dependencies.
       this.initComponentDependencies(name);
 
-      if (isDependency && !isComponentDefined) {
-        // Add component if it is a dependency and not yet defined.
-        this.setAttribute(name, '');
-      } else {
-        if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
-        }
+      // Check if component already initialized.
+      if (name in this.components) { return; }
+      component = this.components[name] = new components[name].Component(this, data);
+      if (this.isPlaying) { playComponent(component, this.sceneEl); }
 
-        // Check if component already initialized.
-        if (name in this.components) { return; }
-
-        component = this.components[name] = new components[name].Component(this);
-        if (this.isPlaying) { playComponent(component, this.sceneEl); }
+      // Components are reflected in the DOM as attributes but the state is not shown
+      // hence we set the attribute to empty string.
+      // The flag justInitialized is for attributeChangedCallback to not overwrite
+      // the component with the empty string.
+      if (!this.hasAttribute(name)) {
+        component.justInitialized = true;
+        HTMLElement.prototype.setAttribute.call(this, name, '');
       }
+
       debug('Component initialized: %s', name);
     },
     writable: window.debug
@@ -287,7 +297,7 @@ var proto = Object.create(ANode.prototype, {
       dependencies = components[name].dependencies;
       if (!dependencies) { return; }
       dependencies.forEach(function (component) {
-        self.initComponent(component, true);
+        self.initComponent(component, undefined, true);
       });
     }
   },
@@ -295,20 +305,51 @@ var proto = Object.create(ANode.prototype, {
   removeComponent: {
     value: function (name) {
       var component = this.components[name];
+      var isDefault = name in this.defaultComponents;
+      var isMixedIn = isComponentMixedIn(name, this.mixinEls);
+      // Don't remove default or mixed in components
+      if (isDefault || isMixedIn) { return; }
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
     }
   },
 
+  /**
+   * Updates all the entity's components. Given by defaults, mixins and attributes
+   * Default components update before the rest.
+   */
   updateComponents: {
     value: function () {
+      var elComponents = {};
       var self = this;
-      var allComponents = Object.keys(components);
-      allComponents.forEach(updateComponent);
+      var i;
+      if (!this.hasLoaded) { return; }
+      // Components defined on the entity element
+      var attributes = this.attributes;
+      for (i = 0; i < attributes.length; ++i) {
+        addComponent(attributes[i].name);
+      }
+      // Components defined as mixins
+      getMixedInComponents(this).forEach(addComponent);
+      // Updates default components first
+      Object.keys(this.defaultComponents).forEach(updateComponent);
+      // Updates the rest of the components
+      Object.keys(elComponents).forEach(updateComponent);
+
+      // add component to the list
+      function addComponent (key) {
+        if (!components[key]) { return; }
+        elComponents[key] = true;
+      }
+      // updates a component with a given name
       function updateComponent (name) {
-        var elValue = self.getAttribute(name);
-        self.updateComponent(name, elValue);
+        var attrValue = self.getAttribute(name);
+        delete elComponents[name];
+        // turn null into undefined because getAttribute
+        // returns null in the absence of an attribute
+        attrValue = attrValue === null ? undefined : attrValue;
+        self.updateComponent(name, attrValue);
       }
     }
   },
@@ -318,30 +359,42 @@ var proto = Object.create(ANode.prototype, {
    *
    * When initializing, we set the component on `this.components`.
    *
-   * @param {string} name - Component name.
-   * @param {object} newData - The new properties assigned to the component
+   * @param {string} attr - Component name.
+   * @param {object} attrValue - The value of the DOM attribute.
    */
   updateComponent: {
-    value: function (name, newData) {
-      var component = this.components[name];
-      var isDefault = name in this.defaultComponents;
-      var isMixedIn = isComponentMixedIn(name, this.mixinEls);
+    value: function (attr, attrValue) {
+      var component = this.components[attr];
       if (component) {
-        // Attribute was removed, remove component if:
-        // 1. If component not defined in the defaults/mixins/attribute.
-        // 2. If new data is null, then not a default component and component is not defined
-        //    via mixins
-        if (!checkComponentDefined(this, name) ||
-            newData === null && !isDefault && !isMixedIn) {
-          this.removeComponent(name);
+        if (attrValue === null) {
+          this.removeComponent(attr);
           return;
         }
         // Component already initialized. Update component.
-        component.updateProperties(newData);
+        component.updateProperties(attrValue);
         return;
       }
       // Component not yet initialized. Initialize component.
-      this.initComponent(name);
+      this.initComponent(attr, attrValue);
+    }
+  },
+
+  /**
+   * Updates one property of the component
+   *
+   * @param {string} name - Component name
+   * @param {string} property - Component property name
+   * @param {any} propertyValue - New property value
+   */
+  updateComponentProperty: {
+    value: function (name, property, propertyValue) {
+      var component = this.components[name];
+      // Cached attribute value
+      var attrValue = component && component.attrValue;
+      // Copy cached value
+      var componentObj = attrValue ? utils.extend({}, attrValue) : {};
+      componentObj[property] = propertyValue;
+      this.updateComponent(name, componentObj);
     }
   },
 
@@ -356,8 +409,9 @@ var proto = Object.create(ANode.prototype, {
       var component = components[attr];
       if (component) {
         this.setEntityAttribute(attr, undefined, null);
+      } else {
+        HTMLElement.prototype.removeAttribute.call(this, attr);
       }
-      HTMLElement.prototype.removeAttribute.call(this, attr);
     }
   },
 
@@ -372,7 +426,7 @@ var proto = Object.create(ANode.prototype, {
       var sceneEl = this.sceneEl;
 
       // Already playing.
-      if (this.isPlaying) { return; }
+      if (this.isPlaying || !this.hasLoaded) { return; }
       this.isPlaying = true;
 
       // Wake up all components.
@@ -427,18 +481,22 @@ var proto = Object.create(ANode.prototype, {
    */
   setEntityAttribute: {
     value: function (attr, oldVal, newVal) {
-      var component = components[attr];
-      oldVal = oldVal || this.getAttribute(attr);
-      // When creating entities programatically and setting attributes, it is not part
-      // of the scene until it is inserted into the DOM. This does not apply to scenes as
-      // scenes depend on its child entities to load.
-      if (!this.hasLoaded && !this.isScene) { return; }
-      if (attr === 'mixin') {
-        this.updateStateMixins(newVal, oldVal);
-        this.updateComponents();
+      if (components[attr]) {
+        this.updateComponent(attr, newVal);
         return;
       }
-      if (component) { this.updateComponent(attr, newVal); }
+      if (attr === 'mixin') {
+        this.mixinUpdate(newVal, oldVal);
+        return;
+      }
+    }
+  },
+
+  mixinUpdate: {
+    value: function (newMixins, oldMixins) {
+      oldMixins = oldMixins || this.getAttribute('mixin');
+      this.updateStateMixins(newMixins, oldMixins);
+      this.updateComponents();
     }
   },
 
@@ -461,37 +519,51 @@ var proto = Object.create(ANode.prototype, {
    */
   setAttribute: {
     value: function (attr, value, componentPropValue) {
-      var self = this;
-      var component = this.components[attr] || components[attr];
-      var partialComponentData;
-      value = value === undefined ? '' : value;
-      var componentObj = value;  // Deserialized value to send to the component.
-      var componentStr = value;  // Serialized value to send to the DOM.
-      var oldValue;
-
-      if (component) {
+      var isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
+      if (components[attr]) {
+        // Just update one of the component properties
         if (typeof value === 'string' && componentPropValue !== undefined) {
-          // Update currently-defined component data with the new property value.
-          // Use native setAttribute in order not to double-parse properties.
-          partialComponentData = styleParser.parse(
-            HTMLElement.prototype.getAttribute.call(this, attr)) || {};
-          partialComponentData[value] = componentPropValue;
-          componentObj = partialComponentData;
+          this.updateComponentProperty(attr, value, componentPropValue);
+        } else {
+          this.updateComponent(attr, value);
         }
-        componentStr = component.stringify(componentObj);
+        // On debug mode we write the component state to the DOM attributes
+        if (isDebugMode) { this.components[attr].flushToDOM(); }
+        return;
       }
-      oldValue = this.getAttribute(attr);
-      ANode.prototype.setAttribute.call(self, attr, componentStr);
-      self.setEntityAttribute(attr, oldValue, componentObj);
+
+      ANode.prototype.setAttribute.call(this, attr, value);
+      if (attr === 'mixin') { this.mixinUpdate(value); }
     },
     writable: window.debug
   },
 
   /**
-   * If `attr` is a component, returns JUST the component data specified in the HTML
-   * by parsing the style-like string into an object. Like a partial version of
-   * `getComputedAttribute` as returned component data does not include applied mixins or
-   * defaults.
+   * To make the DOM attributes reflect the state of the components.
+   *
+   * @param {bool} recursive - Call updateDOM on the children
+   **/
+  flushToDOM: {
+    value: function (recursive) {
+      var components = this.components;
+      var children = this.children;
+      var child;
+      var i;
+      Object.keys(components).forEach(updateDOMAtrribute);
+      if (!recursive) { return; }
+      for (i = 0; i < children.length; ++i) {
+        child = children[i];
+        if (!child.flushToDOM) { continue; }
+        child.flushToDOM(recursive);
+      }
+      function updateDOMAtrribute (name) { components[name].flushToDOM(); }
+    }
+  },
+
+  /**
+   * If `attr` is a component, returns JUST the component data defined on the entity.
+   * Like a partial version of `getComputedAttribute` as returned component data
+   * does not include applied mixins or defaults.
    *
    * If `attr` is not a component, fall back to HTML getAttribute.
    *
@@ -500,10 +572,12 @@ var proto = Object.create(ANode.prototype, {
    */
   getAttribute: {
     value: function (attr) {
-      var component = this.components[attr] || components[attr];
-      var value = HTMLElement.prototype.getAttribute.call(this, attr);
-      if (!component || typeof value !== 'string') { return value; }
-      return component.parse(value, true);
+      var component = this.components[attr];
+      // If there's a cached value we just return it
+      if (component && component.attrValue !== undefined) {
+        return component.attrValue;
+      }
+      return HTMLElement.prototype.getAttribute.call(this, attr);
     },
     writable: window.debug
   },
@@ -568,9 +642,20 @@ function checkComponentDefined (el, name) {
   if (el.defaultComponents[name] !== undefined) { return true; }
 
   // Check if element contains the component.
-  if (el.hasAttribute(name)) { return true; }
+  if (el.components[name] && el.components[name].attrValue) { return true; }
 
   return isComponentMixedIn(name, el.mixinEls);
+}
+
+function getMixedInComponents (entityEl) {
+  var components = [];
+  entityEl.mixinEls.forEach(function getMixedComponents (mixinEl) {
+    Object.keys(mixinEl.componentAttrCache).forEach(addComponent);
+    function addComponent (key) {
+      components.push(key);
+    }
+  });
+  return components;
 }
 
 /**
@@ -582,7 +667,6 @@ function checkComponentDefined (el, name) {
 function isComponentMixedIn (name, mixinEls) {
   var i;
   var inMixin = false;
-
   for (i = 0; i < mixinEls.length; ++i) {
     inMixin = mixinEls[i].hasAttribute(name);
     if (inMixin) { break; }
--- a/src/core/a-mixin.js
+++ b/src/core/a-mixin.js
@@ -1,14 +1,30 @@
 /* global HTMLElement */
-var AComponents = require('./component').components;
 var ANode = require('./a-node');
 var registerElement = require('./a-register-element').registerElement;
+var components = require('./component').components;
 
+/**
+ * @member {object} componentAttrCache - Cache of pre parsed component attributes
+ */
 module.exports = registerElement('a-mixin', {
   prototype: Object.create(
     ANode.prototype,
     {
+      createdCallback: {
+        value: function () {
+          this.componentAttrCache = {};
+        }
+      },
+
+      attributeChangedCallback: {
+        value: function (attr, oldVal, newVal) {
+          this.cacheAttribute(attr, newVal);
+        }
+      },
+
       attachedCallback: {
         value: function () {
+          this.cacheAttributes();
           this.load();
         },
         writable: window.debug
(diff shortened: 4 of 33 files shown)


cvan:
can you add the issue number (#1311) to the commit message?

dmarcos:
This is the final design this patch has converged to:
1. We don't update the DOM anymore to reflect the state of a component.
2. The attribute value is cached in a parsed form within each component.
3. There's an ```updateDOM``` in the entity and component to write back the state to the DOM with the cached value. This can be used by applications that rely on the DOM like our sceneVR or our future editor.
4. When a component is initialized we write the attribute to the DOM without the property values. You will see something like this: ```<a-entity position rotation scale camera wasd-controls look-controls></a-entity>```. One can still query for an entity with a specific component sceneEl.querySelector('[camera]') and standard methods ```el.hasAttribute``` and ```el.attributes``` remain functional.

dmarcos:
The final gains for this patch on the included ```stress-animation``` demo are pretty significant. The example has 369 animations affecting 369 different entities. That translates on 369 at least ```setAttribute``` calls per frame. On my machine and without the patch the demo runs at low 20s FPS. With this patch applied it hits high 30s FPS. That's almost twice as fast.

dmarcos:
Things to look for when reviewing this patch. I like the perf benefits but I'm worried about the form. The complexity of the code to handle all the different cases is phenomenal. It's brittle and very difficult to understand. I cannot keep the complete flow in my head. This patch has been an effort of massaging the code until it covered all the edge cases. This work should have not been possible without the unit tests we have. There should be a simpler more concise way to express the same logic. I will keep thinking about ways to simplify. Bringing more eyes to this patch will help too.

ngokevin:
- [x] Examples work

src/core/a-entity.js
```diff
@@ -246,33 +248,32 @@ var proto = Object.create(ANode.prototype, {
...
-          // For scene default components, expose them in the DOM.
-          HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
-        }
-
-        // Check if component already initialized.
-        if (name in this.components) { return; }
-
-        component = this.components[name] = new components[name].Component(this);
-        if (this.isPlaying) { playComponent(component, this.sceneEl); }
+      // if (this.isScene && !this.hasAttribute(name) && name in this.defaultComponents) {
```
ngokevin: Can remove.

src/components/camera.js
```diff
@@ -19,13 +19,15 @@ module.exports.Component = registerComponent('camera', {
    */
   init: function () {
     var camera = this.camera = new THREE.PerspectiveCamera();
+    this.system.addCamera(this.el);
```
ngokevin: ```addCameraEl``` to be clear between camera entity vs. camera object.

src/core/a-entity.js
```diff
@@ -49,6 +48,11 @@ var proto = Object.create(ANode.prototype, {
    */
   attributeChangedCallback: {
     value: function (attr, oldVal, newVal) {
+      var component = this.components[attr];
+      if (component && component.justInitialized && newVal === '') {
```
ngokevin: Needs a short comment.

src/core/a-entity.js
```diff
@@ -246,33 +248,32 @@ var proto = Object.create(ANode.prototype, {
...
    */
   initComponent: {
-    value: function (name, isDependency) {
+    value: function (name, isDependency, attr) {
       var component;
-      var isComponentDefined = checkComponentDefined(this, name);
+      var isComponentDefined = checkComponentDefined(this, name) || attr !== undefined;
 
-      // Check if component is registered and whether component should be iniitalized.
+      // Check if component is registered and whether component should be initalized.
```
ngokevin: Still a typo: `initialized`.

src/core/a-entity.js
```diff
@@ -246,33 +248,32 @@ var proto = Object.create(ANode.prototype, {
    * Initialize component.
```
ngokevin: Update the docstring to explain the parameters like `attr`.

src/core/a-entity.js
```diff
@@ -246,33 +248,32 @@ var proto = Object.create(ANode.prototype, {
...
+      // For scene default components, expose them in the DOM.
+      //  HTMLElement.prototype.setAttribute.call(this, name, this.defaultComponents[name]);
+      // }
+      // Check if component already initialized.
+      if (name in this.components) { return; }
+      component = this.components[name] = new components[name].Component(this, attr);
+      if (this.isPlaying) { playComponent(component, this.sceneEl); }
+
+      if (!this.hasAttribute(name)) {
+        component.justInitialized = true;
```
ngokevin: Comment what `justInitialized` means and what it will be used for.

src/core/a-entity.js
```diff
@@ -301,14 +302,52 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  getMixedInComponents: {
```
ngokevin: This could be a private helper function ```getMixedInComponents (mixinEls)```

src/core/a-entity.js
```diff
@@ -301,14 +302,52 @@ var proto = Object.create(ANode.prototype, {
...
+    value: function () {
+      var mixinEls = this.mixinEls;
+      var i;
+      var components = [];
+      for (i = 0; i < mixinEls.length; ++i) {
+        var keys = Object.keys(mixinEls[i].attrs);
```
ngokevin: Don't need the intermediate var. Object.keys.forEach is a pretty common pattern.

ngokevin: You could reduce a lot of this function to a `map`.

src/core/a-entity.js
```diff
@@ -301,14 +302,52 @@ var proto = Object.create(ANode.prototype, {
...
-      var allComponents = Object.keys(components);
-      allComponents.forEach(updateComponent);
-      function updateComponent (name) {
-        var elValue = self.getAttribute(name);
-        self.updateComponent(name, elValue);
+      var attrs = {};
+      var componentName;
+      var i;
+      if (!this.hasLoaded) { return; }
+      var elComponents = Object.keys(this.components);
```
ngokevin: Do we still want to declare `var`s at top?

src/core/a-entity.js
```diff
@@ -301,14 +302,52 @@ var proto = Object.create(ANode.prototype, {
...
+        var keys = Object.keys(mixinEls[i].attrs);
+        keys.forEach(function (key) {
+          components.push(key);
+        });
+      }
+      return components;
+    }
+  },
+
   updateComponents: {
```
ngokevin: Could use a docstring.

src/core/a-entity.js
ngokevin: We can separate each block of logic with a comment.

```
// Start building component set using current components
// ...

// Update component set with defined attributes.
// ...

// Update component set with mixins.
// ...

// Update component set with default components.
// ...

// With new component set, update all components.
// ...
```

ngokevin: Some `forEach` and less intermediate vars might make this function easier to read as well.

src/core/a-entity.js
```diff
@@ -322,17 +361,24 @@ var proto = Object.create(ANode.prototype, {
    * @param {object} newData - The new properties assigned to the component
```
ngokevin: Need to update this `param`.

src/core/a-entity.js
```diff
@@ -322,17 +361,24 @@ var proto = Object.create(ANode.prototype, {
...
    */
   updateComponent: {
-    value: function (name, newData) {
+    value: function (name, elData) {
       var component = this.components[name];
-      var isDefault = name in this.defaultComponents;
+      var defaultComponents = this.defaultComponents;
```
ngokevin: Don't need this `var`, it's only used once right below.

ngokevin:
A list of specific bullet-pointed changes in the PR description would help

dmarcos:
@ngokevin the changes are in one of the comments. https://github.com/aframevr/aframe/pull/1323#issuecomment-208617407

src/core/a-entity.js
ngokevin: I would mention what it means here if `elData === undefined` and what it means if `elData === null`

src/core/a-entity.js
```diff
@@ -322,17 +361,24 @@ var proto = Object.create(ANode.prototype, {
...
       if (component) {
         // Attribute was removed, remove component if:
         // 1. If component not defined in the defaults/mixins/attribute.
         // 2. If new data is null, then not a default component and component is not defined
         //    via mixins
-        if (!checkComponentDefined(this, name) ||
-            newData === null && !isDefault && !isMixedIn) {
+        if (elData === null) {
+          // Default components cannot be removed
+          if (isDefault) { return; }
```
ngokevin: Can simplify:

```js
// Don't remove default components or mixed-in components.
if (isDefault || isMixedIn)
```

src/core/a-entity.js
```diff
@@ -246,33 +248,32 @@ var proto = Object.create(ANode.prototype, {
    * Initialize component.
    */
   initComponent: {
-    value: function (name, isDependency) {
+    value: function (name, isDependency, attr) {
```
ngokevin: It makes more sense to do `(name, attr, isDependency)`.

src/core/a-entity.js
```diff
@@ -341,7 +387,7 @@ var proto = Object.create(ANode.prototype, {
         return;
       }
       // Component not yet initialized. Initialize component.
-      this.initComponent(name);
+      this.initComponent(name, false, newData);
```
ngokevin: With https://github.com/aframevr/aframe/pull/1323/files#r59427013, you can just call `this.initComponent(name, newData)` without the mystery boolean.

src/core/a-entity.js
ngokevin: It would help to rename `attr` to `attrValue`. I got confused and thought it might be an `attrName`.

src/core/a-entity.js
```diff
@@ -427,18 +474,17 @@ var proto = Object.create(ANode.prototype, {
    */
   setEntityAttribute: {
     value: function (attr, oldVal, newVal) {
-      var component = components[attr];
-      oldVal = oldVal || this.getAttribute(attr);
-      // When creating entities programatically and setting attributes, it is not part
-      // of the scene until it is inserted into the DOM. This does not apply to scenes as
-      // scenes depend on its child entities to load.
-      if (!this.hasLoaded && !this.isScene) { return; }
+      var component = this.components[attr];
```
ngokevin: How come we need the `this`?

src/core/a-entity.js
```diff
@@ -463,31 +509,52 @@ var proto = Object.create(ANode.prototype, {
...
     },
     writable: window.debug
   },
 
   /**
+   * To make the DOM attributes reflect the state of the components.
+   *
+   * @param {bool} iterative - It calls updateDOM on the children
+   **/
+  updateDOM: {
```
ngokevin: Maybe:

```js
flushToDom: {
  value: function (recurse)
```

I like `flush` since it builds the imagery of all of this data being built up that has been blocked from flowing to the DOM.

src/core/a-entity.js
```diff
@@ -463,31 +509,52 @@ var proto = Object.create(ANode.prototype, {
...
-      self.setEntityAttribute(attr, oldValue, componentObj);
+      if (attr === 'mixin') { self.setEntityAttribute('mixin', undefined, componentObj); }
     },
     writable: window.debug
   },
 
   /**
+   * To make the DOM attributes reflect the state of the components.
+   *
+   * @param {bool} iterative - It calls updateDOM on the children
```
ngokevin: Then: `recurse - Whether to recursively call `flushDOM` onto children.`

src/core/a-entity.js
```diff
@@ -463,31 +509,52 @@ var proto = Object.create(ANode.prototype, {
...
+   **/
+  updateDOM: {
+    value: function (iterative) {
+      var components = this.components;
+      var children = this.children;
+      var child;
+      var i;
+      Object.keys(components).forEach(function updateAttribute (name) {
+        components[name].updateDOM();
```
ngokevin: Should this be defined on the entity or the component? Might consider defining on the entity (or a helper function in the entity module) so you are calling an instance of one function rather than many.

```flushComponentToDOM (component)```

File: src/core/a-entity.js [hunk 20]
```diff
@@ -463,31 +509,52 @@ var proto = Object.create(ANode.prototype, {
...
+      for (i = 0; i < children.length; ++i) {
+        child = children[i];
+        if (!child.updateDOM) { continue; }
+        child.updateDOM(iterative);
+      }
+    }
+  },
+
+  /**
    * If `attr` is a component, returns JUST the component data specified in the HTML
```
ngokevin: Need to update this docstring.

src/core/a-entity.js
```diff
@@ -500,10 +567,12 @@ var proto = Object.create(ANode.prototype, {
    */
   getAttribute: {
     value: function (attr) {
-      var component = this.components[attr] || components[attr];
-      var value = HTMLElement.prototype.getAttribute.call(this, attr);
-      if (!component || typeof value !== 'string') { return value; }
-      return component.parse(value, true);
+      var component = this.components[attr];
+      // If there's a cached value we just return it
+      if (component && component.attr !== undefined) {
```
ngokevin: We should rename `component.attr` to denote that it's a value (`component.value`). `this.components[attr].attr` can be confusing here.

src/core/a-mixin.js
```diff
@@ -1,12 +1,26 @@
 /* global HTMLElement */
-var AComponents = require('./component').components;
 var ANode = require('./a-node');
 var registerElement = require('./a-register-element').registerElement;
+var components = require('./component').components;
+var styleParser = require('../utils/').styleParser;
 
 module.exports = registerElement('a-mixin', {
```
ngokevin: Document `@member {object} attrs`

src/core/a-mixin.js
```diff
@@ -16,23 +30,34 @@ module.exports = registerElement('a-mixin', {
...
-          var component = AComponents[attr];
-          var value = HTMLElement.prototype.getAttribute.call(this, attr);
-          if (!component || typeof value !== 'string') { return value; }
-          return component.parse(value);
+          return this.attrs[attr] || HTMLElement.prototype.getAttribute.call(this, attr);
         },
         writable: window.debug
+      },
+
+      updateAttrs: {
```
ngokevin: Docstring

src/core/a-mixin.js
```diff
@@ -16,23 +30,34 @@ module.exports = registerElement('a-mixin', {
...
+        value: function () {
+          var attrs = this.attributes;
+          var attrName;
+          var i;
+          for (i = 0; i < attrs.length; ++i) {
```
ngokevin: Can do: `for (var i = 0)`

src/core/a-mixin.js
```diff
@@ -16,23 +30,34 @@ module.exports = registerElement('a-mixin', {
...
-          if (!component || typeof value !== 'string') { return value; }
-          return component.parse(value);
+          return this.attrs[attr] || HTMLElement.prototype.getAttribute.call(this, attr);
         },
         writable: window.debug
+      },
+
+      updateAttrs: {
+        value: function () {
+          var attrs = this.attributes;
```
ngokevin: ```var attributes``` to not confuse with ```this.attrs```

src/core/component.js
```diff
@@ -1,4 +1,4 @@
-/* global HTMLElement */
+/* global HTMLElement  */
```
ngokevin: Extra space

src/core/component.js
```diff
@@ -25,20 +25,11 @@ var styleParser = utils.styleParser;
...
-  // Check whether we need to rebuild the schema depending on the data.
-  // Call buildData with silent flag to suppress warnings when parsing data before updating
-  // the schema.
-  if (this.updateSchema) {
-    this.updateSchema(buildData(el, name, this.schema, elData, true));
-  }
-  this.data = buildData(el, name, this.schema, elData);
-  this.init();
-  this.update();
+  this.updateCachedValue(attr);
```
ngokevin: Maybe ```function (el, value) { ... setValue(value) ... }```

src/core/scene/a-scene.js
```diff
@@ -312,6 +311,7 @@ var AScene = module.exports = registerElement('a-scene', {
       },
       writable: window.debug
     }
+
```
ngokevin: Whitespace

src/core/schema.js
```diff
@@ -12,7 +12,7 @@ function isSingleProperty (schema) {
   if ('type' in schema) {
     return typeof schema.type === 'string';
   }
-  return 'default' in schema;
+  return 'default' in schema || Object.keys(schema).length === 0;
```
ngokevin: We explicitly removed this before to allow for shaders having ```schema: {}```.

ngokevin: https://github.com/aframevr/aframe/issues/1163

src/extras/primitives/registerPrimitive.js
```diff
@@ -82,10 +75,11 @@ module.exports = function registerPrimitive (name, definition) {
             if (componentData instanceof Object && Object.keys(componentData).length) {
               Object.keys(componentData).forEach(function setProperty (propName) {
                 // Check if component property already defined.
-                var definedData = self.getAttribute(componentName);
-                if (definedData && definedData[propName] !== undefined) { return; }
-
-                self.setAttribute(componentName, propName, componentData[propName]);
+                var component = components[componentName];
+                var attr = self.getAttribute(componentName) || {};
```
ngokevin: Just note here as well `s/attr/value/`

src/extras/primitives/registerPrimitive.js
```diff
@@ -82,10 +75,11 @@ module.exports = function registerPrimitive (name, definition) {
...
+                var data = component.parse(attr);
+                data[propName] = componentData[propName];
+                self.setAttribute(componentName, data);
```
ngokevin: We want to do an update, not a replace.

src/systems/camera.js
```diff
@@ -10,9 +10,23 @@ var DEFAULT_CAMERA_ATTR = 'data-aframe-default-camera';
 module.exports.System = registerSystem('camera', {
   init: function () {
     this.activeCameraEl = null;
+    this.cameraEls = [];
     this.setupDefaultCamera();
   },
 
+  addCamera: function (el) {
```
ngokevin: Just going to double-note here to call it `cameraEl` when it's an entity, and `camera` when it's an object3D.

src/utils/styleParser.js
```diff
@@ -11,6 +11,9 @@ module.exports.parse = function (value) {
   var parsedData;
   if (typeof value !== 'string') { return value; }
   parsedData = styleParser.parse(value);
+  // The style parser returns an object { "" : "test"} when
```
ngokevin: This could fit in one line. `Handle style parser returning {"": <string>} when fed a string.`

tests/core/component.test.js
```diff
@@ -266,4 +266,40 @@ suite('Component', function () {
...
+
+  suite('update', function () {
+    setup(function () {
+      components.dummy = undefined;
+    });
+
+    test('not called if component data does not change', function () {
+      var updateStub = sinon.stub();
+      var TestComponent = registerComponent('dummy', {
+        schema: {color: { default: 'red' }},
```
ngokevin: Let's start tightening the whitespace in object literals to discern them from blocks.

examples/stress-animation/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+      <a-entity position="0 0 80">
+        <a-entity camera look-controls wasd-controls></a-entity>
+      </a-entity>
+
+      <a-entity id="fans"></a-entity>
+      
+    </a-scene>
+  </body>
+  <script>
```
ngokevin: I'll pull in the template component soon to help with this.

ngokevin:
- [x] First pass

dmarcos:
I added a scene component debug so DOM updates can be enabled/disabled. You can measure the impact of updating the DOM on real time by playing with the flag. @ngokevin This is ready for review again. Almost ready to go.

ngokevin:
K, will look over it tmr morning

src/core/a-entity.js
```diff
@@ -301,14 +308,38 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  /**
+   * Updates all the entity components. Given by defaults, mixins and attributes
+   * Default components update before the rest.
+   */
   updateComponents: {
     value: function () {
+      var entityComponents = {};
```
ngokevin: Maybe `definedComponents`?

ngokevin: Or `elComponents` to be consistent with `elData`.

src/core/a-entity.js
```diff
@@ -319,20 +350,25 @@ var proto = Object.create(ANode.prototype, {
...
    */
   updateComponent: {
-    value: function (name, newData) {
+    value: function (name, elData) {
       var component = this.components[name];
       var isDefault = name in this.defaultComponents;
       var isMixedIn = isComponentMixedIn(name, this.mixinEls);
+      // If the new data is not provided we use the cached value
+      var newData = elData === undefined ? this.getAttribute(name) : elData;
+      newData = newData === null && isDefault ? undefined : newData;
```
ngokevin: I get lost on this line. Could use a comment or a clean up here.

src/core/a-entity.js
```diff
@@ -461,37 +495,60 @@ var proto = Object.create(ANode.prototype, {
...
     value: function (attr, value, componentPropValue) {
+      var debug = this.sceneEl && this.sceneEl.getAttribute('debug');
       var self = this;
       var component = this.components[attr] || components[attr];
-      var partialComponentData;
-      value = value === undefined ? '' : value;
       var componentObj = value;  // Deserialized value to send to the component.
-      var componentStr = value;  // Serialized value to send to the DOM.
-      var oldValue;
+      // if there's a previous value it will be cached in the component.
```
ngokevin: nit: capitalize comments / docstrings to be consistent

src/core/a-entity.js
```diff
@@ -461,37 +495,60 @@ var proto = Object.create(ANode.prototype, {
    */
   setAttribute: {
     value: function (attr, value, componentPropValue) {
+      var debug = this.sceneEl && this.sceneEl.getAttribute('debug');
```
ngokevin: Maybe `isDebugMode`. In case we ever have debug console logs `debug('this is a log')`

src/core/a-entity.js
```diff
@@ -461,37 +495,60 @@ var proto = Object.create(ANode.prototype, {
...
+   *
+   * @param {bool} recursive - It calls updateDOM on the children
+   **/
+  flushToDOM: {
+    value: function (recursive) {
+      var components = this.components;
+      var children = this.children;
+      var child;
+      var i;
+      Object.keys(components).forEach(function updateAttribute (name) {
```
ngokevin: Wrong function name

src/core/a-entity.js
```diff
@@ -461,37 +495,60 @@ var proto = Object.create(ANode.prototype, {
...
   },
 
   /**
-   * If `attr` is a component, returns JUST the component data specified in the HTML
-   * by parsing the style-like string into an object. Like a partial version of
-   * `getComputedAttribute` as returned component data does not include applied mixins or
-   * defaults.
+   * To make the DOM attributes reflect the state of the components.
+   *
+   * @param {bool} recursive - It calls updateDOM on the children
```
ngokevin: nit: Usually don't need to specify `It`

src/core/component.js
ngokevin: Can still rename attr -> value, and then just `setValue`

src/core/component.js
```diff
@@ -136,6 +126,38 @@ Component.prototype = {
     return utils.extend({}, data);
   },
 
+  updateCachedValue: function (value, isSinglePropSchema) {
+    var parsedValue;
+    if (value === '') {
+      this.attrValue = undefined;
```
ngokevin: Let's just call it `value`, not clear what `attr` means

src/core/component.js
```diff
@@ -136,6 +126,38 @@ Component.prototype = {
...
+      return;
+    }
+    if (typeof value === 'string') {
+      if (isSingleProp(this.schema)) {
+        parsedValue = this.schema.parse(value);
+        // To avoid bogus double parsings. Like for example
+        // when parsing a src id to it's url #monster -> models/monster.dae
+        if (typeof parsedValue === 'string') { parsedValue = value; }
```
ngokevin: Can you explain how this avoids double parsing? Like with the `src type: `url(monster.dae)` -> `monster.dae` -> *tries to do a selector*. It every case, it will be a string.

src/core/component.js
ngokevin: I think it makes sense because it's not really a cache. It's now the source of truth value.

src/core/a-mixin.js
```diff
@@ -16,23 +33,38 @@ module.exports = registerElement('a-mixin', {
...
+      /**
+       * Update cache of parsed component attributes
+       *
+       */
+      updateAttrs: {
+        value: function () {
+          var attributes = this.attributes;
+          var attrName;
+          var i;
+          for (i = 0; i < attributes.length; ++i) {
```
ngokevin: Could save a line with just `var i = 0` and still have all variables declared on top

src/core/a-mixin.js
```diff
@@ -16,23 +33,38 @@ module.exports = registerElement('a-mixin', {
...
-          if (!component || typeof value !== 'string') { return value; }
-          return component.parse(value);
+          return this.attrs[attr] || HTMLElement.prototype.getAttribute.call(this, attr);
         },
         writable: window.debug
+      },
+
+      /**
+       * Update cache of parsed component attributes
+       *
```
ngokevin: extra line

src/core/a-mixin.js
```diff
@@ -16,23 +33,38 @@ module.exports = registerElement('a-mixin', {
...
-          var value = HTMLElement.prototype.getAttribute.call(this, attr);
-          if (!component || typeof value !== 'string') { return value; }
-          return component.parse(value);
+          return this.attrs[attr] || HTMLElement.prototype.getAttribute.call(this, attr);
         },
         writable: window.debug
+      },
+
+      /**
+       * Update cache of parsed component attributes
```
ngokevin: probably wouldn't think of it as a cache, right?

src/core/a-mixin.js
```diff
@@ -1,12 +1,29 @@
 /* global HTMLElement */
-var AComponents = require('./component').components;
 var ANode = require('./a-node');
 var registerElement = require('./a-register-element').registerElement;
+var components = require('./component').components;
+var styleParser = require('../utils/').styleParser;
 
+/**
+ * @member {object} attrs - Cache of pre parsed component attributes
```
ngokevin: does it still work with normal attributes?

`<a-mixin id="animation-values" repeat="indefinite"></a-mixin>`

src/core/a-entity.js
```diff
@@ -568,11 +627,22 @@ function checkComponentDefined (el, name) {
...
   // Check if element contains the component.
-  if (el.hasAttribute(name)) { return true; }
+  if (el.components[name] && el.components[name].attrValue) { return true; }
 
   return isComponentMixedIn(name, el.mixinEls);
 }
 
+function getMixedInComponents (entityEl) {
+  var components = [];
+  entityEl.mixinEls.forEach(function (mixinEl) {
```
ngokevin: name function

src/core/a-entity.js
```diff
@@ -568,11 +627,22 @@ function checkComponentDefined (el, name) {
...
+    Object.keys(mixinEl.attrs).forEach(addComponent);
+    function addComponent (key) {
```
ngokevin: IMO moving it down one line doesn't provide much abstraction

src/core/a-entity.js
```diff
@@ -461,37 +495,60 @@ var proto = Object.create(ANode.prototype, {
...
+        // On debug mode we write the component state to the DOM attributes
+        if (debug) { component.flushToDOM(); }
+        return;
       }
-      oldValue = this.getAttribute(attr);
-      ANode.prototype.setAttribute.call(self, attr, componentStr);
-      self.setEntityAttribute(attr, oldValue, componentObj);
+
+      ANode.prototype.setAttribute.call(self, attr, value);
+      if (attr === 'mixin') { self.setEntityAttribute('mixin', undefined, componentObj); }
```
ngokevin: Can do `self.setEnttiyAttribute(attr, ...)`. Could use a comment

src/core/a-entity.js
```diff
@@ -1,4 +1,4 @@
-/* global HTMLElement */
+/* global HTMLElement  */
```
ngokevin: extra space

src/core/a-entity.js
dmarcos: It's not. You're updating the DOM attribute

tests/karma.conf.js
```diff
@@ -20,6 +20,7 @@ module.exports = function (config) {
     envPreprocessor: [
       'TEST_ENV'
     ],
+    //exclude: [ 'tests/components/*.js', 'tests/core/*.js', 'tests/core/scene/*.js', 'tests/utils/*.js', 'tests/extras/primitives/primitives/*.js'],
```
ngokevin: Don't need this

tests/core/a-entity.test.js
```diff
@@ -247,6 +247,20 @@ suite('a-entity', function () {
     });
   });
 
+  suite('flushToDOM', function () {
```
ngokevin: recursive test?

tests/core/a-entity.test.js
```diff
@@ -247,6 +247,20 @@ suite('a-entity', function () {
...
+    test('updates DOM attributes with material properties', function () {
```
ngokevin: don't need to mention material properties, it's agnostic

tests/components/raycaster.test.js
```diff
@@ -7,6 +7,7 @@ suite('raycaster', function () {
     var el = this.el = document.createElement('a-entity');
     el.setAttribute('raycaster', '');
     parentEl.appendChild(el);
+    if (el.hasLoaded) { done(); }
```
ngokevin: You can move the `parentEl.appendChild(el)` after the `addEventListener` instead

src/core/component.js
```diff
@@ -136,6 +126,38 @@ Component.prototype = {
...
+        this.attrValue = parsedValue;
+      } else {
+        // We just parse using the style parser to avoid double parsing
```
ngokevin: could fit in a line, possibly take away some unnecessary words:

```Parse using style parsing to avoid style parsing of individual properties.```

File: src/core/component.js [hunk 57]
```diff
@@ -136,6 +126,38 @@ Component.prototype = {
...
+        // of individual properties.
+        this.attrValue = styleParser.parse(value);
+      }
+      return;
+    }
+    this.attrValue = value !== undefined ? extendProperties({}, value, isSinglePropSchema) : this.attrValue;
+  },
+
+  /**
+   * Writes cached attribute data to the entity DOM element.
```
ngokevin: Double note: we should portray it as "component value" rather than "cached attribute data" to signify what the source of truth is

src/core/a-entity.js
dmarcos: This is just for consistency. Moving the functions below can really improves readability like in this case:
https://github.com/dmarcos/aframe/blob/updateRefactor/src/core/a-entity.js#L328

ngokevin: OK. though i'd probably argue it improves readability there, but doesn't really help here.

src/core/a-mixin.js
dmarcos: consistency.

src/core/component.js
dmarcos: no. It's not the truth value. It's just the value that the DOM attribute would have. The truth value is ```this.data```

src/core/component.js
dmarcos: attr is the value the cached value you set with setAttribute. ```value``` is meaningless. All variables have a value

src/core/component.js
dmarcos: If we parse the cached value then ```#monster``` turns in to ```models/monster```. When calculating the final estate of the component via buildData, ```models/monster``` will be parsed again following the squema. That's not what you want.

src/core/a-entity.js
ngokevin: OK. got confuse since we use `update` and overload `attribute` a lot. might be good to reuse the word `flush`

dmarcos: I can call it ```updateDOMAttribute```

src/core/component.js
ngokevin: I get that, just not clear how this addresses that.

src/core/component.js
ngokevin: OK

src/core/component.js
```diff
@@ -136,6 +126,38 @@ Component.prototype = {
     return utils.extend({}, data);
   },
 
+  updateCachedValue: function (value, isSinglePropSchema) {
```
ngokevin: We call it `cachedValue` in one place and `attrValue` in another. Should we call it `cachedAttrValue` everywhere?

dmarcos:
I made a big clean up.

examples/stress-animation/index.html
```diff
@@ -0,0 +1,77 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Animation</title>
+    <meta name="description" content="Animation - A-Frame">
+    <script src="../../dist/aframe.js"></script>
+  </head>
+  <body>
+    <a-scene stats="true">
+      <a-assets>
+        <a-mixin id="ball" geometry="primitive: sphere; radius:1"  material="shader: flat; color: red">
```
ngokevin: space after radius `radius: 1`, extra space before `material`

examples/stress-animation/index.html
```diff
@@ -0,0 +1,77 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Animation</title>
```
ngokevin: Title should mention perf

examples/stress-animation/index.html
```diff
@@ -0,0 +1,77 @@
...
+        <a-mixin id="spin" attribute="rotation" dur="4000"
+                 repeat="indefinite" easing="linear" dur="3000" to="0 0 360"><a-mixin>
+      </a-assets>
+
+      <a-entity position="0 0 80">
+        <a-entity camera look-controls wasd-controls></a-entity>
+      </a-entity>
+
+      <a-entity id="fans"></a-entity>
+
```
ngokevin: extra line

src/core/a-entity.js
```diff
@@ -49,6 +48,13 @@ var proto = Object.create(ANode.prototype, {
    */
   attributeChangedCallback: {
     value: function (attr, oldVal, newVal) {
+      var component = this.components[attr];
+      // If the empty string is passed by the component initialization
```
ngokevin: why are all the comments clamped with new lines? lots of room

src/core/a-entity.js
```diff
@@ -78,11 +84,15 @@ var proto = Object.create(ANode.prototype, {
...
   applyMixin: {
     value: function (attr) {
-      var attrValue = this.getAttribute(attr);
+      var attrValue;
       if (!attr) {
         this.updateComponents();
         return;
       }
+      attrValue = this.getAttribute(attr);
+      // turn null into undefined because getAttribute
```
ngokevin: can 1line `Make absence of attribute for getAttribute return undefined rather than null`.

src/core/a-entity.js
```diff
@@ -244,35 +252,38 @@ var proto = Object.create(ANode.prototype, {
 
   /**
    * Initialize component.
+   *
+   * @param {string} name - the component name.
```
ngokevin: nit: but could be more concise without `the`, and consistent caps

`Component name.` + `Component data.`

src/core/a-entity.js
```diff
@@ -244,35 +252,38 @@ var proto = Object.create(ANode.prototype, {
...
+   * @param {object} data - The component data
+   * @param {boolean} isDependency. True if the component is a dependency.
```
ngokevin: extra period

src/core/a-entity.js
```diff
@@ -287,28 +298,59 @@ var proto = Object.create(ANode.prototype, {
...
+      // Don't remove default or mixed in components
+      if (isDefault || isMixedIn) { return; }
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
     }
   },
 
+  /**
+   * Updates all the entity components. Given by defaults, mixins and attributes
```
ngokevin: `entity's components` or just `components`

src/core/a-entity.js
```diff
@@ -427,18 +482,22 @@ var proto = Object.create(ANode.prototype, {
...
-        this.updateStateMixins(newVal, oldVal);
-        this.updateComponents();
+        this.mixinUpdate(newVal, oldVal);
         return;
       }
-      if (component) { this.updateComponent(attr, newVal); }
+    }
+  },
+
+  mixinUpdate: {
```
ngokevin: maybe `updateFromMixins`?

src/core/a-mixin.js
```diff
@@ -1,38 +1,72 @@
...
+/**
+ * @member {object} attrsCache - Cache of pre parsed component attributes
+ */
 module.exports = registerElement('a-mixin', {
   prototype: Object.create(
     ANode.prototype,
     {
+      createdCallback: {
+        value: function () {
+          this.attrsCache = {};
```
ngokevin: Can just call this `attrCache`, that will imply plural.

Though you might want to name this `components` or `componentCache` because these changes will make it so mixins will only work with components. Seems they won't work with any attribute like animations/primitives.

src/core/component.js
```diff
@@ -136,6 +126,48 @@ Component.prototype = {
     return utils.extend({}, data);
   },
 
+  updateCachedAttrValue: function (value) {
```
ngokevin: docstring/comment

src/core/component.js
```diff
@@ -136,6 +126,48 @@ Component.prototype = {
...
+      return;
+    }
+    if (typeof value === 'string') {
+      this.attrValue = this.parseAttrValueForCache(value);
+      return;
+    }
+    this.attrValue = value !== undefined ? extendProperties({}, value, isSinglePropSchema) : this.attrValue;
+  },
+
+  parseAttrValueForCache: function (value) {
```
ngokevin: docstring/comment

src/core/component.js
```diff
@@ -136,6 +126,48 @@ Component.prototype = {
...
+    var parsedValue;
+    if (typeof value !== 'string') { return value; }
+    if (isSingleProp(this.schema)) {
+      parsedValue = this.schema.parse(value);
+      // To avoid bogus double parsings. The cached values will
```
ngokevin: double note: can unclamp this comment line

src/core/component.js
ngokevin: Maybe `parseAttrValue` is descriptive enough

ngokevin: This could be a nicely scoped helper/private function

`function parseAttrValue (schema)`

ngokevin: This could be a nicely scoped helper/private function `parseAttrValue (value, schema)`. Do you plan on being able to override this?

src/core/a-mixin.js
dmarcos: mixins will work with anything if you use ```mixinEl.getAttribute('xxx')```. The cache is just to preparse the component backed attributes to minimize parsing passes.

src/core/component.js
```diff
@@ -246,19 +283,16 @@ module.exports.registerComponent = function (name, definition) {
  * @return {object} The component data
  */
 function buildData (el, name, schema, elData, silent) {
-  var componentDefined = elData !== null;
-  var data = {};
+  var componentDefined = elData;
```
ngokevin: should be a boolean, `!!elData` maybe?

tests/components/raycaster.test.js
ngokevin: aka you don't need this line if you do ^^

tests/core/a-entity.test.js
ngokevin: still need a case for recursive flushing test

tests/core/a-entity.test.js
ngokevin: can still remove that part

tests/karma.conf.js
ngokevin: still don't need this

ngokevin:
r+wc

src/core/a-mixin.js
ngokevin: ok, might want to name it for components: `componentAttrCache`

fernandojsg:
Just checked with the editor and after some small changes everything works fine! I tried to inject `debug=true` statically and dynamically and both seems to works as expected, so from my side is looking good.

dmarcos:
All right. I'm going to merge this monster patch. Demos seem to work. I feel better about code quality and it's also much faster :)

ngokevin:
Nice work 🎆 

There're some unaddressed nits, but I can help with those if you want.

dmarcos:
I changed stuff after the last review. Everything should be addressed.

ngokevin:
GitHub shows some unaddressed diff lines, but not a big deal.

src/core/component.js
dmarcos: I think shorter lines are easier to read. Lines that are much longer than the code block above and underneath bother me.

tests/core/component.test.js
dmarcos: I don't have much issues mixing up blocks and object literals. We already use spaces for literals and If I change it here I have to change it everywhere. Not worth the effort IMO.

src/core/a-entity.js
dmarcos: updateFromMixins is not a good name either. This is the method triggered when there's a mixin attribute update (that triggers a components update)

Issue #1033, opened 2016-02-26
Create new linter using standard-engine + ESLint


ngokevin:
We have several rules we want to cherry-pick:

http://eslint.org/docs/rules/

- curly - specify curly brace conventions for all control statements
- no-else-return - disallow else after a return in an if
- vars-on-top - require declaration of all vars at the top of their containing scope
- camelcase - require camel case names
- func-names - require function expressions to have a name
- indent - specify tab or space width for your code 
- no-lonely-if - disallow if as the only statement in an else block

ngokevin:
Let's just follow `semistandard`

PR #1332, opened 2016-04-04
fix registerPrimitive property merging (fixes #1324)

### LINKED ISSUE #1324: Setting fog: false to a-sky is ignored
**Description:**

`<a-scene fog="color: #fff"><a-sky material="fog: false" rotation='0 -45 0' src='https://www.scenevr.com/prerelease/images/sky.png'></a-sky></a-scene>`

The sky isn't rendered, you just see the fog color instead.

- A-Frame Version: 2.0.0
- Platform/Device: Chrome

ngokevin:
**Description:** Mapped attributes were overwriting defined components.

**Changes proposed:**
- Don't keep track of internal component data. Just do non-destructive updates when possible with `setAttribute(component, propName, propValue)` syntax.
- Move default component initialization to separate method.
- Support merging data with default components + defined components.

### DIFF
--- a/src/extras/primitives/registerPrimitive.js
+++ b/src/extras/primitives/registerPrimitive.js
@@ -4,7 +4,7 @@ var registerElement = require('../../core/a-register-element').registerElement;
 var utils = require('../../utils/');
 
 var debug = utils.debug;
-var log = debug('extras:primitives');
+var log = debug('extras:primitives:debug');
 
 module.exports = function registerPrimitive (name, definition) {
   name = name.toLowerCase();
@@ -46,16 +46,10 @@ module.exports = function registerPrimitive (name, definition) {
           var self = this;
           var attributes = this.attributes;
 
-          // Apply default components.
-          this.componentData = cloneObject(this.defaultAttributes);
-          Object.keys(this.componentData).forEach(function (componentName) {
-            if (!self.hasAttribute(componentName)) {
-              self.setAttribute(componentName, self.componentData[componentName]);
-            }
-          });
+          this.applyDefaultComponents();
 
           // Apply initial attributes.
-          Object.keys(attributes).forEach(function (attributeName) {
+          Object.keys(attributes).forEach(function applyInitial (attributeName) {
             var attr = attributes[attributeName];
             self.syncAttributeToComponent(attr.name, attr.value);
           });
@@ -72,6 +66,33 @@ module.exports = function registerPrimitive (name, definition) {
             return;
           }
           this.syncAttributeToComponent(attr, newVal);
+        }
+      },
+
+      applyDefaultComponents: {
+        value: function () {
+          var self = this;
+          var defaultData = this.defaultAttributes;
+
+          // Apply default components.
+          Object.keys(defaultData).forEach(function applyDefault (componentName) {
+            var componentData = defaultData[componentName];
+
+            // Set component properties individually to not overwrite user-defined components.
+            if (componentData instanceof Object && Object.keys(componentData).length) {
+              Object.keys(componentData).forEach(function setProperty (propName) {
+                // Check if component property already defined.
+                var definedData = self.getAttribute(componentName);
+                if (definedData && definedData[propName] !== undefined) { return; }
+
+                self.setAttribute(componentName, propName, componentData[propName]);
+              });
+              return;
+            }
+
+            // Component is single-property schema, just set the attribute.
+            self.setAttribute(componentName, componentData);
+          });
         }
       },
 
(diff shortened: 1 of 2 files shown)


src/extras/primitives/registerPrimitive.js
```diff
@@ -75,6 +69,33 @@ module.exports = function registerPrimitive (name, definition) {
...
+            // Set component properties individually to not overwrite user-defined components.
+            if (componentData instanceof Object && Object.keys(componentData).length) {
+              Object.keys(componentData).forEach(function setProperty (propName) {
+                // Check if component property already defined.
+                var definedData = self.getAttribute(componentName);
+                if (definedData && definedData[propName] !== undefined) { return; }
+
+                self.setAttribute(componentName, propName, componentData[propName]);
+              });
+              return;
```
cvan: perhaps we should `return` only if `self.setAttribute(…)` actually got called?

src/extras/primitives/registerPrimitive.js
```diff
@@ -75,6 +69,33 @@ module.exports = function registerPrimitive (name, definition) {
         }
       },
 
+      applyDefaultComponents: {
+        value: function () {
+          var self = this;
+          var defaultData = cloneObject(this.defaultAttributes);
```
cvan: I wasn't clear from the `cloneObject` docstring - why don't we just use `utils.extend` (which is `Object.assign` under the hood - and polyfilled if not natively available)? `Object.assign` does shallow object cloning, one level only, which seems to be what you want. I'm probably missing something though.

src/extras/primitives/registerPrimitive.js
```diff
@@ -75,6 +69,33 @@ module.exports = function registerPrimitive (name, definition) {
...
+          // Apply default components.
+          Object.keys(defaultData).forEach(function applyDefault (componentName) {
+            var componentData = defaultData[componentName];
+
+            // Set component properties individually to not overwrite user-defined components.
+            if (componentData instanceof Object && Object.keys(componentData).length) {
+              Object.keys(componentData).forEach(function setProperty (propName) {
+                // Check if component property already defined.
+                var definedData = self.getAttribute(componentName);
+                if (definedData && definedData[propName] !== undefined) { return; }
```
cvan: or do we want to just check `propName in definedData`?

src/extras/primitives/registerPrimitive.js
```diff
@@ -75,6 +69,33 @@ module.exports = function registerPrimitive (name, definition) {
...
+          var self = this;
+          var defaultData = cloneObject(this.defaultAttributes);
+
+          // Apply default components.
+          Object.keys(defaultData).forEach(function applyDefault (componentName) {
+            var componentData = defaultData[componentName];
+
+            // Set component properties individually to not overwrite user-defined components.
+            if (componentData instanceof Object && Object.keys(componentData).length) {
+              Object.keys(componentData).forEach(function setProperty (propName) {
```
cvan: I like naming the anonymous functions :+1:

src/extras/primitives/registerPrimitive.js
ngokevin: `definedData` can be null

src/extras/primitives/registerPrimitive.js
ngokevin: I'll try `deepAssign`, maybe it was written before we had that

src/extras/primitives/registerPrimitive.js
ngokevin: I would add it to our linter https://github.com/aframevr/aframe/issues/1033

But I don't want that to catch in our test suite.

src/extras/primitives/registerPrimitive.js
cvan: but couldn't `null` be a valid value for some component properties? probably not with primitives. you know better than I do.

ngokevin: Yeah, it's valid. But we're checking for `null` on the entire component dataset here, not just one property.

src/extras/primitives/registerPrimitive.js
ngokevin: I'm only returning here to mimic an `if/else`. Not a big fan of that style since it's less clear on the intent, but dmarcos likes that.

cvan: okay, I guess I understand

src/extras/primitives/registerPrimitive.js
```diff
@@ -75,6 +69,33 @@ module.exports = function registerPrimitive (name, definition) {
         }
       },
 
+      applyDefaultComponents: {
+        value: function () {
+          var self = this;
+          var defaultData = utils.extendDeep(this.defaultAttributes);
```
dmarcos: Is it really necessary to create a local copy of the object?

ngokevin: yeah. if you modify this object, it will affect all other entities created from this prototype.

dmarcos: I don't see where is the defaultData object modified.

ngokevin: OK, I think since I removed keeping track of component data, this is no longer needed. Good catch.

PR #1338, opened 2016-04-04
default to BufferGeometry for all geometries (fixes #633)

### LINKED ISSUE #633: Support buffer geometries

ngokevin:
**Description:**

http://threejs.org/docs/#Reference/Core/BufferGeometry

```
This class is an efficient alternative to Geometry, because it stores all data, including vertex positions, face indices, normals, colors, UVs, and custom attributes within buffers; this reduces the cost of passing all this data to the GPU. This also makes BufferGeometry harder to work with than Geometry; rather than accessing position data as Vector3 objects, color data as Color objects, and so on, you have to access the raw data from the appropriate attribute buffer. This makes BufferGeometry best-suited for static objects where you don't need to manipulate the geometry much after instantiating it. 
```

**Changes proposed:**
- Default to use BufferGeometry for all geometries using `BufferGeometry.fromGeometry`.
- Have tests not use BufferGeometry to be able to assert geometry type/parameters.
- [ ] replace geometry.translate with a separate `pivot/translate` component, place it in `extras/` since geometry.translate does not work with buffer geometries.

### DIFF
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -12,6 +12,7 @@ var error = utils.debug('components:geometry:error');
  */
 module.exports.Component = registerComponent('geometry', {
   schema: {
+    buffer: {default: true},
     primitive: {default: ''},
     translate: {type: 'vec3'}
   },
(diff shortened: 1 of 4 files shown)


ngokevin:
This saves memory. On the stress-entities test scene with 4000 cubes (two trials):

- Without buffer geometries: 145MB of memory on a fresh Firefox
- With buffer geometries: 80MB of memory on a fresh Firefox

PR #1339, opened 2016-04-04
geometry.translate -> pivot component in extras/ (fixes #1341)

### LINKED ISSUE #1341: Animation demos are no longer working as expected.
**Description:**
  On Current master branch previously working animation examples no longer behave as expected.
- A-Frame Version:  Master
- Platform/Device:  All

Evidence:
1. Checkout Master
2. `npm run dev`
3. Visit animation unfold example.

Expected: Each plane should individually animate delayed from each other.
Seen: The next two planes to animate appear to be 'attached' and share rotation values until the next one is kicked off.

Maybe this behavior is now expected with the updates to primitives and the example code should be updated?

ngokevin:
**Description:**

`geometry.translate` doesn't work with BufferGeometry, and the property was always awkward. The feature also made the geometry component more complex to maintain. A lot of examples use this feature, but we want to remove this feature.

Replacing it with a pivot component that is a helper for wrapping it within another entity. This might be useful for the editor to change the pivot point without having to create/expose additional entities.

I'm putting it `extras/` for now, but it has a case for being a standard component. While it could be done by wrapping it within another entity, we might not want to create an extra entity just to change the pivot point and abstract that completely.

Difference between the pivot and geometry.translate is that rather than translating the entity, `pivot` will internally move the pivot point. This is done by translating the entity and untranslating it at the parent level. The effect is that pivot vectors are the opposite of translate vectors (e.g, `-1 -2 -3` vs `1 2 3`) and since it doesn't translate, you don't have to manually untranslate using position.

Possible caveat is that since we untranslate the pivot using position, a future update to position might mess things up. We would have to make the position component aware of the pivot component. Right now, the pivot component works best on static objects and you can use a wrapper entity for future updates to position.

**Changes proposed:**
- Pivot component (that internally moves the pivot point)
- Remove geometry.translate (which translated the entity while keeping the pivot point)
- Update examples to use pivot over translate
- Lot of examples did not need translate/pivot, updated those


dmarcos:
The main reason to have pivots was being able to change the origin of a loaded model. Now that models are a separate component we could make translate a property of the model components.

ngokevin:
We were previously using `translate` to change the pivot point of meshes for animations. Many of the animation examples were using it to change the anchor of the animation. Not with just models, you might want to change the pivot point for a group of meshes.

Granted that could be done with a wrapper entity. Though we might want to abstract that. In 3D modelling, setting the pivot point doesn't create another object. It just does it internally, so it would be nice to match that expectation.

examples/animation-generic-logo/index.html
```diff
@@ -41,20 +41,20 @@
 
         <a-entity id="dots1" rotation="0 0 45">
           <a-animation attribute="scale" from="1 0 1" to="1 0.6 1" begin="800" dur="750" fill="both" easing="ease-out"></a-animation>
-          <a-cylinder position="0 2 0" radius="0.04" height="2" translate="0 -1 0" color="#EF2D5E" open-ended="true" side="double">
```
dmarcos: I haven't tried but removing translation should change the animation of this demo. Does this look all right?

examples/animation-warps/index.html
```diff
@@ -18,11 +18,11 @@
         <!-- Green -->
         <a-entity>
           <a-animation attribute="scale" from="1 0 1" to="1 1 1" dur="2000" fill="both"></a-animation>
-          <a-cylinder position="0 2 0" radius="2" height="2" translate="0 -1 0" color="#7BC8A4" open-ended="true" side="double">
+          <a-cylinder position="0 2 0" radius="2" height="2" color="#7BC8A4" open-ended="true" side="double">
```
dmarcos: Same for this. How does the animation change without the translation?

dmarcos: Should not the pivot component replace the translate?

examples/animation-generic-logo/index.html
ngokevin: Yeah, I couldn't tell the difference.

examples/animation-warps/index.html
ngokevin: Same for this one, I couldn't tell the difference.

ngokevin: I'll double-check. Maybe it was subtle.

dmarcos: A translation of 0 -1 0 seems pretty big given the size and the rest of the positions.

ngokevin: Yeah, I can't really tell the difference.

src/extras/components/pivot.js
```diff
@@ -0,0 +1,46 @@
...
+
+    // Set outer group as new object3D.
+    el.object3D = outerGroup;
+
+    // Apply pivot to original group.
+    originalGroup.position.set(-1 * data.x, -1 * data.y, -1 * data.z);
+
+    // Transfer rotation to outer group.
+    originalGroup.rotation.set(0, 0, 0);
+    console.log(originalRotation);
```
dmarcos: Do we need this console.log?

ngokevin: removed

src/extras/components/pivot.js
```diff
@@ -0,0 +1,45 @@
...
+
+    // Set outer group as new object3D.
+    el.object3D = outerGroup;
+
+    // Apply pivot to original group.
+    originalGroup.position.set(-1 * data.x, -1 * data.y, -1 * data.z);
+
+    // Transfer rotation to outer group.
+    originalGroup.rotation.set(0, 0, 0);
+    outerGroup.rotation.set(originalRotation.x, originalRotation.y, originalRotation.z);
```
dmarcos: what about: 
````javascript
outerGroup.rotation.copy(originalGroup.rotation);
originalGroup.rotation.set(0, 0, 0);
````

Issue #1334, opened 2016-04-04
animation `from` value not being set for component properties


ngokevin:
**Description:**

`<a-animation attribute="material.color">` won't retrieve the current `from` value because it is currently fetching the value via `var currentValue = el.getComputedAttribute(attribute);` and not doing the `split` to retrieve the component name + component property name.

ngokevin:
Fixed at #1340

PR #1340, opened 2016-04-05
Bugfix/component properties

JohnRodney:
**Description:**
  Adds a method to an entity to understand both 'attribute' and 'attribute.property' and returns the     appropriate value in either case.
**Changes proposed:**
-  Remove special casing of circumstances from animation.js
-  Adds native method to element to understand this syntax.


JohnRodney:
@ngokevin I opted to add the functionality to the element to understand the component attribute rather than write more special cases into animation. If you would prefer a different approach please let me know. This is in reference to #1334

JohnRodney:
Hold off on this... even though all the tests work there are visual regressions in the system.  Found when going to some examples will dig deeper.  A bit strange that all animation tests work when the actual components themselves don't.

JohnRodney:
Hmm the visual regression occurred upstream and not from this PR I will open a new issue.

src/core/a-entity.js
```diff
@@ -525,6 +525,25 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  /**
+   * A mapping of attribute to handle component attributes and singular attributes
+   *
+   * @param {string} attr dot notation or singular 'color' or 'material.color'
+   * @returns {object|string|number} Value of type of component property.
+   */
+  getComputedAttributeFor: {
```
ngokevin: Might be easier to keep this as a utility function within the animation module for now. (e.g., `getComputedAttributeFor(el, attribute)`.

Not sure if we want to fully support dot notation because if we do it here, then we have to do it for `setAttribute` as well to be consistent, and it'll increase API surface area / documentation / code to maintain. If we start to see a bigger need for dot notation, then we could re-eval.

ngokevin: Might be easier to keep this as a utility function within the animation module for now. (e.g., `getComputedAttributeFor(el, attribute)`.

Not sure if we want to fully support dot notation because if we do it here, then we have to do it for `setAttribute` as well to be consistent, and it'll increase API surface area / documentation / code to maintain. If we start to see a bigger need for dot notation, then we could re-eval.

JohnRodney: This has been updated, but the PR is failing to update on the branch push.  Looks like github status at the moment is saying some backend services unavailable.  Will recheck when they report servers are running normally. 

If the issue doesn't repair I will close and open a new PR.

ngokevin: Yeah, GitHub has been acting up last hour.

src/core/a-animation.js
```diff
@@ -463,3 +463,21 @@ function strToBool (str) {
...
 }
+
+/**
+ * A mapping of attribute to handle component attributes and singular attributes
+ *
+ * @param {object} element to look up attribute on
+ * @param {string} attr dot notation or singular 'color' or 'material.color'
+ * @returns {object|string|number} resulting value of type of component property.
+ */
+function getComputedAttributeFor (el, attribute) {
```
ngokevin: Maybe name `getAttributeWithDotNotation`?

src/core/a-animation.js
```diff
@@ -463,3 +463,21 @@ function strToBool (str) {
...
+  var attributeSplit = attribute.split('.');
+  if (attributeSplit.length === 1) {
```
ngokevin: Might be more readable if separated with comments:

```js
// Blah blah.
if (clause) {
  return value;
}
// Blah blah 2.
if (clause2) {
  return value2;
}
```

src/core/a-animation.js
```diff
@@ -463,3 +463,21 @@ function strToBool (str) {
 function boolToNum (bool) {
   return bool ? 1 : 0;
 }
+
+/**
+ * A mapping of attribute to handle component attributes and singular attributes
+ *
+ * @param {object} element to look up attribute on
```
ngokevin: ```@param {Element} el - To look up attribute on.```

src/core/a-animation.js
```diff
@@ -463,3 +463,21 @@ function strToBool (str) {
...
+ * @param {string} attr dot notation or singular 'color' or 'material.color'
```
ngokevin: Can separate argument name with a dash.

```@param {string} attr - Dot notation...```

File: src/core/a-animation.js [hunk 6]
```diff
@@ -463,3 +463,21 @@ function strToBool (str) {
...
+ * @returns {object|string|number} resulting value of type of component property.
```
ngokevin: Probably don't need to specify type here:

```js
@returns Resulting value of component property.
```

ngokevin:
Can we add a test for this?

src/core/a-animation.js
JohnRodney: @ngokevin it gets the attribute no matter if it is dot notation or not.

ngokevin: Yeah, though it seems the purpose of it is to support dot notation, with fallback. Is that right?

ngokevin:
Should we add a test that tests the whole animation grabbing the correct `from` value if using dot notation?

JohnRodney:
Absolutely.  The rest of the refactors are addressed other than changing to the function to get with dot notation because it supports both and that name might be misleading?

ngokevin:
OK, that's fine. I'd emphasize that in the docstring more ```getComputedAttribute that supports dot notation to look up a component property.```

JohnRodney:
Sounds good.

ngokevin:
Did you see my comment about adding a more integrated test?

JohnRodney:
@ngokevin yea I wanted to see why the last refactor didn't pass the integration test even though it did local before I added another test case.  Not sure if this was due to github having issues still so wanted to push the doc refactor first and ensure tests were fine before adding another.

dmarcos:
I know it's a pain the ass (I'm going through this myself now) but we need to isolate the problem. @JohnRodney  Does the failure go away if you comment your test out?

JohnRodney:
@dmarcos sorry didn't get around to trying last night.  I had tried to download a different version of firefox so I could get it to fail on my local, but I never got it to.  Had to wrap up some other stuff on a project today will be revisiting this again in a few hours and figure it out.  As I will need to use an entity generator to get a component to exist on the test.

ngokevin:
Got the tests working https://github.com/aframevr/aframe/pull/1349

I think the problem is you are creating a scene outside of the setup/teardown. We have global setup methods that expect the scene to be created in a setup handler.

ngokevin:
You can apply this diff https://github.com/aframevr/aframe/pull/1349/commits/5bfc71ae63bdb2fdaaca90daf11de5129862bd96

dmarcos:
@JohnRodney What's left to do here?

JohnRodney:
@dmarcos Sorry with all the stuff going on and catching up with work I just haven't had the time to address this.  I can close it and re-open after I finish the feature/animateColor PR.  Or someone can pick up where I left off here.  I know its probably bugging you to have lingering PRs.

ngokevin:
I can volunteer to pick this up.

JohnRodney:
@dmarcos I applied the upstream changes here.

This build seems to be failing due to docker not installing from NPM properly.  This is a problem with NPM itself and not this branch.  https://github.com/npm/npm/issues/9418#issuecomment-170244027

The problem seems to be intermittent waiting to see if this shows up somewhere else.  Or just this branch I could try closing and opening a new PR.  Waiting to see what ya'lls suggestions are.

dmarcos:
@JohnRodney I have restarted the travis build. It seems you need to be an admin of the repo to do it.

src/core/a-animation.js
```diff
@@ -499,6 +500,31 @@ function boolToNum (bool) {
...
+function getComputedAttributeFor (el, attribute) {
+  var attributeSplit = attribute.split('.');
+  var componentName = attributeSplit[0];
+  var componentPropName = attributeSplit[1];
+  // If the attribute is singular call normal getComputedAttribute function
+  if (attributeSplit.length === 1) {
+    return el.getComputedAttribute(componentName);
+  }
+  // If the component exists as an attribute return the property on it
+  if (el.getComputedAttribute(componentName)) {
```
dmarcos: You can declare a variable at the top ```var attributeValue``` and do:
```javascript
attributeValue = el.getComputedAttribute(componentName);
if (attributeValue) { return attributeValue; }
return el.getComputedAttribute(componentPropName);
````

**Comment by dmarcos** [comment]
Almost there. Just one more style change.

File: src/core/a-animation.js (hunk 7 above)
Comment (JohnRodney) [inline src/core/a-animation.js]: @dmarcos I may be misunderstanding, but this doesn't work in the case of an attribute needing to be split.  We need 3 cases here one for a default, one for a attribute split and then fallback to the native if it doesn't work.

Comment (JohnRodney) [inline src/core/a-animation.js]: I pushed it up with the example I showed and all tests are working.  This should be good.

Comment (dmarcos) [inline src/core/a-animation.js]: the code of the whole function I think would be:
````javascript
function getComputedAttributeFor (el, attribute) {
  var attributeSplit = attribute.split('.'); 
  var attributeValue;
  var componentName = attributeSplit[0]; 
  var componentPropName = attributeSplit[1]; 
  // If the attribute is singular call normal getComputedAttribute function
  if (attributeSplit.length === 1) {
    return el.getComputedAttribute(componentName);
   }
  attributeValue = el.getComputedAttribute(componentName);
  if (attributeValue) { return attributeValue; }
  return el.getComputedAttribute(componentPropName);
}
````

File: src/core/a-animation.js [hunk 8]
```diff
@@ -499,6 +500,28 @@ function boolToNum (bool) {
...
+ * @returns Resulting value of component property.
+ */
+function getComputedAttributeFor (el, attribute) {
+  var attributeSplit = attribute.split('.');
+  var componentName = attributeSplit[0];
+  var componentPropName = attributeSplit[1];
+  var attributeValue = el.getComputedAttribute(componentName);
+  // If the attribute is singular call normal getComputedAttribute function
+  if (attributeSplit.length === 1) {
+    return el.getComputedAttribute(componentName);
```
dmarcos: you can do ```return attributeValue```

src/core/a-animation.js
JohnRodney: right but attributeValue would be an object such as in the case of `material.color` value would be `material` and the return you need is actually the value of `color` on material.

src/core/a-animation.js
JohnRodney: In this case attributeValue is an object.

JohnRodney:
Cool those adjustments seem solid tests are still good.  Ping me if anything else gonna step off for a bit for dinner ect.

dmarcos:
Nice. Thanks for the patience :)

PR #1347, opened 2016-04-06
geometry caching system (fixes #1306)

### LINKED ISSUE #1306: cache geometries if we detect they have the same parameters
**Description:** Somewhat different from merging. Only one THREE.Geometry should be created from:

```html
<a-entity geometry="primitive: box; width: 5">
<a-entity geometry="primitive: box; width: 5">
```

Perhaps allow disabling:

```html
<a-entity geometry="primitive: box; width: 5">
<a-entity geometry="primitive: box; width: 5; cache: false">
```

Can utilize systems API to keep track.

ngokevin:
**Description:** 

- Uses **JSON.stringify** on component data to determine which geometry components should share geometries.

**Changes proposed:**
- Add **geometry system** that handles geometry creation, caching, and disposal
- Remove geometry.translate which is going to be removed in https://github.com/aframevr/aframe/pull/1339
- [x] docs
- [x] update tests

**Perf:** 

- In the number of entities stress test, reduced 4000 geometries to just 1.
- Reduced memory usage from 85MB to just 60MB. Previously without buffer geometries, it was at 150MB!
- Cut scene load time to 1/3 (e.g., reduced by 66%). From 6-7 seconds to 2-3 seconds.

screen shot 2016-04-06 at 3 23 29 pm

### DIFF
--- a/src/systems/geometry.js
+++ b/src/systems/geometry.js
@@ -1,0 +1,133 @@
+var geometries = require('../core/geometry').geometries;
+var registerSystem = require('../core/system').registerSystem;
+var THREE = require('../lib/three');
+
+/**
+ * System for geometry component.
+ * Handle geometry caching.
+ *
+ * @member {object} cache - Mapping of stringified component data to THREE.Geometry objects.
+ * @member {object} cacheCount - Keep track of number of entities using a geometry to
+ *         know whether to dispose on removal.
+ */
+module.exports.System = registerSystem('geometry', {
+  init: function () {
+    this.cache = {};
+    this.cacheCount = {};
+  },
+
+  /**
+   * Reset cache. Mainly for testing.
+   */
+  clearCache: function () {
+    this.cache = {};
+    this.cacheCount = {};
+  },
+
+  /**
+   * Attempt to retrieve from cache.
+   *
+   * @returns {Object|null} A geometry if it exists, else null.
+   */
+  getOrCreateGeometry: function (data) {
+    var cache = this.cache;
+    var cachedGeometry;
+    var hash;
+
+    // Skip all caching logic.
+    if (data.skipCache) { return createGeometry(data); }
+
+    // Try to retrieve from cache first.
+    hash = this.hash(data);
+    cachedGeometry = cache[hash];
+    incrementCacheCount(this.cacheCount, hash);
+
+    if (cachedGeometry) { return cachedGeometry; }
+
+    // Create geometry.
+    cachedGeometry = createGeometry(data);
+
+    // Cache and return geometry.
+    cache[hash] = cachedGeometry;
+    return cachedGeometry;
+  },
+
+  /**
+   * Let system know that an entity is no longer using a geometry.
+   */
+  unuseGeometry: function (data) {
+    var cache = this.cache;
+    var cacheCount = this.cacheCount;
+    var geometry;
+    var hash = this.hash(data);
+
+    if (!cache[hash] || data.skipCache) { return; }
+
+    decrementCacheCount(cacheCount, hash);
+
+    // Another entity is still using this geometry. No need to do anything.
+    if (cacheCount[hash] > 0) { return; }
+
+    // No more entities are using this geometry. Dispose.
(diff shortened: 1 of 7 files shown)


ngokevin:
I'm looking into whether three.js already does this internally.

tests/components/geometry.test.js
```diff
@@ -150,128 +109,160 @@ suite('standard geometries', function () {
...
-    geometry = el.getObject3D('mesh').geometry;
-    assert.equal(geometry.type, 'TorusKnotGeometry');
-    assert.equal(geometry.parameters.radius, 1);
-    assert.equal(geometry.parameters.tube, 4);
-    assert.equal(geometry.parameters.radialSegments, 3);
-    assert.equal(geometry.parameters.tubularSegments, 4);
-    assert.equal(geometry.parameters.p, 5);
-    assert.equal(geometry.parameters.q, 6);
+
+    process.nextTick(function () {
```
dmarcos: why do you need next tick. it should not be necessary.

src/systems/geometry.js
```diff
@@ -0,0 +1,140 @@
...
+   * @returns {Object|null} A geometry if it exists, else null.
+   */
+  getOrCreateGeometry: function (data) {
+    var cache = this.cache;
+    var cacheCount = this.cacheCount;
+    var cachedGeometry;
+    var hash;
+
+    // Skip all caching logic.
+    if (data.skipCache) { return Promise.resolve(createGeometry(data)); }
```
dmarcos: Why is this asynchronous?

dmarcos: Asynchrony is horrible to deal with. I don't see why we need it

ngokevin: Mentioned in the description. In the scene with 4000 entities, each entity was looking up the cache, saw the cache was empty, and then started to create their own geometry object, resulting in 4000 geometry objects. I added the promises so when the first entity reaches the cache, it tells the rest that it's creating the object, and the rest of the entities wait.

Should we expect that if 4000 entities are initializing their geometry components, that they go one at a time serially?

ngokevin: I think the root cause of asynchrony is attachedCallback, it just trickles down from there forcing us to do this.

dmarcos: I don't understand the problem. there's not really parallelism in the browser. When the first entity initializes its geometry it will add an entry to the cache and the next 3999 will retrieve it from there. I'm pretty confident that this code can be done synchronous.

dmarcos: We should be extremely vigilant with asynchrony and only introduce it to hide latency (which is its real purpose) when strictly necessary.

ngokevin: I have a synchronous version that we can check out before I added promises: https://github.com/ngokevin/aframe/commit/e56211e27e91ad3a65d3df8a871698bd53b8dc0b

ngokevin: We can pair program on that tmr.

src/systems/geometry.js
```diff
@@ -0,0 +1,133 @@
...
+ * System for geometry component.
+ * Handle geometry caching.
+ *
+ * @member {object} cache - Mapping of stringified component data to THREE.Geometry objects.
+ * @member {object} cacheCount - Keep track of number of entities using a geometry to
+ *         know whether to dispose on removal.
+ */
+module.exports.System = registerSystem('geometry', {
+  init: function () {
+    this.cache = {};
```
dmarcos: Should we merge these two in a single structure of objects ```{ geometry: xxx, count: x }```. We can then retrieve the geometry and increase/decrease the count on a single access.

ngokevin: ok

ngokevin: Hm, actually, it complicates the code a little bit. Because now you have to check whether a key exists and always be creating an inner object if it doesn't. Adds bits of code everywhere. Also it's a bit more clear what the data structure looks like if separate. Is it okay to keep them separate?

ngokevin:
thanks

PR #1351, opened 2016-04-07
use `window.WebVRConfig` if already defined

cvan:


dmarcos:
Do we need to copy it? Can we just do: ```window.WebVRConfig = window.WebVRConfig || {}```

cvan:
> Do we need to copy it? Can we just do: ```window.WebVRConfig = window.WebVRConfig || {}```

well, the thing is some people might do certain overrides like `WebVRConfig = {FORCE_ENABLE_VR: true};` and then `TOUCH_PANNER_DISABLED: true` won't be set.

on second thought, maybe it's okay. we probably shouldn't be too prescriptive and require people to use our `WebVRConfig` options. I'll change.

Issue #1350, opened 2016-04-07
cache materials that are creating using the same component properties


ngokevin:
**Description:**  Like https://github.com/aframevr/aframe/issues/1306 but for materials.

ngokevin:
This depends on untangling it such that each shader instance is not attached to any particular entity. If we want to cache, then a shader/material could be shared by multiple entities.

ngokevin:
Steps:

1. Split material system into material system + texture system.
2. Pass `sceneEl` into Shader constructor so it has access to texture system.
3. Have material component talk to material system to create or retrieve from cache a material.
4. Have material system pub-sub entities to shaders to receive events with `Shader.emit`.

ngokevin:
This adds too much overhead. While it saves memory, it can negatively impact performance of material updates.

ngokevin:
Here is the attempt https://github.com/ngokevin/aframe/commit/44dbd6a7799c97dd211dbfa70ff14c19fed9d4e0

PR #1364, opened 2016-04-11
refactor material + texture system

ngokevin:
**Description:**

This started off as trying to cache materials (#1350), but I found that negatively impacted CPU performance, although it did save relatively a lot of memory. Now it's mostly a refactor with a couple enhancements.

**Changes proposed:**
- Split material system + texture system.
- Move handling of loaded texture to the shader. `standard` and `flat` shaders now handle setting `material.map` and the texture system should only worry about loading/caching the texture.
- Shader talks to texture system.
- Move material/shader creation to material system.
- Set `material.shader` to shader instance such that shader methods can be manually envoked.
- Utility function between `standard`/`flat` shader to load textures for `material.map`.
- Flatten texture cache. Cache is keyed off of serialized data like geometry component. The subset of data to use to make the key is determined by shader.
- Support `repeat` (and future texture properties) for videos.
- Fix unnecessary `` creation when loading img elements.
- Bump rStats to get program stats
- Namespace and link performance tests
- Add more tests for texture caching
- Slightly improve material component perf by not calling `utils.diff`

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -164,6 +164,12 @@
       <li><a href="test-visibility/">Visibility</a></li>
     </ul>
 
+    <h2>Performance</h2>
+    <ul class="links">
+      <li><a href="performance-animations/">Animations</a></li>
+      <li><a href="performance-entity-count/">Entity Count</a></li>
+    </ul>
+
     <h2>Primitives</h2>
 
     <ul class="links">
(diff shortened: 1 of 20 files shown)


ngokevin:
Need to rethink this. The material caching system negatively impacts the performance of updating materials.

dmarcos:
Can you break down this in smaller PRs. This massive patches go over my head.

dmarcos:
It's very very hard to review PRs with a lot of unrelated changes

PR #1365, opened 2016-04-12
Enable Chrome launcher in Karma

### LINKED ISSUE #986: Enable unit tests for Chrome (in Karma)
Might catch things like: https://github.com/aframevr/aframe/issues/983

donmccurdy:
This came up as part of my Controls refactor, thought I'd make a separate PR out of it.

Fixes #986.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test": "karma start ./tests/karma.conf.js",
     "test:firefox": "karma start ./tests/karma.conf.js --browsers firefox_latest",
     "test:chrome": "karma start ./tests/karma.conf.js --browsers Chrome",
-    "test:ci": "TEST_ENV=ci karma start ./tests/karma.conf.js --single-run",
+    "test:ci": "TEST_ENV=ci karma start ./tests/karma.conf.js --single-run --browsers firefox_latest",
     "version": "npm run dist"
   },
   "repository": "aframevr/aframe",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "aframe",
   "version": "0.2.0",
   "description": "Building blocks for the VR Web",
-  "homepage" : "https://aframe.io/",
+  "homepage": "https://aframe.io/",
   "main": "dist/aframe.js",
   "scripts": {
     "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -18,6 +18,8 @@
     "release:push": "npm login && npm publish && git push --follow-tags",
     "start": "npm run dev",
     "test": "karma start ./tests/karma.conf.js",
+    "test:firefox": "karma start ./tests/karma.conf.js --browsers firefox_latest",
+    "test:chrome": "karma start ./tests/karma.conf.js --browsers Chrome",
```
donmccurdy: Eh, is this useful? I guess you can already just do:

```
npm run test -- --browsers Chrome
```

ngokevin: I think it's worth abstracting.

src/components/sound.js
```diff
@@ -66,7 +66,12 @@ module.exports.Component = registerComponent('sound', {
 
   remove: function () {
     this.el.removeObject3D('sound');
-    this.sound.disconnect();
+    try {
```
dmarcos: Is there a way to know if the sound is connected or not instead of doing a try/catch?

donmccurdy: Nothing obvious in [THREE.Audio](https://github.com/mrdoob/three.js/blob/master/src/audio/Audio.js). Reaching into the object a bit, `this.sound.source.context.state === 'running'` looked promising but it's still throwing in that state. There's probably an event or something but i can't find much in the AudioNode spec. :neutral_face:

src/components/obj-model.js
```diff
@@ -58,12 +58,14 @@ module.exports.Component = registerComponent('obj-model', {
...
     objLoader.load(objUrl, function (objModel) {
       // Apply material.
-      var material = el.components.material.material;
-      objModel.traverse(function (child) {
-        if (child instanceof THREE.Mesh) {
-          child.material = material;
-        }
-      });
+      var material = el.components.material;
+      if (material) {
```
ngokevin: Do you think we have a bug? Material is a dependency so it should be initialized by now.

dmarcos: Yeah. Checking for it should not be necessary

ngokevin:
lgtm, we'd file a follow-up issue on that obj-model dep bug. can you squash?

src/components/obj-model.js
ngokevin: Filed it at https://github.com/aframevr/aframe/issues/1369 since this means it was an existing issue.

dmarcos: Can we fix this in a separate PR?

ngokevin: Yeah, I'm suggesting separate PR.

Issue #1346, opened 2016-04-06
Test and ensure A-Frame scenes work with Samsung's Gear VR browser


cvan:

luke-john:
https://github.com/aframevr/aframe/blob/57fc8cbfcd845b2aed547092b321c1f98145193f/src/utils/index.js#L116

To get it working I had this function return false.

Given more mobile devices may have webvr browsers in the future, this function may need renaming.

loving1:
Yes, we have an A/VR headset that has mobile on it already and are not able to use the head tracking. It's not cell phone based which also causes a problem. Any advice on being able to have the browser recognize our device as unique would be helpful and to also connect to the head tracker.

cvan:
Fixed by #1366. Thanks, @dbradleyfl!

Marchhill:
This appears to still be an issue with the samsung internet vr browser (it works fine with the oculus browser), it is set to mobile mode when it shouldn't be

dbradleyfl:
Looks like the isGearVR function isn't being used inside isMobile check anymore [here](https://github.com/aframevr/aframe/blob/739427d84099dd0d92b83dcdf878f032cc7fe81b/src/utils/device.js#L38). The [Gear VR User Agent String](http://developer.samsung.com/technical-doc/view.do?v=T000000203) should still match that regex so that doesn't seem to be the issue. I don't have time to test it right now, but I'd add an if statement under line 47 that sets _isMobile to false if isGearVR is true @Marchhill

PR #1366, opened 2016-04-12
GearVR browser should not be treated as a mobile browser by aframe be…

dbradleyfl:
**Description:**
Samsung GearVR Browser now supports WebVR so this browser should not be treated as a mobile browser by aframe.
**Changes proposed:**
- Check within the isMobile function if the browser isGearVR. If so isMobile should return false.
- Used same style as isIOS check

Tested all examples on Gear VR with Galaxy S7 and this change makes them work.

#1346

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -122,10 +122,10 @@ module.exports.isMobile = function () {
     if (isIOS()) {
       check = true;
     }
+    if (isGearVR()) {
+      check = false;
+    }
   })(navigator.userAgent || navigator.vendor || window.opera);
-  if (isGearVR()) {
-    check = false;
-  }
   return check;
 };
 
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -123,11 +123,18 @@ module.exports.isMobile = function () {
       check = true;
     }
   })(navigator.userAgent || navigator.vendor || window.opera);
+  if (isGearVR()) {
+    check = false;
+  }
   return check;
 };
 
 var isIOS = module.exports.isIOS = function () {
   return /iPad|iPhone|iPod/.test(navigator.platform);
+};
+
+var isGearVR = module.exports.isGearVR = function () {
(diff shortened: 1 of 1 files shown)


src/utils/index.js
```diff
@@ -123,13 +123,20 @@ module.exports.isMobile = function () {
       check = true;
     }
   })(navigator.userAgent || navigator.vendor || window.opera);
+  if (isGearVR()) {
```
dmarcos: I think, this should be inside the function above. Below the If isIOS statement.

dbradleyfl:
@dmarcos you're right. Moved it there and tested. Still works. Thanks.

src/utils/index.js
```diff
@@ -130,6 +133,10 @@ var isIOS = module.exports.isIOS = function () {
   return /iPad|iPhone|iPod/.test(navigator.platform);
 };
 
+var isGearVR = module.exports.isGearVR = function () {
+  return /SamsungBrowser.+Mobile VR/i.test(navigator.userAgent);
```
cvan: nice, thanks! I'd suggest also making a PR against https://github.com/borismus/webvr-polyfill/blob/99c2649/src/util.js#L184 and https://github.com/borismus/webvr-boilerplate/blob/ecd6afc/src/util.js#L22

vincentmorneau:
Tried it, but can't make it work. Gear VR won't let me "Enter VR" anymore, as if it's treated as a desktop browser. Any thoughts?

dbradleyfl:
@vincentmorneau did you "enable WebVR" on the GearVR Browser as mentioned [here](https://mail.mozilla.org/pipermail/web-vr-discuss/2016-April/001054.html)?

vincentmorneau:
@dbradleyfl Oh. No I didn't.

Now I have, but it took me a while to figure out I had to go to `internet://webvr-enable` inside the VR Browser. Thanks a lot, A-Frame works fine with Gear VR now!

PR #1368, opened 2016-04-12
sharpen "Enter VR" and orientation modals, use monospace instead of Fira because web fonts are not used when injected

cvan:
### before

image

image

<hr>

### after

image

image


src/style/aframe.css
```diff
@@ -28,10 +28,37 @@ body {
...
 }
 
+.a-enter-vr-modal,
+.a-orientation-modal {
+  font-family: Andale Mono, Courier New, monospace;
+}
+
+.a-enter-vr-modal {
+  font-size: 11px;
+  line-height: 17px;
```
cvan: might want to reduce `line-height` so the shortest height doesn't pretrude by a few pixels

dmarcos:
I looks better for sure.

PR #1199, opened 2016-03-19
A-Frame-flavoured Shadertoy™

dmarcos:

PR #1384, opened 2016-04-19
[WIP] template component (fixes #930)

### LINKED ISSUE #930: Introduce API to register templates for encapsulating entities (and with optional templating logic)
In 0.1.0 we had the concept of templates:

```html
<template is="a-template"
          element="a-image"
          width="1.75"
          height="1.75"
          opacity="1.0"
          src="">
  <a-entity geometry="primitive: plane;
                      width: ${width};
                      height: ${height}"
            material="shader: flat;
                      src: url(${src});
                      opacity: ${opacity};
                      side: double;
                      transparent: true">
  </a-entity>
</template>
```

Because our primary use case has been to _extend_ entities for common use cases, we agreed to rethink "templates" (formerly known as "mixins" - not to be confused with `<a-mixin>`s, heh). And by rethink I mean remove until we've proven all the interesting use cases (and there certainly are some).

So instead of providing a declarative means of encapsulating one or many entities and exposing it as a custom element by some name (via the `element` attribute) through a `<template is="a-template">`, we've decided to use "primitives" as a way of extending entities.

The "old" templates are now not a high priority, but they're something we should not forget about and revisit some time.

If we don't want to enable devs to define templates from inline HTML (or HTML imports), we could introduce a JS API à la `registerComponent` for registering templates. For inspiration, check out X-Tag's `xtag.register` (go to https://x-tag.github.io/ and check out the examples in `Docs > Content is King`).

<hr>

Relevant: PR #883

ngokevin:
**Description:**

Details at https://github.com/ngokevin/aframe-template-component

**Changes proposed:**
- Move template component to core.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -157,6 +157,7 @@
       <li><a href="test-pivot/">Pivot</a></li>
       <li><a href="test-raycaster/">Raycaster</a></li>
       <li><a href="test-shaders/">Shaders</a></li>
+      <li><a href="test-template/">Template</a></li>
       <li><a href="test-texture/">Texture</a></li>
       <li><a href="test-towers/">Towers</a></li>
       <li><a href="test-video/">Video</a></li>
--- a/examples/test-template/butterfly.template
+++ b/examples/test-template/butterfly.template
@@ -1,0 +1,23 @@
+<a-mixin id="wing" geometry="primitive: box; depth: .005; height: 1; width: 0.5"
+                   pivot="0.25 0 0"
+                   material="color: red" rotation="0 0 0"
+                   scale="0.5 0.625 0.5"></a-mixin>
+<a-mixin id="flap" attribute="rotation" direction="alternate" fill="forwards"
+                   repeat="indefinite"></a-mixin>
+
+<a-entity position="{{ position }}">
+  <a-animation attribute="rotation" to="0 360 0" fill="forwards" dur="3000"
+               easing="linear" repeat="indefinite"></a-animation>
+  <a-entity rotation="0 90 0" position="2.5 0 0">
+    <a-entity rotation="-45 0 90">
+      <a-entity mixin="wing">
+        <a-animation mixin="flap" to="0 120 10"></a-animation>
+      </a-entity>
+    </a-entity>
+    <a-entity rotation="-45 0 90">
+      <a-entity mixin="wing">
+        <a-animation mixin="flap" to="0 -120 10"></a-animation>
+      </a-entity>
+    </a-entity>
+  </a-entity>
+</a-entity>
--- a/examples/test-template/index.html
+++ b/examples/test-template/index.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Template Component</title>
+    <meta name="description" content="Template Component- A-Frame">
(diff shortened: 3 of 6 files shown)


dmarcos:
I'm not sure that the template should be part of the core.

ngokevin:
Maybe not. Though it's immensely useful. I use it in most my scenes, and we have a lot of examples that could be cleaned up with it. It's a useful testbed for testing dynamically injected entities, mixins, and animations. And there's potential for enabling sharing of reusable snippets/templates.

With that, what do you propose?

- Put it in the examples folder, like three.js
- Put it in the `extras/` folder, which is seeming like a place where we shove stuff that we're unsure about it.
- Keep it separate, and promote it elsewhere.

dmarcos:
It's opinionated. I would not pollute the examples with those opinions. They should be as vanilla as possible even if they are a bit uglier.

dmarcos:
I think this should be out of the core and stay as a third party thing.

ngokevin:
I'm fine with that. But if we do this, I'd like to close https://github.com/aframevr/aframe/pull/1199 to keep the story straight.

dmarcos:
Shader toy is my personal bet for the future and something I think will become increasingly important. Being able to create completely procedural scenes (with no geometries) is something I want a-frame to support out of the box to give a chance for people to discover and serendipitously stumble on it.

ngokevin:
I deleted my comment because I was still mulling over it. A lot of the code was Shader Toy agnostic, but I would keep Shader Toy-specific things (if any) out of it if we're going to keep things unopinionated. I'm gonna move discussion there.

dmarcos:
Shader toy is just an extra shader it does not change the perception what a-frame is about. Writing the examples using a template engine obscures the a-frame core message. Is it a template engine? Entity component? declarative 3D DOM? All of the above?

andreasplesch:
Since I am using the template component, I was following this. Let me just add that I agree that it makes sense to have no third party dependencies (other than three.js) such as templating engines in the "core" or "extras". But let me also add that it can only be beneficial to a-frame as a concept to keep an as open as possible mind about what it is about. Its strength is extensibility. Some people may want to use it for performance conscious web games, some for data visualization, some or many for quick to code VR experiences, some purely declaratively, some mostly imperatively or with an important procedural component. So the core message is already pointing to "all of the above", at least in my small mind, and that is a positive. I think a-frame is doing a good job at that as well.
What I feel should be part of the core is support of a higher than asset level of reuse, probably just reuse of what is referred to as snippets.

ngokevin:
I think the core message is declarative ECS, which **enables** all of the above. I do think it is immensely useful to reuse chunks of HTML; I use it in every scene. But whether it's part of the core, or part of a component, I still have the ability to do so.

Later, we will promote third-party components more on the homepage, and support lazy-loading from a registry of components. So the lines of what ships with A-Frame and what's third-party will blur.

dmarcos:
It would be indeed useful to showcase examples of aframe using whatever template engine, d3, react... but making explicit that those are AFRAME + something else that it's not aframe. We could have a section in aframe.io examples that points to those demos.

andreasplesch:
Glad to see that there was a deep exploration of where the boundaries ought to be at #930 . In the end, components may be just bundled with other components and/or with core based mostly on popularity. So it makes sense to me to keep core rather pure.

PR #1387, opened 2016-04-19
split system + material component

ngokevin:
**Description:**

Precursor to doing changes in https://github.com/aframevr/aframe/pull/1364

This is mostly moving code around

**Changes proposed:**
- Split material and texture system
- Rename textureCache -> cache


src/systems/template.js
```diff
@@ -0,0 +1,11 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * Template system.
+ *
+ * Handles lazy-loading template engines, template compilation, caching of compiled templates.
+ */
+module.exports.System = registerSystem('template', {
```
dmarcos: I think we don't need this file

src/utils/material.js
```diff
@@ -10,16 +10,16 @@ module.exports.updateMap = function (shader, data) {
   var el = shader.el;
   var src = data.src;
   var material = shader.material;
-  var materialSystem = el.sceneEl.systems.material;
+  var textureSystem = el.sceneEl.systems.texture;
```
dmarcos: Why making texture a system and not keeping it as utils library?

src/systems/template.js
ngokevin: oops

src/utils/material.js
ngokevin: It's a bit more consistent to use systems when we want components to access some sort of global scope. Also would be nice to inspect the cache ```sceneEl.systems.texture.cache```.

dmarcos: I'm really not sure about it. Why not making all the utils systems then? Keeping the 1:1 relationship between component and system is important . 1 component type should be paired with one system or you will be coupling the state of components of different types via those shared systems (scopes). Systems should not be something you can require or there's no difference between a system and a library.

ngokevin: The only reason I can think of is that a system is specific to a single A-Frame scene, and a utility module is global among all A-Frame scenes.

I'll keep everything in the material system then. I think that doesn't hurt.

PR #1388, opened 2016-04-19
don't use object diff if not necessary

ngokevin:
**Description:** For material + sound, full object diffing was not necessary. Saves some CPU.

**Changes proposed:**
- Don't use object diff in material/sound.
- Compare only properties needed
- make `oldData` to be `undefined` for single prop schemas on first init

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -1,5 +1,4 @@
 var debug = require('../utils/debug');
-var diff = require('../utils').diff;
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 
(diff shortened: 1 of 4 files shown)


src/components/sound.js
```diff
@@ -24,44 +23,33 @@ module.exports.Component = registerComponent('sound', {
...
   update: function (oldData) {
     var data = this.data;
-    var diffData = diff(oldData || {}, data);
     var el = this.el;
     var sound = this.sound;
-    var src = data.src;
-    var srcChanged = 'src' in diffData;
+    var srcChanged;
+
+    oldData = oldData || {};
```
dmarcos: should we always pass empty object on update if there's not a value?

ngokevin: Yeah, I think so. Previously, it was `null` so that we could determine whether it was the first update. But that could be done instead by just setting a flag from `init`. Having to check for `null` gets in the way more than it is useful.

Separate PR? Or this one.

dmarcos: where do check for null value to detect first update? Is still necessary?

ngokevin: I don't think we do. That might have been the original intent. But now it's better to use a flag for that.

dmarcos: let's do an empty object for empty data and add the flag later if there's need. if a conponent wants to do something at inititiization the method init should be used

ngokevin: Yeah, by flag, I meant use `init` and store a variable.

Actually, you already changed it in the refactor PR. It's an empty object at the beginning now. I don't have to do anything here.

ngokevin: Err nvm...going to need to double check that.

ngokevin:
OK updated.

PR #1389, opened 2016-04-19
share updateMap helper between flat/standard shaders

ngokevin:
**Description:**

Precursor to https://github.com/aframevr/aframe/pull/1364 where the shader will be managing the setting of textures on the material. And to emphasize that `map` is just one of many possible textures.

**Changes proposed:**
- Move map updating code to a utility function since shader/flat are the exact same code


dmarcos:
These are rare these days but it always make me happy see a patch that removes more lines than adds :)

PR #1247, opened 2016-03-24
bump changelog

ngokevin:
**Changes proposed:**
- Replaced everything with dashes (sorry in advance for the diff mess).
- Added new changelog items at the bottom of each section.

PR #1394, opened 2016-04-21
organize examples (fixes #1247)

ngokevin:
**Changes proposed:**

- Folders.
- Move asset into the folder of the example they are being used in.
- Put very-commonly used assets in the root `assets/` folder.
- Remove unused assets.
- Remove unnecessary example (i.e., `controls`) that are prevalent throughout all the other examples.
- `s/stats="true"/stats/g`
- Use asset management system when possible


dmarcos:
I want to keep mixin and controls examples. I use them all the time for testing and I've caught more than one bug with them.

dmarcos:
They are also useful to illustrate the feature with no other noise

ngokevin:
I won't argue for mixin. Do you really need controls? Every single example will have controls guaranteed.

dmarcos:
It makes me happy to see all those underscore directory names gone :)

Issue #1372, opened 2016-04-13
Repeat parameter for material doesn't work once the texture is already created


fernandojsg:
**Description:** `repeat` property from from material can't be used dynamically as once the `src` map is loaded, the texture won't be recreated based on a new `repeat` value.

feiss:
I found the same issue, thought it was an inspector problem:
https://github.com/aframevr/aframe-inspector/issues/462

sunyatasattva:
Just wondering if there is any news on this issue @ngokevin since the label was removed.

PR #1399, opened 2016-04-22
make shaders handle setting textures

ngokevin:
**Description:** Make texture caching more agnostic than forcing setting of `material.map`.

Before the flow was:

1. Shader passes entity and material, and asks material system for texture.
2. Material system creates texture, caching with `{src: {repeat: {}}}` data structure.
3. Material system emits event on `el`.
4. Material system sets `material.map`

The new flow is:

1. Shader passes subset of data and callback, and asks material system for texture.
2. Material system creates texture, caching with JSON-hashed data for images.
3. Material system sends texture through callback.
4. The shader does whatever it wants with the material and handles notifying the entity.

Because we hash the data for cache key, we get to remove a lot of code from the previously complex data structure.

**Other Changes proposed:**
- If passing in an `img` element, use it rather than recreate another `img` element.
- Combine video texture video element hash with data when hashing
- Add more tests

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -1,47 +1,52 @@
-var srcLoader = require('./src-loader');
-
 /**
- * Update shader instance's `material.map` given `data.src`.
+ * Update `material.map` given `data.src`. For standard and flat shaders.
  *
  * @param {object} shader - A-Frame shader instance.
  * @param {object} data
  */
 module.exports.updateMap = function (shader, data) {
   var el = shader.el;
-  var src = data.src;
   var material = shader.material;
-  var materialSystem = el.sceneEl.systems.material;
+  var src = data.src;
 
   if (src) {
-    if (src === shader.mapSrc) { return; }
+    if (src === shader.textureSrc) { return; }
     // Texture added or changed.
-    shader.mapSrc = src;
-    srcLoader.validateSrc(
-      src,
-      function loadImageCb (src) { materialSystem.loadImage(el, material, data, src); },
-      function loadVideoCb (src) { materialSystem.loadVideo(el, material, data, src); }
-    );
+    shader.textureSrc = src;
+    el.sceneEl.systems.material.loadTexture(src, {src: src, repeat: data.repeat}, setMap);
     return;
   }
 
   // Texture removed.
-  updateMaterialTexture(material, null);
+  if (!material.map) { return; }
+  setMap(null);
+
+  function setMap (texture) {
+    material.map = texture;
+    material.needsUpdate = true;
+    handleTextureEvents(el, texture);
+  }
 };
 
 /**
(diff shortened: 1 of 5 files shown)


fernandojsg:
Cool, I'll take a look this afternoon on the editor. By the way, will you include here the fix for the https://github.com/aframevr/aframe/issues/1372 ? I just look around the commit and didn't see any reference to it

ngokevin:
Not in this PR. But we can take a look after.

fernandojsg:
I got the editor working with this PR. As we talked it's not as clean as before but it's not a big deal.

dmarcos:
@fernandojsg how this PR makes it not **as clean**?

ngokevin:
The editor side is not as clean because it has to loop through the texture cache and inspect `texture.image` to grab the `src`s.

fernandojsg:
I was talking with @ngokevin on slack that it's better to keep the core simple as possible even if for the editor some functionality will need to do some _hacky tricks_ to access to that info. Just in the case that we'll find a important use case outside from the editor we could go and refactor or expose that functionality in a more useful way. So right now it's not a problem.

ngokevin:
r??

src/utils/material.js
```diff
@@ -1,47 +1,52 @@
...
-  if (texture) { texture.needsUpdate = true; }
-  material.map = texture;
+function handleTextureEvents (el, texture) {
+  if (!texture) { return; }
 
-  // Only need to update three.js material if presence or not of texture has changed.
-  if (oldMap === null && material.map || material.map === null && oldMap) {
-    material.needsUpdate = true;
-  }
+  el.emit('material-texture-loaded', {src: texture.image, texture: texture});
```
dmarcos: consistency. We need to commit to one way of naming events. sometimes we use dashes to separate words and sometimes we don't. For instance: componentchanged or loadeddata.

dmarcos: the dom does not use dashes for event names so I would say we should stick to the standard

dmarcos: https://developer.mozilla.org/en-US/docs/Web/Events

ngokevin: To which one? We also sometimes use past vs present tense.

ngokevin: True. Our event names are a bit longer though, are we okay with like `materialvideoloadeddata`?

dmarcos: The standard DOM does not use dashes. I personally like past tense. The standard DOM is inconsistent on that because you see gamepadconnected, loadeddata. voiceschanged and also languagechange. chargingchange

dmarcos: The dom has also pretty long event names like DOMNodeInsertedIntoDocument or resourcetimingbufferfull

ngokevin: k

PR #1405, opened 2016-04-23
Minor edits to the ECS doc page

bryik:
**Description:** Aside from adding a missing "to", I have slightly rephrased a sentence in the concept section.

**Changes proposed:**
- Add missing "to"
- Rephrase: 

From: "These components would be able to be mixed and matched and even be used with other vehicles..."

To: "These components could be mixed and matched and even used with other vehicles"


ngokevin:
thanks

PR #1406, opened 2016-04-23
Add Texture Position and Wrapping control

bengfarrell:
**Description:**
Control a texture's position (offset) and wrap. My use case was that I wanted to put a transparent texture on a sphere at a specific scale and location. To do this the ThreeJS way, it meant using repeat for scaling down the texture, an alternate form of wrapping so that the texture doesn't actually repeat, and the offset for controlling where the texture goes.

**Changes proposed:**
- Expose "offset" as material property as a space separated list of length 2 (default "0 0")
- Expose wrap as material property as a space separated list of length 2. Values are strings and are can be valid ThreeJS wrapping keys. For less verboseness on the user's end, "RepeatWrapping" and "ClampToEdgeWrapping" is not valid, but "Repeat" and "ClampToEdge" is (the code adds "Wrapping" automatically).
- If wrapping is not set (there is no default), and repeat is used, there is no change to AFrame behavior. Wrapping won't be set unless using repeat - and if repeat is used RepeatWrapping is also used
- If wrapping IS set, the new user defined wrapping will be used regardless of if repeat is utilized or not


src/systems/material.js
```diff
@@ -223,17 +228,32 @@ function loadImageTexture (material, src, repeat) {
      * Texture loaded. Set it.
      */
     function createTexture (texture) {
-      var repeatXY;
+      var repeatXY, offsetXY, wrapST;
```
dmarcos: we declare one variable per line.

src/systems/material.js
```diff
@@ -39,6 +39,9 @@ module.exports.System = registerSystem('material', {
    */
   loadImage: function (el, material, data, src) {
     var repeat = data.repeat || '1 1';
+    var offset = data.offset || '0 0';
+    var wrap = data.wrap;
+
```
dmarcos: We need to add this variables to the schema of the flat and standard shaders (if applies)

dmarcos: We should probably have two variables wrapS and wrapT

src/systems/material.js
```diff
@@ -223,17 +228,32 @@ function loadImageTexture (material, src, repeat) {
...
+      if (offsetXY.length === 2) {
+        texture.offset.set(offsetXY[0], offsetXY[1]);
+      }
+
       // Handle UV repeat.
       repeatXY = repeat.split(' ');
       if (repeatXY.length === 2) {
-        texture.wrapS = THREE.RepeatWrapping;
-        texture.wrapT = THREE.RepeatWrapping;
+        if (!wrap) {
```
dmarcos: if you declare this variable in the schema yo always have one defined and this if won't be necessary since THREE.RepeatWrapping will be the default

ngokevin:
Thanks for PR! Unfortunately, this heavily collides with https://github.com/aframevr/aframe/pull/1399 because of the way texture loading and caching is changed. Let's re-open after that lands.

https://github.com/aframevr/aframe/pull/1399 will make it easier to add more texture properties.

bengfarrell:
No worries! I'm really just learning AFrame, and this was one of the first things I tried to do, saw it wasn't possible and dug in. I wasn't thrilled with the way I did it cause I'd forsee more and more props being stuffed in here, so I'm glad its being rethought.

PR #1415, opened 2016-04-27
fix dynamically-attached entities not playing due to load conditions (fixes #1383)

### LINKED ISSUE #1383: injected animations not working
**Description:** 

Caught this with template component. Creating an animation with JS and appending to scene, animation does not run:

http://codepen.io/team/mozvr/pen/qZYGxj

ngokevin:
**Description:**

Caught this using the template component that dynamically attached entities that also had the template component to dynamically attach entities.

1. Entity is attached.
2. Entity tries to `play()`
3. Entity not yet `loaded`
4. Entity never plays because not loaded (early return statement)

**Changes proposed:**
- Animations wait for entity to both load + play before playing.
- Ensure entities call `play()` after they are loaded.
- Remove unnecessary `.bind` calls.

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -43,20 +43,20 @@ module.exports.AAnimation = registerElement('a-animation', {
     attachedCallback: {
       value: function () {
         var self = this;
-        var el = self.el = self.parentNode;
+        var el = this.el = this.parentNode;
 
-        if (el.isNode) {
-          if (el.hasLoaded) {
-            init();
-          } else {
-            el.addEventListener('loaded', init.bind(self));
-          }
-        } else {
-          // To handle elements that are not yet `<a-entity>`s (e.g., templates).
-          el.addEventListener('nodeready', init.bind(self));
-        }
+        init();
 
         function init () {
+          if (!el.isPlaying) {
+            el.addEventListener('play', init);
+            return;
+          }
+          if (!el.hasLoaded) {
+            el.addEventListener('loaded', init);
+            return;
+          }
+
           self.applyMixin();
           self.update();
           self.load();
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -59,13 +59,14 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  /**
+   * Add to parent, load, play.
+   */
   attachedCallback: {
     value: function () {
       this.addToParent();
-      if (!this.isScene) {
-        this.load();
-        if (this.parentEl.isPlaying) { this.play(); }
-      }
+      if (this.isScene) { return; }
+      this.load();
     }
   },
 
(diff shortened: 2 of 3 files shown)


src/core/a-animation.js
```diff
@@ -43,20 +43,20 @@ module.exports.AAnimation = registerElement('a-animation', {
...
 
-        if (el.isNode) {
-          if (el.hasLoaded) {
-            init();
-          } else {
-            el.addEventListener('loaded', init.bind(self));
-          }
-        } else {
-          // To handle elements that are not yet `<a-entity>`s (e.g., templates).
-          el.addEventListener('nodeready', init.bind(self));
```
dmarcos: we don't listen to the nodeready event anymore?

ngokevin: Those were leftover from `<a-template>`.

src/core/a-entity.js
```diff
@@ -59,13 +59,22 @@ var proto = Object.create(ANode.prototype, {
...
       this.addToParent();
-      if (!this.isScene) {
-        this.load();
-        if (this.parentEl.isPlaying) { this.play(); }
+      if (this.isScene) { return; }
+
+      // Play if parent is playing.
+      if (this.parentNode.isPlaying) {
+        this.addEventListener('loaded', function () {
+          this.play();
```
dmarcos: what about moving ```this.play()``` to the end of the ```load``` function?

src/core/a-entity.js
```diff
@@ -426,7 +435,7 @@ var proto = Object.create(ANode.prototype, {
       var sceneEl = this.sceneEl;
 
       // Already playing.
-      if (this.isPlaying || !this.hasLoaded) { return; }
+      if (this.isPlaying) { return; }
```
dmarcos: I think we still need this check if you do:
````javascript
el.play();
````

**Comment by ngokevin** [comment]
@dmarcos r?

File: src/core/a-entity.js [hunk 4]
```diff
@@ -211,15 +212,26 @@ var proto = Object.create(ANode.prototype, {
...
-        ANode.prototype.load.call(this, this.updateComponents.bind(this));
-      } else {
-        ANode.prototype.load.call(this, this.updateComponents.bind(this),
-                                  function (el) { return el.isEntity; });
+        ANode.prototype.load.call(this, sceneLoadCallback);
+        return;
+      }
+
+      // Entity load.
+      function entityLoadCallback () {
```
dmarcos: Can we move this function to the bottom of the function after ``` ANode.prototype.load.call(this, entityLoadCallback, isEntity);```

ngokevin: any reason to? seems more consistent with doing vars-on-top since these functions get hoisted. and readability-wise, it makes sense to want to know what this function is doing in the context of the method.

dmarcos: The function gets in the way of understanding the main body of the function. I want to see what the function does furst and maybe later want to see what happens asynchronously when the entity loads.

dmarcos: I'll change this later in a follow up PR

ngokevin: the function is integral to understanding what the function does though. it's only separated because it's awkward to inline

PR #1417, opened 2016-04-28
[doc fix] a-curvedimage component/value table is reverted

gasolin:
**Description:**

doc for a-curvedimage component/default value table is reverted
https://aframe.io/docs/primitives/a-curvedimage.html

And some default values are not correspondent to the real value
https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives/a-curvedimage.js

**Changes proposed:**
- reverse the component/default value
- fix default value according to real value


dmarcos:
thanks

Issue #1333, opened 2016-04-04
change default animation easing function `linear`


ngokevin:
**Description:** While on 2D website transitions you want easing most of the time, it seems like games/3D prefer linear transitions. Every time I do an animation, I have to specify `easing="linear"` since I've never found myself wanting to using easing.

ngokevin:
wontfix https://github.com/aframevr/aframe/pull/1421#issuecomment-215512324

PR #1421, opened 2016-04-28
change default animation easing method as linear #1333

gasolin:
**Changes proposed:**
- set default animation type as 'linear' https://github.com/aframevr/aframe/blob/master/src/constants/animation.js#L15
- update doc accordingly

### DIFF
--- a/docs/core/animations.md
+++ b/docs/core/animations.md
@@ -29,7 +29,7 @@ Many attributes and values are used to define animations. We'll delve into more 
 | begin     | Delay (in milliseconds) or event name to wait on before beginning animation.                                           | 0              |
 | direction | Direction of the animation (between `from` and `to`). One of `alternate`, `alternateReverse`, `normal`, `reverse`.     | normal         |
 | dur       | Duration in (milliseconds) of the animation.                                                                           | 1000           |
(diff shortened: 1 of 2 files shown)


ngokevin:
Defaults live at https://github.com/aframevr/aframe/pull/1421/files#diff-cc37208f3a65c67afe166320bf095ca6R10

ngokevin:
@dmarcos Approve the idea? I've built a bunch of scenes, I almost always want `linear` over `ease`.

dmarcos:
The conventional wisdom is that animations should always have some easing because it's how natural movement occurs in real world. Linear animations usually feel unnatural. Look at how biological systems move and operate and how that influence animations in UIs and games. @ngokevin why did you prefer linear over ease in your experiments?

ngokevin:
I've used animations for:

- Falling. A cheap way for gravity.
- Continuous movement. Like a butterfly flying around a try, or something running, or like a plane/car. Or a laser. Something on a conveyor belt.

But those are probably exceptions, and even then they would probably have some miniscule easing. so I agree. Linear is unnatural. Sorry for the goose chase, @gasolin

PR #1426, opened 2016-04-30
faster for loop with length caching

gasolin:
**Description:**

Apply JS best practice of for loop by length caching

Will increase performance when attributes or objects are increasing

http://stackoverflow.com/questions/5349425/whats-the-fastest-way-to-loop-through-an-array-in-javascript

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -25,13 +25,14 @@ module.exports = registerElement('a-assets', {
         var imgs = this.querySelectorAll('img');
         var timeout = parseInt(this.getAttribute('timeout'), 10) || 3000;
         var videos = this.querySelectorAll('video');
+        var i, len;
 
         if (this.parentNode.tagName !== 'A-SCENE') {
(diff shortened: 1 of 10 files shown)


dmarcos:
This does not make the code harder to maintain (for loops leak an extra variable) and a bit more brittle? If you for example have two nested for loops. You need two variables and make sure you use the appropriate one. I've quickly checked backbone and react (just picking two random popular libraries) and they don't use that form of for loops. Why is that?

https://github.com/facebook/react/blob/master/src/renderers/dom/client/ReactDOMTreeTraversal.js#L120

https://github.com/jashkenas/backbone/blob/master/backbone.js#L145

Also most of the Stack Overflow discussion you pointed out is from 2011. Are the claimed perf gains still relevant in the JS VMs of 2016?

gasolin:
Thanks for the question. More recent result shows normal for loop is better, http://stackoverflow.com/questions/17989270/for-loop-performance-storing-array-length-in-a-variable

At least I learned micro-optimization seems harmful :/

PR #1427, opened 2016-04-30
Removed `cursor-` prefixes (not firing in 0.2.x)

maxkrieger:
This is just a fix which worked for me -- but it seems that all `cursor`-related events are now firing without the `cursor-` prefix as of `0.2.x`. For example `cursor-mouseenter` is now working as `mouseenter`.

If this is completely wrong, feel free to close this, but if not, the relevant doc has been updated in this PR.

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -13,7 +13,7 @@ The cursor component lets us interact with entities through clicking and gazing.
 - Emits special mouse and hover events (e.g., relating to mouse down/up/enter/leave).
 - Has additional states for hovering.
 
-When the mouse is clicked, the closest visible entity intersecting the cursor, if any, will emit a `cursor-click` event. Note the cursor component only applies the raycasting behavior. To provide a shape or appearance to the cursor, you could apply the [geometry][geometry] and [material][material] components.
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks! The convention we were trying to introducing for event names was: componentname-eventname. @ngokevin why do you think the new events are not working as expected?

maxkrieger:
Ope, should have opened an issue I guess

ngokevin:
I wanted to remove the prefixes too. I think we can make these events standard.

dmarcos:
I see that the cursor component in master indeed emits events of the form cursor-eventname . The problem is that the  published docs don't reflect 0.2.0 but master. Someone reported that the color animations don't work as stated in the docs https://aframe.io/docs/core/animations.html#Color-Attributes. That's a feature that recently landed in master @cvan are we recently republished the docs? We should deploy 0.2.0 tag instead of master

ngokevin:
Yeah, we've been deploying master documentation. I agree, I can make it go back to the 0.2.0 tag and I can put a dropdown to view old/master docs.

dmarcos:
The current docs correctly reflect the API. The problem is that the published version on the website corresponds to master instead of the latest public release 0.2.0. I'm closing this and I filed https://github.com/aframevr/aframe/issues/1428 to take care of the issue

xzvfinet:
Hi, I've been looking through your showcase and just realized your showcase source doesn't have the latest version. It still has cursor- prefix. 
Thanks for the great library btw.

PR #1430, opened 2016-05-03
Fix issue building when installed as a dependency with npm3

sjando:
**Description:**

The way three.js examples (loaders, controls, effects) are currently required with an explicit node_modules path breaks the ability to build aframe from source when aframe is installed as a dependency with npm3 (flat dependency structure). This scenario should be supported given requiring from source instead of prebuilt is sometimes more appropriate (e.g. avoiding duplicate THREE object in prebuilt file if another three dependency exists in the project outside of aframe)

**Changes proposed:**
- Do not provide explicit node_modules path, allowing browserify/webpack/etc to find the appropriate three location

### DIFF
--- a/src/lib/three.js
+++ b/src/lib/three.js
@@ -19,10 +19,10 @@ if (THREE.Cache) {
 }
 
 // TODO: Eventually include these only if they are needed by a component.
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks!

Issue #1431, opened 2016-05-03
Emit event when component has been removed


mayognaise:
**Emit event when a component has been removed.**

example: `el.emit('componentremoved')`

https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L314

- A-Frame Version: All
- Platform/Device: All

ngokevin:
https://github.com/aframevr/aframe/pull/1434/files

PR #1434, opened 2016-05-03
Emit event when component has been removed

mayognaise:
This PR is for this issue:
https://github.com/aframevr/aframe/issues/1431

Additionally it's been attached `name` of component

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -312,6 +312,7 @@ var proto = Object.create(ANode.prototype, {
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
+      this.emit('componentremoved');
     }
   },
 
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -312,7 +312,7 @@ var proto = Object.create(ANode.prototype, {
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
-      this.emit('componentremoved');
+      this.emit('componentremoved', { name: name });
     }
   },
 
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -56,6 +56,16 @@ suite('a-entity', function () {
     parentEl.appendChild(el);
   });
 
+  test('emits `componentremoved` event when element itself has been removed', function (done) {
+    var el = this.el;
+    el.setAttribute('geometry', 'primitive:plane');
+    el.addEventListener('componentremoved', function (event) {
+      assert.equal(event.detail.name, 'geometry');
+      done();
+    });
+    el.removeAttribute('geometry');
+  });
+
   suite('attachedCallback', function () {
     test('initializes 3D object', function (done) {
       var el = entityFactory();


src/core/a-entity.js
```diff
@@ -312,6 +312,7 @@ var proto = Object.create(ANode.prototype, {
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
+      this.emit('componentremoved');
```
donmccurdy: Might be helpful to attach the name of the component, as well –

```javascript
this.emit('componentremoved', {name: 'name'});
```

mayognaise: @donmccurdy great idea let me add it.

ngokevin:
Can you write a quick unit test for this in `a-entity.test.js`? Thanks!

mayognaise:
@ngokevin hmm actually the test hasn't been passed.. the event is not called :/ do you think it's because it's already removed?

ngokevin:
Don't need to create a Mesh, you should set a random component, use `removeAttribute` on it, and check the event detail has the component name

mayognaise:
@ngokevin using `entityFactory();`?

ngokevin:
In the test suite, there should already be a `setup` that calls `entityFactory` for you.

`removeAttribute` detaches components, which is what you're testing here. `removeChild` would detach all components I guess, but it's simpler to test with `removeAttribute`.

ngokevin:
ex:

```js
  test('emits `componentremoved` event when element itself has been removed', function (done) {
    var el = this.el;
    el.setAttribute('geometry');
    el.addEventListener('componentremoved', function (evt) {
      assert.equal(evt.detail.name, 'geometry');
      done();
    });
    el.removeAttribute('geometry');
  });
```

mayognaise:
ohh, oh i see 
so this doesn’t mean when the `a-entity` has been removed from dom, correct?

ngokevin:
yes just testing component detach

mayognaise:
@ngokevin i see. `componentremoved` hasn't been fired.. 😭

ngokevin:
`el.setAttribute('geometry', '');` fixed this line

ngokevin:
you might want something for detachedCallback, but this PR is ok too

ngokevin:
I'll check this PR out later and try to add a proper test.

mayognaise:
@ngokevin the test has been passed 🎉

src/core/a-entity.js
```diff
@@ -312,6 +312,7 @@ var proto = Object.create(ANode.prototype, {
       pauseComponent(component, this.sceneEl);
       component.remove();
       delete this.components[name];
+      this.emit('componentremoved', { name: name });
```
dmarcos: should we pass the entity as well?

dmarcos: not sure if we should pass it. it would be done this way:
```this.emit('componentremoved', { el: this, name: name });```

Comment (mayognaise) [inline src/core/a-entity.js]: @dmarcos sure let me add `el`

Comment (donmccurdy) [inline src/core/a-entity.js]: would it already be available as `event.target`? that seems like enough if so, otherwise maybe helpful to add it.

Comment (mayognaise) [inline src/core/a-entity.js]: ah true, let me check

Comment (mayognaise) [inline src/core/a-entity.js]: This has passed. so maybe dont we need to add it?
```js
  test('emits `componentremoved` event when element itself has been removed', function (done) {
    var el = this.el;
    el.setAttribute('geometry', 'primitive:plane');
    el.addEventListener('componentremoved', function (event) {
      assert.equal(event.target, el);
      assert.equal(event.detail.name, 'geometry');
      done();
    });
    el.removeAttribute('geometry');
  });
```

dmarcos:
Nice work. @mayognaise thanks!

PR #1437, opened 2016-05-04
Graph based behavior execution scheduling and post processing support

wizgrav:
Following the discussions with @ngokevin @dmarcos @donmccurdy and @fernandojsg :

Enable post processing by extending components with a .tock() method which runs after the scene renders.

A-scene maintains a .renderTarget and switches rendering to that in the presence of at least one tock behavior.

A graph based mechanism is employed to allow for explicitly definable execution ordering for both tick and tock behaviors.

Components can define two way connections in the graph using the properties .precedents and .following, similar to .dependencies.

Sorting to define the execution order takes place only in registerComponent() and batching is done efficiently in a-scene using caches.

The graph based mechanism works fast and comfortable for the developer. Cyclic dependencies in the graph are errors and throw exceptions.

An example utilizing most of the functionality added by this PR at http://wizgrav.github.io/aframe/examples/test/post-processing/index.html

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -229,6 +229,7 @@
       <li><a href="test/video/">Video</a></li>
       <li><a href="test/videosphere/">Video 360&deg;</a></li>
       <li><a href="test/visibility/">Visibility</a></li>
+      <li><a href="test/post-processing/">Post Processing</a></li>
     </ul>
 
     <h2>Primitives</h2>
--- a/examples/test/post-processing/components/bloom.js
+++ b/examples/test/post-processing/components/bloom.js
@@ -1,0 +1,119 @@
+/* global AFRAME THREE */
+
+/*
+Kawase bloom, described here
+https://software.intel.com/en-us/blogs/2014/07/15/an-investigation-of-fast-real-time-gpu-based-image-blur-algorithms
+*/
+
+AFRAME.registerComponent('bloom', {
+  dependencies: ['post-process'],
+  following: true, // boolean true is shortcut for reusing the  definition of the dependencies property.
+
+  schema: {
+    passes: { default: '-0.5 0 1 2 2 3' },
+    intensity: { default: 1.0 }
+  },
+
+  setupPostState: function () {
+    this.scenePost = new THREE.Scene();
+
+    // The camera is not actually used.
+    this.cameraPost = new THREE.OrthographicCamera(1, 1, 1, 1, 1, 1);
+
+    this.quadPost = new THREE.Mesh(new THREE.PlaneBufferGeometry(2, 2), null);
+    this.scenePost.add(this.quadPost);
+  },
+
+  init: function () {
+    this.setupPostState();
+    this.quadPost.material = new THREE.ShaderMaterial({
+      uniforms: {
+        srcTexture: { type: 't', value: null },
+        resolution: {type: 'v2', value: new THREE.Vector2(0, 0)},
+        step: { type: 'v2', value: new THREE.Vector2(0, 0) },
+        threxp: { type: 'v2', value: new THREE.Vector2(0, 0) }
+      },
+      vertexShader: this.vertexText,
+      fragmentShader: this.fragmentText,
+
+      depthWrite: false
+
+    });
+    var rtOptions = { minFilter: THREE.LinearFilter, magFilter: THREE.NearestFilter, format: THREE.RGBAFormat };
+    this.renderTargets = [
+      new THREE.WebGLRenderTarget(1, 1, rtOptions),
+      new THREE.WebGLRenderTarget(1, 1, rtOptions)
+    ];
+    this.el.emit('post-modified');
+  },
+
+  remove: function () {
+    this.el.emit('post-modified');
+  },
+
+  update: function () {
+    this.postIntensity = this.data.intensity;
+  },
+
+  // We're doing bloom at half resolution for performance. Most blurry effects can get away with that.
+  getTargets: function () {
+    var srt = this.el.sceneEl.renderTarget;
+    for (var i = 0; i < 2; i++) {
+      var rt = this.renderTargets[i];
+      if (rt.width !== srt.width / 2 || rt.height !== srt.height / 2) {
+        rt.setSize(srt.width / 2, srt.height / 2);
+      }
+    }
+    return this.renderTargets;
+  },
+
+  tock: function () {
+    var rts = this.getTargets();
+    var el = this.el;
+    var scene = el.sceneEl;
+    var uns = this.quadPost.material.uniforms;
+    var rt = scene.renderTarget;
+
+    // Ping pong the render targets progressively blurring more on each pass.
+    var passes = this.data.passes.split(' ');
+    for (var i = 0; i < passes.length; i++) {
+      var ps = +(passes[i]) + 0.5;
+      uns.step.value.set(ps, ps);
+      uns.srcTexture.value = rt;
+      if (!i) uns.threxp.value.set(16, 0); else uns.threxp.value.set(1, 0);
+      rt = rts[i & 1];
+      uns.resolution.value.set(rt.width, rt.height);
+
+      scene.renderer.render(this.scenePost, this.cameraPost, rt);
+    }
+    this.renderTarget = rt;
+  },
+
+  vertexText: [
+    'uniform vec2 resolution;',
+
+    'void main() {',
+    '    gl_Position =  vec4(( position.xy ) * resolution, 1.0, 1.0 );',
+    '}'
+  ].join('\n'),
+
+  fragmentText: [
+    'uniform sampler2D srcTexture;',
+    'uniform vec2 resolution;',
+    'uniform vec2 threxp;',
+    'uniform vec2 step;',
+
+    'void main(){',
+    '    vec2 uv = gl_FragCoord.xy/resolution;',
+    '    vec2 st = step/resolution;',
+    '    vec4 color = vec4(0.0);',
+
+    '   color += texture2D(srcTexture,uv + st);',
+    '    color += texture2D(srcTexture,uv + vec2(st.x,-st.y));',
+    '    color += texture2D(srcTexture,uv - vec2(st.x,-st.y));',
+    '    color += texture2D(srcTexture,uv - st);',
+    '    color /= 4.0;',
+    '    gl_FragColor = vec4(pow(color.rgb,vec3(threxp.x))+vec3(threxp.y), 1.0);',
+    '}'
+  ].join('\n')
+});
--- a/examples/test/post-processing/components/compositor.js
+++ b/examples/test/post-processing/components/compositor.js
@@ -1,0 +1,198 @@
+/* global AFRAME THREE */
+
+// The compositor. Blends the blurred renderTarget from the bloom component.
+// It also checks opacity and converts parts of the frame to ascii(Alpha masking).
+AFRAME.registerComponent('post-process', {
+
+  init: function () {
+    this.setupPostState();
+    this.postModified();
+
+    this.el.sceneEl.renderTarget.depthTexture = new THREE.DepthTexture();
+    this.el.sceneEl.renderTarget.stencilBuffer = false;
+    this.el.sceneEl.enablePostProcessing = true;
+
+    this.evh = this.postModified.bind(this);
+    this.el.sceneEl.addEventListener('post-modified', this.evh);
+  },
+
+  setupPostState: function () {
+    this.scenePost = new THREE.Scene();
+
+    // The camera is not actually used.
+    this.cameraPost = new THREE.OrthographicCamera(1, 1, 1, 1, 1, 1);
+
+    this.quadPost = new THREE.Mesh(new THREE.PlaneBufferGeometry(2, 2), null);
+    this.scenePost.add(this.quadPost);
+  },
+
+  postModified: function (ev) {
+    this.needsConfig = true;
+  },
+
+  // Whenever the postproc setup changes, the compositor shader is recompiled
+  config: function () {
+    var behaviors = this.el.sceneEl.behaviors.tock('bloom');
+    var bloom = behaviors.length ? behaviors[0] : null;
+    console.log('POST-MODIFIED', behaviors);
+    this.needsConfig = false;
+
+    var uniforms = {
+      srcTexture: { type: 't', value: this.el.sceneEl.renderTarget },
+      resolution: {type: 'v2', value: new THREE.Vector2(0, 0)}
+    };
+    var defines = {};
+    var scene = this.el.sceneEl;
+    if (bloom) {
+      uniforms.bloomTexture = { type: 't', value: null };
+      uniforms.bloomIntensity = { type: 'f', value: bloom.postIntensity };
+      defines['BLEND_BLOOM'] = 1;
+      this.bloomComponent = bloom;
+    }
+
+    uniforms.cameraNear = { type: 'f', value: scene.camera.near };
+    uniforms.cameraFar = { type: 'f', value: scene.camera.far };
+
+    if (scene.renderTarget.depthTexture) {
+      uniforms.depthTexture = { type: 't', value: null };
+      defines['BLEND_DEPTH'] = 1;
+    }
+
+    this.quadPost.material = new THREE.ShaderMaterial({
+      defines: defines,
+      uniforms: uniforms,
+      vertexShader: this.vertexText,
+      fragmentShader: this.fragmentText,
+
+      depthWrite: false
+
+    });
+  },
+
+  // Traverse scene objects and set opacity based on the postOpacity attribute set by the pulse component.
+  tick: function (time) {
+    this.el.sceneEl.object3D.traverse(function (obj) {
+      var mat = obj.material;
+      if (mat && mat.postOpacity !== undefined) {
+        mat.opacity = mat.postOpacity;
+      }
+    });
+    var renderTarget = this.el.sceneEl.renderTarget;
+    if (!renderTarget.depthTexture) {
+      renderTarget.depthTexture = new THREE.DepthTexture();
+      renderTarget.stencilBuffer = false;
+    }
+  },
+
+  // Cleanup the scene objects opacities.
+  remove: function () {
+    this.el.sceneEl.object3D.traverse(function (obj) {
+      var mat = obj.material;
+      if (mat && mat.postOpacity !== undefined) {
+        mat.opacity = 1;
+      }
+    });
+    this.el.sceneEl.removeEventListener('post-modified', this.evh);
+  },
+
+  tock: function () {
+    var el = this.el;
+    var scene = el.sceneEl;
+    var rt = scene.renderTarget;
+
+    if (this.needsConfig) {
+      this.config();
+    }
+
+    var uns = this.quadPost.material.uniforms;
+    uns.resolution.value.set(rt.width, rt.height);
+
+    if (uns.bloomTexture) {
+      uns.bloomTexture.value = this.bloomComponent.renderTarget;
+    }
+
+    if (uns.depthTexture) {
+      uns.depthTexture.value = scene.renderTarget.depthTexture;
+    }
+
+    scene.renderer.render(this.scenePost, this.cameraPost);
+  },
+
+  vertexText: [
+    'uniform vec2 resolution;',
+
+    'void main() {',
+    '    gl_Position =  vec4(( position.xy ) * resolution, 1.0, 1.0 );',
+    '}'
+  ].join('\n'),
+
+  fragmentText: [
+    'uniform sampler2D srcTexture;',
+    'uniform vec2 resolution;',
+    'uniform float cameraNear;',
+    'uniform float cameraFar;',
+
+    '#if defined BLEND_BLOOM',
+    'uniform sampler2D bloomTexture;',
+    'uniform float bloomIntensity;',
+    '#endif',
+
+    '#if defined BLEND_DEPTH',
+    'uniform sampler2D depthTexture;',
+
+    'float readDepth (vec2 uv) {',
+    '  float cameraFarPlusNear = cameraFar + cameraNear;',
+    '  float cameraFarMinusNear = cameraFar - cameraNear;',
+    '  float cameraCoef = 2.0 * cameraNear;',
+    '  return cameraCoef / (cameraFarPlusNear - texture2D(depthTexture, uv).x * cameraFarMinusNear);',
+    '}',
+    '#endif',
+
+    '// Bitmap to ASCII (not really) fragment shader by movAX13h, September 2013',
+    '// --- This shader is now used in Pixi JS ---',
+
+    'float character(float n, vec2 p) // some compilers have the word "char" reserved',
+    '{',
+    '  p = floor(p*vec2(4.0, -4.0) + 2.5);',
+    '  if (clamp(p.x, 0.0, 4.0) == p.x && clamp(p.y, 0.0, 4.0) == p.y)',
+    '  {',
+    '    if (int(mod(n/exp2(p.x + 5.0*p.y), 2.0)) == 1) return 1.0;',
+    '  }	',
+    '  return 0.0;',
+    '}',
+
+    'void main()',
+    '{',
+    '  vec2 uv = gl_FragCoord.xy/resolution;',
+    '  vec4 col = texture2D(srcTexture, floor(gl_FragCoord.xy/8.0)*8.0/resolution);	',
+    '  ',
+    '  float gray = (col.r + col.g + col.b)/3.0;',
+    '  ',
+    '  float n = 65536.0;       // .',
+    '  if (gray > 0.2) n = 65600.0;  // :',
+    '  if (gray > 0.3) n = 332772.0;  // *',
+    '  if (gray > 0.4) n = 15255086.0; // o ',
+    '  if (gray > 0.5) n = 23385164.0; // &',
+    '  if (gray > 0.6) n = 15252014.0; // 8',
+    '  if (gray > 0.7) n = 13199452.0; // @',
+    '  if (gray > 0.8) n = 11512810.0; // #',
+    '  float split = 0.33;',
+    '  #if defined BLEND_DEPTH',
+    '  split = smoothstep(0.3,0.9,readDepth(uv));',
+    '  #endif',
+
+    '  vec2 p = mod(gl_FragCoord.xy/4.0, 2.0) - vec2(1.0);',
+    '  vec4 ocol = texture2D(srcTexture, uv );',
+    '  vec3 acol = ((1.0-split) * col.rgb * character(n, p) + split * ocol.rgb);',
+
+    '  vec3 color = mix(ocol.rgb,acol.rgb,ocol.a);',
+    '  ',
+
+    '  #if defined BLEND_BLOOM',
+    '  color += texture2D(bloomTexture,uv).rgb * bloomIntensity;',
+    '  #endif',
+
+    '  gl_FragColor = vec4(color,1.0);',
+    '}'
+  ].join('\n')
+});
--- a/examples/test/post-processing/index.html
+++ b/examples/test/post-processing/index.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Post-processing</title>
+    <meta name="description" content="Post-processing — A-Frame">
+    <script src="../../../dist/aframe.js"></script>
+    <script src="components/pulse.js"></script>
+    <script src="components/envmap.js"></script>
+    <script src="components/bloom.js"></script>
+    <script src="components/compositor.js"></script>   
+    <script src="components/bloom-toggler.js"></script>
+  </head>
+  <body>
+    <a-scene title="right click to toggle bloom" stats bloom="" bloom-toggler="">
+      <a-assets>
+        <video id="video" src="https://ucarecdn.com/bcece0a8-86ce-460e-856b-40dac4875f15/"
+               autoplay loop></video>
+      </a-assets>
+      <a-entity position="0 1.8 5">
+        <a-entity camera="far:10000; near:10;" look-controls></a-entity>
+      </a-entity>
+      <a-sphere envmap="" pulse="phase:1.58" rotation="0 180 0" color="white" position="0 -50 -160" radius="40"></a-sphere>
+      <a-videosphere pulse="phase:0; exponent:5;" src="#video" rotation="0 180 0"></a-videosphere>
+    </a-scene>
+  </body>
+</html>
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -441,8 +441,8 @@ var proto = Object.create(ANode.prototype, {
    */
   play: {
     value: function () {
-      var components = this.components;
-      var componentKeys = Object.keys(components);
+      var entityComponents = this.components;
+      var componentKeys = Object.keys(entityComponents);
       var sceneEl = this.sceneEl;
 
       // Already playing.
@@ -451,7 +451,7 @@ var proto = Object.create(ANode.prototype, {
 
       // Wake up all components.
       componentKeys.forEach(function _playComponent (key) {
-        playComponent(components[key], sceneEl);
+        playComponent(entityComponents[key], sceneEl);
       });
 
       // Tell all child entities to play.
@@ -695,6 +695,13 @@ function isComponentMixedIn (name, mixinEls) {
 }
 
 /**
+ * Checks if a component has defined a method that needs to run every frame.
+ */
+function hasBehavior (component) {
+  return component.tick || component.tock;
+}
+
+/**
  * Pause component by removing tick behavior and calling pause handler.
  *
  * @param component {object} - Component to pause.
@@ -703,7 +710,9 @@ function isComponentMixedIn (name, mixinEls) {
 function pauseComponent (component, sceneEl) {
   component.pause();
   // Remove tick behavior.
-  if (!component.tick) { return; }
+  if (!hasBehavior(component)) {
+    return;
+  }
   sceneEl.removeBehavior(component);
 }
 
@@ -716,7 +725,9 @@ function pauseComponent (component, sceneEl) {
 function playComponent (component, sceneEl) {
   component.play();
   // Add tick behavior.
-  if (!component.tick) { return; }
+  if (!hasBehavior(component)) {
+    return;
+  }
   sceneEl.addBehavior(component);
 }
 
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -68,6 +68,15 @@ Component.prototype = {
   tick: undefined,
 
   /**
+   * Tock handler.
+   * Called after the base scene renders. Used for post processing purposes.
+   *
+   * @param {number} time - Scene tick time.
+   * @param {number} timeDelta - Difference in current render time and previous render time.
+   */
+  tock: undefined,
+
+  /**
    * Called to start any dynamic behavior (e.g., animation, AI, events, physics).
    */
   play: function () { /* no-op */ },
@@ -267,12 +276,19 @@ module.exports.registerComponent = function (name, definition) {
   components[name] = {
     Component: NewComponent,
     dependencies: NewComponent.prototype.dependencies,
+    precedents: NewComponent.prototype.precedents,
+    following: NewComponent.prototype.following,
     parse: NewComponent.prototype.parse,
     parseAttrValueForCache: NewComponent.prototype.parseAttrValueForCache,
     schema: utils.extend(processSchema(NewComponent.prototype.schema)),
     stringify: NewComponent.prototype.stringify,
     type: NewComponent.prototype.type
   };
+
+  // Force a resorting on the next .getOrder('component:execution') call.
+  utils.precedence.setGraph('component:execution');
+  // The graph for dependencies will get sorted when getOrder(...) is called.
+  utils.precedence.setGraph('component:dependency');
   return NewComponent;
 };
 
@@ -346,3 +362,23 @@ function extendProperties (dest, source, isSinglePropSchema) {
   if (isSinglePropSchema) { return source; }
   return utils.extend(dest, source);
 }
+
+// Setup a graph for the order of component execution. Only tick/tock components are taken into account.
+utils.precedence.setGraph('component:execution', {
+  map: components,
+  collect: function (self, name) {
+    var item = self.map[name];
+    if (!item) { return {name: ' '}; }
+    if (item.Component.prototype.tick || item.Component.prototype.tock) {
+      return { incoming: item.precedents, outgoing: item.following };
+    }
+  }
+});
+
+// Setup a graph for calculating the dependency order. To be sorted. What can we use this order for?
+utils.precedence.setGraph('component:dependency', {
+  map: components,
+  collect: function (self, name) {
+    return { incoming: self.map[name].dependencies };
+  }
+});
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -56,7 +56,11 @@ module.exports = registerElement('a-scene', {
 
     init: {
       value: function () {
-        this.behaviors = [];
+        this.behaviors = {
+          tick: utils.precedence.newStore('component:execution', Array),
+          tock: utils.precedence.newStore('component:execution', Array)
+        };
+        this.precedence = utils.precedence; // Expose getOrder() to components, useful for post proc.
         this.hasLoaded = false;
         this.isPlaying = false;
         this.originalHTML = this.innerHTML;
@@ -111,15 +115,24 @@ module.exports = registerElement('a-scene', {
       }
     },
 
-    /**
+     /**
      * @param {object} behavior - Generally a component. Must implement a .update() method to
      *        be called on every tick.
      */
     addBehavior: {
       value: function (behavior) {
+        var self = this;
         var behaviors = this.behaviors;
-        if (behaviors.indexOf(behavior) !== -1) { return; }
-        behaviors.push(behavior);
+        var name = Object.getPrototypeOf(behavior).name || ' '; // unnamed components are used in tests.
+
+        // Check if behavior has tick and/or tock and add to the appropriate structure.
+        Object.keys(behaviors).forEach(function (behaviorType) {
+          if (!behavior[behaviorType]) { return; }
+          var behaviorArr = self.behaviors[behaviorType](name); // eg. behaviors.tick("material")
+          if (behaviorArr.indexOf(behavior) === -1) {
+            behaviorArr.push(behavior);
+          }
+        });
       }
     },
 
@@ -168,10 +181,17 @@ module.exports = registerElement('a-scene', {
      */
     removeBehavior: {
       value: function (behavior) {
+        var self = this;
         var behaviors = this.behaviors;
-        var index = behaviors.indexOf(behavior);
-        if (index === -1) { return; }
-        behaviors.splice(index, 1);
+        var name = Object.getPrototypeOf(behavior).name || ' ';
+
+        Object.keys(behaviors).forEach(function (behaviorType) {
+          var behaviorArr = self.behaviors[behaviorType](name);
+          var index = behaviorArr.indexOf(behavior);
+          if (index !== -1) {
+            behaviorArr.splice(index, 1);
+          }
+        });
       }
     },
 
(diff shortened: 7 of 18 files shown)


donmccurdy:
This is really neat! Nice example.

As far as the API, it feels strange having DOM events hooked into the render loop, firing every 8-16ms. I'm not sure exactly what to recommend though. Other ideas:

```javascript
// 1.
sceneEl.setPostProcessingCallback(function () { /* ... */ });

// 2.
sceneEl.addPostProcessingCallback(function () { /* ... */ });

// 3.
AFRAME.registerComponent('my-component', {
  prerender: function () { /* called only if my-component is attached to sceneEl */ },
  postrender: function () { /* called only if my-component is attached to sceneEl */ },
});
```

wizgrav:
@donmccurdy you proposals #1/#2 where also my original thoughts. I went for emit because it seemed to be more consistent with the patterns used in the library. I'm almost certain that performance shouldn't degrade with dom events. Of course this needs investigation but 8/16ms is still a good chunk

wizgrav:
Hmm on the other hand I'm not sure if the events are dispatched synchronously

dmarcos:
Thanks for tackling this. I don't think emitting events from the render loop is a very good idea. It adds overhead (and most of the time you won't even use it) and compromises performance. Having a way to do postprocessing effects would be indeed very cool. I like more @donmccurdy approach of providing two methods at the component level. You can currently insert your logic in the render loop via the `tick` function of components and systems. It would be nice to have a way to throw warnings or stats on console if  any those user defined functions in the render loop are too slow.

src/core/system.js
```diff
@@ -33,6 +33,16 @@ System.prototype = {
...
+    
+  /**
+   * Post handler.
+   * Called after each iteration of the scene render loop.
+   * Affected by play and pause.
+   *
+   * @param {number} time - Scene tick time.
+   * @param {number} timeDelta - Time it took to render the scene.
+   */
+  tick: undefined,
```
ngokevin: We probably shouldn't be using the `system` API to handle post-processing. Systems are sort of global managers for components. We should just use the plain component API.

ngokevin: I don't think we need Systems to have a hook into post-render callback yet.

src/core/component.js
```diff
@@ -66,7 +66,17 @@ Component.prototype = {
...
+  
+  /**
+   * Post handler.
+   * Called after each iteration of the scene render loop.
+   * Affected by play and pause.
+   *
+   * @param {number} time - Scene tick time.
+   * @param {number} timeDelta - Time it took to render the scene.
+   */
+  post: undefined,
```
ngokevin: Let's call it `postRender` to be more specific.

ngokevin:
Can you remove the `dist` bumps?

examples/boilerplate/360-video-post/index.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
```
ngokevin: Can you move this example to `examples/test/post-processing` and rename the titles/metatags to match?

examples/boilerplate/360-video-post/index.html
```diff
@@ -0,0 +1,117 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>360 Video</title>
+    <meta name="description" content="360 Video — A-Frame">
+    <script src="../../../dist/aframe.js"></script>
+    <script id="vertexShader" type="x-shader/x-vertex">
+
```
ngokevin: A lot of whitespace. Can one-line this one:

```html
<script>
  void main() { gl_Position ...
</script>
```

examples/boilerplate/360-video-post/index.html
```diff
@@ -0,0 +1,117 @@
...
+      <a-assets>
+        <video id="video" src="https://ucarecdn.com/bcece0a8-86ce-460e-856b-40dac4875f15/"
+               autoplay loop></video>
+      </a-assets>
+
+      <a-videosphere src="#video" rotation="0 180 0"></a-videosphere>
+    </a-scene>
+    
+    <script>
+        AFRAME.registerComponent('effect', {
```
ngokevin: Can you put this in the `<head>` so that it is available before the scene runs?

ngokevin:
Hmm, after going over the PR, I'm not sure about piggybacking off the component API. It assumes `postRender` might have a wider use case than just post-processing. Maybe we need a tighter API for register post-processing effects. Thinking about it...

wizgrav:
based on @dmarcos comments I added a post method on components. Behaviors are split in tick and post types and handled separately. Performance has indeed improved noticeably(from 50% -> 40% on my box for the example) so this seems like the optimal path to take.

I also added a postPriority property on components. This is intended to facilitate common post proc setups where several components will produce their own renderTargets containing bloom, godrays, ssao etc and then a composer component can pick these up and blend the to screen. Prioritizations happens only on post behaviors(which should only be a handful) and sorting happens only once per frame if there was a change in the post pipeline.

Finally an event is emitted when the sorting happens to indicate that the pipeline has changed. This can be picked up by the composer components so they can reconfigure themselves. I'm not sure about this one maybe a simple timestamp could work. What are your thoughts on this, and the whole pr in general?

wizgrav:
@ngokevin sorry for the mess with dist and such, I was having great pain with npm and setting up the build in general. I'll fix everything

wizgrav:
I also included the tweaked VREffect inside lib/three.js until it is fixed in three.js itself. I see no reason why they shouldn't do the tweaks themselves. The VREffect render() should match the signature of the base WebGLRenderer since it's a wrapper of it. Another note of interest is the use of depth textures. I think it recently landed on three and fortunately they implemented it as a property on the renderTarget so when aframe upgrades to the latest three.js all that will be needed to utilize depth textures will be

`this.renderTarget.depthTexture = new THREE.DepthTexture()`

and of course resizing it along with the renderTarget itself

ngokevin:
I don't like using the component API too much. We are adding a special handler and property that from what I can tell would only be used for post-processing, and then we have to touch play/pause code, and such.

Brainstorming, but here's an alternative API, which would be harder to implement, but would be a bit more explicit. Using the System API, this would be a bit similar to Don's Controls API:

```
AFRAME.registerPostProcessor('some-effect', {
  schema: {},
  vertexShader: 'vertexshaderstring',
  fragmentShader: 'fragmentshaderstring'
});

<a-scene postprocess="bloom, hdr, ssao" bloom="intensity: 0.5>
```

wizgrav:
@ngokevin I'm not sure a postproc effect can be described like a shader, it's more of a setup. For instance bloom commonly uses a separable gaussian blur so it would need at least two passes. Good bloom would also entail downsampling several times, blurring, and blending again on the way up. I think components make for a natural expression even if meant to be attached on the scene only. But I don't know I haven't yetlooked over all of aframe's source and patterns so you'd know best

wizgrav:
@ngokevin another thing that should be taken into account while brainstorming is that customisation logic ends up in the composer usually. Postproc steps that provide intermediate renderTargets like bloom and ssao tend to be straightforward and so could be cast in place as reusable components. Only the composer needs special treatment, both for doing the composition of all the rest and also to apply the simple effects like vignette aberations, ascii etc At least for production one shouldn't use extra passes for these

ngokevin:
Ah, right, thanks for the details. I should clarify the sample API I had would still make post-processing effects a component, but just a subset API of a component. Just because `postRender` + `postRenderPriority` would only be used by effects, seems to make sense to fence it off.

wizgrav:
Yeah I understand your concerns. Defining post specific components seems to be the cleaner route then unless there would be a reason for post components to also run some logic at tick time. To be honest though I cannot think of such a case right now, I need to think about it

wizgrav:
Ok so I tried to think of cases where having a component tap into both tick and postRender and I have indeed came up with something. I've updated the example which you can see on http://wizgrav.github.io/aframe/examples/test/post-processing/

The alpha channel of the render target can be used to pass information to the post processing chain, the most common case being masking. In the example the pulse component sets up properties on its underlying mesh. The postproc component then traverses the scene and adjusts the opacity according to these properties. It also handles cleanup.

Alternatively the pulse component itself could setup it's opacity by itself but it would need to know the state of post processing. It could do that by listening for something like the "post-modified" event.

I'm not arguing in favor of either approach, it would be convenient to have both options availble.

I would also like to mention something, and I know you're going to hate me @ngokevin , but it really needs to be said. It is inevitable that components will need to be extended. At some point, a component will need to access the results of computations performed on another component on another entity. A way to guarantee the order of execution will be needed. I'm sorry if I've missed something, I still study the source but I think there isn't something like that in currently.

Doing something like postRenderPriority may work for postproc because it's by nature limited in scale but for tick behaviors I don't think it's appropriate at all. The solution unity came up for this situation was to add a LateUpdate method which runs after all update methods. So aframe will also need a tock() to go along with tick(). Again excuse me if I'm missing something. This does not concern this PR, it's something else all together but it's relevant and I feel that it should be taken into account

wizgrav:
Following the line of thought from the previous comment, if a tick()/tock() pair is indeed desirable, the api could get some symmetry with a postTick()/postTock() pair for the post processing, getting postRenderPriority out of the picture. Two lanes for the post processing should cover most cases. The third fxaa step I mentioned could be integrated in the composer itself in postTock(). So that setup could also work

wizgrav:
@ngokevin @dmarcos Added tock() on components as a straight equivalent of unity's LateUpdate. An equivalent pair is implemented for post render as postRenderTick, postRenderTock. These should be enough to guarantee the order of execution so postRenderPriority was removed. The post-modified event is still emitted as it is useful for reconfiguring the compositor when the post chain changes.

The example at http://wizgrav.github.io/aframe/examples/test/post-processing/ was commented and updated to reflect all functionality in the PR. Some crude bloom was added to highlight the tick/tock pattern, you can toggle it by right clicking.

ngokevin:
Thanks, will take a look when I can

ngokevin:
Can you remove the `dist` files from the PR?

examples/test/post-processing/index.html
```diff
@@ -0,0 +1,309 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Post Process</title>
+    <meta name="description" content="Post process — A-Frame">
+    <script src="../../../dist/aframe.js"></script>
+    
+    <script>
+        AFRAME.registerComponent('pulse', {
```
ngokevin: Let's register these components in separate files: e.g., `components/pulse-effect.js`, `components/envmap.js`, `components/bloom.js`

wizgrav:
Hmm, I split the example but I'm not sure I did the right thing with the dist files. I did a checkout on the last commit before I touched them. Can you suggest how could they get out of the PR completely?

There's also the issue with VREffect being included in lib/three.js which is not so good. I can try making a PR to three so they can fix it on their source but to be honest, requiring such a core part of aframe from an examples/ folder kind of makes me nervous. Since you control the revision of three included anyway, maybe you should consider including VREffect.js in aframe's source somewhere.

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,141 @@
...
+  vertexText: [
+    "uniform vec2 resolution;",
+        
+    "void main() {",
+    "    gl_Position =  vec4(( position.xy ) * resolution, 1.0, 1.0 );",
+    "}"
+  ].join("\n"),
+    
+  fragmentText: [
+    "uniform sampler2D srcTexture;",
```
ngokevin: nice, thanks

ngokevin:
- `dist/` files are out and look OK now.
- Yeah, that's not great. But @dmarcos just pulled them into `vendor/` for WebVR 1.0 API reasons. We still wouldn't want to touch them. Maybe we need that module to live in its own home?

ngokevin:
Needs a rebase. I'm not sure how we should handle the `VREffect` change collisions. Ideally, all the changes would be upstreamed to three.js, but maybe we should just move it to `lib/` and maintain it for now...

wizgrav:
Yes I agree. Do you want me to do it here?

examples/test/post-processing/index.html
```diff
@@ -0,0 +1,36 @@
...
+      <a-assets>
+        <video id="video" src="https://ucarecdn.com/bcece0a8-86ce-460e-856b-40dac4875f15/"
+               autoplay loop></video>
+      </a-assets>
+      <a-sphere envmap="" pulse="phase:1.58" rotation="0 180 0" color="white" position="0 -50 -160" radius="40"></a-sphere>
+      <a-videosphere pulse="phase:0; exponent:5;" src="#video" rotation="0 180 0"></a-videosphere>
+    </a-scene>
+      
+    <script>
+        var scene = document.querySelector("a-scene");
```
ngokevin: This can be a scene component like `components/bloom-toggler.js`

examples/test/post-processing/index.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Post Process</title>
```
ngokevin: "Post-processing" here and below would match better

src/core/a-entity.js
```diff
@@ -683,7 +683,9 @@ function isComponentMixedIn (name, mixinEls) {
 function pauseComponent (component, sceneEl) {
   component.pause();
   // Remove tick behavior.
-  if (!component.tick) { return; }
+  if (!component.tick && !component.tock && !component.postRenderTick && !component.postRenderTock) {
```
ngokevin: Hm, maybe a separate function

```js
function hasBehavior (component) {
  return component.tick...;
}
```

src/core/scene/a-scene.js
```diff
@@ -118,8 +124,15 @@ var AScene = module.exports = registerElement('a-scene', {
     addBehavior: {
       value: function (behavior) {
         var behaviors = this.behaviors;
-        if (behaviors.indexOf(behavior) !== -1) { return; }
-        behaviors.push(behavior);
+        for (var b in behaviors) {
```
ngokevin: We usually use `Object.keys().forEach` pattern for object enumeration, which I read to be a bit faster.

ngokevin: avoid 1-letter var names: s/b/behaviorType

src/core/a-entity.js
wizgrav: Yeap this line has gotten way too long

src/core/scene/a-scene.js
```diff
@@ -118,8 +124,15 @@ var AScene = module.exports = registerElement('a-scene', {
...
+          var bs = behaviors[b];
```
ngokevin: s/bs/behaviorTypeArr or behaviorArr or something

src/core/system.js
```diff
@@ -34,6 +34,16 @@ System.prototype = {
    */
   tick: undefined,
 
+ /**
```
ngokevin: Do we need tock support for systems yet?

examples/test/post-processing/index.html
wizgrav: Not sure I got that, you mean renaming the file and component or something else?

src/core/scene/a-scene.js
```diff
@@ -57,7 +57,13 @@ var AScene = module.exports = registerElement('a-scene', {
 
     init: {
       value: function () {
-        this.behaviors = [];
+        this.behaviors = {
+          tick: [],
+          tock: [],
+          postRenderTick: [],
```
ngokevin: Maybe `tick`, `tock`, `tickPostRender`, `tockPostRender`?

src/core/scene/a-scene.js
```diff
@@ -317,10 +395,33 @@ var AScene = module.exports = registerElement('a-scene', {
...
+
+          this.postRenderTick(time, timeDelta);
+          this.postRenderTock(time, timeDelta);
+        } else {
+          this.renderer.render(this.object3D, camera, null);
         }
-        this.renderer.render(this.object3D, camera);
 
         this.time = time;
+
```
ngokevin: extra line

src/core/scene/a-scene.js
```diff
@@ -317,10 +395,33 @@ var AScene = module.exports = registerElement('a-scene', {
 
         if (this.isPlaying) {
           this.tick(time, timeDelta);
+          this.tock(time, timeDelta);
+        }
+
+        if (this.behaviors.postRenderTick.length > 0 || this.behaviors.postRenderTock.length > 0) {
```
ngokevin: need the `> 0`?

src/core/scene/a-scene.js
```diff
@@ -317,10 +395,33 @@ var AScene = module.exports = registerElement('a-scene', {
...
+          // Post processing setup has been modified, emit a notification.
+          if (this.postNeedsNotify) {
+            var bs = this.behaviors;
+            this.postNeedsNotify = false;
+            this.emit('post-modified', { tick: bs.postRenderTick, tock: bs.postRenderTock });
+          }
+
+          this.renderer.render(this.object3D, camera, this.renderTarget);
+
+          this.time = time;
```
ngokevin: Can put all the `time` line next to each other:

```
this.time...
time = ...
timeDelta = ...
```

examples/test/post-processing/index.html
ngokevin: I mean move the whole `script` logic to a component

wizgrav: Ok got it you mean the last script on bottom, yeap will fix that

src/core/scene/a-scene.js
wizgrav: I don't mind with either option. Your call

src/core/scene/a-scene.js
wizgrav: not really, should I take it out?

src/core/scene/a-scene.js
ngokevin: Probably go with the option I had to namespace it

src/core/scene/a-scene.js
ngokevin: yeah

src/core/scene/a-scene.js
wizgrav: On a related note @dmarcos mentioned that he'd wanted a means to check if post processing takes too long. Should we take another timestamp with performance.now() and I don't know spew some debug with that here  or should that be done inside a component in postRenderTock?

ngokevin: As long as the performance of non-post-processed scenes don't degrade, it's OK.

wizgrav: it won't even run if there's not post processing but what kind of debug should we output? I don't like the idea of outputing on the console every frame. Three.js has a stats object inside where it collects various metrics like draw calls etc Maybe we should do something similar?

ngokevin: Try out `<a-scene stats>` and be amazed

ngokevin:
Bunch of `npm run lint` errors

src/core/scene/a-scene.js
wizgrav: awesome I totally missed that, so should I just window.performance.mark it?

wizgrav:
the bulk of lint errors come from lib/three.js cause of the inclusion of VREffect.js, I'll take it on it's own file out as we said but should we ignore it in lint checks?

ngokevin:
Yeah, still need to resolve it with the VREffect in `vendor/`, but there'll be a lot of lint errors for `a-scene`, and all the example JS files.

src/core/scene/a-scene.js
wizgrav: forEach makes me weary as it would create a closure but I'm not really sure how the browsers will optimise it. I went on to make a test case in jsperf but it seems their site is down for maintenance. I'll try again later

ngokevin: I've read `for...in``` makes hasOwnProperty calls for each key. But that's just from skimming SO

wizgrav: The body of the function is so short that I think in either case it will be optimised to the same native code. I'll switch to forEach for consistency

wizgrav:
@ngokevin  Rebased and squashed. Did also the minor changes needed for renderTarget support in @dmarcos  VREffect.js in vendor/

vendor/VREffect.js
```diff
@@ -232,7 +232,7 @@ THREE.VREffect = function ( renderer, onError ) {
 	var cameraR = new THREE.PerspectiveCamera();
 	cameraR.layers.enable( 2 );
 
-	this.render = function ( scene, camera ) {
+	this.render = function ( scene, camera, renderTarget, forceClear ) {
```
ngokevin: Can we upstream this to three.js?

src/lib/three.js
```diff
@@ -26,3 +26,4 @@ require('../../vendor/VRControls');  // THREE.VRControls
 require('../../vendor/VREffect');  // THREE.VREffect
 
 module.exports = THREE;
+
```
ngokevin: whiteline

src/core/scene/a-scene.js
```diff
@@ -312,10 +397,36 @@ module.exports = registerElement('a-scene', {
...
+          this.tickPostRender(time, timeDelta);
+          this.tockPostRender(time, timeDelta);
+          window.performance.mark('post-process-finished');
+        } else {
+          this.effect.render(this.object3D, camera, null);
         }
-        this.effect.render(this.object3D, camera);
 
         this.time = time;
+
```
ngokevin: whitespace

ngokevin:
@fernandojsg r?

src/core/scene/a-scene.js
```diff
@@ -312,10 +397,36 @@ module.exports = registerElement('a-scene', {
 
         if (this.isPlaying) {
           this.tick(time, timeDelta);
+          this.tock(time, timeDelta);
+        }
+
+        if (this.behaviors.tickPostRender.length || this.behaviors.tockPostRender.length) {
```
ngokevin: How about just a `behavior-added` event when you add a behavior?

src/core/scene/a-scene.js
```diff
@@ -312,10 +397,36 @@ module.exports = registerElement('a-scene', {
...
+          // Post processing setup has been modified, emit a notification.
+          if (this.postNeedsNotify) {
```
ngokevin: How about just a behavior-added event when you add a behavior?

src/core/scene/a-scene.js
```diff
@@ -312,10 +397,36 @@ module.exports = registerElement('a-scene', {
...
+              tick: behaviors.tickPostRender,
+              tock: behaviors.tockPostRender
+            });
+          }
+
+          this.effect.render(this.object3D, camera, this.renderTarget);
+
+          this.time = time;
+          time = window.performance.now();
+          timeDelta = time - this.time;
```
ngokevin: I think we can just reuse the timeDelta from before. Makes sense that it's timeDelta between each frame

src/core/scene/a-scene.js
```diff
@@ -299,6 +334,56 @@ module.exports = registerElement('a-scene', {
...
+
+    /**
+     * Behavior-updater meant to be called from scene after it renders.
+     * Abstracted to a different function to facilitate unit testing (`scene.tickPostRender()`) without
+     * needing to render. timeDelta is the time it took to render the scene.
+     */
+    tickPostRender: {
+      value: function (time, timeDelta) {
+        // Components.
+        this.behaviors.tickPostRender.forEach(function (component) {
```
ngokevin: Might want to abstract this to a function outside of the scene prototype below:

```
runBehavior('tickPostRender', time, timeDelta)

// ...

function runBehavior (type, time, timeDelta)
```

src/core/scene/a-scene.js
```diff
@@ -56,7 +56,13 @@ module.exports = registerElement('a-scene', {
...
     init: {
       value: function () {
-        this.behaviors = [];
+        this.behaviors = {
+          tick: [],
+          tock: [],
+          tickPostRender: [],
+          tockPostRender: []
+        };
+        this.postNeedsNotify = false;
```
ngokevin: Yeah, let's just emit right when we add/remove rather than keeping this state var and emitting during render loop

vendor/VREffect.js
wizgrav: I believe so, this is the signature of WebGLRenderer.render() and since VREffect behaves similarly it should match it.

src/core/scene/a-scene.js
wizgrav: Sure but it only applies to post behaviors, I think doing it for ticks especially would cause lots of traffic. It's also emmited when a post behavior is removed. I don't think separate events are needed for these. The compositor will mainly listen to these in order to recompile itself when something changes in the post chain. Maybe "post-behavior-changed" or something like that?

src/core/scene/a-scene.js
wizgrav: Well this will lead to several events handled where one would be sufficient, usually you'll pass several components to setup the chain at once. The compositor will catch the events and reconfigure itself. It's not really a problem as this will generally not happen often but If you mean to generalize the notification events for all types of behaviors then this will probably lead to alot of event traffic.

src/core/scene/a-scene.js
wizgrav: Sure, to be honest I can't think of a case for using the second timedelta besides knowing how long the base rendering took, maybe we could just mark for stats purposes?

src/core/scene/a-scene.js
wizgrav: Yeah good idea, lots of duplication in there

wizgrav: hmmm, not all behaviors are handled the same. The post behaviors don't care if the component's el isPlaying, I think we said that it shouldn't matter for these? As it is right now even if the scene is not playing the base rendering still happens so post render should as well

src/core/scene/a-scene.js
ngokevin: Perhaps just an event for added/removed post-processing behaviors then? Not too worried about batching them.

wizgrav: Ok I'll take out the dirty flag and emit in add/remove behavior, how does "post-behaviors-changed" sounds?

ngokevin: Yeah or `modified`

fernandojsg:
I'm looking around the API and I'm thinking how to deal with multiple postprocessing on the scene.
Sometimes you may want to add gloom + ao for the whole scene but also want to active from time to time other effects like nightvision or some kind of underwater distorsion.
Ideally you could be able to chain them, switch on/off and change the parameters for each one.
Did you already think about a way to solve this with the current implementation?
I was thinking on something like:
```
<a-postprocessing>
    <a-effect id="sparkles1" type="sparkles" img="sparkes1.jpg" opacity="0.5" color="#fff">
    <a-effect id="sparkles2" type="sparkles" img="sparkes2.jpg" opacity="1.0" color="#f00">
    <a-effect id="gloom" type="gloom" intensity="1.2" radius="2.0">
    <a-effect id="nightvision" type="nightvision" intensity="2.0" noise="1.2">
</a-postprocessing>
```
Everyone will be loaded but just some of them will be used, defined by the `post-process` param on `a-scene`, and just the last one is the one that will render back to the framebuffer.
```
<a-scene post-process="gloom, nightvision"/>
```
So you could add or remove effects from the postprocessing chain modifying `post-process` effect. And also you could modify the parameters for each effect accessing directly to the element defining it.

ngokevin:
Those individual effects could be components rather than adding a new element too.

@wizgrav I'm rethinking the tick/tock...It would be more flexible to explicitly define an order like you had before right?

fernandojsg:
but if you want to reuse some like the `sparkles` but with different parameters, you need to create a new component? or will you create a `a-mixin` for each one?

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,111 @@
+/* global AFRAME THREE */
+
+/*
+Kawase bloom, described here
+https://software.intel.com/en-us/blogs/2014/07/15/an-investigation-of-fast-real-time-gpu-based-image-blur-algorithms
+*/
+
+AFRAME.registerComponent('bloom', {
```
dmarcos: Can we reuse the registerShader API to register post processing effects?

examples/test/post-processing/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <meta name="description" content="Post-processing — A-Frame">
+    <script src="../../../dist/aframe.js"></script>
+    <script src="components/pulse.js"></script>
+    <script src="components/envmap.js"></script>
+    <script src="components/bloom.js"></script>
+    <script src="components/compositor.js"></script>   
+    <script src="components/bloom-toggler.js"></script>
+  </head>
+  <body>
+    <a-scene post-process="" bloom-toggler="">
```
dmarcos: If we reuse the register shader API to register effects one could do 
````html
<a-scene post-process="effects: bloom, nightVision; bloomIntensity: 1.0; nightVisionIntensity: 0.6">...</a-scene>
````
The name of the effect is prepended to the attribute names defined in the schema of the effect (aka shader)

**Comment by dmarcos** [comment]
What about this:
````html
<a-scene post-process="effects: sparkles, sparkles, bloom, nightVision; bloomIntensity: 1.0; nightVisionIntensity: 0.6; sparkles1Color: red; sparkles2Color: blue">...</a-scene>
````
The name of the effect and the index are automatically prepended based on the schema of the effect and the order in the list.

**Comment by ngokevin** [comment]
I think we should decide on a post-processing API later. This PR is really about being able to order component `tick` calls and being able to hook into the post-render. That is very flexible for now, and when we have more experience, we can maybe define an explicit PP API later.

**Comment by dmarcos** [comment]
I think the post-processing component should be the one calling the tick methods of each effect.

**Comment by ngokevin** [comment]
That could be layered in or decided later. Note that PP is just a component in `examples/`; it's just one use case. The PR is actually about ordering components. Another use case for that is the controls API needs support for ordering `controls` before `velocity`.

**Comment by dmarcos** [comment]
What I mean is I think we just need a new method tock called after the render call and that's all. If there are multiple tick functions to call on a specific order should be managed by the component post-processing. We don't need a generic API at this point. What other use case are there? The API proposed is over killing in my opinion.

**Comment by ngokevin** [comment]
So if we had two components `controls` and `velocity`, where `controls` needs to be run before `velocity`, we wouldn't be able to control that with just `tick`/`tock`. We need a way to order those `tick`s as well. That could be a different PR though, yeah.

Yeah, I think the bare minimum for this PR, is a `tock` and events on `add/removeBehavior`. I think PP could be accomplished within those constraints?

**Comment by dmarcos** [comment]
You can express that with the dependencies components. If a component is a dependency it will be initialized in first place and the tick method called before the dependent component one.

**Comment by fernandojsg** [comment]
As the active PP `effects` could be dynamically modified I believe the order must be controlled from a `post-process` component that is the responsible to access the PP system. I'm not sure why do you want to implement `tick/tock` for PP instead of doing a composer similar to the one used in threejs' `THREE.EffectComposer` or Jaume's `Wagner`. So you could just inject a call in the `a-scene` render when postprocessing is active and the order, passes and so on will be delegated to the `post-process` components, instead of worry about `tick` order for each effect's component.

**Comment by wizgrav** [comment]
@fernandojsg This is how I think the post processing could work with both flexibility and efficiency. Here's how the markup would look like

`<a-scene post-process="gloom, nightvision" gloom="intensity:0.5" nightvision="intensity:0.33"></a-scene>`

In this imagine that gloom would generate a texture that needs to be additively blended with the base rendering but nightvision is a simple equation applied to the frag color. The gloom component would do it's processing in tickPostRender and produce the texture. It would also define a postInfo property(not part of the components prototype) that would contain a unique id, uniform definitions for intensity etc and the fragment shader text it needs for the composition later kind of like

`postInfo: {
  name: "gloom",
  uniforms: {
    "srcTexture": {type: "t", value: this.gloomTexture} // Produced by the component
    "intensity": {type: "1f", value: 0.5} //Settable by the component in it's update()  
  },
  fragmentText: " void gloomPostProcess(in out vec4 currentColor, vec2 uv){ currentColor += texture2D(gloomInfo.srcTexture, uv) * gloomInfo.intensity; }"
}`

The "post-process" component, the compositor, would collect this information and create on shader with a setup like

`{
  uniforms: {
    "gloomInfo.srcTexture": gloomComponent.uniforms.srcTexture,
    "gloomInfo.intensity": gloomComponent.uniforms.intensity
  },
  fragmentShader: "
   sample2D sceneTexture; // scene.renderTarget
   varying vec2 vUv;
    struct gloomInfoStruct
    {
       sampler2D srcTexture;
       float intensity;
    } gloomInfo; // This is generated by the compositor based on the uniforms from gloom component

    void gloomPostProcess(in out vec4 currentColor, vec2 uv){ currentColor += texture2D(gloomInfo.srcTexture, uv) * gloomInfo.intensity; } // This comes from the gloom component itself

    void main(){
        vec4 color = texture2D(sceneTexture, vUv);
        gloomPostProcess(color,vUv);
        nightvisionPostProcess(color, vUv);
        gl_FragColor = vec4(color.rgb,1.0);
    }
  "
}`

Nightvision would not need to produce a texture so it would just provide uniforms for control and it's shader chunk. When you change attributes text components would set their uniforms in their update() and the compositor who has them by reference would apply them with no further intervention

**Comment by fernandojsg** [comment]
@wizgrav With that approach you're avoiding more drawcalls but if the effects change dynamically (going inside/outside water, nightvision on/off) you're rebuilding the whole shader on each switch which is not very performant :\
Also the `fragmentText` should be defined in that way by the user? or it's dynamically generated from the original fragment shader from the user?

**Comment by wizgrav** [comment]
when you say switch, you mean taking out the nightvision completely or just playing with it's intensity to make it apply more or less? In the former case yes the post chain will change and the compositor shader will need to be recompiled but if you just want to toggle the effect setting the intensity will just update the uniform. 

The fragmentText will be part of the gloomComponent's definition. The compositor will generate the gloomInfo struct text and include the fragmentText from gloomComponent + dynamically generate it's main as a chain. Some conventions are used, the components will have to include a unique prefix (postInfo.name) to aid the compositor for it's dynamic generation of text. 

The fragmentText of gloomComponent will be defined with these conventions in mind, that it will have access to a gloomInfo struct based on it's unique name containing it's uniforms and it will need to provide all the rest and name it's main function as gloomPostProcess so the compositor can call it

**Comment by fernandojsg** [comment]
I believe it will be quite hard to come out with an hybrid system to control the passes in order to optimize the drawcalls as creating just one whole shader is not possible for many effects so it should be necessary to create a logic to separate the "full" passes from the "composed" passes as you described.
For example if you want to apply blur, it's way much faster to do two passes with horizontal blur -> rotate -> horizontal blur, instead of one pass. Also you will have some passes that will be applied to RT of different sizes, so the system should be intelligent enough to deal with this...
Maybe you could define which passes should be executes separately in the API to prevent overengineering it.

**Comment by wizgrav** [comment]
Yes you can check out exactly that with bloom in the example. It's actually doing it in 5 passes(I wanted to try out kawase bloom) Components that need to create textures will do that in their tickPostRender(). They'd still need to provide some fragmentText for the blending. Bloom would need to add on top while ssao for instance would need to modulate. The compositor would be the only component that defines postRenderTock so it would run after all the others. 

There's still one caveat. Some effects need the final composited frame to work on. Fxaa or glitch for instance. For these instead of a postInfo, the component will provide a function with a signature of postProcess(inRenderTarget, outRenderTarget) and again it will be the compositor who will apply them after he's done with the uber shader

**Comment by wizgrav** [comment]
We could have the mechanism I described as standard for post processing in frame. I like the idea that it could instead be implemented on top of the existing component infrastructure though. This would make it easily exchangeable if a better scheme comes up afterwards.

**Comment by wizgrav** [comment]
@dmarcos @ngokevin is right, the postproc related stuff, addition of a .renderTarget and a switch to render on that instead of the screen are pretty much imposed by the apis provided by three itself. This PR has come to mostly concern with defining priorities for the execution of behaviors that need to run every frame. I'm leaning towards the, not so flexible, tick/tock mechanism because of the precedent in unity. It seems like a safe choice. Of course just because unity did it doesn't mean it is the final answer

**Comment by wizgrav** [comment]
@dmarcos reusing the dependencies property you mentioned makes sense and is attractive as an option. It is not enough as it is though as it ensures that controls will run before position for the components attached on a single entity only where what we really need is all control behaviors from all components executing before the first position component of any entity. Maybe not for the specific case but you should check unity's LateUpdate() for the rationale behind the need for ordering behavior groups/types. It's all about entity data interchange. 

We could accomplish that by making scene.behaviors a dynamically updating structure adding new symbols as they appear in addBehavior(), sorting only when a new symbol gets registered in runtime so as to be efficient. There may be some issues with circular dependencies but those will be error cases anyway. 

For post processing we could use tock() to tap on the post render side, allow systems to tap into it as well and define the compositor as a system to ensure it would run after all post component behaviors. I'd prefer this scheme than having to define numeric priorities, what do you think @ngokevin?

**Comment by wizgrav** [comment]
Following the discussions with @ngokevin @dmarcos @donmccurdy and @fernandojsg :

Enable post processing by extending components with a .tock() method which runs after the scene renders.

A-scene maintains a .renderTarget and switches rendering to that in the presence of at least one tock behavior.

A graph based mechanism is employed to allow for an explicitly definable execution order for both tick and tock behaviors.

Components can define two way connections in the graph using the properties .precedents and .following, similar to .dependencies.

Sorting to define the execution order takes place only in registerComponent() and batching is done efficiently in a-scene using caches.

The graph based mechanism works is fast and comfortable for the developer. Cyclic dependencies in the graph are errors and throw exceptions.

Post processing adapts naturally to the graph and becomes flexible and easy, aided by a-scene's new methods .getBehaviors() and .getPrecedence().

An example utilizing most of the functionality added by this PR can be seen at http://wizgrav.github.io/examples/test/post-processing/index.html

**Comment by wizgrav** [comment]
While fixing the tests, I found some issues that you should take note of @ngokevin @dmarcos. In a-entity.test.js components are manually cleared with "components[componentName] = undefined". I thought that this was the same as "delete components[componentName]". Apparently it isn't, the keys are still enumerable with an undefined value. I iterate the component keys to do the topological sort and undefined values were not handled. Should I handle these or just fix the tests to use delete instead?

File: examples/test/post-processing/components/bloom.js (hunk 26 above)
Comment (wizgrav) [inline examples/test/post-processing/components/bloom.js]: We could but it would be limiting. A postproc effect usually employs several shaders and passes to produce. For instance an ssao pass first runs the ssao shader which gets out noisy so it also applies a special blur to it, taking depth into account so as to avoid spilling the ao to near objects. The components are the best solution imo, and with the graph mechanism could be used in creative ways, like defining entities that each contain a whole chain of postprocessing filters and combine these entities

**Comment by ngokevin** [comment]
Yeah, we can fix the tests to `delete`

**Comment by ngokevin** [comment]
The `precedents`/`following` ain't too bad. It's better than just hijacking the `dependencies` API because that is meant for a different purpose (i.e., controls initialization rather than tick execution). It's also better than assigning numbers because it's much tighter and explicit.

Will have to look at the graph-based API because it is fairly complex. But sounds cool.

**Comment by wizgrav** [comment]
As the dependencies property was already defining a graph, the concept is not really foreign to the framework, it's more of an upgrade and generalization of an existing mechanism

**Comment by wizgrav** [comment]
Tidied it up, graphing related functionality is isolated in utils/precedence.js. The most drastic changes are in a-scene, the batching of components, but at least they have been done with potential future generalization in mind. The way the graph sorting is setup already makes it generally applicable. It can be used with systems and could even be used for shaders, geometries and anything that maps names to definitions. Wherever it makes sense it would be trivial to add it.

**Comment by ngokevin** [comment]
Thanks. We'll try to get to this and understand it. Though it's Memorial Day Weekend, I'll be on a trip next week, and we'll all be in London for a week right after. Focuses will be on controls, documentation, and 0.3.0 release, but I hope we can get to this!

**Comment by wizgrav** [comment]
Graphing is generally useful so I took it out into it's own package https://www.npmjs.com/package/precedence-maps and pull it as a dependency. The package also provides ordered stores so the code in a-scene is much cleaner now

File: vendor/VREffect.js (hunk 18 above)
Comment (ngokevin) [inline vendor/VREffect.js]: Want to do it or me? https://github.com/mrdoob/three.js/blob/master/examples/js/effects/VREffect.js

File: tests/core/scene/a-scene.test.js [hunk 28]
```diff
@@ -110,7 +110,7 @@ helpers.getSkipCISuite()('a-scene (with renderer)', function () {
 
   test('calls behaviors', function () {
     var scene = this.el;
-    var Component = { el: { isPlaying: true }, tick: function () {} };
+    var Component = {el: { isPlaying: true }, tick: function () {}};
```
ngokevin: you're my type of guy

ngokevin: also around the isPlaying too

src/core/system.js
```diff
@@ -75,6 +85,7 @@ module.exports.registerSystem = function (name, definition) {
   NewSystem.prototype = Object.create(System.prototype, proto);
   NewSystem.prototype.name = name;
   NewSystem.prototype.constructor = NewSystem;
+
```
ngokevin: no need

src/core/scene/a-scene.js
```diff
@@ -313,9 +367,24 @@ module.exports = registerElement('a-scene', {
...
+
+          this.tock(time, timeDelta);
+        } else {
+          this.effect.render(this.object3D, camera, null);
+        }
+
+        window.performance.mark('render-iteration-finished');
 
         this.time = time;
+
```
ngokevin: no need

src/core/scene/a-scene.js
```diff
@@ -313,9 +367,24 @@ module.exports = registerElement('a-scene', {
         if (this.isPlaying) {
           this.tick(time, timeDelta);
         }
-        this.effect.render(this.object3D, camera);
+
+        window.performance.mark('render-iteration-started');
```
ngokevin: can get rid of the whitespace lines until `render-iteration-finished` and move the comment above this entire block such that it's clear this whole block is about PP

src/core/scene/a-scene.js
```diff
@@ -212,6 +236,10 @@ module.exports = registerElement('a-scene', {
           antialias: antialias || window.hasNativeWebVRImplementation,
           alpha: true
         });
+
+        this.renderTarget = new THREE.WebGLRenderTarget(window.innerWidth, window.innerHeight, { minFilter: THREE.LinearFilter, magFilter: THREE.NearestFilter, format: THREE.RGBAFormat });
```
ngokevin: let's start steering towards no whitespace around `{}`

examples/index.html
```diff
@@ -229,6 +229,7 @@
       <li><a href="test/video/">Video</a></li>
       <li><a href="test/videosphere/">Video 360&deg;</a></li>
       <li><a href="test/visibility/">Visibility</a></li>
+      <li><a href="test/post-processing/">Post Processing</a></li>
```
ngokevin: alphabetize

ngokevin:
Do you think it'd be a good idea to split this PR up into three parts so it'd be easier to review?

- `tock`
- execution ordering
- post processing exaple

wizgrav:
@ngokevin yes I was also thinking about splitting it. The only issue is that the PRs would conflict with each other. They both touch this.behaviors in a-scene.js which is split in tick and tock. We could do the ordering as the first PR for tick() and then do tock(). 

Should the example for tock be a PR on it's own?

ngokevin:
Is it possible to do `tock` addition first? I think that would be easiest. And then later, we can layer on the ordering?

wizgrav:
it's certainly possible but mind that the example actually depends on the order guarantees. As the bloom component gets added/removed by bloom-toggler it's execution order will change. This means that the compositor component(post-process) which outputs to screen will blend the current scene's frame with the bloom texture from the previous frame. We can probably workaround that by also removing and readding the post-process in order as well

ngokevin:
Yeah, I was suggesting that we save the post-processing example for last.

wizgrav:
Ok I'll start off with a new PR for tock() then

ngokevin:
Cool. We can keep this one open to track the overall work.

cvan:
@wizgrav did you want to wait until PR #1564 is merged before fixing this one up too?

wizgrav:
@cvan the tock() method and the behavior ordering overlap on some parts. Should I merge both in this one?

wizgrav:
@cvan #1564 is rebased and clear for merge

cvan:
@wizgrav thanks. @ngokevin and @dmarcos are best to review #1564. thanks; excited to see this get merged 👍

ngokevin:
I think we can close. Not much demand for ordering past tick/tock and it seems aframe-effects was able to get by.

PR #1448, opened 2016-05-11
remove translate mapping from primitives, fix primitive bug

ngokevin:
**Description:**

Was previously updating Pano for guide. Will move that to separate repo.

**Changes proposed:**
- remove old translate mapping from primitives
- fix primitives when trying to define `material` properties

### DIFF
--- a/src/extras/primitives/registerPrimitive.js
+++ b/src/extras/primitives/registerPrimitive.js
@@ -78,7 +78,7 @@ module.exports = function registerPrimitive (name, definition) {
             var componentData = defaultData[componentName];
 
             // Set component properties individually to not overwrite user-defined components.
-            if (componentData instanceof Object && Object.keys(componentData).length) {
+            if (componentData instanceof Object) {
               var component = components[componentName];
               var attrValues = self.getAttribute(componentName) || {};
               var data = component.parse(attrValues);
@@ -88,7 +88,6 @@ module.exports = function registerPrimitive (name, definition) {
                 if (data[propName]) { return; }
                 data[propName] = componentData[propName];
               });
-              console.log(self, data);
               self.setAttribute(componentName, data);
               return;
             }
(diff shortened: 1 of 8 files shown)


dmarcos:
I would prefer to keep the examples as plain as possible only using aframe built in features (without 3rd party components) Also I don't believe on promoting a way to declaratively express logic.

dmarcos:
Can we have a separate repo for use cases, advance interaction techniques... ?

ngokevin:
I don't think we can compromise by just moving it to a separate repo. The promoted examples should come with A-Frame repo so that people can check it out like they can with three.js (which have tons of third-party modules).

- One of A-Frame's propositions is to be able to use components that other people have created; I don't think we should avoid showing that to feel pure.
- I have my preference, and you have yours, but neither are absolute. The side-projects I've worked on so far, this pattern has worked well for a framework of development. For an example of this scope, it is a lot easier to read. I want this to start writing the guide for 0.3.0.

ngokevin:
Updated.

src/extras/primitives/registerPrimitive.js
```diff
@@ -72,16 +72,19 @@ module.exports = function registerPrimitive (name, definition) {
...
           var self = this;
           var defaultData = this.defaultComponentsFromPrimitive;
+
           // Apply default components.
           Object.keys(defaultData).forEach(function applyDefault (componentName) {
             var componentData = defaultData[componentName];
+
             // Set component properties individually to not overwrite user-defined components.
-            if (componentData instanceof Object && Object.keys(componentData).length) {
+            if (componentData instanceof Object) {
```
dmarcos: Why is this check not necessary anymore?

ngokevin: It shouldn't have been there in the first place. It was there thinking that default components would only overwrite values if they had properties attached, but they were overwriting values regardless.

ngokevin:
r?

Issue #1285, opened 2016-03-28
add phi/theta start/length attributes to a-sphere primitive


ngokevin:
**Description:** The a-sphere primitive is missing some mappings to the geometry component.

- Expected: https://aframe.io/docs/components/geometry.html#Sphere
- Actual: https://aframe.io/docs/primitives/a-sphere.html#Attributes

- A-Frame Version: master
- Platform/Device: all

ngokevin:
https://github.com/aframevr/aframe/pull/1454

PR #1454, opened 2016-05-12
Expose segments, phi-length, phi-start, theta-length, and theta-start to the sphere component.

msimpson:
Fixing #1285


ngokevin:
nice

Issue #42, opened 2015-10-13
Add sound primitive


jcarpenter:
Enables devs to add audio to their scenes.

User can specify:

* Source URL
* Position of sound
* Bounding box and falloff (not sure how to articulate this in values)
* Loop / play-once
* User triggered (true/false. If true, sound is played when user enters bounding box)

@nickdesaulniers created a component for this under the original architecture. We may be able to re-use some of that logic?

Is dependent on (or related to?) https://github.com/MozVR/vr-markup/issues/255

cvan:
after @ngokevin's PR #883 is merged, we can do this with ease

ngokevin:
https://github.com/aframevr/aframe/pull/1455

PR #1455, opened 2016-05-12
Add the Sound Primitive.

msimpson:
Fix for #42. I assume this is all that is needed to make this work as it seemed to work in my project.


ngokevin:
thanks!

PR #1460, opened 2016-05-12
Change 'defaultAttributes' to 'defaultComponents' in Primitive Definitions

### LINKED ISSUE #1330: Rename `defaultAttributes` in extras/primitives to `defaultComponents`
**Description:** For primitives, we specify a set of default components in the primitive definition. These are named defaultAttributes, but it would be more correct to call them defaultComponents. A simple grep job.

msimpson:
This needs review as I had to alter the prototype to use `defaultComponentsFromPrimitive` in `registerPrimitive` instead of simply `defaultComponents` since Object.create will override the `defaultComponents` from the `a-entity` prototype. This, if allowed, would cause test failures as `getAttribute` subsequently returns `null` for those overridden attributes (i.e. position, scale, etc.).

Fixes #1330.


dmarcos:
It looks good to me. I let @ngokevin merge it.

PR #1462, opened 2016-05-13
On the first called to tick time and delta were undefined because ren…

dmarcos:
**Description:**

**Changes proposed:**
-
-
-

…der was not passed any argument

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -242,7 +242,7 @@ module.exports = registerElement('a-scene', {
             if (window.performance) {
               window.performance.mark('render-started');
             }
-            this.render();
+            this.render(0);
             this.renderStarted = true;
             this.emit('renderstart');
           }


src/core/scene/a-scene.js
```diff
@@ -308,10 +308,10 @@ module.exports = registerElement('a-scene', {
     render: {
       value: function (time) {
         var camera = this.camera;
-        var timeDelta = time - this.time;
```
ngokevin: timeDelta seemed fine to me

dmarcos: delta is pretty much standard in this context and more compact

ngokevin: better to be explicit and not optimize for typing

dmarcos: actually time and timeDelta kind of mix when reading and typing. it's not optimizing for typing efficiency but to avoid confusing both by making look different.

ngokevin: deltaTime? === dt

PR #1465, opened 2016-05-14
Fix of head tracking disabled on non vr mode on mobile

dmarcos:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Having troubles entering VR on iPhone and getting blocks of white on the screen. But that happens on master too.

PR #1468, opened 2016-05-16
glTF Loader

xirvr:
**Description:**
Adding the glTF loader from Three.

**Changes proposed:**
- Additional loader, works the same as the Collada loader.


xirvr:
Not working yet, making an early PR to get feedback on approach and to ask how I can get the glTFParser available to the glTFLoader?  The glTFLoader needs to have glTFParser available, but at run time it has a delegate instead of the object.

PR #1469, opened 2016-05-17
glTF Loader

xirvr:
**Description:**
Supporting the glTF Loader alongside the Collada Loader.

**Changes proposed:**
- Adding the glTFLoader

### DIFF
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -1,0 +1,57 @@
+var registerComponent = require('../core/component').registerComponent;
+var THREE = require('../lib/three');
+
+var GLTFLoader = THREE.glTFLoader; // To appease the linter
+var loader = new GLTFLoader();
+var gltf = null;
+
+module.exports.Component = registerComponent('gltf-model', {
+  schema: {
+    src: {type: 'src'},
+    loop: {default: true},
+    auto: {default: true}
+  },
+
+  init: function () {
+    this.model = null;
+  },
+
+  update: function () {
+    var self = this;
+    var el = this.el;
+    var src = this.data.src;
+    var loop = this.data.loop;
+    var auto = this.data.auto;
+
+    if (!src) { return; }
+
+    this.remove();
+
+    loader.load(src, function (gltfModel) {
+      var i = null;
+      var len = null;
+      var animation = null;
+      gltf = gltfModel;
+
+      self.model = gltfModel.scene;
+      el.setObject3D('mesh', self.model);
+      el.emit('model-loaded', {format: 'gltf', model: self.model});
+
+      if (gltf.animations && gltf.animations.length) {
+        len = gltf.animations.length;
+        for (i = 0; i < len; i++) {
+          animation = gltf.animations[i];
+          animation.loop = loop;
+          if (auto) {
+            animation.play();
+          }
+        }
+      }
+    });
+  },
+
+  remove: function () {
+    if (!this.model) { return; }
+    this.el.removeObject3D('mesh');
+  }
+});
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -1,5 +1,6 @@
 require('./camera');
 require('./collada-model');
+require('./gltf-model');
 require('./cursor');
 require('./geometry');
 require('./light');
--- a/src/extras/primitives/primitives/a-gltf-model.js
+++ b/src/extras/primitives/primitives/a-gltf-model.js
@@ -1,0 +1,9 @@
+var getMeshMixin = require('../getMeshMixin');
+var registerPrimitive = require('../registerPrimitive');
+var utils = require('../../../utils/');
+
+registerPrimitive('a-gltf-model', utils.extendDeep({}, getMeshMixin(), {
+  mappings: {
+    src: 'gltf-model'
+  }
+}));
--- a/src/extras/primitives/primitives/a-model.js
+++ b/src/extras/primitives/primitives/a-model.js
@@ -3,7 +3,7 @@ var registerPrimitive = require('../registerPrimitive');
 var utils = require('../../../utils/');
 
 registerPrimitive('a-model', utils.extend({}, getMeshMixin(), {
-  deprecated: '<a-model> is deprecated. Use <a-obj-model> or <a-collada-model> instead.',
+  deprecated: '<a-model> is deprecated. Use <a-obj-model>, <a-collada-model>, or <a-gltf-model> instead.',
 
   defaultComponents: {
     loader: {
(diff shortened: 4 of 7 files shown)


src/components/gltf-model.js
```diff
@@ -0,0 +1,61 @@
...
+
+      self.model = gltfModel.scene;
+      el.setObject3D('mesh', self.model);
+      el.emit('model-loaded', {format: 'gltf', model: self.model});
+
+      if (gltf.animations && gltf.animations.length) {
+        len = gltf.animations.length;
+        for (i = 0; i < len; i++) {
+          animation = gltf.animations[i];
+          animation.loop = true;
```
xirvr: @ngokevin Here's where I'm wanting to override with attributes passed in.

ngokevin: Use data.loop

donmccurdy: Do animations in glTF have unique names? We probably don't want to play all of them at once, but having an `animation` property in the schema would be a nice way to specify which animation should play, or to change the current animation. [Example](https://github.com/donmccurdy/aframe-extras/blob/master/src/loaders/three-model.js#L56).

xirvr: They have the object they animate nested in, but it requires expensive traversal to do that.  I'd recommend that since there's already a huge overhead for the glTF libraries not to do that but maybe add a utility to the component where you can find / cache animations to play / pause after load?

src/components/gltf-model.js
```diff
@@ -0,0 +1,61 @@
...
+  },
+
+  remove: function () {
+    if (!this.model) { return; }
+    this.el.removeObject3D('mesh');
+  },
+
+  tick: function () {
+    if (gltf && camera) {
+      THREE.glTFAnimator.update();
```
donmccurdy: This needs to be called once per frame, it looks like, not per model. I'm confused about why it's different from the generic `THREE.AnimationMixer` API.

xirvr: I still don't really understand the new animation system -- I copied in from the ThreeJS example page and was hoping someone who knew it better could recommend how to optimize this :)   The glTFShaders does need to be called, but I hadn't tried taking this out -- will give it a shot.

xirvr: I tried commenting it out, and the animation doesn't play.  If the THREE.AnimationMixer is being called elsewhere (I assume it is), then that would mean this is also required.

Moving it out of the component into the main event loop does make sense, but would make more work later to decouple the glTF functionality in a mix-and-match component system.

donmccurdy: Ah sorry, meant to say it needs to be called exactly once no matter how many models there are (not that it's already called somewhere else). Does it work if you put two glTF models into the same scene, or do the animations start going twice as fast? I don't think THREE.AnimationMixer is called anywhere either, yet, but I don't know.

xirvr: We're actually talking two things here:

1) Is THREE.AnimationMixer and glTFAnimator both needed?  We should probably understand this better before adding it in. 

2) Aside from THREE.AnimationMixer, should glTFAnimator be called for each object or only once for the app?  Even if it checks for duplicate calls, I think it would be clearer and technically more efficient to only call it once; but if we want to keep it to the glTF component, we might want to at least move it to a controller to avoid the duplicate calls.  My vote would just be put it in the event loop (wherever that is... still new here :)

dmarcos:
You can create a system and move your tick function there so it's called once per frame regardless of the number of models in the scene

src/systems/animation.js
```diff
@@ -0,0 +1,31 @@
...
+    }, true);
+  },
+
+  /**
+   * Update animation mixers and shaders.
+   *
+   */
+  tick: function () {
+    if (scene && camera) {
+      THREE.glTFAnimator.update();
```
xirvr: Moved these here per @dmarcos 's suggestion -- I called it animation because I'd assume we'd want to use the regular animation mixer here per @donmccurdy 's comments.  

Not making any command decisions here, easy to rename to this to glTF-specific, or call it something relating to imported objects / models?  Or tack it into another system?

src/components/gltf-model.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  init: function () {
+    this.model = null;
+  },
+
+  update: function () {
+    var self = this;
+    var el = this.el;
+    var src = this.data;
```
xirvr: @ngokevin I'm missing something small here, I think.  this.data points to schema.type['value'], if I change schema to { type: 'src', loop: 'loop', auto: 'auto' } this.data is still just the src value.  I also tried using this.schema directly, but that didn't work and seemed like going around the intended use anyway.

src/systems/animation.js
donmccurdy: There's also a global animation update in [a-scene.js::tick()](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L286-L287) ... would it make sense to move that here, or vice versa?

dmarcos: I would create a ```gltf-model``` system specific to your component (there's a 1:1 relationship) Systems are global API/State available to a component type. For instance, The geometry system has a cache that the all geometry components can use.

src/components/gltf-model.js
ngokevin: You can check the schema docs. It should look like:

```
schema: {
  src: {type: 'src'},
  loop: {default: true},
  auto: {default: true{
}
```

xirvr:
@ngokevin auto and loop are working, @dmarcos named the system gltf-animations -- gltf-model seemed confusing given it's the name of the component (3d model, data model), and there's no actual data associated with this system, more of a controller.

There's some other features I'd like to add (like play/pause animations associated with particular objects within a glTF scene), but seems like the most basic functionality is best for the initial PR.  Let me know if there's anything else I should add/remove/change.

src/systems/gltf-animations.js
```diff
@@ -0,0 +1,31 @@
+var registerSystem = require('../core/system').registerSystem;
+var THREE = require('../lib/three');
```
dmarcos: Components and systems should have the same name. So you can do ```this.system``` within the component.

xirvr: That'll be a handy feature when/if I/someone adds extra animation features (like from the component, this.system.addAnimation(animation); and in system, function playAllAnimations() { /* loop over this.animations */ }

xirvr:
Per @dmarcos going to start off putting this in a component for now because of the glTF library overhead : https://github.com/xirvr/aframe-gltf

ngokevin:
Yeah, that would be easiest. People can consume it right away, and later we will allow lazy-loading components in.

xirvr:
Converted to 3rd party component: https://github.com/xirvr/aframe-gltf

ngokevin:
Thanks for doing the good work. Shall we close this?

PR #1474, opened 2016-05-18
It removes the ability to provide an external canvas. The canvas size…

dmarcos:
… and positioning can be controled styling the a-scene when adding the embedded attribute


src/core/scene/a-scene.js
```diff
@@ -65,11 +65,30 @@ module.exports = registerElement('a-scene', {
           this.setupRenderer();
           this.resize();
         });
+        if (!this.hasAttribute('embedded')) {
```
ngokevin: Componentize to keep scene uncluttered?

ngokevin:
Should we add a test example showing multiple scenes on a page?

dmarcos:
I don't think multiple scenes per page works quite yet

src/components/scene/embedded.js
```diff
@@ -0,0 +1,13 @@
+var register = require('../../core/component').registerComponent;
```
ngokevin: registerComponent to be consistent

src/components/scene/embedded.js
```diff
@@ -0,0 +1,13 @@
+var register = require('../../core/component').registerComponent;
+
+module.exports.Component = register('embedded', {
+  schema: { default: true },
+
+  update: function () {
+    if (this.data === true) {
+      this.el.removeFullScreenStyles();
```
ngokevin: can name this `var sceneEl` to differentiate from entity

src/core/scene/a-scene.js
```diff
@@ -65,11 +65,30 @@ module.exports = registerElement('a-scene', {
...
         });
+        this.addFullScreenStyles();
         initPostMessageAPI(this);
       },
       writable: true
     },
 
+    addFullScreenStyles: {
+      value: function () {
+        var htmlEl = document.querySelector('html');
```
ngokevin: can do `document.documentElement`

src/components/scene/embedded.js
ngokevin: It might be more clear if an `embedded` component called `addEmbeddedStyles`/`removeEmbeddedStyles`. Right now, the code has a few double-negatives.

src/core/scene/a-scene.js
```diff
@@ -65,11 +65,30 @@ module.exports = registerElement('a-scene', {
...
+        htmlEl.classList.add('a-html');
+        document.body.classList.add('a-body');
+        this.classList.add('fullscreen');
```
ngokevin: I would prefix styles `aframe-` to reduce collision chances

src/core/scene/a-scene.js
```diff
@@ -65,11 +65,30 @@ module.exports = registerElement('a-scene', {
...
         initPostMessageAPI(this);
       },
       writable: true
     },
 
+    addFullScreenStyles: {
+      value: function () {
+        var htmlEl = document.querySelector('html');
+        htmlEl.classList.add('a-html');
+        document.body.classList.add('a-body');
```
ngokevin: `aframe-body`/`aframe-html` to be more clear. On large pages, `a-html` might not be parsed as A-Frame-related.

src/core/scene/a-scene.js
dmarcos: I think the collisions chances are minimal and it's already the established pattern. Not worth to go and change all class names now.

src/core/scene/a-scene.js
dmarcos: a-html is only for non embedded use cases. I expect to see on the page very little else besides the a-frame scene.

src/core/scene/a-scene.js
ngokevin: I wouldn't call the collision chances minimal. fullscreen is a very generic name, I wouldn't be surprised if a site had a class for that. it's very easy to make the chances zero now. The established pattern is that we prefix things

ngokevin: At least `a-fullscreen`

src/components/scene/canvas.js
```diff
@@ -1,47 +1,58 @@
...
-      canvas.dataset.aframeDefault = true;
-      scene.appendChild(canvas);
-    }
+
+  init: function () {
+    var sceneEl = this.el;
+    var canvasEl = document.createElement('canvas');
+    canvasEl.classList.add('a-canvas');
+    // Mark canvas as provided/injected by A-Frame.
+    canvasEl.dataset.aframeCanvas = true;
```
ngokevin: `aframeDefault` was the normalized class we used to determine all the injected/default elements from A-Frame.

src/components/scene/canvas.js
```diff
@@ -1,47 +1,58 @@
...
+        sceneEl.exitVR();
+      }
+      document.activeElement.blur();
+      document.body.focus();
+      setTimeout(sceneEl.resize.bind(sceneEl), 0);
+    }
   }
+
 });
+
```
ngokevin: don't need these extra lines

src/components/scene/embedded.js
```diff
@@ -0,0 +1,14 @@
+var registerComponent = require('../../core/component').registerComponent;
+
+module.exports.Component = registerComponent('embedded', {
+  schema: { default: true },
+
+  update: function () {
+    var sceneEl = this.el;
+    if (this.data === true) {
```
ngokevin: wouldn't just `this.data` do?

src/components/scene/canvas.js
```diff
@@ -1,47 +1,58 @@
...
+    function addFullscreenClass (event) {
+      canvasEl.classList.add('fullscreen');
+    }
+
+    function removeFullscreenClass (event) {
+      canvasEl.classList.remove('fullscreen');
+    }
+
+    function onFullScreenChange () {
```
ngokevin: function needs documentation on the general purpose + document steps along the way

src/core/scene/a-scene.js
```diff
@@ -65,11 +66,30 @@ module.exports = registerElement('a-scene', {
...
       },
       writable: true
     },
 
+    addFullScreenStyles: {
+      value: function () {
+        var htmlEl = document.documentElement;
+        htmlEl.classList.add('a-html');
+        document.body.classList.add('a-body');
+        this.classList.add('fullscreen');
```
ngokevin: prefix this class

src/style/aframe.css
```diff
@@ -31,6 +33,24 @@ body {
   width: 100%;
 }
 
+.a-canvas.fullscreen {
```
ngokevin: just `.a-fullscreen` would be specific enough

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
     enterVR: {
       value: function (event) {
         var self = this;
-        return this.effect.requestPresent().then(enterVRSuccess, enterVRFailure);
+        if (this.is('vr-mode')) { return; }
+        if (checkHeadsetConnected() || this.isMobile) {
+          // We send the canvas fullscreen so mirroring overlays
+          // any other elements on the page. Important with
+          // embedded scenes.
+          if (!this.isMobile) { this.requestFullscreen(); }
```
ngokevin: This might need to be in a separate `if` clause since it's weird to have `if (connected || mobile) { if (not mobile) ...`

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
+        var canvas = this.canvas;
+        var requestFullscreen =
+          canvas.requestFullScreen ||
+          canvas.webkitRequestFullScreen ||
+          canvas.mozRequestFullScreen;
+        requestFullscreen.apply(canvas);
+      }
+    },
+
     exitVR: {
```
ngokevin: function could use comments on generally what it does + the steps.

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
       value: function () {
         var self = this;
-        return this.effect.exitPresent().then(exitVRSuccess, exitVRFailure);
+        if (!this.is('vr-mode')) { return; }
```
ngokevin: could explain `Can't exit VR if not in VR mode. Return.`

ngokevin: Could also collapse everything into one if statement:

```
// Can't exit VR if not already in VR, or not using a VR-compatible device. Return.
if (!this.is('vr-mode') || (!checkHeadsetConnected() && !this.isMobile)
```

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
+        if (checkHeadsetConnected() || this.isMobile) {
+          return this.effect.exitPresent().then(exitVRSuccess, exitVRFailure);
+        }
+        exitVRSuccess();
```
ngokevin: How come we are calling this by itself?

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
         function exitVRSuccess () {
+          var embedded =
```
ngokevin: maybe mark a TODO for fixing getAttribute for boolean components

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
+            self.hasAttribute('embedded') && self.getAttribute('embedded') !== 'false';
           self.removeState('vr-mode');
           // Lock to landscape orientation on mobile.
           if (self.isMobile && window.screen.orientation) {
             window.screen.orientation.unlock();
           }
+          if (embedded) { self.removeFullScreenStyles(); }
```
ngokevin: comment. something like `Exiting VR in embedded mode, no longer need fullscreen styles.`

src/core/scene/a-scene.js
```diff
@@ -179,24 +226,16 @@ module.exports = registerElement('a-scene', {
       value: function () {
         var camera = this.camera;
         var canvas = this.canvas;
+        var embedded = this.hasAttribute('embedded') && !this.is('vr-mode');
```
ngokevin: don't you need to check for embedded/false here too

src/core/scene/a-scene.js
```diff
@@ -323,15 +360,15 @@ module.exports = registerElement('a-scene', {
   })
 });
 
-function getCanvasSize (canvas) {
-  if (isMobile) {
+function getCanvasSize (canvas, embedded) {
```
ngokevin: this could use a docstring now on the effects of passing `embedded`, return cases, and why

src/core/scene/a-scene.js
```diff
@@ -123,37 +143,64 @@ module.exports = registerElement('a-scene', {
...
-        return this.effect.requestPresent().then(enterVRSuccess, enterVRFailure);
+        if (this.is('vr-mode')) { return; }
+        if (checkHeadsetConnected() || this.isMobile) {
+          // We send the canvas fullscreen so mirroring overlays
+          // any other elements on the page. Important with
+          // embedded scenes.
+          if (!this.isMobile) { this.requestFullscreen(); }
+          return this.effect.requestPresent().then(enterVRSuccess, enterVRFailure);
+        }
+        // Just go fullscreen with no headset
```
ngokevin: nit: most our comments end with period

src/style/aframe.css
```diff
@@ -189,14 +209,4 @@ a-scene audio {
...
-@media (min-width: 480px) {
-  .a-enter-vr {
-    bottom: 20px;
-    right: 20px;
-  }
-  .a-enter-vr-modal {
-    width: 400px;
-  }
-}
+}
```
ngokevin: needs a newline

src/utils/checkHeadsetConnected.js
```diff
@@ -0,0 +1,12 @@
+var THREE = require('../lib/three');
+var dummyDolly = new THREE.Object3D();
+var controls = new THREE.VRControls(dummyDolly);
+
+module.exports = function checkHeadsetConnected () {
```
ngokevin: simple docstring

src/components/scene/embedded.js
```diff
@@ -0,0 +1,14 @@
+var registerComponent = require('../../core/component').registerComponent;
+
+module.exports.Component = registerComponent('embedded', {
```
ngokevin: simple docstring

ngokevin:
Need to move the A-Frame scene above the fold, else it looks like a normal Wikipedia page:

screen shot 2016-06-10 at 3 24 54 pm

src/components/scene/canvas.js
dmarcos: The word default does not make sense because the user cannot specify one anymore. For a camera or lights make sense but not for canvas.

src/components/scene/embedded.js
dmarcos: the less you rely on implicit type coercion the more robust and less error prone is your code.

src/core/scene/a-scene.js
dmarcos: I don't think prefixing all classes make sense that's a qualifier specific for a-scene

src/core/scene/a-scene.js
dmarcos: Complicated if clauses are hard to read and understand. I prefer to break them down in smaller bite sized chunks

src/style/aframe.css
dmarcos: using qualifiers is mainly to avoid collisions with existing class in the page. if we make fullscreen specific to a-canvas the prefix is no longer necessary and makes things a bit prettier to read and write.

src/core/scene/a-scene.js
ngokevin: it's more readable overall because you do all your checks at once and there's a logical progression to the function

1. Check if I can exit VR
2. Exit VR 

Rather than

1. Check if I'm in VR to see if I can exit VR 
2. If I can exit VR, exit VR

src/style/aframe.css
ngokevin: OK

src/core/scene/a-scene.js
ngokevin: OK

src/components/scene/canvas.js
ngokevin: ok i see

dmarcos:
This puppy is ready to go home.

ngokevin:
ok will look when i can

ngokevin:
Clicking on the Goggles for `Anime UI`, `Dynamic Lights`, other examples, the icon just disappears. For `Composite`, you get teleported into the cube. It works for `Cube` though.

ngokevin:
Can we maintain some of the padding in non-embedded mode? It's a bit too close to the corner now.

dmarcos:
This is ready to go. I made an important change I think it's sensible at that point. The default camera position was `x: 0; y:1.8; z: -4` so objects at `x: 0; y: 0; z: 0` are visible in front of the user. Now the camera starts by default at `x: 0; y: 1.8; z:0` so it's consistent with the position returned by a headset when entering VR.

ngokevin:
Can we maintain that offset if we detect non-VR fullscreen?

dmarcos:
No because the scenes will look different. It's better to be consistent and behave in the same way that other libraries operate. I think it only adds confusion.

ngokevin:
Right now, if I play a scene on desktop and enter fullscreen, I'm on the ground or below the ground.

dmarcos:
The offset in the y axis should still apply

ngokevin:
Yeah, I'm just talking about the Y-axis. Try the geometry gallery scene and enter fullscreen (non-VR).

dmarcos:
I believed I covered all the cases: Desktop with and without headset, mobile, scenes with and without default camera. I have to do one more round of tests with a headset.

src/systems/camera.js
```diff
@@ -39,12 +49,14 @@ module.exports.System = registerSystem('camera', {
...
       }
       defaultCameraEl = document.createElement('a-entity');
-      defaultCameraEl.setAttribute('position', DEFAULT_CAMERA_POSITION);
+      defaultCameraEl.setAttribute('position', '0 0 0');
       defaultCameraEl.setAttribute(DEFAULT_CAMERA_ATTR, '');
       defaultCameraEl.setAttribute('camera', {'active': true});
       defaultCameraEl.setAttribute('wasd-controls', '');
       defaultCameraEl.setAttribute('look-controls', '');
       sceneEl.appendChild(defaultCameraEl);
+      // Adding defautlt offset to the camera
```
ngokevin: typo

src/systems/camera.js
```diff
@@ -57,30 +69,58 @@ module.exports.System = registerSystem('camera', {
...
   removeDefaultOffset: function () {
     // Remove default camera if present.
-    var defaultCamera = this.sceneEl.querySelector('[' + DEFAULT_CAMERA_ATTR + ']');
+    var cameraOffset = this.cameraOffset;
     var currentPosition;
-    if (!defaultCamera) { return; }
+    var sceneEl = this.sceneEl;
+    var defaultCamera = sceneEl.querySelector('[' + DEFAULT_CAMERA_ATTR + ']');
+    // Checking this.headsetConnected to make the value injectable for unit tests.
+    var headsetConnected = this.headsetConnected || checkHeadsetConnected();
```
ngokevin: `this.headsetConnected` check doesn't seem needed.

If so then we can call `if (!defaultCamera || !cameraOffset || !checkHeadsetConnected()) { return; }` directly

src/systems/camera.js
```diff
@@ -57,30 +69,58 @@ module.exports.System = registerSystem('camera', {
...
-      z: currentPosition.z - DEFAULT_CAMERA_POSITION.z
+      x: currentPosition.x - cameraOffset.x,
+      y: currentPosition.y - cameraOffset.y,
+      z: currentPosition.z - cameraOffset.z
     });
   },
 
+  saveCameraPose: function () {
+    var activeCameraEl = this.activeCameraEl;
+    this.savedPose = {
```
ngokevin: can put `this.savedPose = null` in the `init`

ngokevin:
1. Consider configurable camera property `nonVRHeight` or similar name such that user-defined cameras can benefit from the non-VR offset. Default camera can set this to 1.8 by default.
2. Consider, now or later, to differentiate between VR and flat fullscreen. The save/restore camera pose when entering in and out seems useful for only VR, not flat fullscreen.

dmarcos:
Ready to go. The offset in the y is now applied to all cameras and configured via the `userHeight` property in camera. The offset is removed in the presence of a headset. When exiting VR the previous (before entering VR mode) position of the camera is restored when a headset is connected.

ngokevin:
very nice

ngokevin:
r+ when you ready to merge

ngokevin:
I would probably add `userHeight: 1.8` to `<a-camera>` as a default.

dmarcos:
@ngokevin can you test a bit to see if everything make sense?

ngokevin:
I tested around a few examples on my laptop, everything makes sense there:

- Flat:
  - Default camera, 1.8m offset.
  - Enter fullscreen, offset still applies.
  - Exit fullscreen, pose is not overwritten.

I haven't tested in VR. But I assume you have this:

- VR:
  - Enter VR, offset is removed.
  - Exit VR, pose is restored to before entering VR.

Just the one note that if I did `<a-camera></a-camera>`, I would sort of expect to have similar defaults as the default camera.

ngokevin:
We can decide on the user-defined `<a-camera>` default later.

ngokevin:
Nice. Love resolving the camera height inside and outside VR.

FrederickDesimpel:
If we're nitpicking **eyeheight** would rather be 1.6 m ...

PR #1478, opened 2016-05-19
tweak default lighting to make less flat

ngokevin:
**Description:**

Everything was looking flat.

**Changes proposed:**
- lower default ambient light
- strengthen default directional light and reposition

### DIFF
--- a/src/systems/light.js
+++ b/src/systems/light.js
@@ -45,13 +45,12 @@ module.exports.System = registerSystem('light', {
     var ambientLight = document.createElement('a-entity');
     var directionalLight = document.createElement('a-entity');
 
-    ambientLight.setAttribute('light',
-                              {color: '#fff', type: 'ambient'});
+    ambientLight.setAttribute('light', {color: '#BBB', type: 'ambient'});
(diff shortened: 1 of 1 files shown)


dmarcos:
Can you post a couple of screen shots with the before and after?

ngokevin:
screen shot 2016-05-19 at 11 02 04 am


screen shot 2016-05-19 at 11 04 16 am

dmarcos:
@fernandojsg what do you think?

fernandojsg:
I'll probably go for a lower value for the directional light intensity, something like ~0.6 instead of 0.8 and ambient light for #BBB instead of #AAA. 
Having very dark ambient light could produce results that looks more like and old engine with a phong model, and using a high direct light intensity will burn some scenes:
With 0.8 and #AAA:

screenshot 2016-05-19 20 50 49
screenshot 2016-05-19 20 50 44

With 0.6 and #BBB:
screenshot 2016-05-19 20 50 18
screenshot 2016-05-19 20 50 14
screenshot 2016-05-19 20 50 09

ngokevin:
updated to .6/bbb

PR #1483, opened 2016-05-19
fix parentNode error on load callback (fixes #1471)

### LINKED ISSUE #1471: Anime UI throws a-entity.js:232 Uncaught (in promise) TypeError: Cannot read property 'isPlaying' of null(…)
**Description:**

1. In Chrome, load [http://localhost:9000/examples/showcase/anime-UI/](http://localhost:9000/examples/showcase/anime-UI/)
2. Notice JS error:

```js
a-entity.js:232 Uncaught (in promise) TypeError: Cannot read property 'isPlaying' of null(…)
```

- A-Frame Version: `273622e`
- Platform/Device: Mac + Windows; Chrome + Chromium (Firefox is fine)

image

ngokevin:
**Description:**

@dmarcos and I don't know why this is needed. Before the callback, parentNode is defined. At the callback parentNode is briefly null. parentEl works for some reason. Have no idea what is happening to the read-only parentNode.

**Changes proposed:**
-
-
-

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -229,7 +229,8 @@ var proto = Object.create(ANode.prototype, {
       // Entity load.
       function entityLoadCallback () {
         self.updateComponents();
-        if (self.parentNode.isPlaying) { self.play(); }
+        // self.parentNode should work but that is null during this cb for unknown (#1483).
+        if (self.parentEl.isPlaying) { self.play(); }
       }
       ANode.prototype.load.call(this, entityLoadCallback, isEntity);
     },


src/core/a-entity.js
```diff
@@ -229,7 +229,7 @@ var proto = Object.create(ANode.prototype, {
       // Entity load.
       function entityLoadCallback () {
         self.updateComponents();
-        if (self.parentNode.isPlaying) { self.play(); }
+        if (self.parentEl.isPlaying) { self.play(); }
```
dmarcos: Let's add a comment here.

darkwing:
The logo demo on master is currently broken and `parentEl` is `null`, so looks like this is still an issue

darkwing:
Still hitting this but here's what I've noticed with the A-Frame Logo demo where this happens twice:

* `self` is `<a-entity light="" data-aframe-default-light="" position="" rotation="" scale="" visible=""></a-entity>`;   `parentEl` and `parentNode` are both null

* When inspecting the DOM to find elements, they're gone; instead there's:  `<a-light type="directional" color="#FFF" intensity="0.45" position="4 2 1" light="" rotation="" scale="" visible=""></a-light>`

If elements are being created/swapped then a reference isn't being updated which is why `parentNode` is null.

ngokevin:
Not being swapped, the defining lights replaces the default ones. They shouldn't be related to one another. But seems this might be a bug.

FrederickDesimpel:
I'm running into this too, while working on lighttarget.

FrederickDesimpel:
Could it have something to do with this:
In systems/light.js:
32 `sceneEl.removeChild(defaultLights[i]);`
a regular DOM removechild deletes the element, but what happens to the light component on it then? maybe first some cleanup is needed with the threejs scenegraph ?

PR #1486, opened 2016-05-19
getComputedAttribute is needed to calculate the direction in case rot…

dmarcos:
…ation is defined in the defaults or via mixins


ngokevin:
I wonder if there is a benefit to what `getAttribute` does vs `getComputedAttribute`. I haven't needed the former case so far.

PR #1487, opened 2016-05-20
postMessage scene events for detecting load state, managing entering/exiting VR

cvan:
**Description:**

I want to be able to know when a scene is loaded. And I want a programmatic way of entering VR - from the console, from a JS file included on the page, a content script (via Greasemonkey/dotjs/etc.), or from an `<iframe>`/`<webview>`.

**Changes proposed:**
- because the parent script needs to know when the scene(s) has loaded, we first `postMessage` an event indicating whether the scene(s) is loaded, and then the parent script can _then_ reliably `postMessage` to enter VR and exit VR. (previously we were blindly receiving `postMessage` events, which were not getting received unless you were working with `<iframe>`'d scenes on your origin that you  could introspect to get `load` events.)
- introduce `enter-vr-error` and `exit-vr-error` events.
- introduce `load-error` event (based on a timeout, which defaults to 30 seconds). not completely sold on this. thoughts?
- `postMessage` success + error events for loading, entering VR, and exiting VR.
- do not restrict only `<iframe>`s to be able to `postMessage` (see example cases above) - example code: `window.postMessage({type: 'aframe-scene', data: 'enter-vr'}, '*')`

### DIFF
--- a/tests/core/a-node.test.js
+++ b/tests/core/a-node.test.js
@@ -50,6 +50,33 @@ suite('a-node', function () {
         done();
       }, 50);
     });
+
+    test('can postMessage', function (done) {
+      var el = this.el;
+      var child = document.createElement('a-node');
+      el.appendChild(child);
+      window.top.addEventListener('message', function (event) {
+        assert.equal(event.detail.type, 'event');
+        assert.equal(event.detail.data.type, 'hadouken');
+        done();
+      });
+      child.emit('hadouken', {}, undefined, true);
+      setTimeout(function () { done(); }, 50);
+    });
+
+    test('defaults to not postMessage', function (done) {
+      var el = this.el;
+      var child = document.createElement('a-node');
+      el.appendChild(child);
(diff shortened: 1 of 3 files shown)


tests/core/a-node.test.js
```diff
@@ -50,6 +50,33 @@ suite('a-node', function () {
         done();
       }, 50);
     });
+
+    test('can postMessage', function (done) {
```
cvan: ignore tests for now 👍

dmarcos:
Is there any work left to do here?

cvan:
> Is there any work left to do here?

I wanted feedback on this approach. any concerns?

cvan:
I'm going to close this. I don't think it's necessary yet, given the functionality of the [WebVR Agent](https://github.com/webvrrocks/webvr-agent) (which is interoperable, library-agnostic and doesn't require modifications to A-Frame/three.js sites). If someone has a use for these `postMessage` hooks, let me know.

PR #1491, opened 2016-05-20
Add an `end` attribute to the animation element.

### LINKED ISSUE #1036: add `end` attribute (similar to `begin`) to a-animation to specify when to stop animation
Right now, there can be conflicting animations running. Imagine you have two animations:
- a animation that triggers on `click` that has a duration of 4 seconds and scales an element up
- another animation that triggers on `mouseleave` and scales that element back to its original size in 1 second

Once the user clicks on the element, the element animates, eg. 2 seconds, but then the user decides to cancel the zoom/scale by moving the cursor out. The second animation then triggers and scales the element back to the initial size. So far so good.

However, the first animation is still running in the background! And will eventually fullfil its animation by setting the element to the increased scale, Which results in a strange flicker because the other element scales down. etc.

I propose automatically cancel sibling animations that modify the same attribute.
Additionally, there could be a `cancel-other` attribute to `animation` that simply cancels other animations in the same scope.

msimpson:
Should fix #1036.


ngokevin:
Do you think you could add a unit test for this?

msimpson:
@ngokevin 
Okay. I'll try to expand what exists for a-animation to cover it.

msimpson:
@ngokevin 
Tucked it in at the bottom of the file and based it off the `click` event `begin` test.

ngokevin:
Thanks!

PR #1497, opened 2016-05-24
Make `setObject3D` bind `el` inside each child of `obj` when it is a `THREE.Group` (fixes #1403)

### LINKED ISSUE #1403: intersectedEl is null error
**Description:**

```
// Emit intersected on intersected entity per intersected entity.
    intersections.forEach(function emitEvents (intersection) {
      var intersectedEl = intersection.object.el;
      intersectedEl.emit('raycaster-intersected', {el: el, intersection: intersection});
    });
```

reported by @drakon58

- A-Frame Version: master
- Platform/Device:

msimpson:
While this fixed #1403 for me, it needs review. I have no idea what else this could effect, regardless of the existing tests passing.


src/core/a-entity.js
```diff
@@ -145,12 +145,18 @@ var proto = Object.create(ANode.prototype, {
...
+      var self = this;
       var oldObj = this.object3DMap[type];
       if (oldObj) { this.object3D.remove(oldObj); }
       if (obj instanceof THREE.Object3D) {
-        obj.el = this;
+        obj.el = self;
         this.object3D.add(obj);
       }
+      if (obj instanceof THREE.Group && obj.children.length) {
+        obj.traverse(function (child) {
```
ngokevin: Can name this function `bindEl`

ngokevin:
@dmarcos Looks good to me. The alternative is to have the raycaster traverse up, but that's not performant.

src/core/a-entity.js
```diff
@@ -153,7 +153,7 @@ var proto = Object.create(ANode.prototype, {
         this.object3D.add(obj);
       }
       if (obj instanceof THREE.Group && obj.children.length) {
-        obj.traverse(function (child) {
+        obj.traverse(function bindEl (child) {
```
msimpson: For debugging I assume.

ngokevin: Just to help readability by providing more information.

msimpson: Ah, I see.

src/core/a-entity.js
```diff
@@ -145,12 +145,18 @@ var proto = Object.create(ANode.prototype, {
...
+      var self = this;
       var oldObj = this.object3DMap[type];
       if (oldObj) { this.object3D.remove(oldObj); }
       if (obj instanceof THREE.Object3D) {
-        obj.el = this;
+        obj.el = self;
         this.object3D.add(obj);
       }
+      if (obj instanceof THREE.Group && obj.children.length) {
+        obj.traverse(function bindEl (child) {
```
dmarcos: Do we want to do this with only ```THREE.Group``` or for any Object3D with children?

msimpson: Yeah, that's a good point. In fact, do we need to be recursive as well?

ngokevin: Does this solution work with any OBJ mesh if we don't recurse?

msimpson: From what I can see inside Three's [OBJLoader](https://github.com/mrdoob/three.js/blob/44844c77f491e3115c3473f0237c25d6b1146d32/examples/js/loaders/OBJLoader.js#L163), all groups and objects are throw into a flat list. So this code would work with any OBJ borne mesh. I only worry about other formats; but if that's not the case, this could work:

```
if (obj instanceof THREE.Object3D) {
  obj.el = self;
  this.object3D.add(obj);
  if (obj.children.length) {
    obj.traverse(function bindEl (child) {
      child.el = self;
    });
  }
}
```

ngokevin: I think let's just not recurse for now, and do it if we find we have to. Don't want to unnecessarily do this.

ngokevin:
Can you add a small test in the entity suite?

msimpson:
Sure.

msimpson:
Hmm. Travis CI seems to fail a lot.

dmarcos:
@msimpson Thanks!

PR #1503, opened 2016-05-25
Added flatShading parameter to material component

fernandojsg:
**Description:** Added `flatShading` boolean attribute to `material` component to map to `shading=THREE.FlatShading or THREE.SmoothShading` in three.js


dmarcos:
@rlamana this might help to achieve the style you wanted for your demo

Issue #567, opened 2016-01-13
Investigate whether to switch to using webcomponents.js' Custom Elements polyfill


ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Tuesday Oct 20, 2015 at 23:17 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/294_

---

We're using the WebReflection polyfill for browsers that don't yet ship with native support for Custom Elements (i.e., every browser except Chrome):

https://github.com/WebReflection/document-register-element
https://github.com/WebReflection/document-register-element/blob/master/build/document-register-element.max.js

Polymer has maintained its polyfills as part of the [webcomponents.js](https://github.com/webcomponents/webcomponentsjs) project. I've heard from folks that the WebReflection polyfill is not a [spec](http://www.w3.org/TR/custom-elements/)-compliant polyfill.

@dmarcos fixed in d3fb0f99 an issue with Firefox/the polyfill's implementation for Firefox where `attributeChangedCallback` gets called even if the attribute values don't change.

And @dmarcos found [this issue](https://github.com/WebReflection/document-register-element/issues/33), which seems similar to the issues we've experienced with the ordering of `attachedCallback` getting fired. Related issue: #281.

This issue is to investigate whether we can drop in the webcomponents.js polyfill and get some better (if any?) stability:

https://github.com/webcomponents/webcomponentsjs/tree/master/src/CustomElements
https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.15/CustomElements.js

ngokevin:
<a href="https://github.com/dmarcos"><img src="https://avatars.githubusercontent.com/u/39342?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dmarcos](https://github.com/dmarcos)**
_Saturday Nov 28, 2015 at 07:13 GMT_

----

@cvan is this something we still want to do?

ngokevin:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cvan](https://github.com/cvan)**
_Sunday Nov 29, 2015 at 03:33 GMT_

----

Yup

ngokevin:
[...]

PR #1504, opened 2016-05-25
use webcomponents-lite polyfill #567

gasolin:
**Description:**

experiment with webcomponents-lite polyfill, 

the exported file (via `npm run dist`) is 30k bigger than the origin one (863k vs 892.6k)

The webcomponents-lite does not contain shadowdom polyfill, and the last update is a year old
https://github.com/webcomponents/webcomponents-lite

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,13 +29,13 @@
     "browserify-css": "^0.8.2",
     "debug": "^2.2.0",
     "deep-assign": "^2.0.0",
-    "document-register-element": "dmarcos/document-register-element#8ccc532b7",
-    "promise-polyfill": "^3.1.0",
     "object-assign": "^4.0.1",
     "present": "0.0.6",
+    "promise-polyfill": "^3.1.0",
     "style-attr": "^1.0.2",
     "three": "^0.76.1",
     "tween.js": "^15.0.0",
+    "webcomponents.js": "^0.7.22",
     "webvr-polyfill": "borismus/webvr-polyfill#f45f87a"
   },
   "devDependencies": {
--- a/src/core/a-register-element.js
+++ b/src/core/a-register-element.js
@@ -1,5 +1,5 @@
 // Polyfill `document.registerElement`.
-require('document-register-element');
+require('webcomponents.js/CustomElements.js');
 
 /*
  ------------------------------------------------------------


gasolin:
The correct one will be get webcomponents-lite.js from webcomponents.js

gasolin:
in 1st test fail https://travis-ci.org/aframevr/aframe/jobs/132840836#L559

seems `el.setAttribute('camera', 'fov: 65');` does not set correctly with webcomponents.js
https://github.com/aframevr/aframe/blob/master/tests/components/camera.test.js#L32

cvan:
thanks for doing this 👍

[`webcomponents-lite`](https://www.npmjs.com/package/webcomponents-lite) (updated a year ago) still includes Custom Elements, HTML Imports, and Templates. we use Custom Elements but not the other two in A-Frame (we used HTML Imports in 0.1.0 but they're gone now).

could try [`webcomponentsjs/micro`](https://www.npmjs.com/package/webcomponentsjs) (updated two months ago) includes only Custom Elements and Templates. can we try to see if that works?

gasolin:
The [webcomponentsjs](https://www.npmjs.com/package/webcomponentsjs) seems no star and no fork on github https://github.com/dlmma/webcomponentsjs , seems not reliable.

The patch is changed to use `webcomponents.js/webcomponents-lite.js` instead of outdated `webcomponents-lite` npm. The possible approach is build a customized version from https://github.com/webcomponents/webcomponentsjs that only include the custom element polyfil.

Though with current https://github.com/webcomponents/webcomponentsjs version, it failed A-Frame auto tests (but seems work well on examples by manual test). 

Might need help to identify what's the issue around the polyfill.

ngokevin:
Am not sure that'll you find a working one, we had to fix a couple of very subtle things with the polyfill we're currently using. If we wanted to save some of the bytes, we'd probably hack on https://github.com/dmarcos/document-register-element and remove lines of code. Not sure that is worth doing because then we are really maintaining a true fork.

cvan:
@gasolin: @ngokevin: I've heard from numerous folks that the WebReflection document.registerElement polyfill is incomplete and not spec compliant (and there are issues filed for them). @csuwildcat told me we should be using the webcomponentsjs Custom Element polyfill instead.

csuwildcat:
@gasolin @ngokevin I am not 100% sure of the coverage of the WebReflection polyfill at this point (it has been a while), but I can assure you that the WebComponentsJS polyfill strives to be as accurate as humanly possible, and is relied upon my millions of web pages.

May I ask why you are seeking another polyfill besides the most widely used?

csuwildcat:
FYI: X-Tag uses the webcomponentsls-lite build of the polyfill for our Core lib, so I know that one works well in prod.

ngokevin:
OK, we can try if @gasolin you want to help out with that, would be much appreciated. There may be quirks as A-Frame was developed on top of the other polyfill so it will take a lot of investigating and massaging.

dmarcos:
The webreflection polyfill has served us well. besides saving a few KBs what's the benefit of this?

gasolin:
If we took CustomElements.js from webcomponent.js, 
The aframe.min.js will grow from 861k to 873k

ngokevin:
Added a WIP tag. Let us know you can get the tests passing and investigate the quirks in switching.

ngokevin:
Going to close for inactivity, let us know if you make progress on this. Thanks!

PR #1505, opened 2016-05-26
update license

ngokevin:
r? @cvan 

What three.js does, and @donmccurdy says it would allow him to do blanket PRs. It looks harmless, but I am not sure of the implications.


cvan:
I'm fine with it

donmccurdy:
Thanks! ❤️

The wording is mysterious to me but this does help the PR approval process somehow.

PR #1506, opened 2016-05-27
Correct geometries schemas and add segments parameters for box and plane

fernandojsg:
**Description:**
I've been fixing the `type` and some values like `min` or `max` for most of the geometries based on the valid values for the three.js geometry implementation.
I've also exposed `segments(Width/Height/Depht)` for both `plane` and `box`.

### DIFF
--- a/src/geometries/torusKnot.js
+++ b/src/geometries/torusKnot.js
@@ -3,12 +3,12 @@ var THREE = require('../lib/three');
 
 registerGeometry('torusKnot', {
   schema: {
-    p: {default: 2, type: 'int'},
-    q: {default: 3, type: 'int'},
+    p: {default: 2, min: 1, type: 'int'},
+    q: {default: 3, min: 1, type: 'int'},
     radius: {default: 1, min: 0},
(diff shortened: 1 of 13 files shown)


src/geometries/torusKnot.js
```diff
@@ -3,12 +3,12 @@ var THREE = require('../lib/three');
...
   schema: {
-    p: {default: 2, type: 'int'},
-    q: {default: 3, type: 'int'},
+    p: {default: 2, min: 1, type: 'int'},
+    q: {default: 3, min: 1, type: 'int'},
     radius: {default: 1, min: 0},
     radiusTubular: {default: 0.2, min: 0},
-    segmentsRadial: {efault: 36, min: 0, type: 'int'},
-    segmentsTubular: {default: 32, min: 0, type: 'int'}
+    segmentsRadial: {efault: 36, min: 3, type: 'int'},
```
ngokevin: typo

fernandojsg: yep, fixed. thank you

Issue #1401, opened 2016-04-22
Stagger animations not supported.


JohnRodney:
if you do a `<a-animation begin="200" dur="1000" repeate="indefinite">` then the repeated will also stagger its delay by 200 milliseconds making it impossible to stagger a animation through a group.

Suggested that an animation should ignore the delay after the first run on the animation if it is set to repeat the animation.

msimpson:
This makes me wonder whether there should be a separate `delay` instead of overloading begin, as well.

ngokevin:
I think that would be pragmatic.

ngokevin:
We've deprecated a-animation. Try https://github.com/ngokevin/kframe/tree/master/components/animation and file issues, and component will be merged in in the future.

The component won't take into account the delay on subsequent runs. If you do want delay on subsequent runs, you can control the anim with a bit of JS

PR #1508, opened 2016-05-27
Deprecate usage of `begin` as a delay and provide a `delay` attribute, instead.

msimpson:
Start to address the concerns raised in #1401.


src/core/a-animation.js
```diff
@@ -158,6 +158,12 @@ module.exports.AAnimation = registerElement('a-animation', {
     update: {
       value: function () {
         var data = this.data;
+        // Deprecation warning for begin when used as a delay.
+        if (data.begin !== '' && !isNaN(data.begin)) {
+          console.warn("Using 'begin' to specify a delay is deprecated. Use 'delay' instead.");
+          data.delay = cloneValue(data.begin);
```
ngokevin: Don't think you need to clone because it's just a number.

src/core/a-animation.js
```diff
@@ -161,7 +161,7 @@ module.exports.AAnimation = registerElement('a-animation', {
         // Deprecation warning for begin when used as a delay.
         if (data.begin !== '' && !isNaN(data.begin)) {
           console.warn("Using 'begin' to specify a delay is deprecated. Use 'delay' instead.");
-          data.delay = cloneValue(data.begin);
+          data.delay = data.begin;
```
msimpson: Gotcha.

src/core/a-animation.js
```diff
@@ -114,7 +114,7 @@ module.exports.AAnimation = registerElement('a-animation', {
           self.count = repeat === Infinity ? 0 : parseInt(data.repeat, 10);
         }
 
-        if (isNaN(begin)) { begin = 0; }
+        if (isNaN(delay)) { delay = 0; }
```
msimpson: Actually, is it safe to remove this check as well?

ngokevin: Maybe good to leave in to handle weird input

msimpson: Okay.

PR #1511, opened 2016-05-28
Create new instance of THREE.ColladaLoader on collada-model init

### LINKED ISSUE #1510: Using multiple collada models sometimes causes latter models to render incorrectly
**Description:**
When loading multiple collada files, using a single collada loader for multiple models can result in issues rendering each model--causing an imploded/exploded appearance. I believe it may be related to this ThreeJS issue: https://github.com/mrdoob/three.js/issues/5721. The issue can be avoided by creating a new THREE.ColladaLoader instance for each collada model.

- A-Frame Version: 0.2.0
- Platform/Device: Mac OS 10.11.4

iandeboisblanc:
**Description:**
Fixes #1510. Creates new instance of collada loader for each collada-model component to prevent issue where using multiple collada models causes them to render incorrectly. 

**Changes proposed:**
- Move new 'THREE.ColladaLoader()' from outside component registration to within init method of collada-model component

### DIFF
--- a/src/components/collada-model.js
+++ b/src/components/collada-model.js
@@ -1,9 +1,6 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 
-var loader = new THREE.ColladaLoader();
-loader.options.convertUpAxis = true;
-
 module.exports.Component = registerComponent('collada-model', {
   schema: {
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks!

Issue #1514, opened 2016-05-28
Error while trying to add an empty geometry component to an entity


fernandojsg:
**Description:**
If you try to add an empty `geometry` component to an `entity` you'll get an error as it will expect to have a `primitive` defined. It should let you add it anyway, and after that you could add the primitive you want.
<img width="707" alt="screenshot 2016-05-28 21 44 20" src="https://cloud.githubusercontent.com/assets/782511/15629604/adac67aa-251d-11e6-8428-18ae2c52428f.png">

fernandojsg:
At first I thought maybe it could be as easy a to prevent the update schema to be executed
```javascript
  updateSchema: function (data) {
    if (!data.primitive)
      return;
  ...
```
But it will call anyway to `createGeometry` and it should returns a `bufferGeometry` so I guess it's not so simple to skip all these steps.

PR #1523, opened 2016-05-31
Added default primitive to geometry

fernandojsg:
**Description:** 
When you try to add an empty entity with an empty geometry Aframe throws an error as it looks for a `primitive` value to return the expected geometryBuffer. 
We should add a default value so empty component doesn't give you any error and also give you some kind of basic behaviour (the same as if you put an empty material).
So basically what I've done is add a `default: 'box'` to the primitive schema and you'll get a box by default.
Fixes https://github.com/aframevr/aframe/issues/1514

### DIFF
--- a/tests/components/geometry.test.js
+++ b/tests/components/geometry.test.js
@@ -16,6 +16,10 @@ suite('geometry', function () {
   });
 
   suite('update', function () {
+    test('allows empty geometry', function () {
+      this.el.setAttribute('geometry', '');
+    });
+
     test('creates geometry', function () {
       var mesh = this.el.getObject3D('mesh');
       assert.ok(mesh.geometry);
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -15,7 +15,7 @@ module.exports.Component = registerComponent('geometry', {
   schema: {
     buffer: { default: true },
     mergeTo: { type: 'selector' },
-    primitive: { default: '', oneOf: geometryNames },
+    primitive: { default: 'box', oneOf: geometryNames },
     skipCache: { default: false }
   },
 


ngokevin:
Small test for geometry component just to check an empty primitive doesn't throw error?

fernandojsg:
Test added.

tests/components/geometry.test.js
```diff
@@ -16,6 +16,11 @@ suite('geometry', function () {
   });
 
   suite('update', function () {
+    test('empty geometry', function () {
```
ngokevin: coudl do `allows empty geometry` to be more specific

tests/components/geometry.test.js
```diff
@@ -16,6 +16,11 @@ suite('geometry', function () {
...
+      var el = this.el;
+      el.setAttribute('geometry', '');
```
ngokevin: we can just do `this.el.setAttribute('geometry', '');` in one line

tests/components/geometry.test.js
ngokevin: not a big deal though, can merge anyways

tests/components/geometry.test.js
ngokevin: not a big deal though, can merge anyways

fernandojsg: sure! I didn't noticed it after copypaste :)

tests/components/geometry.test.js
fernandojsg: done!

ngokevin:
Thanks, I was waiting for the tests to pass, then eventually forgot about it.

PR #1524, opened 2016-05-31
add includeMeshProperties property for primitive registration

ngokevin:
**Description:** I am documenting primitive registration. Most of our primitives was using a mixin for common material mappings, but that was public. I made it public by adding an easy flag. This will make it easy for people (and for us) to register primitives.

**Changes proposed:**
- includeMeshProperties attribute for primitive registration
- remove deprecated `<a-model>` primitive


src/extras/primitives/primitives/a-box.js
```diff
@@ -1,17 +1,15 @@
-var getMeshMixin = require('../getMeshMixin');
 var registerPrimitive = require('../registerPrimitive');
-var utils = require('../../../utils/');
 
-registerPrimitive('a-box', utils.extendDeep({}, getMeshMixin(), {
+registerPrimitive('a-box', {
+  includeMeshProperties: true,
```
dmarcos: The flag is too specific ad-hoc to mesh primitives. If you establish the pattern you will have to create other flags for other property groups. I think the previous pattern is less magic and also scale better for future extensions.

PR #1539, opened 2016-06-05
Fixed two closing tags in Primitives description

Zeouterlimits:
**Description:**
Noticed incorrect sky closing tag instead of box closing tag
Noticed a-video was missing a closing tag


dmarcos:
Thanks!

PR #1544, opened 2016-06-09
fix <a-scene debug> equating to false (fixes #1466)

### LINKED ISSUE #1466: <scene debug> should work. But only <scene debug="true"> works
**Description:**

debug mode enables serialization of component data to the DOM. if this is on, you will be able to see DOM updates like with animations or setAttributes. This is done in `<a-entity>` code. Grep for  `isDebugMode`

- A-Frame Version: master

AVGP:
**Description:**

- Added tests for the `debug` property of `a-entity`
- Fixed behaviour when `debug` is used as a boolean property

**Changes proposed:**
- `debug` property can either be a normal string property (`debug=true`) or a boolean one (`debug` without any value)

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -539,7 +539,7 @@ var proto = Object.create(ANode.prototype, {
    */
   setAttribute: {
     value: function (attr, value, componentPropValue) {
-      var isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
+      var isDebugMode = this.sceneEl && this.sceneEl.hasAttribute('debug');
       if (components[attr]) {
         // Just update one of the component properties
         if (typeof value === 'string' && componentPropValue !== undefined) {
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -660,11 +660,13 @@ suite('a-entity', function () {
       el.setAttribute('debug', true);
       el.addEventListener('loaded', function () {
         var child = document.createElement('a-entity');
+        child.addEventListener('loaded', function () {
+          child.setAttribute('position', '1 2 3');
+          assert.equal(child.attributes[0].value, '1 2 3');
+          done();
+        });
         el.appendChild(child);
         document.body.appendChild(el);
-        child.setAttribute('position', '1 2 3');
-        assert.equal(child.attributes[0].value, '1 2 3');
-        done();
       });
       document.body.appendChild(el);
     });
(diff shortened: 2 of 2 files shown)


src/core/a-entity.js
```diff
@@ -539,7 +539,7 @@ var proto = Object.create(ANode.prototype, {
    */
   setAttribute: {
     value: function (attr, value, componentPropValue) {
-      var isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
+      var isDebugMode = this.sceneEl && this.sceneEl.hasAttribute('debug');
```
dmarcos: This is still not correct. Debug mode will trigger in this case: 
````html
<a-scene debug="false"></a-scene>
````

dmarcos: I think the proper solution is that for singleProperty components if the attribute is an empty string we should return the default value that in case of the debug component is `true`

AVGP: Hm, but according to spec, if an attribute is considered boolean, the only way to disable it is to omit it, see https://html.spec.whatwg.org/#boolean-attributes:

> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

I also would consider this to be a boolean attribute and the "omit or true" mechanism makes sense IMHO as you'd normally leave it out and only specify it, if you want debug, no?

ngokevin: The spec refers to boolean attributes, though A-Frame components have a little bit more juice. That spec always confused people (`disabled="false"` means disabled?) too, so if we can avoid that confusion, it would be nice.

I think @dmarcos is right, making `getAttribute('debug')` on `<a-scene debug>` return the default component value seems best because we've had to do this in other places: https://github.com/aframevr/aframe/pull/1474/files#diff-eb9bb1084863a2f65678ea67d6e7cf5eR197

Thanks for the work so far, I'll look into advice on how to do this.

ngokevin: Steps look like:

- https://github.com/aframevr/aframe/blob/master/src/core/schema.js#L124 - Pass in the default value from the prop definition to to parse function.
- https://github.com/aframevr/aframe/blob/master/src/core/propertyTypes.js#L67 - Treat empty string as wanting the default value.

ngokevin:
We got working a proper solution at https://github.com/aframevr/aframe/commit/365829f688a420631e0f575c2598fe07a6abe554 which was a bit more involved and fiddly.

Thanks so much for the work though!

PR #1545, opened 2016-06-10
include reason in error message for scene enterVR/exitVR errors

cvan:
**Description:**

The true error messages for `vrDisplay.requestPresent()` and `vrDisplay.exitPresent()` calls were getting swallowed.

**Changes proposed:**
- Include the error's `message` in the message we print.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -138,8 +138,12 @@ module.exports = registerElement('a-scene', {
             window.screen.orientation.lock('landscape');
           }
         }
-        function enterVRFailure () {
-          throw new Error('enter VR mode error. requestPresent failed');
+        function enterVRFailure (err) {
+          if (err && err.message) {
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -157,8 +159,10 @@ module.exports = registerElement('a-scene', {
           self.resize();
           self.emit('exit-vr', {target: self});
         }
-        function exitVRFailure () {
-          throw new Error('exit VR mode error. exitPresent failed');
+        function exitVRFailure (err) {
+          var reason = err && err.message;
+          throw new Error('Failed to exit VR mode (`exitPresent`).' +
+            (reason ? '\n' + reason : ''));
```
dmarcos: can you move this to a variable?
`var reasonString = reason ? '\n' + reason : '';`

cvan: can do. I did that originally, but so many `var`s

cvan:
thanks for deleting, @ngokevin

PR #1548, opened 2016-06-10
fix unclosed <a-mixin> tags in performance animation example

cvan:


ngokevin:
Can you create branches on your fork and not `aframevr`'s?

cvan:
@ngokevin: I did here. you're thinking of the #1545.

as long as I delete them afterwards, is there an issue? for non-code changes especially, the Edit button is great for editing docs.

ngokevin:
OK, didn't realize those were from the Edit button. nvm

PR #1549, opened 2016-06-10
don't require shader to return materials, only set

ngokevin:
This matches geometry registration. And the extra return is not needed since it's set on the instance, and it's weird for init/update methods to be returning.

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -82,12 +82,15 @@ module.exports.Component = registerComponent('material', {
   updateShader: function (shaderName) {
     var data = this.data;
     var Shader = shaders[shaderName] && shaders[shaderName].Shader;
-    var material;
+    var shaderInstance;
+
     if (!Shader) { throw new Error('Unknown shader ' + shaderName); }
-    this.shader = new Shader();
-    this.shader.el = this.el;
-    material = this.shader.init(data);
-    this.setMaterial(material);
+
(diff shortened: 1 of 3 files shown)


src/components/material.js
```diff
@@ -81,13 +81,16 @@ module.exports.Component = registerComponent('material', {
...
   updateShader: function (shaderName) {
     var data = this.data;
-    var Shader = shaders[shaderName] && shaders[shaderName].Shader;
-    var material;
-    if (!Shader) { throw new Error('Unknown shader ' + shaderName); }
-    this.shader = new Shader();
-    this.shader.el = this.el;
-    material = this.shader.init(data);
-    this.setMaterial(material);
+    var ShaderClass = shaders[shaderName] && shaders[shaderName].Shader;
```
dmarcos: I think the world class here is not a good fit. It's confusing. We don't use classes. The capitalization of the variable already indicates a constructor. It's probably the most widely spread convention in JS.

ngokevin: Is there another word we can use? Because there's `shaders`, `.Shader`, `this.shader`, `shader`.

dmarcos: I think the capitalization is enough to discriminate.

ngokevin: k

ngokevin: k

PR #1557, opened 2016-06-14
Preserve old geometry metadata.

### LINKED ISSUE #1538: ______BufferGeometry ➡️  BufferGeometry change breaking extras physics
Question related to #1338, I follow that BufferGeometry is more efficient than standard Geometry, but, why is creating a BoxGeometry first and converting it to BufferGeometry better than just creating a BoxBufferGeometry? Are the _____BufferGeometry subclasses less efficient than BufferGeometry?

Physics for aframe-extras have broken against aframe/master – [I construct CANNON.Shape instances by looking at the THREE.js geometry](https://github.com/donmccurdy/aframe-extras/blob/master/lib/CANNON-mesh2shape.js#L29-L50) type. ____BufferGeometry is useable, but there's not much to go on with plain BufferGeometry instances.

donmccurdy:
Fixes #1538

If we're using BufferGeometry instead of BoxBufferGeometry / PlaneBufferGeometry (not sure what the performance implications are on that switch), it would be helpful to preserve some of the metadata. For example, I use this to [construct CANNON.Shape instances from the geometry](https://github.com/donmccurdy/aframe-extras/blob/master/lib/CANNON-mesh2shape.js).

### DIFF
--- a/src/systems/geometry.js
+++ b/src/systems/geometry.js
@@ -132,6 +132,7 @@ function toBufferGeometry (geometry, doBuffer) {
   if (!doBuffer) { return geometry; }
 
   bufferGeometry = new THREE.BufferGeometry().fromGeometry(geometry);
+  bufferGeometry.oldMetadata = {type: geometry.type, parameters: geometry.parameters || {}};
   geometry.dispose();  // Dispose no longer needed non-buffer geometry.
   return bufferGeometry;
 }
--- a/src/systems/geometry.js
+++ b/src/systems/geometry.js
@@ -132,7 +132,7 @@ function toBufferGeometry (geometry, doBuffer) {
   if (!doBuffer) { return geometry; }
 
   bufferGeometry = new THREE.BufferGeometry().fromGeometry(geometry);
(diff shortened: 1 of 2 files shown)


src/systems/geometry.js
```diff
@@ -132,6 +132,7 @@ function toBufferGeometry (geometry, doBuffer) {
   if (!doBuffer) { return geometry; }
 
   bufferGeometry = new THREE.BufferGeometry().fromGeometry(geometry);
+  bufferGeometry.oldMetadata = {type: geometry.type, parameters: geometry.parameters || {}};
```
ngokevin: What about just `metadata`?

donmccurdy: No preference – only precedent I could find for this was `mesh.userData[*]` but it would be a little more complicated to pass this metadata up to the Mesh probably

ngokevin: Was thinking we name it `metadata` since the data is still current, but just destroyed.

donmccurdy: Sounds good / done!

PR #1564, opened 2016-06-18
Component tock() method for post scene rendering operations.

wizgrav:
**Description:**

Enables post scene rendering operations in aframe.

**Changes proposed:**

Extends a-scene element with a .renderTarget(to be set by a component/system on demand) to enable the rendering of the scene in a texture.
Adds a tock() method on components/systems where post scene rendering logic can be performed.
An example showcasing the functionality is included.

https://wizgrav.github.io/aframe-clubber/demo/


vendor/VREffect.js
```diff
@@ -232,7 +232,7 @@ THREE.VREffect = function ( renderer, onError ) {
 	var cameraR = new THREE.PerspectiveCamera();
 	cameraR.layers.enable( 2 );
 
-	this.render = function ( scene, camera ) {
+	this.render = function ( scene, camera, renderTarget, forceClear ) {
```
dmarcos: We should not be changing the code in the vendor folder. If there's something to be changed in VREffect we should submit a PR to three.js https://github.com/mrdoob/three.js/blob/master/examples/js/effects/VREffect.js

dmarcos:
This code is much cleaner that the previous attempt. @fernandojsg, What do you think?

wizgrav:
@dmarcos implementing the final compositor as a system does away with the need for ordered execution of behaviors so this PR provides all that's needed for post processing. This setup is more proper since there can be only one compositor active. I also included the example as there is no need anymore for further PRs for ordering as discussed with @ngokevin . 

Execution ordering could be implemented with something complex like the graph approach or something simple like just using a boolean flag on components to switch between behaviors.unshift/push in addBehavior, but that's now unrelated to post proc and can be decided on a later time

vendor/VREffect.js
wizgrav: I agree and the change is both trivial and makes VREffect more consistent with the rest of three. I'll try submiting a PR to threejs as well but I suggest we keep the patched version for now.

dmarcos: The vendor directory must not have any patched code. It's messy to maintain and easy to forget that we have ad-hoc changes. This PR assumes new functionality on the VR Effect and Three.js might not agree on the approach. This would force us to redo this patch or maintain a fork of VREffect. Let's share the changes on a PR on three.js to get their feedback before merging this.

wizgrav: Ok I'm getting on it

wizgrav:
VREffect additions submitted to three.js dev mrdoob/three.js/issues#9172

dmarcos:
@wizgrav nice, thanks!

dmarcos:
Nice it seems mrdoob accepted the patch

dmarcos:
We can now copy and paste to the vendor directory

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+      if (mat && mat.postOpacity !== undefined) {
+        mat.opacity = mat.postOpacity;
+      }
+    });
+  },
+
+  tock: function () {
+    var scene = this.sceneEl;
+    var rt = scene.renderTarget;
+    if (!scene.enablePostProcessing) return;
```
dmarcos: missing curly brackets around return

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+    if (this.needsConfig) {
```
dmarcos: it can be a one liner

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+      this.config();
+    }
+    var uns = this.quadPost.material.uniforms;
```
dmarcos: call it uniforms instead of the more cryptic uns

examples/test/post-processing/components/envmap.js
```diff
@@ -0,0 +1,14 @@
+/* global AFRAME THREE */
+
+AFRAME.registerComponent('envmap', {
+
+  update: function () {
+    this.el.envEl = document.querySelector('a-videosphere');
+  },
+
+  tick: function () {
+    var et = this.el.envEl.object3DMap.mesh.material.map;
```
dmarcos: can we find a better variable name than et?

examples/test/post-processing/components/pulse.js
```diff
@@ -0,0 +1,25 @@
...
+AFRAME.registerComponent('pulse', {
+  dependencies: ['geometry', 'material'],
+
+  schema: {
+    phase: { default: 0, min: 0, max: 1 },
+    exponent: { default: 1 }
+  },
+
+  remove: function () {
+    var m = this.el.object3DMap.mesh;
```
dmarcos: don't access the map directly use this.el.getObject3D('mesh')

examples/test/post-processing/components/pulse.js
```diff
@@ -0,0 +1,25 @@
...
+    delete m.postOpacity;
+  },
```
dmarcos: we need a better variable name. single letter variables are cryptic

examples/test/post-processing/components/pulse.js
```diff
@@ -0,0 +1,25 @@
...
+
+  tick: function (time) {
+    var m = this.el.object3DMap.mesh;
```
dmarcos: better variable name. use getObject3D('mesh') method

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+      var mat = obj.material;
+      if (mat && mat.postOpacity !== undefined) {
+        mat.opacity = mat.postOpacity;
+      }
+    });
+  },
+
+  tock: function () {
+    var scene = this.sceneEl;
+    var rt = scene.renderTarget;
```
dmarcos: better variable name

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+
+    });
+  },
+
+  // Traverse scene objects and set opacity based on the postOpacity attribute set by the pulse component.
+  tick: function (time) {
+    var scene = this.sceneEl;
+    if (!scene.enablePostProcessing) return;
+    scene.object3D.traverse(function (obj) {
+      var mat = obj.material;
```
dmarcos: no abrevations. use material as variable name

examples/test/post-processing/components/compositor.js
```diff
@@ -0,0 +1,196 @@
...
+  postModified: function (ev) {
+    this.needsConfig = true;
+  },
+
+  // Whenever the postproc setup changes, the compositor shader is recompiled
+  config: function () {
+    var scene = this.sceneEl;
+    var behaviors = scene.behaviors.tock;
+    var bloom = behaviors.length ? behaviors[0] : null;
+    console.log('POST-MODIFIED', behaviors);
```
dmarcos: is this logging needed?

examples/test/post-processing/components/post-toggler.js
```diff
@@ -0,0 +1,21 @@
+/* global AFRAME */
+
+AFRAME.registerComponent('post-toggler', {
+  init: function () {
+    this.evHandler = this.clickHandler.bind(this);
```
dmarcos: we can use clickHandler as variable name

wizgrav:
@dmarcos Yeah that was very responsive and kind of them actually. I'll include the latest VREffect from three/dev in this PR. Also, a guy from the slack channel asked me and so I've also made another example. The approach taken on that, while not as efficient as I'd like, is more akin to the effect composer from three.js. Post effect components are very easy to write as they're just wrappers over the various three shaders. They just provide a shaderMaterial to the compositor system for blending to screen and a namespace to update uniforms. Should I include that as well? 

http://wizgrav.github.io/aframe/examples/test/post-processing-grain/

examples/test/post-processing/components/compositor.js
wizgrav: not really, I put the logs in just so there can be a reference of things that could be useful for anyone writing post proc. Should I take them out?

examples/test/post-processing/components/envmap.js
wizgrav: yeah I'll change it to environmentMap. Is there an event emmited somewhere that would indicate that the videosphere is ready with mesh and material? If we could catch something like that we could get rid of the tick method here, the mapping only needs to be set once

fernandojsg:
I still didn't check the code, by now I'm trying to run the demos with a headset and on chromium they won't work as the video won't load so maybe the demo version could be something simpler using just geometry?
I've tried also to run it on Firefox nightly 50.0a1 with Oculus Rift and I just get rendered the right eye, the left eye is black. 
Has someone successfully make them work on windows with headset?

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,161 @@
...
+        pixelSize: { type: 'v2', value: new THREE.Vector2(0, 0) }
+      },
+      vertexShader: this.vertexBlurText,
+      fragmentShader: this.fragmentBlurText,
+
+      depthWrite: false
+
+    });
+    this.quadPost.material = this.brightMaterial;
+    var rtOptions = { minFilter: THREE.LinearFilter, magFilter: THREE.NearestFilter, format: THREE.RGBAFormat };
```
fernandojsg: better use whole words for variables instead of abbreviations: `renderTargetOptions`

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,161 @@
...
+    this.el.emit('post-modified');
+  },
+
+  remove: function () {
+    this.el.emit('post-modified');
+  },
+
+  update: function (oldData) {
+    this.postIntensity = this.data.intensity;
+    console.log(this.postIntensity);
```
fernandojsg: remove the log

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,161 @@
...
+  },
+
+  // We're doing bloom at half resolution for performance. Most blurry effects can get away with that.
+  getTargets: function () {
+    var srt = this.el.sceneEl.renderTarget;
+    for (var i = 0; i < 2; i++) {
+      var rt = this.renderTargets[i];
```
fernandojsg: renderTarget and sceneRenderTarget less cryptic

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,161 @@
...
+    this.scenePost.overrideMaterial = this.brightMaterial;
+    uns.threxp.value = this.data.threxp;
+    uns.srcTexture.value = rt;
+    rt = rts[0]; // Output to the first of the ping pong targets.
+    scene.renderer.render(this.scenePost, this.cameraPost, rt);
+
+    // Ping pong the render targets progressively blurring more on each pass.
+    uns = this.blurMaterial.uniforms;
+    this.scenePost.overrideMaterial = this.blurMaterial;
+    var passes = this.data.passes.split(' ');
```
fernandojsg: probably define `this.passes` on the component and update just when the component's attribute `passes` changes, instead of splitting the string on every tock

examples/test/post-processing/components/bloom.js
```diff
@@ -0,0 +1,161 @@
...
+    for (var i = 0; i < passes.length; i++) {
```
fernandojsg: Do we really need `passes` to be a sequential string `"1 2 3 4 5..."`? maybe we could just have `num_passes`? and loop until that number?

src/core/scene/a-scene.js
```diff
@@ -320,7 +364,16 @@ module.exports = registerElement('a-scene', {
         if (this.isPlaying) {
           this.tick(time, timeDelta);
         }
-        this.effect.render(this.object3D, camera);
+        // Postproc must be explicitly enabled by a component or system.
```
fernandojsg: Instead of using `enablePostProcessing` that is specific from the `postprocess` component, just call `this.tock` all the time, if the scene doesn't have any `tock` behaviours it won't execute anything anyway

examples/test/post-processing/components/bloom.js
wizgrav: we could but kawase bloom is supposed to be tunable with these params. Using different not linearly incrementing values would result in a different kind of bloom, check the intel article which explains it. In practice it doesn't give very good results for small blur kernels, even though it's very fast with large ones. A standard sepparable gaussian would be more straightforward, should I revert to that? https://software.intel.com/en-us/blogs/2014/07/15/an-investigation-of-fast-real-time-gpu-based-image-blur-algorithms

wizgrav:
@fernandojsg I've updated the compositor to use VREffect instead of the vanilla WebGLRenderer for outputting to screen, can you check the examples now? Unfortunately I don't have an hmd available currently. I hacked a bit on webvr-polyfill and vr-mode-ui to enable vr on desktop and got an image that looked proper but failed to make the view rotate. I'll check on a mobile later on

src/core/scene/a-scene.js
wizgrav: the final render to screen happens in a compositor system, the setup now is more akin to @donmccurdy physics where a proxy component controls the system. This approach is actually more proper as there should be only one place where the screen is drawn. Hence the need to have a way to explicitly toggle post proc. I also find it more convenient as otherwise you'd have to remove and readd several components to check the base render vs the post proced one

wizgrav:
@fernandojsg I patched VREffect for hmd frame submission to be performed manually. I haven't updated this PR with the changes, which will also have to be propagated to three.js. Please check the examples to see if it fixes the issue.

http://wizgrav.github.io/aframe/examples/test/post-processing/
http://wizgrav.github.io/aframe/examples/test/post-processing-grain/

fernandojsg:
@wizgrav I've just tried the new version and it's still painting just one eye. I'll try to debug it this afternoon!

wizgrav:
@fernandojsg I think the issue with the single eye rendering was that the renderer's viewport was left on the second half of the frame in VREffect.render(). I've now updated the live examples to deal with this as well

fernandojsg:
Now both eyes works, but it looks weird. The left eye looks at the front but the texture is stretched and the right eye looks like the camera is rotated to the left:

wizgrav:
Ok at least we're getting somewhere, this seems like a size mismatch between renderer and renderTarget at some point in the pipeline. We can debug it further later when I get back from work, if you're available on slack

fernandojsg:
@wizgrav sure! just drop me a line on slack and we'll do it!

wizgrav:
The last commit adds several stuff in VREffect, notably the ability to manually submit the frame to the hmd and also graceful access to the hmd.requestAnimationFrame to take advantage of the higher fps. An alternative approach would be a single method to get the hmd object back but having VREffect handle the functionality seems more natural. The manual submission flag will be necessary in any case. The frame has to be submitted after post proc has finished.

Render target size setting is done more tightly now, I'd appreciate some feedback from people with hmds to see how it actually works (I don't currently have access on one)

examples/test/post-processing/components/compositor.js
ngokevin: Could probably take it out to be more tidy. Anyone is free to add console logs if they're writing their own effects.

src/core/scene/a-scene.js
ngokevin: What about a `post-processing` component that when attached, does everything the `enablePostProcessing` does?

Seems `this.tock` should be available without post-processing as well.

ngokevin:
Would still be easier to review if it was broken down further like:

1. tock (purely)
2. post-processing
3. graph-ordered execution

src/core/scene/a-scene.js
wizgrav: At minimum there needs to be a way to instruct a-scene to render to texture instead of directly to the canvas. Tocks are meant to execute after the base render, so they're not really useful for updating the scene at that point. We could go with @fernandojsg suggestion and just check for the presence of tock components but then the developer would have to constantly add and remove a bunch of components. The .enablePostProcessing flag is purely for convenience

ngokevin: The scene could check whether that post-processing component is attached to the scene rather than having to set it in the PP components. 

Though you'd need to add a standard post-processing component, but it might be easier to do on a separate PR and purely just add the tock logic in for now?

wizgrav:
@ngokevin the compositor is better implemented as a system. This is more in line with existing component sets like @donmccurdy physics. Besides being more logically correct, there can be only one compositor in the scene, It also aleviates the need for ordered execution for post processing.

I also thought that we could get away with a two liner for having a minimum of ordering guarantees. Something like defining an "important" boolean flag on components that would switch between unshift() and push() to the behaviors.tick/tock arrays in a-scene.addBehavior. This would at least provide us with the level of flexibility unity has with LateUpdate() with minimal code impact. Or we could go to the more complex graph based approach. But post processing doesn't currently need any of these approaches, it can work as it is using the component/system combo.

Right now I'm extending VREffect for submission to three.js. There are several things needed in there to actually work in hmds. Some changes will be needed in a-scene as well, most notably tapping the .requestAnimationFrame on the hmd instead of the window to enable the higher fps on the headsets. So I think we'll have to first see if/what they accept before carrying on

src/core/scene/a-scene.js
wizgrav: you mean having something like the debug component and checking for it's presence? In the current design there is a post-process component who's purpose is to pass the order of application for the post processing effects to the compositor system as depicted here http://wizgrav.github.io/aframe/examples/test/post-processing-simple/ . Mind that the aframe.js used by the example is patched with the updated VREffect that works with actual hmds (not with the higher fps though)

wizgrav: I will take out the example from this PR, the design is pretty outdated. I'll just leave the logic and put examples in a different repo for inspection

wizgrav:
Added update() for systems to allow dynamic reconfiguration of the compositor. Took out the example for now, you can check out one with the current dist here http://wizgrav.github.io/aframe/examples/test/post-processing/

ngokevin:
@wizgrav @fernandojsg ready for review right?

wizgrav:
@fernandojsg  @ngokevin  should a simple example be included in this PR as well?

ngokevin:
Not an example, but could use unit tests for adding/remove `tock` and that components that define `tock` have those behaviors registered on the scene.

src/components/index.js
```diff
@@ -24,5 +24,6 @@ require('./scene/embedded');
 require('./scene/inspector');
 require('./scene/fog');
 require('./scene/keyboard-shortcuts');
+require('./scene/post-processing');
```
ngokevin: Maybe just `postprocessing` without hyphen.

src/core/scene/a-scene.js
```diff
@@ -403,13 +461,28 @@ module.exports = registerElement('a-scene', {
...
 
-        this.animationFrameID = effect.requestAnimationFrame(this.render);
-        effect.render(this.object3D, this.camera);
-        this.time = time;
+        window.performance.mark('render-iteration-started');
+        if (this.getAttribute('post-processing')) {
+          var size = this.renderer.getSize();
+          var renderTarget = this.renderTarget;
+          if (size.width !== renderTarget.width || size.height !== renderTarget.height) {
+            renderTarget.setSize(size.width, size.height);
```
ngokevin: Does setSize have to be done on every frame? Can we listen for a size change instead?

src/components/index.js
wizgrav: sure sounds better, and an ever shorter form could be nice, it seems too long as it is. Maybe something like deffered?

src/core/scene/a-scene.js
wizgrav: It does a check first so setSize won't actually happen every frame. The reason I went that route, checking every frame, was because VREffect will switch the framebuffer(canvas) size internally to fit the hmd without providing any feedback

src/components/index.js
ngokevin: `postprocessing` is clearest to me, i wouldn't know `deferred` unless i had the context. just wanted to rm the hyphen so you can do `sceneEl.systems.postprocessing` :)

src/components/scene/post-processing.js
```diff
@@ -0,0 +1,5 @@
+var register = require('../../core/component').registerComponent;
+
+module.exports.Component = register('post-processing', {
```
ngokevin: should this be a system?

fernandojsg:
@ngokevin @wizgrav do you mind if I leave this review to next week as I'm quite busy right now? I'll give it a try and try to use in another context with other effects to see how it will scale

AnnCaryn:
Was wondering if it might be faster to review if we unbundle the post processing effects - with each PR having its own effect.  Bloom, film grain, technicolor etc - kind of like http://stemkoski.github.io/Three.js/#shader-glow

wizgrav:
@AnnCaryn that won't be necessary, the effects themselves are not included in the PR, just the core functionality to enable them. The examples posted in this thread are not included either, they just reside on my gh-pages for reference. Components providing the various effects will be provided by third parties after the core is enabled

fernandojsg:
@AnnCaryn Yes, the effects probably will be added to the https://aframe.io/aframe-registry so people could register their own posteffects and then you could easy include them in your project, they won't belong to the core, just the base functionality, as @wizgrav explained.

wizgrav:
@fernandojsg have in mind this pull request as well https://github.com/mrdoob/three.js/pull/9848 this will be needed for some effects that need per eye calculations, like godrays which need to calc the 2D sun position from the camera matrix. There is no easy way around it with the latest WebVR spec because these matrices will come directly from vrDisplay. I think it does not pose a problem with this PR as some solution will have to be included in VREffect anyway and that will trickle back into AFrame. The comments in that pull request are interesting though as VREffect may become redundant altogether in future three.js releases. Some thought should be put in that prospect so it doesn't catch us off guard

wizgrav:
@ngokevin @dmarcos @fernandojsg rebased the PR. A minor check slipped in VREffect. This was needed to pass test "a-scene (with renderer)". It seems that at some point it calls VREffect.cancelAnimationFrame which in turn tries to call the non existent vrDisplay.cancelAnimationFrame. I suspect this is an issue with webvr-polyfill and should go away with an update of that. In the meantime the check is really minor and won't affect anything else

cvan:
@wizgrav @ngokevin is this change still something that is desired to be merged?

johnnyshankman:
@cvan I'm dying for post processing support personally and would love to see this get merged soon.

src/core/scene/a-scene.js
```diff
@@ -312,6 +343,11 @@ module.exports = registerElement('a-scene', {
         renderer.setPixelRatio(window.devicePixelRatio);
         renderer.sortObjects = false;
         this.effect = new THREE.VREffect(renderer);
+        this.effect.autoSubmitFrame = false;
+
+        // Create a render target for post processing.
+        this.renderTarget = new THREE.WebGLRenderTarget(1, 1, { minFilter: THREE.LinearFilter, magFilter: THREE.NearestFilter, format: THREE.RGBAFormat });
+        this.renderTarget.texture.generateMipmaps = false;
```
johnnyshankman: Seems like this could be be lazily instantiated either here or on line L484 only when `this.getAttribute('postprocessing')` is set to `true` instead of always instantiated right here.

Seems inefficient unless this is a scoping/ease-of-use thing because `this.renderTarget` is never used otherwise.

wizgrav:
@johnnyshankman in the meantime you can easily put in some postproc by overriding scene.render() check an example here https://github.com/wizgrav/copernicus/blob/master/src/main.js line 230 is where it's done

src/core/scene/a-scene.js
wizgrav: yeah it could even though the framebuffer is created only on the scene element so probably only once in the whole app.

ngokevin:
Since this thread got bumped, the update for post-processing is there's pending work in three.js to get it to work well for VR (a camera for each eye). The `tock` looks OK, I've since merged the System update pieces.

wizgrav:
@ngokevin I can rebase the PR to the current master in the w/e if everyone's ok with the functionality

johnnyshankman:
@wizgrav Having trouble getting the scene to double across the two eyes instead of spread across the two eyes but probably an implementation error on my side. Not sure what I did differently.

screen shot 2017-04-20 at 3 17 43 pm

Anyways though yes I'm super happy with this PR so I'll be following along to see where this goes. Thank you @wizgrav and @ngokevin!

johnnyshankman:
@ngokevin Is that related to this PR?

vendor/VREffect.js
```diff
@@ -260,25 +262,43 @@ THREE.VREffect = function ( renderer, onError ) {
...
+			var layers = vrDisplay.getLayers();
+			var leftBounds;
+			var rightBounds;
+
+			if ( layers.length ) {
+
+				var layer = layers[ 0 ];
+
+				leftBounds = layer.leftBounds !== null && layer.leftBounds.length === 4 ? layer.leftBounds : defaultLeftBounds;
+				rightBounds = layer.rightBounds !== null && layer.rightBounds.length === 4 ? layer.rightBounds : defaultRightBounds;
```
johnnyshankman: I had to do the following to get this to avoid getting `length` of `undefined` errors when testing on iPhone or Chrome's mobile simulator.

```
leftBounds = layer.leftBounds && layer.leftBounds.length === 4 ? layer.leftBounds : defaultLeftBounds;
rightBounds = layer.rightBounds && layer.rightBounds.length === 4 ? layer.rightBounds : defaultRightBounds;
```

johnnyshankman:
@wizgrav @ngokevin I got a little antsy and rebased this branch locally with aframe's current master. Additionally, I made that one fix with `layer.leftBounds.length`. After those two things this branch runs really well 👍 

I don't have push rights to wiz's fork of aframe so I can't push my rebased+modified branch to the `tock` branch myself unfortunately.

wizgrav:
@ngokevin @johnnyshankman I don't know why the null/undefined issue happens so I set the test against a generic false value, I don't expect any sideffects from that

src/components/scene/postprocessing.js
```diff
@@ -0,0 +1,5 @@
+var register = require('../../core/component').registerComponent;
```
dmarcos: Can we keep the `tock` related code and move the postprocessing component to a different PR?

src/core/scene/a-scene.js
dmarcos: Can we move this postprocessing logic to a different PR?

src/core/scene/a-scene.js
```diff
@@ -389,9 +412,9 @@ module.exports = registerElement('a-scene', {
                 window.performance.mark('render-started');
               }
               sceneEl.clock = new THREE.Clock();
-              sceneEl.render();
               sceneEl.renderStarted = true;
               sceneEl.emit('renderstart');
+              sceneEl.render();
```
dmarcos: This is incorrect now. The event should be emitted after rendering has started.

src/core/scene/a-scene.js
```diff
@@ -467,11 +512,29 @@ module.exports = registerElement('a-scene', {
...
+        var camera = this.camera;
         var delta = this.clock.getDelta() * 1000;
         this.time = this.clock.elapsedTime * 1000;
+
         if (this.isPlaying) { this.tick(this.time, delta); }
+
         this.animationFrameID = effect.requestAnimationFrame(this.render);
-        effect.render(this.object3D, this.camera);
+
+        window.performance.mark('render-iteration-started');
```
dmarcos: Can we move this post processing code to a different PR?

dmarcos:
Thanks for the patience. I would like to keep the `tock` related logic and move the postprocessing logic to a different PR. Can we do that?

src/components/scene/postprocessing.js
wizgrav: We can just take it out, the post processing logic is triggered with .getAttribute("postprocessing") so the component is in fact not needed

src/core/scene/a-scene.js
wizgrav: Ok I had this to notify the post proc system that the rendertargets are ready in case it wants to set a depth texture. That was when aframe was using an older version of three, now we can just attach the depth texture on the render target when creating them

src/core/scene/a-scene.js
wizgrav: You mean the timings with performance.mark? I so, sure these should be done in a timing PR

wizgrav:
@dmarcos I'm not sure I understand. You mean keeping the tock infrastructure in but not actually calling it in render()?

src/core/scene/a-scene.js
dmarcos: I eas refering to the lines below tha check for the posproceesing component and do renderTarget stuff

wizgrav: so should I just leave in the renderTarget creation in setupRenderer and the tock definition in component? the render to texture is straightforward, there's no other way to do it. If you're weary about the renderTarget size checking and resize, unfortunately that's necessary with the current way VREffect works(doesn't hint when it changes the canvas size when going in/out of VR)

dmarcos: I see the tock infrastructure as a way to introduce logic after rendering the scene. It is useful beyond postprocessing. I would make a separate PR that introduces all the postprocessing pieces on top of the `tock` work that is postprocessing agnostic

dmarcos: I would start by just doing (as analog to the tick):
````javascript
effect.render(this.object3D, camera, null);
if (this.isPlaying) { this.tock(this.time, delta): }
````
`

wizgrav: I can't readily think of any logic besides post processing that would need to happen after the scene rendering. Having tick for scene updates and tock for post proc is clean, if tock where to be generic it would lead to new issues like having to check in components whether post processing took place. Without these checks the final compositor would overwrite the screen with redundant frames

dmarcos:
@wizgrav A use case for example we needed this for (and we hacked temporarly) is the ability to capture an animated gif of a running scene.

dmarcos:
@wizgrav we would call `tock` in the render function the same way we call `tick`. This is what the render method would look like (renderTarget is null by default)
```javascript
render: {
      value: function () {
        var effect = this.effect;
        var camera = this.camera;
        var delta = this.clock.getDelta() * 1000;
        this.time = this.clock.elapsedTime * 1000;

        if (this.isPlaying) { this.tick(this.time, delta); }

        this.animationFrameID = effect.requestAnimationFrame(this.render);
        effect.render(this.object3D, camera, renderTarget);

        if (this.isPlaying) { this.tock(this.time, delta); }

        this.effect.submitFrame();
        window.performance.mark('render-iteration-finished');
}
````

wizgrav:
@dmarcos I understand your desire to keep something so core as generic as possible, The scene video recording would still need the scene rendered to the renderTarget but I understand why you want to keep something so core as simple and generic as possible, I'm thinking more from a convenience standpoint..

Maybe we could add a boolean property to the component, to denote that it shouldn't execute when the renderTarget is not enabled from the frame. Then the "postprocessing" attribute check would just toggle the renderTarget on/off. What do you think of this scheme?

dmarcos:
I would not check for the `postprocessing` attribute inside `a-scene`. The postprocessing logic should be decoupled from the scene. The postprocessing component could instantiate the renderTarget for the scene and you can move the logic to the `tick` method (that happens before render) so nothing specific to postProcessing remains in the render loop. We might want to consider a separate scene component called renderTarget that instantiates the renderTarget for those components that require one.

wizgrav:
The scene.renderTarget setting could also get done by the final compositor system itself utilizing both tick and tock. It sounds ok

wizgrav:
@dmarcos removed renderTarget handling from scene, that can now be set by a component/system, example here https://wizgrav.github.io/aframe-clubber/demo/

One minor thing I found was that system update is not called when changing attributes from the dom inspector, only with setAttribute(). I used to override setEntityAttribute to do that but took it out cause it caused a test to fail. This is unrelated to tock though and could be fixed in another PR.

dmarcos:
Thanks a lot for the patience. Longest running PR in aframe history. It need a rebase.

wizgrav:
@dmarcos rebased and also proposed mrdoob/three.js#9848 to threejs. It's as minimal as it can be and If accepted, it will solve all issues concerning post processing in VR(proper godrays and such).

wizgrav:
@dmarcos just saw that ArrayCamera has landed in three dev and that is indeed the best way to handle it, so I'll wait for when AFrame utilizes WebVRCamera to get the twin projection matrices after the scene renders. In the meantime tock is ready and sufficient for most effects

dmarcos:
Yeah! Thanks very much for sticking with it.

dmarcos:
@wizgrav one last thing. I see changes on VREffect. Are those coming from the three repo? Do we need then!

wizgrav:
@dmarcos yes I made the checks for layer.leftBounds and layer.rightBounds less strict as  @johnnyshankman reported that they were undefined instead of null which the original code checked against. I hadn't experienced that myself tbh so I can take it out and if encountered again we can open a new issue. Though I don't expect no side effects even if we leave it in

wizgrav:
@dmarcos reverted VREffect. Also updated the tock related tests

dmarcos:
Amazing! it only took a year to land this 😄 @wizgrav It's fantastic you were able to bring it back home

PR #1565, opened 2016-06-18
Component play should not execute if the component is not initialized

dmarcos:
**Description:**

There's a bogus situation if `play` is called on a component that has not been yet initialized and the logic relies on something done in the `init` method. 

The bug has manifested when the camera system pauses the inactive camera entity and plays the new active one. With the default camera the camera system calls play on the entity. The camera component is initialized before `look-controls` and its `play` method invoked before `init`. The `play` method of `look-controls` expects the methods to be bound before attaching the event listeners. 

**Changes proposed:**
Ensure that `play` is never invoked before `init` and also prevent play/pause logic to be executed multiple times if the component is already playing or paused respectively.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -203,6 +203,8 @@ Component.prototype = {
     if (!this.initialized) {
       this.init();
       this.initialized = true;
+      // Play the component if the entity is playing.
+      if (el.isPlaying) { this.play(); }
     }
     this.update(oldData);
 
@@ -259,10 +261,13 @@ module.exports.registerComponent = function (name, definition) {
   NewComponent = function (el, attr) {
     Component.call(this, el, attr);
   };
+
   NewComponent.prototype = Object.create(Component.prototype, proto);
   NewComponent.prototype.name = name;
   NewComponent.prototype.constructor = NewComponent;
   NewComponent.prototype.system = systems && systems.systems[name];
+  NewComponent.prototype.play = wrapPlay(NewComponent.prototype.play);
+  NewComponent.prototype.pause = wrapPause(NewComponent.prototype.pause);
 
   components[name] = {
     Component: NewComponent,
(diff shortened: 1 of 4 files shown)


src/core/component.js
```diff
@@ -259,10 +260,14 @@ module.exports.registerComponent = function (name, definition) {
   NewComponent = function (el, attr) {
     Component.call(this, el, attr);
   };
+
   NewComponent.prototype = Object.create(Component.prototype, proto);
   NewComponent.prototype.name = name;
   NewComponent.prototype.constructor = NewComponent;
   NewComponent.prototype.system = systems && systems.systems[name];
+  // Wraps method to meet needed gurantees
```
ngokevin: `Wraps methods to meet needed guarantees.`

Can also explain what the "guarantees" are.

src/core/component.js
```diff
@@ -203,6 +203,7 @@ Component.prototype = {
     if (!this.initialized) {
       this.init();
       this.initialized = true;
+      if (el.isPlaying) { this.play(); }
```
ngokevin: Sort of confusing since seems like it should read `should be playing` rather than `is playing`. A comment might help.

src/core/component.js
```diff
@@ -346,3 +351,40 @@ function extendProperties (dest, source, isSinglePropSchema) {
   if (isSinglePropSchema) { return source; }
   return utils.extend(dest, source);
 }
+
+/**
+ * Pause component by removing tick behavior and calling pause handler.
+ *
+ * @param component {object} - Component to pause.
```
ngokevin: Incorrect parameters in docstring.

src/core/component.js
```diff
@@ -346,3 +351,40 @@ function extendProperties (dest, source, isSinglePropSchema) {
...
+ * @param sceneEl {Element} - Scene, needed to remove the tick behavior.
+ */
+function wrapPause (pauseMethod) {
+  return function pause () {
+    var sceneEl = this.el.sceneEl;
```
ngokevin: Could document `@this` is the component

src/core/component.js
```diff
@@ -346,3 +351,40 @@ function extendProperties (dest, source, isSinglePropSchema) {
...
+/**
+ * Play component by adding tick behavior and calling play handler.
+ *
+ * @param component {object} - Component to play.
+ * @param sceneEl {Element} - Scene, needed to add the tick behavior.
+ */
+function wrapPlay (playMethod) {
+  return function play () {
+    var sceneEl = this.el.sceneEl;
+    var canPlay = this.el.isPlaying && !this.isPlaying;
```
ngokevin: Probably `shouldPlay`

tests/core/component.test.js
```diff
@@ -343,4 +343,87 @@ suite('Component', function () {
       assert.equal(HTMLElement.prototype.getAttribute.call(el, 'dummy'), 'color:blue');
     });
   });
+
+  suite('play', function () {
+    setup(function () {
+      components.dummy = undefined;
+      var playStub = this.playStub = sinon.stub();
+      registerComponent('dummy', {
+        schema: {color: { default: 'red' }},
```
ngokevin: don't really need to define a schema

src/core/component.js
dmarcos: Can you rephrase? What's the confusing part? What the line does it's to call the play method on the component if the corresponding entity is playing.

ngokevin: The confusing part is that `isPlaying` implies that it's already playing, and therefore we wouldn't need to call `play()`.

ngokevin: Ah never mind, I confuse entity playing vs component playing.

dmarcos: `el` referes to the entity and `this` to the component

dmarcos: for an entity that is playing we have to play its components

ngokevin: Wouldn't hurt to comment anyways.

```Play the component if the entity is playing.```

PR #1566, opened 2016-06-19
Multiple components

dmarcos:
**Description:**

It allows defining multiple component of a specific kind by appending an id to the component name followed by a `__` character. e.g:

````javascript
el.setAttribute('sound__1', { src: '...' });
el.setAttribute('sound__explosion', { src: '...' });
````
Components have to opt into multiplicity otherwise an error is thrown.

This PR depends on https://github.com/aframevr/aframe/pull/1565

Needs feedback from @fernandojsg

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -16,17 +16,20 @@ module.exports.Component = registerComponent('sound', {
     volume: { default: 1 }
   },
 
+  multiple: true,
+
   init: function () {
+    var id = this.id;
     this.listener = null;
     this.sound = null;
+    this.object3DName = id ? 'sound__' + id : 'sound';
+    this.playSound = this.playSound.bind(this);
   },
 
   update: function (oldData) {
     var data = this.data;
-    var el = this.el;
     var sound = this.sound;
     var srcChanged = data.src !== oldData.src;
-
     // Create new sound if not yet created or changing `src`.
     if (srcChanged) {
       if (!data.src) {
@@ -41,22 +44,47 @@ module.exports.Component = registerComponent('sound', {
     sound.setVolume(data.volume);
 
     if (data.on !== oldData.on) {
-      if (oldData.on) { el.removeEventListener(oldData.on); }
-      el.addEventListener(data.on, this.play.bind(this));
+      this.addEventListener(oldData.on);
     }
 
     // All sound values set. Load in `src`.
     if (srcChanged) { sound.load(data.src); }
   },
 
+  addEventListener: function (oldEvt) {
+    var el = this.el;
+    if (oldEvt) { el.removeEventListener(oldEvt, this.playSound); }
+    el.addEventListener(this.data.on, this.playSound);
+  },
+
+  removeEventListener: function () {
+    this.el.removeEventListener(this.data.on, this.playSound);
+  },
+
   remove: function () {
-    this.el.removeObject3D('sound');
+    this.el.removeObject3D(this.object3DName);
     try {
       this.sound.disconnect();
     } catch (e) {
       // disconnect() will throw if it was never connected initially.
       warn('Audio source not properly disconnected');
     }
+  },
+
+  play: function () {
+    if (!this.sound) { return; }
+    if (this.sound.source.buffer && this.data.autoplay) {
+      this.sound.play();
+    }
+    this.addEventListener();
+  },
+
+  pause: function () {
+    if (!this.sound) { return; }
+    if (this.sound.source.buffer && this.sound.isPlaying) {
+      this.sound.pause();
+    }
+    this.removeEventListener();
   },
 
   /**
@@ -70,7 +98,7 @@ module.exports.Component = registerComponent('sound', {
     var sound = this.sound;
 
     if (sound) {
-      this.stop();
+      this.stopSound();
       el.removeObject3D('sound');
     }
 
@@ -88,7 +116,7 @@ module.exports.Component = registerComponent('sound', {
     });
 
     sound = this.sound = new THREE.PositionalAudio(listener);
-    el.setObject3D('sound', sound);
+    el.setObject3D(this.object3DName, sound);
 
     sound.source.onended = function () {
       sound.onEnded();
@@ -98,18 +126,13 @@ module.exports.Component = registerComponent('sound', {
     return sound;
   },
 
-  play: function () {
+  playSound: function () {
     if (!this.sound.source.buffer) { return; }
     this.sound.play();
   },
 
-  stop: function () {
+  stopSound: function () {
     if (!this.sound.source.buffer) { return; }
     this.sound.stop();
-  },
-
-  pause: function () {
-    if (!this.sound.source.buffer || !this.sound.isPlaying) { return; }
-    this.sound.pause();
   }
 });
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -278,7 +278,11 @@ var proto = Object.create(ANode.prototype, {
   initComponent: {
     value: function (name, data, isDependency) {
       var component;
+      var componentInfo = name.split('__');
+      var id = componentInfo[1];
       var isComponentDefined = checkComponentDefined(this, name) || data !== undefined;
+      var attrName = name;
+      name = componentInfo[0];
 
       // Check if component is registered and whether component should be initialized.
       if (!components[name] || (!isComponentDefined && !isDependency)) {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ngokevin** [comment]
The overall API looks good to me. We might consider a double underscore to further make it look different from normal component name hyphens (like OOCSS namespaces).

**Comment by fernandojsg** [comment]
I agree with @ngokevin having double underscore maybe looks uglier but it will make it easier to differentiate between normal components and multiple instances

**Comment by fernandojsg** [comment]
I'm trying this PR on the editor. Please share what you think about the way to show this feature: https://github.com/aframevr/aframe-editor/issues/72 /cc @ngokevin @dmarcos

**Comment by fernandojsg** [comment]
I've been playing with your example `examples/test/cube`:
Querying each component return the expected results:
screenshot 2016-06-29 11 03 07

But when I try to change one instance's attribute something weird happens: it generates a new `sound` component, copying the same values from `sound_2` but the volume is being copied as `string` instead of `number`.
screenshot 2016-06-29 11 04 59

**Comment by fernandojsg** [comment]
It fails just on `debug=true` as in the `flushToDOM` you're using just the `this.name` for the component (sound) instead of the whole `component_id` name:
This:
```javascript
  flushToDOM: function () {
    var attrValue = this.attrValue;
    if (!attrValue) { return; }
    HTMLElement.prototype.setAttribute.call(this.el, this.name, this.stringify(attrValue));
  },
```
Should be something like:
```javascript
  flushToDOM: function () {
    var attrValue = this.attrValue;
    if (!attrValue) { return; }
    var nameId = this.name + (this.id ? '_' + this.id : '');
    HTMLElement.prototype.setAttribute.call(this.el, nameId, this.stringify(attrValue));
  },
```

I'm thinking also about `componentChanged` event:
```javascript
    el.emit('componentchanged', {
      name: this.name,
      newData: this.getData(),
      oldData: oldData
    });
```

With the current implementation it will emit just `sound` name event without any additional information, we should add the full name `sound_{id}` or at least and additional `id` parameter

src/components/sound.js
```diff
@@ -16,17 +16,20 @@ module.exports.Component = registerComponent('sound', {
     volume: { default: 1 }
   },
 
+  multiple: true,
+
   init: function () {
+    var id = this.id;
     this.listener = null;
     this.sound = null;
+    this.object3DName = id ? 'sound-' + id : 'sound';
```
fernandojsg: why don't we use here the same style as the component's name? `sound__`

src/core/component.js
```diff
@@ -245,6 +251,12 @@ module.exports.registerComponent = function (name, definition) {
   var NewComponent;
   var proto = {};
 
+  if (name.indexOf('_') !== -1) {
```
fernandojsg: Are we preventing using single underscore too? Why not check for `indexOf('__')`

src/core/a-entity.js
```diff
@@ -278,7 +278,11 @@ var proto = Object.create(ANode.prototype, {
   initComponent: {
     value: function (name, data, isDependency) {
       var component;
+      var componentInfo = name.split('__');
+      var id = componentInfo[1];
```
ngokevin: Won't this error? `'geometry'.split('__')[1]`

src/components/sound.js
```diff
@@ -16,17 +16,20 @@ module.exports.Component = registerComponent('sound', {
     volume: { default: 1 }
   },
 
+  multiple: true,
+
   init: function () {
+    var id = this.id;
     this.listener = null;
     this.sound = null;
+    this.object3DName = id ? 'sound__' + id : 'sound';
```
ngokevin: We're just getting the attribute name. Is there an easier way? Like `this.name` or `this.fullName`.

src/components/sound.js
```diff
@@ -41,16 +44,25 @@ module.exports.Component = registerComponent('sound', {
...
-      if (oldData.on) { el.removeEventListener(oldData.on); }
-      el.addEventListener(data.on, this.play.bind(this));
+      this.addEventListener(oldData.on);
     }
 
     // All sound values set. Load in `src`.
     if (srcChanged) { sound.load(data.src); }
   },
 
+  addEventListener: function (oldEvt) {
```
ngokevin: could use docstring to explain that it helps handle updates to `on`

ngokevin: `updateEventListener` might be more accurate

src/core/a-entity.js
ngokevin: Name `componentId` might help differentiate from just HTML ID.

fernandojsg: it will return `undefined`, we could leave it that way or do something like `componentInfo[1] || null`. In any case it should works as expected on the rest of the code where the full name is built using: `componentName = id ? component + '__' + id : component` and both `undefined` or `null` will return false.

src/core/a-entity.js
```diff
@@ -290,19 +294,23 @@ var proto = Object.create(ANode.prototype, {
 
       // Check if component already initialized.
       if (name in this.components) { return; }
-      component = this.components[name] = new components[name].Component(this, data);
+      if (id && !components[name].multiple) {
+        throw new Error('Trying to initialize multiple components of type ' + name +
```
ngokevin: Maybe visual indent + backticks around component name + period:

```js
throw new Error('Trying to initialize multiple components of type `' + name + '`. ' +
                'There can only be one component of this type per entity.');
```

src/core/a-entity.js
dmarcos: No it just returns `undefined`

fernandojsg: umm, maybe `instanceId` ? and the `componentId` could be the whole name `componentName__instanceId`?

src/core/a-entity.js
```diff
@@ -359,7 +367,8 @@ var proto = Object.create(ANode.prototype, {
 
       // add component to the list
       function addComponent (key) {
-        if (!components[key]) { return; }
+        var name = key.split('_')[0];
```
ngokevin: Double underscore, make sure we have test coverage here

src/core/component.js
```diff
@@ -177,8 +178,10 @@ Component.prototype = {
    */
   flushToDOM: function () {
     var attrValue = this.attrValue;
+    var attrName;
     if (!attrValue) { return; }
-    HTMLElement.prototype.setAttribute.call(this.el, this.name, this.stringify(attrValue));
+    attrName = this.name + (this.id ? '__' + this.id : '');
```
ngokevin: Starting to seem more useful to store the entire attribute name rather than the raw ID.

src/core/component.js
```diff
@@ -245,6 +251,12 @@ module.exports.registerComponent = function (name, definition) {
   var NewComponent;
   var proto = {};
 
+  if (name.indexOf('__') !== -1) {
+    throw new Error('The component name `' + name + '` is not allowed. ' +
+                    'The sequence __ (double underscore) is reserved and cannot be used ' +
```
ngokevin: Would update to `The sequence `__` (double underscore) is reserved for multiple components of the same type.` since  the "cannot be used" is sort of restating the first sentence

src/core/component.js
```diff
@@ -25,8 +25,9 @@ var styleParser = utils.styleParser;
  * @member {object} el - Reference to the entity element.
  * @member {string} name - Component name exposed as an HTML attribute.
  */
-var Component = module.exports.Component = function (el, attr) {
+var Component = module.exports.Component = function (el, attr, id) {
```
ngokevin: Would call `componentId` here too to not confuse with entity ID.

tests/core/a-entity.test.js
```diff
@@ -549,6 +549,24 @@ suite('a-entity', function () {
...
+      var el = this.el;
+      assert.throws(function setAttribute () {
+        el.setAttribute('geometry__1', {primitive: 'box'});
+      }, Error);
+      assert.notOk(el.components.geometry_1);
+    });
+
+    test('initializes components with id if it opts in to multiple', function () {
+      var el = this.el;
+      el.setAttribute('sound__1', {'src': 'mysoundfile.mp3'});
```
ngokevin: Need to test without the component ID too

tests/core/a-entity.test.js
```diff
@@ -549,6 +549,24 @@ suite('a-entity', function () {
...
+
+    test('does not initialize with id if the component is not multiple', function () {
+      var el = this.el;
+      assert.throws(function setAttribute () {
+        el.setAttribute('geometry__1', {primitive: 'box'});
+      }, Error);
+      assert.notOk(el.components.geometry_1);
+    });
+
+    test('initializes components with id if it opts in to multiple', function () {
```
ngokevin: Need some `getAttribute` coverage for `multiple` components

tests/core/component.test.js
```diff
@@ -144,6 +144,14 @@ suite('Component', function () {
         done();
       });
     });
+
+    test('cannot be registered if it uses the character _ in the name', function () {
```
ngokevin: double underscore

src/core/component.js
```diff
@@ -177,8 +178,10 @@ Component.prototype = {
    */
   flushToDOM: function () {
```
ngokevin: Needs test coverage

src/core/a-entity.js
ngokevin: Needs test coverage

src/components/sound.js
```diff
@@ -59,6 +71,22 @@ module.exports.Component = registerComponent('sound', {
     }
   },
 
+  play: function () {
```
ngokevin: Would be good to test play/pause too

src/core/a-entity.js
dmarcos: I think the id has to be still easily accessible. Each component is going to handle multiplicity in a different way and it's nice to have access to the id directly without having to parse any string. I can add this.name, this.attrName, this.id to the component.

PR #1568, opened 2016-06-21
Allow material repeat to be float

mkeblx:
**Description:**
Repeat need not be an integer, like underlying threejs material:texture property.

**Changes proposed:**
Change repeat values to parse as float instead of int.


dmarcos:
thanks!

PR #1575, opened 2016-06-23
add three scenes for link traversal

dmarcos:
**Description:**
Creates a showcase to illustrate link traversal, characteristic of the WebVR 1.0 spec. A hover and click effects for the links are still pending.

@cvan I know you have opinions on how pages should handle link traversal while remaining in VR. Let's discus and incorporate those ideas into this PR.

This is a link to the demo if you want to quickly try it out:
http://swimminglessonsformodernlife.com/aframe/examples/showcase/link-traversal/


examples/showcase/link-traversal/sunrise.html
```diff
@@ -0,0 +1,37 @@
...
+        <img id="thumbCity" crossOrigin="anonymous" src="https://media.aframe.io/link-traversal/thumbs/city.png">
+        <img id="thumbMountains" crossOrigin="anonymous" src="https://media.aframe.io/link-traversal/thumbs/mountains.png">
+      </a-assets>
+      <a-link href="index.html" position="-1.5 1.8 -1" src="#thumbHome" highlight></a-link>
+      <a-link href="city.html" position="0 1.8 -1" src="#thumbCity" highlight></a-link>
+      <a-link href="mountains.html" position="1.5 1.8 -1" src="#thumbMountains" highlight></a-link>
+      <a-entity id="sky"
+                 geometry="primitive: sphere; radius: 65;"
+                 material="shader: skyGradient; colorTop: #353449; colorBottom: #BC483E; side: back"></a-entity>
+       <a-entity ground='url: https://media.aframe.io/link-traversal/models/groundSunrise.json'></a-entity>
```
cvan: why single ticket marks?

dmarcos: a problem of having different conventions for different contexts. JS vs HTML

src/components/link.js
```diff
@@ -0,0 +1,39 @@
+var registerComponent = require('../core/component').registerComponent;
+
+module.exports.Component = registerComponent('link', {
+  schema: {
+    url: { default: '' },
```
ngokevin: Maybe `href` to be consistent with the primitive that is abstracting it as with as with `<a>`?

cvan:
can you rebase?

src/components/link.js
cvan: yes, please

cvan: or, rather can we call the component `href`? I know it conflicts with the [`aframe-href-component`](https://github.com/gasolin/aframe-href-component). another approach I thought about, in keeping with the behaviour of 2D/classic anchors, perhaps we could actually wrap the entities in `<a>` elements. any emitted `click` would bubble up to the `<a>` and handle the page navigation (i.e., modifying `window.location` from JS). thoughts?

ngokevin: Wrapping in `<a>` isn't great. It doesn't fit with the framework, is syntactically more complicated (both in HTML, and you can't just synchronously do `setAttribute('link', ..>)` or use mixins/`aframe-mss`, and will cause tons of complications in introducing a non-A-Frame element into the scene graph.

No strong preference for href/link, but `href` might be confusing because people will be passing a URL string to it, when it needs both an `href` and an event name (`href="href: blah.com; on: grab`).

ngokevin: `link="href: foo.bar; on: hit"` is fine. Separate attribute/component isn't consistent with the framework, prescribing an event is inflexible to different use cases.

micahstubbs:
as a developer, I like the idea of `link`,  as it feels a bit more intuitive

dmarcos:
Tests pending

src/systems/link.js
```diff
@@ -0,0 +1,16 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * Tracked controls system.
+ * It maintains a list with the available tracked controllers
+ */
+module.exports.System = registerSystem('link', {
```
ngokevin: I don't think this needs to be a system, it doesn't provide any services to the link component. It can be part of the scene module, which is where all the other WebVR-related handlers live.

examples/showcase/link-traversal/js/components/aframe-tooltip-component.js
```diff
@@ -0,0 +1,372 @@
+/*eslint-disable */
```
ngokevin: We can pull the 3rd party components in from unpkg.com

src/systems/link.js
dmarcos: It does. I should probably initialize the variable but the currently peeked link is referenced on the system

ngokevin: That's fine to put any variables you need here, but the activate/deactivate event handlers belong on the scene module vs. links IMO, they're not link-specific.

src/components/link.js
```diff
@@ -0,0 +1,345 @@
...
+      primitive: 'sphere',
+      radius: 1.5,
+      phiStart: 0,
+      segmentsWidth: 64,
+      segmentsHeight: 64,
+      phiLength: 180,
+      thetaStart: 0,
+      thetaLength: 360
+    });
+
```
ngokevin: can remove a couple of these blank lines to group the entity modifications together

src/components/link.js
```diff
@@ -0,0 +1,345 @@
...
+    el.appendChild(sphereEl);
+  },
+
+  redirect: function () {
+    window.location = this.data.href;
+  },
+
+  /**
+   * The tick handles:
+   * 1. It swaps the plane the represents the portal with a sphere with a hole when the camera is close
```
ngokevin: Don't need to put `It` behind anywhere since it's ambiguous and we can infer from context

src/components/link.js
```diff
@@ -0,0 +1,345 @@
...
+      if (!this.previousQuaternion) { this.previousQuaternion = quaternion.clone(); }
+
+      // If the portal is far away from the user the portal faces the camera
+      if (distance > 20) {
+        object3D.lookAt(cameraWorldPosition);
+      } else { // When the portal is close to the user (camera)
+        cameraPortalOrientation = this.calculateCameraPortalOrientation();
+        // If the user gets very close to the portal it is replaced
+        // by a holed sphere where see can peek inside
+        if (distance < 1.0) {
```
ngokevin: Can we do closer than 1 meter?

ngokevin:
Alright, Nightly enters for me now. Camera had a low height for a bit until I restarted. But I experienced that the enter VR wasn't kicking in after navigation, and the portals are too large/far away (I can't walk up to them, and or poke my head thru because they look huge)

stevel4857:
Very cool. It works in nighlty, nightly actually started working again
after today's oculus upgrade. I noticed that if I start in sunrise.html the
labels for the Oculus controllers overlap.

Steve Luiting
http://sluiting.weebly.com
P: 720.563.9377

ngokevin:
In link.js

`if (previousSelectedLinkEl || selectedLinkEl.components.link === undefined) { return; }`

This threw an error for me `selectedLinkEl` is not yet set.

```
link-controls.js:302 Uncaught TypeError: Cannot read property 'components' of undefined
    at i.onMouseEnter (http://localhost:9005/js/link-controls.js:302:49)
    at Object.module.exports.fireEvent (http://localhost:9005/js/aframe-master.min.js:450:2078)
    at http://localhost:9005/js/aframe-master.min.js:293:2801
    at Array.map (native)
    at HTMLElement.value (http://localhost:9005/js/aframe-master.min.js:293:2772)
    at i.twoWayEmit (http://localhost:9005/js/aframe-master.min.js:202:3787)
    at i.onIntersection (http://localhost:9005/js/aframe-master.min.js:202:3035)
    at HTMLElement.<anonymous> (http://localhost:9005/js/aframe-master.min.js:432:132)
    at Object.module.exports.fireEvent (http://localhost:9005/js/aframe-master.min.js:450:2078)
```

ngokevin:
Still looking into this, but creating the links programmatically with

```js
var LINKS = [
  {href: '?env=arches', image: 'img/arches.png', title: 'Arches'},
  {href: '?env=checkboard', image: 'img/checkerboard.png', title: 'Checkerboard'},
  {href: '?env=egypt', image: 'img/egypt.png', title: 'Egypt'},
  {href: '?env=forest', image: 'img/forest.png', title: 'Forest'},
  {href: '?env=goaland', image: 'img/goaland.png', title: 'Goaland'},
  {href: '?env=goldmine', image: 'img/goldmine.png', title: 'Goldmine'},
  {href: '?env=japan', image: 'img/japan.png', title: 'Japan'},
  {href: '?env=poison', image: 'img/poison.png', title: 'Poison'},
  {href: '?env=threetowers', image: 'img/threetowers.png', title: 'Three Towers'},
  {href: '?env=tron', image: 'img/tron.png', title: 'Tron'},
  {href: '?env=yavapai', image: 'img/yavapai.png', title: 'Yavapai'},
];

AFRAME.registerComponent('generate-links', {
  init: function () {
    LINKS.forEach(link => {
      var linkEl;

      // Current page.
      if (window.location.search.indexOf(link.href.substring(1)) !== -1) { return; }

      linkEl = document.createElement('a-entity');
      linkEl.setAttribute('link', {
        href: link.href,
        src: link.image,
        title: link.title
      });
      this.el.appendChild(linkEl);
    });
  }
});
```

Gives errors like:

```
core:schema:warn Unknown property `0` for component/system `material`. 
aframe-master.min.js:21 core:schema:warn Unknown property `1` for component/system `material`. 
aframe-master.min.js:21 core:schema:warn Unknown property `2` for component/system `material`. 
aframe-master.min.js:21 core:schema:warn Unknown property `3` for component/system `material`. 

aframe-master.min.js:297 Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'strokeColor' in pano
    at i.updateCachedAttrValue (aframe-master.min.js:297)
    at i.updateProperties (aframe-master.min.js:297)
    at HTMLElement.value (aframe-master.min.js:289)
    at HTMLElement.value (aframe-master.min.js:289)
    at i.update (aframe-master.min.js:222)
    at i.updateProperties (aframe-master.min.js:297)
    at HTMLElement.value (aframe-master.min.js:289)
    at e (aframe-master.min.js:289)
    at Array.forEach (<anonymous>)
    at HTMLElement.value (aframe-master.min.js:289)
```

ngokevin:
I think an issue is `el.setAttribute('material', 'pano', data.src.src);` in link.js. If I pass in a raw URL vs `` from `<a-assets>`, then it sort of breaks. This has been an awkward part of the API, we should just have the link component check if it's an `` or string, or hope it just works as a string.

ngokevin:
It doesn't look like the portals are doing `look-at` on the camera

ngokevin:
Should `<a-entity link scale="0.5 0.5 1">` work?

ngokevin:
I think the title text above the portals need double-sided materials in case you go to the other side

ngokevin:
r+

philipardeljan:
Am I doing something wrong or maybe I don't get it, should we be able to "click" on them with mouse/trackpad?

ngokevin:
This is for VR to VR scene traversal.

src/components/link.js
```diff
@@ -0,0 +1,345 @@
...
+var THREE = require('../lib/three');
+
+/**
+ * Link component. Connect experiences and traverse between them in VR
+ *
+ * @member {object} hiddenEls - Stores the hidden elements during peek mode.
+ */
+module.exports.Component = registerComponent('link', {
+  schema: {
+    href: {default: ''},
```
vik66: Hi

PR #1584, opened 2016-06-28
fix fog property for standard material (fixes #1572)

### LINKED ISSUE #1572: `Fog: false` only works with `shader: flat`
**Description:**

I found that `fog: false` wasn't working with the following:

```html
 <a-scene fog="type: linear; color: #AAB; far: 30; near: 0">
    <a-entity geometry="primitive: sphere; radius: 50" material="src: url(../_skies/greencubes.png); side: back; fog: false;"></a-entity>
 </a-scene>
```

However, `<a-sky>` plus `material="fog: false"` *was* working, so I dug into the sky primitive [source](https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives/a-sky.js), and found it was using `shader: flat`

So I added `shader: flat` to my entity, and everything works:

```html
<a-scene fog="type: linear; color: #AAB; far: 30; near: 0">
    <a-entity geometry="primitive: sphere; radius: 50" material="src: url(../_skies/greencubes.png); shader: flat; side: back; fog: false;"></a-entity>
</a-scene>
```

I feel we should call this out in the [docs](https://aframe.io/docs/0.2.0/components/fog.html).

- A-Frame Version: master
- Platform/Device: OSX El Capitan, Chrome 51

ngokevin:
**Description:**

Fog is a base material property. https://github.com/mrdoob/three.js/blob/master/src/materials/Material.js#L15

Wasn't being set on standard material.

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -18,12 +18,13 @@ var shaderNames = shader.shaderNames;
  */
 module.exports.Component = registerComponent('material', {
   schema: {
-    shader: { default: 'standard', oneOf: shaderNames },
-    transparent: { default: false },
-    opacity: { default: 1.0, min: 0.0, max: 1.0 },
-    side: { default: 'front', oneOf: ['front', 'back', 'double'] },
-    depthTest: { default: true },
-    flatShading: { default: false }
+    depthTest: {default: true},
+    flatShading: {default: false},
+    fog: {default: true},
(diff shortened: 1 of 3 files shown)


dmarcos:
I don't think we can assume that fog applies to all materials. The flat and standard shaders happen to consider fog but that's not always the case. This is why I think fog belongs to the shader level.

PR #1586, opened 2016-06-28
Initial .gitattributes to have clean release archives

### LINKED ISSUE #1578: .gitattributes with export-ignore for release archives
Shouldn't this repo use .gitattributes to have bit cleaner release archives?

**Possible contents could be:**
````
# Exclude git files and other files starting with .
.* export-ignore
# As in .npmignore
examples export-ignore
test export-ignore
````
> Besides these suggested above almost everything besides `dist` `LICENSE` and some essential files could be ignored, since that won't affect `npm` and usage of release archives becomes handy only when you want to use browser files in production. As in development you would never use release archives anyway!

- A-Frame Version: *
- Platform/Device: *

### PR DESCRIPTION
**Pull request by mkungla**
This addition would close #1578 

- A-Frame Version: *
- Platform/Device: *

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,6 @@
 # A-Frame .gitattributes
 # Directories and files to be ignored from release archives
-# keep it consistent with .npmignore
-
-tests export-ignore
+# (none) archives are snapshot of entire repo
 
 # Avoid creation of unnecessary big commit objects
 # For these files we do not want to see text diff for
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,16 +1,38 @@
-# Directories to ignore from release archives
-docs export-ignore
-examples export-ignore
-scripts export-ignore
-src export-ignore
-tests export-ignore
-vendor export-ignore
+# A-Frame .gitattributes
+# Directories and files to be ignored from release archives
+# keep it consistent with .npmignore
 
-# Files to ignore from release archives
-.* export-ignore
-*.html export-ignore
-CHANGELOG.md export-ignore
-CONTRIBUTING.md export-ignore
-ORGANIZATION.md export-ignore
-package.json export-ignore
-ROADMAP.md export-ignore
+examples export-ignore
+tests export-ignore
+
+# Avoid creation of unnecessary big commit objects
+# For these files we do not want to see text diff for
+
+*.min.* -diff binary minified
+*js.map* -diff binary
+*.svg -diff binary
+*.mtl -diff binary
+*.obj -diff binary
+*.jpg  -diff binary
+*.jpeg -diff binary
+*.png -diff binary
+*.wav -diff binary
+
+# FIX CRLF always when developer has not set
+# Linux/Mac: git config --global core.autocrlf input
+# Windows: git config --global core.autocrlf true
+# Auto detect text files and perform LF normalization
+* text=auto
+* eol=lf
+
+*.js text
+*.html text
+*.npmignore text
+*.md text
+.ackrc text
+.gitattributes text
+.gitignore text
+.jshintrc text
+.nojekyll text
+.travis.yml text
+LICENSE text
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,16 @@
+# Directories to ignore from release archives
+docs export-ignore
+examples export-ignore
+scripts export-ignore
+src export-ignore
+tests export-ignore
+vendor export-ignore
+
+# Files to ignore from release archives
+.* export-ignore
+*.html export-ignore
+CHANGELOG.md export-ignore
+CONTRIBUTING.md export-ignore
+ORGANIZATION.md export-ignore
+package.json export-ignore
+ROADMAP.md export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,7 +2,6 @@
 # Directories and files to be ignored from release archives
 # keep it consistent with .npmignore
 
-examples export-ignore
 tests export-ignore
 
 # Avoid creation of unnecessary big commit objects

### CODE REVIEW

**Comment by ngokevin** [comment]
@dmarcos r? This will make the archives only have `dist/` + LICENSE + README. I'm mostly okay with it.

**Comment by dmarcos** [comment]
I don't remember having ever used those archives on any of my projects. What's is the expectation when downloading one? Having a snapshot of the whole repo or just the library dist files?

**Comment by ngokevin** [comment]
So after some talking, it seems that while it is nice to have it slightly cleaner, the expectation is that it should serve as a snapshot of the repository. It's a bit more flexible to leave those files in, and follow what other big projects do.

**Comment by mkungla** [comment]
You would still get full snapshot from tag related to release if you want to get snapshot of the repository. 
`$ git checkout tags/<tag_name>`
Isn't point of **release** something which is production ready, which this case is only `dist/*`

Other option would be to use [GitHub API Upload a release asset](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) which enables you to upload extra archives under release downloads like I uploaded, see example aframe-0.2.0-dist.zip at [Example Tag 0.2.0](https://github.com/mkungla/aframe/releases/tag/0.2.0).

Since you can upload release assets any given point then you could upload cleaner dist archives for also for actual 0.2.0 version.

````
curl -u <your-github-username> -XPOST -H "Content-Type:<archive-content-type>" --data-binary @<archive-name> "https://uploads.github.com/repos/aframevr/aframe/releases/<releas-id>/assets?name=<archive-name>"
````

**Comment by dmarcos** [comment]
@mkungla Do you use any project that follows that pattern? Using the archives to only package up the dist files

**Comment by mkungla** [comment]
@dmarcos Yes I do use browser files in cases where it is appropriate/required to use them and these cases I'm only interested on dist files and most annoying is to manually select dist and LICENSE when you really do not want to have other `src` files included your production env, almost all projects which can be used with browser files follow one of the patterns, for instance [twbs/bootstrap - releases](https://github.com/twbs/bootstrap/releases) and [Facebook/react - releases](https://github.com/facebook/react/releases). Most use second method I described here, as you see there is `-dist` archive uploaded for each release. For instance there is also github account [components](https://github.com/components) which is creating dist packages out of popular repositories and started since many of these did not distribute clean archives by themselves, which is slowly changing these days and commonly one or other method is used. Up to date various packages from vendor space [components](https://github.com/components)) were downloaded a lot before original repos started to use cleaner distribution packages. End of the day it comes down to that if your library can be distributed as browser files then release archives should contain only production files and if your library should be used only with npm and bundled for production, then archives don't matter anyway. 
While my personal opinion is that since full snapshot of release can obtained from git tag or npm then archives could be really just distribution files, but if there is any possible argument why One could be interested to find in archive full snapshot then suggestion would be to use .gitattributes only to exlude same pattern like you ignore in  `.npmignore` for consistency and have a release post hook to upload third archive with only dist files, so:

````
# .gitignore
# Directories to ignore from release archives
examples export-ignore
tests export-ignore

# Files to ignore from release archives
````

besides some point you might want to add also something like this into your git attributes
````

# Auto detect text files and perform LF normalization
* text=auto
# JS files must always use LF for tools to work
*.js eol=lf
````

**Comment by mkungla** [comment]
Eventually .gitattributes should be present anyway, perhaps not only to discuss under this pull request what should release archive look like, but what should initial .gitatributes look like. In this case just gie this pull request new title. Some suggestions would be then.
````
# .gitattributes
# Directories and files to ignore from release archives
examples export-ignore
tests export-ignore
.gitignor export-ignore
.gitattributes export-ignore

# Auto detect text files and perform LF normalization
# JS files must always use LF for tools to work
* text=auto
* eol=lf
dist/aframe-v0.2.0.min.js -diff binary minified
dist/aframe-v0.2.0.js.map -diff binary minified
dist/aframe.min.js -diff binary minified
*.js text
*.html text
*.npmignore text
*.md text
.ackrc text
.gitattributes text
.gitignore text
.jshintrc text
.nojekyll text
.travis.yml text
LICENSE text

*.jpg  -diff binary image
*.jpeg -diff binary image
*.png -diff binary image

# not sure about these
*.svg -diff binary
*.mtl -diff binary
*.obj -diff binary
````

**Comment by dmarcos** [comment]
[React for example](https://github.com/facebook/react/releases) offers two kinds of downloads in the release (they're even represented with different icons)
1. A zip file with the dist files
2. A zip with the archive of the repo

The purpose of git archiving is to actually snapshot the directory tree of a repo:
https://git-scm.com/docs/git-archive

@mkungla I don't know how the `.zip` with the dist files is generated but React is not using `.gitattributes` (https://github.com/facebook/react/blob/master/.gitattributes). Can we change this PR to let the archiving be the whole repo snapshot and generate another zip with the dist files like React does?

**Comment by mkungla** [comment]
@dmarcos the extra zip is uploaded via GitHub API. My previous comment includes `CURL` example how to upload extra archives to releases, I also made example tag under my fork where you see third archive see `aframe-0.2.0-dist.zip`: [Example Tag 0.2.0](https://github.com/mkungla/aframe/releases/tag/0.2.0) which is uploaded via API not with .gitattributes. I also mentioned above that perhaps indeed just use this method to upload extra archive and use this PR just for initial .gitattributes for other purpose. Example contents above.

**Comment by ngokevin** [comment]
FWIW, here are my current thoughts:

It seems more flexible and less contentious to have it serve both purpose of release and snapshot. Downloading the `.zip` will still give you access to the `dist/` files, it shouldn't be too obstructing to also have other files in the archive.

I agree that most people that would want the other files would just clone it. And it would make it cleaner to exclude those files, but it's a marginal benefit.

You mentioned that we might want a `.gitattributes` for other purposes. The one you posted, is all that does is just add EOF newlines? Just not sure we want to maintain another file unless it's really useful.

**Comment by mkungla** [comment]
`*-diff binary*` would keep your diffs clean

**Comment by mkungla** [comment]
@ngokevin 
> The one you posted, is all that does is just add EOF newlines?

* `* text=auto * eol=lf` will not only add correct line feeds if missing, but also corrects EOF if any of the files end with incorrect line feed so if contributors used editor inserts `CRLF` instead of `LF` then upon `git commit` git will show self explaining message: 
````
warning: CRLF will be replaced by LF in aframe.js.
The file will have its original line endings in your working directory.
````
This is real life example since latest state of `master` branch `/dist/aframe.js` seems to have  `CRLF` line feed instead of `LF`. As we all know we are past of the age of *Carriage Return* and should exclusively use `LF` instead.

* `binary`
Would tell which files git should not attend to create text diffs. This will make your repo cleaner and eventually commit objects smaller. for instance `*.png diff=exif` option would let you actually compare diffs of images if really would need that.

* `-diff` just don't generate diff object for file or pattern

As conclusion of this PR for me it seems that #1578 should be detached from this PR and discuss whether to upload 3rd archive via GitHub API as release asset like React and others too or not. 
And use this PR just use for `.gitattributes`. Are you going to use it and if yes then initial contents can be commited to this PR. If you are certain that you will not use `.gitattributes` for any purpose then this PR can be closed.

**Comment by mkungla** [comment]
http://stackoverflow.com/a/25125870/1318830

File: .gitattributes [hunk 1]
```diff
@@ -0,0 +1,38 @@
+# A-Frame .gitattributes
+# Directories and files to be ignored from release archives
+# keep it consistent with .npmignore
+
+examples export-ignore
```
ngokevin: I think we should keep these for now, just in case anyone wants them for whatever reason. And then we can merge the rest.

mkungla:
@ngokevin ready to go.

.gitattributes
```diff
@@ -0,0 +1,37 @@
+# A-Frame .gitattributes
+# Directories and files to be ignored from release archives
+# keep it consistent with .npmignore
+
+tests export-ignore
```
ngokevin: might as well just keep everything so we don't have to explain anything

mkungla:
Aight everything it is!

PR #1588, opened 2016-06-29
tweaks/nits to canvas texture example

ngokevin:
**Description:**

Some tweaks I caught to @MatthewShotton's example.

**Changes proposed:**
- move component to separate file for easier copy/pasting + encapsulation + enable linting
- use hyphens for ID
- use selector type for component to not have to do querySelector

### DIFF
--- a/examples/test/canvas-texture/index.html
+++ b/examples/test/canvas-texture/index.html
@@ -5,49 +5,18 @@
     <title>Canvas Texture</title>
     <meta name="description" content="Canvas Texture - A-Frame">
     <script src="../../../dist/aframe.js"></script>
+    <script src="./components/draw-canvas-rectangles.js"></script>
   </head>
   <body>
-    <script type="text/javascript">
-      AFRAME.registerComponent('update-canvas', {
-        schema: {default: ''},
-
-        init: function () {
-          this.canvas = document.getElementById(this.data);
-          this.ctx = this.canvas.getContext('2d');
-          this.ctx.fillStyle = "rgb(0,0,0)";
-          this.ctx.fillRect(0, 0, this.canvas.width, this.canvas.height);
-        },
-
-        tick: function (t) {
-          let hue = t / 10;
-          this.ctx.fillStyle = 'hsl(' + hue + ', 50%, 80%)';
-          this.ctx.fillRect(0,0,this.canvas.width,this.canvas.height);
-
-          hue = t / 15;
-          this.ctx.fillStyle = 'hsl(' + hue + ', 50%, 60%)';
-          var x = this.canvas.width / 10;
-          var y = this.canvas.height / 10;
-          this.ctx.fillRect(x, y, this.canvas.width - x * 2, this.canvas.height - y * 2)
-
-          hue = t / 20;
-          this.ctx.fillStyle = 'hsl(' + hue + ', 50%, 40%)';
-          x = this.canvas.width / 5;
-          y = this.canvas.height / 5;
-          this.ctx.fillRect(x, y, this.canvas.width - x * 2, this.canvas.height - y * 2);
-        }
-      });
-    </script>
     <a-scene stats>
       <a-assets>
-        <canvas id="helloworldCanvas" crossOrigin="anonymous"></canvas>
+        <canvas id="hello-world-canvas" crossOrigin="anonymous"></canvas>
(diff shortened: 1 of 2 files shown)


examples/test/canvas-texture/index.html
```diff
@@ -5,49 +5,18 @@
...
-          x = this.canvas.width / 5;
-          y = this.canvas.height / 5;
-          this.ctx.fillRect(x, y, this.canvas.width - x * 2, this.canvas.height - y * 2);
-        }
-      });
-    </script>
     <a-scene stats>
       <a-assets>
-        <canvas id="helloworldCanvas" crossOrigin="anonymous"></canvas>
+        <canvas id="hello-world-canvas" crossOrigin="anonymous"></canvas>
```
dmarcos: Why should we use dashes on the id? I don't like to have different conventions in different contexts if there's not a strong reason to.

ngokevin: It's just the more common convention in HTML to use dashes. We do it everywhere else, and it's in other style guides.

dmarcos: use dashes in HTML where? attribute names? attribute values? class names? I would keep ids camel cased. We should write a style guide soon so these things are easy on code reviews.

dmarcos: Having different conventions in different contexts adds a lot of cognitive load and space for inconsistencies. Unless is a widespread convention we should follow whatever we do in JS by default.

ngokevin: IDs, class names, attribute names.

https://google.github.io/styleguide/htmlcssguide.xml

I'd either say not care, or use hyphens since that's how we do it everywhere else. But not camel-case, since that's uncommon.

dmarcos: That's google's style guide. Those dashes really hurt my eyes. Let's do camelcase

ngokevin: You asked for a convention, I referred one of the largest and biggest companies, and also noted that we do it everywhere else, and it is the most common convention. Retreating to "it hurts my eyes" is silly.

Though if you want another argument, relying on case-sensitivity adds more opportunity for mistakes.

dmarcos: I stated my arguments above: cognitive load and risk for inconsistencies. One thing already bothers me is using single quotes for strings in JS while in HTML forces as to use double quotes. I would replace all singles with doubles so we have the same convention everywhere.

dmarcos: I don't want to spend more time discussing this. Let's do camelcase and move on please

ngokevin: If you don't want to spend time discussing it, just leave it as is and let's move on please. It's really disrespectful to just ignore comments and say 'I don't care, let's do it my way'.

dmarcos: I've already wasted too much of my life with inconsistency errors based on context switches. Doing server side Ruby / Python and then moving to JS / HTML. The more homogeneous the conventions the better.

ngokevin: It's better to stick with established conventions than to go against the grain. The context switches are natural given they're different syntax and languages.

dmarcos: You claim that dashes in HTML id's is a established convention. But I argue that it's not.

ngokevin: OK, your points are valid, and will switch it back so we can move on.

Given how we're arguing from both sides, I don't think we'll arrive to a conclusion in terms of style guide, so it'll most likely be "both ways are accepted, do it how you want", but fortunately, this PR is not about a style guide.

ngokevin: Though I'll probably think about it more. camelcase doesn't look like the worse thing actually.

dmarcos: Open the react website (https://facebook.github.io/react/) for example and do:
`document.body.querySelectorAll('[id]')`
You'll see that dashes is far from a convention. You will see dashes, underscores and camelcase wins.

dmarcos: In the absence of a widespread convention I would default to `do whatever JS does`

ngokevin: ok agreed

PR #1589, opened 2016-06-29
allow system configuration (fixes #1464)

### LINKED ISSUE #1464: Consider allowing configurable systems with schema
**Description:**

Allow configuration of systems through the DOM like we have for components. 

For example to adjust physics gravity `<a-scene physics="gravity: 9.8">`

Or Websocket server `<a-scene multiuser="http://myserver.5000">`

ngokevin:
**Description:**

Configurable systems. Use cases so far have been configurable Firebase system for API credentials, and configurable Physics system such as for gravity. 

```html
<a-scene firebase="apiKey: dsakldasd; authDomain: github.com">
```

So far we've had to work around using proxy components.

Currently does not support updates.

**Changes proposed:**
- Add schema and initial parsing to Systems.
- Have `Scene.getAttribute` and `Scene.setAttribute` to return system data (if exists) rather than component data that might have the same name (e.g., if you had a `physics` system and `physics` component, `sceneEl.getAttribute('physics')` should return physics system data.
- No actual changes to Entity, just shuffled code (to do early returns) around from leftover implementation.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -606,11 +606,9 @@ var proto = Object.create(ANode.prototype, {
    */
   getAttribute: {
     value: function (attr) {
+      // If cached value exists, return partial component data.
       var component = this.components[attr];
-      // If there's a cached value we just return it
-      if (component && component.attrValue !== undefined) {
-        return component.attrValue;
-      }
+      if (component && component.attrValue !== undefined) { return component.attrValue; }
       return HTMLElement.prototype.getAttribute.call(this, attr);
     },
     writable: window.debug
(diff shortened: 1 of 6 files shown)


src/core/a-entity.js
```diff
@@ -538,20 +539,23 @@ var proto = Object.create(ANode.prototype, {
...
-          this.updateComponentProperty(attr, value, componentPropValue);
-        } else {
-          this.updateComponent(attr, value);
-        }
-        // On debug mode we write the component state to the DOM attributes
-        if (isDebugMode) { this.components[attr].flushToDOM(); }
+
+      // If attribute is not a component or if attribute is a system for the scene.
+      // If attribute is a system that exists for the scene, then we ignore the component.
+      if (!components[attr] || (this.isScene && systems[attr])) {
```
dmarcos: This code would probably be part of  `a-scene` implementation of `setAttribute`

ngokevin: OK, updated + ready.

dmarcos:
@fernandojsg Do you see any problem with the editor handling this?

dmarcos:
After this change I'm starting to think that A-Frame should prescribe a physics engine for 0.4.0. You want one that works well and integrates consistently with components.

PR #1590, opened 2016-06-29
tweaks/nits to canvas texture example

ngokevin:
- Move component to separate JS file to get linting
- Use selector prop type


dmarcos:
Don't we have to add a link to the example in the index.html?

ngokevin:
It's already there, though I need to alphabetize it.

ngokevin:
thanks!

PR #1597, opened 2016-06-30
revert cursor event renames (fixes #1593)

### LINKED ISSUE #1593: Consider standard interaction event names
**Description:**

Earlier, I changed `click` to `cursor-click` for the cursor component to know where the event comes from.

But if we want to support multiple interaction mechanisms (mouse, gaze, controller), should we come up with standard event names such that developers only have to listen to one event (`click` vs. `cursor-click` + `trigger`)? Right now, we have "mousedown" even though you might not be using a mouse, should we keep those old names?

Example names:

- `cursor-click`, `trigger` -> `click`
- `cursor-mouseenter` -> `hoverenter`
- `cursor-mouseleave` -> `hoverleave`
- `cursor-mousedown` -> `clickdown`
- `cursor-mouseup` -> `clickup`

ngokevin:
**Description:**

This attempts to make the cursor event names (e.g., `click`, `mouseenter`) "standard". Meaning we won't expect only the cursor component to emit these events.

If someone creates a box, hopefully they'll only have to listen to `click` rather than having to listen to a new event name for each type of interaction method or controller.

The question is how cursor event names, or mouse event names, should interoperate with controller event names. How should we try to normalize them?

Or maybe if we don't want to normalize them at all, and consider them completely separate, then maybe we should be keeping the event names as is.

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -13,7 +13,7 @@ The cursor component lets us interact with entities through clicking and gazing.
 - Emits special mouse and hover events (e.g., relating to mouse down/up/enter/leave).
 - Has additional states for hovering.
 
-When the mouse is clicked, the closest visible entity intersecting the cursor, if any, will emit a `cursor-click` event. Note the cursor component only applies the raycasting behavior. To provide a shape or appearance to the cursor, you could apply the [geometry][geometry] and [material][material] components.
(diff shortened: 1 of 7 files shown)


dmarcos:
We have to change the tests too. There are a bunch failing

PR #1599, opened 2016-07-01
Shortcut to inject the editor

dmarcos:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aframe",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "description": "Building blocks for the VR Web",
   "homepage": "https://aframe.io/",
   "main": "dist/aframe.js",
--- a/src/components/scene/inspector.js
+++ b/src/components/scene/inspector.js
@@ -1,0 +1,41 @@
+/* global AFRAME */
+var registerComponent = require('../../core/component').registerComponent;
+var pkg = require('../../../package');
+
+var INSPECTOR_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+
+module.exports.Component = registerComponent('inspector', {
+  schema: {
+    url: {default: INSPECTOR_URL}
+  },
+
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + i
+    var shortcutPressed = evt.keyCode === 73 && evt.ctrlKey && evt.altKey;
+    if (!this.data || !shortcutPressed) { return; }
+    this.injectInspector();
+  },
+
+  injectInspector: function () {
+    var self = this;
+    var script;
+    if (AFRAME.inspector || AFRAME.inspectorInjected) { return; }
+    script = document.createElement('script');
+    script.src = this.data.url;
+    script.onload = function () {
+      AFRAME.inspector.open();
+      window.removeEventListener('keydown', self.onKeydown);
+    };
+    document.head.appendChild(script);
+    AFRAME.inspectorInjected = true;
+  },
+
+  remove: function () {
+    window.removeEventListener('keydown', this.onKeydown);
+  }
+});
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -22,6 +22,7 @@ require('./wasd-controls');
 require('./scene/canvas');
 require('./scene/debug');
 require('./scene/embedded');
+require('./scene/inspector');
 require('./scene/fog');
 require('./scene/keyboard-shortcuts');
 require('./scene/stats');
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -38,6 +38,7 @@ module.exports = registerElement('a-scene', {
     defaultComponents: {
       value: {
         'canvas': '',
+        'inspector': '',
         'keyboard-shortcuts': '',
         'vr-mode-ui': ''
       }


src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    if (!this.data || this.editorInjected || !shortcutPressed) { return; }
+    this.injectEditor();
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
+    script.type = 'text/javascript';
+    script.async = true;
+    script.onload = function () { console.log('A-Frame Editor Loaded'); };
+    script.src = 'https://fernandojsg.github.io/aframe-editor/build/aframe-editor.js';
```
cvan: can we have a const defined at the top of this file?

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    window.addEventListener('keydown', this.onKeydown, false);
+  },
+
+  pause: function () {
+    window.removeEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
```
cvan: we have [this helper function](https://github.com/aframevr/aframe/blob/f0c2b9f/src/utils/index.js#L126-L136). either use that, or check `document.activeElement` or make sure that `evt.target` is not a form field, `[contexteditable]` element, etc.

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    this.injectEditor();
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
+    script.type = 'text/javascript';
+    script.async = true;
+    script.onload = function () { console.log('A-Frame Editor Loaded'); };
+    script.src = 'https://fernandojsg.github.io/aframe-editor/build/aframe-editor.js';
+    document.getElementsByTagName('head')[0].appendChild(script);
```
cvan: you can just use `document.head`

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    this.editorInjected = true;
```
cvan: technically not true [yet](https://github.com/aframevr/aframe/pull/1599/files#diff-565a020f4c51daed65e6dca6c67d7e8eR29), but I [understand](https://github.com/aframevr/aframe/pull/1599/files#diff-565a020f4c51daed65e6dca6c67d7e8eR21) why you've done this.

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+
+  pause: function () {
+    window.removeEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
+    if (!this.data || this.editorInjected || !shortcutPressed) { return; }
```
cvan: what if `this.data.enabled === false`?

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    this.injectEditor();
```
cvan: if I have multiple scenes on the page, the Editor script will still get injected again

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
+    script.type = 'text/javascript';
+    script.async = true;
```
cvan: [> Scripts that are dynamically created and added to the document are async by default](http://www.html5rocks.com/en/tutorials/speed/script-loading/)

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
+    if (!this.data || this.editorInjected || !shortcutPressed) { return; }
+    this.injectEditor();
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
+    script.type = 'text/javascript';
```
cvan: can remove

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+    script.async = true;
+    script.onload = function () { console.log('A-Frame Editor Loaded'); };
```
cvan: shouldn't the Editor do its logging? (and perhaps spit out `A-Frame Editor Version: 0.1.0`?)

src/components/scene/editor.js
```diff
@@ -0,0 +1,38 @@
...
+
+module.exports.Component = registerComponent('editor', {
+  schema: { default: true },
+
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+  },
+
+  play: function () {
+    window.addEventListener('keydown', this.onKeydown, false);
```
cvan: `false` is the default

src/components/scene/editor.js
cvan: or in `src/index.js`?

src/components/scene/editor.js
```diff
@@ -0,0 +1,40 @@
...
+        AFRAME.editorInjected || !shortcutPressed) { return; }
+    this.injectEditor();
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
+    script.type = 'text/javascript';
+    script.src = EDITOR_URL;
+    document.getElementsByTagName('head')[0].appendChild(script);
+    AFRAME.editorInjected = true;
```
ngokevin: Does this need to be global rather than `this.injected`?

src/components/scene/editor.js
```diff
@@ -0,0 +1,40 @@
...
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
+    if (document.activeElement !== document.body || !this.data ||
+        AFRAME.editorInjected || !shortcutPressed) { return; }
+    this.injectEditor();
+  },
+
+  injectEditor: function () {
+    var script = this.script = document.createElement('script');
```
ngokevin: Doesn't look like `this.script` is used.

src/components/scene/editor.js
```diff
@@ -0,0 +1,40 @@
+/* global AFRAME */
+var registerComponent = require('../../core/component').registerComponent;
+
+var EDITOR_URL = 'https://aframe.io/aframe-editor/build/aframe-editor.js';
```
ngokevin: We should version this later `https://aframe.io/aframe-editor/builds/0.3.0/aframe-editor.js`

ngokevin:
r+wc

dmarcos:
This is ready to go but I first need @fernandojsg input first. In order to lock the version of the editor to a version of A-Frame we have to agree on the url schema. @ngokevin suggested `https://aframe.io/aframe-editor/builds/0.3.0/aframe-editor.js` where the version refers to the A-Frame version. I made the editor url a property of the component so people filling adventurous can try experimental versions

fernandojsg:
Yep, I agree that probably will be more useful to follow the same version number for the builds to match aframe. We could use `https://aframe.io/aframe-editor/builds/0.3.0/aframe-editor.js` for the first release

ngokevin:
Alternatively `https://aframe.io/releases/0.3.0/aframe-editor.min.js` to match A-Frame.

fernandojsg:
@ngokevin :+1:

dmarcos:
@ngokevin @fernandojsg How are we going to deal with master? If you use a-frame master it should point to the master version of the editor as well.

ngokevin:
Maybe we can immediately point A-Frame Master to Editor Master after we release. But A-Frame 0.3.0 should be locked to Editor 0.3.0.

Or we always point to the last release. If developers are using Master, we prescribe to configure their Editor URL. `<a-scene editor="url: mastereeditorurl">`

dmarcos:
@ngokevin are you want to add the release command for the editor? As you mentioned master should point to the latest release of the editor and devs can just change the url if they want bleeding edge. Before merging this It would great to have a functional URL `ttps://aframe.io/aframe-editor/builds/0.3.0/aframe-editor.js`

ngokevin:
How about I copy one to `https://aframe.io/releases/0.3.0/aframe-editor.min.js`?

dmarcos:
Yes we can copy it manually for now.

fernandojsg:
Remember to `s/editor/inspector`

dmarcos:
Let's push a build to `https://aframe.io/releases/0.3.0/aframe-inspector.min.js` so I can point this PR to it.

src/components/scene/editor.js
```diff
@@ -0,0 +1,42 @@
...
+  play: function () {
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  pause: function () {
+    window.removeEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
```
ngokevin: We should change this to `i` for Inspector.

ngokevin:
`ctrl + alt + i/e` loads the editor and immediately jumps into it the first time.

When I exit, and press `ctrl + alt + i/e` again, I expected it to open the Editor again, just like the Browser DOM Inspector you can toggle on/off with the same shortcut.

src/components/scene/editor.js
ngokevin: Also nice because it matches Chrome's `cmd + alt + i` shortcut.

dmarcos:
In Windows the shortcut is `ctrl+shift+i` . `windows key +alt+e` opens the browser menu in chrome.

dmarcos:
@fernandojsg how can I programmatically exit the editor so I can close/open it with the shortcut? Maybe the shortcut should be handled by the editor instead of aframe. What do you think?

fernandojsg:
@dmarcos once the editor is injected you can use `editor.enable()` and `editor.disable()` (I just realised that I should change also s/editor/inspector there :)
I could include the shortcut in the inspector to toggle it, but to inject the first time should be done outside the editor, how to handle this two entry points?

ngokevin:
Just a nitpick, could be `editor.open()/editor.close()` or `editor.enter()/enter.exit()`.

fernandojsg:
@ngokevin yep, just after writing that, I was thinking how bad naming these functions have. I'll change them for `open/close`

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+  play: function () {
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  pause: function () {
+    window.removeEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
```
ngokevin: `alt + ctrl + i` ?

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+  toggleInspector: function () {
+    if (AFRAME.inspector.opened) {
+      AFRAME.inspector.close();
+    } else {
+      AFRAME.inspector.open();
+    }
+  },
+
+  injectInspector: function () {
+    var script = document.createElement('script');
```
ngokevin: Declare before logic. So we don't unnecessarily create a script element if the inspector isn't injected.

ngokevin:
r+wc

src/components/scene/inspector.js
fernandojsg: Yep probably that will match better for `inspector` as similar as devtools, and leave `e` for the incoming VR-editor

fernandojsg:
One question regarding this shortcut. If you press `ctrl+alt+i` to inject the `inspector` and open it, should I add also the same shortcut once injected to toggle the inspector right? (Even if after that we could create a easier shortcut like `Esc` or whatever for doing the same)

ngokevin:
Yeah, would be nice to toggle like the Chrome Devtools.

FrederickDesimpel:
just wondering... since these are the same shortcuts... will this replace the browsers inspector? can' t we have both?

ngokevin:
I think we're using alt instead of shift.

cvan:
> I think we're using alt instead of shift.

nice, ok - can we update the docs?

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+  },
+
+  play: function () {
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  pause: function () {
+    window.removeEventListener('keydown', this.onKeydown);
```
fernandojsg: When the editor starts, it pause the scene, causing the shortcut to be removed (https://github.com/aframevr/aframe-inspector/issues/257).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aframe",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
ngokevin: I'd probably move this out to its own commit, but same PR

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
```
cvan: there might be some form field on the page. so this should be checking if we should be listening for the keypress. you can either use [our `utils.shouldCaptureKeyEvent` method](https://github.com/aframevr/aframe/blob/master/src/utils/index.js#L127-L137).

what was that case when you said `document.activeElement` was not `<body>` in some browser in some case? I can't reproduce that issue you were talking about.

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  },
+
+  injectInspector: function () {
+    var script = document.createElement('script');
+    if (AFRAME.inspectorInjected) { return; }
+    script.type = 'text/javascript';
+    script.src = this.data.url;
+    document.getElementsByTagName('head')[0].appendChild(script);
+    AFRAME.inspectorInjected = true;
```
cvan: why is this not `AFRAME.inspector.injected`?

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+      AFRAME.inspector.open();
+    }
+  },
+
+  injectInspector: function () {
+    var script = document.createElement('script');
+    if (AFRAME.inspectorInjected) { return; }
+    script.type = 'text/javascript';
+    script.src = this.data.url;
+    document.getElementsByTagName('head')[0].appendChild(script);
```
cvan: `document.head` works fine

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+      AFRAME.inspector.close();
+    } else {
+      AFRAME.inspector.open();
+    }
+  },
+
+  injectInspector: function () {
+    var script = document.createElement('script');
+    if (AFRAME.inspectorInjected) { return; }
+    script.type = 'text/javascript';
```
cvan: don't need

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
+    var shortcutPressed = evt.keyCode === 69 && evt.ctrlKey && evt.altKey;
+    if (!this.data || !shortcutPressed) { return; }
+    this.injectInspector();
+    this.toggleInspector();
+  },
+
+  toggleInspector: function () {
+    if (AFRAME.inspector.opened) {
```
cvan: should a method perhaps live in the Inspector repo called `toggle`?

src/components/scene/inspector.js
dmarcos: In the case of embedded scenes the activeElement might not be the body. The editor widgets can capture the focus and you still want the shortcut to work.

src/components/scene/inspector.js
cvan: it's possible that this code gets run before the `<script>` is actually injected by `injectInspector`.

even checking for `AFRAME.inspectorInjected` isn't enough. you'll want to add a callback or Promise for the `script.onload` for `injectInspector`.

src/components/scene/inspector.js
dmarcos: inspector is the scope where the inspector inject itself. inspectorInjected is the flag that aframe uses that the inspector has been already injected and we don't load it multiple times.

src/components/scene/inspector.js
cvan: good call

src/components/scene/inspector.js
```diff
@@ -0,0 +1,52 @@
...
+    var script = document.createElement('script');
+    if (AFRAME.inspectorInjected) { return; }
+    script.type = 'text/javascript';
+    script.src = this.data.url;
+    document.getElementsByTagName('head')[0].appendChild(script);
+    AFRAME.inspectorInjected = true;
+  },
+
+  remove: function () {
+    window.removeEventListener('keyup', this.listener);
```
cvan: `this.listener` isn't actually defined above

dmarcos:
I suggest to not do reviews if the label is not set to `review` this PR only went through a rebase cycle

src/components/scene/inspector.js
dmarcos: The shouldCaptureKeyEvent also returns if the alt or ctrl keys are pressed. We are capturing `ctrl + alt + e`

src/components/scene/inspector.js
```diff
@@ -0,0 +1,62 @@
...
+    url: {default: INSPECTOR_URL}
+  },
+
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + e
```
ngokevin: Should be `ctrl + alt + i`

ngokevin:
r+

can do later:

- unit tests
- bump package.json
- test with most recent inspector build that ctrl/alt/i can toggle inspector open and closed

mkungla:
restarted the build

src/components/scene/inspector.js
cvan: in the case of embedded scenes, the `<iframe>` can't listen to any keypresses until you first click it. on page load, the `document.activeElement` will be `<body>` of the parent window.

and, it's probably rare, but the keyboard shortcut might get accidentally triggered if a user typed it while typing in a form field (e.g., a `<textarea>` on a Wikipedia page or a comment field of a page with embedded scenes made with A-Frame).

also, what happens on the page when I have two embedded scenes? we probably don't want both listeners to fire?

src/components/scene/inspector.js
```diff
@@ -0,0 +1,61 @@
...
+    var script;
+    if (AFRAME.inspectorInjected) { return; }
+    script = document.createElement('script');
+    script.src = this.data.url;
+    document.head.appendChild(script);
+    AFRAME.inspectorInjected = true;
+  },
+
+  remove: function () {
+    window.removeEventListener('keydown', this.keydown);
```
cvan: typo: should be `this.onKeydown`

src/components/scene/inspector.js
```diff
@@ -0,0 +1,61 @@
...
+
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + i
+    var shortcutPressed = evt.keyCode === 73 && evt.ctrlKey && evt.altKey;
+    var escape = evt.keyCode === 27;
```
cvan: IMO, this should also check `evt.target`. we're assuming that the scene won't be listening for Escape at all? that's a pretty common keypress to listen for.

cvan:
just marked `needs changes` for the typo in the `remove` method

src/components/scene/inspector.js
ngokevin: Code is a little misleading because it only applies when the Inspector is open, which the scene shouldn't be listening to anything at that point.

dmarcos: Yes. The escape shortcut is to close the editor that will be occupying the whole screen with the scene paused. The editor should be the only thing in the page listening for key presses.

src/components/scene/inspector.js
```diff
@@ -0,0 +1,41 @@
...
+  init: function () {
+    this.onKeydown = this.onKeydown.bind(this);
+    window.addEventListener('keydown', this.onKeydown);
+  },
+
+  onKeydown: function (evt) {
+    // Alt + Ctrl + i
+    var shortcutPressed = evt.keyCode === 73 && evt.ctrlKey && evt.altKey;
+    if (!this.data || !shortcutPressed) { return; }
+    this.injectInspector();
```
dmarcos: I simplified this. The component just injects the editor the first time `ctrl + alt + i` and then the editor takes over listening to keyboard events.

PR #1601, opened 2016-07-02
Return early from exitVR if not in VR mode.

donmccurdy:
Prevents an exception when <kbd>Esc</kbd> is pressed and not in VR mode. I use pointerlock while debugging components, as well as the "Pause On Caught Exceptions" option in dev tools, so the otherwise harmless exception gets in the way a bit.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -150,6 +150,7 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
+        if (!this.is('vr-mode')) { return; }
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -150,6 +150,7 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
+        if (this.states.indexOf('vr-mode') === -1) { return; }
```
ngokevin: Try `this.is('vr-mode')`

donmccurdy: Oh cool.

ngokevin:
Thanks!

PR #1605, opened 2016-07-05
fix a-cursor not being registered

ngokevin:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Should be simple enough, just gonna merge to fix the master error

PR #1610, opened 2016-07-06
support runtime mixins (fixes #895)

### LINKED ISSUE #895: Mixins should still apply to entities even if they are attached after the entity has already loaded.
Have a scene:

```html
<a-assets></a-assets>
<a-scene>
  <a-entity mixin="async-mixin"></a-entity>
</a-scene>
```

The entity has no mixin applied. Inject a mixin.

```js
var mixin = document.createElement('a-mixin');
assets.appendChild(mixin);
```

The mixin should now be applied to the entity even if it was added after the entity.

ngokevin:
**Description:**

Support mixins being attached/injected after entities have already loaded.

**Changes proposed:**
- When a mixin is attached, find all entities using it that have already loaded, and apply the mixin.
- Change `Component.updateProperties` to take an undefined `attrValue`. This means continue updating properties using mixins and such, but use cached attr value.
- Rename `applyMixin` to `handleMixinUpdate` to signify that it is asynchronous and that it doesn't apply when the mixin is attached. It only handles observed mutations.
- Rename `componentAttrCache` to `componentCache` because after parsing, anything related to the attribute value becomes irrelevant. It becomes purely component data.
- Various docstrings/comments and max 100-char style stuff.

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -58,7 +58,7 @@ module.exports.AAnimation = registerElement('a-animation', {
             return;
           }
 
-          self.applyMixin();
+          self.handleMixinUpdate();
           self.update();
           self.load();
         }
@@ -69,7 +69,7 @@ module.exports.AAnimation = registerElement('a-animation', {
       value: function (attr, oldVal, newVal) {
         if (!this.hasLoaded || !this.isRunning) { return; }
         this.stop();
-        this.applyMixin();
+        this.handleMixinUpdate();
         this.update();
       }
     },
@@ -326,7 +326,7 @@ module.exports.AAnimation = registerElement('a-animation', {
      * Works the same as component mixins but reimplemented because animations
      * aren't components.
      */
-    applyMixin: {
+    handleMixinUpdate: {
       value: function () {
         var data = {};
         var elData;
--- a/src/core/a-mixin.js
+++ b/src/core/a-mixin.js
@@ -4,13 +4,15 @@ var registerElement = require('./a-register-element').registerElement;
 var components = require('./component').components;
 
 /**
- * @member {object} componentAttrCache - Cache of pre parsed component attributes
+ * @member {object} componentCache - Cache of pre-parsed values. An object where the keys
+ *         are component names and the values are already parsed by the component.
  */
 module.exports = registerElement('a-mixin', {
   prototype: Object.create(ANode.prototype, {
     createdCallback: {
       value: function () {
-        this.componentAttrCache = {};
+        this.componentCache = {};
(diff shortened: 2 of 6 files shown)


src/core/a-mixin.js
```diff
@@ -48,23 +51,38 @@ module.exports = registerElement('a-mixin', {
...
           this.cacheAttribute(attrName);
         }
       }
+    },
+
+    /**
+     * For entities that already have been loaded by the time the mixin was attached, tell
+     * those entities to register the mixin and refresh their component data.
+     */
+    refreshEntities: {
```
dmarcos: Can we use `updateEntities` to use a consistent language?

src/core/a-mixin.js
```diff
@@ -22,49 +24,80 @@ module.exports = registerElement('a-mixin', {
...
+    /**
+     * If `attr` is a component, then parse the value using the schema and store it.
+     */
     cacheAttribute: {
       value: function (attr, value) {
         var component = components[attr];
         if (!component) { return; }
-        value = value === undefined ? HTMLElement.prototype.getAttribute.call(this, attr) : value;
-        this.componentAttrCache[attr] = component.parseAttrValueForCache(value);
+        value = (
```
dmarcos: What is this parenthesis? Can we collapse to one line?

ngokevin: ok. any other comments for this PR?

src/core/a-animation.js
```diff
@@ -326,7 +326,7 @@ module.exports.AAnimation = registerElement('a-animation', {
      * Works the same as component mixins but reimplemented because animations
      * aren't components.
      */
-    applyMixin: {
+    handleMixinUpdate: {
```
dmarcos: apply mixin seems a better name than handleMixinUpdate. When invoked from the `attachedCallback` method there's no mixin update anywhere.

ngokevin: Yeah, I guess so only because of `<a-animation>`. Otherwise for entities, it's purely a handler for the MutationObserver. I can rename it back for now.

---

Just to implant the idea in your head, I want to kill `<a-animation>` for a [component-based approach](https://github.com/ngokevin/aframe-animation-component#why). For timeline and such, I think we could find an API that meshes with ECS. Just bringing this up now because if we didn't have `<a-animation>`, we could transfer the mixin logic down to the Entity level, and `ANode` would purely be for load order/events.

dmarcos: We can discuss about removing `a-animation`. I believe that should be solved by 3rd parties anyways.

PR #1611, opened 2016-07-07
Return promise when exitVR() returns early.

donmccurdy:
More consistent, pointed out by @cvan in #1601.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -149,7 +149,7 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
-        if (!this.is('vr-mode')) { return; }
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -149,7 +149,7 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
-        if (!this.is('vr-mode')) { return; }
+        if (!this.is('vr-mode')) { return Promise.resolve(); }
```
cvan: I'd probably `return Promise.resolve(false);` or `return Promise.resolve('Already in VR mode.');`. either way, this fixes it so thanks 👍

donmccurdy: I don't know what the precedent is or should be, here. Which do you prefer?

PR #1615, opened 2016-07-07
fix pixel ratio for VR mode (fixes #1541)

### LINKED ISSUE #1541: Different texture quality in release/v0.2.0 and master branch
I tried to play with the [Composite Example](https://aframe.io/examples/showcase/composite/) in the project and found that texture quality is different between the v0.2.0 release version and master branch when in VR Mode on Mobile, both iOS and Android.

In release/v0.2.0, the texture quality is fine and working properly in VR Mode on Mobile.
In master branch, the texture quality is downgraded and all images are blur in VR Mode on Mobile.

I made two html pages which have the same source code but includes the two different versions of Aframe framework.

- A-Frame Version: v0.2.0
- Platform/Device: VR Mode on Mobile
- Description: Fine texture quality
- [Composite Example with release/v0.2.0](http://bit.ly/1U8Zk8r)

---

- A-Frame Version: master branch
- Platform/Device: VR Mode on Mobile
- Description: Texture quality is downgraded, images are blur
- [Composite Example with branch/master](http://bit.ly/1UCAqg0)

ngokevin:
**Description:**

VREffect wasn't listening to A-Frame's set pixel ratio.

To test, open VR mode on mobile with the Composite example. Compare resolution quality before/after.

**Changes proposed:**
- Proposed upstream of VREffect changes to https://github.com/mrdoob/three.js/pull/9304

### DIFF
--- a/vendor/VREffect.js
+++ b/vendor/VREffect.js
@@ -73,7 +73,7 @@ THREE.VREffect = function ( renderer, onError ) {
 		if ( isPresenting ) {
 
 			var eyeParamsL = vrHMD.getEyeParameters( 'left' );
-			renderer.setPixelRatio( 1 );
+			renderer.setPixelRatio( rendererPixelRatio );
 
 			if ( isDeprecatedAPI ) {
 
@@ -124,7 +124,7 @@ THREE.VREffect = function ( renderer, onError ) {
 
 			}
 
-			renderer.setPixelRatio( 1 );
+			renderer.setPixelRatio( rendererPixelRatio );
 			renderer.setSize( eyeWidth * 2, eyeHeight, false );
 
 		} else {
@@ -146,7 +146,7 @@ THREE.VREffect = function ( renderer, onError ) {
 			rendererSize = renderer.getSize();
 
 			var eyeParamsL = vrHMD.getEyeParameters( 'left' );
-			renderer.setPixelRatio( 1 );
+			renderer.setPixelRatio( rendererPixelRatio );
 			renderer.setSize( eyeParamsL.renderWidth * 2, eyeParamsL.renderHeight, false );
 
 		} else {


dmarcos:
Have you tried in mobile?

ngokevin:
Yeah. This issue is all about mobile.

dmarcos:
How do you know the pixelRatio of a headset display? In mobile it might work because the display used for 2D and VR mode is the same.

dmarcos:
I think that's the reason why this is handled by the polyfill

ngokevin:
Where is this handled in the polyfill?

dmarcos:
I've never completely understood the logic. You can dig and try to understand what this value does:
https://github.com/borismus/webvr-polyfill/blob/master/src/main.js#L57

dmarcos:
I would first bump the polyfill to see if there's any difference and the problems are still there.

ngokevin:
I've played with the BUFFER_SCALE and poked all over the polyfill. It always comes down to VREffect for this issue.

dmarcos:
Are the polyfill examples affected by the low resolution issues?

ngokevin:
No, it only happens because through VREffect. Non-VR on mobile looks fine either way.

dmarcos:
The basic example of the polyfill uses three.js and hence `VREffect`. 
https://borismus.github.io/webvr-polyfill/examples/basic/
Is it affected?

ngokevin:
Hmm...OK, will look further.

dmarcos:
VREffect sets pixelRatio to 1 because the buffer size is calculated manually based on the info reported by `getEyeParameters`

dmarcos:
On mobile and using the polyfill you will be calling this version of `getEyeParameters`:

https://github.com/borismus/webvr-polyfill/blob/master/src/cardboard-vr-display.js#L100

It uses `BUFFER_SCALE`. I would take that method as the first suspect.

PR #1620, opened 2016-07-08
Fixed toggle stats component visibility with attr, (fixes #1619)

### LINKED ISSUE #1619: Turning `stats=""`/`stats="true"` to `stats="false"` does not remove the `stats` component
**Description:**

Turning `stats=""`/`stats="true"` to `stats="false"` does not remove the `stats` component. I have to explicitly remove the `stats` attribute.

- A-Frame Version: `master`
- Platform/Device: all/any

mkungla:
**Description:**
Enables you to toggle stats component visibility with DOM attribute
without need of explicitly remove the stats attribute which is expected
result next to already available methods stats.hide() and stats.show()

**Changes proposed:**
- Use component `data` to hold `true|false`
- Upon DOM update determine either to hide or to show stats

### DIFF
--- a/src/components/scene/stats.js
+++ b/src/components/scene/stats.js
@@ -11,10 +11,7 @@ var AFrameStats = window.aframeStats;
  * Stats appended to document.body by RStats.
  */
 module.exports.Component = registerComponent('stats', {
-  schema: {
-    type: 'boolean',
-    default: true
-  },
+  schema: { default: true },
 
   init: function () {
     var scene = this.el;
--- a/src/components/scene/stats.js
+++ b/src/components/scene/stats.js
@@ -11,6 +11,11 @@ var AFrameStats = window.aframeStats;
  * Stats appended to document.body by RStats.
  */
 module.exports.Component = registerComponent('stats', {
+  schema: {
+    type: 'boolean',
+    default: true
+  },
+
   init: function () {
     var scene = this.el;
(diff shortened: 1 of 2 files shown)


src/components/scene/stats.js
```diff
@@ -11,6 +11,11 @@ var AFrameStats = window.aframeStats;
  * Stats appended to document.body by RStats.
  */
 module.exports.Component = registerComponent('stats', {
+  schema: {
+    type: 'boolean',
```
dmarcos: No need to add type since it's inferred from the default.

mkungla: @dmarcos so that can be removed then also from `visible component` so just having
`schema: { default: true }` right?

dmarcos: Correct.

dmarcos: I believe there's a bug associated to this. Can you try the following with your patch:
```<a-scene stats=""></a-scene>``` Do the stats show up?

mkungla: Bug for stats is #1619, So removed  `type: 'boolean'` from stats component and visible tested

- `<a-scene stats></a-scene>` OK
- `<a-entity visible geometry="primitive: box; depth: 1; height: 1; width: 1;" ></a-entity>` OK
- and both hide / show when toggling `true|false`

PR #1621, opened 2016-07-08
fix spot lights

ngokevin:
**Description:**

three.js used to be exponent, now it's penumbra


FrederickDesimpel:
Spotlights also have a position bug, their cone is a unit or so higher then their origin: http://www.frederickdesimpel.be/test/webVR/Aframe/spot-bug.jpg

Also lookat component doesn't work as expected with spotlights.

PR #1629, opened 2016-07-12
Fixes THREE.Audio.load is deprecated, closes #1472

### LINKED ISSUE #1472: Use `THREE.AudioLoader` instead of deprecated `THREE.AUDIO#load`
**Description:**

1. In Chrome, load [http://localhost:9000/examples/showcase/anime-UI/](http://localhost:9000/examples/showcase/anime-UI/)
2. Notice JS warning:

`THREE.Audio: .load has been deprecated. Please use THREE.AudioLoader.`

- A-Frame Version: `273622e`
- Platform/Device: all

mkungla:
**srcChanged**
I had to update audio src which caused these warnings.
`THREE.Audio: .load has been deprecated. Please use THREE.AudioLoader.`

Although this is not big issue I came across another problem that I could not play previous audio again.
so changing `src` on same entity back to audio which was already played causes
````
The buffer passed to decodeAudioData contains an unknown content type.x
EncodingError: The given encoding is not supported.
````
This happens since Three. `THREE.XHRLoader.load` returns for already played audio cache which is instance of `ArrayBuffer`, but THREE.context.decodeAudioData expects instance of `AudioBuffer`

**Changes proposed:**
- use THREE.AudioLoader  instead of THREE.Audio.load 
- Do not let `Tree` to cache audio files

### DIFF
--- a/docs/components/sound.md
+++ b/docs/components/sound.md
@@ -20,7 +20,7 @@ The sound component defines the entity as a source of sound or audio. The sound 
 | Property | Description                                                           | Default Value |
 |----------|-----------------------------------------------------------------------|---------------|
 | autoplay | Whether to automatically play sound once set.                         | false         |
-| on       | An event for the entity to listen to before playing sound.            | click         |
+| on       | An event for the entity to listen to before playing sound.            | null          |
 | loop     | Whether to loop the sound once the sound finishes playing.            | false         |
 | src      | Selector to an asset `<audio>`or `url()`-enclosed path to sound file. | null          |
 | volume   | How loud to play the sound.                                           | 1             |
--- a/docs/components/sound.md
+++ b/docs/components/sound.md
@@ -12,18 +12,18 @@ The sound component defines the entity as a source of sound or audio. The sound 
 
 ```html
 <a-entity id="river" geometry="primitive: plane" material="color: blue"
-          position="-10 0 0" sound="src: river.mp3; autoplay: true"></a-entity>
+          position="-10 0 0" sound="src: url(river.mp3); autoplay: true"></a-entity>
 ```
 
 ## Properties
 
-| Property | Description                                                | Default Value |
-|----------|------------------------------------------------------------|---------------|
-| autoplay | Whether to automatically play sound once set.              | false         |
-| on       | An event for the entity to listen to before playing sound. | click         |
-| loop     | Whether to loop the sound once the sound finishes playing. | false         |
-| src      | Path to sound file.                                        | null          |
-| volume   | How loud to play the sound.                                | 1             |
+| Property | Description                                                           | Default Value |
+|----------|-----------------------------------------------------------------------|---------------|
+| autoplay | Whether to automatically play sound once set.                         | false         |
+| on       | An event for the entity to listen to before playing sound.            | click         |
+| loop     | Whether to loop the sound once the sound finishes playing.            | false         |
+| src      | Selector to an asset `<audio>`or `url()`-enclosed path to sound file. | null          |
+| volume   | How loud to play the sound.                                           | 1             |
 
 ## Events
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/components/sound.js [hunk 1]
```diff
@@ -46,7 +47,14 @@ module.exports.Component = registerComponent('sound', {
     }
 
     // All sound values set. Load in `src`.
-    if (srcChanged) { sound.load(data.src); }
+    if (srcChanged) {
+      var audioLoader = this.audioLoader || new THREE.AudioLoader();
```
dmarcos: Can we maybe initialize this in the `init` method?

src/components/sound.js
```diff
@@ -46,7 +47,14 @@ module.exports.Component = registerComponent('sound', {
...
+      audioLoader.load(data.src, function (buffer) {
+        sound.setBuffer(buffer);
+        // Remove this key from cache, otherwise we can't play it again
+        THREE.Cache.remove(data.src);
```
dmarcos: Why do we need to remove from cache the audio to play it again?

mkungla:
@dmarcos  
> Can we maybe initialize this in the init method?

Yes, I added commit which initializes `AudioLoader` in the `init` method now.

> Why do we need to remove from cache the audio to play it again?

Consider following scene using current aframe.js and that you have some tracks like `1.mp3` and `2.mp3`

````html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,shrink-to-fit=no,user-scalable=no,minimal-ui">
<script src="https://raw.githubusercontent.com/aframevr/aframe/master/dist/aframe.js"></script>
</head>
<body>
<a-scene>
	<a-entity sound="src: 1.mp3; autoplay: true" position="0 1 0"></a-entity>
</a-scene>
</body>
</html>
````

-  As you load your scene track 1 will start playing
- Now modify DOM  audio src to `src: 2.mp3` and you'll hear second track
- When you modify DOM back to `src: 1.mp3` you get silence and following errors

````
The buffer passed to decodeAudioData contains an unknown content type.{url}
EncodingError: The given encoding is not supported.
````

If you now use following aframe.js which has this PR applied
````html
<script src="https://raw.githubusercontent.com/mkungla/aframe/sound-AudioLoader-test/dist/aframe.js"></script>
````
So removing file from TREE.Cache removes these errors and you can shuffle the tracks without any problems. 

It seems to be problem with TREE.js Cache which passes ArrayBuffer to AudioLoader.decodeAudioData while last expects instance of AudioBuffer. I have opend issue at TREE as well about that and keep eye on where it goes. For now I find that using  `THREE.Cache.remove(data.src);` is securest way to ensure that aframe's audio src can modified, setting new sources and setting it also back to already played source.

File: src/components/sound.js [hunk 3]
```diff
@@ -9,7 +9,7 @@ var warn = debug('components:sound:warn');
  */
 module.exports.Component = registerComponent('sound', {
   schema: {
-    src: { default: '' },
+    src: { type: 'src' },
     on: { default: 'click' },
```
ngokevin: Can you make this `default: ''` while you're here?

ngokevin:
Yeah, can you update the examples and docs as well? Thanks!

mkungla:
@ngokevin I'll go over all filed issues and collect some additional information before I file new issue about sound component what we talked about. Since this PR improves a lot compared to masters `if (srcChanged) { sound.load(data.src); }` and it self does not introduce new issues then I find that these other questions should handled separately.

docs/components/sound.md
```diff
@@ -12,18 +12,18 @@ The sound component defines the entity as a source of sound or audio. The sound
...
-|----------|------------------------------------------------------------|---------------|
-| autoplay | Whether to automatically play sound once set.              | false         |
-| on       | An event for the entity to listen to before playing sound. | click         |
-| loop     | Whether to loop the sound once the sound finishes playing. | false         |
-| src      | Path to sound file.                                        | null          |
-| volume   | How loud to play the sound.                                | 1             |
+| Property | Description                                                           | Default Value |
+|----------|-----------------------------------------------------------------------|---------------|
+| autoplay | Whether to automatically play sound once set.                         | false         |
+| on       | An event for the entity to listen to before playing sound.            | click         |
```
ngokevin: Default value for this is no longer click

examples/test/cursor/index.html
```diff
@@ -9,6 +9,8 @@
   <body>
     <a-scene stats>
       <a-assets>
+      	<audio id="blip1" src="../../showcase/anime-UI/audio/321103__nsstudios__blip1.wav"></audio>
```
ngokevin: missed indent

PR #1631, opened 2016-07-14
Single property components have to return the default value when the …

dmarcos:


ngokevin:
Going to assume work-in-progress based on discussion yesterday.

dmarcos:
The code is simpler now and works as expected but It introduces a wrinkle in the consistency of the system that I don't know how to do differently. Look at the following entity with a single property component call `dummy` with `true` as default value.
````html
  <a-entity position="" rotation="" scale="" visible="" dummy=""></a-entity>
````
position, rotation, scale and visible are default components so they're not really defined on the entity (a mixin would take precedence) On the contrary dummy is not a default component and hence really defined on the entity. When calling `getAttribute` default and non default components will produce different values:
`entityEl.getAttribute('dummy') === true`
`entityEl.getAttribute('position') === undefined`
`entityEl.getComputedAttribute('position') === { x: 0, y: 0, z: 0 }`
Does it make sense?

**Comment by dmarcos** [comment]
After discussion with @ngokevin this is how the API behaves with the latest changes. Given a single property component `dummy`with a default value `true` and the following entity:
```html
<a-entity dummy></a-entity>
````
These are the values returned for the components
```javascript
el.getAttribute('dummy') === true;
el.getAttribute('position') === null;
el.getComputedAttribute('position') === { x: 0, y: 0, z: 0 };
````
The only inconsistency left it's when opening the DOM inspector the entity above will appear like `<a-entity position="" rotation="" scale="" scale="" dummy></a-entity>`

When calling `getAttribute` on those components you would expect to get the default values as it happens with `dummy`. Default components will return `null` instead because they're not really defined in the entity element despite of being shown as attributes. To retrieve the value of the default components one has to call `getComputedAttribute`

File: src/core/component.js [hunk 1]
```diff
@@ -135,15 +135,8 @@ Component.prototype = {
...
-    if (value === '') {
-      this.attrValue = undefined;
-      return;
-    }
-    if (typeof value === 'string') {
-      this.attrValue = this.parseAttrValueForCache(value);
-      return;
-    }
-    this.attrValue = value !== undefined ? extendProperties({}, value, isSinglePropSchema) : this.attrValue;
+    var attrValue = this.attrValue = this.parseAttrValueForCache(value);
```
ngokevin: Don't need to assign to `this.attrValue` here since it gets replaced.

ngokevin:
r+

PR #1632, opened 2016-07-14
Add Getting Started for Beginners

AnnCaryn:
@ngokevin 

**Description:**

**Changes proposed:**
-
-
-

A quick step by step for brand new users/beginners to get them started with A-frame.

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -11,22 +11,57 @@ nav_slug: install
 
 <script async src="//assets.codepen.io/assets/embed/ei.js"></script>
 
-There are several ways to get started with A-Frame:
-
-* [Play with CodePen](#CodePen)
-* [Grab the Boilerplate](#Boilerplate)
-* [Include the JS Build](#Builds)
-* [Install from npm](#npm)
-
-## CodePen
+## Beginners:
+You can quickly check out and play with A-frame [in CodePen](#CodePen):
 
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks for the work. I do feel workflow instructions are important. Though for the purpose of endorsing one, I feel these instructions might be:

- Prescriptive (must use Git, GitHub, Node, Atom)
- Challenging (introduces a ton of new technologies into the stack)
- Out-of-scope (tutorial for getting started with web development tools in a section about getting started with A-Frame)

However, I feel like this might be a great blog post material if there're more screenshots and elaboration. And then we can also put your face on it. Something along the lines of "Workflow for Web Development Beginners".

ngokevin:
Going to think about this a bit more, actually. Still on the fence. It might not be prescriptive since we should recommend Git, GitHub, Node, and I re-read and saw you didn't actually enforce a text editor. If the section is small and contained enough, it might be useful.

AnnCaryn:
Yup.  Totally understand the endorsement aspect.  Just wanted to get something down for the total beginner when they are looking at the docs and then work from there.  Most important is to get this in the right place for new people in the funnel when they learn about A-frame, wherever that may be.

ngokevin:
Thought about it more. I don't think we should recommend this workflow for total beginners. GitHub + Node + CLI is too much. This would make a good blog post for easing people into web development. For the next version of the A-Frame though, we will be featuring different workflows for different types of creators.

AnnCaryn:
Do you think there is a path for beginners to start now with A-frame or do they need to wait until the next version?

ngokevin:
The homepage we will have something that acts like a playground for beginners. There's also Codepen, and workflows like MagicaVoxel.

ngokevin:
Will close for now, but will loop you in homepage posts.

I recommend publishing this workflow online somewhere for now though for those that need it! (like Medium)

PR #1634, opened 2016-07-14
deprecate declarative events in favor of event set component

ngokevin:
**Description:**

https://github.com/ngokevin/aframe-event-set-component

Replaced with a component-form API. Advantage is that listeners can be added via a simple `setAttribute`, can work with mixins (or aframe-mss), and is faster due to removing the overhead of creating/attach a DOM element.

**Changes proposed:**
- Deprecate Declarative Events.
- Remove examples.


dmarcos:
You removed the examples but the implementation of `a-events` is still there right?

ngokevin:
Yeah, leaving it for one version to sunset.

PR #1637, opened 2016-07-18
Add HTC Vive to index page on docs

tjhorner:
**Description:**
WebVR uses the OpenVR SDK and thus supports the HTC Vive. I've tested it myself :)

**Changes proposed:**
- Add HTC Vive to list of compatible platforms on the index page.


dmarcos:
Thanks!

PR #1638, opened 2016-07-18
Remove declarative example links from index

darkwing:
**Description:**
The "Declarative" links on the A-Frame Examples page are 404'ing  :(

**Changes proposed:**
- Remove said links until the examples are in place.


ngokevin:
Thanks! We removed those examples, forgot those links.

PR #1639, opened 2016-07-19
Example cleanup; update 'begin' attribute to 'delay'

darkwing:
**Description:**
Many examples use the `begin` attribute instead of new `delay` attribute so I've fixed them to avoid warning.  I've also updated a broken image path.


dmarcos:
Nice! Thanks!

PR #1640, opened 2016-07-19
Fix #704: Add warning for use of repeat=infinite

### LINKED ISSUE #704: warning if repeat="infinite" instead of repeat="indefinite"
<a href="https://github.com/cvan"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Thursday Dec 03, 2015 at 04:38 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/584_

----

# 1

Though [SMIL](https://css-tricks.com/guide-svg-animations-smil/) (since it appears what this is based off of) calls it `repeatCount="indefinite"`, `repeat="infinite"` would make more sense to me.

<hr>

# 2

If you enter anything that gets `parseInt`'d as not a number (e.g., `repeat="infinitite"` instead of `repeat="indefinite"`), `NaN` gets used which makes the thing repeat forever.

<hr>

# 3

We `parseInt` so if you pass `repeat="1.5"`, we actually do it only `1` times. [`tween.js`](https://github.com/tweenjs/tween.js/blob/e19a4c9/src/Tween.js#L335) does `parseFloat`, for what it's worth. So maybe we should do it.

<hr>

# 4

The animation [`DEFAULTS.repeat` is `0`](https://github.com/MozVR/aframe-core/blob/dev/src/constants/animation.js#L92). 
But then the [`getTween` method](https://github.com/MozVR/aframe-core/blob/dev/src/core/a-animation.js#L119) checks if `repeat="indefinite"`.
And this [`Infinity` check here](https://github.com/MozVR/aframe-core/blob/dev/src/core/a-animation.js#L119).
Not entirely sure why it's written the way it is. Can you explain?

<hr>

# 5

Per [SMIL](https://css-tricks.com/guide-svg-animations-smil/) (since it appears what this is based off of):

> Also note that, even if you set the `repeatCount` to `indefinite`, it will be overridden by the `end` values and will not repeat indefinitely.

So, that would mean that if the `to` condition has been met, the `repeat="indefinite"` would be ignored. That's not what I'd expect, but SMIL is complicated. I'd say ignore that, but it seemed worth calling out.

darkwing:
**Description:**
Fixes #704 

**Changes proposed:**
- One reservation here is that the `value` method can get called many times and that this warning may become overwhelming.  Let me know what you think.

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -159,6 +159,10 @@ module.exports.AAnimation = registerElement('a-animation', {
     update: {
       value: function () {
         var data = this.data;
+        // Terminology warning if infinite used instead of indefinite
+        if (data.repeat === 'infinite') {
+          console.warn("Using 'infinite' as 'repeat' value is invalid.  Use 'indefinite' instead.");
+        }
         // Deprecation warning for begin when used as a delay.
         if (data.begin !== '' && !isNaN(data.begin)) {
           console.warn("Using 'begin' to specify a delay is deprecated. Use 'delay' instead.");


ngokevin:
Thanks!

Also check out https://github.com/ngokevin/aframe-animation-component, I will be repping this implementation from now on.

Issue #1129, opened 2016-03-10
selectorAll property type should convert nodeList to array


ngokevin:
Follow up to @legendtang's https://github.com/aframevr/aframe/pull/1099

cvan:
I know the naming is unrelated, but perhaps could call it just `query` à la `xtag.query`:

https://x-tag.github.io/

![image](https://cloud.githubusercontent.com/assets/203725/13662508/e4c1ce54-e64f-11e5-84c3-5bd005169f49.png)

legendtang:
@cvan But `x-tag` requires additional polyfills for mainstream browsers for now. Maybe too early to use this name?

cvan:
@cvan But x-tag requires additional polyfills for mainstream browsers for now. Maybe too early to use this name?

I was not suggesting to use X-Tag, just using its method names as inspiration/prior art. And we are already using a polyfill for Custom Elements: https://github.com/WebReflection/document-register-element - I've heard bugs with the polyfill, and we have an issue (#567) on file to investigate if the more "popular" [WebComponentsJS Custom Elements polyfill](https://github.com/webcomponents/webcomponentsjs/tree/master/src/CustomElements) fixes the bugs/inconsistencies we've seen between native and polyfilled.

ngokevin:
Hard to tell whether you are selecting one element or multiple with ```query```. And I think selector/selectorAll is easier to get what it means. (query)selector/selectorAll are in the browser and probably make a stronger for "prior art".

cvan:
yeah, I think that's a good point. though jQuery seems to handle this fine - `$()` vs `$()[0]`. I've always liked `$` vs `$$`. but, the naming you have is fine.

ngokevin:
Yeh. This is for property types though, not for actually running queries.

PR #1642, opened 2016-07-19
Update selectorAllParse to return an Array instead of a NodeList

darkwing:
**Description:**
Fixes https://github.com/aframevr/aframe/issues/1129

I suspect this one will require much more work.  I've run tests and all passes but things to consider:

*  Is this a breaking change moving forward where API docs must be updated?  Is it work the breaking change?
*  What other components may need updating to accommodate an array instead of a NodeList?

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -85,7 +85,7 @@ function selectorParse (value) {
 function selectorAllParse (value) {
   if (!value) { return null; }
   if (typeof value !== 'string') { return value; }
-  return document.querySelectorAll(value);
+  return Array.from(document.querySelectorAll(value));
 }
 
 function selectorStringify (value) {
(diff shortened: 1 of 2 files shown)


src/core/propertyTypes.js
```diff
@@ -85,7 +85,7 @@ function selectorParse (value) {
 function selectorAllParse (value) {
   if (!value) { return null; }
   if (typeof value !== 'string') { return value; }
```
darkwing: Should `value` also be wrapped in `Array.from()`?  What is the use case?  In case a `NodeList` instance is passed in?

dmarcos: We might want to return `[]` in the cases above?

ngokevin: I think the first case may be for `{default: ''}` so we don't need to run that.

The second case is if an array or NodeList is passed in yeah.

ngokevin:
looks good to me

ngokevin:
@darkwing thanks!

PR #1647, opened 2016-07-21
set use-credentials for crossorigin loading

ngokevin:
**Description:**

Setting to `use-credentials` seems harmless even the asset does not need credentials. But opens up in case people are hosting assets on S3 or something.


dmarcos:
Any downsides?

ngokevin:
Besides the request attaching browser session ID/cookie. We do it for videos https://github.com/aframevr/aframe/blob/master/src/systems/material.js#L277

Hopefully, this is the right way to set it.

PR #1648, opened 2016-07-21
Fix typo removeEventLister -> removeEventListener

bryik:
**Description:** Just ran into this while testing pause() on a component attached to the same entity as vive-controls.

**Changes proposed:**
- Change "removeEventLister" to "removeEventListener"


dmarcos:
thanks!

PR #1655, opened 2016-07-22
fix parents loading before children because parents tried to wait on children that were not ready (fixes #1652)

### LINKED ISSUE #1652: Parent entities are initializing before their child entities are loaded.
**Description:**

Entities should wait for all of their children to initialize before it starts initializing itself. We have logic for this where an entity grabs all of its child entities and waits for the `loaded` event. We use `el.isEntity` to filter its children. However, if the child entities are not yet fully attached (`nodeready`), then `isEntity` will be false.

- A-Frame Version: 0.2.0 + master

ngokevin:
**Description:**

Consider:

```html
<a-entity id="parent">
  <a-entity id="child"></a-entity>
</a-entity>
```

Currently, we want the child to completely attach + initialize (including all components) before the parent starts initializing. If the parent starts initializing when the child is not yet fully attached, then components are not initialized, then `getComputedAttribute('position')` will return a falsy value.

I ran into this case. The parent generally looks at all child **entities** and wait for them to load. We use `el.isEntity` to filter for entities. **However**, the children were not yet attached, and the `child.isEntity` was `undefined`. This makes the parent think it has no child entities and then it starts to initialize.

**Solution:**

We can detect that the child was attached by using the `nodeready` events. So we first wait for `nodeready` (if it is `isNode`), THEN wait for `loaded`. It seems that `isNode` is somehow initialized by this time

There was a hitch with assets because `` and `<video>` and such aren't nodes. So we have to add an argument that allows `<a-assets>` to configure which elements to wait for `nodeready`, e.g., only `<a-asset-item>`.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -63,8 +63,10 @@ module.exports = registerElement('a-assets', {
 
     load: {
       value: function () {
-        ANode.prototype.load.call(this, null, function waitOnFilter (el) {
+        ANode.prototype.load.call(this, null, function waitChildLoadedFilter (el) {
           return el.isAssetItem && el.hasAttribute('src');
+        }, function waitChildReadyFilter (el) {
+          return el.isAssetItem;
         });
       }
     }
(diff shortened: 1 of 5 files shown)


ngokevin:
Investigating an issue.

ngokevin:
Ready again.

ngokevin:
The root cause seems to have been fixed with the `registerElement` fixes. Closing for now.

PR #1657, opened 2016-07-22
revert back to crossorigin=anonymous

ngokevin:
**Description:**

I observed a case with UploadCare that setting use-credentials when you don't need it, or if the corresponding server header is not sent, then the request seems to fail. Revert back to **anonymous**.
Can still configure `use-credentials` on video/img elements, and later can add support for `<a-asset-item crossorigin>`.


donmccurdy:
Same issue with images in rawgit :/

PR #1663, opened 2016-07-25
removed unnecessary uses of 'a-scene'

blairmacintyre:
**Description:**
A number of places in the source assumed that the only possible scene would be an 'a-scene' element.  In most places in the code, however, the .sceneEl (property on any element contained in a scene) and .isScene (property on a scene) are used.  By directly using 'a-scene' (when searching for elements, and so on), we cannot create new kinds of scenes. 

**Changes proposed:**
- replace different references with existing alternatives (el.sceneEl, el.isScene)
- add new closestScene() method to a-node.js, and use it instead of closest('a-scene')
- add new findAllScenes() method to utils.js, and use it instead of document.querySelectorAll('a-scene');

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -99,7 +99,7 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   removeEventListeners: function () {
-    var sceneEl = document.querySelector('a-scene');
+    var sceneEl = this.el.sceneEl;
     var canvasEl = sceneEl && sceneEl.canvas;
     if (!canvasEl) { return; }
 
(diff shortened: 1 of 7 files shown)


blairmacintyre:
Ignore this.  I need to clean up the commit (I can't figure out why it decided all of scene.js was changed, but I doubt you want this in your build).  I'm going to start again.

ngokevin:
maybe because we use spaces and not hard tabs?

blairmacintyre:
I am not sure.  It only did it to that file.  I don't think I'm using hard tabs (using vs-code, I thought it defaulted to spaces).

PR #1665, opened 2016-07-25
fixes explicit uses of 'a-scene'

blairmacintyre:
Description:
A number of places in the source assumed that the only possible scene would be an 'a-scene' element. In most places in the code, however, the .sceneEl (property on any element contained in a scene) and .isScene (property on a scene) are used. By directly using 'a-scene' (when searching for elements, and so on), we cannot create new kinds of scenes.

Changes proposed:

replace different references with existing alternatives (el.sceneEl, el.isScene)
add new closestScene() method to a-node.js, and use it instead of closest('a-scene')
add new findAllScenes() method to utils.js, and use it instead of document.querySelectorAll('a-scene');

### DIFF
--- a/src/extras/declarative-events/index.js
+++ b/src/extras/declarative-events/index.js
@@ -38,7 +38,7 @@ module.exports = registerElement('a-event', {
           'Use https://github.com/ngokevin/aframe-event-set-component instead.');
 
         if (targetSelector) {
-          this.targetEls = this.closest('a-scene').querySelectorAll(targetSelector);
+          this.targetEls = this.el.sceneEl.querySelectorAll(targetSelector);
         } else {
           this.targetEls = [this.el];
         }
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -99,7 +99,7 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   removeEventListeners: function () {
-    var sceneEl = document.querySelector('a-scene');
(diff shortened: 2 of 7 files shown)


blairmacintyre:
I'd submitted a PR on this already, but it had extraneous stuff in it.  So, I cleaned out my repository, re-applied the changes an resubmitted.

dmarcos:
It looks good. The only concern is some problems we had in the past that might not be issues anymore. When using HTML templates the isScene might not be available (because the custom element is not yet initialized) while you can still check for a tagName. @ngokevin @cvan Do you have any use case in mind where these changes could break things?

src/extras/declarative-events/index.js
```diff
@@ -38,7 +38,7 @@ module.exports = registerElement('a-event', {
           'Use https://github.com/ngokevin/aframe-event-set-component instead.');
 
         if (targetSelector) {
-          this.targetEls = this.closest('a-scene').querySelectorAll(targetSelector);
+          this.targetEls = this.el.sceneEl.querySelectorAll(targetSelector);
```
ngokevin: We can probably leave this one since it's deprecated, just to make sure we don't break it

ngokevin: Actually, probably fine.

ngokevin:
Don't think there will be much side effects. HTML templates are no longer used. And the changes here look like `isScene` will be available.

PR #1669, opened 2016-07-26
Replace goggles icon with a fullscreen one when webVR is not availabl…

dmarcos:


dmarcos:
embedded

anime

ngokevin:
We can do this after, but:

- The fullscreen icon can be less pronounced. It looks a bit like a puzzle piece or a pattern.
- Reintroduce the modal that mentions VR.

dmarcos:
@ngokevin I would not reintroduce the modals. Modals are annoying and most people don't read them. The button behaves properly given the environment capabilities so the modals are redundant.

dmarcos:
An application seamlessly behaving in the best possible way given the available capabilities is what graceful degradation is about.

ngokevin:
ok sure

PR #1670, opened 2016-07-26
initialize systems only after scene attach

ngokevin:
**Description:**

Was trying to unit test a system that required a schema. But was unable to set the schema data before the system was initialized.

```js
var scene = document.createElement('a-scene');
// `my-system` already initialized when not ready.
scene.setAttribute('my-system', config);
```

- Move system init to attachedCallback for better control.

### DIFF
--- a/tests/components/scene/fog.test.js
+++ b/tests/components/scene/fog.test.js
@@ -2,16 +2,20 @@
 var entityFactory = require('../../helpers').entityFactory;
 
 suite('fog', function () {
-  'use strict';
-
-  setup(function () {
+  setup(function (done) {
     this.entityEl = entityFactory();
     var el = this.el = this.entityEl.parentNode;
-    this.updateMaterialsSpy = this.sinon.spy(el.systems.material, 'updateMaterials');
+    var self = this;
 
-    // Stub scene load to avoid WebGL code.
-    el.hasLoaded = true;
-    el.setAttribute('fog', '');
+    el.addEventListener('loaded', function () {
+      self.updateMaterialsSpy = self.sinon.spy(el.systems.material, 'updateMaterials');
+
+      // Stub scene load to avoid WebGL code.
+      el.hasLoaded = true;
+      el.setAttribute('fog', '');
+
+      done();
+    });
   });
 
   test('does not set fog for entities', function () {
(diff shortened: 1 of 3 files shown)


tests/components/scene/fog.test.js
```diff
@@ -2,16 +2,20 @@
...
-  'use strict';
-
-  setup(function () {
+  setup(function (done) {
     this.entityEl = entityFactory();
     var el = this.el = this.entityEl.parentNode;
-    this.updateMaterialsSpy = this.sinon.spy(el.systems.material, 'updateMaterials');
+    var self = this;
+
+    el.addEventListener('loaded', function () {
```
dmarcos: name this function

ngokevin: this is in a test, there are over 700 other instances in the tests of unnamed functions that just wait for loaded.

tests/core/scene/a-scene.test.js
```diff
@@ -28,6 +28,24 @@ suite('a-scene (without renderer)', function () {
     test('initializes scene object', function () {
       assert.equal(this.el.object3D.type, 'Scene');
     });
+
+    test('does not initialize systems', function (done) {
+      var sceneEl = document.createElement('a-scene');
+      process.nextTick(function () {
```
dmarcos: Do you really need `nextTick` here? I believe that createdCallback is called synchronously

ngokevin: thanks, removed

dmarcos: it's still around right?

dmarcos: weird. cache problems

dmarcos: In the conversation section of the PR I see the nextTick still but If I go to changes it's not there anymore. nextTick

tests/core/scene/a-scene.test.js
```diff
@@ -28,6 +28,21 @@ suite('a-scene (without renderer)', function () {
...
+    test('does not initialize systems', function () {
+      var sceneEl = document.createElement('a-scene');
+      assert.notOk(Object.keys(sceneEl.systems).length);
+    });
+  });
+
+  suite('attachedCallback', function () {
+    test('initializes systems', function (done) {
+      var self = this;
+      process.nextTick(function () {
```
dmarcos: Instead of doing next tick here it's probably more clear to wait for `loaded` in the setup function like we do here 
https://github.com/aframevr/aframe/blob/master/tests/components/geometry.test.js#L13

ngokevin: updated

Issue #1277, opened 2016-03-26
switch to browsersync


cvan:
maybe we should use something

> Budo is laggy

ngokevin:
#1671 browsersync

dmarcos:
I'm ok on how budo is working now and nobody else seems to have been complaining.

PR #1671, opened 2016-07-27
Replace budo with browserSync. Budo is not relialable and extremely s…

dmarcos:
…low on windows

@cvan @ngokevin I don't know if this is the most elegant way to deal with this but it's 100x faster on Windows :). I should have done this a looooong time ago. I was suffering in silence.


dmarcos:
@cvan I need your input on this

cvan:
yippie, I'm glad this is faster - testing…

dmarcos:
@cvan How does this look like?

scripts/bs-config.js
```diff
@@ -0,0 +1,48 @@
...
+      // dev against the examples :)
+      var path = urlParse(req.url).pathname;
+      if (path.indexOf('/' + 'dist/aframe.js') !== -1) {
+        req.url = req.url.replace('/dist/', '/build/');
+      }
+      next();
+    }
+  ],
+  rewriteRules: [],
+  files: ['build/aframe.js'],
```
cvan: I notice the auto-reloading no longer works for the Examples. also, should whitelist `examples/**`.

cvan:
sorry, forgot about this PR - besides the comment above, looks good. it's quite fast.

scripts/bs-config.js
```diff
@@ -0,0 +1,48 @@
+/*
```
ngokevin: perhaps name `browsersync-config.js`

package.json
```diff
@@ -5,9 +5,10 @@
   "homepage": "https://aframe.io/",
   "main": "dist/aframe.js",
   "scripts": {
-    "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
     "build": "mkdirp build/ && npm run browserify -- --debug -o build/aframe.js",
-    "dev": "npm run build && node ./scripts/budo",
+    "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
```
ngokevin: nit: perhaps keep alphabetized

scripts/bs-config.js
```diff
@@ -0,0 +1,48 @@
...
+      if (path.indexOf('/' + 'dist/aframe.js') !== -1) {
+        req.url = req.url.replace('/dist/', '/build/');
+      }
+      next();
+    }
+  ],
+  rewriteRules: [],
+  files: ['build/aframe.js'],
+  watchOptions: {ignoreInitial: true},
+  open: getEnvVar('BS_OPEN', false),
```
ngokevin: Do we want `true`? I sort of like having it open in the browser when I start the server.

ngokevin:
r+wc fast and 👍

fernandojsg:
Yes yes yes! :)

scripts/bs-config.js
cvan: IMO, it can get annoying, but I could go either way

scripts/bs-config.js
cvan: just wanted to make sure you saw this comment

cvan:
can you add "#1277" to the commit message?

scripts/bs-config.js
dmarcos: @cvan watchify observes the files that affect the browserify bundle. I don't know how to make it observe files that are not part of it. Any suggestion?

cvan: can't you just do this?

```js
files: [
  'build/aframe.js',
  'examples/**'
],
```

dmarcos: I don't believe that files attribute is to indicate what it's going to be watched. I've tried and it does not work. What is files exactly for? It's not a very descriptive name.

scripts/browserify-config.js
```diff
@@ -0,0 +1,48 @@
+/*
```
ngokevin: s/browserify/browsersync

ngokevin:
r+wc

scripts/browserify-config.js
```diff
@@ -0,0 +1,48 @@
...
+      // dev against the examples :)
+      var path = urlParse(req.url).pathname;
+      if (path.indexOf('/' + 'dist/aframe.js') !== -1) {
+        req.url = req.url.replace('/dist/', '/build/');
+      }
+      next();
+    }
+  ],
+  rewriteRules: [],
+  files: ['build/aframe.js'],
```
cvan: this is still missing `examples/**`

dmarcos: I had asked you. What does `files` do? Adding `examples/**` does not seem to do the trick

package.json
```diff
@@ -5,9 +5,10 @@
...
+    "browsersync" : "browser-sync start --config ./scripts/browserify-config.js",
```
cvan: this should be called `browsersync-config.js`, right?

package.json
```diff
@@ -5,9 +5,10 @@
...
+    "dev": "npm run watchify | npm run browsersync",
```
cvan: `browsersync` has the ability to watch files; why are we using both?

scripts/browserify-config.js
cvan: the comment got removed, so I didn't see this. because watchify's output is being piped to the browser-sync command, the changes aren't getting picked up. I'm working on a diff right now; will ping you when everything's working.

dmarcos:
@cvan As far as I understand `browserSync` just monitors the build file to reload the browser. Watchify observes the `src` files to regenerate the build. Any suggestion on how both should be combined?

cvan:
@cvan As far as I understand `browserSync` just monitors the build file to reload the browser. `Watchify` observes the `src` files to regenerate the build. Any suggestion on how both should be combined?

that's correct. if you're going to use `watchify`, you'll need to watch `{src,vendor}/**` to know when to rebuild `src/index.js` -> `build/aframe.js` and then `browser-sync` needs to watch `{build/aframe.js,examples/**}` to know when to refresh the JS bundle or the HTML page.

I spent some time trying to get it all to work without `watchify`. I have a patch, but it's not all working correctly:

```diff
diff --git a/package.json b/package.json
index bf10f65..af63cc5 100644
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "scripts": {
     "build": "mkdirp build/ && npm run browserify -- --debug -o build/aframe.js",
     "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
-    "browsersync" : "browser-sync start --config ./scripts/browserify-config.js",
-    "dev": "npm run watchify | npm run browsersync",
+    "browsersync" : "node ./scripts/browserify-config.js",
+    "dev": "npm run browsersync",
     "dist": "mkdirp dist/ && npm run browserify -s -- --debug | exorcist dist/aframe.js.map > dist/aframe.js && uglifyjs dist/aframe.js -c warnings=false -m -o dist/aframe.min.js",
     "gh-pages": "npm run ghpages",
     "ghpages": "node ./scripts/gh-pages",
@@ -22,7 +22,6 @@
     "test:firefox": "karma start ./tests/karma.conf.js --browsers firefox_latest",
     "test:chrome": "karma start ./tests/karma.conf.js --browsers Chrome",
     "test:ci": "TEST_ENV=ci karma start ./tests/karma.conf.js --single-run --browsers firefox_latest",
-    "watchify": "watchify src/index.js -o build/aframe.js",
     "version": "npm run dist"
   },
   "repository": "aframevr/aframe",
diff --git a/scripts/browserify-config.js b/scripts/browserify-config.js
index fb4dd2f..10d15dd 100644
--- a/scripts/browserify-config.js
+++ b/scripts/browserify-config.js
@@ -14,15 +14,22 @@
 
 var urllib = require('url');
 var urlParse = urllib.parse;
+
 function isEnabled (val) {
   val = val || '';
   return val !== '' && val !== '0' && val !== 'false' && val !== 'off';
 }
 
-function getEnvVar (name, defaultVal) {
+function getBoolEnvVar (name, defaultVal) {
   return name in process.env ? isEnabled(name) : defaultVal;
 }
 
+function getIntEnvVar (name, defaultVal) {
+  return name in process.env ? process.env[name] : defaultVal;
+}
+
+var bs = require('browser-sync').create();
+
 var opts = {
   server: {baseDir: './'},
   middleware: [
@@ -37,12 +44,34 @@ var opts = {
     }
   ],
   rewriteRules: [],
-  files: ['build/aframe.js'],
-  watchOptions: {ignoreInitial: true},
-  open: getEnvVar('BS_OPEN', false),
-  notify: getEnvVar('BS_NOTIFY', false),
-  tunnel: getEnvVar('BS_TUNNEL', false),
-  minify: getEnvVar('BS_MINIFY', false)
+  files: [
+    'examples/**',
+    {
+      match: ['src/**', 'vendor/**'],
+      fn: function (event, file) {
+        console.log('matched', event, file);
+        if (event === 'change') {
+
+          var browserify = require('browserify');
+          var b = browserify();
+          b.add('./src/index.js');
+          var fs = require('fs');
+          b.bundle().pipe(fs.createWriteStream('build/aframe.js'));
+          browsersync.reload('build/aframe.js');
+
+        }
+        /** Custom event handler **/
+        // console.log(event, file);
+      }
+    }
+  ],
+  // files: ['build/aframe.js', 'examples/*.html'],
+  // watchOptions: {ignoreInitial: true},
+  port: getIntEnvVar('PORT', 9000),
+  open: getBoolEnvVar('BS_OPEN', false),
+  notify: getBoolEnvVar('BS_NOTIFY', false),
+  tunnel: getBoolEnvVar('BS_TUNNEL', false),
+  minify: getBoolEnvVar('BS_MINIFY', false)
 };
 
-module.exports = opts;
+bs.init(opts);
```

I can try to keep digging into it. but if you want to keep `watchify`, we just have to make sure the reloading works correctly. because currently it doesn't, at least for me on Mac.

Even [`budo`](https://github.com/mattdesl/budo/blob/c4aa94428/package.json#L39) uses a [custom middleware for `connect` (the dev web server)](https://github.com/mattdesl/watchify-middleware`).

I tried seeing if anyone else has tried to combine `browserify` and `browser-sync`. They have with `gulp`, but I don't think it's worth adding yet another build tool.

My patch is nearly close to getting it to work, but it might be slower than just using `browser-sync` + `watchify`. What are your thoughts? IMO, let's do whichever's easiest (without adding Gulp though).

cvan:
in addition to changing the commit message, can we fix the typo of "reliable" ("relialable")?

cvan:
I have a _hunch_ that budo was unreliable and slow because of the connect server and middleware. Want to try switching out the [`connect`](https://github.com/senchalabs/connect) middleware ([`connect-pushstate`](https://github.com/brentertz/connect-pushstate)) for a BrowserSync one? Here's [some sample code](https://github.com/mattdesl/budo/issues/140#issuecomment-193360011) from a `budo` issue.

[`budo`](https://github.com/mattdesl/budo/) is powerful, designed specifically for browserify, is actively maintained by the community, and @mattdesl updates it very frequently. If BrowserSync makes Budo more reliable, I'd suggest trying that first. And, if that's the case for you, Diego, on Windows, let's file an issue for @mattdesl to consider making the same change.

mattdesl:
This is the first I've heard of budo being slow, I think as with most code it's just a matter of benchmarking to determine the culprit. Can you provide more details about what is slow? Maybe a test repo I can try on Windows?

cvan:
> This is the first I've heard of budo being slow, I think as with most code it's just a matter of benchmarking to determine the culprit. Can you provide more details about what is slow? Maybe a test repo I can try on Windows?

I think it happened when the `ecstatic` server was replaced by `connect`. I develop primarily on Mac OS X, and only test my scenes on Windows (and Chromebooks).

For other projects, I have resorted to using BrowserSync instead of `budo` for mainly these reasons:

1. a plethora of options (though I see you have an `--ssl` flag in `budo 0.9.0`, which is awesome.
2. out-of-the-box support for https://localtunnel.me/.
3. the main reason: the WebSocket hanging issues  I've had a lot of budo-specific issues with multiple HTML pages open and directories [i.e., any HTML file whose url is not `'/index.html'`] - I've filed issues for all those in the past. I don't think others could reproduce my issues. (FYI: I'm referring to budo/issues#79 and mattdesl#80 mostly, which I cannot tell whether they are fixed or not because they are intermittent but common enough to be a problem/nuisance.)

The HTTP server alone feels significantly faster. The headers are the mostly the same, and neither is sending gzip'd responses. So, I'm not sure what is accounting for the "hangs" (or "drops"?) @dmarcos is experiencing on Windows. All I can attest to lately is that `browser-sync` _feels_ much faster than `budo`.

`budo` is so easy to drop in a Browserify project and get going. I'd rather we not switch dev servers - or worse, go the Webpack route 😏

mattdesl:
Just FYI — I noticed you guys are still using budo in AFrame. If anybody is experiencing LiveReload hangs/issues with budo, try installing the following branch which does not rely on tiny-lr, but a custom WebSocket implementation. It should be a bit faster and more reliable and runs on the same server/port as budo.

```sh
npm install mattdesl/budo#feature/livereload-rework
```

See here for details:  
https://github.com/mattdesl/budo/issues/194

cvan:
@mattdesl: thanks for chiming in here. Since budo was using connect, I was having some problems with the speed (not sure if it was tiny-lr hanging or what). For other projects, I tried BrowserSync (which is just express under the hood, with a nice plugin system).

I saw your new branch and am actively using it for development of another project. Nice work on the perf improvements - definitely faster than any other watcher with a build step.

I haven't filed an issue because I'm trying to find reliable STR, but every so often when I make file changes, the server will die on me on some socket error. I'll capture it next time I see it.

Having multiple pages open with multiple browsers open, watching files *added*, etc. - those seem to be the factors I've noticed when I have to restart budo to pick up my changes.

I have a few related budo questions for you. Would be easier to chat over Slack or Twitter DM. When you have some time, hmu. Thanks!

P.S. I'll make a PR to A-Frame using the latest budo once this has all been figured out.

ngokevin:
@dmarcos Do you still have issues with Budo? I haven't run into much issues.

cvan:
I haven't had many issues. If we wanted to simplify the workflow and since most component authors are using webpack, we could switch to webpack 2. As complicated as I think webpack's config system is, version 2 supports a local dev server and easy prod building out of the box. @ngokevin: does webpack provide the nice error logging and https support that budō does?

ngokevin:
Yeah, pretty sure it does, given the huge ecosystem behind it. The workflow won't be simplified that much, you still run a dev server that rewrites paths to an in-memory build and you still have a build step to pack, just as we do now.

The benefits are smaller build size (someone reported shaved 100KB in one of the issues) and **simpler** config. If you see our current Browserify steps, they're a bit complicated to modify, and really hard to add additional steps like license-prepend or autoprefixer especially without messing source maps. But not high priority given other work.

dmarcos:
I have not seen much problems lately with budo. Let me spend a bit more time with it and I'll close this if it's not relevant anymore.

ngokevin:
I think we can close?

cvan:
I haven't seen any issues with budo. I think can close

PR #1678, opened 2016-07-27
grab/grabbing cursor on mousedown/mouseup (fixes #1167)

### LINKED ISSUE #1167: Change cursor to grab or crosshair when mousedrag is available
Description: Change cursor to grab or crosshair when mousedrag is available.

- A-Frame Version: Master
- Platform/Device: Desktop

ngokevin:
**Description:**

Visually indicate that canvas is draggable.

Or should this be done in look controls? Though perhaps it might be nice for other look controls to not have to duplicate this logic.

grab

**Changes proposed:**
- On canvas mousedown, set grab class.
- On document body mouseup, remove grab class.
- On canvas active and grab class active, set grabbing style.
- Add more canvas tests
- Typos

### DIFF
--- a/src/components/scene/canvas.js
+++ b/src/components/scene/canvas.js
@@ -1,10 +1,16 @@
-var register = require('../../core/component').registerComponent;
+var registerComponent = require('../../core/component').registerComponent;
 
-module.exports.Component = register('canvas', {
+var FULLSCREEN_CLASS = 'fullscreen';
+var GRABBING_CLASS = 'a-canvas-grabbing';
 
+/**
+ * Create canvas that toggles full screen styles when entering and exiting VR.
+ */
+module.exports.Component = registerComponent('canvas', {
   init: function () {
     var sceneEl = this.el;
     var canvasEl = document.createElement('canvas');
+
     canvasEl.classList.add('a-canvas');
     // Mark canvas as provided/injected by A-Frame.
     canvasEl.dataset.aframeCanvas = true;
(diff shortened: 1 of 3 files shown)


src/style/aframe.css
```diff
@@ -33,6 +33,19 @@
   width: 100%;
 }
 
+.a-canvas:hover {
+  cursor: grab;
+  cursor: -moz-grabbing;
+  cursor: -webkit-grab;
+}
+
+.a-canvas-grabbing,
```
dmarcos: `.a-canvas.grabbing`?

src/components/scene/canvas.js
```diff
@@ -14,42 +20,40 @@ module.exports.Component = register('canvas', {
...
+    });
+    document.body.addEventListener('mouseup', function removeGrabbingClass (event) {
+      document.body.classList.remove(GRABBING_CLASS);
     });
 
-    // Handle fullscreeen styling
-    sceneEl.addEventListener('enter-vr', addFullscreenClass);
-    sceneEl.addEventListener('exit-vr', removeFullscreenClass);
+    // Handle fullscreeen styling.
+    sceneEl.addEventListener('enter-vr', function addFullscreenClass () {
```
dmarcos: Can you move these down where they were?

dmarcos:
Yeah I think this should part of look-controls. What happens if there's no `look-controls`? Would I still get a grabbing hand?

src/components/scene/canvas.js
ngokevin: I think moving it down connotes that it's reused multiple times. And when they're separated from where they're used, it takes effort to see where it's being used.

src/style/aframe.css
ngokevin: This is on the body, could use a better class name, perhaps.

src/components/scene/canvas.js
ngokevin: Will probably be moot argument if I move it to look-controls though.

ngokevin: But there are good arguments for defining variables and functions close to where you are using them, I think it really reduces maintainance: https://www.youtube.com/watch?v=bOAJTR6rcPA

Issue #1679, opened 2016-07-27
declare-vars-on-top vs. declare nearby usage


ngokevin:
**Description:**

This coding style has been bothering me a bit. I know we should [avoid bikeshedding about code style](https://github.com/feross/standard#i-disagree-with-rule-x-can-you-change-it), but I find myself hesistant to make changes/move rapidly to avoid style wars. I want to resolve this by either coming to a clear consensus to a style, or to allow more freedom and don't enforce anything.

Currently, we are loosely enforcing [`declare-vars-on-top` style](http://eslint.org/docs/rules/vars-on-top). All variables must be declared at the top of their containing scope. The intent is to visually represent hoisting.

I think this style slows maintenance and coding. Below are several reasons.

### Difficult to Tell Where Variables are Used

We must read through and load into our heads several lines of code, and read the entire function to understand when and where they are being used. Also, when reading `Block B`, you must scroll up to see what the values are.

```js
function varsOnTop () {
  var a = 2;
  var b = 3;
  var c = this.c;
  var d = 4;
  var e = 5;

  // Block A.
  if (!c) { return 0; }

  // Block B.
  return a + b + d + e;
}
```

Versus here. We only need to know about `c` and understand that it returns early. If we wanted, we could stop reading. But reading further, it is very clear that `a`/`b`/`d`/`e` are used in `Block B` and not related to `Block A`.

```js
function varsWhereUsed () {
  // Block A.
  var c = this.c;
  if (!c) { return 0; }

  // Block B.
  var a = 2;
  var b = 3;
  var d = 4;
  var e = 5;
  return a + b + d + e;
}
```

### Hinders Refactoring

If we wanted to refactor `Block B` to a different function, we would have to scan and extract separated portions of the code.

```js
function varsOnTop () {
  var a = 2;  // Take this.
  var b = 3;  // This this.
  var c = this.c;  // Leave this.
  var d = 4;  // Take this.
  var e = 5;  // Take this.

  // Block A.
  if (!c) { return 0; }  // Leave this.

  // Block B.
  return a + b + d + e;  // Take this.
}
```

Versus just extracting the consolidated block altogether:

```js
function varsWhereUsed () {
  // Block A.
  var c = this.c;
  if (!c) { return 0; }

  // Take this.
  // Block B.
  var a = 2;
  var b = 3;
  var d = 4;
  var e = 5;
  return a + b + d + e;
}
```

### Forces Constant Code Restructuring for Minor Changes

[...]

PR #1680, opened 2016-07-27
grab/grabbing cursor on mousedown/mouseup (fixes #1167)

### LINKED ISSUE #1167: Change cursor to grab or crosshair when mousedrag is available

ngokevin:
**Description:**

On look-controls this time. #1678

grab

**Changes proposed:**

- Add class to canvas such that CSS handles grab class.
- On canvas mousedown, set grab class on body. On window mouseup, remove grab class from body.
- Add more canvas tests, from previous PR
- Typos

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -16,10 +16,20 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   init: function () {
+    var sceneEl = this.el.sceneEl;
+
     this.previousHMDPosition = new THREE.Vector3();
     this.setupMouseControls();
     this.setupHMDControls();
     this.bindMethods();
+
+    // Enable grab cursor class on canvas.
+    function enableGrabCursor () { sceneEl.canvas.classList.add('a-grab-cursor'); }
+    if (!sceneEl.canvas) {
+      sceneEl.addEventListener('render-target-loaded', enableGrabCursor);
+    } else {
+      enableGrabCursor();
+    }
   },
 
   update: function (oldData) {
@@ -236,11 +246,13 @@ module.exports.Component = registerComponent('look-controls', {
   onMouseDown: function (event) {
     this.mouseDown = true;
     this.previousMouseEvent = event;
+    document.body.classList.add('a-grabbing');
     event.preventDefault();
   },
 
   releaseMouse: function () {
     this.mouseDown = false;
+    document.body.classList.remove('a-grabbing');
   },
 
   onTouchStart: function (e) {
--- a/src/style/aframe.css
+++ b/src/style/aframe.css
@@ -33,6 +33,19 @@
   width: 100%;
 }
 
+.a-canvas.a-grab-cursor:hover {
+  cursor: grab;
+  cursor: -moz-grabbing;
+  cursor: -webkit-grab;
+}
+
+.a-grabbing,
+.a-canvas.a-grab-cursor:active {
+  cursor: grabbing;
+  cursor: -moz-grabbing;
+  cursor: -webkit-grabbing;
+}
+
 .a-canvas.fullscreen {
   width: 100% !important;
   height: 100% !important;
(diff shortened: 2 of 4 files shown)


src/components/look-controls.js
```diff
@@ -20,6 +20,15 @@ module.exports.Component = registerComponent('look-controls', {
     this.setupMouseControls();
     this.setupHMDControls();
     this.bindMethods();
+
+    // Enable grab cursor class on canvas.
+    var sceneEl = this.el.sceneEl;
```
ngokevin: @dmarcos 

This goes against the grain of the current style. See https://github.com/aframevr/aframe/issues/1679

I really like it this way because it consolidates everything related together in one visual chunk. If we wanted to extract, refactor, or remove this block, we simply take out this block and nothing else. If we moved `var sceneEl` above everything else, it'd take some reading that it's only used here.

I also put the function on top because it's a bit more logic. Declare it, and then use it.

dmarcos: Move this variable up. If you need it somwhere else in the function is easy to miss. All declarations go on top. No exceptions.

ngokevin: Can you please read the issue I filed? The current case is I don't need it somewhere else in the function. If you move it up, it implies it is used somewhere else and it is hard to tell where it is used.

https://github.com/aframevr/aframe/issues/1679

dmarcos: sceneEl might be needed in other part of the function and you won't see is defined. If it's only needed within the function make a local variable or create a closure to hide ir. There's a usual case to define a variable in a function that it's only use in a small section of the body. I still think it has to be moved up.

ngokevin: If we are worried about duplicate definitions, linters will catch that instantly.

I am not worried about scoping. Been following this style for months, I have been bothered and felt bogged down due to increased maintenance. I detailed the examples in #1679, every little change requires more code in different places while downgrading readability.

dmarcos: sceneEl is also a convention in our code use all over the place. Seeing the declaration does not add value but a little bit of visual clutter.

ngokevin: True, but just because that is the way it is now doesn't mean we couldn't improve it. Even in the case of `sceneEl`, I question the convention, there is still my problem of declaring variables up top and not knowing where/how they are used. And if I delete the block around the area where it *is* used, I have to scroll up and delete the declaration.

dmarcos: It adds cognitive load. is it this variable already defined? Is it naming the same thing I want? Where is it? Yes, the linter will catch some of the issues but the it's extra work to ammend and rename variables that could introduce new bugs. I really think there's zero value on having the variable here. Let's move it up. Let's be strict with code conventions and save the creativity and energy for features

dmarcos:
If we're 100% strict there's no room for misinterpretations or broken expectations

src/components/look-controls.js
ngokevin: Sorry, this code convention reduces my creativity and energy because it is such a drag on maintenance and readability. I've adopted most styles so far, but I wouldn't make a big fuss if I didn't think it really impacted my ability to quickly make changes.

Have you checked out #1679? I made arguments that it drastically **increases** cognitive load because you have to load more variables into your head for every function you read and write, and deduce which variables are used where. 

- `Is this variable already defined?` - Linter and text editors catch that. 
- `Is it naming the same thing I want?` - I am not sure what this means.
- `Where is it?` - I am not sure what that means either. I propose variables are declared where you use them, so they will be right next to where you are reading.

ngokevin:
Strictness is relative though. You could be strict one way or the other. I am hoping that either #1679 is accepted/refuted, or we agree to not debate on styles that our linter doesn't catch and follow our own ways.

PR #1692, opened 2016-07-30
Restructure flushToDOM code

darkwing:
This seemed structured weirdly so I made a more simple update.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -170,8 +170,9 @@ Component.prototype = {
    */
   flushToDOM: function () {
     var attrValue = this.attrValue;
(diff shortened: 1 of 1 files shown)


ngokevin:
A-Frame's style guide is a combination of https://github.com/Flet/semistandard and @dmarcos preferences (one if which is early returns and flattened logic blocks). So some things might be weird, but style guide is mostly set in stone.

darkwing:
My bad, didn't know that existed!  Sorry!

ngokevin:
No problem, takes getting used to :)

PR #1693, opened 2016-07-30
Nits for updateStateMixins

darkwing:
Another small nit to knock down  code.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -121,12 +121,11 @@ var proto = Object.create(ANode.prototype, {
       var diff = oldMixinsIds.filter(function (i) { return newMixinsIds.indexOf(i) < 0; });
       // Remove the mixins that are gone on update
       diff.forEach(function (mixinId) {
-        var forEach = Array.prototype.forEach;
-        // State Mixins
-        var stateMixinsEls = document.querySelectorAll('[id^=' + mixinId + '-]');
-        var stateMixinIds = [];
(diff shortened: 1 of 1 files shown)


src/core/a-entity.js
```diff
@@ -121,12 +121,11 @@ var proto = Object.create(ANode.prototype, {
...
       // Remove the mixins that are gone on update
       diff.forEach(function (mixinId) {
-        var forEach = Array.prototype.forEach;
-        // State Mixins
-        var stateMixinsEls = document.querySelectorAll('[id^=' + mixinId + '-]');
-        var stateMixinIds = [];
-        forEach.call(stateMixinsEls, function (el) { stateMixinIds.push(el.id); });
-        stateMixinIds.forEach(self.unregisterMixin.bind(self));
+        // State mixins
+        var stateMixinsEls = Array.from(document.querySelectorAll('[id^=' + mixinId + '-]'));
```
dmarcos: It seems that the support of `Array.from` is more limited than the previous form.

darkwing:
If we want to support old browsers without `Array.from`, we should back this out (https://github.com/aframevr/aframe/pull/1642) and use `Array.prototype.slice` (https://davidwalsh.name/nodelist-array).

dmarcos:
Yes. I think we should get rid of Array.from. We can open a different PR to remove the one introduced by #1642

darkwing:
Updated.  Feel free to just close if you don't want it

darkwing:
Closing in favor of another.

PR #1700, opened 2016-08-01
have entities wait on assets in order to make use of xhrloader cache

ngokevin:
**Description:**

Entities were not waiting for assets to load This caused duplicate asset fetching if an asset was defined in `<a-assets>` and an entity was also fetching an asset through a component.

This makes entities wait for `<a-assets>` (if exists) to load. This ensures all `<a-asset-item>`s have been fetched and made into the `THREE.Cache` used by `THREE.XHRLoader`. Subsequent fetches from loaders (e.g., `ColladaLoader`, `OBJLoader`, `MTLLoader`) will automatically pull from the cache.

Depended on https://github.com/aframevr/aframe/pull/1689, which has been merged.

**Changes proposed:**
- Entity wait for `<a-assets>` to load if exists.
- Update asset system documentation.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -83,9 +83,9 @@ registerElement('a-asset-item', {
     attachedCallback: {
       value: function () {
         var self = this;
-        var src = this.getAttribute('src');
-
+        var src = src = this.getAttribute('src');
         xhrLoader.load(src, function (textResponse) {
+          THREE.Cache.files[src] = textResponse;
           self.data = textResponse;
           ANode.prototype.load.call(self);
         });
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -68,8 +68,27 @@ var proto = Object.create(ANode.prototype, {
    */
   attachedCallback: {
     value: function () {
+      var assetsEl;  // Asset management system element.
+      var sceneEl = this.sceneEl;
+      var self = this;  // Component.
+
       this.addToParent();
+
+      // Don't .load() scene on attachedCallback.
       if (this.isScene) { return; }
+
+      // Gracefully not error when outside of <a-scene> (e.g., tests).
+      if (!sceneEl) {
+        this.load();
+        return;
+      }
+
+      // Wait for asset management system to finish before loading.
+      assetsEl = sceneEl.querySelector('a-assets');
+      if (assetsEl && !assetsEl.hasLoaded) {
+        assetsEl.addEventListener('loaded', function () { self.load(); });
+        return;
+      }
       this.load();
     }
   },
--- a/docs/core/asset-management-system.md
+++ b/docs/core/asset-management-system.md
@@ -6,29 +6,31 @@ parent_section: core
 order: 9
 ---
 
+A-Frame has an asset management system that allows us to place all of our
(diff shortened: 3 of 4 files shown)


src/core/a-assets.js
```diff
@@ -83,9 +83,11 @@ registerElement('a-asset-item', {
     attachedCallback: {
       value: function () {
         var self = this;
-        var src = this.getAttribute('src');
+        var src;
 
+        src = this.getAttribute('src');
```
dmarcos: Can declaration and assignment be in the same line?

src/core/a-entity.js
```diff
@@ -68,8 +68,18 @@ var proto = Object.create(ANode.prototype, {
...
+      var assetsEl;  // Asset management system element.
+      var self = this;  // Component.
+
       this.addToParent();
       if (this.isScene) { return; }
+
+      // Wait for asset management system to finish before loading.
+      assetsEl = this.sceneEl.querySelector('a-assets');
+      if (assetsEl && !assetsEl.hasLoaded) {
+        assetsEl.addEventListener('loaded', function () { self.load(); });
```
dmarcos: What about `assetsEl.addEventListener('loaded', this.load.bind(this));` and we can get rid of self.

dmarcos: Do we have a test to cover this?

src/core/a-assets.js
ngokevin: If we're going to do separate declarations, I prefer to keep them pure declarations when we can. If the intent is to separate declaration and logic and make the style 100% unquestionable with no exception, this does that.

I've found mixing declarations and initializations is what requires me to do frequent code restructuring (like in case we added an early return, or we needed assignments to go in a certain order).

src/core/a-entity.js
ngokevin: We've found binding is slow though. This will happen for every entity.

ngokevin: I can add a test.

ngokevin: Test added.

src/core/a-assets.js
dmarcos: So [in this function for instance](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L264) do I have to do?
````javascript
function () {
  var camera;
  var canvas;
  var embedded;
  var size;
  camera = this.camera;
  canvas = this.canvas;
  embedded =  this.getAttribute('embedded') && !this.is('vr-mode');
}
`````
It feels like a lot of clutter.

ngokevin: In that case, note that we got tempted to initialize `embedded` even though it might not even get used if it returns early: https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L266 

Mixing declaration/logic also messes up the alphabetization sometimes if the variables depend on each other.

I am not sure where to draw the line. Perhaps only allow pure declarations and aliasing (e.g., `var el = this.el`) in the declaration block. Everything else is logic and goes after.

ngokevin: I updated it. I guess I will just use discretion on when to declare + assign (small functions), and when to declare + assign later (large functions or functions with multiple paths).

PR #1701, opened 2016-08-02
Fix for the low display resolution on Android when going into VR mode

dmarcos:
**Description:**

a-scene `resize` method is attached to the resize event on the windows element. The resize event is triggered when entering/exiting fullscreen. The polyfill on mobile uses fullscreen to enter VR mode and the the `resize` method was replacing the renderer size set by VREffect (based on the getEyeParameters API) with the window height and width using the pixelRatio === 1 set by VREffect when entering VR.

**Changes proposed:**
1. the `resize` method returns if we're in mobile + VR mode. This way the renderer size set by VREffect is not replaced anymore.
2. Bumps the webvr-polyfill to the latest version.
3. Bumps three.js VREffect to avoid a problem when using the undefined renderRect value returned by the polyfilled `getEyeParameters` API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "style-attr": "^1.0.2",
     "three": "^0.76.1",
     "tween.js": "^15.0.0",
-    "webvr-polyfill": "borismus/webvr-polyfill#5fdd44b"
+    "webvr-polyfill": "0.9.15"
   },
   "devDependencies": {
     "browserify": "^11.0.1",
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -266,7 +266,11 @@ module.exports = registerElement('a-scene', {
         var embedded = this.getAttribute('embedded') && !this.is('vr-mode');
         var size;
         // Possible camera or canvas not injected yet.
-        if (!camera || !canvas) { return; }
+        // ON MOBILE the webvr-polyfill relies on the fullscreen API to enter
+        // VR mode. The canvas is resized by VREffect following the values returned
+        // by getEyeParameters. We don't want to overwrite the size with the
+        // windows width and height.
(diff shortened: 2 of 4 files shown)


ngokevin:
I tested locally and made sure I bumped my dependencies. I don't see any differences these changes made on iOS or Android.

dmarcos:
@ngokevin This is the results with master on a Galaxy S6
master

This the result with the patch:
patch

ngokevin:
Still not quite as clear as 0.2.0. What should I expect on iOS?

dmarcos:
Well, there's distortion applied so it looks... distorted. It's supposed to be better through the cardboard lenses. The polyfill returns what it considered the appropriate buffer side based on a database with info about each device.

dmarcos:
The improvement is the same in iOS. 
master branch master 
with this PR applied patched

dmarcos:
We're doing the correct thing now on the a-frame side. If there's any other concerns about image quality there should be dealt with in the `webvr-polyfill` by improving the [distorter](https://github.com/borismus/webvr-polyfill/blob/master/src/cardboard-distorter.js) or changing [the logic that returns the render target size](https://github.com/borismus/webvr-polyfill/blob/master/src/cardboard-vr-display.js#L96)

ngokevin:
Thanks for looking at this. This is a bit better. I do wonder if we can get the image quality up to 0.2.0 standards else it will still seem like a regression. Or perhaps the image quality differences don't matter when you're in VR. Will try them out side by side.

dmarcos:
@ngokevin Please let me merge the patches. Sometimes I have small tweaks in the comments or other things I want to add before merge.

ngokevin:
OK, sorry. It would help to adjust the GitHub label and the last comment sounded like you were waiting for a merge.

PR #1702, opened 2016-08-02
update docs

ngokevin:
**Description:**

I'm continuously updating the documentation and guides, but I want to sync what I have now.

**Changes proposed:**
- Organize documentation. *Introduction* -> *Guides* -> *Concepts & API* -> *Components* -> *Primitives*
- MagicaVoxel workflow guide.
- Move FAQ into documentation.


docs/guides/build-with-magicavoxel.md
```diff
@@ -0,0 +1,182 @@
...
+title: Build with MagicaVoxel
+type: guides
+layout: docs
+parent_section: guides
+order: 3
+---
+
+[magicavoxel]: https://ephtracy.github.io/
+
+[MagicaVoxel][magicavoxel] is a free and user-friendly tool that lets you build
```
dmarcos: Should documentation about workflows live somewhere else?

ngokevin: They should live here because people need discoverable guides on how to use A-Frame practically. Having them in the documentation ensures discoverability, versioning, and the ability to use relative links.

dmarcos: Where do we draw the line about what belongs to the aframe repo and what not? Teaching about basic web development and how to deal with CORS and set local web servers is something we should include as well? This docs also reference 3rd party components that create a liability if those parts break.

dmarcos: What about if the magicavoxel workflow gets outdated?

docs/primitives/a-camera.md
```diff
@@ -27,11 +29,14 @@ The camera primitive places the user somewhere within the scene. It is an entity
 | fov                   | camera.fov            | 80            |
 | look-controls-enabled | look-controls.enabled | true          |
 | near                  | camera.near           | 0.5           |
+| user-height           | camera.userHeight     | 0             |
```
mkungla: a-camera default user-height is now 1.6

docs/guides/build-with-magicavoxel.md
ngokevin: For documentation, teaching is paramount to drawing technical lines. Whatever helps people on their way. If people are constantly running into Slack/SO asking about CORS and web servers, it would serve us well to have guides. I wouldn't think of this as part of the core A-Frame repo, they're resources that happen to live in the repo.

The 3rd party components featured are curated and thoroughly QAed. If A-Frame is going to lean on third-party components, we should promote using them. Especially as we are becoming stricter on what is considered standard. 

We haven't had to do retrospective updates to the docs so far. But if needed, we can promote the `master` version of the workflow such that we can update freely. If there is a critical change that needs to be done on an old version, I am able to point the docs to different commits and tags.

dmarcos:
We should not promote any 3rd party components. We should not oficially play favorites and let the market decide. I know that the component is QAed but it will be in a year from now? That's a cost we would have to absorbe and one more distraction and moving piece. I know that we will do a manual curation of components for the first incarnation of our asset story but it will be temporary until it proves its value and we can justify the logistics of a community based curation

docs/primitives/a-camera.md
dmarcos: good catch!

ngokevin:
Ecosystem and community are among A-Frame's top line goals, not promoting solid third-party components goes against that. The cost of maintaining the components falls upon me, I already do it with over a dozen components. I don't think of giving developers what they need is a distraction.

These components actually are the current state of the market and ecosystem. I have a pretty complete grasp of component/scenes out there and their adoption, having searched/curated/built/published. And most scenes out there involve third-party components,

In a year, there will be a new set of documentation and guides that recommend other components, and that's fine.

PR #1703, opened 2016-08-02
Remove Array.from usage, use Array.prototype.slice instead

darkwing:
@dmarcos mentioned Array.from doesn't meet support requirements so this PR reverts to old method of NodeList to array conversion.


dmarcos:
Nice! Thanks!

PR #1705, opened 2016-08-02
Fix hand-controls events being attached twice (fixes #1704)

### LINKED ISSUE #1704: Hand-controls events being attached twice
**Description:** I think the hand-controls component is adding events twice, once during a [manual "addEventListeners()" call](https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L27) in init() and a second time when play() is called after update().

Also the ["stop()" method](https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L34) should be "pause()", right?

I can do a PR, if desired.

- A-Frame Version: master
- Platform/Device: all

bryik:
**Description:** The hand-controls component is adding events twice, once during a manual "addEventListeners()" call in init() and a second time when play() is automatically called after update(). Additionally, a few other changes were made based on [@ngokevin's comment on the issue](https://github.com/aframevr/aframe/issues/1704).

**Changes proposed:**
- Remove "addEventListeners()" call in init()
- Rename "stop()" to "pause()"
- ~~Move addEventListeners/removeEventListeners into pause() and play()~~
- Fix typo in component docstring ("HAndle" -> "Handle")

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -5,7 +5,7 @@ var RIGHT_HAND_MODEL_URL = 'https://media.aframe.io/controllers/hands/rightHand.
 /**
 *
 * Hand Controls component
-* HAndle events coming from the vive-controls
+* Handle events coming from the vive-controls
 * Translate button events to hand related actions:
 * gripclose, gripopen, thumbup, thumbdown, pointup, pointdown
 * Load a hand model with gestures that are applied based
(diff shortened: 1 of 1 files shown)


src/components/hand-controls.js
```diff
@@ -24,18 +24,9 @@ module.exports.Component = registerComponent('hand-controls', {
     this.onTrackpadUp = function () { self.handleButton('trackpad', 'up'); };
     this.onTriggerDown = function () { self.handleButton('trigger', 'down'); };
     this.onTriggerUp = function () { self.handleButton('trigger', 'up'); };
-    this.addEventListeners();
```
dmarcos: I think removing only this line will do the trick.

bryik: @dmarcos I don't think I removed anything else; the add/remove event listener methods were just rolled into pause() and play().

dmarcos: I like to keep event listener attachment logic in a separate method. Can we keep that code as it was?

bryik: Sure, I'll change it back.

bryik:
OK. I redid the last commit. The add/remove eventListener methods should be back as they used to be. 

Still new at this, so let me know if the PR didn't update correctly.

ngokevin:
Thanks!

PR #1706, opened 2016-08-02
a-node getChildren shorter array syntax

darkwing:
Much shorter than a `for` loop.

This is the third or fourth usage of `A.p.s.call` -- should we consider a `ANode.prototype.query` method for querying children and returning an `Array` as a standard?


dmarcos:
Thanks!

Issue #1628, opened 2016-07-11
Docs for position component - Y axis incorrect?


mike7ant:
**Description:**

Docs for 'position' component state the following for the y axis: "Positive Y Axis extends down.".  I believe the positive Y axis actually points up, which would be the expected orientation for right-handed coordinates.

- A-Frame Version:
0.2.0
- Platform/Device:
Any

ngokevin:
Yes, should be up :)

ngokevin:
Fixed thanks!

PR #1710, opened 2016-08-03
Fixed a mistake in docs for position component

halfzebra:
**Description:**
Issue #1628 

Documentation suggests, that: 

> Positive Y Axis extends down

It's the other way around.

**Changes proposed:**
Literally this:
` Negative Y axis extends down. Positive Y Axis extends up.`


dmarcos:
Nice thanks! Good catch

PR #1711, opened 2016-08-03
 Docs: event section of entity.md

FrederickDesimpel:
Not sure why git keeps adding the other commits... that's just keeping things synched right?

Anyways, had an error in my previous PR about this docs section, and some more updates:

**Docs: event section of entity.md**
fixed error in list of events
added event schemachanged
fixed error in componentchanged example
 using evt.detail instead of evt directly
added child-attached example


docs/core/entity.md
```diff
@@ -361,33 +361,49 @@ entity.is('selected');  // >> false
...
   }
 });
 ```
 
+#### Listening for child elements being attached
+
+We can use the `child-attached` event to listen for elements being attached:
+
+```js
+entity.addEventListener('child-attached', function(evt) {
```
dmarcos: is checking for nodeType necessary? is not checking for tagName enough?

FrederickDesimpel: yeah, maybe that is overdoing it...

ehm. Not sure if/how i need to change the pr?

dmarcos: iYeah that would be great. entities are the only elements that emit `child-attached` so the nodeType check is not necessary

dmarcos:
Nice Thanks!

PR #1713, opened 2016-08-04
Updated VREffect.js from three.js repo.

mkungla:


ngokevin:
thanks!

dmarcos:
Wait. what vreffect is this? master or dev branch?

mkungla:
@dmarcos I see, I got it from `dev` while it should be synced with `master` however this does not introduce major changes just some clean up. Since it will be in three master soon and vreffect is at this point manually synced with aframe I guess that will not cause problems now. However if your vreffect will stay in use then in future it should follow master.  Came across this update since I'm playing with this #1541 issue, but seems that vreffect can not provide same result that StereoCamera did at 0.2.0. So is it regression or how would user understand better that they deal with totally different effects!?

PR #1720, opened 2016-08-04
Entity: first attach the object3d then emit the child-attached event.

FrederickDesimpel:
**Description:**
Saw this when looking for info for the docs.

I might be wrong, but wouldn't it make more sense that the object3D is already present when the event is fired?

**Changes proposed:**
first attach the object3d then emit the child-attached event.


dmarcos:
thanks

PR #1721, opened 2016-08-04
Modified camera userHeight under examples.

mkungla:
After update default camera.userHeight and user-height some
examples needed logical adjustment to reflect that. And when you
are creating cameras with specific position you better use camera
component instead a-camera since it would make more sense there.
So to avoid that you'll have to position your camera
(prefered positon - camera.userHeight)


ngokevin:
I think rather than switching to use `<a-entity camera>`, we should set `user-height="0"` for some of these examples since they have no concept of standing.

mkungla:
that makes sense, will change to `<a-camera user-height...`

mkungla:
#1737 has to be fixed before I can confirm that changes have correct effect.

mkungla:
I checked all the examples

ngokevin:
Thanks!

PR #1725, opened 2016-08-05
Workaround for webkit bug where a webgl canvas element is not composited properly

dmarcos:
Workaround for https://bugs.webkit.org/show_bug.cgi?id=152556 that makes the canvas not have the proper size when changing its dimensions. This fix cover all the cases except when the VR mode is engaged in landscape orientation. An orientation change is needed to readjust the size of the canvas. I've spent a whole day and I could not find a workaround for that last glitch. I propose to live with it and wait for the fix to ship with iOS 10 (the bug above has been marked as fixed in April).

### DIFF
--- a/src/utils/forceCanvasResizeSafariMobile.js
+++ b/src/utils/forceCanvasResizeSafariMobile.js
@@ -1,0 +1,14 @@
+module.exports = function forceCanvasResizeSafariMobile (canvasEl) {
+  var width = canvasEl.style.width;
+  var height = canvasEl.style.height;
+  // Taken from webvr-polyfill (https://github.com/borismus/webvr-polyfill/blob/85f657cd502ec9417bf26b87c3cb2afa6a70e079/src/util.js#L200)
+  // iOS only workaround for https://bugs.webkit.org/show_bug.cgi?id=152556
+  // By changing the size 1 pixel and restoring the previous value
+  // we trigger a size recalculation cycle.
+  canvasEl.style.width = (parseInt(width, 10) + 1) + 'px';
+  canvasEl.style.height = (parseInt(height, 10) + 1) + 'px';
+  setTimeout(function () {
+    canvasEl.style.width = width;
+    canvasEl.style.height = height;
+  }, 200);
+};
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -8,6 +8,7 @@ module.exports.coordinates = require('./coordinates');
 module.exports.checkHeadsetConnected = require('./checkHeadsetConnected');
 module.exports.debug = require('./debug');
 module.exports.entity = require('./entity');
+module.exports.forceCanvasResizeSafariMobile = require('./forceCanvasResizeSafariMobile');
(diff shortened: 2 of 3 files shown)


dmarcos:
@mayognaise Can you try this patch to see if it looks better? Read the description above. There's a case that is not covered.

mayognaise:
@dmarcos cool ill check now!

mayognaise:
@dmarcos IT'S FIXED!!!! YEAHHHHHHH 🎉 🎈 🍰🎂🍭 💯

src/utils/index.js
```diff
@@ -8,6 +8,7 @@ module.exports.coordinates = require('./coordinates');
 module.exports.checkHeadsetConnected = require('./checkHeadsetConnected');
 module.exports.debug = require('./debug');
 module.exports.entity = require('./entity');
+module.exports.forceCanvasResizeSafariMobile = require('./forceCanvasResizeSafariMobile');
```
ngokevin: we could have a `mobile` namespace `utils.mobile.forceCanvasResizeSafari`

mayognaise:
Verified on:
- Safari/FF - iPhone5s/iPhone6
- Safari- iPad
- Chrome - Android Nexus5x

Remaining issue:
- Chrome - iPhone5s/iPhone6
image

- Safari/FF/Chrome - iPhone6s (The scale isn't correct by URL bar. It's fixed after orientation change)
image

dmarcos:
The bug @mayognaise mentions affects the webvr-polyfill as well. I filled https://github.com/borismus/webvr-polyfill/issues/140 to keep track of a solution. I don't think there's work specific to a-frame that has to be done.

ngokevin:
r+wc

src/utils/index.js
dmarcos: we can have a namespace is we have more mobile specific utils. I don't think it's worth it for a single one.

ngokevin: OK. We can merge, and I can update it to have a mobile namespace. We have 5 now, and it'd be best to future proof where it lives since utils are public.

src/utils/forceCanvasResizeSafariMobile.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function forceCanvasResizeSafariMobile (canvasEl) {
+  var width = canvasEl.style.width;
+  var height = canvasEl.style.height;
+  // Taken from webvr-polyfill (https://github.com/borismus/webvr-polyfill/blob/85f657cd502ec9417bf26b87c3cb2afa6a70e079/src/util.js#L200)
+  // iOS only workaround for https://bugs.webkit.org/show_bug.cgi?id=152556
+  // By changing the size 1 pixel and restoring the previous value
+  // we trigger a size recalculation cycle.
+  canvasEl.style.width = (parseInt(width, 10) + 1) + 'px';
```
cvan: I'm surprised adding `0.00001px` doesn't trigger a reflow

PR #1726, opened 2016-08-05
addToParent was invoked twice: In the load and attachedCallback methods

dmarcos:
This fixes the error of `parentEl === null`  seen on console in some of the demo like anime-ui.

This removes also what I believe is unnecessary sceneEl specific code since the same logic (updating components) is run for every entity (scene is just another entity).

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -43,25 +43,10 @@ module.exports.AAnimation = registerElement('a-animation', {
 
     attachedCallback: {
       value: function () {
-        var self = this;
-        var el = this.el = this.parentNode;
-
-        init();
-
-        function init () {
-          if (!el.isPlaying) {
-            el.addEventListener('play', init);
-            return;
-          }
-          if (!el.hasLoaded) {
-            el.addEventListener('loaded', init);
-            return;
-          }
-
-          self.handleMixinUpdate();
-          self.update();
-          self.load();
-        }
+        this.el = this.parentNode;
+        this.handleMixinUpdate();
+        this.update();
+        this.load();
       }
     },
 
@@ -172,7 +157,7 @@ module.exports.AAnimation = registerElement('a-animation', {
         var begin = data.begin;
         var end = data.end;
         // Cancel previous event listeners
-        if (this.evt) this.removeEventListeners(this.evt);
+        if (this.evt) { this.removeEventListeners(this.evt); }
         // Store new event name.
         this.evt = { begin: begin, end: end };
         // Add new event listeners
@@ -214,6 +199,12 @@ module.exports.AAnimation = registerElement('a-animation', {
 
     start: {
       value: function () {
+        var self = this;
+        // Postpone animation start until the entity has loaded
+        if (!this.el.hasLoaded) {
+          this.el.addEventListener('loaded', function () { self.start(); });
+          return;
+        }
         if (this.isRunning || !this.el.isPlaying) { return; }
         this.tween = this.getTween();
         this.isRunning = true;
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -241,23 +241,12 @@ var proto = Object.create(ANode.prototype, {
 
       if (this.hasLoaded) { return; }
 
-      // Attach to parent object3D.
-      this.addToParent();
-
-      // Scene load.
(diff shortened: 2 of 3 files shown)


src/core/a-entity.js
```diff
@@ -241,21 +240,11 @@ var proto = Object.create(ANode.prototype, {
 
       if (this.hasLoaded) { return; }
 
-      // Attach to parent object3D.
-      this.addToParent();
-
-      // Scene load.
-      function sceneLoadCallback () { self.updateComponents(); }
-      if (this.isScene) {
-        ANode.prototype.load.call(this, sceneLoadCallback);
```
ngokevin: The scene should call ANode.load *without* the `isEntity` filter. We want the scene to wait on *all* chk.d Nodes (including `<a-assets>`, whereas entities need only wait on all child entities.

src/core/a-entity.js
```diff
@@ -226,7 +226,6 @@ var proto = Object.create(ANode.prototype, {
   addToParent: {
     value: function () {
       var parentNode = this.parentEl = this.parentNode;
-
```
ngokevin: Not a big deal, but I liked to separate declaration with logic using whitespace.

src/core/a-entity.js
ngokevin: I wonder how the asset tests passed.

dmarcos: All entities are also nodes. Do we need to specify what each waits for? Both scene and entities wait for all the children nodes to load.

ngokevin: By default, nodes will wait for all child nodes to load. Entities have a special filter 'isEntity' that makes it only wait for entities. But scenes should wait on all nodes such as <a-assets< 

By maintaining the isEntity filter below, scenes will ignore assets theoretically. Perhaps we can remove that filter altogether, that might just be for animations? Though it'd be an additional change to the PR

dmarcos: working on it.

darkwing:
All references to `parentEl` could be removed here and elsewhere since it's not needed; `parentNode` should cover everything.  Or is `___El` a standard?

src/core/a-entity.js
```diff
@@ -241,21 +240,11 @@ var proto = Object.create(ANode.prototype, {
...
-      function sceneLoadCallback () { self.updateComponents(); }
-      if (this.isScene) {
-        ANode.prototype.load.call(this, sceneLoadCallback);
-        return;
-      }
-
       // Entity load.
       function entityLoadCallback () {
         self.updateComponents();
         // self.parentNode should work but that is null during this cb for unknown (#1483).
```
darkwing: This comment could probably be removed now.

dmarcos:
The naming is for consistency.

dmarcos:
I think I got equivalent and simpler code. I added a test to cover for https://github.com/aframevr/aframe/issues/1383

Now all entities wait for all the children nodes to load so we don't need specific logic for the scene anymore.

tests/core/a-animation.test.js
```diff
@@ -682,4 +679,42 @@ suite('a-animation', function () {
       });
     });
   });
+
+  suite('dynamic animations', function () {
+    test('animation plays when both entity and animation are dinamically created', function (done) {
```
ngokevin: typo s/`dinamically`/`dynamically`

tests/core/a-animation.test.js
```diff
@@ -682,4 +679,42 @@ suite('a-animation', function () {
...
+      var sceneEl = document.createElement('a-scene');
+      sceneEl.addEventListener('loaded', function () { createAnimation(); });
```
ngokevin: Don't need to create a function `addEventListener('loaded', createAnimation)`

src/core/a-animation.js
```diff
@@ -214,6 +199,11 @@ module.exports.AAnimation = registerElement('a-animation', {
 
     start: {
       value: function () {
+        // Postpone animation start until the entity has loaded
+        if (!this.el.hasLoaded) {
+          this.el.addEventListener('loaded', this.start.bind(this));
```
ngokevin: I'd use `self.start` since we've found that it's faster than binding

ngokevin:
r+wc

Issue #1624, opened 2016-07-10
Spot light position bug


FrederickDesimpel:
the cone of a spot light seems to be a unit higher then it's origin :
http://www.frederickdesimpel.be/test/webVR/Aframe/spot-bug.jpg

FrederickDesimpel:
not sure if it is the spotlight itself that is in the wrong place or the editor's cone visualisation.

msimpson:
Is this related to #1621?

FrederickDesimpel:
no this is just about the position of the object

FrederickDesimpel:
it has to do with the editor, position in aframe is correct.

fernandojsg:
@FrederickDesimpel can I have the example scene that you're using in your screenshot to test it on the editor?

FrederickDesimpel:
@fernandojsg i don't seem to have it around anymore... should be easy to recreate, it was just a spotlight and a sphere to check it's position...

can't seem to get into the editor anymore for the moment, so i can't recreate it now.

FrederickDesimpel:
@fernandojsg i have a scene that clearly shows the problem:

http://www.frederickdesimpel.be/test/webVR/Aframe/test11-spot-showbug-spotPos.html

the yellow line is from the light aframe element position, the white light is from the position of the light attached to it's object3D. it is exactly one unit too high.

FrederickDesimpel:
Made a temporary fix [here](https://github.com/aframevr/aframe/issues/1728). The functionality to the end user should be as expected. The real cause can be hunted upon later...

ngokevin:
fixed

PR #1728, opened 2016-08-05
Light target support. Fixed a bug with spotlight angle. temp fix for a bug with spotlight position.

FrederickDesimpel:
spot and directional light have a target, this lets the user set it with a selector.

when no target is set, the target is added to the light itself, with an offset of (0 0 -1), this intuitively let's you control the spotlight direction with it's rotation. (pointing -Z).
haven't done this with directional lights, so not to break anything, when someone sets directional by position offset, pointing to 0 0 0 by default, or to the new target.

a test is here : http://www.frederickdesimpel.be/test/webVR/Aframe/lighttarget/

might need further tests to see if anything strange happens when removing the component.

Also fixed a bug where spotlight angle was only set with degToRad when it is created, further updates were set in radians.

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -20,8 +20,8 @@ module.exports.Component = registerComponent('light', {
                 if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     penumbra: {default: 0, min: 0, max: 1, if: {type: ['spot']}},
     type: {default: 'directional',
-           oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']
-   }
+           oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']},
+    target: {type: 'selector', if: {type: ['spot', 'directional']}}
   },
 
   /**
@@ -30,6 +30,7 @@ module.exports.Component = registerComponent('light', {
   init: function () {
     var el = this.el;
     this.light = null;
+    this.defaultTarget = null;
     this.system.registerLight(el);
   },
 
@@ -40,16 +41,47 @@ module.exports.Component = registerComponent('light', {
     var data = this.data;
     var diffData = diff(data, oldData);
     var light = this.light;
+    var self = this;
 
     // Existing light.
     if (light && !('type' in diffData)) {
       // Light type has not changed. Update light.
       Object.keys(diffData).forEach(function (key) {
         var value = data[key];
-        if (['color', 'groundColor'].indexOf(key) !== -1) {
-          value = new THREE.Color(value);
+
+        switch (key) {
+          case 'color': {
+            light.color.set(value);
+            break;
+          }
+
+          case 'groundcolor': {
+            light.groundColor.set(value);
+            break;
+          }
+
+          case 'angle': {
+            light.angle = degToRad(value);
+            break;
+          }
+
+          case 'target': {
+            // Reset target if selector is null.
+            if (value === null) {
+              if (data.type === 'spot' || data.type === 'directional') {
+                light.target = self.defaultTarget;
+              }
+            } else {
+              // Target specified, set target using entity's `object3D`.
+              light.target = value.object3D;
+            }
+            break;
+          }
+
+          default: {
+            light[key] = value;
+          }
         }
-        light[key] = value;
       });
       return;
     }
@@ -60,8 +92,7 @@ module.exports.Component = registerComponent('light', {
 
   setLight: function (data) {
     var el = this.el;
-
-    var newLight = getLight(data);
+    var newLight = this.getLight(data);
     if (newLight) {
       if (this.light) {
         el.removeObject3D('light');
@@ -70,6 +101,68 @@ module.exports.Component = registerComponent('light', {
       this.light = newLight;
       this.light.el = el;
       el.setObject3D('light', this.light);
+
+      if (data.type === 'spot') {
+        // HACK solution for issue #1624
+        el.getObject3D('light').translateY(-1);
+        // set and position default lighttarget as a child to enable spotlight orientation
+        el.setObject3D('light-target', this.defaultTarget);
+        el.getObject3D('light-target').position.set(0, 0, -1);
+      }
+    }
+  },
+
+  /**
+   * Creates a new three.js light object given data object defining the light.
+   *
+   * @param {object} data
+   */
+  getLight: function (data) {
+    var angle = data.angle;
+    var color = new THREE.Color(data.color).getHex();
+    var decay = data.decay;
+    var distance = data.distance;
+    var groundColor = new THREE.Color(data.groundColor).getHex();
+    var intensity = data.intensity;
+    var type = data.type;
+    var target = data.target;
+    var light = null;
+
+    switch (type.toLowerCase()) {
+      case 'ambient': {
+        return new THREE.AmbientLight(color, intensity);
+      }
+
+      case 'directional': {
+        light = new THREE.DirectionalLight(color, intensity);
+        this.defaultTarget = light.target;
+        if (target) {
+          light.target = target.object3D;
+        }
+        return light;
+      }
+
+      case 'hemisphere': {
+        return new THREE.HemisphereLight(color, groundColor, intensity);
+      }
+
+      case 'point': {
+        return new THREE.PointLight(color, intensity, distance, decay);
+      }
+
+      case 'spot': {
+        light = new THREE.SpotLight(color, intensity, distance, degToRad(angle), data.penumbra, decay);
+        this.defaultTarget = light.target;
+        if (target) {
+          light.target = target.object3D;
+        }
+        return light;
+      }
+
+      default: {
+        warn('%s is not a valid light type. ' +
+           'Choose from ambient, directional, hemisphere, point, spot.', type);
+      }
     }
   },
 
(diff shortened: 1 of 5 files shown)


FrederickDesimpel:
Just found out it doesn't work on firefox.

Probably because directly setting the target which is an object3D, to another object3D.

Will try with a more elaborate way... to add() the object3D target to the seleted object...

FrederickDesimpel:
This is weird.

It seems that Chrome finds the key with

`if (['target'].indexOf(key) !== -1)`

while firefox doesn't...

FrederickDesimpel:
Maybe this whole foreach has to be rewritten as a regular loop?  i actually wanted to add the target to the object instead of setting the target the same as the object, which would need a continue..., and with three if checks, ...  it isn't really a foreach anymore...

FrederickDesimpel:
2nd commit fixes firefox... i didn't set the target on light creation. Which is odd that it worked in chrome then... does chrome fire the update handler multiple times ?

I added a switch case to check for the different cases of the key. Now you can also set target back to null.

[test example here](http://www.frederickdesimpel.be/test/webVR/Aframe/lighttarget/test2.html). try changing the target between #redsphere and #whitesphere or null in DOM inspector.

src/components/light.js
```diff
@@ -46,10 +46,37 @@ module.exports.Component = registerComponent('light', {
       // Light type has not changed. Update light.
       Object.keys(diffData).forEach(function (key) {
         var value = data[key];
-        if (['color', 'groundColor'].indexOf(key) !== -1) {
-          value = new THREE.Color(value);
+
+        switch (key) {
+          case 'color':
+            light[key] = new THREE.Color(value);
```
dmarcos: You don't need to instanciate a new color. You can just set it `light.color.set(value)`.

FrederickDesimpel: ah, that code was already there... I'll change it.

FrederickDesimpel: Since i'm checking for specific keys in the switch, i could also set the light.angle and light.target values directly, instead of with light[key] = ... would i change that too ?

dmarcos: Yeah I think you can also change the angle and the target

src/components/light.js
```diff
@@ -46,10 +46,37 @@ module.exports.Component = registerComponent('light', {
...
+          case 'color':
+            light.color.set(value);
+            break;
+
+          case 'groundcolor':
+            light.groundColor.set(value);
+            break;
+
+          case 'angle':
+            light.angle = degToRad(value);
```
ngokevin: This is already done in the spot light initialization.

src/components/light.js
```diff
@@ -46,10 +46,37 @@ module.exports.Component = registerComponent('light', {
       // Light type has not changed. Update light.
       Object.keys(diffData).forEach(function (key) {
         var value = data[key];
-        if (['color', 'groundColor'].indexOf(key) !== -1) {
-          value = new THREE.Color(value);
+
+        switch (key) {
+          case 'color':
```
ngokevin: Can we add braces around our switch cases?

```
case 'groundcolor': {
  // ...
  break;
}
```

src/components/light.js
```diff
@@ -20,8 +20,8 @@ module.exports.Component = registerComponent('light', {
                 if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     penumbra: {default: 0, min: 0, max: 1, if: {type: ['spot']}},
     type: {default: 'directional',
-           oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']
-   }
+           oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']},
+    target: {default: null, type: 'selector', if: {type: ['spot', 'directional']}}
```
ngokevin: Implicit default is `null` for `selector`.

src/components/light.js
ngokevin: Was color/groundColor not being set before?

src/components/light.js
```diff
@@ -46,10 +46,37 @@ module.exports.Component = registerComponent('light', {
...
+          case 'groundcolor':
+            light.groundColor.set(value);
+            break;
+
+          case 'angle':
+            light.angle = degToRad(value);
+            break;
+
+          case 'target':
+            if (value === null) {
```
ngokevin: I'd comment this block and also check that there's a target already set. Don't reset if there's no target.

```js
// Reset target if selector is null.
if (light.target && value === null) {

}
```

src/components/light.js
```diff
@@ -46,10 +46,37 @@ module.exports.Component = registerComponent('light', {
...
+              if (data.type === 'directional') {
+                light.target = new THREE.Object3D();
+              } else if (data.type === 'spot') {
+                light.target = new THREE.Object3D();
+                light.add(light.target);
+                light.target.position.set(0, 0, -1);
+              }
+            } else {
+              light.target = value.object3D;
```
ngokevin: And then comment:

```js
// Target specified, set target using entity's `object3D`.
light.target = value.object3D;
```

src/components/light.js
FrederickDesimpel: There's always a target. ThreeJs has it defined, but not added to the scene.

src/components/light.js
FrederickDesimpel: i'll leave that out.

src/components/light.js
ngokevin: OK I see. You could also check if there was a target existing before.

`oldData.target && value === null`

src/components/light.js
FrederickDesimpel: yes, and it need to be set again when updating.

src/components/light.js
FrederickDesimpel: they were being set to a new THREE.Color, since i changed the key checks to a switch, i set it there. I guess that is to convert color values?

I'll add the curly braces.

src/components/light.js
FrederickDesimpel: I thought the code was pretty self-explanatory...

ngokevin: Yeah. I just like to add comments to separate blocks of logic.

src/components/light.js
```diff
@@ -71,6 +105,15 @@ module.exports.Component = registerComponent('light', {
       this.light.el = el;
       el.setObject3D('light', this.light);
     }
+
+    if (!data.target && data.type === 'spot') {
```
ngokevin: Can you add a comment on the line above?

Not quite sure what this line does.

ngokevin: I think this and below should be done within `getLight` which handles creation of the light?

ngokevin:
Would you be okay adding tests for setting a target, updating a target, removing a target, updating angle?

`tests/components/light.test.js`

src/components/light.js
FrederickDesimpel: the diff is already checking that...

src/components/light.js
FrederickDesimpel: maybe...

FrederickDesimpel: it does the same as in the switch case.

FrederickDesimpel:
I've looked into testing, but it is still eluding me how to write them...

FrederickDesimpel:
I'm looking into testing... still eluding me a bit.

ngokevin:
I added a README https://github.com/aframevr/aframe/tree/master/tests#a-frame-unit-tests

src/components/light.js
```diff
@@ -46,10 +46,44 @@ module.exports.Component = registerComponent('light', {
...
+          case 'angle': {
+            light.angle = degToRad(value);
+            break;
+          }
+
+          case 'target': {
+            // Reset target if selector is null.
+            if (value === null) {
+              if (data.type === 'directional') {
+                light.target = new THREE.Object3D();
```
ngokevin: We could reuse a dummy so we aren't creating duplicate ones.

`var dummyTarget = new Object3D()` near the top of the file

tests/components/light.test.js
```diff
@@ -112,4 +121,68 @@ suite('light', function () {
...
+
+      assert.strictEqual(lightEl.object3D.children[0].target, othertargetEl.object3D);
+    });
+
+    test('when spotlight is created, light target is child object positioned at 0 0 -1', function () {
+      var lightEl = this.el;
+
+      lightEl.setAttribute('light', 'type: spot');
+
+      assert.typeOf(lightEl.object3D.children[0].target, 'object');
```
ngokevin: you can do `lightEl.getObject3D('light')`

tests/components/light.test.js
```diff
@@ -112,4 +121,130 @@ suite('light', function () {
...
+      var targetEl = document.createElement('a-entity');
+
+      targetEl.setAttribute('id', 'target');
+      sceneEl.appendChild(targetEl);
+
+      lightEl.setAttribute('light', 'type: spot');
+      lightEl.setAttribute('light', 'target', '#target');
+      lightEl.setAttribute('light', 'target', null);
+
+      assert.typeOf(lightEl.object3D.children[0].target, 'object');
```
ngokevin: you could use `var lightObject3D = lightEl.getObject3D('light')`

FrederickDesimpel:
Hold on.. i still have a bug...

FrederickDesimpel:
Ok. It should work pretty robust now.

Added a temporary fix for [issue #1624](https://github.com/aframevr/aframe/issues/1624), otherwise i couldn't get spotlight orientation correct.

there's a suite of unit test for light target, as wel as for both bugfixes.

FrederickDesimpel:
rebased and squached..

tests/components/light.test.js
```diff
@@ -109,5 +118,143 @@ suite('light', function () {
...
+
+      assert.strictEqual(lightEl.object3D.children[0].target, othertargetEl.object3D);
+    });
+
+    test('directional: when created, light target is positioned at 0 0 0, but NOT a child', function () {
+      var lightEl = this.el;
+
+      lightEl.setAttribute('light', 'type: directional');
+
+      assert.typeOf(lightEl.object3D.children[0].target, 'object');
```
ngokevin: You can do `lightEl.getObject3D('light')` all the other places in this file. And I would store it in a `var` to reduce clutter.

dmarcos: Yes. this is not robust.

PR #1734, opened 2016-08-08
Fix #1732: Don't bubble 'componentchanged' event

### LINKED ISSUE #1732: don't bubble `componentchanged` event
**Description:**

`Entity.emit` third argument is a boolean for whether to bubble. Flip that to false for `componentchanged`.

@blairmacintyre:

```

>Based on the documentation and example in entity.md, I set up a componentchanged listener on an entity that looks for changes to it's rotation component.  BUT, I have an animated entity in the entity graph below this node with an animation component on it.  Component.js does not set "bubbles" to false, so the events are bubbling up!  So, I was happily rotating whenever the child rotated (which was always).  Arg. (edited)

[5:54]  
I added
```      if (evt.target != this.el) { return; }```
to the start of my event handler.

[5:55]  
Which dealt with it.  But the question is:  should component changed events have the bubbles set to false (should the be bubbling??) or should the examples start including this test?

blairmacintyre [6:01 AM]  
It's understandable why bubbling is good by default:  I could set up one handler at the top of a subgraph, and handle things by explicitly looking at the target element.  BUT, the examples in the documentation should start doing this, even in the simple examples, so people have that in their minds when they write components or eventListeners.  :disappointed:

```

- A-Frame Version:
- Platform/Device:

darkwing:
As requested.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -210,7 +210,7 @@ Component.prototype = {
       name: this.name,
       newData: this.getData(),
       oldData: oldData
-    });
+    }, false);
   },
 
   /**


dmarcos:
Thanks!

PR #1736, opened 2016-08-08
Fix #798: Add 'aframe-injected' attribute to elements injected by aframe

### LINKED ISSUE #798: Add a class to the elements injected by aframe
<a href="https://github.com/fernandojsg"></img></a> **Issue by [fernandojsg](https://github.com/fernandojsg)**
_Monday Dec 28, 2015 at 21:18 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/758_

----

Add a class ```aframe-auto-whatever``` to all the elements injected by aframe that they were not part of the original html, to detect them with querySelector.

darkwing:
re: https://github.com/aframevr/aframe/issues/798

Added as an attribute to not muddle the CSS classes, while still able to select with QSA.  This also adds to only `a-` elements as well as the `<video>` created by `material.js`.  

Purposely did not add to meta tag, the original canvas, or the `vr-mode-ui` elements as they're simply for VR mode in/out.

Let me know what you think of this.

### DIFF
--- a/src/systems/camera.js
+++ b/src/systems/camera.js
@@ -1,4 +1,5 @@
 var registerSystem = require('../core/system').registerSystem;
+var constants = require('../constants/');
 
 var DEFAULT_CAMERA_ATTR = 'data-aframe-default-camera';
 var DEFAULT_USER_HEIGHT = 1.6;
@@ -39,6 +40,7 @@ module.exports.System = registerSystem('camera', {
       defaultCameraEl.setAttribute('camera', {active: true, userHeight: DEFAULT_USER_HEIGHT});
       defaultCameraEl.setAttribute('wasd-controls', '');
       defaultCameraEl.setAttribute('look-controls', '');
+      defaultCameraEl.setAttribute(constants.AFRAME_INJECTED, '');
       sceneEl.appendChild(defaultCameraEl);
       sceneEl.addEventListener('enter-vr', self.removeDefaultOffset);
       sceneEl.addEventListener('exit-vr', self.addDefaultOffset);
--- a/src/components/scene/vr-mode-ui.js
+++ b/src/components/scene/vr-mode-ui.js
@@ -1,4 +1,5 @@
 var registerComponent = require('../../core/component').registerComponent;
+var constants = require('../../constants/');
 var utils = require('../../utils/');
 
 var ENTER_VR_CLASS = 'a-enter-vr';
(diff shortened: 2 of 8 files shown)


src/systems/material.js
```diff
@@ -279,6 +279,7 @@ function createVideoEl (src, width, height) {
     warn('`$s` is not a valid video', src);
   }, true);
   videoEl.src = src;
+  videoEl.setAttribute('aframe-injected', '');
```
ngokevin: These videos are never injected to the DOM so don't need it here.

src/systems/camera.js
```diff
@@ -39,6 +39,7 @@ module.exports.System = registerSystem('camera', {
       defaultCameraEl.setAttribute('camera', {active: true, userHeight: DEFAULT_USER_HEIGHT});
       defaultCameraEl.setAttribute('wasd-controls', '');
       defaultCameraEl.setAttribute('look-controls', '');
+      defaultCameraEl.setAttribute('aframe-injected', '');
```
ngokevin: Could perhaps define this string as a constant somewhere. Maybe `constants/index.js`

ngokevin:
@fernandojsg The inspector is involved with this. Thoughts?

fernandojsg:
It's ok for me, as I could look for those `aframe-injected` and remove them before exporting or copying to clipboard the scene.

src/systems/camera.js
ngokevin: Let's define it there.

`AFRAME_INJECTED`

darkwing:
Updated!

ngokevin:
@fernandojsg Do you want the aframe-injected also on the VR mode UI, canvas, meta tags?

fernandojsg:
Yep that would be great! As right now I'm removing them by hand: 
https://github.com/aframevr/aframe-inspector/blob/b9e46216e005a1117a708943342cb589aed81524/src/lib/exporter.js#L39-L47
So what you said VR Mode UI, canvas, metatags, CSS, ...

ngokevin:
@darkwing would you mind remove those as well? I think the issue was primarily for @fernandojsg's inspector to generate HTML

darkwing:
Added to VR elements and metas, updated test.

ngokevin:
thanks!

fernandojsg:
@ngokevin @darkwing  ops, sorry I was a little bit late to the conversation, maybe it could be another PR, but what about including `aframe-injected` to the CSS?
`<style type="text/css" data-href="src/style/aframe.css">`

And probably for the stats too?
`<style type="text/css" data-href="src/style/rStats.css">`

PR #1738, opened 2016-08-08
Code cleanup for assets

darkwing:
- Shortens code to ones less QSA and one less `for` 
- Using forEach since `this.qsa` returns an `Array`


ngokevin:
@darkwing actually, how are we sure qsa returns an array again?

PR #1740, opened 2016-08-08
Update scene docs about VR effect

mkungla:
Update scene docs there are no props `monoRenderer` or `stereoRenderer`


ngokevin:
thanks!

PR #1741, opened 2016-08-09
count all relevant entities for aframe stats

FrederickDesimpel:
**Description:**
Only literal a-entity elements were counted for aframe stats

**Changes proposed:**
counts all <a-*> elements excluding a-animations,  a-assets and it's children.

Not sure yet how and if this stat is updated when children are parsed / attached.
An alternative is to keep track of entities being created.
An addition could be to give stats on the different kinds of entities.

code review please if this is good programming / performance, ...

### DIFF
--- a/src/lib/rStatsAframe.js
+++ b/src/lib/rStatsAframe.js
@@ -15,29 +15,16 @@ window.aframeStats = function (scene) {
   } ];
 
   function _update () {
-    _rS('te').set(getEntitiesCount());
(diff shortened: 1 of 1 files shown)


dmarcos:
It would be more robust to traverse all the scene children and check for the `isEntity` flag on them. Nothing prevents people to register a primitive with a prefix other than `a-XXX`. By checking `isEntity`  you don't need a special case for `a-animation` either.

PR #1742, opened 2016-08-09
closestScene already checks current node so don't duplicate check

darkwing:
Tiny cleanup; `closestScene` starts with the current node so no need to repeat it with the ternary.


dmarcos:
Thanks

PR #1743, opened 2016-08-09
Avoid forEach on NodeList for optimal backward support

darkwing:
@ngokevin made a good point about using `NodeList.prototype.forEach`/`map` which probably wont fly on older Android/iOS.  Stick with the old I guess.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -21,26 +21,28 @@ module.exports = registerElement('a-assets', {
       value: function () {
         var self = this;
         var loaded = [];
-        var medias = this.querySelectorAll('audio, videos');
+        var medias = this.querySelectorAll('audio, video');
         var imgs = this.querySelectorAll('img');
         var timeout = parseInt(this.getAttribute('timeout'), 10) || 3000;
+        var i;
 
         if (!this.parentNode.isScene) {
           throw new Error('<a-assets> must be a child of a <a-scene>.');
         }
 
         // Wait for <img>s.
-        imgs.forEach(function (img) {
+        for (i = 0; i < imgs.length; i++) {
+          var img = imgs[i];
           loaded.push(new Promise(function (resolve, reject) {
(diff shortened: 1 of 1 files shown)


mkungla:
@darkwing and @ngokevin wouldn't following be a solution? and perhaps add then polyfill into lib for compatibility.
```js
var medias = Array.from(this.querySelectorAll('audio, videos'));
var imgs = Array.from(this.querySelectorAll('img'));
```

darkwing:
@mkungla I submitted a PR with Array.from last week but @dmarcos thought we should use `Array.prototype.slice` instead so we could accommodate older browsers.

mkungla:
makes sense while there would be possibility to add polyfill for Array.from

darkwing:
@mkungla We could consider adding `querySelectorAll` to the `ANode` prototype which would return an array via `Array.prototype.slice.call(NodeList)` ; that way we're always working with an array when we query.

mkungla:
Selecting nodes without the need to traverse the DOM in tedious loops is nice, however @ngokevin and @dmarcos probably have better overview what performance impact querySelectorAll can have in this context for A-Frame.

ngokevin:
@davidwalsh I noticed a bug, it should be `video` not `videos`.

src/core/a-assets.js
```diff
@@ -30,17 +30,18 @@ module.exports = registerElement('a-assets', {
         }
 
         // Wait for <img>s.
-        imgs.forEach(function (img) {
+        for (var i = 0; i < imgs.length; i++) {
```
ngokevin: A-Frame likes `vars on top`, might have to add that to our linter.

ngokevin:
I'd file an issue for polyfilling NodeList forEach/map.

darkwing:
Issues fixed.

ngokevin:
Thanks!

PR #1744, opened 2016-08-09
Fixes #1737 and enables to update userHeight

### LINKED ISSUE #1737: camera.userHeight cannot be updated
**Description:**

Currently, changing camera height after camera initialization does not update the offset.

This was made apparent with the `<a-camera>` primitive that first sets the default userHeight to 1.6, and then mappings update the camera after that.

We should fix primitives such that they gather all data before initializing the component, but we could also fix the camera user height updating.

- A-Frame Version:
- Platform/Device:

mkungla:
Added camera.updateHeightOffset so that `camera.userHeight` and attr `user-height` can be updated when headset is not connected without conflicting `removeHeightOffset` and `addHeightOffset`

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
     "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
     "build": "mkdirp build/ && npm run browserify -- --debug -o build/aframe.js",
     "dev": "npm run build && node ./scripts/budo",
-    "dist": "mkdirp dist/ && npm run browserify -s -- --debug | exorcist dist/aframe.js.map > dist/aframe.js && uglifyjs dist/aframe.js -c warnings=false -m -o dist/aframe.min.js",
+    "dist": "npm run dist:min && npm run dist:max",
+    "dist:max": "npm run browserify -s -- --debug | exorcist dist/aframe.js.map > dist/aframe.js",
+    "dist:min": "npm run browserify -s -- --debug -p [minifyify --map dist/aframe.min.js.map --output dist/aframe.min.js.map] -o dist/aframe.min.js",
     "gh-pages": "npm run ghpages",
     "ghpages": "node ./scripts/gh-pages",
     "lint": "semistandard -v | snazzy",
@@ -26,7 +28,6 @@
   "repository": "aframevr/aframe",
   "license": "MIT",
   "dependencies": {
-    "browserify-css": "^0.8.2",
     "debug": "^2.2.0",
     "deep-assign": "^2.0.0",
     "document-register-element": "dmarcos/document-register-element#8ccc532b7",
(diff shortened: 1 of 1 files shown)


src/components/camera.js
```diff
@@ -87,6 +88,30 @@ module.exports.Component = registerComponent('camera', {
...
+  /**
+   * Enables updating camera.userHeight
+   */
+  updateHeightOffset: function () {
+    var currentPosition;
+    var headsetConnected;
+    var el = this.el;
+    var oldHeightOffset = this.userHeightOffset;
+
+    headsetConnected = this.headsetConnected || checkHeadsetConnected();
```
ngokevin: Could move this to the block where it is used:

```js

// Do not update camera.userHeight if headset is connected.
headsetConnected = this.headsetConnected || checkHeadsetConnected();
if (headsetConnected) { return; }

```

src/components/camera.js
```diff
@@ -87,6 +88,30 @@ module.exports.Component = registerComponent('camera', {
...
   },
 
+  /**
+   * Enables updating camera.userHeight
+   */
+  updateHeightOffset: function () {
+    var currentPosition;
+    var headsetConnected;
+    var el = this.el;
+    var oldHeightOffset = this.userHeightOffset;
```
ngokevin: Should we get rid of `this.userHeightOffset`? This is always available at `this.data.userHeight` and the old height is available through `update (oldData)`, and we don't want extra state.

ngokevin:
Could you add a test in `tests/components/camera.test.js` to test the offset is updated? Thanks!

src/components/camera.js
mkungla: agree `userHeightOffset` is superfluous

ngokevin:
Need a proper rebase.

mkungla:
yes I see

ngokevin:
What steps were you trying to rebase? Might be good to add to CONTRIBUTING.md guide.

Should be `git pull --rebase origin master` (if origin points to this aframevr/aframe upstream).

mkungla:
I had to rewind it that why that got closed.

ngokevin:
Hm, I usually never have to close a PR. Just want to know so I can document. I usually do `git rebase -i HEAD~10` to squash.

mkungla:
just do `git rebase upstream/,master` !
this one here got closed since rewinded and push state which was alredy merged github auto closes and marks it merged.

mkungla:
in my case I have `git remote -v`
```bash
origin	git@github.com:mkungla/aframe.git (fetch)
origin	git@github.com:mkungla/aframe.git (push)
upstream	git@github.com:aframevr/aframe.git (fetch)
upstream	git@github.com:aframevr/aframe.git (push)
```

so what ever branch you want to rebase just run
`git fetch --all` and
`git rebase upstream/master`

PR #1748, opened 2016-08-09
Fixes #1737 and enables to update userHeight

### LINKED ISSUE #1737: camera.userHeight cannot be updated

mkungla:
Added camera.updateHeightOffset so that `camera.userHeight` and attr `user-height` can be updated when headset is not connected without conflicting `removeHeightOffset` and `addHeightOffset`

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -36,6 +36,7 @@ module.exports.Component = registerComponent('camera', {
   bindMethods: function () {
     this.addHeightOffset = this.addHeightOffset.bind(this);
     this.removeHeightOffset = this.removeHeightOffset.bind(this);
+    this.updateHeightOffset = this.updateHeightOffset.bind(this);
     this.saveCameraPose = this.saveCameraPose.bind(this);
     this.restoreCameraPose = this.restoreCameraPose.bind(this);
   },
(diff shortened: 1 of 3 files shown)


mkungla:
continues #1744

tests/extras/primitives/primitives/a-camera.test.js
```diff
@@ -1,14 +1,38 @@
...
+  suite('active camera', function () {
+    test('is the active camera when applied', function (done) {
+      var camera = this.camera;
       assert.ok(camera.getComputedAttribute('camera').active);
       done();
     });
   });
+
+  suite('user-height', function () {
+    test('updates height offset when DOM user-height attibute changes', function () {
```
ngokevin: If you have asynchrony on your test, you have to accept the `done` argument and call it when finished.

```js
function (done) {
  process.nextTick(function () {
    assert...
    camera.setAttribute...
    process.nextTick(function () {
      assert...
      done();
    });
  });
}
```

src/components/camera.js
```diff
@@ -87,6 +88,30 @@ module.exports.Component = registerComponent('camera', {
...
   },
 
+  /**
+   * Enables updating camera.userHeight
+   */
+  updateHeightOffset: function () {
+    var currentPosition;
+    var headsetConnected;
+    var el = this.el;
+    var oldHeightOffset = this.userHeightOffset;
```
ngokevin: Are we going to remove `this.userHeightOffset` this PR?

mkungla: I think no since I can't test headset at the moment. and to remove `userHeightOffset` the logic of  `enter-vr` and  `exit-vr` have to be modified a bit.

ngokevin: OK

src/components/camera.js
```diff
@@ -128,8 +153,8 @@ module.exports.Component = registerComponent('camera', {
     var camera = this.camera;
     var system = this.system;
     if (!oldData || oldData.userHeight !== data.userHeight) {
-      this.removeHeightOffset();
-      this.addHeightOffset();
+      this.userHeightOffset = oldData.userHeight || 0;
```
ngokevin: How come we are setting this here?

mkungla: when you have scene where you would constantly update `userHeight` while you don't want to be applied when in VR with headset, but I guess when you exit-vr you would like to "restore" userHeight to what it was changed not to when you entered into VR - right?

ngokevin: OK, so this addresses if you update `userHeight` while in VR? That'd be an odd case, but we can maybe clean up later.

ngokevin:
Thanks! Tested, I was able to reset camera user height with a-camera.

PR #1749, opened 2016-08-09
Sets near plane to 5mm so closed objects are not clipped. Specially i…

dmarcos:
…mportant when manipulating with hand controls


ngokevin:
Can you update the docs?

PR #1756, opened 2016-08-10
wait for camera to be active before rendering (fixes #1733, fixes #1754)

### LINKED ISSUE #1733: renderer resolution is blurry on desktop
**Description:**

1. Open the Cube example.
2. See it is blurry.
3. window resize fixes it.

- A-Frame Version: master
- Platform/Device: desktop

### Blurry on desktop

screen shot 2016-08-09 at 3 29 03 pm

### After resize

screen shot 2016-08-09 at 3 29 06 pm

### LINKED ISSUE #1754: "THREE.WebGLRenderer.render: camera is not an instance of THREE.Camera.”
**Description:**

Error: `"THREE.WebGLRenderer.render: camera is not an instance of THREE.Camera.”`

I think this is because we are now delaying default camera injection until scene is `loaded`. Meaning it will try to render without a camera set up.

Perhaps have to listen to an earlier event for camera setup like before render. Or ask the scene to wait for the camera if doesn't exist.

- A-Frame Version:
- Platform/Device:

ngokevin:
**Description:**

Caused by https://github.com/aframevr/aframe/pull/1724

Since we waited for scene to load before setting up the camera. We have to tell the scene to wait for the camera as well.

- [x] unit tests

**Changes proposed:**
- Wait on `camera-set-active` before render if camera does not yet exist.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -313,19 +313,29 @@ module.exports = registerElement('a-scene', {
         }
 
         this.addEventListener('loaded', function () {
-          if (this.renderStarted) { return; }
+          AEntity.prototype.play.call(this);  // .play() *before* render.
 
-          AEntity.prototype.play.call(this);
-          this.resize();
+          // Wait for camera if necessary before rendering.
+          if (this.camera) {
+            startRender(this);
+            return;
+          }
+          this.addEventListener('camera-set-active', function () { startRender(this); });
 
-          // Kick off render loop.
-          if (this.renderer) {
-            if (window.performance) {
-              window.performance.mark('render-started');
+          function startRender (sceneEl) {
+            if (sceneEl.renderStarted) { return; }
+
+            sceneEl.resize();
+
+            // Kick off render loop.
+            if (sceneEl.renderer) {
+              if (window.performance) {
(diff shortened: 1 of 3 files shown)


src/core/scene/a-scene.js
```diff
@@ -313,19 +313,29 @@ module.exports = registerElement('a-scene', {
         }
 
         this.addEventListener('loaded', function () {
-          if (this.renderStarted) { return; }
+          AEntity.prototype.play.call(this);  // .play() *before* render.
```
dmarcos: Are you sure you want to start playing before rendering? What if there's components that rely on having an active camera?

ngokevin: All entities initialize before the scene gets here. With our without this patch, I believe they would have had to listen to the `camera-set-active` event, so nothing is changed here.

I had it within the `startRender` before but it broke some tests. I can try to play with it more.

ngokevin: But yeah, perhaps it's best to start execution and rendering at the same time.

ngokevin: I tried to get it working by calling `play` only after camera, but couldn't get the tests passing. I might have some code quality improvements though.

PR #1757, opened 2016-08-10
Removes assignment to CustomEvent.target

hugopeixoto:
**Description:**
`CustomEvent#target` is a read only property.
The removed line yielded an error when using strict mode.

**Changes proposed:**
- Remove assignment to `CustomEvent#target`


dmarcos:
Thanks

PR #1766, opened 2016-08-10
setTimeout to let conection close (fix #1765)

### LINKED ISSUE #1765: Composite and anime-ui examples broken in Chrome.
**Description:**

ColladaLoader does not work as expected if the model is defined in `a-assets` in Chrome. The logic of all entities to wait for `a-assets` before loading has introduced a side effect for the `colllada-model` component.  

Chrome behaves in a strange way in the case of  a nested `XMLHttpRequest` to the same URL. The inner request never finishes properly (if this is expected behavior or a bug I don't know). The problem is observed when adding a collada model as part of `a-assets` and a `collada-model` component in the scene that refers to the same url. The file is going to be fetched first by the `a-assets` request. The callback of that request invokes the `load` event that can trigger all the scene entities and their components to load. When the `collada-model` component updates is going to also do an `XMLHttpRequest` to load the model that will never return (notice that the ColladaLoader uses raw `XMLHttpRequest` instead of the `THREE.XHRLoader` so it does not benefit from `THREE.Cache` and the redundant `XMLHttpRequest` is sent).

dmarcos:
**Changes proposed:**
Postpones the load method call so the `XMLHttpRequest` closes before any `load` logic executes

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -82,7 +82,11 @@ registerElement('a-asset-item', {
         xhrLoader.load(src, function (textResponse) {
           THREE.Cache.files[src] = textResponse;
           self.data = textResponse;
-          ANode.prototype.load.call(self);
+          // In Chrome if another XMLHttpRequest is sent to the same url
+          // before the previous one closes. The second request never finishes.
+          // setTimeout finishes the first request and lets the logic
+          // triggered by load open subsequent requests.
+          setTimeout(function load () { ANode.prototype.load.call(self); });
         });
       }
     }


src/core/a-assets.js
```diff
@@ -82,7 +82,11 @@ registerElement('a-asset-item', {
         xhrLoader.load(src, function (textResponse) {
           THREE.Cache.files[src] = textResponse;
           self.data = textResponse;
-          ANode.prototype.load.call(self);
+          // In Chrome if another XHTMLHttpRequest is sent to the same url
+          // before the previous one closes. The second request never finishes.
+          // setTimeout finishes the first request and lets the logic
+          // triggered by load open subsequent requests.
+          setTimeout(function load () { ANode.prototype.load.call(self); }, 0);
```
ngokevin: Don't need a 0 right?

ngokevin: I might link to the issue

ngokevin:
r+wc

src/core/a-assets.js
```diff
@@ -82,7 +82,11 @@ registerElement('a-asset-item', {
         xhrLoader.load(src, function (textResponse) {
           THREE.Cache.files[src] = textResponse;
           self.data = textResponse;
-          ANode.prototype.load.call(self);
+          // In Chrome if another XHTMLHttpRequest is sent to the same url
```
ngokevin: XMLHttpRequest?

ngokevin:
Thanks for investigating!

PR #1782, opened 2016-08-12
Fix #1762: Implement faster bind function

### LINKED ISSUE #1762: Use a faster `Function.prototype.bind` utility function
**Description:**

[Bind is slow](http://stackoverflow.com/questions/18895305/will-function-prototype-bind-always-be-slow). The answer says that the native one does a lot of unnecessary stuff when we just want:

```js
function bind(fn, ctx) {
    return function bound() {
        return fn.apply(ctx, arguments);
    };
}
```

I'd probably put this in `src/utils/function.js` so it is resistant to future possible namespacing. And then replace all our calls of `bind` with the new faster bind

darkwing:
One idea I had and have seen before is, instead of `this.blah, this` all the time, doing `'blah', this` and assuming that `this` holds the function if the first argument is a string. i.e:

```
module.exports.bind = function (fn, ctx) {
  if(typeof fn === 'string') fn = ctx[fn];
  return function bound () {
    return fn.apply(ctx, arguments);
  };
};

bind('onKeyUp', this);
```

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -1,6 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 var utils = require('../utils/');
+var bind = utils.bind;
 
 var checkHeadsetConnected = utils.checkHeadsetConnected;
 
@@ -34,8 +35,8 @@ module.exports.Component = registerComponent('camera', {
     el.setObject3D('camera', camera);
 
     // Add listeners to save and restore camera pose if headset is present.
-    this.onEnterVR = this.onEnterVR.bind(this);
-    this.onExitVR = this.onExitVR.bind(this);
+    this.onEnterVR = bind(this.onEnterVR, this);
+    this.onExitVR = bind(this.onExitVR, this);
     sceneEl.addEventListener('enter-vr', this.onEnterVR);
     sceneEl.addEventListener('exit-vr', this.onExitVR);
   },
--- a/src/utils/bind.js
+++ b/src/utils/bind.js
@@ -1,0 +1,15 @@
+/**
+ * Faster version of Function.prototype.bind
+ * @param {Function} fn - Function to wrap.
+ * @param {Object} ctx - What to bind as context.
+ * @param {...*} arguments - Arguments to pass through.
+ */
+module.exports = function bind (fn, ctx/* , arg1, arg2 */) {
+  return (function (prependedArgs) {
+    return function bound () {
+      // Concat the bound function arguments with those passed to original bind
+      var args = prependedArgs.concat(Array.prototype.slice.call(arguments, 0));
+      return fn.apply(ctx, args);
+    };
+  })(Array.prototype.slice.call(arguments, 2));
+};
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -4,6 +4,7 @@
 var deepAssign = require('deep-assign');
 var objectAssign = require('object-assign');
 
+module.exports.bind = require('./bind');
 module.exports.coordinates = require('./coordinates');
 module.exports.checkHeadsetConnected = require('./checkHeadsetConnected');
 module.exports.debug = require('./debug');
--- a/tests/utils/bind.test.js
+++ b/tests/utils/bind.test.js
@@ -1,0 +1,47 @@
+/* global assert, suite, test */
+var bind = require('utils/bind');
+
+suite('utils.bind', function () {
+  test('utils.bind binds to object', function () {
+    var obj = {
+      propName: 'aframe',
+      getProp: function (arg) {
+        return this.propName;
+      }
+    };
+    assert.equal(obj.getProp(), bind(obj.getProp, obj)());
+  });
+
+  test('utils.bind binds properly when called by other object', function () {
+    var obj = {
+      propName: 'aframe',
+      getProp: function (arg) {
+        return this.propName;
+      },
+      getPropByCallback: function (cb) {
+        return cb();
+      }
+    };
+    var obj2 = {
+      propName: 'webvr'
+    };
+    var bound = bind(obj.getProp, obj2);
+    assert.equal(obj2.propName, bound());
+    assert.equal(obj2.propName, obj.getPropByCallback(bound));
+  });
+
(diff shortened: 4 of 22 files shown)


ngokevin:
Can we get some unit test coverage on it?

darkwing:
Yep; left off late at night.  Will add tests today!

src/utils/index.js
```diff
@@ -9,6 +9,7 @@ module.exports.checkHeadsetConnected = require('./checkHeadsetConnected');
 module.exports.debug = require('./debug');
 module.exports.entity = require('./entity');
 module.exports.forceCanvasResizeSafariMobile = require('./forceCanvasResizeSafariMobile');
+module.exports.functions = require('./functions');
```
ngokevin: Actually might be more consistent to have this as `function` rather than `functions`. (I imagine we'd have `object` and we already have `entity`). Hopefully not too much hassle to replace.

dmarcos: I don't think functions is a very good name. All modules export functions. I would just name the file `bind.js` an do `require('./bind');`

ngokevin: They'd be utility related to wrapping functions. I could imagine we'd add more function-related utilities in the future (e.g., debounce), and it'd be more organized and consistent to namespace them.

darkwing: Agreed @ngokevin.  We could add `forEach` (in some places we're doing `[].forEach` and others we're doing `Array.prototype.forEach(NodeList)`, etc.  I'm sure it will grow.

src/components/camera.js
```diff
@@ -1,6 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 var utils = require('../utils/');
+var bind = utils.function.bind;
```
dmarcos: I still think that the function namespace does not convey anything. Just call it `utils.bind`

ngokevin: It conveys function binding. What would `util.bind` bind?

I'd like it for organization and consistency above anything else though.

darkwing:
Test added.

src/components/camera.js
ngokevin: I don't think by itself it useful yeah. But I'd just like to be flexible in case we add http://underscorejs.org/#functions in the future.

tests/utils/function.test.js
```diff
@@ -0,0 +1,14 @@
...
+
+suite('utils.function', function () {
+  test('utils.function.bind binds to correct object', function () {
+    var obj = {
+      propName: 'aframe',
+      getProp: function () {
+        return this.propName;
+      }
+    };
+    assert.equal(obj.getProp(), bind(obj.getProp, obj)());
```
ngokevin: Would it be useful to have a test that binds to a different `obj` with different value?

tests/utils/function.test.js
```diff
@@ -0,0 +1,14 @@
+/* global assert, suite, test */
+var bind = require('utils/function').bind;
+
+suite('utils.function', function () {
+  test('utils.function.bind binds to correct object', function () {
```
ngokevin: I'd also add a test passing in arguments. Thanks!

src/components/camera.js
dmarcos: If we add other functions in the future we can add the namespace. Let's keep it simple for now and not anticipate.

ngokevin: Could, but I don't understand why we wouldn't want to avoid a potential breaking change for the sake of a typing a few extra characters.

dmarcos: I'm not sure we will need more functions other than `bind`. Why optimizing for something we are not sure that will ever happen. `function` is a such a generic name that the namespace does not add any information. it just feels like clutter.

dmarcos: `function` resonates with you because you're familiarized with `underscore`. Without that context is meaningless.

ngokevin: We may or may not, but I think it's smart to prepare for the possible scenarios. If we don't ever add more functions, then it might look cluttered (subjectively, I think it's organized). If we do and we end up namespacing it later, we have to call a breaking change and components using the function (possibly third-party) have to be migrated.

ngokevin: I wouldn't assume what resonates with me. It makes sense to me because the `function` utilities wrap functions, and there are various use cases of wrapping functions.

dmarcos: If you code trying to anticipate everything you will have super abstract unmaintainable code with optimizations that are there for no reason. Let's just add a bind file and generalize, create a namespace when there's a need.

dmarcos: We should not make those functions part of the public API

src/utils/bind.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Faster version of Function.prototype.bind
+ */
+module.exports = function bind (fn, ctx/* , arg1, arg2 */) {
```
darkwing: This function got more complicated because it doesn't accommodate for passing more arguments to `bind`, i.e. `bind(fn, context, VALUE1, VALUE2)`.  So I had to update it. :/

darkwing: We use the extra `VALUE1` in light.js

tests/utils/bind.test.js
```diff
@@ -0,0 +1,49 @@
+/* global assert, suite, test */
+var bind = require('utils/bind');
+
+suite('utils.bind', function () {
```
ngokevin: We can de-nest this suite now.

src/utils/index.js
```diff
@@ -9,6 +9,7 @@ module.exports.checkHeadsetConnected = require('./checkHeadsetConnected');
 module.exports.debug = require('./debug');
 module.exports.entity = require('./entity');
 module.exports.forceCanvasResizeSafariMobile = require('./forceCanvasResizeSafariMobile');
+module.exports.bind = require('./bind');
```
ngokevin: nit: could alphabetize these

src/utils/bind.js
ngokevin: Ah, good to have.

We can document this in the docstring

```js
* @params {...*} arguments - Arguments to pass through.
```

darkwing:
Updates made!

PR #1786, opened 2016-08-12
Fix for setting light target. updated unit tests and docs.

FrederickDesimpel:
**Description:**
A fix for an error with light target not being set when target declared after light.

**Changes proposed:**
Made setting light target async, upon target entity load.
Updated temp position bugfix #1624 to fix directional and hemisphere lights too.
Updated unit tests for both async fix and bugfix.
Added section to docs about setting directional light with orientation.

### DIFF
--- a/docs/components/light.md
+++ b/docs/components/light.md
@@ -50,6 +50,14 @@ The example below creates a light source shining from the upper-left at a 45-deg
 <a-entity light="type: directional; color: #EEE; intensity: 0.5" position="-1 1 0"></a-entity>
 ```
 
+It is also possible to specify the direction of the directional light with it's orientation by creating a child entity it targets. for example pointing down it's -Z axis:
+
+```html
+<a-light type="directional" position="0 0 0" rotation="-90 0 0" target="#directionaltarget">
+	<a-entity id="directionaltarget" position="0 0 -1"></a-entity>
+</a-light>
+```
+
 ### Hemisphere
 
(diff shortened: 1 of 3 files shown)


docs/components/light.md
```diff
@@ -50,6 +50,14 @@ The example below creates a light source shining from the upper-left at a 45-deg
 <a-entity light="type: directional; color: #EEE; intensity: 0.5" position="-1 1 0"></a-entity>
 ```
 
+It is also possible to specify the direction of the directional light with it's orientation by creating a child entity it targets. for example pointing down it's -Z axis:
```
ngokevin: Why would we just do `<a-light position="0 0 -1">`?

FrederickDesimpel: i'd guess some people would be more used to setting a direction by rotation.

ngokevin: I think the example below is a bit confusing because it includes `position` + `rotation` + `target` vs. just `position` or just `rotation`. Not sure why we'd want a nested target with `position` when we can just do `position`?

FrederickDesimpel: the nested target needs a position offset to set what relative position from the light object3D to point at. if it was 0 0 0 it wouldn't make sense, pointing from 000 to 000... So setting position of the target to 0 0 -1 sets it to target along it's Z axis.

last thing i am wondering about concerning this light target is if we should handle spotlights the same as above, ie with a nested target to allow for rotation instead of what is now comitted with spots having the nested target by default. This will break scenes that have a positioned spotlight that depended on the target to be at world 0 0 0. So either make that clear in the change notes... That to keep the old behaviour one must create a not nested target entity at 0 0 0. Or i need to commit a fix to do the same like directional lights, and we add the info how to do rotation to the spotlight docs too.

PR #1788, opened 2016-08-12
fix userHeight not being applied on mobile (fixes #1780)

### LINKED ISSUE #1780: camera.userHeight offset seems to not be applying on mobile
**Description:**

1. Open up Geometry Gallery on mobile http://localhost:9000/examples/test/geometry-gallery/
2. See you are clipped through the ground.

- A-Frame Version: master
- Platform/Device: iOS iPhone 6 Safari

ngokevin:
**Description:**

`updateHeightOffset` was not taking into account mobile.

**Changes proposed:**
- Simplify by removing `updateHeightOffset` since `addHeightOffset` can handle it if `oldData` is passed in.
- Fix enter VR since sometimes `removeHeightOffset` was getting called before `saveCameraPose`.
- Should not have to do `addHeightOffset` on exit VR now since `restoreCameraPose` handles that.

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -23,22 +23,21 @@ module.exports.Component = registerComponent('camera', {
    * Add reference from scene to this entity as the camera.
    */
   init: function () {
-    var camera = this.camera = new THREE.PerspectiveCamera();
-    var sceneEl = this.el.sceneEl;
-    this.el.setObject3D('camera', camera);
-    this.bindMethods();
-    sceneEl.addEventListener('enter-vr', this.removeHeightOffset);
-    sceneEl.addEventListener('enter-vr', this.saveCameraPose);
-    sceneEl.addEventListener('exit-vr', this.restoreCameraPose);
-    sceneEl.addEventListener('exit-vr', this.addHeightOffset);
-  },
+    var camera;
+    var el = this.el;
+    var sceneEl = el.sceneEl;
 
-  bindMethods: function () {
-    this.addHeightOffset = this.addHeightOffset.bind(this);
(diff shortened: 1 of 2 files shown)


src/components/camera.js
```diff
@@ -23,22 +23,25 @@ module.exports.Component = registerComponent('camera', {
...
-    sceneEl.addEventListener('exit-vr', this.restoreCameraPose);
-    sceneEl.addEventListener('exit-vr', this.addHeightOffset);
-  },
+    var camera;
+    var el = this.el;
+    var sceneEl = el.sceneEl;
+    var self = this;
+
+    this.savedPose = null;
+    this.enterVRHandler = function enterVRHandler () {
```
dmarcos: this should be a method of the component: 
````javascript
   onEnterVR: function() {
      this.saveCaemeraPose();
      this.removeHeightOffset();
   }
````

File: src/components/camera.js [hunk 2]
```diff
@@ -23,22 +23,25 @@ module.exports.Component = registerComponent('camera', {
...
+      self.saveCameraPose();
+      self.removeHeightOffset();
+    };
+    this.exitVRHandler = this.restoreCameraPose.bind(this);
```
dmarcos: this level of interaction just makes things harder to understand. You can just do 
````javascript
  sceneEl.addEventListener('exit-vr', this.restoreCameraPose.bind(this));
````

ngokevin: We need a reference to it if you want to detach the event listener later. I can move it to a method though.

dmarcos: This is I was I had a `bindMethods` function.

PR #1789, opened 2016-08-12
don't require people to set crossorigin on assets (fixes #1775)

### LINKED ISSUE #1775: Attempt automatically setting crossorigin flags on all assets within `<a-assets>`
**Description:**

If `crossorigin` is not defined.

- A-Frame Version:
- Platform/Device:

ngokevin:
**Description:**

If cross-origin media element is defined in `<a-assets>` without `crossorigin` attribute set, then `<a-assets>` will re-append the media element **with** `crossorigin` set to re-request with correct headers. three.js will then be able to access the contents of the resource.

- [ ] tests

This means **no more CORS confusions**!

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -20,33 +20,40 @@ module.exports = registerElement('a-assets', {
     attachedCallback: {
       value: function () {
         var self = this;
-        var loaded = [];
-        var medias = this.querySelectorAll('audio, video');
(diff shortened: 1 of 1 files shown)


wahengchang:
I am desperately needed to fix a cross domain issue, apparently this is not the solution, it shows the same error like before to me:

image

PR #1792, opened 2016-08-13
Address the parentNode === null error in Firefox 

dmarcos:
It is due to the use of `MutationObserver` in the `registerElement` polyfill. Default lights are first added to the scene and immediately removed by the light system when the scene lights initialize.  By the time the `attachedCallback` method is called for the default lights `parentNode === null` due to have been removed. The solution is postponing the default light creation until the scene has loaded and prevent the spurious default lights. This also makes the light and camera systems consistent.

### DIFF
--- a/src/systems/camera.js
+++ b/src/systems/camera.js
@@ -12,7 +12,10 @@ var DEFAULT_USER_HEIGHT = 1.6;
 module.exports.System = registerSystem('camera', {
   init: function () {
     this.activeCameraEl = null;
-    this.setupDefaultCamera();
+    // Wait for all entities to fully load before checking for existence of camera.
+    // Since entities wait for <a-assets> to load, any cameras attaching to the scene
+    // will do so asynchronously.
+    this.sceneEl.addEventListener('loaded', this.setupDefaultCamera.bind(this));
   },
 
   /**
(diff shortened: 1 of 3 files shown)


src/systems/light.js
```diff
@@ -1,62 +1,65 @@
...
+ *
+ * @param {bool} defaultLightsEnabled - Whether default lighting is active.
+ */
+module.exports.System = registerSystem('light', {
+  init: function () {
+    this.defaultLightsEnabled = null;
+    // Wait for all entities to fully load before checking for existence of lights.
+    // Since entities wait for <a-assets> to load, any lights attaching to the scene
+    // will do so asynchronously.
+    this.sceneEl.addEventListener('loaded', this.setupDefaultLights.bind(this));
```
dmarcos: This line is the only actual change in the file. The rest is just windows endline removal. I still have not figure out a way in Sublime to enforce unix line endings in every case. Sometimes switches automatically based on some heuristic I don't understand.

src/systems/camera.js
```diff
@@ -12,7 +12,10 @@ var DEFAULT_USER_HEIGHT = 1.6;
 module.exports.System = registerSystem('camera', {
   init: function () {
     this.activeCameraEl = null;
-    this.setupDefaultCamera();
+    // Wait for all entities to fully load before checking for existence of camera.
+    // Since entities wait for <a-assets> to load, any cameras attaching to the scene
+    // will do so asynchronously.
+    this.sceneEl.addEventListener('loaded', this.setupDefaultCamera.bind(this));
```
dmarcos: I moved the asynchrony outside of the `setupDefaultCamera` method.

tests/systems/light.test.js
```diff
@@ -1,50 +1,48 @@
...
+      assert.ok(lights[i].hasAttribute(constants.AFRAME_INJECTED));
+    }
+    assert.equal(lightsNum, 2);
+  });
+
+  test('removes default lights when more lights are added', function (done) {
+    var el = this.el;
+    var lightEl = document.createElement('a-entity');
+    lightEl.setAttribute('light', '');
+    lightEl.addEventListener('loaded', function () {
```
dmarcos: This test is the only change in the file. The rest is windows endlines removal.

FrederickDesimpel:
Since you wait for the scene to be loaded... the lights defined in the html will be there already right?
Don't you need to check for lights in the scene at the start of setupDefualtLights, and not create them then in the first place? That would also make it even more in line with camera.

dmarcos:
@FrederickDesimpel yeah! good point! I'm going to make some changes to adjust.

ngokevin:
r+

PR #1803, opened 2016-08-16
reduce buffer scale to 0.5 for iOS safari < 10 (fixes #1714)

### LINKED ISSUE #1714: The scene looks zoomed in after exiting VR on mobile
on Mobile, after exit VR mode (stereo view), the scene looks like zoomed in.
This only happens on master branch. `0.2.0` works fine.

- A-Frame Version: master branch
- Platform/Device: Mobile

ngokevin:
**Description:**

Workaround for [Webkit bug on iOS Safari < 10](https://bugs.webkit.org/show_bug.cgi?id=152556).

### Before

Upon Entering VR from landscape without orientation change:

img_3576

### After

Stereo now sized correctly, at the cost of decreased resolution. With the version check, this should automatically be addressed once people upgrade to iOS 10 as the Webkit patch rolls out.

img_3575

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,12 @@ window.WebVRConfig = window.WebVRConfig || {
   MOUSE_KEYBOARD_CONTROLS_DISABLED: true
 };
 
+// Workaround for iOS Safari canvas sizing issues in stereo (webvr-polyfill/issues/102).
+// Should be fixed in iOS 10.
+if (/(iphone|ipod|ipad).*os.*(7|8|9)/i.test(navigator.userAgent)) {
+  window.WebVRConfig.BUFFER_SCALE = 1 / window.devicePixelRatio;
+}
+
 // WebVR polyfill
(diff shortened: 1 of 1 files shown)


dmarcos:
I have to try this tomorrow on my iphone 6 Plus that has `devicePixelRatio === 3`

dmarcos:
The canvas size gets fixed when turning the phone to portrait and back to landscape. Is avoiding that glitch more important than keeping the high resolution?

ngokevin:
I think so, I don't think it's acceptable that people (not developers) have to know to rotate their phone back and forth. Many would think it was just broken and leave. With the version check, trying to have it automatically be addressed once iOS 10 starts rolling out.

I'll change it to `1 / window.devicePixelRatio` then for iPhone Plus.

dmarcos:
Yes, I confirm we need `1 / window.devicePixelRatio` to make it work on the iphone6 Plus

PR #1806, opened 2016-08-17
refactor canvas fullscreen code and add unit tests for enter/exit VR

ngokevin:
**Description:**

- Add unit tests for enter/exit VR also covering the canvas fullscreen styles.


src/core/scene/a-scene.js
```diff
@@ -76,7 +75,7 @@ module.exports = registerElement('a-scene', {
         var htmlEl = document.documentElement;
         htmlEl.classList.add('a-html');
         document.body.classList.add('a-body');
-        this.classList.add('fullscreen');
+        this.canvas.classList.add('fullscreen');
```
ngokevin: This is the relevant change.

src/core/scene/a-scene.js
```diff
@@ -85,7 +84,7 @@ module.exports = registerElement('a-scene', {
         var htmlEl = document.documentElement;
         htmlEl.classList.remove('a-html');
         document.body.classList.remove('a-body');
-        this.classList.remove('fullscreen');
+        this.canvas.classList.remove('fullscreen');
```
ngokevin: This is the relevant change.

src/core/scene/a-scene.js
```diff
@@ -64,8 +63,8 @@ module.exports = registerElement('a-scene', {
         this.addEventListener('render-target-loaded', function () {
           this.setupRenderer();
           this.resize();
+          this.addFullScreenStyles();
```
ngokevin: This is required so tests pass. Otherwise it will try to set canvas style without a canvas.

src/core/scene/a-scene.js
```diff
@@ -163,47 +185,60 @@ module.exports = registerElement('a-scene', {
           // TODO: 07/16 Chromium builds break when `requestFullscreen`ing on a canvas
           // that we are also `requestPresent`ing. Until then, don't fullscreen if headset
           // connected.
-          if (!self.isMobile && !checkHeadsetConnected()) { requestFullscreen(self.canvas); }
+          if (!self.isMobile && !self.checkHeadsetConnected()) {
```
dmarcos: Can we leave this as a on liner?

src/core/scene/a-scene.js
```diff
@@ -163,47 +185,60 @@ module.exports = registerElement('a-scene', {
...
         }
 
         function enterVRFailure (err) {
-          if (err && err.message) {
-            throw new Error('Failed to enter VR mode (`requestPresent`): ' + err.message);
-          } else {
-            throw new Error('Failed to enter VR mode (`requestPresent`).');
-          }
+          throw new Error('Failed to enter VR mode (`requestPresent`): ' +
+                          (err.message || ''));
```
dmarcos: What happens if `err` is not defined?

src/core/scene/a-scene.js
ngokevin: It went over 100 chars, might be hard to read on some terminals.

src/core/scene/a-scene.js
ngokevin: Fixed

dmarcos:
The problem is caused by the styles applied by the polyfill on the canvas element:
https://github.com/borismus/webvr-polyfill/blob/v0.9.15/src/cardboard-vr-display.js#L224
Our `position: fixed` is overridden 
https://github.com/aframevr/aframe/blob/master/src/style/aframe.css#L49
Making `position: fixed !important` solves the problem

dmarcos:
This one is risky. I would postpone this post 0.3.0

PR #1808, opened 2016-08-17
don't bind render on each tick

ngokevin:


src/core/scene/a-scene.js
```diff
@@ -45,13 +45,14 @@ module.exports = registerElement('a-scene', {
 
     createdCallback: {
       value: function () {
-        this.isMobile = isMobile;
+        this.init();
         this.isIOS = isIOS;
+        this.isMobile = isMobile;
         this.isScene = true;
         this.object3D = new THREE.Scene();
+        this.renderBinded = this.render.bind(this);
```
dmarcos: we can just do `this.render = this.render.bind(this)` as we do in other modules

src/core/scene/a-scene.js
```diff
@@ -45,13 +45,14 @@ module.exports = registerElement('a-scene', {
...
         this.systems = {};
         this.time = 0;
-        this.init();
```
dmarcos: I would kip init at the bottom. There's logic that might rely on the variables set above. For instance the logic that initializes systems might want to check for `isIOS` or `isMobile`

src/core/scene/a-scene.js
ngokevin: I get `a-scene.js:405 Uncaught TypeError: Cannot read property 'render' of undefined`

src/core/scene/a-scene.js
ngokevin: good call

src/core/scene/a-scene.js
dmarcos: Do you have to set ` writable: true` in the `render` method?

ngokevin: ok thanks that works

ngokevin:
Will fix the Codecov Travis failures soon. Need to remove the constraint that all PRs must improve coverage.

PR #1810, opened 2016-08-17
set important on canvas fullscreen style to not get overridden by webvr-polyfill

ngokevin:
**Description:**

Discovered and investigated by @dmarcos

### DIFF
--- a/src/style/aframe.css
+++ b/src/style/aframe.css
@@ -46,6 +46,7 @@
   cursor: -webkit-grabbing;
 }
 
+// Class is removed when doing <a-scene embedded>.
 .a-canvas.fullscreen {
   width: 100% !important;
   height: 100% !important;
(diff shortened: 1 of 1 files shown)


src/style/aframe.css
```diff
@@ -54,7 +54,7 @@
   right: 0 !important;
   bottom: 0 !important;
   z-index: 999999 !important;
-  position: fixed;
+  position: fixed !important;
```
cvan: can you add a comment? I hope nobody needs to ever override any of these important values. perhaps these styles would be better set with a `a.fullscreen.forced` class? I could be wrong. I haven't touched this code in a while.

ngokevin: Added a comment. If developers want to override, they can do `<a-scene embedded>` which removes these styles and allows styling of the canvas. That's also already documented.

PR #1815, opened 2016-08-18
Make spacing around objects consistent

darkwing:
This has been bothering me; need consistency! :)


ngokevin:
Can we consistency-tize it the other way?

darkwing:
We could but it seems a bit out of place; we require a space between `function` and `(`, and then `)` and `{`.  IMO spaces looks much more elegant and easy to read.

ngokevin:
I'd already been slowly moving it towards without spaces. I think spaces make them look like if or function blocks, and we don't have spaces for arrays.

```js
if (hello) { return this; }
var a = { b: this.c };
```

PR #1658, opened 2016-07-22
Revert "Initial .gitattributes to have clean release archives"

dmarcos:
Reverts aframevr/aframe#1586

PR #1825, opened 2016-08-19
Aframe gitattributes based on conversations #1586

mkungla:
Main purpose is to fix mixed linefeeds and have cleaner
master branch by flaging some files as binary so that git
won't create unnecessary big diff objects. Related to #1586, #1658
This is necessary so that added .gitattributes
could be dealt same as regular PR's and would not affect
new clones. No action needed by Collaborators since together with
second commit including fixed line feeds this PR can be treated as regular!
Some currently open PR's may need rebase with this commit if PR's
if PR have common files fixed with this PR.


mkungla:
After merge only thing to do is to bump new dist!

PR #1830, opened 2016-08-19
Ask for link to reproducable code within Issue template

darkwing:
This information would help us identify more easily if there's a real bug or if it's user error, and give us a realistic code snippet to use to fix a bug.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,3 +2,4 @@
 
 - A-Frame Version:
 - Platform/Device:
+- Reproducible Code Snippet or Demo URL [highly encouraged]:


mkungla:
@darkwing **Steps to reproduce** is definitely needed info in many cases. Besides that I find really useful how [three.js](https://github.com/mrdoob/three.js/issues/new) new issue templates are.

So even if you find a bug in `Windows`, `Chrome`, `Aframe 0.2.0` and I test it and get same bug in `Linux`, `FF`, `Aframe 0.3.1` then adding these gives better overview about extent of issue.

But I don't know what others think about that. @ngokevin , @dmarcos ?
If there is something else to be considered to add into new issue template perhaps you can add it to this PR.

darkwing:
Steps to reproduce are great but even better, IMO, is a working demo if available.  It's easier to ask for more and get less than visa versa.

ngokevin:
Most cases we can track down the issue given a simple description. In the contributing guide, it mentions a test case as optional/extra.

dmarcos:
Is it fair to replace the text with?
`Reproducible Code Snippet or Demo URL [highly encouraged]:`
If someone does not have the time to provide the test case... Do we prefer them not to file the issue?

darkwing:
Updated with @dmarcos 's idea; much softer.

dmarcos:
Thanks!

PR #1831, opened 2016-08-20
Add browserify env to .jshintrc

donmccurdy:
Nit that's been bugging me, in both Sublime and Atom I have JSHint plugins configured that automatically apply the nearest .jshintrc. In A-Frame's case there's nothing to indicate that `require()` and `module.exports` are allowed, so they're flagged:

screen shot 2016-08-19 at 11 34 19 pm


mkungla:
Same here has bugging me too with Atom

Issue #1490, opened 2016-05-20
Request: Make arrow keys work the same way as AWSD keys.


flukeout:
**Description:**
Allow movement through a scene using the arrow keys in addition to ASWD, which only really make sense for people who have played first person shooters on PC in the past.

While using a mouse and arrow keys together feels awkward position-wise, it seems like a good idea to allow it for the initial experience.

- A-Frame Version: Any
- Platform/Device: Desktop w/ Keyboard

ngokevin:
You can try https://github.com/donmccurdy/aframe-extras for now

There's a PR open for merging extensible controls in.

donmccurdy:
@dmarcos an simple-ish fix for this, #333, and #1409 (short of the whole controls refactor) would be to (1) support the arrow keys, and (2) use `event.code` instead of `event.keyCode` if it's provided. In browsers that implement it, that will reflect the physical location of the key so that Dvorak and ZQSD keyboards work as expected.

Example:

```javascript
// Tiny KeyboardEvent.code polyfill.
var KEYCODE_TO_CODE = {
    '38': 'ArrowUp',
    '37': 'ArrowLeft',
    '40': 'ArrowDown',
    '39': 'ArrowRight',
    '87': 'KeyW',
    '65': 'KeyA',
    '83': 'KeyS',
    '68': 'KeyD'
};

var code = event.code || KEYCODE_TO_CODE[event.keyCode];

```

darkwing:
Did a quick test on an example and I believe this can be closed with @donmccurdy 's PR

ngokevin:
thanks

Issue #1409, opened 2016-04-25
Enable Dvorak Controls?


prichey:
I'm a Dvorak user who is annoyed by having to switch modes every time I want to test out something with the WASD controls. I'm wondering if there has been any thought put into alternative keyboard layouts? I was thinking about making a PR for a new `dvorak-controls` component, but ultimately it'd just get used in development since the vast majority of production users will use QWERTY. I also had the idea of changing the `wasd-controls` component to something more general (`keyboard-controls`?), which could potentially allow the user to swap between WASD and Dvorak modes, maybe with some (optional) small text in the lower left that would appear on hover?

I'm willing to work towards a PR for this, I just wanted to open the issue to see if anybody else had any ideas, or if anyone has pursued this line of thinking as of yet.

donmccurdy:
Hi! Dvorak user here as well. :) I think this will be fixed in #1248. It's a larger PR but the relevant part is that using the newer KeyboardEvent.code API lets us map hotkeys by *locations*, without considering keyboard layout. This works better for ZQSD layouts as well. 

(only applies in browsers that support KeybardEvent.code, though - otherwise there's no good way to support both modes simultaneously)

prichey:
Excellent! Thanks @donmccurdy!

ngokevin:
dupe #333

PR #1832, opened 2016-08-20
Support for non-QWERTY keyboard layouts (ZQSD, AWSD, Dvorak, ...)

### LINKED ISSUE #333: Add ZQSD for controller

donmccurdy:
Uses `event.code` in browsers that implement it, which is layout-agnostic and should consistently correspond to the physical locations of the WASD keys, not their letter-equivalents.

Also adds <kbd>←</kbd><kbd>→</kbd><kbd>↑</kbd><kbd>↓</kbd> for good measure.

Would understand if we don't want to merge this while #1248 is still open, but this seems like a nice-to-have for users outside the U.S.

Fixes #333, #1409, and #1490.


dmarcos:
Thanks!

Issue #1635, opened 2016-07-15
A-Frame 0.4.0 Dependencies update


mkungla:
As dependencies where update last time around release of 0.2.0 then it would be wise to check over and update these for 0.4.0 since only some got updated for 0.3.1. Since some dependency updates may fix some issues, introduce new issues and some dependencies currently used are about become deprecated (or have their dependencies which are about to become deprecated)!

<table>
<caption><h3>Dependencies (Last updated @ 20 Aug 2016)</h3></caption>
<thead><tr><th>Package</th><th>Currently used </th><th>Latest Version</th><th>To be used for 0.4.0</th></tr></thead>

<tbody>

<tr>
<td><b>browserify-css</b></td>
<td><code>0.8.4</code></td>
<td><code>0.9.1</code></td>
<td>:grey_question: </td>
</tr><tr><td></td><td colspan="3">
This update adds a onFlush option and  bug fixes for `processRelativeUrl` and few other BUGS
<br>
</td></tr>

<tr>
<td><b>debug</b></td>
<td><code>^2.2.0</code></td>
<td><code>2.2.0</code></td>
<td><code>^2.2.0</code></td>
</tr><tr><td></td><td colspan="3">
No changes there
</td></tr>

<tr>
<td><b>deep-assign</b></td>
<td><code>^2.0.0</code></td>
<td><code>2.0.0</code></td>
<td><code>^2.0.0</code></td>
</tr><tr><td></td><td colspan="3">
No changes there
</td></tr>

<tr>
<td><b>document-register-element</b></td>
<td><code>#8ccc532b7</code> </td>
<td><code>1.0.0</code></td>
<td>:grey_question: </td>
</tr><tr><td></td><td colspan="3">
Should not use commits as dependency
<a href="https://github.com/WebReflection/document-register-element/pull/55"  target="_blank">WebReflection/document-register-element/pull/55</a>
</td></tr>

<tr>
<td><b>object-assign</b></td>
<td><code>^4.0.1 = 4.1.0</code> </td>
<td><code>4.1.0</code></td>
<td>:grey_question: </td>
</tr><tr><td></td><td colspan="3">
Current semver uses currently latest release
</td></tr>

<tr>
<td><b>present</b></td>
<td><code>0.0.6</code> </td>
<td><code>1.0.0</code></td>
<td>:grey_question: </td>
</tr><tr><td></td><td colspan="3">
No changes in code just 0.0.6 is considered to be 1.0.0
</td></tr>

[...]

PR #1837, opened 2016-08-20
[WIP] Updated dependencies related to #1635

mkungla:
I have tested these updates fairly extensively. Failed to update three to 0.79 because of failing tests on textures and material

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "present": "0.0.6",
     "promise-polyfill": "^3.1.0",
     "style-attr": "^1.0.2",
-    "three": "^0.76.1",
+    "three": "^0.81.0",
     "tween.js": "^15.0.0",
     "webvr-polyfill": "0.9.15"
   },


dmarcos:
Update three.js dependency is definitively due.

dmarcos:
I believe dependency update has to be deliberate: why do we want to update?

mkungla:
@dmarcos Better to do it now than later. As aframe is running for 0.4.0 many new futures will be added! Right? As currently there are long time warnings for some deps. about to become deprecated. while now you already see warnings that your deps will stop working since their deps will not be included in future releases etc. I find that this is a must that at the beginning of next version roadmap you update your deps and solve any exposed issues before you start building new futures on top of something which may not work anymore when you finally to the update. Another point here is that webvr and aframe are cutting edge therefore should not take to easy keeping deps up to date. Quality of Aframe can suffer in long term. Can't say that this will happen, but risk factor it is for sure. For example it was known after three 79 came out that Aframe has issues with it, but so far nothing has been done to fix that while three is nr 1 dep. I don't know if that makes any sense to you and do you get what I'm worried about, but I shared my opinion anyway :). And I totally understand your point ex with polyfill, but can that be as excuse not to update nevre again or if then when everything craches?

mkungla:
And if you have solid plan how to track that need of dep update, based on what to consider when to update - then please share. It would be useful info.

dmarcos:
> As currently there are long time warnings for some deps. about to become deprecated. while now you already see warnings that your deps will stop working since their deps will not be included in future releases etc. 

@mkungla What warnings are you referring to?

dmarcos:
> I find that this is a must that at the beginning of next version roadmap you update your deps and solve any exposed issues before you start building new futures on top of something which may not work anymore when you finally to the update.

Dependencies like `object-assign`,  `promise-polyfill` or `style-attr` play a very specific role and they fulfilled their role flawlessly. Why fixing what it's not broken and expose us to possible regressions or carrying additional features we don't need? A library increasing a version number does not mean that there are less bugs than the previous version or regressions have not been introduced.

dmarcos:
Can we change this PR to just update `three.js` and evaluate each dependency individually. For the `devDependencies` I'm not that worried.

mkungla:
i'm on gihub mobile at the moment and can't tell from head right now. Im back online within 3-4 h and can check then. But you can see by your self by installing aframe and (verbose) output of npm speaks for it self.

mkungla:
i'm on gihub mobile at the moment and can't tell from head right now. Im back online within 3-4 h and can check then. But you can see by your self by installing aframe and (verbose) output of npm speaks for it self.

mkungla:
Sure lets to three then.

mkungla:
@dmarcos 
So I revert then all, but devDependencies?

> Can we change this PR to just update three.js

Sure, but which version? `0.77.1` which has been tested and working or `0.79` which breaks and requires to fix blocking issues with `textures`.

> and evaluate each dependency individually

That lives here for a while #1635

dmarcos:
@mkungla can we move the devDependencies to a different PR?

mkungla:
sure, what is going to be then with three?

dmarcos:
@mkungla anything wrong with 0.80.0?

mkungla:
Yes, as build finishes check the build logs. These failures are with three  >= 0.78

mkungla:
@dmarcos I guess we have to start digging into these issues first and reference each issue to this PR?

dmarcos:
@mkungla Yeah, it seems something is different with the `THREE.TextureLoader`. We have to investigate what has changed. Thanks for looking into this.

ngokevin:
This has been done. Thanks though! We still need r82.

PR #1842, opened 2016-08-22
Fix #899: Add warning for wrong URL format

### LINKED ISSUE #899: Throw warnings if developer forgets the `url( )` wrapper for srcs.
For properties of type `src` (propertyTypes.js), if developer does ```srcProp: myimage.png``` instead of ```srcProp: url(myimage.png)```, throw a console warning. This can be detected if the querySelector fails and there isn't a ```#``` at the beginning of the string.

https://github.com/aframevr/aframe/blob/master/src/core/propertyTypes.js#L117

darkwing:
**Description:**
Fix for #899 

**Changes proposed:**
- Simply add warning for it
- Added a few tests for giggles.

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -2,6 +2,7 @@ var coordinates = require('../utils/coordinates');
 var debug = require('debug');
 
 var error = debug('core:propertyTypes:warn');
+var warn = debug('core:propertyTypes:warn');
 
 var propertyTypes = module.exports.propertyTypes = {};
 
@@ -116,6 +117,11 @@ function srcParse (value) {
 
   var el = selectorParse(value);
   if (el) { return el.getAttribute('src'); }
+
+  if (value.charAt(0) !== '#') {
+    warn('"' + value + '" is not a valid `src` attribute. ' +
+         'Value must be an ID selector (i.e. "#someElement") or wrapped in `url()`.');
+  }
 
   return '';
 }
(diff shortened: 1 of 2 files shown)


darkwing:
On PTO today (ill); will add test for selector as soon as I get a moment.

darkwing:
Updated

src/core/propertyTypes.js
```diff
@@ -117,6 +118,10 @@ function srcParse (value) {
   var el = selectorParse(value);
   if (el) { return el.getAttribute('src'); }
 
+  if (value.charAt(0) !== '#') {
```
ngokevin: Selectors can be a class or attribute or data selector.

ngokevin: Or any element name

darkwing: Right; I was following your logic here:  https://github.com/aframevr/aframe/issues/899

I'd assumed you made ID selection a standard.  Is your `#` comment no longer true?

ngokevin: Sorry, you're right. I was thinking selector not src

src/core/propertyTypes.js
```diff
@@ -117,6 +118,10 @@ function srcParse (value) {
...
+    warn('Value provided for `src` attribute is not valid: ' + value + '. Value must be selector notation or wrapped in `url()`.');
```
ngokevin: Should we mention this is either ID or `url()`, rather than just "selector notation"?

darkwing: Yes, I'll update the language this morning.

darkwing:
Updated with mores specific message.

PR #1847, opened 2016-08-24
Inspector test

darkwing:
**Description:**
Adds test for Inspector

**Changes proposed:**
None

### DIFF
--- a/tests/components/scene/inspector.test.js
+++ b/tests/components/scene/inspector.test.js
@@ -1,0 +1,29 @@
+/* global assert, setup, suite, teardown, test */
+
+suite('inspector', function () {
+  setup(function (done) {
+    var el = this.sceneEl = document.createElement('a-scene');
+    document.body.appendChild(el);
+
+    el.addEventListener('loaded', function () { done(); });
+  });
+
+  teardown(function () {
+    var el = this.sceneEl;
+    el.parentNode.removeChild(el);
+  });
+
+  test('adds inspector SCRIPT element to page', function () {
+    var el = this.sceneEl;
+
+    // Have to call method directly because Chrome doesn't provide
+    // a reliable method for KeyboardEvent
+    el.components.inspector.onKeydown({
+      keyCode: 73,
+      ctrlKey: true,
+      altKey: true
+    });
+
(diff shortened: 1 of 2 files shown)


tests/components/scene/inspector.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+  teardown(function () {
+    var el = this.sceneEl;
+    el.parentNode.removeChild(el);
+  });
+
+  test('adds inspector SCRIPT element to page', function (done) {
+    var el = this.sceneEl;
+
+    el.addEventListener('loaded', function () {
```
ngokevin: We could call this in the setup

ngokevin:
We can get full coverage if we also test the `remove` (detached) callback. This should remove the event listener when the components is detached (with `removeAttribute`).

tests/components/scene/inspector.test.js
```diff
@@ -0,0 +1,42 @@
...
+      var i;
+      var matchingScripts = 0;
+      var scripts;
+
+      // Have to call method directly because Chrome doesn't provide
+      // a reliable method for KeyboardEvent
```
ngokevin: Maybe we should emit a CustomEvent on `window`?

tests/components/scene/inspector.test.js
```diff
@@ -0,0 +1,42 @@
...
+      el.components.inspector.onKeydown({
+        keyCode: 73,
+        ctrlKey: true,
+        altKey: true
+      });
+
+      scripts = document.querySelectorAll('script');
```
ngokevin: something like `assert.ok(document.querySelector(script[src="..."]))` ?

darkwing: Illegal selector :/

ngokevin: We could add `data-name="aframe-inspector" aframe-injected` attributes to the injected script tag.

darkwing:
Updates:

1.  Added the attributes you suggested; I held off originally to avoid code bloat for sake of test but since we're already using `aframe-injected` elsewhere, probably good to be consistent.

2.  Not adding `remove()` test since default components cannot be removed at present.

ngokevin:
Thanks!!

Issue #1293, opened 2016-03-29
Link to Chrome for Android `autoplay` bug in FAQ + video docs


cvan:
https://bugs.chromium.org/p/chromium/issues/detail?id=178297

PR #1850, opened 2016-08-24
Fix #1373, #1293 - Update video docs with regard to webkit-playsinline

### LINKED ISSUE #1373: Update docs to mention that `webkit-playsinline` gets set on all videos
https://github.com/aframevr/aframe/blob/faddf6f/docs/primitives/a-videosphere.md#caveats
https://github.com/aframevr/aframe/blob/faddf6f/docs/primitives/a-video.md#caveats

```
2. Set the `webkit-playsinline` attribute to the video element.
```

Since #1179, all videos now have this attribute and can be played inline when added to homescreen.

darkwing:
Like the `apple-mobile-web-app-capable` point, adding note that this attribute is automatically added by A-Frame so users don't think they need to add it.

### DIFF
--- a/docs/introduction/faq.md
+++ b/docs/introduction/faq.md
@@ -83,7 +83,9 @@ Because of an [iOS platform restriction][iosvideo] in order to get inline video
 Inline video support on iOS 10 may change this. On certain Android devices or
 browsers, we must:
 
-- Require user interaction to trigger the video (such as a click or tap event).
+[android-touch-bug]: https://bugs.chromium.org/p/chromium/issues/detail?id=178297
+
+- Require user interaction to trigger the video (such as a click or tap event). See [Chromium Bug 178297][android-touch-bug].
 
 We will try to create a video boilerplate that has all the configurations to
 work on mobile devices.
--- a/docs/introduction/faq.md
+++ b/docs/introduction/faq.md
@@ -76,8 +76,8 @@ Mobile browsers have limitations with displaying inline video.
 Because of an [iOS platform restriction][iosvideo] in order to get inline video
 (with or without autoplay), we must:
 
-- Set the `<meta name="apple-mobile-web-app-capable" content="yes">` metatag.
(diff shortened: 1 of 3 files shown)


docs/introduction/faq.md
```diff
@@ -71,19 +71,21 @@ Uploader][uploader], a service that serves files with CORS headers set.
 
 [iosvideo]: https://developer.apple.com/library/iad/documentation/UserExperience/Conceptual/iAdJSProgGuide/PlayingVideosinAds/PlayingVideosinAds.html
 
+[android-touch-bug]: https://bugs.chromium.org/p/chromium/issues/detail?id=178297
```
ngokevin: I've been starting to add the link definition above the paragraph where it's used

docs/introduction/faq.md
```diff
@@ -71,19 +71,21 @@ Uploader][uploader], a service that serves files with CORS headers set.
...
+
 Mobile browsers have limitations with displaying inline video.
 
 Because of an [iOS platform restriction][iosvideo] in order to get inline video
 (with or without autoplay), we must:
 
-- Set the `<meta name="apple-mobile-web-app-capable" content="yes">` metatag.
-- Set the `webkit-playsinline` attribute to the video element.
+- Set the `<meta name="apple-mobile-web-app-capable" content="yes">` meta tag (will be injected if missing).
+- Set the `webkit-playsinline` attribute to the video element (is automatically added to all videos).
```
ngokevin: I wonder if iOS will honor it if we add it dynamically

docs/introduction/faq.md
darkwing: Updated!

ngokevin: Looks like it's still here?

darkwing:
I thought you had meant section; updated.

ngokevin:
Thanks, we can apply the same to the other couple parts

PR #1851, opened 2016-08-24
Updates to guide language

darkwing:
There's been talk of updating the scenes with a more expansive rewrite but I've quickly made some language updates to the guides as they appear now.

### DIFF
--- a/docs/guides/build-with-components.md
+++ b/docs/guides/build-with-components.md
@@ -10,7 +10,7 @@ order: 2
 
 ![360&deg; Image Viewer](/images/docs/360-image-viewer.png)
 
-Let's go through an example building a scene using an
+Let's create an example of building a scene using an
 [entity-component-system][ecs] workflow. This guide will introduce three
 concepts:
 
--- a/docs/guides/build-with-magicavoxel.md
+++ b/docs/guides/build-with-magicavoxel.md
@@ -8,8 +8,8 @@ order: 3
 
 [magicavoxel]: https://ephtracy.github.io/
 
-[MagicaVoxel][magicavoxel] is a free and user-friendly tool that lets you build
-3D scenes and models using *voxels* (i.e., blocks). This makes modeling super
+[MagicaVoxel][magicavoxel] is a free and user-friendly tool for building
+3D scenes and models using *voxels* (i.e., blocks). MagicaVoxel makes modeling super
 easy, similar to building in Minecraft.
 
 ![magicavoxel](https://i.imgur.com/XYYXjIn.jpg)
@@ -19,7 +19,7 @@ easy, similar to building in Minecraft.
 ## Installation
 
 MagicaVoxel works on Windows and OS X. On the [MagicaVoxel
-homepage][magicavoxel], click the *Download* button and install it:
+homepage][magicavoxel], click the *Download* button and install the application:
 
 ![magicavoxel install](https://i.imgur.com/GmgdyHs.jpg)
 
@@ -30,22 +30,20 @@ from the App Store in the *Security & Privacy* settings:
 
 ## Tutorial
 
-A good basic introduction to using MagicaVoxel is to watch the official
-tutorial video as the author quickly runs through a lot of the features in 10
-minutes:
+The official tutorial video serves as a good introduction to using MagicaVoxel.
+The author quickly runs through many basic features in 10 minutes:
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/PPu7SJ1_bwc" frameborder="0" allowfullscreen></iframe>
 
-The best way to learn is to click around, play with the UI, and pay attention
-to the tooltips on hover. We were able to create a scene in less than 20
-minutes without having ever used the program nor watched any tutorials. But
-given the absence of official tutorials, we will provide guidance on navigating
-the UI.
+The best way to learn the MagicaVoxel UI is to click around, play with application controls,
+and pay attention to the tooltips on hover. Many developers are able to create their first
+scene within a half hour!  Given the absence of official tutorials, however,
+the following serves as a quick guide to navigating the UI.
 
 ### Viewport
 
-In the center is the viewport where we build our model. We can change the
-viewport with our mouse/trackpad or the `wasd` keys:
+The center panel hosts the viewport where we build our model. We can change the
+viewport our mouse/trackpad or the `wasd` keys:
 
 - **Pan**: Hold `<space> + <right-click>` and move the mouse or trackpad, or
   hold `<space>` and press one of the `<a>` or `<d>` keys.
@@ -56,12 +54,13 @@ viewport with our mouse/trackpad or the `wasd` keys:
 
 ![magicavoxel viewport](http://imgur.com/vq34Mkk.jpg)
 
-On the top right, we can change the bounding dimensions of our model. Note, in
-MagicaVoxel, the XY plane is horizontal, and the Z axis points up.
+The top right corner of the viewport allows you to change the bounding dimensions
+of the model. Note that in MagicaVoxel, the XY plane is horizontal and the Z
+axis points up.
 
 ### Color Palette
 
-On the left panel is the color palette. You can use the default palettes,
+The left panel is the color palette. You can use the default palettes,
 modify them, create your own, and save them. The palette is tied to the scene
 so if you change from the palette a color that a voxel is using, the voxel will
 update its color. At the bottom, there are tools to define your own color with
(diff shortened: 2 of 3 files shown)


docs/guides/build-with-magicavoxel.md
```diff
@@ -30,22 +30,20 @@ from the App Store in the *Security & Privacy* settings:
...
+The author quickly runs through many basic features in 10 minutes:
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/PPu7SJ1_bwc" frameborder="0" allowfullscreen></iframe>
 
-The best way to learn is to click around, play with the UI, and pay attention
-to the tooltips on hover. We were able to create a scene in less than 20
-minutes without having ever used the program nor watched any tutorials. But
-given the absence of official tutorials, we will provide guidance on navigating
-the UI.
+The best way to navigate the MagicaVoxel UI is to click around, play with application controls,
```
ngokevin: This paragraph was addressing the best way to learn, not the best way to navigate.

docs/guides/build-with-magicavoxel.md
```diff
@@ -30,22 +30,20 @@ from the App Store in the *Security & Privacy* settings:
...
+and pay attention to the tooltips on hover. Many developers are able to create their first
+scene within minutes!  Given the absence of official tutorials, however,
```
ngokevin: `Within minutes` is a bit too hype. I think it's a bit more honest just to say "We were able to create our first scene in 20 minutes" because I don't have anything to back up people creating scenes in minutes.

docs/guides/build-with-magicavoxel.md
```diff
@@ -56,12 +54,12 @@ viewport with our mouse/trackpad or the `wasd` keys:
 
 ![magicavoxel viewport](http://imgur.com/vq34Mkk.jpg)
 
-On the top right, we can change the bounding dimensions of our model. Note, in
-MagicaVoxel, the XY plane is horizontal, and the Z axis points up.
+The top right panell allows you to change the bounding dimensions of the model.
```
ngokevin: typo

ngokevin: It's more like the top right of the viewport, it's not its own panel

docs/guides/build-with-magicavoxel.md
```diff
@@ -81,11 +79,11 @@ There are six brushes:
 
 - **V (Voxel)**: Works on individual or discrete groups of voxels.
 - **F (Face)**: Works on groups of adjacent voxels that are the same shape or color.
-- **B (Box)**: Works on a 2-dimensional box of voxels where we control the area
+- **B (Box)**: Works on a 2-dimensional box of voxels where you control the area
```
ngokevin: We use `we` everywhere in the documentation. It's more personal

docs/guides/index.md
```diff
@@ -10,14 +10,14 @@ section_order: 2
...
-building blocks.
+Let's build a scene using A-Frame's basic [primitive][primitives] HTML
+building blocks!
 
 <!--toc-->
 
 ## Adding a Box
 
-A *Hello World* scene is a scene with a box:
+This sample *Hello World* scene starts with a box:
```
ngokevin: Maybe emit `sample`

docs/guides/index.md
```diff
@@ -93,10 +93,13 @@ back color to white:
...
 
-It is desirable to cache the texture and block the scene from rendering until
-the texture is loaded.  We can move the texture into the [asset management
-system][asset]. We define it as an `<img>` tag, give it an ID, and point to it
-using a selector:
+It is best to cache the texture and block the scene from rendering until
+the texture has loaded.  To do so we must move the texture into the [asset management
+system][asset] by:
+
+- defining the asset as an `<img>` tag
```
ngokevin: capitalize list items, might not have to mention "tag"

docs/guides/index.md
```diff
@@ -93,10 +93,13 @@ back color to white:
...
+- giving it an ID (`texture`)
```
ngokevin: `id="texture"`

docs/guides/index.md
```diff
@@ -93,10 +93,13 @@ back color to white:
...
+- referring to the asset using an ID selector (`#texture`):
```
ngokevin: extra colon

ngokevin: "using the ID (in selector form)"

docs/guides/index.md
```diff
@@ -260,7 +262,7 @@ created by creating a sphere that wraps the scene:
 
 [next]: ./build-with-components.md
 
-And that is a very basic scene that places static objects in a 3D space using
-HTML. A good VR experience requires rich interaction and dynamic behavior. With
-the help of using and writing A-Frame components, we can [build a more advanced
+That is all it takes to create a very basic scene which places static objects in a 3D space using
+HTML! A good VR experience requires rich interaction and dynamic behavior. With
```
ngokevin: Maybe exclude the exclamation, don't really intend this hello world scene to be impressive :)

docs/guides/build-with-magicavoxel.md
```diff
@@ -30,22 +30,20 @@ from the App Store in the *Security & Privacy* settings:
...
+The best way to navigate the MagicaVoxel UI is to click around, play with application controls,
+and pay attention to the tooltips on hover. Many developers are able to create their first
+scene within minutes!  Given the absence of official tutorials, however,
+the following serves as a quick guide to navigating the UI.
 
 ### Viewport
 
-In the center is the viewport where we build our model. We can change the
-viewport with our mouse/trackpad or the `wasd` keys:
+The center panel hosts the viewport where the model is built. The
```
ngokevin: the viewport isn't really a panel

ngokevin: "we build our model" is more active voice than "where the model is built"

docs/guides/build-with-magicavoxel.md
```diff
@@ -30,22 +30,20 @@ from the App Store in the *Security & Privacy* settings:
...
+viewport can be manipulated with the mouse/trackpad or the `wasd` keys:
```
ngokevin: This is also a bit more passive than the previous `we` form

docs/guides/index.md
```diff
@@ -138,11 +141,11 @@ some motion into the scene:
 To interact with the box via clicking or gazing, we can use a cursor which we
 place as a child of the camera such that it is fixed to the screen. When we
 don't define a camera, the scene will inject a default camera, but in this case
-to add a cursor, we will need to define one.
+to add a cursor, we will need to explicitly define a camera.
 
-Then what we can tell the animation only to start when the cursor clicks the
-box event, using the animation's `begin` attribute. The cursor will emit the
-`click` event on the box, and the animation will listen to it:
+Next we configure the animation to start when the cursor clicks the
```
ngokevin: seems we should use prose when we can, `tell the animation` vs `configure the animation`

docs/guides/build-with-components.md
```diff
@@ -10,7 +10,7 @@ order: 2
 
 ![360&deg; Image Viewer](/images/docs/360-image-viewer.png)
 
-Let's go through an example building a scene using an
+The following is an example of building a scene using an
```
ngokevin: The former was a bit more personal

docs/guides/build-with-magicavoxel.md
darkwing: I wanted to avoid putting a number to it; puts a standard in people's head about how fast they should learn.  Maybe "within a few hours" would be better though?  Not sure.

ngokevin: Hm, maybe "under half an hour" then? I'm confident that it's that easy and people won't have any troubles. I think the reason I mention the time is to emphasize how easy it is. "within a few hours" makes it sound hard and people will be discouraged.

darkwing:
I'll revisit we vs. you vs. passive once I get more sleep.  The guides use both "we" and "you" at the moment, thus their voice jumps about.

ngokevin:
Thanks, if you catch any `you`s in a guide, we could make it consistent.

docs/guides/index.md
darkwing: The exclamation is finishing off the "That's all it takes" part; we're meaning to hype people up! :)

docs/guides/build-with-magicavoxel.md
darkwing: I see it as a panel, but can't think of a better word; regardless of what word we choose, it's better than "in the middle is"

ngokevin: there's a picture with a giant red circle highlighting the viewport. i think that's fine? for 3D programs, the viewport is separate from panels

docs/guides/index.md
ngokevin: not sure if hello world is worthy of hype. i sort of preferred the intention before, "that was a basic scene, now we are moving on to do cooler things" rather than "wasn't that amazing?"

darkwing: I disagree, I think we should show "!" as encouragement, but I've updated to your punctuation.

ngokevin: thanks. i think we should use emphasizers wisely, else they lose their value and it ends up seeming spammy.

darkwing:
This should be good for review @ngokevin

ngokevin:
Still a few "you"s instead of "we"s

PR #1853, opened 2016-08-25
unit test look controls

ngokevin:
**Description:**

Add unit tests for look-controls coverage. They are more like integration tests because they have no insight into the internals of look-controls. I test on the default camera such that if we ever change how look-controls works or to a different type of component, the tests should still cover everything.

 I have no idea why I need a `process.nextTick` in the tests after calling `sceneEl.tick()` and before calling `getAttribute`. I dug into it and could not find out where the asynchrony was coming from.

**Changes proposed:**
- clean up/comment/docstring look controls
  - don't conflate the component callbacks with methods (i.e., don't call `update` from `tick`, separate out)
- caught and fixed a bug where touch drag would throw `this.sceneEl is undefined`
- add unit tests for hmd (by stubbing VRControls), mouse drag, and touch

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -1,6 +1,5 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
-var isMobile = require('../utils/').device.isMobile();
 var bind = require('../utils/bind');
 
 var radToDeg = THREE.Math.radToDeg;
@@ -185,7 +184,7 @@ module.exports.Component = registerComponent('look-controls', {
       hmdQuaternion = this.calculateHMDQuaternion();
       hmdEuler.setFromQuaternion(hmdQuaternion, 'YXZ');
 
-      if (isMobile) {
+      if (el.sceneEl.isMobile) {
         // In mobile, allow camera rotation with touch events and sensors.
         rotation = {
           x: radToDeg(hmdEuler.x) + radToDeg(pitchObject.rotation.x),
@@ -229,17 +228,12 @@ module.exports.Component = registerComponent('look-controls', {
    * @member previousRotationY
    */
   calculateDeltaRotation: (function () {
-    var previousRotationX;
-    var previousRotationY;
+    var previousRotationX = 0;
+    var previousRotationY = 0;
     return function () {
       var currentRotationX = radToDeg(this.pitchObject.rotation.x);
       var currentRotationY = radToDeg(this.yawObject.rotation.y);
-      var deltaRotation;
-
-      previousRotationX = previousRotationX || currentRotationX;
-      previousRotationY = previousRotationY || currentRotationY;
-
-      deltaRotation = {
+      var deltaRotation = {
         x: currentRotationX - previousRotationX,
         y: currentRotationY - previousRotationY
       };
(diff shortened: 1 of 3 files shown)


tests/core/controls.test.js
```diff
@@ -0,0 +1,279 @@
...
+
+  test('does not rotate camera when dragging in VR with headset', function (done) {
+    var el = this.el;
+    this.dolly.quaternion.setFromEuler(new THREE.Euler(0, PI, 0));
+    el.sceneEl.addState('vr-mode');
+    el.sceneEl.canvas.dispatchEvent(new Event('mousedown'));
+
+    process.nextTick(function afterMousedown () {
+      var mouseMoveEvent = new Event('mousemove');
+      mouseMoveEvent.movementX = 1000;
```
darkwing: Super lame nit but you use `1000` quite a bit; should we use a var?

ngokevin: it's just a dummy number, not really meant to represent anything

src/components/look-controls.js
```diff
@@ -55,136 +68,177 @@ module.exports.Component = registerComponent('look-controls', {
...
-    var previousRotationX;
-    var previousRotationY;
+    var previousRotationX = 0;
+    var previousRotationY = 0;
     return function () {
       var currentRotationX = radToDeg(this.pitchObject.rotation.x);
       var currentRotationY = radToDeg(this.yawObject.rotation.y);
-      var deltaRotation;
-      previousRotationX = previousRotationX || currentRotationX;
-      previousRotationY = previousRotationY || currentRotationY;
```
dmarcos: I believe this code is incorrect. The orientation of the headset can start at any angle. The initial orientation won't be zero so the first calculated delta won't be
correct. If you start with your head turned 90 degrees and you move to 120 you want to apply a delta of 30 (120 of current rotation - 90 of inital one) Your code will return 120 instead.

ngokevin: Actually, `calculateDeltaRotation` is not used for headsets. It is only used for mouse drag. I changed it to `0` because I caught in the unit tests that because it ended up doing:

```js
var deltaRotation = {
  x: currentRotationX - (previousRotationX || currentRotationX),
  y: currentRotationY - (previousRotationY || currentRotationY)
};
```

On the first tick then since `previousRotation` was undefined, we ended up:

```js
var deltaRotation = {
  x: currentRotationX - currentRotationX,
  y: currentRotationY - currentRotationY
};
```

And achieving no rotation

ngokevin: I'll add tests with the camera in a pre-rotated position to make sure. I think `previousRotation` is previous rotation relative to the delta?

ngokevin: Added a couple tests:

- HMD rotation is absolute. It replaces rotation without regard for deltas.
- Mouse drag rotation is relative. It applies delta rotation relative to previous rotation

ngokevin:
rebased

ngokevin:
Oops...need to fix the rebase.

ngokevin:
r?

cvan:
@dmarcos @donmccurdy does this look good?

src/components/look-controls.js
```diff
@@ -56,99 +69,132 @@ module.exports.Component = registerComponent('look-controls', {
...
       var rotation;
+      var sceneEl = el.sceneEl;
+      var yawObject = this.yawObject;
+
+      hmdQuaternion = this.calculateHMDQuaternion();
       hmdEuler.setFromQuaternion(hmdQuaternion, 'YXZ');
-      if (isMobile) {
-        // In mobile we allow camera rotation with touch events and sensors
+
+      if (el.sceneEl.isMobile) {
```
donmccurdy: What's this change, and should external components be doing this?

ngokevin: The scene exposes this property so no need to recalculate.

ngokevin:
Will re-open the PR

Issue #1535, opened 2016-06-03
Camera position is set each frame even with no movement


tbalouet:
In the "wasd-controls" component, the update function retrieves a movementVector, and update the camera in consequence. But it doesn't check if the movement vector is null, in which case the camera position shouldn't be affected, right?

It's starting line 58500 in dist/aframe.js

- A-Frame Version: tot
- Platform/Device: google chrome 50

ngokevin:
https://github.com/aframevr/aframe/pull/1904

PR #1854, opened 2016-08-25
Fix 1535: Don't set camera position for Null vector

darkwing:
Took a stab at #1535 ; reuses existing function, avoids unnecessary processing.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -2,6 +2,7 @@ var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 var isMobile = require('../utils/').isMobile();
 var bind = require('../utils/bind');
+var isNullVector = require('../utils/coordinates').isNullVector;
 
 // To avoid recalculation at every mouse movement tick
 var PI_2 = Math.PI / 2;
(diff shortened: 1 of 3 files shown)


src/utils/vectors.js
```diff
@@ -0,0 +1,4 @@
+
+module.exports.isNullVector = function (vector) {
```
ngokevin: You could put this in `utils/coordinates.js` where all the other vector stuff is.

ngokevin:
brave enough for a test? i spent a lot of today testing look-controls, would be great to have a start on wasd-controls. i'd mimic https://github.com/aframevr/aframe/pull/1853/files#diff-387522ff670c30773ff46795b9f7dc84R7 where tests are in `core/controls.test.js` and we integration test using the default camera.

ngokevin:
could use a small test

darkwing:
I plan to do so early this week.

darkwing:
Starting these tests tonight.

darkwing:
I was writing a test for this today, it kept failing, so upon further review I don't think this is needed due to @ngokevin's recent PR and this line: https://github.com/aframevr/aframe/pull/1904/files#diff-6231789e7e7fac0fbea47834d09b3bdcR56

Let me know if I'm wrong but I think this can be closed.

ngokevin:
Cool. Thanks. I was just writing the tests to accompany the look-controls and guess I accidentally did this too.

PR #1857, opened 2016-08-25
treat gearvr as mobile to apply userHeight (fixes #1856)

### LINKED ISSUE #1856: GearVR camera is the wrong height
**GearVR camera height incorrect:**
- A-Frame Version: 3.0
- Platform/Device: GearVR
## To reproduce
- Install browser inside GearVR Oculus homepage
- Turn on WebVR (using `internet://webvr-enable`)
- goto https://aframe.io/
- click to enter VR.

You are at height 0, the floor, not 1.8, like Carboard and 2D screen versions.

I see in the documentation: https://aframe.io/docs/0.3.0/components/camera.html

_"When not in VR mode, userHeight translates the camera up to simulate human height. The injected camera has this set to 1.8 (meters). When entering VR, this height offset is removed such that the absolute position return from the VR headset is used. This is very convenient for experiences that work both in and out of VR, as well as making experiences look decent from the monitor as opposed to clipping the ground if the headset was resting on the ground."_

So I suspect that we need a special case for fixed-position headsets, like the GearVR, to use the 1.8.

ngokevin:


dmarcos:
gear vr is not really mobile. It could have undesired side effects. When we say mobile we refer to mobile devices without a native webvr API. To determine if user height has to be applied it might be better to check if the headset has positional tracking

tests/utils/mobile.test.js
```diff
@@ -0,0 +1,14 @@
+/* global assert, suite, test */
+var utils = require('utils/');
+
+suite('utils.mobile', function () {
+  suite('isMobile', function () {
+    test('includes GearVR', function () {
+      var originalUa = navigator.userAgent;
+      assert.notOk(utils.isMobile());
+      navigator.__defineGetter__('userAgent', function () { return 'SamsungBrowser'; });
```
darkwing: `__defineGetter__` is deprecated.  How about this:

```js
Object.defineProperty(navigator, 'userAgent', {
    get: function() {
        return 'SamsungBrowser';
    }
});
```

ngokevin:
Redoing...

Issue #1846, opened 2016-08-24
a-videosphere no working on iOS


fuleinist:
a-videosphere is not working in mobile safari on iOS device.

A-Frame Version: 0.3.0
Platform/Device: iPhone 6 Safari iOS ver 9.3.4

Example page shows a full black screen, finger tab will not initiate video playback.

Compared side by side with with a-video. followed steps in https://aframe.io/docs/0.3.0/primitives/a-videosphere.html#caveats. a-video works perfectly fine while a-videosphere is not on iOS

```
<a-assets>
      <video webkit-playsinline autoplay loop crossorigin  id="video"  src="city.mp4"></video>
      <video webkit-playsinline autoplay loop crossorigin id="video1"  src="john.mp4"></video>
</a-assets>

<a-videosphere src="#video" rotation="0 180 0" ></a-videosphere>
<a-video src="#video1" width="16" height="9" position="20 0 -10"></a-video>
```

ngokevin:
I found some stuff here: https://github.com/bfred-it/iphone-inline-video

I wonder if you need to add `muted` to get it to play without user interaction?

fuleinist:
@ngokevin added iphone-inline-video as well as muted inline. still has the same problem.

I double it's related to user interaction on video tag since a-video works fine once I  pined the webpage to the iOS homescreen.

ngokevin:
Hm, they should both follow down the same code path. plane/sphere shouldn't make a difference.

What if it's just one video? And does the city.mp4 video play on desktop? What if we try to play `#video1` on the videosphere?

fuleinist:
wow, videosphere actually worked after I reversed the video content....

the city video is the same 360 video on aframe.io demo page. The other one is just a random mp4 video. Now it's confirmed to be a problem with the video file.

After some google search I believe now this is the root: 

One important thing to note is that "iDevices" doesn't support all the varieties of mp4. Here is a small list about MPEG-4 and H.264 supported varieties.
According to Apple official website, iPad or iPhone 4S is only compatible with H.264 or MPEG-4 video format with the following specifications:
If it is MPEG-4 video, it should meet: up to **2.5 Mbps**, 640 by 480 pixels, 30 frames per second, Simple Profile with AAC-LC audio up to 160 Kbps per channel, 48kHz, stereo audio in .m4v, .mp4, and .mov file formats;

[...]

PR #1858, opened 2016-08-25
Low Rate Video File to be uploaded to cdn

fuleinist:
**Description:**
For issue #1846 task 
**Changes proposed:**
Please download the video via https://drive.google.com/file/d/0B4IIadesQuavN0puUjNqLV84bG8/view?usp=sharing

### DIFF
--- a/examples/boilerplate/360-video/index.html
+++ b/examples/boilerplate/360-video/index.html
@@ -5,6 +5,15 @@
     <title>360 Video</title>
     <meta name="description" content="360 Video — A-Frame">
     <script src="../../../dist/aframe.js"></script>
+    <script "text/javascript">
+    var cityvideo = document.getElementById("video");
+		if( /iPhone|iPad|iPod/i.test(navigator.userAgent) ) {
+			//src to be changed to the low bitrate one.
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks, should we just use that video on desktop too?

fuleinist:
Hi Kevin,

You can but the resolution is too low on big screen.... And this issue only
happen on iOS so far

Best Regards,

ngokevin:
I see, thanks. I suppose the proper API might be something like:

```
<a-assets>
  <video id="city"></video>
  <video id="cityMobile></video>
</a-assets>

<a-videosphere degrading-src="desktopSrc: #city; mobileSrc: #cityMobile"></a-videosphere>
```

where degrading-src is a component...although it's a lot to bend backwards just for iOS. I wonder if we should just document the limitations and move on

fuleinist:
I'd suggest to wait and see if mobile safari's html5 player will be upgraded to resolve this after iOS10's release

ngokevin:
Thanks, will file an issue to have that example swap to mobile using a component and/or just wait for iOS 10.

Utopiah:
Having encountered similar problems I usually serve 2 different video, 1 for desktop in high res and 1 for mobile (not just iOS) that is usually significantly lower.

PR #1861, opened 2016-08-27
Use SSL with budo

darkwing:
Do we want to create a separate command for SSL to avoid the annoying SSL warnings the first time we hit the dev site?


ngokevin:
There's a currently a PR open to swap to BrowserSync. Will close in favor of that: https://github.com/aframevr/aframe/pull/1671

PR #1863, opened 2016-08-27
Fix #830: Allow stats to be disabled via querystring

### LINKED ISSUE #830: Allow `stats` to be disabled from postMessage/queryString
Useful for when embedding scenes as `<iframe>`s. But there are cases when iframe'd I still want the stats visible.

darkwing:


src/components/scene/stats.js
```diff
@@ -1,6 +1,7 @@
 var registerComponent = require('../../core/component').registerComponent;
 var RStats = require('../../../vendor/rStats');
-var bind = require('../../utils/bind');
+var utils = require('../../utils');
+var bind = utils.bind;
```
ngokevin: can separate requires from variable setup

```
var this = require('this');

var that = this.that
```

ngokevin:
needs unit tests

ngokevin:
thanks!

cvan:
FWIW, I opened a PR about `postMessage`-ing data in #1487. probably not necessary any time soon, but I can see the query-string parameters getting messy over time.

ngokevin:
query string params for stats would've been useful in several cases when i wanted to showcase scenes that had stats on.

darkwing:
Test added.

dmarcos:
Where are `queryStrings` documented? I agree with @cvan. It is a slippery slope to sprinkle queryParameter based switches all over the place. Is there a way to make it component agnostic and self contained by having a `queryParameter` component?

ngokevin:
IMO, the behavior should be contained within the component. I don't think it's a slippery slope, having query params here isn't going to encourage us to have query params in other components. We have some precedent on not trying to anticipate. To date, we only have two: `stats` + `vr-mode-ui`. We'd document the behavior from within the relevant component.

Though alternatively, you could have an agnostic component that detaches/attaches components based on query parameters. Something like `queryparam-detach="component: stats; queryParam: stats"`. I don't like this as much because it's more API, more default components to add, and it's indirect.

cvan:
What about going with a strictly `postMessage` API? Seems lightweight and changing the settings wouldn't cause a page refresh. Or if we don't want to go there yet, use hashes (since those won't reload the page upon changing, but you can listen for `hashchange`).

ngokevin:
postMessage is for I-Frames and is more complex. My use case is just when someone publishes a scene with stats on, I want to share it with stats off. We don't need dynamic change of the settings here, don't think we're gonna be toggling the stats on and off across origins. And also web pages should only have one hash, giving it functionality for stats makes things odd.

cvan:
> postMessage is for I-Frames and is more complex.

`postMessage` is not necessarily for iframes. can be used for these following use cases:

* a document can `postMessage` to itself
* a scene can `postMessage` to another scene
* a Web Worker can `postMessage` to scene(s)
* a Service Worker can `postMessage` to scene(s)

> And also web pages should only have one hash, giving it functionality for stats makes things odd.

Do you think that will hold true? Already the author of the [`aframe-href-component`](https://github.com/gasolin/aframe-href-component/blob/e1e49f3/basic/anchor.html#L16) is using a hash to jump to an entity in the scene. How would you load https://gasolin.github.io/aframe-href-component/basic/anchor.html#blue-cube and disable stats if we're assuming a "single hash"?

I don't have a perfect solution in mind, but I think it'd be more future-proof, extensible, and user-friendly to introduce a `postMessage`-based solution and introduce a method on the `AFRAME` global, such as `AFRAME.stats.disable()` - which would disable stats and store a value in `localStorage` to disable stats upon subsequent page loads.

(Related: For the [eleVR Web Player](https://github.com/hawksley/eleVR-Web-Player#ability-to-load-external-video-urls-via-query-string-hash-postmessage), I introduced a few (albeit complicated but versatile options) for configuring the player without having to write a bunch of boilerplate code.)

ngokevin:
RE: postMessage use cases. None of those have anything to do with stats. I mentioned I-Frames because I assumed that was your use case.

> How would you load https://gasolin.github.io/aframe-href-component/basic/anchor.html#blue-cube and disable stats if we're assuming a "single hash"?

I wouldn't, which is why I'm in favor of using queryparams and not hashes as you proposed `use hashes`.

> I don't have a perfect solution in mind, but I think it'd be more future-proof, extensible, and user-friendly to introduce a postMessage-based solution and introduce a method on the AFRAME global, such as AFRAME.stats.disable() - which would disable stats and store a value in localStorage to disable stats upon subsequent page loads.

Don't see why all this over a queryparam. I think we're missing why we're introducing a flag to disable stats. If a developer wanted to disable stats on their own, they would just remove it, they wouldn't call a global method. The use case is to disable stats when the developer has set it and we want to showcase it without.

cvan:
> Don't see why all this over a queryparam.

what do you mean by `all this`? as I'd expect in any other PR, I think we can agree it's fair to question additions to the codebase, regardless of the seeming size or scope of the change.

> I think we're missing why we're introducing a flag to disable stats. If a developer wanted to disable stats on their own, they would just remove it, they wouldn't call a global method. The use case is to disable stats when the developer has set it and we want to showcase it without.

apologies, either the past conversation I think got truncated here, or I was just getting confused between PRs (as this is over three weeks old now).

> If a developer wanted to disable stats on their own, they would just remove it, they wouldn't call a global method. The use case is to disable stats when the developer has set it and we want to showcase it without.

what do you mean by "showcase"? do you mean fire up the URL of an A-Frame scene, manually/automatically take a video/GIF/screenshot? or do  you mean link to a URL from our "showcase" (i.e., the A-Frame Blog or from Twitter)?

in any case, I would imagine that we would want to normalise all the presentation of the page (meaning disable stats UI, debug mode, cursor style of `<body>`, hide the VR goggles/any other button UI). if that's what we're going for, I would recommend instead of having a one-off for `stats` (which as @dmarcos says could be a slippery slope), perhaps what we want instead is a `presentation ` mode? thoughts?

ngokevin:
> what do you mean by all this? 

I was questioning the complexity of the solution (which actually doesn't solve the problem).

> what do you mean by "showcase"?

Both

> in any case, I would imagine that we would want to normalize all the presentation of the page...could be a slippery slope

The only thing we currently desire is the ability to disable stats. Debug mode has no visual effect, don't had a need for cursor style customization, don't have a need for hiding the goggles. I don't imagine a slippery slope, adding a query parameter here is not going to encourage other uses, it's just the simplest solution that effectively solves the problem so we can move on.

src/components/scene/stats.js
```diff
@@ -16,6 +17,9 @@ module.exports.Component = registerComponent('stats', {
 
   init: function () {
     var scene = this.el;
+
+    if (utils.getUrlParameter('stats') === 'false') { return; }
```
cvan: I would expect `?stats=true` to enable stats

ngokevin: We can't because this code is only run when you actually have stats attached. And there's not really a use case for needed `?stats=true`, but if we wanted to see them, we can attach stats via devtools.

cvan: > And there's not really a use case for needed `?stats=true`, but if we wanted to see them, we can attach stats via devtools.

Why would there not be a use case for them? I can imagine someone writing a script that fires up a handful of scenes to quickly look at the stats.

Just for the sake of expectations, if I see a URL with `?stats=false`, I'm going to assume that `?stats=true` turns stats on. Why would not enable both?

> We can't because this code is only run when you actually have stats attached.

Yeah, but wouldn't you just need to call this when `?stats=true`:

```js
this.el.setAttribute('stats', '');
```

ngokevin: > Why would there not be a use case for them? I can imagine someone writing a script that fires up a handful of scenes to quickly look at the stats.

That's not really a use case. Because that's not an actual flow I've needed outside of A-Frame nor do I imagine anyone having a handful of scenes that they want to do a stats pass over. Stats are usually used during a deep inspection when you dig down into a single scene, rather than taking a brief look at many. Even then, a console command or coding into the scene is quick enough.

> Just for the sake of expectations, if I see a URL with ?stats=false, I'm going to assume that ?stats=true turns stats on. Why would not enable both?

There's a single very narrow, minor, and non-imagined use case (me sharing other people's scenes that have stats on) that this PR was addressing, and we have a concise one-line solution for it. If someone was clicking a scene on the blog or Twitter, they aren't going to try messing with the URL or wanting to look at the stats.

darkwing:
Any more thoughts on his?  Just want to keep it moving....

ngokevin:
I'd r+ it, but making sure everyone's OK with it. Simple use case, simple solution, should be cut/dry.

cvan:
go ahead

PR #1864, opened 2016-08-27
Fix #1862 - Ensure inspector is destroyed upon test completion

### LINKED ISSUE #1862: Inspector test making sound test fail
I believe the inspector test is causing transient CI failures for the sound test:

"does not call sound pause if not playing"

I don't believe the inspector is ever removed from the page once it has been injected.

darkwing:
Appears to fix transient test failures.

### DIFF
--- a/tests/components/scene/inspector.test.js
+++ b/tests/components/scene/inspector.test.js
@@ -1,8 +1,12 @@
 /* global assert, setup, suite, teardown, test */
 
 suite('inspector', function () {
+  var SCRIPT_SELECTOR = 'script[data-name=aframe-inspector]';
+
   setup(function (done) {
     var el = this.sceneEl = document.createElement('a-scene');
+    // Set a fake URL so the inspector doesn't interfere with other tests
+    el.setAttribute('inspector', 'url: fake.js');
     document.body.appendChild(el);
 
     el.addEventListener('loaded', function () { done(); });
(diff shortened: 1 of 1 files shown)


darkwing:
Removing the element in `inspector.test.js` doesn't remove it completely.

tests/__init.test.js
```diff
@@ -28,7 +28,7 @@ setup(function () {
 
 teardown(function () {
   // Clean up any attached elements.
-  ['canvas', 'a-assets', 'a-scene'].forEach(function (tagName) {
+  ['canvas', 'a-assets', 'a-scene', '#aframe-inspector'].forEach(function (tagName) {
```
ngokevin: I think we can isolate this tear down to the inspector test suite

darkwing: Tried that...somehow it doesn't leave

ngokevin: Even just doing `#aframe-inspector` instead of `div`?

ngokevin: OK, I would just set a `url` property on the inspector component in the test to a dummy script URL, rather than loading from the real URL.

darkwing:
Spoofed URL -- good call!  Updated!

ngokevin:
Seems like we could also assert the URL now for good measure: https://github.com/aframevr/aframe/pull/1864/files#diff-4f03fc30c4b0a8c6c5d3703fc9d6e58aR29

darkwing:
IMO a URL check is essentially a check that component settings are passed along properly, which isn't specific to this change.

ngokevin:
The `<script>` tag would still be there and not destroyed right? We can tear it down in the test teardown.

darkwing:
Updated to axe `script` tag.

PR #1874, opened 2016-08-30
Fix #1630: Add white background highlighting for rStats alarm red text

### LINKED ISSUE #1630: Make stats colour more readable
When debugging the frame rate on an iOS device, if the frame rate is less than 30 the text is red against the pink background making it difficult to read. Can the background or the text colour be more contrasting?

- A-Frame Version: 0.2.0
- Platform/Device: iPhone 6s iOS 9.3

framerate

kevinzwh:
**Description:**
rStats currently displays red text to alarm the user of stats that are less than optimal. This PR attempts to address #1630 ( red text behind the current gray background is hard to read ).

screen shot 2016-08-29 at 9 15 06 pm

**Changes proposed:**
- When a counter/label enters the alarm state in rStat, a white background highlights the red text to make it more visible and readable.

rstatcolorchange


dmarcos:
Thanks!

PR #1875, opened 2016-08-31
Tracked controls still update the pose if a mesh is not present

dmarcos:
The tick method was returning if a mesh is not defined. `tracked-controls` should work with any entity with and without a mesh defined.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -33,8 +33,8 @@ module.exports.Component = registerComponent('tracked-controls', {
 
   tick: function (time, delta) {
     var mesh = this.el.getObject3D('mesh');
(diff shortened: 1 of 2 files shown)


src/components/tracked-controls.js
```diff
@@ -33,8 +33,8 @@ module.exports.Component = registerComponent('tracked-controls', {
 
   tick: function (time, delta) {
     var mesh = this.el.getObject3D('mesh');
-    if (!mesh) { return; }
-    if (mesh.update) { mesh.update(delta / 1000); }
+    // Update mesh animations;
```
ngokevin: semicolon

ngokevin:
could add a test

PR #1880, opened 2016-09-01
Replace alarm white background with improved white text shadow in rStats.js

kevinzwh:
**Description:**
Currently, rStats displays a large white background behind counters under the alarm state. This was added for visibility in #1874, however it's aggressive compared to the rest of the rStat dashboard.

rstatcolorchange

**Changes proposed:**
- When a counter/label enters the alarm state, a white text shadow highlights the red text to make it more readible and visible without being as intrusive as a wide white background. The text shadow makes use of 9 levels of shadow to form a composite & dense coloring around the red text.

- This solution was discussed in #1630.

rstatcolorchange7

### DIFF
--- a/vendor/rStats.js
+++ b/vendor/rStats.js
@@ -251,7 +251,7 @@ window.rStats = function rStats ( settings ) {
             var a = ( _def && ( ( _def.below && _value < _def.below ) || ( _def.over && _value > _def.over ) ) );
             _graph.draw( _value, a );
             _dom.style.color = a ? '#b70000' : '#ffffff';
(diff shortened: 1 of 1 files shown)


ngokevin:
Can you PR to rstats: https://github.com/spite/rstats to add a **class** to those under/over-threshold fields? That way we don't have to hack the JS file and apply through CSS

kevinzwh:
Sure, I can do that later tonight

kevinzwh:
Went ahead and opened a PR to rstats with a CSS class based solution here -> https://github.com/spite/rstats/pull/24

kevinzwh:
Revisiting this

PR #1883, opened 2016-09-01
make cursor mouseleave with multiple objects more reliable (fixes #1882)

### LINKED ISSUE #1882: Cursor events don't fire correctly when multiple objects intersect the path
In Aframe 0.3.0 the mouse events do not fire correctly when multiple objects intersect the raycaster/cursor. Unfortunately, this makes the cursor events unusable for any UI that has overlapping objects. This worked as expected in 0.2.0. 

Demo of the bug in 0.3.0: https://codepen.io/ryanbetts/pen/yaLVGL
Demo of it working in 0.2.0: https://codepen.io/ryanbetts/pen/bwGgoG

ryanbetts:
**Description:**
A fix to the cursor.js component to ensure that `mouseenter` and `mouseleave` events fire reliably when multiple objects intersect the cursor's raycaster. 
 
**Changes proposed:**
- created a new function called `leaveCurrentIntersection` to handle triggering leave events, resetting the fuse timer, etc.
- leaveCurrentIntersection is now triggered in the `onIntersection` function if a current intersection already exists
- leaveCurrentIntersection is triggered in `onIntersectionCleared` only if the event is fired on the currently intersected element
- added test cases to the Cursor test example

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -89,10 +89,7 @@ module.exports.Component = registerComponent('cursor', {
     // Set intersected entity if not already intersecting.
     if (this.intersectedEl === intersectedEl) { return; }
 
-    if (this.intersectedEl) {
-      this.leaveCurrentIntersection();
-    }
-
+    if (this.intersectedEl) { this.clearCurrentIntersection(); }
     this.intersectedEl = intersectedEl;
 
     // Hovering.
@@ -120,10 +117,10 @@ module.exports.Component = registerComponent('cursor', {
     // ignore if the event didn't occur on the current intersection
     if (intersectedEl !== this.intersectedEl) { return; }
 
-    this.leaveCurrentIntersection(intersectedEl);
+    this.clearCurrentIntersection(intersectedEl);
   },
 
-  leaveCurrentIntersection: function () {
+  clearCurrentIntersection: function () {
     var cursorEl = this.el;
 
     // No longer hovering (or fusing).
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -88,6 +88,11 @@ module.exports.Component = registerComponent('cursor', {
 
     // Set intersected entity if not already intersecting.
     if (this.intersectedEl === intersectedEl) { return; }
+
+    if (this.intersectedEl) {
+      this.leaveCurrentIntersection();
+    }
+
     this.intersectedEl = intersectedEl;
 
     // Hovering.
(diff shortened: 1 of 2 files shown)


src/components/cursor.js
```diff
@@ -88,6 +88,11 @@ module.exports.Component = registerComponent('cursor', {
 
     // Set intersected entity if not already intersecting.
     if (this.intersectedEl === intersectedEl) { return; }
+
+    if (this.intersectedEl) {
```
dmarcos: let's do one liner

dmarcos:
I think the bug that cause the regression is that the cursor is listening for the wrong event names `raycaster-intersection` and `raycaster-intersection-cleared`. Replacing those with `intersection` and `intersactioncleared` should fix the problem

https://github.com/aframevr/aframe/blob/master/src/components/cursor.js#L53

dmarcos:
Nevermind. Disregard. I was looking at 0.2.0 :)

ryanbetts:
I've added to the pull request to attach intersection information to all of the cursor event payloads to match the specs of this stackoverflow post:
http://stackoverflow.com/a/38667473/6781113

src/components/cursor.js
```diff
@@ -132,8 +143,10 @@ module.exports.Component = registerComponent('cursor', {
    */
   twoWayEmit: function (evtName) {
     var intersectedEl = this.intersectedEl;
-    this.el.emit(evtName, {intersectedEl: this.intersectedEl});
+    var cursorPayload = { intersectedEl: this.intersectedEl, intersection: this.intersection };
```
dmarcos: can we call it `evtDetail`?

src/components/cursor.js
```diff
@@ -132,8 +143,10 @@ module.exports.Component = registerComponent('cursor', {
...
+    var intersectedElPayload = { cursorEl: this.el, intersection: this.intersection };
```
dmarcos: Can we call it `intersectedElEvtDetail`

ngokevin:
There are a couple of failing tests

src/components/cursor.js
```diff
@@ -132,8 +143,10 @@ module.exports.Component = registerComponent('cursor', {
    */
   twoWayEmit: function (evtName) {
     var intersectedEl = this.intersectedEl;
```
ngokevin: I'd style it like:

```js
var cursorEl = this.el;
var intersectedEl = this.intersectedEl;
var intersection = this.intersection;

cursorEl.emit(evtName, {intersectedEl: intersectedEl, intersection: intersection});

if (!intersectedEl) { return; }
intersectedEl.emit(evtName, {cursorEl: el, intersection: intersection});
```

ngokevin:
addressed at:

- https://github.com/aframevr/aframe/commit/e4900e16ea9228af39d2a4fef6798393e79bd82a
- https://github.com/aframevr/aframe/commit/37c33d2cc88de369feff7027f3d1e24b305ef496

PR #1887, opened 2016-09-02
Refresh list of raycasted objects when a new entity is added to the s…

dmarcos:


src/components/raycaster.js
```diff
@@ -31,6 +31,15 @@ module.exports.Component = registerComponent('raycaster', {
     this.prevCheckTime = undefined;
     this.raycaster = new THREE.Raycaster();
     this.updateOriginDirection();
+    this.refreshObjects = this.refreshObjects.bind(this);
```
darkwing: To be consistent we probably want `bind(this.refreshObjects, this)`

src/components/raycaster.js
```diff
@@ -31,6 +31,15 @@ module.exports.Component = registerComponent('raycaster', {
...
+  },
+
+  play: function () {
+    this.el.sceneEl.addEventListener('child-attached', this.refreshObjects);
```
ngokevin: Rather than refresh all objects again, we can check if the child matches the selector, and append to array.

src/components/raycaster.js
```diff
@@ -31,6 +31,15 @@ module.exports.Component = registerComponent('raycaster', {
...
+  },
+
+  stop: function () {
```
ngokevin: pause?

src/components/raycaster.js
```diff
@@ -31,6 +31,15 @@ module.exports.Component = registerComponent('raycaster', {
...
+    this.el.sceneEl.removeEventListener('child-attached', this.refreshObjects);
```
ngokevin: we can cover this line to protect from mistakes like misnamed method ^

src/components/raycaster.js
dmarcos: We would have to run `el.match` and `el.querySelectorAll` on the child and remove one by one all the query matches when a child is detached. For simplicity sake I prefer to just `refreshObjects` on attach / detach and make it more granular later if needed.

ngokevin: I don't think we should compromise on performance. Imagine a shooting game where the gun is a raycaster aimed at monsters, and you are spawning lots of bullets. This would rebuild the entire raycaster target list on each bullet when we actually don't have to do anything. You don't have to worry about children because `child-attached` will be called on each child and the raycaster can handle recursion.

ngokevin:
Tagged needs-answers for https://github.com/aframevr/aframe/pull/1887#discussion_r77955718

src/components/raycaster.js
dmarcos: In a shooting game you will be using a pool of bullets not recreating and destroying entities.

dmarcos: We can optimize later. I need this patch for the a-painter and don't have the bandwidth right now to do something more elaborate.

dmarcos: We're actually working on a shooting demo. It this becomes a problem we're going to know soon.

src/core/a-entity.js
```diff
@@ -103,6 +103,7 @@ var proto = Object.create(ANode.prototype, {
       // Remove components.
       Object.keys(this.components).forEach(bind(this.removeComponent, this));
       this.parentEl.remove(this);
+      this.parentEl.emit('child-detached', { el: this });
```
ngokevin: no spaces

src/components/raycaster.js
ngokevin: The painter isn't much a factor since you're using a custom/dev build of A-Frame anyways. But want me to just do it after? It's not that much work (run a match and push, run a match and pop).

ngokevin:
Okay, let's just r+wc, I can add perf improvements quickly so we can move on.

PR #1903, opened 2016-09-13
[docs] Updating Fuse Timemout to new syntax in cursor docs

mininao:
**Description:**
It seems that you can no longer set the fuse timeout of a cursor using `timeout`. This looks only partially reflected in the docs.
**Changes proposed:**
- Change `timeout` to `fuseTimeout` in the cursor docs

Anyway, Thanks a lot for the great work that you've done on this project ! It's awesome :smiley:


ngokevin:
thanks!

Issue #1881, opened 2016-09-01
Dynamic src attribute broken in 0.3.0


vrimaginar:
We are using a dynamic source attribute to the video tag and that's not working anymore in version 0.3.0. The reason we do this is to stream different bitrate and or resolution video to client devices. The error is: TypeError: r is null, 	

...ossorigin")?e:(r=e.getAttribute("src"),r.indexOf("://")===-1?e:extractDomain(r)=...

I suppose it has something to do with the error handling for wrong formatted src attributes implemented a couple of days ago (#899).

Question: can we have the implmentation back from 0.2.0 where a src attribute was not mandated in the video tag? 

- A-Frame Version: 0.3.0
- Platform/Device: all
- Reproducible Code Snippet or Demo URL [highly encouraged]:

<a-scene>
     <a-assets>
         <video id="#video"/>
     </a-asset>
     <a-videospehere src="#video"/>
</a-scene>

       [javascript to determine device/os]
       video =  document.getElementById('video');
       source = document.createElement('source');
			
	source.setAttribute('src', src);
	video.appendChild(source);
	video.play();

The above is simpliflied code (there should be some code there???)

ngokevin:
If you use 0.3.0 stable, then #899 shouldn't apply. Are you using master or stable 0.3.0?

vrimaginar:
We use 0.3.0 stable and the difference with 0.2.0 is that there seems to be an added check if the src attribute exists and if not it throws the above error. In 0.2.0 there was no check for the src attribute in the video tag (assets --> video).

ngokevin:
I see. Will think about how to resolve that.

You can do a workaround by using entities.

`<a-entity geometry="primitive: plane" material>`

When you are ready, set the material src:

`el.setAttribute('material', 'src', 'url(myvideo.mp4)');`

vrimaginar:
I wonder if autoplay works on the workaround you propose. I will try and see if that works. And by the way, thank you (and the other on the team) for a great framework and the effort you put in it.

ngokevin:
This was introduced in https://github.com/aframevr/aframe/pull/1789

Need to update that code to not error if there's not a `src`

ngokevin:
Fixed thanks

PR #1909, opened 2016-09-14
Issue #1881 src not set throws error for dynamic src

Zeouterlimits:
**Description:**
Fixing [Issue #1881 ](https://github.com/aframevr/aframe/issues/1881#issuecomment-244658276), where a previous change caused a regression for usage of the dynamic src element.
Now handling when src is not present, where it will return null or an empty string - [https://developer.mozilla.org/en/docs/Web/API/Element/getAttribute](https://developer.mozilla.org/en/docs/Web/API/Element/getAttribute)

**Changes proposed:**
- If src is null, return the media element, like if it does not have the protocol set

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -156,8 +156,8 @@ function setCrossOrigin (mediaEl) {
 
   src = mediaEl.getAttribute('src');
 
-  // Does not have protocol.
-  if (src.indexOf('://') === -1) { return mediaEl; }
+  // Does not have src or src does not have protocol.
+  if (src === null || src.indexOf('://') === -1) { return mediaEl; }
 
   // Determine if cross origin is actually needed.
   if (extractDomain(src) === window.location.host) { return mediaEl; }
--- a/tests/core/a-assets.test.js
+++ b/tests/core/a-assets.test.js
@@ -107,6 +107,26 @@ suite('a-assets', function () {
       document.body.appendChild(scene);
     });
 
+    test('does not recreate media element if no src set', function (done) {
+      var el = this.el;
+      var scene = this.scene;
+      var img = document.createElement('img');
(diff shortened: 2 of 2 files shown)


src/core/a-assets.js
```diff
@@ -156,8 +156,8 @@ function setCrossOrigin (mediaEl) {
 
   src = mediaEl.getAttribute('src');
 
-  // Does not have protocol.
-  if (src.indexOf('://') === -1) { return mediaEl; }
+  // Does not have src or src does not have protocol.
+  if (src === null || src.indexOf('://') === -1) { return mediaEl; }
 
   // Determine if cross origin is actually needed.
   if (extractDomain(src) === window.location.host) { return mediaEl; }
```
ngokevin: I would add the check here so it will set get crossorigin applied, but not error.

Zeouterlimits: Like so?

```
function setCrossOrigin (mediaEl) {
  var newMediaEl;
  var src;

  // Already has crossorigin set.
  if (mediaEl.hasAttribute('crossorigin')) { return mediaEl; }

  src = mediaEl.getAttribute('src');

  if (src != null) {
    // Does not have protocol.
    if(src.indexOf('://') === -1) { return mediaEl; }

    // Determine if cross origin is actually needed.
    if (extractDomain(src) === window.location.host) { return mediaEl; })
  }

  warn('Cross-origin element was requested without `crossorigin` set. ' +
       'A-Frame will re-request the asset with `crossorigin` attribute set. src:', src);
  mediaEl.crossOrigin = 'anonymous';
  newMediaEl = mediaEl.cloneNode(true);
  mediaEl.parentNode.appendChild(newMediaEl);
  mediaEl.parentNode.removeChild(mediaEl);
  return newMediaEl;
}
```

Zeouterlimits:
Will re-open. #rebaseShame

PR #1910, opened 2016-09-14
Issue #1881 - Fix for null src, will not set crossorigin rather than …

Zeouterlimits:
**Description:**
Fixing [Issue #1881 ](https://github.com/aframevr/aframe/issues/1881#issuecomment-244658276), where a previous change caused a regression for usage of the dynamic src element.
Now handling when src is not present, where it will return null or an empty string - [https://developer.mozilla.org/en/docs/Web/API/Element/getAttribute](https://developer.mozilla.org/en/docs/Web/API/Element/getAttribute)

**Changes proposed:**
- If src is null, set cross origin instead of throwing error

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -156,11 +156,13 @@ function setCrossOrigin (mediaEl) {
 
   src = mediaEl.getAttribute('src');
 
-  // Does not have protocol.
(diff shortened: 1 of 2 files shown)


ngokevin:
Perfect, thanks!

PR #1920, opened 2016-09-18
Add intersection data to cursor 'click' events.

### LINKED ISSUE #1917: Attach intersection event details to cursor click events
**Description:**

These details about an intersection are attached to the `raycaster-intersected` event:

```
distance: number
face: THREE.Face3
faceIndex: number
object: THREE.Mesh
point: THREE.Vector3
uv: THREE.Vector2
```

It'd be lovely if that same intersection data were also available on `click` events.

donmccurdy:
Fixes #1917.


ngokevin:
looks good, thanks!

PR #1923, opened 2016-09-19
Raise 'asking a question' to top level in contributing.md

donmccurdy:
May be worth creating a top-level header on how/where to ask questions (A: Stack Overflow), rather than having that under the 'File an Issue' header. The distinction of a "question rather than an issue" may not be self-evident.


dmarcos:
Thanks for this!

PR #1925, opened 2016-09-19
getAttribute -> getDefinedAttribute, getComputedAttribute -> getAttribute (fixes #1590)

ngokevin:
**Description:**

I'd expect that `getAttribute` returns full component data. 99% of the time we're grabbing component data, we want the computed component data, not just what is defined in the attribute. Since everyone knows DOM's `getAttribute` rather than A-Frame's `getComputedAttribute`, this is the method they'll reach for first. Since `getAttribute` currently returns only *defined* data, it is confusing when `el.getAttribute('position')` returns `null`.

**Changes proposed:**
- Old `getAttribute` is now an explicit `getDefinedAttribute`
- Old `getComputedAttribute` is now `getAttribute`
- [x] docs

### DIFF
--- a/docs/core/entity.md
+++ b/docs/core/entity.md
@@ -189,12 +189,12 @@ entity.getAttribute('data-position');
 // >> "0 1 1"
 ```
 
-### `getDefinedAttribute (componentName)`
+### `getDOMAttribute (componentName)`
 
-`getDefinedAttribute` retrieves only parsed component data that is explicitly
+`getDOMAttribute` retrieves only parsed component data that is explicitly
 defined in the DOM or via `setAttribute`. If `componentName` is the name of a
-registered component, `getDefinedAttribute` will return only the component data
-defined in the HTML as a parsed object. `getDefinedAttribute` for components is
+registered component, `getDOMAttribute` will return only the component data
+defined in the HTML as a parsed object. `getDOMAttribute` for components is
 the partial form of `getAttribute` since the returned component data does not
 include applied mixins or default values:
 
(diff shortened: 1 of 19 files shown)


donmccurdy:
I like `getAttribute` having the default behavior of returning full component data: this should make a lot of my code more consistent.

To me `getDefinedAttribute()` isn't a super-intuitive name for the alternative. Perhaps `getShallowAttribute()`, `getAttributeShallow()`, `getNonDefaultAttribute()`, `getAttributeOverrides()`? Are there cases when the explicitly overridden subset of properties should be preferred, for performance or something, and if not should it even be exposed?

ngokevin:
I like `getDefinedAttribute` because it is getting data from only explicitly *defined* HTML attributes (either via DOM or `setAttribute`). Does that clear it up, or do we need more alternatives? At least this method will rarely ever be called in the context of components.

`shallow` isn't as immediately clear to me (I think of shallow vs. deep object cloning). `getNonDefaultAttribute` isn't great because the discrepancy not related to default attributes, it is also factoring in mixins. And I don't think setting HTML attributes as `overrides`.

dmarcos:
What about `getDOMAttribute` or `getElementAttribute` ?

ngokevin:
They're actually not bad, the only confusion is when you do `el.setAttribute('geometry' {primitive: 'box'})`, then our new method should return that data, but it is not reflected in the actual DOM nor element. Which is why I liked `defined` because it takes into account whether defined via DOM or setAttribute.

donmccurdy:
Kk, all fine with me.

dmarcos:
I still think `getDefinedAttribute` is a very cryptic name. `getElementAttribute` and `getDOMAttribute` are more explicit. They might be inconsistent if you look at the DOM inspector but I expect that use case to become more and more rare as people transition to the inspector.

ngokevin:
updated

PR #1928, opened 2016-09-20
Remove parentEl references when entity is detached. (fixes #1444)

### LINKED ISSUE #1444: entities do not reinitialize when detached and then reattached to DOM
**Description:**

Able to use **detach** to remove an a-entity, but **append** does not work to place it back on the scene.

See here, where .redCube is an a-entity that is a cube, and masterScene is the class applied to a-scene.

```
$( ".redCube" ).click(function() {
  $( this ).toggleClass( "off" );
});
var p;
$( "button" ).click(function() {
  if ( p ) {
    p.appendTo( ".masterScene" );
    p = null;
  } else {
    p = $( ".redCube" ).detach();
  }
});
```

```
  <a-assets>
  <a-mixin id="cube" geometry="primitive: box"></a-mixin>
  <a-mixin id="cube-hovered" material="color: magenta"></a-mixin>
  <a-mixin id="cube-selected" material="color: cyan"></a-mixin>
  <a-mixin id="red" material="color: red"></a-mixin>
  <a-mixin id="green" material="color: green" data></a-mixin>
  </a-assets>

 <a-entity position="0 2.2 4">
    <a-entity camera look-controls wasd-controls>
      <a-entity position="0 0 -3"
                geometry="primitive: ring; radiusOuter: 0.10;
                          radiusInner: 0.08;"
                material="color: cyan; shader: flat"
                cursor="maxDistance: 30; fuse: true">
        <a-animation begin="click" easing="ease-in" attribute="scale"
             fill="backwards" from="0.1 0.1 0.1" to="1 1 1" dur="150"></a-animation>
        <a-animation begin="fusing" easing="ease-in" attribute="scale"
             fill="forwards" from="1 1 1" to="0.1 0.1 0.1" dur="1500"></a-animation>
      </a-entity>
    </a-entity>
  </a-entity>

<a-scene class="masterScene" >
      <a-entity position="0 0 -6" class="greenCube mixin="cube green">
                   <a-animation begin="click" attribute="rotation" to="0 360 0"
                   easing="linear" dur="2000" fill="backwards"></a-animation>
     </a-entity>
      <a-entity position="5 2 -6" class="redCube" mixin="cube red">
                   <a-animation begin="click" attribute="rotation" to="0 360 0"
                   easing="linear" dur="2000" fill="backwards"></a-animation>
     </a-entity>
</a-scene>
```

- A-Frame Version: 0.2.0
- Platform/Device: Mac/Chrome and tested on iphone6+/ios 9.3.1

donmccurdy:
Failure to remove parentEl references blocks the re-attachment callback
from completing, which prevents the element from being added back to the
scene graph.

Fixes #1444.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -102,7 +102,7 @@ var proto = Object.create(ANode.prototype, {
       if (!this.parentEl || this.isScene) { return; }
       // Remove components.
       Object.keys(this.components).forEach(bind(this.removeComponent, this));
-      this.parentEl.remove(this);
+      this.removeFromParent();
     }
   },
 
@@ -233,6 +233,17 @@ var proto = Object.create(ANode.prototype, {
 
       parentNode.add(this);
       this.attachedToParent = true;  // To prevent multiple attachments to same parent.
+    }
+  },
+
+  /**
+   * Tell parentNode to remove this entity from itself.
+   */
+  removeFromParent: {
+    value: function () {
+      this.parentEl.remove(this);
+      this.parentEl = this.parentNode = null;
+      this.attachedToParent = false;
     }
   },
 
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -103,6 +103,7 @@ var proto = Object.create(ANode.prototype, {
       // Remove components.
       Object.keys(this.components).forEach(bind(this.removeComponent, this));
       this.removeFromParent();
+      ANode.prototype.detachedCallback.call(this);
     }
   },
 
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -72,7 +72,9 @@ module.exports = registerElement('a-node', {
     },
 
     detachedCallback: {
-      value: function () { /* no-op */ }
+      value: function () {
+        this.hasLoaded = false;
+      }
     },
 
     /**
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -23,9 +23,13 @@ module.exports = registerElement('a-node', {
 
     attachedCallback: {
       value: function () {
-        var mixins = this.getAttribute('mixin');
+        var mixins;
+
+        this.hasLoaded = false;
         this.sceneEl = this.closestScene();
         this.emit('nodeready', {}, false);
+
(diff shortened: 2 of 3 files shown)


ngokevin:
I think we should add a test to cover detaching an entity, then reattaching it.

donmccurdy:
Here's a test on detachment and re-attachment.

Components do work for my manual checks – geometry re-appears, and with the "cursor" demo I can remove one of the cubes, re-attach it, and it will animate on click as expected.

However, adding a test to cover that (verify that `el.components.foo` is truthy after re-attachment) does *not* pass. I didn't go too far down the rabbit hole, but in `a-entity::load()` we hit this line:

```
```

hasLoaded is set to `false` in `a-node.js` on *creation*, not when attached to the DOM, so resetting it to false on detachment would feel weird.

ngokevin:
If we set `hasLoaded` on detach, it'll work? If so, we should maybe do that with a comment on why.

ngokevin:
Or maybe we also set `hasLoaded` on `attachedCallback`

donmccurdy:
Eh, yes I think that works but it feels bad. And (the reason for that lovely new MutationObserver), `loaded` is not going to fire a second time, so any code depending on that will be sorely disappointed.

I don't see why that `hasLoaded` check is there, actually. The tests pass without it. `entity.load()` is invoked in `attachedCallback` and attachment is inherently something that can happen multiple times.

donmccurdy:
Maybe the alternative fix here is, if we can't depend on components to clean up after themselves correctly (which is likely if the entity/node code has trouble with it) it might be better to flag the element as 'dirty' on detachment, and throw an error if it's re-attached.

donmccurdy:
Components do get entirely re-instantiated, so the concerns are simpler there. Here's an updated test that covers components, by setting hasLoaded=false in `detachedCallback`.

ngokevin:
I think the `hasLoaded` check is so it doesn't try to initialize itself twice?

Did you see the comment about maybe setting it on `attachedCallback` rather than `detachedCallback`? That seems less icky.

ngokevin:
I pushed a patch to also do it on attachedCallback, for good measure.

PR #1932, opened 2016-09-23
Docs: fix typos and links in "Building with Components" guide

cvan:


docs/guides/building-with-components.md
```diff
@@ -85,10 +85,10 @@ HTML attribute:
          material></a-plane>
 ```
 
-Then we the specify the component data using a syntax that looks like inline
-CSS styles. We set `shader` to `flat` so the image isn't affected negatively by
-lighting. And we set `src` to `#cubes-thumb`, a selector to one of the images
-defined in the asset management system.
+Then, we the specify the component data using a syntax that resembles that of
```
ngokevin: don't need to put a comma after then/and/so/but, not intending a pause.

docs/guides/building-with-components.md
```diff
@@ -42,7 +42,7 @@ This is the starting point for our scene:
     <img id="sechelt-thumb" src="img/thumb-sechelt.png">
   </a-assets>
 
-  <!-- 360-degree image. -->
+  <!-- 360° image. -->
```
ngokevin: i usually don't have unicode characters in code. might flub up some editors if they're not configured

ngokevin:
Addressed the comments

PR #1934, opened 2016-09-23
Added sounds pool to sound component

fernandojsg:
**Description:** Added `pool` attribute to create a soundpool for the `sound` component. With that we could play multiple sound on the same instance at the same time, for example when you're shooting a gun multiple times.

**Changes proposed:**
- Add `pool` attribute, by default 1
- Replace `this.sound` from the component with `this.soundsPool` that will store the multiple instances of the sound.
- I've modified the `cursor` example adding a new cube that you could click quickly to see the effect.

I didn't update the test yet until you validate the proposal.


src/components/sound.js
```diff
@@ -50,10 +52,12 @@ module.exports.Component = registerComponent('sound', {
...
     if (srcChanged) {
       this.audioLoader.load(data.src, function (buffer) {
-        sound.setBuffer(buffer);
+        this.soundsPool.forEach(function (sound) {
+          sound.setBuffer(buffer);
+        });
         // Remove this key from cache, otherwise we can't play it again
         THREE.Cache.remove(data.src);
-      });
+      }.bind(this));
```
dmarcos: Do `var self = this;` at the top instead of binding

src/components/sound.js
```diff
@@ -125,24 +131,47 @@ module.exports.Component = registerComponent('sound', {
...
-    sound = this.sound = new THREE.PositionalAudio(listener);
-    el.setObject3D(this.attrName, sound);
-
-    sound.source.onended = function () {
-      sound.onEnded();
-      el.emit('sound-ended');
-    };
+    this.soundsPoolGroup = new THREE.Group();
+    this.soundsPool = [];
+    // var soundGroup = new THREE.Group();
```
dmarcos: Remove

src/components/sound.js
```diff
@@ -12,36 +12,38 @@ module.exports.Component = registerComponent('sound', {
   schema: {
     src: {type: 'src'},
     on: { default: '' },
-    autoplay: {default: false},
-    loop: {default: false},
-    volume: {default: 1}
+    autoplay: { default: false },
+    loop: { default: false },
+    volume: { default: 1 },
+    pool: { default: 1 }
```
dmarcos: `poolSize` might be a better name

src/components/sound.js
```diff
@@ -125,24 +131,47 @@ module.exports.Component = registerComponent('sound', {
...
-    if (!this.sound.source.buffer) { return; }
-    this.sound.play();
+    for (var i = 0; i < this.soundsPool.length; i++) {
+      var sound = this.soundsPool[i];
+      if (!sound.isPlaying && sound.source.buffer) {
+        this.soundsPool[i].play();
+        return;
+      }
+    }
+    warn('Not unused sound found on pool, please increate its size using the `pool` attribute');
```
dmarcos: Maybe:
`All the sounds are playing. If you need to play more sounds simultaneously consider increasing the size of pool with the `poolSize` attribute.`

src/components/sound.js
```diff
@@ -125,24 +131,47 @@ module.exports.Component = registerComponent('sound', {
...
   },
 
   stopSound: function () {
-    if (!this.sound.source.buffer) { return; }
-    this.sound.stop();
+    this.soundsPool.forEach(function (sound) {
```
dmarcos: Is it possible you just want to stop one of the sounds of the pool?

fernandojsg: I don't think so, as the user won't have the control of the playing buffer. For example if you have a gun you just keep shooting and suddenly you want to silent it but don't have any idea or way to know the exactly buffer that it's currently playing.
I believe we could leave it this way and if we get the need we could reevaluate it, as for example we'll need to give some feedback back to the entity to use it.

src/components/sound.js
fernandojsg: Yep, thanks

src/components/sound.js
fernandojsg: :+1:

src/components/sound.js
fernandojsg: done

tests/components/sound.test.js
```diff
@@ -45,35 +48,35 @@ suite('sound', function () {
...
 
-  suite('pause', function () {
-    test('does not call sound pause if not playing', function () {
+  suite('soundPause', function () {
+    test('does not call sound soundPause if not playing', function () {
       var el = this.el;
       var sound = el.components.sound.sound = {
         disconnect: sinon.stub(),
-        pause: sinon.stub(),
+        soundPause: sinon.stub(),
```
ngokevin: It's named `pauseSound`, right?

src/components/sound.js
```diff
@@ -12,36 +12,38 @@ module.exports.Component = registerComponent('sound', {
   schema: {
     src: {type: 'src'},
     on: { default: '' },
-    autoplay: {default: false},
-    loop: {default: false},
-    volume: {default: 1}
+    autoplay: { default: false },
```
ngokevin: nit: could alphabetize properties and remove spaces

fernandojsg: :+1:

tests/components/sound.test.js
fernandojsg: Yeah right!

fernandojsg:
@ngokevin @dmarcos updated with the proposed changes and tests fixed.

src/components/sound.js
```diff
@@ -22,35 +23,39 @@ module.exports.Component = registerComponent('sound', {
   init: function () {
     this.listener = null;
     this.audioLoader = new THREE.AudioLoader();
-    this.sound = null;
+    this.soundsPool = new THREE.Group();
```
ngokevin: This could be `this.pool` (since we're already in the sound component) or `this.soundPool` (since `pool` already connotes plural)

src/components/sound.js
```diff
@@ -125,24 +127,48 @@ module.exports.Component = registerComponent('sound', {
...
+        el.emit('sound-ended', {index: i});
+      };
+    });
   },
 
   playSound: function () {
-    if (!this.sound.source.buffer) { return; }
-    this.sound.play();
+    // Look for an unused sound in the pool and play it if found.
+    for (var i = 0; i < this.soundsPool.children.length; i++) {
```
ngokevin: can use forEach?

src/components/sound.js
```diff
@@ -74,29 +79,27 @@ module.exports.Component = registerComponent('sound', {
...
       // disconnect() will throw if it was never connected initially.
       warn('Audio source not properly disconnected');
     }
   },
 
   play: function () {
-    if (!this.sound) { return; }
-    if (this.sound.source.buffer && this.data.autoplay) {
-      this.sound.play();
+    // Look for an unused sound in the pool and play it if found.
```
ngokevin: this can call `playSound` right?

src/components/sound.js
fernandojsg: I used because I use `return` to exit the method if a instance is found on the pool, otherwise I show the warning message. I'll change it to use foreach and some boolean variable `isFound` that will determine if the message will be shown or not

src/components/sound.js
fernandojsg: Yep! done.

donmccurdy:
👏  i was working on collision sound effects earlier, w/ physics, and this looks really promising

PR #1939, opened 2016-09-26
Modify the begin attribute of a-animation

shaabhishek:
**Description:**

**Changes proposed:**
-
-
-

"fusing" doesn't work, while "cursor-fusing" does, as mentioned in the cursor docs


ngokevin:
thanks!

PR #1943, opened 2016-09-27
fix unrendered "Text Geometry Component" link in FAQ

cvan:
Sorry for another PR; I didn't catch this earlier. I would've fixed this along with the other link I fixed in PR #1942.


ngokevin:
np thanks

PR #1950, opened 2016-09-29
component-initialized event

ngokevin:
**Description:**

Wanted a way to depend on other components dynamically. (e.g., I have a component that draws line, where the line data resides in another component. Thus the line component must wait for the other component to init, without knowing beforehand what that component is nor hardcoding it to avoid decoupling).

**Changes proposed:**
- add component-initialized event

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -201,16 +201,20 @@ Component.prototype = {
       // Play the component if the entity is playing.
       this.update(oldData);
       if (el.isPlaying) { this.play(); }
+      el.emit('componentinitialized', {
+        id: this.id,
+        name: this.name,
+        data: this.getData()
+      }, false);
     } else {
       this.update(oldData);
+      el.emit('componentchanged', {
+        id: this.id,
+        name: this.name,
+        newData: this.getData(),
+        oldData: oldData
+      }, false);
     }
-
-    el.emit('componentchanged', {
-      id: this.id,
-      name: this.name,
-      newData: this.getData(),
-      oldData: oldData
-    }, false);
   },
 
   /**
--- a/tests/components/raycaster.test.js
+++ b/tests/components/raycaster.test.js
@@ -79,7 +79,7 @@ suite('raycaster', function () {
       var el = this.el;
       var newEl = this.targetEl = document.createElement('a-entity');
       var numObjects = el.components.raycaster.objects.length;
-      newEl.addEventListener('loaded', function () {
+      el.sceneEl.addEventListener('child-attached', function () {
         assert.equal(el.components.raycaster.objects.length, numObjects + 1);
         done();
       });
--- a/docs/core/entity.md
+++ b/docs/core/entity.md
@@ -364,8 +364,9 @@ entity.is('selected');  // >> false
 | Event Name       | Description                                                         |
 |------------------|---------------------------------------------------------------------|
 | child-attached   | A child was attached to the entity.                                 |
-| componentchanged | One of the entity's component's data was modified.                  |
-| componentremoved | One of the entity's component was removed.                          |
+| componentchanged | One of the entity's components was modified.                        |
+| componentinit    | One of the entity's components was initialized.                     |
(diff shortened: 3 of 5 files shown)


donmccurdy:
In practice, how would I use this when writing a component with these dependencies?

1. During `myComponent.init()`, check for whatever I need. 
2. If missing, listen for `component-initialized`.
3. On `component-initialized`, check for dependency again. If missing, go back to 2. Else, go to 4.
4. Finish initialization.

In cases like you the 'line data' example, consistent event names are probably most important. For example, having all loaders emit the `model-loaded` event, regardless of the format, is very helpful.

(Note – I think adding a `component-initialized` event seems worthwhile regardless. Just brainstorming whether this is the ideal way to handle dynamic component dependencies.)

ngokevin:
I made a hacky API.

```html
<a-entity draw-line="pointer: movement-pattern.movementPattern.spline" movement-pattern">
```

It will grab the `spline` object from `el.components['movement-pattern'].movementPattern.spline` and draw it. So I needed to wait on that component to initialize, at which it'd be ready.

Maybe I should just couple them, or make `draw-line` just an internal function. This is dirty, I admit.

donmccurdy:
Ok cool. Yeah for most cases I can think of, I'd probably use a custom event or method. But `component-initialized` seems at least as important as `componentchanged`, from a lifecycle perspective, so this event works fine.

(Difference in hyphenation is because the all-one-word thing is phasing out, i assume?)

src/core/component.js
```diff
@@ -201,16 +201,20 @@ Component.prototype = {
       // Play the component if the entity is playing.
       this.update(oldData);
       if (el.isPlaying) { this.play(); }
+      el.emit('component-initialized', {
```
dmarcos: remove dash from the event name for consistency?

tests/components/raycaster.test.js
```diff
@@ -80,8 +80,10 @@ suite('raycaster', function () {
       var newEl = this.targetEl = document.createElement('a-entity');
       var numObjects = el.components.raycaster.objects.length;
       newEl.addEventListener('loaded', function () {
-        assert.equal(el.components.raycaster.objects.length, numObjects + 1);
-        done();
+        process.nextTick(function () {
```
dmarcos: Why do we need a nextTick here?

src/core/component.js
ngokevin: OK. Seems like half our event names are hyphened and half are not (e.g., `enter-vr`). But I can keep it without here.

tests/components/raycaster.test.js
ngokevin: Saw the test was failing. Maybe the raycaster handler was getting called after this handler.

ngokevin: Actually, could also change this tets to listen to child-attached.

ngokevin:
re: event name hyphen/no space. Well, half the codebase uses hyphen, half no space. But I guess I'll keep it consistent within the module

ngokevin:
updated. also added ID to componentremoved

tests/components/raycaster.test.js
```diff
@@ -79,7 +79,7 @@ suite('raycaster', function () {
       var el = this.el;
       var newEl = this.targetEl = document.createElement('a-entity');
       var numObjects = el.components.raycaster.objects.length;
-      newEl.addEventListener('loaded', function () {
+      el.sceneEl.addEventListener('child-attached', function () {
```
dmarcos: We should remove dashes from event names for consistentcy

ngokevin: ok. will have to do that separately since a pretty large change.

dmarcos:
Question. What's wrong with your `draw-line` having `movement-pattern` as dependency?

ngokevin:
It might've been preemptive, but it's possible that component could work to draw any spline. Just wanted to decouple and make it reusable.

PR #1954, opened 2016-09-30
Add entities pool

dmarcos:


examples/test/pool/components/spaceship.js
```diff
@@ -0,0 +1,53 @@
...
+  pause: function () {
+    this.removeKeyEventListeners();
+  },
+
+  onKeyDown: function (evt) {
+    var el = this.el;
+    var laserEl;
+    var position;
+    if (evt.keyCode !== 32) { return; }
+    laserEl = el.sceneEl.components.pool.requestEntity();
```
fernandojsg: What about `requestEntity('laser')`?

examples/test/pool/components/spaceship.js
```diff
@@ -0,0 +1,53 @@
...
+
+  laserDestroyer: function (el) {
+    var laserEl = el;
+    var component = this;
+    return function () {
+      if (!laserEl.isPlaying) {
+        setTimeout(component.laserDestroyer(laserEl), 1000);
+        return;
+      }
+      this.el.sceneEl.components.pool.returnEntity(laserEl);
```
fernandojsg: the same as before `returnEntity('laser', laserEl)`?

src/components/scene/pool.js
```diff
@@ -0,0 +1,65 @@
+var debug = require('../../utils/debug');
+var register = require('../../core/component').registerComponent;
+
+var warn = debug('components:pool:warn');
+
+module.exports.Component = register('pool', {
+  schema: {
+    mixin: {default: ''},
+    size: {default: 0}
+  },
```
fernandojsg: Add another parameter to force the `requestEntity` to return a new entity even if there's no available entity on the pool, it will create a new one and pushed to the pool and return it

src/components/scene/pool.js
```diff
@@ -0,0 +1,65 @@
...
+
+  createEntities: function () {
+    var el;
+    var mixin = this.data.mixin;
+    if (!mixin) { return; }
+    this.pool = [];
+    this.pooledEls = [];
+    for (var i = 0; i < this.data.size; ++i) {
+      el = document.createElement('a-entity');
```
fernandojsg: Takes this part out from the loop to a another function `createEntity/pushNewEntity/...`, so it could be reused on `requestEntity` when the previous described param is set to true and there's no available entity

examples/test/pool/components/spaceship.js
dmarcos: what does that mean?

dmarcos: if you want to have multiple pools you add multiple pool components and call requestEntity on them: 
`<a-entity pool__laser="mixin: laser; size: 10" pool__bomb="mixin: bomb; size: 10"></a-entity>`

dmarcos: `sceneEl.components.pool__laser.requestEntity()`

fernandojsg: aps sorry I didn't notice that `pool` was the name of the component and not a global that will store all the pools. I mistook this use case with accessing directly the system.

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+ * It initialized a pool of entities that will be reused.
+ * It avoids creating and destroying the same kind entities
+ * in dynamic scenes. Useful for example in a game where you want
+ * to reuse enemies entities
+ *
+ * @member {array} poolEls - Available entities in the pool.
+ * @member {array} pooledEls - Entities of the pool in use.
+ *
+ */
+module.exports.Component = register('pool', {
```
ngokevin: Common stuff I see online refer to it as "Object Pooling" so consider `entity-pool` to be more clear. I might've not immediately understood what a `pool` component did, but `entity-pool` I might've.

examples/test/pool/components/laser.js
```diff
@@ -0,0 +1,22 @@
+/* global AFRAME */
+
+AFRAME.registerComponent('laser', {
+  schema: {
+    speed: { default: 1 }
+  },
+
+  init: function () {
+    var el = this.el;
+    var geometry = 'primitive: box; height: 2; width: 0.1; depth: 0.1';
```
ngokevin: Intermediate vars necessary?

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
+var debug = require('../../utils/debug');
+var register = require('../../core/component').registerComponent;
```
ngokevin: `s/register/registerComponent`

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @member {array} poolEls - Available entities in the pool.
+ * @member {array} pooledEls - Entities of the pool in use.
+ *
+ */
+module.exports.Component = register('pool', {
+  schema: {
+    mixin: {default: ''},
+    size: {default: 0},
+    dynamicSize: {default: false}
```
ngokevin: `dynamicallySize` might differentiate it as a boolean. `dynamicSize` sounds like it'd take a size.

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+    var mixin = this.data.mixin;
+    if (!mixin) { return; }
+    this.poolEls = [];
+    this.pooledEls = [];
+    for (var i = 0; i < this.data.size; ++i) {
+      this.createEntity();
+    }
+  },
+
+  update: function (oldData) {
```
ngokevin: i like all lifecycle handlers on top in order they're invoked, and then internal methods

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+  init: function () {
+    this.createEntities();
+  },
+
+  createEntities: function () {
+    var mixin = this.data.mixin;
+    if (!mixin) { return; }
+    this.poolEls = [];
+    this.pooledEls = [];
+    for (var i = 0; i < this.data.size; ++i) {
```
ngokevin: vars on top

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+var warn = debug('components:pool:warn');
+
+/**
+ * Pool component.
+ * It initialized a pool of entities that will be reused.
+ * It avoids creating and destroying the same kind entities
+ * in dynamic scenes. Useful for example in a game where you want
+ * to reuse enemies entities
+ *
+ * @member {array} poolEls - Available entities in the pool.
```
ngokevin: Maybe just `entities` and `activeEntities`, `poolEls` and `pooledEls` look the same

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+  multiple: true,
+
+  init: function () {
+    this.createEntities();
+  },
+
+  createEntities: function () {
+    var mixin = this.data.mixin;
+    if (!mixin) { return; }
+    this.poolEls = [];
```
ngokevin: I'd comment that this clears the pool

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Play wrapper for pooled entities. When pausing and playing
+   * a scene we don't want to play the entities that are not in use
+   */
+  wrapPlay: function (playMethod) {
+    var pooledEls = this.pooledEls;
+    return function () {
+      if (pooledEls.indexOf(this) === -1) { return; }
+      playMethod.call(this);
```
ngokevin: This line doesn't have test coverage

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+   * Used to request one of the available entities of the pool
+   */
+  requestEntity: function () {
+    var el;
+    if (this.poolEls.length === 0) {
+      if (this.data.dynamicSize === false) {
+        warn('Requested entity from empty pool ' + this.name);
+        return;
+      }
+      this.createEntity();
```
ngokevin: This line doesn't have test coverage

src/components/scene/pool.js
```diff
@@ -0,0 +1,103 @@
...
+    el = this.poolEls.shift();
+    this.pooledEls.push(el);
+    el.setAttribute('visible', true);
+    return el;
+  },
+
+  /**
+   * Used to return a used entity to the pool
+   */
+  returnEntity: function (el) {
```
ngokevin: Maybe `releaseEntity`? `returnEntity` sounds like it'll return an entity

tests/components/scene/pool.test.js
```diff
@@ -0,0 +1,80 @@
+/* global assert, process, setup, suite, test */
+
+suite('pool', function () {
+  setup(function (done) {
+    var el = this.sceneEl = document.createElement('a-scene');
```
ngokevin: You can do: 

```js
var sceneEl = helpers.entityFactory().sceneEl;
var mixinEl = helpers.mixinFactory('test', {material: 'color: red'}, sceneEl);
```

This will create `<a-assets>` and `<a-mixin>` for you.

examples/test/pool/index.html
```diff
@@ -0,0 +1,55 @@
...
+      <a-entity position="0 0 20">
+        <a-entity camera="fov: 45"></a-entity>
+      </a-entity>
+    </a-scene>
+  </body>
+  <script>
+    var sceneEl = document.querySelector('a-scene');
+    var message = '<p>Left and right arrows to move and space bar to shoot</p><p>Used Pool Entities ';
+    sceneEl.addEventListener('loaded', updateMessage);
+    function updateMessage() {
```
ngokevin: I think we should encourage a component here:

- Don't have to wait for the scene to load. Lifecycle is handled for you
- Don't have to create your own RAF.

src/components/scene/pool.js
dmarcos: It actually returns the entity to the pool. Release is a term used in memory allocation / deallocation

examples/test/pool/components/laser.js
dmarcos: It makes lines shorter. Easier to read

src/components/scene/pool.js
dmarcos: I would not be too anal about 100% coverage

examples/test/pool/index.html
dmarcos: I prefer to see this right there on the index.html and not hidden in a component.

src/components/scene/pool.js
dmarcos: the word `active` does not really convey what those variables do.

src/components/scene/pool.js
dmarcos: dynamicallySize is a mouthful. Adverbs should be ban from earth.

src/components/scene/pool.js
dmarcos: Where is that stated? Is not a policy difficult to understand, enforce and context dependent? Sometimes we sort things alphabetically and sometimes we don't. I recently realized that the order of variables or methods in a file does not help me much. On the other hand it introduces friction and sometimes you have to do weird contortions to conform to the policy like for instance when sorting the variables alphabetically.

src/components/scene/pool.js
dmarcos: When an application requires pools I expect to often see multiple instances `pool__bullets`, `pool__enemies`... If we rename `pool` to `entity-pool` you will end up with `entity-pool__enemies` , `entity-pool_bullets`, Is not that to verbose? @fernandojsg

src/components/scene/pool.js
fernandojsg: I agree with @ngokevin that `dynamicSize` sounds like it expects a numeric value instead of boolean, but I don't like `dynamicallySize` either, what about just `dynamic`?

src/components/scene/pool.js
fernandojsg: `Entity pool` is more clear than just `pool` but due its nature, we'll have plenty of instances on a-scene therefore we'll have `entity-pool__whatever` as @dmarcos said and it looks quite ugly, so I would choose to use just `pool`

src/components/scene/pool.js
dmarcos: ... and to be correct you would have to pair it with a participle: `dynamicallySized`

examples/test/pool/index.html
ngokevin: You can still define it in the html file

src/components/scene/pool.js
ngokevin: inPoolEls / outPoolEls? Any suggestions would help differentiate since they both sound like they're in the pool

src/components/scene/pool.js
ngokevin: Small test to prevent potential bugs, good trade off (we've had plenty bugs that could've been caught with tests). At least test the below case of dynamically fetching from any empty pool.

src/components/scene/pool.js
fernandojsg: used/unused, available... ?

fernandojsg:
It's looking good to me, @ngokevin ?

src/components/scene/pool.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Pool component.
+ * A pool of entities that will be reused.
+ * It avoids creating and destroying the same kind of entities
+ * in dynamic scenes. It will help reduce GC pauses. Useful for example
+ * in a game where you want to reuse enemies entities.
+ *
+ * @member {array} availableEls - Available entities in the pool.
+ * @member {array} useedEls - Entities of the pool in use.
```
ngokevin: typo

docs/components/pool.md
```diff
@@ -0,0 +1,23 @@
...
+
+A pool of entities that will be reused. It avoids creating and destroying the same kind of entities in dynamic scenes. It will help reduce GC pauses. Useful for instance in a game where you want to reuse enemy entities.
+
+## Example
+
+```html
+<a-scene pool="mixin: enemy; size : 10"></a-scene>
+```
+
+## Properties
```
ngokevin: should probably document the methods as well since those are part of the API

examples/test/pool/index.html
ngokevin: If you want it, it's a bit cleaner:

```js
var message = '<p>Left and right arrows to move and space bar to shoot</p><p>Used Pool Entities ';
AFRAME.registerComponent('update-pool-message', {
  schema: {type: 'selector'},
  tick: function () {
    var poolComponent = this.el.components.pool;
    var poolSize = poolComponent.availableEls.length + poolComponent.usedEls.length;
    var usedPoolEntities = this.el.components.pool.usedEls.length;
    this.data.innerHTML = message + usedPoolEntities + '/' + poolSize + '</p>';
  }
});
```

tests/components/scene/pool.test.js
```diff
@@ -0,0 +1,75 @@
+/* global assert, process, setup, suite, test */
+var helpers = require('../../helpers');
+
+suite('pool', function () {
+  setup(function (done) {
+    var el = helpers.entityFactory();
+    var sceneEl = this.sceneEl = el.parentNode;
+    sceneEl.setAttribute('pool', 'mixin: test; size: 1');
```
ngokevin: test `dynamic`?

dmarcos:
I added a test for a dynamic pool and made the example screen message a component. I'm merging this....

PR #1968, opened 2016-10-10
[docs] Update scene.md

theatlasroom:
**Description:**
The sentence didnt make much sense

**Changes proposed:**
-
-
-

I think this might be a typo


dmarcos:
Good catch! Thank you

PR #1971, opened 2016-10-11
Add wireframe params to Flat and Standard Material

AdaRoseEdwards:
**Description:**

Moved in a couple more of the THREE.js properties WRT wireframes. 

The default width is set to 2 so that it looks cleaner on VR Headsets

If there are reasons for them not being included that is fine and I just thought it could be useful.

**Changes proposed:**

Allow standard and flat materials to use wireframes.

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -63,7 +63,7 @@ A-Frame ships with a few built-in materials.
 
 ### `standard`
 
-The `standard` material is the default material. It uses the physically-based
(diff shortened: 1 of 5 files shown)


dmarcos:
Thanks! Can you update the docs too?

AdaRoseEdwards:
Just checking you were interested before tests and docs, will do them now. :smile:

ngokevin:
thanks!

PR #1972, opened 2016-10-11
add questions about performance concerns to FAQ

ngokevin:


docs/introduction/faq.md
```diff
@@ -36,26 +36,48 @@ a vehicle in which to make WebVR successful.
...
+objects themselves.
+- When changing an entity's properties, data is not serialized back to the DOM. This
+reduces the DOM overhead and most modifications are done directly in memory,
+similar to Virtual DOM.
+- HTML attributes explicitly set from an HTML file are cached in memory.
+
+[bestpractices]: ../introduction/best-pratices.md
+
+However, the performance of complex A-Frame scenes depends on the optimizations
+that developers employ. In 3D and game development, resources are more scarce,
```
dmarcos: `Resource are more scarce` sounds like a strange statement.  What about something like this:

`In a-frame we provide a set of reasonable defaults that yield good performance for the most common use cases. However, performance is ultimately determined by the complexity and characteristics of each individual scene. In order to get the best use of resources developers will need understanding on how a 3D graphics pipeline works. These are some guidelines that will help you get started`

docs/introduction/faq.md
```diff
@@ -36,26 +36,48 @@ a vehicle in which to make WebVR successful.
...
 
-If you need additional support, [contact us][contact]! We are very interested
-in understanding learning curve difficulties.
+Past that, you can browse the resources on the [`awesome-aframe`
+repository][awesome]. If you need additional support, [ask on
+StackOverflow][stackoverflow].
+
+## How is A-Frame's performance?
+
+Historically, JavaScript and the browser have a bad reputation for performance
```
dmarcos: I would take out `Javascript` and focus just on the DOM as the performance bottleneck. What about this?
`The main performance bottleneck for Web Applications comes from the DOM and its layout system which was originally designed for static pages and later re purposed for dynamic applications. A-Frame uses the DOM to represent a scene graph of a 3D application using custom elements. Those elements are not handled by the browser rendering system but managed by A-Frame and ultimately rendered with WebGL. This allows us to run 90+ FPS experiences in the HTC Vive. See [A-Painter](https://blog.mozvr.com/a-painter/) for an example of a full application`

ngokevin:
addressed, thanks!

Issue #1959, opened 2016-10-05
leaveCurrentIntersection() is not defined etc.


rong1129:
Function leaveCurrentIntersection() on [this line](https://github.com/aframevr/aframe/blob/37c33d2cc88de369feff7027f3d1e24b305ef496/src/components/cursor.js#L106) doesn't exist. I changed it to clearCurrentIntersection(), it seemed working. 

Probably unrelated to this, but [this line](https://github.com/aframevr/aframe/blob/37c33d2cc88de369feff7027f3d1e24b305ef496/src/components/cursor.js#L139) doesn't seem right either as clearCurrentIntersection() doesn't take any argument.

BTW, not sure if anyone tried this component with raycaster as described [here](https://aframe.io/docs/0.3.0/components/cursor.html#configuring-the-cursor-through-the-raycaster-component), the object selector doesn't work.
 
- A-Frame Version: master
- Platform/Device: all
- Reproducible Code Snippet or Demo URL [highly encouraged]:

ngokevin:
Thanks, I think there were some merge conflicts. Will get that fixed.

rong1129:
Regarding the raycaster issue, I looked a little further at it, and this line looks suspicious

https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L358
 
> self.initComponent(component, undefined, true);

It looks like that the raycaster component was initialized as a dependency, and the above line initialized it without its attributes. If that's right, chances are many other components may also be initialized without their attributes.

dmarcos:
Component dependencies don't have attributes. They could but the current API only allows to define the list of dependencies to guarantee an order of initialization. If the component is also defined in the entity we call `updateComponent` with the proper data defined in the DOM attribute:

https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L425

ryanbetts:
Gonna take a look at this right now!

ngokevin:
Thanks Ryan, my bad for regressing it!

PR #1975, opened 2016-10-13
fixed bug #1959 and added a test for it

ryanbetts:
**Description:**

Fixed the issue described in bug #1959 and added a test.


ngokevin:
thanks!

PR #1987, opened 2016-10-18
reverse mouse drag look direction

ngokevin:
**Description:**

Based on feedback from what I've seen in Slack and HackerNews, this is more commonly the intuitive direction.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -239,8 +239,8 @@ module.exports.Component = registerComponent('look-controls', {
     }
     this.previousMouseEvent = event;
 
-    yawObject.rotation.y -= movementX * 0.002;
-    pitchObject.rotation.x -= movementY * 0.002;
+    yawObject.rotation.y += movementX * 0.002;
+    pitchObject.rotation.x += movementY * 0.002;
     pitchObject.rotation.x = Math.max(-PI_2, Math.min(PI_2, pitchObject.rotation.x));
   },
 


dmarcos:
What about in mobile? I don't think there's right or wrong and the feeling of it depends on your background and the application. In an FPS kind of application it makes more sense to drag the mouse the way you want to go. Drag to the left to make the character turn to the left (the way it is right now) instead of grabbing the world (you are controlling the character not the world). If you are not used to 3D applications and looking at a panorama I understand that grabbing the picture might make more sense. People probably expect that panning a panorama should feel like manipulating a 2D picture or a map in your cellphone instead of moving a virtual camera. We could add a flag to the component reverse the direction if people want to do it (I believe there are already flags on the wasd-controls already). Two people have complained in HN. It doesn not mean we're wrong and they're right or the other way around. Remember when Lion changed the scrolling direction in OSX? Did it feel weird to you? Does it feel weird now? @fernandojsg what do you think?

fernandojsg:
A friend of mine also complains that it looked strange while watching 360 images, I honestly didn't notice it and I agree with @dmarcos that it really depends on the kind of experience, so I'll just leave it that way but add a option to use the inverse direction if needed.

ngokevin:
This isn't just a couple comments HN (although it says something when a large chunk of comments talk about this). There's been a lot of comments in Slack over the months from multiple companies about trying it out on new people/clients, and it wasn't the intuitive direction. We can't really force reason on what feels intuitive, it either is or it isn't. The FPS example is also a bit flawed because mouse-controlling the player is only intuitive if it's mouse-locked.

We might not notice it because we've been using A-Frame so much. This is just about deciding on a more sane default because most people won't know which to choose or go through the trouble.

ngokevin:
Large scale production examples:

- [Youtube 360](https://www.youtube.com/watch?v=-xNN-bJQ4vI)
- [iStaging](http://vrviewer.istaging.co/#!/684173)

fernandojsg:
So what about adding reverse by default but including an option to change the direction if needed?

dmarcos:
What were those companies trying to do? First person 3D applications or panorama / 360 video viewers?

What I'm saying is that there's not right or wrong. If we change the default there will be another 50% of users that will complain. In the same HN entry you point out there's this comment:

"There is no consensus on drag direction. I've built several projects and have alternated the drag direction on each and have always received vociferous complaints about both. " by Sean McBeth that has been working and demoing WebVR content for a while.

I think we can just leave the default as is and add a flag to the `mouse-look` component to reverse the axis direction if creators want to do so.

ngokevin:
I don't think real-time vs. panorama doesn't matter in regards to camera rotation on desktop.

I don't think it's 50/50 as Sean says. But if there's no clear right or wrong, why not just switch it to what's commonly used in large-scale today, and leave the current one as an option?

dmarcos:
What you pointed out are 360 video / panorama players not 3d applications. Those are examples of move the world (vs camera / character). Our focus is on the latter. Let's add a flag to the component to let content creators decide based on their user base, background and application and keep the current default.

ngokevin:
I don't believe 3D vs. media player makes a difference in regards to camera rotation. What matters is the action. Click-drag for move the world, pointer-lock for move the camera. In the end, what's on the desktop will still be a 2D display.

ngokevin:
OK, actually I think it is pretty weird in combination with WASD. Hopefully, it's not just because I'm used to A-Frame. Will at least add an option then.

PR #1995, opened 2016-10-23
Update cursor intersection metadata for intersected entities.

### LINKED ISSUE #1994: Cursor intersection metadata does not update over time
When `onIntersection` is called repeatedly for the same entity, the intersection metadata should be updated even though the entity has not changed. Currently we return early:

```js
102:    // Already intersecting this entity.
103:    if (this.intersectedEl === intersectedEl) { return; }
```

As a result, consecutive clicks on the same element may only return the original intersection data.

donmccurdy:
Fixes #1994.


ngokevin:
nice!

PR #1997, opened 2016-10-24
allow opening inspector through postMessage

ngokevin:
**Description:**

For A-Frame site to have a button on each example to pop up the Inspector.


ngokevin:
aframe

PR #1998, opened 2016-10-24
a-asset-item progress (fixes #1937)

### LINKED ISSUE #1937: No export API  to listen loader/request loading progress ?
How to listen e.g. OBJ loader loading progress?
We know all loaders in aframe were using THREE.XHRLoader( url, onLoad, onProgress, onError ), which has 'onProgress' to handle XHR progress event. But in aframe, loaders just call THREE.XHRLoader( url, onLoad) with first two vars. Giving us no API to handle loading progress.

I think we need add some apis to do that.

ngokevin:
**Description:**

`progress` and `loaded` events for XHRs on `<a-asset-item>`.

### DIFF
--- a/docs/core/asset-management-system.md
+++ b/docs/core/asset-management-system.md
@@ -134,9 +134,16 @@ the `loaded` event when they say they have finished loading.
 
 ### `<a-asset-item>`
 
-| Event Name | Description                           |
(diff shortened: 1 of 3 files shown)


ngokevin:
ready

PR #2002, opened 2016-10-24
Add 'position' support for Trackedcontrols

fernandojsg:
**Description:**
When developing the teleport component I noticed that you could set a `position` value for the camera entity when jumping to another position, but if you set a `position` for the hand controllers (to follow your head) it won't be taken into consideration when computing the position as right now this position it's just based on the current controller absolute position.

**Changes proposed:**
- Compute the delta of absolute movement from the controller (from VRPose) and add it to the `position` component value for each hand.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -43,8 +43,10 @@ module.exports.Component = registerComponent('tracked-controls', {
     var controllerEuler = new THREE.Euler();
     var controllerPosition = new THREE.Vector3();
     var controllerQuaternion = new THREE.Quaternion();
+    var deltaControllerPosition = new THREE.Vector3();
     var dolly = new THREE.Object3D();
     var standingMatrix = new THREE.Matrix4();
+    var previousControllerPosition = new THREE.Vector3();
     controllerEuler.order = 'YXZ';
     return function () {
       var controller;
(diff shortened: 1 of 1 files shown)


src/components/tracked-controls.js
```diff
@@ -43,6 +44,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     var controllerEuler = new THREE.Euler();
     var controllerPosition = new THREE.Vector3();
     var controllerQuaternion = new THREE.Quaternion();
+    var deltaControllerPosition = new THREE.Vector3();
```
dmarcos: This has been always kind of crappy. Can we create a closure and initialize and reuse these variables instead instancing once per frame? Like this:

https://github.com/aframevr/aframe/blob/master/src/components/raycaster.js#L142

src/components/tracked-controls.js
```diff
@@ -18,6 +18,7 @@ module.exports.Component = registerComponent('tracked-controls', {
   },
 
   init: function () {
+    this.previousControllerPosition = new THREE.Vector3();
```
dmarcos: Can we move this variable to a closure as I indicated below?

src/components/tracked-controls.js
fernandojsg: They're already in a closure

src/components/tracked-controls.js
fernandojsg: done!

src/components/tracked-controls.js
```diff
@@ -74,10 +76,15 @@ module.exports.Component = registerComponent('tracked-controls', {
         y: THREE.Math.radToDeg(controllerEuler.y),
         z: THREE.Math.radToDeg(controllerEuler.z)
       });
+
+      deltaControllerPosition.copy(controllerPosition).sub(previousControllerPosition);
```
ngokevin: Could use a good ol' unit test

PR #2006, opened 2016-10-25
Add loading feedback while the inspector library is fetched over the network

dmarcos:


src/components/scene/inspector.js
```diff
@@ -12,12 +12,35 @@ module.exports.Component = registerComponent('inspector', {
...
+  initOverlay: function () {
+    this.overlayEl = document.createElement('div');
+    this.overlayEl.style.backgroundColor = '#222';
+    this.overlayEl.style.opacity = '0.90';
+    this.overlayEl.style.position = 'absolute';
+    this.overlayEl.style.width = '800px';
+    this.overlayEl.style.height = '600px';
+    this.overlayEl.style.textAlign = 'left';
+
+    this.loadingMessageEl = document.createElement('div');
```
ngokevin: I'd give it `text-transform: capitalize`, increased font size, and `monospace` font-family.

src/components/scene/inspector.js
```diff
@@ -12,12 +12,35 @@ module.exports.Component = registerComponent('inspector', {
...
     this.onKeydown = bind(this.onKeydown, this);
     this.onMessage = bind(this.onMessage, this);
+    this.initOverlay();
     window.addEventListener('keydown', this.onKeydown);
     window.addEventListener('message', this.onMessage);
   },
 
+  initOverlay: function () {
+    this.overlayEl = document.createElement('div');
+    this.overlayEl.style.backgroundColor = '#222';
```
ngokevin: Could do `setAttribute('style', ...)` here and below to be more compact.

src/components/scene/inspector.js
```diff
@@ -31,6 +54,50 @@ module.exports.Component = registerComponent('inspector', {
...
+      }
+      if (step === 50) {
+        this.loadingMessageEl.innerHTML = message + '..';
+        return;
+      }
+      if (step === 75) {
+        this.loadingMessageEl.innerHTML = message + '...';
+        return;
+      }
+      console.log(step);
```
ngokevin: unneeded log

src/components/scene/inspector.js
```diff
@@ -12,12 +12,35 @@ module.exports.Component = registerComponent('inspector', {
   },
 
   init: function () {
+    this.loadingMessage = 'Loading Inspector';
```
ngokevin: this could just be a constant rather than a property of the component

`var LOADING_MESSAGE`

dmarcos:
If you add a delay to the overlay you won't see immediate feedback from your action (the keyboard shortcut). I like to see something as soon as I press the shortcut. 500ms is a lot of time and it feels slow

dmarcos:
What do you mean by buggy? It works fine on my machine

dmarcos:
I thought forcing the overlay to show up for at least 500ms but considered unfair to slow down the UI if things can go faster.

ngokevin:
You can tweak the value, but less than that, then the immediate feedback is simply the Inspector opening up. On faster connections, it already feels immediate.

fernandojsg:
What about having something in a corner instead of the whole page. I agree with @ngokevin that being on fast connections you can't read anything but looks like something buggy.
So adding it to a corner if it's fast you won't notice, and if it's slow you could see something bright there in the bottom.
Maybe something using similar style with the enter/exit editor? 
image

ngokevin:
I like that. Then don't need to worry about the full-screen flash. Maybe on the top left where the Open Inspector / Back to Scene buttons are. When the scene is open though, a neutral color looks better than the A-Frame red.

dmarcos:
The feedback should also be hard to miss. When pressing the shortcut is important to clearly show that something is going on. The loading feedback is specially important for people that use / discover the inspector for the first time. Something tiny on one corner might pass unnoticed for those users.

fernandojsg:
But I believe users already got used of looking around if something is not working. For example if you click a link and something is not working people could look at the windows clock, or the chrome spinner on the tab. Also when you use apps like googledocs or so, the saving feedback, even if it's taking a while, it's shown on the corners instead of fullscreen. I can't remember now any example of such big screen feedback. Maybe we could make it brighter or bigger, if it's taking some time they'll notice that there's something there, it's not that hard to miss that something completely different from the scene just pop up there

dmarcos:
It's not only that people are used to it. People cope with a lot unconvenient and bad designed things. We should think about what's the least likely design to produce undesired outcomes and the one with lowest risk of confusion. I can try introducing a 100ms delay on the overlay.

dmarcos:
In the embedded case putting the loader in the corner is specially bad.

src/components/scene/inspector.js
dmarcos: Is not constructing the string to set multiple CSS attributes more cumbersome than just setting `style.XXX = YYY`

ngokevin: You could also put it in the A-Frame CSS stylesheet.

dmarcos:
I made the loader similar to the inspector button: 

loader

dmarcos:
I also replaced the JS based animation with CSS

fernandojsg:
I like it!

ngokevin:
🎉

PR #2016, opened 2016-10-27
"var src = src =" should be "var src =" in a-assets.js (fixes #2015)

### LINKED ISSUE #2015: "var src = src =" should be "var src =" in a-assets.js
**Description:**

Hi, I'm new to A-Frame.

I think this line

https://github.com/ngokevin/aframe/blob/5af7916d8460d7f5ecc61fcdcbc9bf6d3afb6f0e/src/core/a-assets.js#L86

    var src = src = this.getAttribute('src');

should be

    var src = this.getAttribute('src');

correct?

If I'm right, I'll make PR to fix.

takahirox:
**Description:**

Fixed typo. #2015

**Changes proposed:**

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -86,7 +86,7 @@ registerElement('a-asset-item', {
     attachedCallback: {
       value: function () {
         var self = this;
-        var src = src = this.getAttribute('src');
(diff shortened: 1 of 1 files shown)


ngokevin:
thanks!

PR #2023, opened 2016-10-28
make xhrloader public from <a-assets> (fixes #2022)

### LINKED ISSUE #2022: expose xhrloader in `<a-assets>`
**Description:**

So we can do `document.querySelector('a-assets').xhrLoader`

ngokevin:


ngokevin:
Need to rename to fileLoader.

dmarcos:
Should we expose this or let people use `THREE.XHRLoader`. We're exporting `THREE` in the global scope anyways.  Mapping `THREE.XHRLoader` to our API could lead to confusion and hides what it really is. Not sure if the slightly more convenient name is worth the increase in API surface.

dmarcos:
Disregard my previous comment. I misunderstood.

PR #2027, opened 2016-10-30
make loadedData public for <a-assets>,  download percentage of xhr+mediaElements

jonalex:
**Description:**
ISSUE:
 I’ve tried to utilize a combination of assets on my scene which has lead me down the path of buffering and optimization in order to achieve a relatively smooth experience.

Specifically, I’ve tried to attach a 300mb .mp4 to a <a-videosphere/>, and even though the rest of the assets don’t present much of an issue, the video will freeze and hit download bumps every so often.

The conclusion: download all assets before starting the scene.

assets.js checks that all mediaElements have a combination of autoplay and preload=“auto” before buffering, but also, there is no way of monitoring the progress of the download of ALL assets (img, a-asset-item, mediaElements).

**Changes proposed:**
The goal is to have a progress bar that feeds from easily-accessible data.

I decided to make a global object that holds individual loaded/total combinations per asset and triggers a function that adds up all these values to estimate a percentage of progress.

To the current structure of assets.js, I added a few lines that do the following:
- I created the global object assetsElement
- within createdCallback in registerElement(‘a-assets’) I added this.loadedData = 0; which makes progress updates easily accessible
- I add any xhr items to assetsElement
- created a new callback to the xhrLoaded which updates values to assetsElement object for any <a-asset-item>
- I removed the validation for the existence of autoplay in mediaElementLoaded(el), instead, it only checks if preload==‘none’
- in function checkProgress() I add new mediaElements to assetsElement object
-  in checkProgress(), after running through the buffer cycle,  assetsElement loaded value is updated for each mediaElement key
- Both in the xhrLoaded and the mediaElementLoaded the function assetsProgress() is used to add up all the loaded and total values of assetsElement, it returns a percentage which is used to update loadedData in the parentNode (<a-assets>).
- Finally, both the xhrLoaded and mediaElementLoaded emit(‘progress’) to the parentNode (<a-assets>)

USAGE:
- preload=“auto” for any media elements (img, video) will add it to the percentage tally. xhr's are added by default.
- the value of loadedData can be grabbed as follow:
  var loadingAssets = document.querySelector('a-assets');
  loadingAssets.addEventListener("progress",function(evt){
  if(this.hasLoaded){
      console.log(this.loadedData)
  }
  });

PROBLEMS:
- The only way to buffer audio files is by adding autoplay, which starts playing all the audio files at once from the beginning.  For now, I’ve left all audios with preload=“none” and no autoplay
- image file sizes should be accessible as well, just to be consistent.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -16,7 +16,11 @@ module.exports = registerElement('a-assets', {
       value: function () {
         this.isAssets = true;
         this.fileLoader = fileLoader;
-        this.loadedData = 0;
+        this.loadedBytes = 0;
+        this.percent = 0;
+        this.progress = {};
+        this.progressInterval = null;
+        this.totalBytes = 0;
         this.timeout = null;
       }
     },
@@ -56,9 +60,19 @@ module.exports = registerElement('a-assets', {
         // Trigger loaded for scene to start rendering.
         Promise.all(loaded).then(bind(this.load, this));
 
+        // Progress checker.
+        this.progressInterval = setInterval(function updateProgress () {
+          self.emit('progress', {
+            percent: self.percent,
+            loadedBytes: self.loadedBytes,
+            totalBytes: self.totalBytes
+          });
+        }, 100);
+
         // Timeout to start loading anyways.
         timeout = parseInt(this.getAttribute('timeout'), 10) || 3000;
         this.timeout = setTimeout(function () {
+          clearInterval(self.progressInterval);
           if (self.hasLoaded) { return; }
           warn('Asset loading timed out in ', timeout, 'ms');
           self.emit('timeout');
@@ -67,9 +81,28 @@ module.exports = registerElement('a-assets', {
       }
     },
 
+    /**
+     * Calculate overall progress as a percentage.
+     */
+    calculateProgressPercent: {
+      value: function () {
+        var loaded = 0;
+        var total = 0;
+        for (var assetId in this.progress) {
+          loaded += this.progress[assetId].loaded;
+          total += this.progress[assetId].total;
+        }
+        this.loadedBytes = loaded;
+        this.totalBytes = total;
+        this.progress = Math.round(loaded * 100 / total);
+        return this.progress;
+      }
+    },
+
     detachedCallback: {
       value: function () {
         if (this.timeout) { clearTimeout(this.timeout); }
+        if (this.progressInterval) { clearInterval(this.progressInterval); }
       }
     },
 
@@ -82,22 +115,6 @@ module.exports = registerElement('a-assets', {
     }
   })
 });
-
-// Lets create the object that will hold asset's loaded/total info.
-var assetsElement = {};
-function assetsProgress () {
-  var assetsLoaded = 0;
-  var assetsTotal = 0;
-  var mediaPercent = 0;
-  for (var key in assetsElement) {
-    assetsLoaded += assetsElement[key].loaded;
-    assetsTotal += assetsElement[key].total;
-  }
-  if (assetsTotal > 0) {
-    mediaPercent = Math.round((assetsLoaded * 100) / assetsTotal);
-  }
-  return mediaPercent;
-}
 
 /**
  * Preload using XHRLoader for any type of asset.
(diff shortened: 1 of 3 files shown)


ngokevin:
Need to `npm run lint`

jonalex:
I fixed the mentioned lines, but still failed Travis.

Trying to figure out what this means:

FAILED TESTS:

  obj-model

    ✖ "before each" hook

      Firefox 49.0.0 (Linux 0.0.0)

    ReferenceError: assetsManager is not defined
(/tmp/src/core/a-assets.js:9:0 <-
/tmp/61855f8826522c2327878f1e193f248a.browserify:57952)

  a-assets

    ✖ does not wait for media element without preload attribute

      Firefox 49.0.0 (Linux 0.0.0)

    timeout of 2000ms exceeded. Ensure the done() callback is being
called in this test.


I've console.dir(xhr) and I swear there is a *xhr.currentTarget.responseURL*
and/or *xhr.srcElement.responseURL* with the complete path of the obj.  Not
sure why its bouncing it out as not defined.

As far as the next error, I have no idea.

src/core/a-assets.js
```diff
@@ -74,6 +75,22 @@ module.exports = registerElement('a-assets', {
...
+	var assetsLoaded = 0;
+	var assetsTotal = 0;
+	var mediaPercent = 0;
+	for (var key in assetsElement){
+		assetsLoaded += assetsElement[key].loaded
+		assetsTotal += assetsElement[key].total
+	}
+	if(assetsTotal>0){
+		mediaPercent = Math.round((assetsLoaded*100)/assetsTotal);
+		if(mediaPercent>=100){clearInterval(assetsManager);};
```
ngokevin: It just means this variable `assetsManager` is not defined

ngokevin:
Looks like we have to bump Node for the lint tests to run well on Travis. But can you do `npm run lint` to fix lint errors?

jonalex:
@ngokevin is there anything I need to do here?

ngokevin:
Yeah, can you run `npm run lint` locally and fix those errors? It also now needs a rebase.

ngokevin:
Thanks for addressing the lint. Working on reviewing this now. I have fetched and rebased this. I'm currently doing some cleanup and refactoring and will get merged/approved shortly.

ngokevin:
How come we are doing `el.getAttribute('preload') === 'none' || !el.hasAttribute('preload')` to check for media element progress. What's the situation on what audio and video need to preload?

Video: needs autoplay
Audio: needs preload

?

jonalex:
When I first removed the autoplay + preload combination I started getting an error related to asset items that didnt have a preload, which is why im treating no-preload tag the same as a "none".  Ideally all media should have a preaload value, but audios are tricky since they don't begin preloading unless autoplay is included.

ngokevin:
I dug into this more. There is an issue. There is no way to get the number of bytes through HTMLMediaElement progress events, only time/duration. Ideally, we'd be able to calculate total size of assets in the scene in bytes. But right now, XHRs give bytes and HTMLMediaElements give seconds. There isn't an easy way to get an accurate percentage or progress.

I think the solution is to not rely on HTMLMediaElement progress, and use XHRs for everything. However, with HTMLMediaElement, I was confident that the browser would cache these types of assets. My question is: is there a way to make the browser cache video/img/audio that are requested first via XHRs and ensure that only one XHR is sent per asset?

cvan:
> There is no way to get the number of bytes through HTMLMediaElement progress events, only time/duration.

Did you check on responses with and without `Content-Length` HTTP response headers?

cvan:
> My question is: is there a way to make the browser cache video/img/audio that are requested first via XHRs and ensure that only one XHR is sent per asset?

It's a hack, but probably the most bulletproof way that comes to mind: I'd suggest monkeypatching three.js' XHR loader.

cvan:
@jonalex let me know if you need help with this; I'm excited about this functionality

jonalex:
That would be great.  I managed to get this working, but after it was
approved on GitHub Kevin dug a bit further and bumped into the same issues
I had found during my work: I can't get a percentage read on
HTMLMediaElements.

This forced me to group these loads in everything XHR and everything that
was media.  The goal of this was to be able to Cache media and minimize
future loads, but that's not happening.

Have you taken a look at the code?  I haven't touched on this for a while
(I jumped on mastering VR in after effects, learning Unity and learning
React)... but it'd be nice to get back on aframe.io

I have a huge interest in getting video to work right on WebVR, and this is
definitely and stepping stone in that direction.

cvan:
Yeah, I've been looking through this code just recently, which I was commented here. I'll follow up at some point here and let you know if/where the assets logic could use some help; stay tuned. Thanks for your help - it's much appreciated.

cvan:
is this close to graduating from WIP? is there anything I can do test or simplify this?

ngokevin:
If you can get it to read byte progress from media elements, be my guest.

ngokevin:
Closing to archive. This code may be useful in the future, but hard to nail this correctly with not being able to read video download progress. Perhaps even VR browsers can handle progress bars automatically. And there's been a lot of changes like the loading-screen that will need rethink of the PR. As many A-Frame scenes load quickly, and large videos should not be 100% preloaded, it is not a huge priority ATM. Thanks though!

Issue #1848, opened 2016-08-24
Order in obj-model & material components affects result


feiss:
I'm trying to use an OBJ model without MTL, and specify the material by hand, using a `material` component. If I put the obj-model first:

``` html
<a-entity obj-model="obj: #my_model" material="color:#ff0000"></a-entity>
```

the material is ignored, and a default material is added (checked with inspector). But if I swap the order and put the material component first:

``` html
<a-entity material="color:#ff0000" obj-model="obj: #my_model" ></a-entity>
```

...the correct material is applied
- A-Frame Version: 0.3.0
- Platform/Device: Chromium + Windows + HTC Vive

ngokevin:
Thanks for catching this.

The `obj-model` component has an initialization dependency on the `material` component. But it seems that it's not being honored?

ngokevin:
fixed

Issue #1841, opened 2016-08-22
Configuring the cursor's raycaster does not work (component ordering/dep bug)


bryik:
**Description:** The documentation says we can ["customize the raycasting pieces of the cursor...by changing the raycaster component properties"](https://aframe.io/docs/0.3.0/components/cursor.html#configuring-the-cursor-through-the-raycaster-component). But attempting to restrict the cursor to a certain class using the raycaster's 'objects' property does not work (all entities remain active).

CodePen: https://codepen.io/bryik/pen/VjRgyk
- A-Frame Version: 0.3.0
- Platform/Device: Chrome/Windows

bryik:
This may be caused by confusion between the raycaster created by the cursor (as a dependency) and the raycaster added in markup.

I had the same problem with the vive-cursor component and "solved" it by having the raycaster's 'objects' property set by the vive-cursor component during initialization.

ngokevin:
I got it working in https://github.com/aframevr/aframe/pull/1845

I wonder if it's a component initialization order issue.

bryik:
>I wonder if it's a component initialization order issue.

Putting the raycaster component before the cursor component [works as it should]( https://codepen.io/bryik/pen/PzAGRa).

The code sample on the cursor's doc page could be amended, but ideally the order wouldn't matter.

ngokevin:
fixed

PR #2036, opened 2016-11-01
fix component dependencies being initialized with no data (fixes #2033, #1841, #1848)

### LINKED ISSUE #2033: Dependent instantiations do not correctly populate data
Its appear that `init()` is getting called with `this.data` not populated when it is early-instantiated via the dependency logic.

`this.data.foo` will be undefined in `init()` for this entity `<a-entity component-that-is-depended-on="foo: bar" component-with-dependency></a-entity>`

https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L358

My guess is that the solution may end up involve retrieving the `attrValue` if `isDependency` is `true` before calling this:
https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L331

ngokevin:
**Description:**

With investigation help from @GavanWilhite.

`initComponent` in `initComponentDependencies` was being called with `undefined` rather than reading data from the DOM. The component later initializes with a null data set because components themselves never fetch data from the DOM; they depend on the entity to pass data to them.

**Changes proposed:**
- Call `initComponent` for dependencies using data from the DOM (`HTMLElement.prototype.getAttribute`).
- Increase confidence of dependency system with tests.
- [x] Specific tests for cursor + raycaster combo.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -369,7 +369,7 @@ var proto = Object.create(ANode.prototype, {
         // Call getAttribute to initialize the data from the DOM.
         self.initComponent(
           componentName,
-          HTMLElement.prototype.getAttribute.call(self, componentName),
+          HTMLElement.prototype.getAttribute.call(self, componentName) || undefined,
           true
         );
       });
@@ -611,7 +611,7 @@ var proto = Object.create(ANode.prototype, {
   setAttribute: {
     value: function (attr, value, componentPropValue) {
       var componentName;
-      var isDebugMode;
+      var isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
 
       componentName = attr.split(MULTIPLE_COMPONENT_DELIMITER)[0];
       if (COMPONENTS[componentName]) {
@@ -622,8 +622,7 @@ var proto = Object.create(ANode.prototype, {
           this.updateComponent(attr, value);
         }
 
-        // On debug mode we write the component state to the DOM attributes.
-        isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
+        // In debug mode, write component data up to the DOM.
         if (isDebugMode) { this.components[attr].flushToDOM(); }
         return;
       }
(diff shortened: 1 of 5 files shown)


src/core/a-entity.js
```diff
@@ -594,16 +610,20 @@ var proto = Object.create(ANode.prototype, {
...
+        // Just update one of the component properties.
         if (typeof value === 'string' && componentPropValue !== undefined) {
           this.updateComponentProperty(attr, value, componentPropValue);
         } else {
           this.updateComponent(attr, value);
         }
-        // On debug mode we write the component state to the DOM attributes
+
+        // On debug mode we write the component state to the DOM attributes.
+        isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
```
dmarcos: I would leave the variables at top. If you want to reuse the same variable somewhere else in the function you have to remember to move it to the top or at least above the other place you want to use it.

src/core/a-entity.js
```diff
@@ -346,16 +346,32 @@ var proto = Object.create(ANode.prototype, {
...
       if (!dependencies) { return; }
-      dependencies.forEach(function (component) {
-        self.initComponent(component, undefined, true);
+
+      // Initialize dependencies.
+      dependencies.forEach(function initializeDependency (componentName) {
+        // Call getAttribute to initialize the data from the DOM.
+        self.initComponent(
+          componentName,
+          HTMLElement.prototype.getAttribute.call(self, componentName),
```
dmarcos: If the dependency is not defined in the entity this will pass a `null` to the `initComponet` function. What are the consequences?

ngokevin: I guess I should do `|| undefined` and add a test case.

ngokevin:
updated

src/core/a-entity.js
dmarcos: At some point (I cannot longer tell you if the case) a null value was interpreted as a component being removed. This is why `undefined` value was passed before.

PR #2041, opened 2016-11-02
add link for device&platform support page to to >find out more: table

rttomlinson:
**Description:**
add link to support page on aframe.io under find out more:
**Changes proposed:**
-
-
-


dmarcos:
Thanks!

PR #2045, opened 2016-11-03
lax the asset parser, remove no longer needed primitives.transform (fixes #2038)

### LINKED ISSUE #2038: Original url('srcPath') string is not maintained in element
**Description:**

I'm writing a A-Frame DOM->JSON scene save/loader.

OBJ-Models that have property "obj: url(assets/asset.obj)" (and mtls) get their sources trimmed. When iterating through the elements, you get something like this:

    Object {mtl: "assets/asset.mtl", obj: "assets/asset.obj"}

When loading this back as an element by doing:
    element.setAttribute( propertyName, propertyObject );

A-frame will throw an error:

    aframe.js:62939 Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': 'assets/BlackCW_Rand_LOD0.mtl' is not a valid selector.

This is because 

1. The url('') part is stripped and

2. This function

    function assetParse (value) {
      var parsedUrl = value.match(/\url\((.+)\)/);
    ...

is too lax with matching url() and when that fails it looks for an element id immediately. 

What I think A-Frame should do is match the url('') part, then if that fails, match the hashtag ('#') part for an existing element, then finally if all else fails, simply load the argument as the url.

Either that, or provide the original, unmodified string with the url('') in-tact. 

- A-Frame Version: Master
- Platform/Device:

ngokevin:
**Description:**

Allows URL properties to be just a string, for a more lenient API. Keep `url()` as a solution to escape semicolons or for data URIs. Lets us remove the primitives transform API.

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -233,13 +233,13 @@ types:
 | Property Type   | Description                                                                                                       | Default Value            |
 | --------------- | -------------                                                                                                     | -------------            |
 | array           | Comma-separated values to array (e.g., `"1, 2, 3" to ['1', '2', '3'].                                             | []                       |
+| asset           | Parse URL out of `url(<url>)`. If a selector, call `querySelector` and `getAttribute('src')`. Can take a string.  | ''                       |
 | boolean         | Convert to boolean (i.e., `"false"` to false, everything else truthy).                                            | false                    |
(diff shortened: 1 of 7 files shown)


ngokevin:
rebased

ngokevin:
Thanks for the reviews 👍

PR #2050, opened 2016-11-03
Fix typo and add clarification

thoragio:
**Description:**
Typo and making a point clearer

**Changes proposed:**
- I did not understand what "block" meant until I read the very end of the page, so adding this to make the point clearer sooner.

### DIFF
--- a/docs/core/asset-management-system.md
+++ b/docs/core/asset-management-system.md
@@ -11,7 +11,7 @@ assets in one place and to preload and cache assets for better performance.
 
 Games and rich 3D experiences traditionally preload many of their assets, such
 as models or textures, before rendering their scenes. This makes sure that
-assets aren't missing visually, and this is benenficial for performance to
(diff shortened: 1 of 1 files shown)


docs/core/asset-management-system.md
```diff
@@ -22,8 +22,8 @@ Assets are placed within `<a-assets>`, and `<a-assets>` is placed within
 - `<img>` - Image textures
 - `<video>` - Video textures
 
-The scene and all entities will block until all of these types of assets are
-fetched (or error out) before playing.
+The scene and all entities will block (i.e., won't render) until all of these
```
dmarcos: I would clarify even more  `(i.e, the scene won't render)` since entities don't render

thoragio: Good point. Thanks for clarifying. I've made the change as you suggested.

PR #2051, opened 2016-11-04
Add runtime env variable to load inspector from master in developmen

dmarcos:
The build commands are getting messy. Ideas to clean them up are more than welcome.

Budo is also messy. I could not pass `-t [envify --BUILD_TYPE development]` as an argument. Whatever parser budo is using splits the string in substrings using the space as separator. It took my a while to understand what was going on. At the end I used `cross-env` to be able to set an environment variable so I can just pass `-t envify` to budo.

### DIFF
--- a/src/components/scene/inspector.js
+++ b/src/components/scene/inspector.js
@@ -4,7 +4,9 @@ var bind = require('../../utils/bind');
 var pkg = require('../../../package');
 var registerComponent = require('../../core/component').registerComponent;
 
-var INSPECTOR_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+var INSPECTOR_DEV_URL = 'https://aframe.io/aframe-inspector/build/aframe-inspector.js';
+var INSPECTOR_RELEASE_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+var INSPECTOR_URL = process.env.INSPECTOR_VERSION === 'dev' ? INSPECTOR_DEV_URL : INSPECTOR_RELEASE_URL;
 var LOADING_MESSAGE = 'Loading Inspector';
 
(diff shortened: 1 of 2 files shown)


src/components/scene/inspector.js
```diff
@@ -4,7 +4,9 @@ var bind = require('../../utils/bind');
 var pkg = require('../../../package');
 var registerComponent = require('../../core/component').registerComponent;
 
-var INSPECTOR_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+var INSPECTOR_MASTER_URL = 'https://aframe.io/aframe-inspector/build/aframe-inspector.js';
+var INSPECTOR_RELEASE_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+var INSPECTOR_URL = process.env.BUILD_TYPE === 'development' ? INSPECTOR_MASTER_URL : INSPECTOR_RELEASE_URL;
```
ngokevin: We should name the environment variable for what it does, pointing to Inspector dev builds. It doesn't have any effect on the overall build type of A-Frame.

ngokevin: Actually, might be fine if we start adding more flags.

src/components/scene/inspector.js
```diff
@@ -4,7 +4,9 @@ var bind = require('../../utils/bind');
 var pkg = require('../../../package');
 var registerComponent = require('../../core/component').registerComponent;
 
-var INSPECTOR_URL = pkg.homepage + 'releases/' + pkg.version + '/aframe-inspector.min.js';
+var INSPECTOR_MASTER_URL = 'https://aframe.io/aframe-inspector/build/aframe-inspector.js';
```
ngokevin: we'll start calling them dev builds `INSPECTOR_DEV_URL`

PR #2052, opened 2016-11-04
Fix syntax around link to minified K-Frame

jpetto:
**Description:**

Just fixes a markdown typo around the link to the minified K-Frame source.

### DIFF
--- a/docs/guides/building-with-components.md
+++ b/docs/guides/building-with-components.md
@@ -121,7 +121,7 @@ event-set. First, we have to include them. [K-Frame][kframe] is a component
 pack by [Kevin Ngo][ngokevin], an A-Frame core developer, that conveniently
 includes all three of these components in one bundle.
 
-To drop in K-Frame, download [`k-frame.min.js`](kmin) from the project's
+To drop in K-Frame, download [`k-frame.min.js`][kmin] from the project's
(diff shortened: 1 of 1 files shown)


dmarcos:
I believe the syntax for a link is already correct. Is not the link working for you?

https://github.com/jpetto/aframe/blob/44a9ca12f66ab4e846cb9992c5d45487a4f71a8e/docs/guides/building-with-components.md

On the markdown docs I see that a link is expressed as `[link to Google!](http://google.com)`

https://guides.github.com/features/mastering-markdown/

jpetto:
The link is currently broken on the [live page](https://aframe.io/docs/0.3.0/guides/building-with-components.html#using-third-party-components). The `k-frame.min.js` link in the **Using Third-Party Components** section points to `https://aframe.io/docs/0.3.0/guides/kmin`, which is a 404. The current/broken markdown is `[k-frame.min.js](kmin)` (`kmin` is not a valid URL).

The majority of the links on that page are using variable substitution in the form of `[displayed text][variable for href]`. If you look at the other links in that same section (and really the whole file), they are in this format. The variables are defined at the bottom of [the markdown file](https://raw.githubusercontent.com/aframevr/aframe/master/docs/guides/building-with-components.md).

dmarcos:
understood! Thanks!

Issue #2047, opened 2016-11-03
consider trimming the registerShader API and prescribe using ShaderMaterial directly for custom glsl shaders


ngokevin:
**Description:**

`registerShader` registers three.js materials for the material component. `registerShader` has a shortcut to instantiating a `ShaderMaterial` via `vertexShader/fragmentShader` and A-Frame-to-three.js-uniform mappings.

My concern is that we are tightly trying to wrap `ShaderMaterial`. This is a large API surface to cover. We have to replicate all the [three.js uniform types](https://github.com/mrdoob/three.js/wiki/Uniforms-types), and support all the `ShaderMaterial` properties, and keep it up to date with three.js.

```
shading — Define shading type. Default is THREE.SmoothShading.
fog — Define whether the material color is affected by global fog settings. Default is true.
wireframe — render geometry as wireframe. Default is false.
wireframeLinewidth — Line thickness. Default is 1.
vertexColors — Define how the vertices gets colored. Default is THREE.NoColors.
skinning — Define whether the material uses skinning. Default is false.
morphTargets — Define whether the material uses morphTargets. Default is false.
```

It seems more intuitive (and more documented) to just use `ShaderMaterial` directly, possibly with helpers to handle the update process. This way you could copy/paste existing three.js shaders without running into the many limitations that the A-Frame wrapper provides.

```
AFRAME.registerShader('myshader', {  // or registerMaterial to map to three.js API
  schema: {},

  init: function () {
    this.material = new ShaderMaterial({
      // ...
    });
  }
});
```

ngokevin:
Will pass on this for now. registerShader does help in handling updates.

PR #2054, opened 2016-11-05
add asset-specific property types: map, model, audio (fixes #1329)

### LINKED ISSUE #1329: Add parameter type for maps
**Description:** Currently the maps parameters are treated as simple strings, eg:

```
    src: { default: '' },
    envMap: { default: '' },
```

We should give them a proper type, `texture` ?, `map` ?

ngokevin:
**Description:**

Add new property types: `map`, `model`, `audio`.

**Changes proposed:**
- Convert some existing components to use the new property types (e.g., `asset` and `map).
- Add them to property type mappings in A-Frame's API for wrapping ShaderMaterial, but I'm convinced we should kill that #2047.
- Move canvas texture loading outside of the funky `src` validation code, have property type return the canvas, and pass it straight to three.js.

### DIFF
--- a/src/components/blend-character-model.js
+++ b/src/components/blend-character-model.js
@@ -6,7 +6,7 @@ var THREE = require('../lib/three');
  * Loads a model with skeletal animation blending.
  */
 module.exports.Component = registerComponent('blend-character-model', {
-  schema: {type: 'src'},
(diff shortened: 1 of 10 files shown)


fernandojsg:
cool! :+1:

PR #2056, opened 2016-11-05
Fix broken link, improve example code

thoragio:
**Description:**

Minor fixes

**Changes proposed:**
- Broken link
- The 1-unit plane entities were overlapping due to insufficient margin in the example code, so upped it to 1.5 units to give them some spacing


ngokevin:
thanks!

PR #2064, opened 2016-11-08
flushToDOM default component values (fixes #1944)

### LINKED ISSUE #1944: Entity.flushToDOM() does not flush default component at default value (e.g, position)
**Description:**

Component.flushToDOM() is working.

ngokevin:
If we do `flushToDOM`, `position=""`  should become `position="0 0 0"`.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -4,9 +4,9 @@ var COMPONENTS = require('./component').components;
 var registerElement = require('./a-register-element').registerElement;
 var THREE = require('../lib/three');
 var utils = require('../utils/');
-var bind = utils.bind;
 
 var AEntity;
+var bind = utils.bind;
 var debug = utils.debug('core:a-entity:debug');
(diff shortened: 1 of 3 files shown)


src/core/a-entity.js
```diff
@@ -634,24 +634,30 @@ var proto = Object.create(ANode.prototype, {
...
       var components = this.components;
-      var children = this.children;
+      var defaultComponents = this.defaultComponents;
       var child;
+      var children = this.children;
       var i;
-      Object.keys(components).forEach(updateDOMAtrribute);
+
+      // Flush entity's components to DOM.
+      Object.keys(components).forEach(function flushComponent (componentName) {
```
dmarcos: Are component dependencies properly flashed now?

ngokevin: Should be, dependency components are part of the entity's components so it'll be picked up.

PR #2068, opened 2016-11-08
[WIP] system share component registration/constructor/prototype (fixes #1947)

### LINKED ISSUE #1947: Make System prototype inherit from Component prototype
**Description:**

I've found cases where we want to be able to update systems (e.g., `enemy` system). Since systems already have schema/play/pause/init/data...with update, they will have the same API as components. We should be able to reuse the prototypes, but maintain the same registration API + system-to-component referencing

ngokevin:
**Description:**

This enables system updates, play, and pause.

**Changes proposed:**
- Reuse code between components and systems.
- Refactor functions out of `AFRAME.registerComponent` to be agnostic enough to use with systems (or other types of modules) as well
- `AFRAME.components[myComponent].Component` -> `AFRAME.components[myComponent].Constructor`...follow this pattern for registerable modules.
- [ ] Have Entity handle System initialization / updates just as it does with components (if the entity is the scene).
- [ ] Update geometry/material constructors to use the new functions for DRY.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -14,14 +14,16 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
 
 /**
  * Entity is a container object that components are plugged into to comprise everything in
- * the scene. In A-Frame, they inherently have position, rotation, and scale.
+ * the scene. In A-Frame, they inherently have position, rotation, and scale. The entity
+ * handles component lifecycle.
  *
  * To be able to take components, the scene element inherits from the entity definition.
+ * Since systems are similar to components, the entity also handles system lifecycle.
  *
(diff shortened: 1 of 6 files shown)


dmarcos:
I'm not sure this is worth it. I actually prefer not to couple systems and components since I expect systems to evolve soon and diverge from components.

ngokevin:
We can still evolve systems on top of this, but the behavior/lifecycle between systems and components are 99% the same. I don't anticipate much evolving in effort to reduce API surface.

ngokevin:
Closed in favor of considering https://github.com/aframevr/aframe/issues/2071, pushing back System improvements to 0.5.0.

PR #2070, opened 2016-11-09
Fix for mixins not caching the values of multiple components (fix #1699)

### LINKED ISSUE #1699: multiple instanced components do not work with mixins
**Description:**

Does not work:

``` html
<a-assets>
  <a-mixin id="sounds" sound__1="..." sound__2="..."></a-mixin
</a-assets>
<a-entity mixin="sounds"></a-entity>
```

dmarcos:


tests/core/a-entity.test.js
```diff
@@ -935,6 +935,18 @@ suite('a-entity', function () {
...
+    test('applies multiple components from mixin', function () {
+      var el = this.el;
+      var mixinId = 'sound';
+      mixinFactory(mixinId, {
+        sound__1: 'url(mysoundfile.mp3)',
+        sound__2: 'url(mysoundfile.mp3)'
+      });
+      el.setAttribute('mixin', mixinId);
+      assert.ok(el.components.sound__1);
+      assert.ok(el.components.sound__1);
```
ngokevin: This should test `sound__2`.

ngokevin: I'd also assert the data.

ngokevin:
r+wc

ngokevin:
Marked Needs Changes since I remember you ran into a bug.

dmarcos:
It now works. The remaining piece was passing the attribute name to `buildData` so the correct attribute is used for multiple components

PR #2072, opened 2016-11-10
[WIP] Initialize primitive mappings with one setAttribute() per component.

### LINKED ISSUE #1900: Components used in primitives are initialized before mappings are applied
**Description:**

Code:

``` javascript
AFRAME.registerPrimitive('a-tube', {
  defaultComponents: {
    tube:           {},
  },
  mappings: {
    path:           'tube.path',
    segments:       'tube.segments',
    radius:         'tube.radius',
    radialSegments: 'tube.radialSegments',
    closed:         'tube.closed'
  }
});

AFRAME.registerComponent('tube', {
  schema: {
    path:           {default: []},
    segments:       {default: 64},
    radius:         {default: 1},
    radialSegments: {default: 8},
    closed:         {default: false}
  },

  init: function () {
    var el = this.el,
        data = this.data,
        material = el.components.material;

    var curve = new THREE.CatmullRomCurve3(data.path.map(function (point) {
      point = point.split(' ');
      return new THREE.Vector3(Number(point[0]), Number(point[1]), Number(point[2]));
    }));
    var geometry = new THREE.TubeGeometry(
      curve, data.segments, data.radius, data.radialSegments, data.closed
    );

    this.mesh = new THREE.Mesh(geometry, material.material);
    this.el.setObject3D('mesh', this.mesh);
  },

  remove: function () {
    if (this.mesh) this.el.removeObject3D('mesh');
  }
});
```

Markup:

``` html
<a-tube path="5 0 5, 5 0 -5, -5 0 -5" radius="0.1" segments="8" closed="true"></a-tube>
```

After some debugging, here's what's happening:
1. [Default properties are applied](https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives.js#L57), and component is instantiated.
2. [The first mapping is applied](https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives.js#L59), and `tube.init()` is called.
3. For each subsequent mapping, `tube.update()` is called.

I could / should implement `tube.update()` of course, but this would result in re-creating the geometry up to six times during scene load, so it doesn't solve the issue.

Related: https://github.com/donmccurdy/aframe-extras/issues/80, https://github.com/donmccurdy/aframe-extras/issues/85 (and possibly #1609 ?).
- A-Frame Version: v0.3.1

donmccurdy:
Fixes #1900.

This is orthogonal to the `setAttribute` API changes @ngokevin suggests, although that sounds valuable independently. Tests aren't currently passing, and I'm not sure if this is too risky for v0.4.0, but here's the work in progress.

### DIFF
--- a/src/extras/primitives/primitives.js
+++ b/src/extras/primitives/primitives.js
@@ -5,6 +5,7 @@ var utils = require('../../utils/');
 
 var bind = utils.bind;
 var debug = utils.debug;
+var getComponentPropertyPath = utils.entity.getComponentPropertyPath;
 var setComponentProperty = utils.entity.setComponentProperty;
 var log = debug('extras:primitives:debug');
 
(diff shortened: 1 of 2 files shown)


donmccurdy:
Never mind, nothing to see here.

jetpack

PR #2075, opened 2016-11-10
add events for set/removeObject3D (fixes #2061)

### LINKED ISSUE #2061: have setObject3D emit an event
**Description:**

To listen for when things are attached.

ngokevin:
**Description:**

**Changes proposed:**
- `object3dset` and `object3dremove` events.
- Enforce `setObject3D` to pass a valid instance of `THREE.Object3D`.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -167,27 +167,56 @@ var proto = Object.create(ANode.prototype, {
     }
   },
 
+  /**
+   * Set a THREE.Object3D into the map.
+   *
+   * @param {string} type - Developer-set name of the type of object, will be unique per type.
+   * @param {object} obj - A THREE.Object3D.
+   */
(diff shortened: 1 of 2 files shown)


src/core/a-entity.js
```diff
@@ -167,36 +167,61 @@ var proto = Object.create(ANode.prototype, {
...
   },
 
+  /**
+   * Remove object from scene and entity object3D map.
+   */
   removeObject3D: {
     value: function (type) {
-      this.setObject3D(type, null);
+      var obj = this.getObject3D(type);
+      this.object3D.remove(obj);
```
dmarcos: What happens if `removeObject3D` is called with a `type` that does not exist? I think the event should not be emitted.

ngokevin:
Updated.

PR #2082, opened 2016-11-10
rename master builds to aframe-master* (fixes #2007)

### LINKED ISSUE #2007: rename aframe master builds in dist
**Description:**

In `dist/` folder, I think everyone doesn't see `aframe.min.js` are the unstable builds.

Rename to like `aframe-master.min.js`

ngokevin:
- `aframe*` to `aframe-master*`
- Release script to automatically copy/rename the master builds to new stable version as well as update documentation and README.


dmarcos:
Should it be `aframe-dev` or `aframe-master`? That way is it's easy to know where they can from

ngokevin:
I initially thought `master` in the issue, but I thought `dev` would be more clear on what it was in terms of instability. "dev build" vs. "master build".

dmarcos:
@fernandojsg @feiss what do you think? Using `aframe-dev.js` or `aframe-master.js`. One indicates the development nature of one of the builds and the other the branch where it was generated from.

dmarcos:
Can we automate the generation of these builds per commit?

ngokevin:
Yeah. We could do it nightly maybe? If so, that opens up `aframe-nightly`.

dmarcos:
My preference is `aframe-master.js` , always in sync with the latest commit. I believe is clear enough since all github repos have a `master` branch.

ngokevin:
I don't have a strong preference. When I link people to it, I'll be linking to a heading in the README that'll explain it.

dmarcos:
If we use `master` it will also be consistent with the release versions `aframe-v0.4.0` and the git tags `v0.4.0`. You can just do `git checkout build-sufix` to see the code

ngokevin:
Updated

ngokevin:
Test Slack bot

ngokevin:
test slack bot

Issue #2069, opened 2016-11-09
Enhancement: Add Matrix4 Property Type


james-h-stephens:
Addition of the Matrix4 property type for creation of Matrix4 objects to assist in the creation of quaternions.

donmccurdy:
@james-h-stephens working on this, but also there is a `vec4` property type that [works well for quaternions](https://github.com/donmccurdy/aframe-physics-system/blob/master/src/components/math/quaternion.js).

foobar404:
@dmarcos can we close this? Seems like its already been completed twice.

PR #2105, opened 2016-11-16
Add Matrix3/4 property type

takahirox:
Hi, I added Matrix4 property type #2069
Would you review the code because I'm still new to A-Frame.
LMK if there's any issues.

**Description:**

Add 'matrix4' property type.

Default value is
[1, 0, 0, 0,
 0, 1, 0, 0,
 0, 0, 1, 0,
 0, 0, 0, 1]

Input value is assumed as space separated strings.
ex: '0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15'

If elements# is greater than 16, it'll be forced to be 16.
If elements# is less than 16, the shortage will be filled with default value.

**Changes proposed:**

### DIFF
--- a/tests/core/propertyTypes.test.js
+++ b/tests/core/propertyTypes.test.js
@@ -156,26 +156,6 @@ suite('propertyTypes', function () {
     });
   });
 
-  suite('matrix4', function () {
-    var parse = propertyTypes.matrix4.parse;
-    var stringify = propertyTypes.matrix4.stringify;
-
-    test('parses matrix4', function () {
-      assert.deepEqual(parse(''), [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
-      assert.deepEqual(parse({}), [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
-      assert.deepEqual(parse([]), [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
-      assert.deepEqual(parse('0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15'), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
-      assert.deepEqual(parse('0 1 2 3'), [0, 1, 2, 3, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
-      assert.deepEqual(parse('0   1  2 3 4 5   6 7 8 9 10   11  12 13 14 15   16 17 18    19 20'), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
-      assert.deepEqual(parse([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
-    });
-
-    test('stringifies matrix4', function () {
-      assert.equal(stringify([]), '');
-      assert.equal(stringify([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), '0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15');
-    });
-  });
-
   suite('src', function () {
     var parse = propertyTypes.src.parse;
 
--- a/tests/core/propertyTypes.test.js
+++ b/tests/core/propertyTypes.test.js
@@ -172,7 +172,7 @@ suite('propertyTypes', function () {
 
     test('stringifies matrix4', function () {
       assert.equal(stringify([]), '');
-      assert.equal(stringify([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), '0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15');
+      assert.equal(stringify([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), '0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15');
     });
   });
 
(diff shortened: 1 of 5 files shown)


ngokevin:
@takahiro Thanks for this! 

@james-h-stephens To get a better picture, can you give a specific use case for this?

Need to think because this property type would probably only be set by other components since it's awkward to hand-code it.

tests/core/propertyTypes.test.js
```diff
@@ -156,6 +156,26 @@ suite('propertyTypes', function () {
...
+      assert.deepEqual(parse([]), [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
+      assert.deepEqual(parse('0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15'), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+      assert.deepEqual(parse('0 1 2 3'), [0, 1, 2, 3, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
+      assert.deepEqual(parse('0   1  2 3 4 5   6 7 8 9 10   11  12 13 14 15   16 17 18    19 20'), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+      assert.deepEqual(parse([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+    });
+
+    test('stringifies matrix4', function () {
+      assert.equal(stringify([]), '');
+      assert.equal(stringify([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]), '0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15');
```
ngokevin: I'd expect no commas in the stringification.

tests/core/propertyTypes.test.js
```diff
@@ -156,6 +156,26 @@ suite('propertyTypes', function () {
     });
   });
 
+  suite('matrix4', function () {
+    var parse = propertyTypes.matrix4.parse;
+    var stringify = propertyTypes.matrix4.stringify;
+
+    test('parses matrix4', function () {
```
ngokevin: I would test being able to pass a `THREE.Matrix4` or even a `THREE.Quaternion` and hopefully it would still work. I think perhaps it should always resolve a three.js object rather than an array.

tests/core/propertyTypes.test.js
takahirox: Thanks, I've updated.

fernandojsg:
The rightmost column of affine 4x4 matrix is always [0,0,0,1] that's why some engines uses 4x3 matrices instead of 4x4, maybe we could simplify this one using just 4x3 (and then it will be parsed still to a THREE.Matrix4) but without explicitly define the latest column?

donmccurdy:
Should the property be explicit about column-major vs row-major format? The latter is more readable to me, but the former is required for `THREE.Matrix4.fromArray(...)`.

From an ergonomics perspective, I think I like this syntax:

```
<a-entity my-matrix="1 2 3 4,
                     5 6 7 8,
                     9 10 11 12,
                     13 14 15 16">
</a-entity>
```

.... but the output should be a flat array, regardless, as it is here.

tests/core/propertyTypes.test.js
donmccurdy: @ngokevin resolving a THREE.js object sounds handy, but is inconsistent with the vector types. Compatibility with THREE.Matrix4.prototype.toArray() and fromArray() seems critical.

(that said, it would be neat for all vectors and matrices to resolve as THREE types, but outside scope here)

donmccurdy:
I screwed up and started a duplicate PR, #2113. If we're still undecided on syntax here feel free to compare/contrast.

takahirox:
I prefer row-major format as user input because easier to read for A-Frame users.
They can easily edit in their HTML.

donmccurdy:
Agreed. Summary of my vote:

Serialized / user input: row-major, either `n1 n2 n3 n4` or `n1 n2, n3 n4`.
Deserialized / output: Either column-major array, or THREE.Matrix4.

takahirox:
Is it expected that A-Frame users call `stringify` method?
If so, I prefer row-major format for deserialization
because row-major input while column-major output(stringify) would be confusing users.

takahirox:
I've updated.

- export as `utils/matrices.js` like `utils/coordinates.js`
- add `Matrix3`
- type names are `mat3/mat4` like `vec2/vec3/vec4`
- End user sees matrix as row-major format while it's handled as column-major in A-Frame/Three.js
- parse(): row-major space separated string input and column-major array output
- stringify(): column-major array input and row-major space separated string output

cvan:
> what type/format of parse() and stringify() input/output/defaultValue should be

any resolution on this?

donmccurdy:
I'm happy with @takahirox's implementation as it stands:

- in serialized form: row-major space-separated string.
- in deserialized form: column-major array.

I do realize that this preference sounds inconsistent. IMO the deserialized form _must_ be either column-major or a THREE.MatrixX — those are the two forms that will play well with three.js. Row-major is nice for readability, but if we think readability of matrices in HTML is less critical than API consistency, we could change the serialized form to column-major as well.

dmarcos:
This needs rebase

dmarcos:
@takahirox what use cases do you have in mind for this type?

takahirox:
About rebase, ok. 
Regarding use cases, for example rotation matrix.

dmarcos:
Not sure what to do with this PR. No complains about code or the API. My issue is that I struggle to find good use cases for the matrix type. I also think that the components APIs are a layer on top of three and passing matrix values around in my view breaks the level of abstraction. I propose shelving / archiving this PR until we find a legitimate use case. What do you think @donmccurdy @ngokevin @fernandojsg ?

donmccurdy:
@dmarcos I don't think it breaks the abstraction — reasonable that someone might want to encapsulate scale/position/rotation, and matrices are the appropriate primitive for that — but I don't have an immediate use case, either. We've done OK without even an official `quaternion` component, so maybe it is best to table this.

I'm certainly OK with re-opening this PR in the future though.

dmarcos:
As per prior conversations. I'm archiving this for future reference.

PR #2106, opened 2016-11-16
hook primitives into the core entity init process (fixes #2104)

### LINKED ISSUE #2104: Dependency not satisfied during primitive initialization
**Description:**

- A-Frame Version: master

```js
AFRAME.registerComponent('shadow', {
  dependencies: ['geometry'],

  schema: {
    cast:     { default: false },
    receive:  { default: false }
  },

  init: function () {
    this.el.addEventListener('model-loaded', this.update.bind(this));
  },

  update: function () {
    var data = this.data;

    // BUG: this.el.object3D has no children here, but should.

    // Applied recursively to support imported models.
    this.el.object3D.traverse(function(node) {
      if (node instanceof THREE.Mesh) {
        node.castShadow = data.cast;
        node.receiveShadow = data.receive;
      }
    });
  },

  remove: function () {}
});
```

```html
<a-plane position="0 0 0" rotation="-90 0 0" width="32" height="32" color="#ffffff " shadow="receive: true"></a-plane>
```

ngokevin:
**Description:**

The root issue was the dependencies were not being initialized for primitives because primitives were trying to do `setAttribute` on `attachedCallback`, but at that time `Entity.load` had not been called yet, so dependencies were being blocked from initializing on `el.hasLoaded`.

I initially tried keep them more-or-less separate and give primitives their own `load` method, but it  got messy fast and it became much easier to have the `Entity` prototype have knowledge of the extra data supplied by primitives (through `defaultComponents` and `mappings`). Thus I made primitives implement a method called `getInitialComponents` which the Entity can pull from. 

As an extra bonus, this also means primitives no longer have to duplicate some of the Entity data building process. Primitives only have to worry about default components + mappings, and the entity handles the other data sources.

I made `getInitialComponents` a method rather than having primitives set a property because when an instance of a primitive was setting the property, it would overwrite that property used by other instances sharing the same prototype.

**Changes proposed:**
- Primitives implement `getInitialComponents`.
- `Entity.updateComponents` includes `getInitialComponents`, merging it on the level of DOM components

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -287,12 +287,10 @@ var proto = Object.create(ANode.prototype, {
 
       if (this.hasLoaded) { return; }
 
-      ANode.prototype.load.call(this, entityLoadCallback);
-      // Entity load.
-      function entityLoadCallback () {
+      ANode.prototype.load.call(this, function entityLoadCallback () {
         self.updateComponents();
         if (self.isScene || self.parentEl.isPlaying) { self.play(); }
-      }
+      });
     },
     writable: window.debug
   },
@@ -341,13 +339,16 @@ var proto = Object.create(ANode.prototype, {
       var componentId = componentInfo[1];
       var componentName = componentInfo[0];
       var isComponentDefined = checkComponentDefined(this, attrName) || data !== undefined;
-      // Check if component is registered and whether component should be initialized.
-      if (!COMPONENTS[componentName] ||
-          (!isComponentDefined && !isDependency) ||
-          // If component already initialized.
-          (attrName in this.components)) {
-        return;
-      }
+
+      // Not a registered component.
+      if (!COMPONENTS[componentName]) { return; }
+
+      // Component is not a dependency and is undefined.
+      // If a component is a dependency, then it is okay to have no data.
+      if (!isComponentDefined && !isDependency) { return; }
+
+      // Component already initialized.
+      if (attrName in this.components) { return; }
 
       // Initialize dependencies first
       this.initComponentDependencies(componentName);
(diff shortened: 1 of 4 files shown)


src/core/a-entity.js
```diff
@@ -449,28 +462,41 @@ var proto = Object.create(ANode.prototype, {
       // Gather mixin-defined components.
       getMixedInComponents(this).forEach(addComponent);
 
-      // Set default components.
-      Object.keys(this.defaultComponents).forEach(updateComponent);
+      // Initialze or update default components first.
```
donmccurdy: s/Initialze/Initialize

dmarcos:
I would try to avoid adding extra complexity on `a-entity` to fix primitives. I need to think about this and see if there's another way to solve this.

dmarcos:
I'm not worried about duplicated code in primitives.

ngokevin:
Primitives need a way to hook in early into the component initialization process or else it's extremely difficult to maintain. I've refactored primitives over and over but there's always been issues, I'd like to just end it with a tiny hook and not have to maintain it. The hook isn't primitive-specific code and won't affect existing entities.

dmarcos:
I spent a couple of hours trying to mixin the primitive `defaultComponentsFromPrimitive` into `defaultComponents`. I have the hunch that might be the cleaner way to fix this. In the process I discovered an inconsistency in the system. Values in the default components are not honored. For instance if you do:
````javascript
var proto = Object.create(ANode.prototype, {
  defaultComponents: {
    value: {
      position: '2 2 2',
      rotation: '',
      scale: '',
      visible: ''
    }
  },
````

donmccurdy:
@dmarcos that would need to be `position: {x: 2, y: 2, z: 2}` — I patched this into one of my examples and default position still works.

PR #2107, opened 2016-11-17
fix primitive bugs: missing material mappings and failed object cloning

ngokevin:
**Description:**

**Changes proposed:**
- Standard material props were in the mapping, but not core material props like `side`.
- Fix primitive not cloning the default components properly causing primitives to affect each other.


ngokevin:
Going to move to https://github.com/aframevr/aframe/pull/2106

PR #2108, opened 2016-11-17
Copy default values for types when assigned to properties so they're …

dmarcos:


src/utils/coordinates.js
```diff
@@ -1,5 +1,5 @@
 /* global THREE */
-
+var extend = require('object-assign');
```
ngokevin: Should probably move all the object utils (extend, deep extend, diff, clone) to a separate module in the future.

tests/core/component.test.js
```diff
@@ -129,6 +129,22 @@ suite('Component', function () {
       data = buildData(el, 'dummy', 'dummy', TestComponent.prototype.schema, 'red');
       assert.equal(data, 'red');
     });
+
+    test('multiple vec3 properties do not share same default value object', function () {
+      var data;
+      var el = entityFactory();
+      var TestComponent = registerComponent('dummy', {
```
ngokevin: delete the component at the end of the test.

tests/core/component.test.js
```diff
@@ -129,6 +129,22 @@ suite('Component', function () {
       data = buildData(el, 'dummy', 'dummy', TestComponent.prototype.schema, 'red');
       assert.equal(data, 'red');
     });
+
+    test('multiple vec3 properties do not share same default value object', function () {
```
ngokevin: you need to specify `done` argument here since it has asyncrony

tests/core/component.test.js
```diff
@@ -129,6 +129,22 @@ suite('Component', function () {
...
+      var data;
+      var el = entityFactory();
+      var TestComponent = registerComponent('dummy', {
+        direction: { type: 'vec3' },
+        position: { type: 'vec3' }
+      });
+      el.addEventListener('loaded', function () {
```
ngokevin: this test is not actually doing anything because it is asynchronous and not using the `done` handler

tests/core/component.test.js
```diff
@@ -129,6 +129,22 @@ suite('Component', function () {
       data = buildData(el, 'dummy', 'dummy', TestComponent.prototype.schema, 'red');
       assert.equal(data, 'red');
     });
+
+    test('multiple vec3 properties do not share same default value object', function () {
+      var data;
+      var el = entityFactory();
+      var TestComponent = registerComponent('dummy', {
+        direction: { type: 'vec3' },
```
ngokevin: nit: no spaces

tests/core/component.test.js
```diff
@@ -129,6 +129,22 @@ suite('Component', function () {
...
+        position: { type: 'vec3' }
+      });
+      el.addEventListener('loaded', function () {
+        el.setAttribute('dummy', '');
+        data = el.getAttribute('dummy');
+        assert.deepEqual(data.direction, TestComponent.prototype.schema.direction.default);
```
ngokevin: nit: could do `shallowDeepEqual`

src/utils/coordinates.js
ngokevin: I'm doing something similar in another PR. A `cloneIfObject` util would be helpful

tests/core/component.test.js
dmarcos: it's already deleted in the `setup` function

src/utils/coordinates.js
dmarcos: We can factor it out later when needed.

ngokevin:
r+

i can handle the `cloneIfObject` util

PR #2120, opened 2016-11-21
Fix material update referencing the sceneEl when it has not loaded

dmarcos:
This bug was unveiled by the inspector. The material update method was triggering a `sceneEl.addBehaviour(self)`. If the entity has not loaded there's no reference to `sceneEl`. It turns that that call was also unnecessary since the `play` / `pause` methods will take care of adding / removing the behavior when appropriate. I made sure that the sky shader example still works since it uses a property with `time` type. This is an example that reproduces the bug bug

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -57,7 +57,6 @@ module.exports.Component = registerComponent('material', {
   },
 
   updateBehavior: function () {
-    var scene = this.el.sceneEl;
     var schema = this.schema;
     var self = this;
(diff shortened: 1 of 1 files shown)


ngokevin:
Can we reproduce in a test?

ngokevin:
r+

dmarcos:
This was not finished. It has a bug

ngokevin:
OK, you can keep the label up-to-date to WIP

PR #2124, opened 2016-11-22
support Oculus Touch in Chromium builds, along with Vive

machenmusik:
**Description:**
support Oculus Touch in Chromium builds, along with Vive

**Changes proposed:**
- added rotationOffset in tracked-controls to allow adjustment of hand model rotation
- modify vive-controls to understand both Oculus Touch and OpenVR Gamepad, and pass rotationOffset
- modify hand-controls to pass per-hand rotationOffset through vive-controls to tracked-controls


ngokevin:
Do you have this tested with an Oculus Touch?

ngokevin:
Don't need to commit the dists.

machenmusik:
tested with Touch, kfarr confirmed I believe.  also tested with Vive wands.

machenmusik:
original intent was to make hand-controls just work.  but hand-controls is hardcoded to use vive-controls which uses tracked-controls.  maybe what is really needed is a generic modeled-tracked-controls or something, and then vive-controls and oculus-touch-controls could have their own models just like hand-controls.  but that seemed a larger refactor than needed to just get things working... especially assuming that many folks just want to use hand-controls to abstract left and right hand without having to know which kind...

src/components/vive-controls.js
```diff
@@ -66,9 +67,34 @@ module.exports.Component = registerComponent('vive-controls', {
...
     // handId: 0 - right, 1 - left
     var controller = data.hand === 'right' ? 0 : 1;
-    el.setAttribute('tracked-controls', 'controller', controller);
+
+    // interrogate gamepads ourselves to see what we need to specify
+    // (to do a priori, we need to figure out whether to use Vive or Rift)
+    var gpads = navigator.getGamepads && navigator.getGamepads();
+    if (typeof gpads !== 'undefined') {
+      for (var i = 0; i < gpads.length; i++) {
+        if (gpads[i].id.indexOf('Oculus Touch') === 0) {
```
machenmusik: the original intent was to make hand-controls left and right hands just work whether Oculus Touch or Vive wands.  internally, hand-controls uses vive-controls, which uses tracked-controls.

maybe what is really needed is to refactor vive-controls into modeled-tracked-controls, and then vive-controls, hand-controls and oculus-touch-controls could all leverage that -- but those are much larger changes (and maybe not backward compatible) so thought getting things functional first might be wise.

if what you are saying is that tracked-controls should be doing what vive-controls / modeled-tracked-controls does, that could work as well, but I assumed there was an explicit decision to keep tracked-controls more minimal.  (may want to remove the default to OpenVR Gamepad in there, since the code will already filter out gamepads without poses)

ngokevin:
Thanks. That will require some thought on how to untangle and handle different controllers gracefully. Not too worried on getting it out too soon since we are about to ship a release, and Touch controllers are still rolling out. Here is what we have now:

```
hand-controls
<adds hand model and uses events from>
vive-controls
<adds button mappings and uses tracking from>
tracked-controls
```

What do we need? Maybe:

```
hand-controls, vive-controls, oculus-controls, X-model-controls
tracked-controls: handles both tracking + button mappings?
```

But @dmarcos and @fernandojsg have played with controls more than me.

machenmusik:
my thoughts were:

**tracked-controls** 
... stays as low-level tracking 
... suggest the filtering to OpenVR Gamepad be removed as previously mentioned
... should we consider calling this tracked-entities or similar, since they may not necessarily be controllers?
**modeled-tracked-controls**
... does most of what what vive-controls does now, but no model (or maybe generic model)
... button mapping, determination of which controllers to bind for specified hand
... maybe modeled-controllers would be a better name
**hand-controls**, **vive-controls**, **oculus-touch-controls**
... add specific models and gamepad id filtering (e.g. vive = OpenVR Gamepad, oculus-touch = Touch naming)

ngokevin:
Is there a need to make `modeled-tracked-controls` separate? We can assume controllers will always have models + mappings.

machenmusik:
perhaps not, but that makes `tracked-controls` fatter and may inflict upon it model-specific filtering and button mapping -- at the moment tracking is uniform regardless of controller vendor, but button/event mapping may not be... so in theory `modeled-controllers` could use `tracked-controls` generically for positioning, but then inject model-specific components for button/event mapping based on gamepad id pattern?
then again, I may be overthinking it...

ngokevin:
I see, maybe it's OK to just duplicate the mappings between the uppermost layer? We can expose common mappings as a constant. As opposed to what I think you are proposing, a thin intermediary component just to handle mappings.

machenmusik:
thoughts on https://github.com/chenzlabs/aframe/tree/refactor-hand-controls
(still WIP) ?

tracked-controls

modeled-controllers ... looks at available controllers and inject mapping
- vive-controls ... has mapping and model, matches specified hand and injects tracked-controls
- oculus-touch-controls ... has mapping and model, matches specified hand and injects tracked-controls

hand-controls ... uses modeled-controllers, which is agnostic

that way, oculus-touch-controls and vive-controls can implement events both common and unique, they select specific controllers to use according to their logic; pattern for others is clear; one can specifically request touch and/or vive as well

dmarcos:
I would start simple and just add `oculus-controls` and a logic to `hand-controls` to select either `oculus-controls` or `vive-controls` at runtime.

machenmusik:
@dmarcos I didn't want to bind that up with the hand model in hand-controls, thus the intermediate layer modeled-controllers.

https://github.com/chenzlabs/aframe/tree/refactor-hand-controls 
works for Touch on Chromium, any chance you can test on Vive?  happy to close this and do a pull request for that one if desired, although Touch needs diifferent models...

dmarcos:
I don't think `modeled-controls` is an abstraction we need. What about adding a `modelSrc` property to `tracked-controls`?

dmarcos:
We would have `tracked-controls` that optionally loads a model through a `modelSrc` property and then `oculus-controls` and `vive-controls` will manage button mappings, controller specific events and animations / highlights related to the model.  `oculus-controls` and `vive-controls` will use `tracked-controls` , set the appropriate `modelSrc` and pass the corresponding string for the gamepad id `OpenVR Gamepad` \ `Oculus Touch`. You could add both to an entity if you want to support both Touch and Vive:

`<a-entity oculus-controls vive-controls></a-entity>`

If either the Oculus or Vive gamepad is not present then the corresponding controller will have no effect on the entity. 

More abstract controls like `hand-controls` will set both `oculus-controls` and `vive-controls` on the entity (and maybe `daydream-controls` in the future) , set the `model` property to false and map the buttons to the specific high level actions: pointing, grip, grab...

machenmusik:
ok you prefer a heavier tracked-controls, fine but I don't think having tracked-controls do modelSrc will suffice given that there are different types of models that could be used etc. (actually i am having similar discussion that video cannot be fully specified only looking at src) - in the refactor I was working on, that seems relatively close to what you describe except that you prefer the modeled-controllers functionality to be merged into tracked-controls; in the meantime I will close this pull request in favor of the refactor.  thanks!

dmarcos:
`tracked-controls` won't be much heavier. It would just handle the model loading.  If you have exotic model formats or want to make something procedural you could always handle that yourself in your higher level component: `hand-controls`,  `magic-wand-controls` ...

An alternative is to leave `oculus-controls` and `vive-controls` to handle the model loading. That would mean some potential code duplication but everything is simple and clear

machenmusik:
actually though, I disagree that authors should have to enumerate all the controller types they could possibly support, including ones that don' exist yet, the framework should handle that internally unless specified IMO... it should be possible for the author to simply say that want hand-controls left and right, and in the refactor, hand-controls does that by using modeled-controllers; the author is free to explicitly use vive-controls or oculus-controls independently.

machenmusik:
the alternative you suggest is basically the current refactor, with modeled-controllers basically meaning "platform please detect and inject for me" - so to add autodetect of daydream-controls, one would simply add to prefixMapping in modeled-controllers which would detect the controller via gamepad extensions id prefix

(i will submit the refactor as pull request soon so we can transition this discussion)

dmarcos:
Authors don't have to enumerate the controllers manually. For instance `hand-controls` (and other high level controls schema) would do the instantiate and configure them in the init method:
``` javascript
el.setAttribute('vive-controls' { ... });
el.setAttribute('oculus-controls' { ... });
````
The capabilities of each controller are very different. Daydream is only 3DOF, Vive has a touchpad, Touch has two analog joysticks and two triggers. Each application and type of controls should decide what input to use and how integrate it. Integration is not straight forward and there's going to be cases where you cannot simply support all inputs for your interaction model. You cannot for instance do a-painter with a Daydream controlller or there might be an application that uses the two analog triggeres of the Oculus Touch in a way that won't work in a Vive controller. We should favor highly reusable non opinionated components in the core and let 3rd parties to compete on the higher levels of abstraction.

machenmusik:
(Further discussion appears in #2132.)

Issue #2126, opened 2016-11-22
Video autoplays and loops


paddeee:
Adding `<a-video>` or `<a-videosphere>` primitives in the latest master version results in the video playing automatically and looping with seemingly no way to override this behaviour.

- A-Frame Version: Master
- Platform/Device: Windows 10/ Chromium

`<a-scene>
    <a-assets>
        <video src="assets/video/test.webm" preload="auto" id="who"></video>
    </a-assets>
    <a-video id="Video"
             src="#who"
             height="0.6"
             width="1.2"
             position="0 2 -1">
    </a-video>
</a-scene>`

machenmusik:
can you see if this is fixed by https://github.com/aframevr/aframe/pull/2129 ?

paddeee:
Thanks machenmusik. 

Yes, can confirm that this has fixed the issue. Works as expected again.

PR #2129, opened 2016-11-23
if we get a video element, pass it all the way through like canvas

machenmusik:


machenmusik:
(this fixes regression that broke https://www.w3.org/TR/media-source/ for video playback)

ngokevin:
This might be a breaking change. Components might be expecting a string. If we do this for video, we should maybe be creating video elements if a string is passed, and same for audio and images, and to also inject to a-assets so it's referencable. Will need to look into that

machenmusik:
that is what I am doing manually, you may not be able to correctly create the video element from just src.  my understanding is that when a string is specified, it will not have a tagName attribute and so will not be passed as either a canvas or video element directly.  am i missing something?

machenmusik:
(for clarity, the existing a-frame examples work fine with this change, and so do MSE videos)

ngokevin:
I'm worried that previously components always accepted a string, now they can sometimes get a string or an element. I think I'd like to ensure that an element is always received if we do this to keep things consistent:

1. If it is a selector, just return the element. Ensure it has a `src` set.
2. If it is a URL, then maybe we create a media element for them, setting the src, ensure `crossorigin` and `playsinline`. And we append it to `<a-assets>` so they have a reference to it if they want to control media playback.

This PR does look like it fixes things without breaking any official examples, but I think I'd like to resolve it completely. For your case, you could work with an A-Frame fork for a while. Then once it's committed, we'll start having auto-generated master builds you can use.

jesstelford:
I just came across this same issue - completely breaks [`aframe-video-billboard`](https://github.com/jesstelford/aframe-video-billboard) component (I opened a duplicate PR in #2141).

> previously components always accepted a string

Which components are you referring to? The only place that this is used is within the `material` component, and before #2054, the logic was the same as what @machenmusik has provided above (but in a different place).

> Ensure it has a src set.

What about [when it's `srcObject`](https://github.com/jesstelford/aframe-video-billboard/blob/d985dda5fef90a12662274494ad8dc8e6093ea73/src/component.js#L175)? The aframe code is simplified by skipping those extra checks and passing it on straight to three.js to let it handle the specifics of using a `<video>` element as a texture.

> If it is a URL, then maybe [...]

Could that be a future enhancement / separate PR? This PR is good to fix the regression, and what you're suggesting is a new feature to add. (my personal use-case just needs the regression fixed on `master` - I'm using a fork with this PR applied in the meantime)

machenmusik:
@ngokevin I think the difference of opinion we have is your assertion that the previous behavior always returned string, which @jesstelford described above.  really what happened is #2054 broke video cases other than trivial examples, and we'd like to restore the original (working) behavior for video elements as part of 0.4.0 release while further steps are discussed.

machenmusik:
for those not aware, it is common video player practice to attach the player logic (event listeners etc.) to the provided video element, so creating a different video element with same src may not yield the same behaviors (i.e. may not play properly) - the same is true for many a-frame elements...

machenmusik:
> If it is a selector, just return the element. Ensure it has a src set.

i am not sure that it should be a-frame's place to attempt to validate all possible ways an element could be considered valid, that seems brittle and not easy to maintain going forward (e.g. @jesstelford srcObject rather than src).  if you agree with "if it is a selector, just return the element" then merging this PR will help ensure that prior use cases that expected and accepted said behavior will continue to work (not only canvas, but also video)

machenmusik:
> If it is a selector, just return the element. Ensure it has a src set.
If it is a URL, then maybe we create a media element for them, setting the src, ensure crossorigin and playsinline. And we append it to <a-assets> so they have a reference to it if they want to control media playback.

I believe the video material handling already does 2; this PR provides 1 other than the src validation per prior comment.  in theory, we could compromise by adding check for either src or srcObject in the case of video, but really THREE.js is permissive and it seems better to just pass through the video element.  if you feel strongly that the check for src or srcObject is truly necessary to merge the pull request, please let me know ASAP.

machenmusik:
@ngokevin this also fixes #2126, and thanks to https://github.com/aframevr/aframe/blob/master/src/systems/material.js#L123, it behaves as you describe, except for "Ensure it has a src set" - primarily due to cases such as @jesstelford where not having a src set is valid.  Rather than coming up with a different fix for #2126, can we work to get this merged?  At minimum, we would need to check for either src or srcObject as mentioned above, if you feel strongly that a-frame should take on the validation burden, but otherwise I think we are all good here...

machenmusik:
@ngokevin @dmarcos still needs changes?

>this also fixes #2126, and thanks to https://github.com/aframevr/aframe/blob/master/src/systems/material.js#L123, it behaves as you describe, except for "Ensure it has a src set" - primarily due to cases such as @jesstelford where not having a src set is valid. Rather than coming up with a different fix for #2126, can we work to get this merged? At minimum, we would need to check for either src or srcObject as mentioned above, if you feel strongly that a-frame should take on the validation burden, but otherwise I think we are all good here...

machenmusik:
as compromise, added warning if video element has neither src nor srcObject
(because existing behavior for invalid video is to warn)
hopefully we are now OK to merge

> thanks to https://github.com/aframevr/aframe/blob/master/src/systems/material.js#L123, it behaves as you describe

> If it is a selector, just return the element. Ensure it has a src set.
If it is a URL, then maybe we create a media element for them, setting the src, ensure crossorigin and playsinline. And we append it to <a-assets> so they have a reference to it if they want to control media playback.

machenmusik:
(@dmarcos @ngokevin added warning if no src or srcObject, this one should be good to go now.)

machenmusik:
@dmarcos @ngokevin is there something holding up merge of this PR?

ngokevin:
Sorry, just got back from vacation and work trips, will look at it tomorrow as I prepare release. Thanks!

jesstelford:
Thanks for the merge @ngokevin :tada:

PR #2131, opened 2016-11-23
Moved the recommended way (component) first

Utopiah:
**Description:** positive to encourage newcomers to write component and embrace the ECS pattern

**Changes proposed:**
- switched the recommended way (component) up and scene.hasLoaded down as specific-case only alternative.


dmarcos:
Thanks

PR #2135, opened 2016-11-23
Fix deprecation warnings within the examples

darkwing:
Just some cleanup of examples so they don't give off the deprecation error for `getComputedAttribute`


dmarcos:
Nice. I was noticing those too :)

PR #2137, opened 2016-11-23
Fix material update referencing the sceneEl when it has not loaded

dmarcos:


ngokevin:
r+

test if easy/possible

PR #2138, opened 2016-11-23
Fix a stringification bug when a null value is used for a property wi…

dmarcos:
**Description:**

**Changes proposed:**
-
-
-

…th an

object type like vec3


ngokevin:
r+

PR #2144, opened 2016-11-27
Preserve type of default property values.

### LINKED ISSUE #2140: [0.4.0] #2108 incorrectly returns empty object when default value is `null`
**Description:**

- A-Frame Version: `master` (#2108)
- Platform/Device: All

When setting a component schema such as:

```javascript
schema: {
  foo: {
    type: 'string',
    default: null,
  }
}
```

The values set on `this.data` don't reflect the default `null`, instead, we end up with an empty object.

The problem is introduced by [this line](https://github.com/dmarcos/aframe/blob/53eae6d00a6eca53e494b6e3f2887251a36549a3/src/core/component.js#L334):

```javascript
data[key] = typeof defaultValue === 'object' ? utils.extend({}, defaultValue) : defaultValue;
```

The `utils.extend` returns a new empty object.

---

## The Fix

Instead, it should check to see if it's actually an object, maybe via one of these two methods:

```javascript
data[key] = Object.prototype.toString.call(defaultValue) === '[object Object]' ? utils.extend({}, defaultValue) : defaultValue;
```

or

```javascript
data[key] = (defaultValue.constructor && defaultValue.constructor === Object) ? utils.extend({}, defaultValue) : defaultValue;
```

Personally; I prefer the first option as it protects against the mutable `.constructor` property being modified by userland code.

---

_Edit_: Am happy to open a PR; just give the :+1: for one of the two solutions above, and I'll get it done.

### LINKED ISSUE #2143: Default values of array properties are ignored.
Similar to #2140, an array provided in schema as a default value is [first cast to an Object](https://github.com/dmarcos/aframe/blob/53eae6d00a6eca53e494b6e3f2887251a36549a3/src/core/component.js#L334), and [then converted to an empty array](https://github.com/dmarcos/aframe/blob/53eae6d00a6eca53e494b6e3f2887251a36549a3/src/core/propertyTypes.js#L53-L54).

- A-Frame Version: master / unreleased

```js
AFRAME.registerComponent('array-container', {
  schema: {
    list: {default: ['a', 'b', 'c', 'd']}
  },
  init: function () {
    console.log('[array-container] %o', this.data.list); // []
  }
});
```

donmccurdy:
Fixes #2140.
Fixes #2143.

Not sure we have consensus on this as a method of detecting POJOs (maybe it should be a util?), but this fixes the cases I'm aware of.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -332,7 +332,9 @@ function buildData (el, name, attrName, schema, elData, silent) {
     data = {};
     Object.keys(schema).forEach(function applyDefault (key) {
       var defaultValue = schema[key].default;
-      data[key] = typeof defaultValue === 'object' ? utils.extend({}, defaultValue) : defaultValue;
+      data[key] = defaultValue && defaultValue.constructor === Object
+        ? utils.extend({}, defaultValue)
+        : defaultValue;
     });
   }
 
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -52,6 +52,19 @@ suite('Component', function () {
       assert.equal(data, 'blue');
     });
 
+    test('preserves type of default values', function () {
+      var schema = processSchema({
+        list: {default: [1, 2, 3, 4]},
+        none: {default: null},
+        string: {default: ''}
+      });
+      var el = document.createElement('a-entity');
+      var data = buildData(el, 'dummy', 'dummy', schema, undefined, null);
+      assert.shallowDeepEqual(data.list, [1, 2, 3, 4]);
+      assert.equal(data.none, null);
+      assert.equal(data.string, '');
+    });
+
     test('uses mixin values', function () {
       var data;
       var TestComponent = registerComponent('dummy', {


src/core/component.js
```diff
@@ -332,7 +332,9 @@ function buildData (el, name, attrName, schema, elData, silent) {
     data = {};
     Object.keys(schema).forEach(function applyDefault (key) {
       var defaultValue = schema[key].default;
-      data[key] = typeof defaultValue === 'object' ? utils.extend({}, defaultValue) : defaultValue;
+      data[key] = defaultValue && defaultValue.constructor === Object
```
dmarcos: I would be invert the conditions and be explicit with the check for null:
`defaultValue.constructor === Object && and defaultValue !== null`
In other parts of the code we use `typeof defaultValue === 'object' `. Is there any advantage on using `defaultValue.constructor === Object`. If they're equivalent I would use the first form for consistency.

dmarcos:
Thanks for the PR

src/core/component.js
donmccurdy: `defaultValue.constructor === Object && defaultValue !== null` will throw an error, given `null`.

The disadvantage of `typeof defaultValue === 'object'` is that it evaluates true for `[1,2,3]`. If other code care about the distinction between Array and Object, perhaps a `util.isPlainObject()` method would be helpful. Here's Lodash's equivalent:

```js
    /**
     * Checks if `value` is a plain object, that is, an object created by the
     * `Object` constructor or one with a `[[Prototype]]` of `null`.
     *
     * @static
     * @memberOf _
     * @since 0.8.0
     * @category Lang
     * @param {*} value The value to check.
     * @returns {boolean} Returns `true` if `value` is a plain object, else `false`.
     * @example
     *
     * function Foo() {
     *   this.a = 1;
     * }
     *
     * _.isPlainObject(new Foo);
     * // => false
     *
     * _.isPlainObject([1, 2, 3]);
     * // => false
     *
     * _.isPlainObject({ 'x': 0, 'y': 0 });
     * // => true
     *
     * _.isPlainObject(Object.create(null));
     * // => true
     */
    function isPlainObject(value) {
      if (!isObjectLike(value) || baseGetTag(value) != objectTag) {
        return false;
      }
      var proto = getPrototype(value);
      if (proto === null) {
        return true;
      }
      var Ctor = hasOwnProperty.call(proto, 'constructor') && proto.constructor;
      return typeof Ctor == 'function' && Ctor instanceof Ctor &&
        funcToString.call(Ctor) == objectCtorString;
    }
```

donmccurdy: `undefined` would still throw, but otherwise I think your suggestion works. Is `undefined` considered a valid default value?

dmarcos:
Thanks!

PR #2145, opened 2016-11-27
Ignore primitive mappings that have the same name as an existing component (fix #2142)

### LINKED ISSUE #2142: [0.4.0] Cannot opt-out of fog / apply flat shader
**Description:**

Opting out of fog with `material="fog: false; shader: flat"` no longer works on `master`:

```
<a-scene fog="type: linear; color: #AAB; far: 30; near: 0">
  <a-sphere
    position="0 10 0"
    radius="10"
    color="#93648D"
    material="fog: false; shader: flat"
  ></a-sphere>
</a-scene>
```

- A-Frame Version: `master` (909e5ec89cd2e0e11f56db51819deef170f0a115)
- Platform/Device: Desktop Chrome
  - `master` (not working): http://codepen.io/anon/pen/YpxPxz
  - `0.3.2` (working): http://codepen.io/anon/pen/NbvPvj

_Edit_: It appears to be working correctly on 32b1378109361e898eb47e51d1a7691013402c8b, which means something went wrong between [32b1378...909e5ec](https://github.com/aframevr/aframe/compare/32b1378109361e898eb47e51d1a7691013402c8b...909e5ec89cd2e0e11f56db51819deef170f0a115/)

_Edit 2_; Some `git bisect`ing reveals, the first bad commit is 8b0df2365c77dd43666afb6f63e922b4d4037753 / #2106 

_Edit 3_: Confirming; reverting the commits in #2106 fixes this issue.

dmarcos:


src/extras/primitives/getMeshMixin.js
```diff
@@ -11,7 +11,9 @@ Object.keys(shaders.standard.schema).forEach(addMapping);
 
 function addMapping (prop) {
   // To hyphenated.
-  var htmlAttrName = prop.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+  var htmlAttrName;
+  if (components[prop]) { return; }
+  htmlAttrName = prop.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
```
ngokevin: Seems a bit ad-hoc to apply it only to meshMixin primitives, which are only a handful. We probably want to update the core primitive code to detect this and throw a warning.

The symptom problem is `visible: material.visible` mapping could be renamed to `material-visible: material.visible`.

ngokevin:
r+

src/extras/primitives/primitives.js
```diff
@@ -30,6 +31,29 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
...
+       */
+      resolveMappingCollisions: {
+        value: function () {
+          var mappings = this.mappings;
+          var self = this;
+          Object.keys(mappings).forEach(function resolveCollision (key) {
+            var newAttribute;
+            if (components[key]) {
+              newAttribute = mappings[key].replace('.', '-');
+              mappings[newAttribute] = mappings[key];
```
jesstelford: What if there's also a component that matches, say, `material-visible`? Admittedly less likely than the original issue, but the door is still open for a collision.

jesstelford:
Thanks for figuring out the original issue! :+1:

src/extras/primitives/primitives.js
ngokevin: Alternative API might be: `<a-box style="width: 5; color: red">`, containing the mappings within a reserved component?

PR #2151, opened 2016-12-01
Fixes #2088

### LINKED ISSUE #2088: Hemisphere light stops emitting light when groundColor is changed runtime
**Description:**

Changing attributes of `light` components works great in general, but I think there's a bug specifically when changing the `groundColor` on a hemisphere light.

As demonstrated in the Codepen example linked below, setting the `color`of a hemisphere light run time works as expected, but setting the `groundColor` results in the whole model being rendered without any light on it at all.

I would expect to be able to change the `groundColor` just as I can change the `color` of a hemisphere light.

- A-Frame Version: 0.3.2
- Platform/Device: Chrome on Mac laptop
- Reproducible Code Snippet or Demo URL:
 http://codepen.io/fabricius/pen/GNqJdr

magnus-bergman:
**Description:**
The docs refer to the ground color prop in hemisphere light as groundColor. But the code refers to it as groundcolor. Should this be case insensitive?

**Changes proposed:**
- Update groundcolor to groundColor in light.js.

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -55,7 +55,7 @@ module.exports.Component = registerComponent('light', {
             break;
           }
 
-          case 'groundcolor': {
+          case 'groundColor': {
             light.groundColor.set(value);
             break;
           }


dmarcos:
Thanks!

PR #2153, opened 2016-12-01
Loader

dmarcos:
It provides a loader for scenes to show something on screen while assets and the scene initializes. It demotes `canvas` and `vr-mode-ui` to libraries from components. The loader needs them before components initialize.


donmccurdy:
Glad to review this if it’s still on the roadmap. Or perhaps we should create an external component if we’re hesitant to put this in core?

dmarcos:
This needs more work. We will postpone for 0.9.0

VasuBhog:
Hi I wanted to know if this is still in implementation or if there is another component out there that has same functionality to create a "load screen" before all assets are loaded fluidly.

dmarcos:
This functionality has landed on a different PR.

PR #2154, opened 2016-12-02
Remove modifier key conditions from util.shouldCaptureKeyEvent(e)

### LINKED ISSUE #1479: ASDF movement bug when using SHIFT key
**Description:**
- In one of the Aframe demo scenes, start walking left by holding down the `A` key.
  - https://aframe.io/examples/showcase/helloworld/
- While holding the `A` key, press and hold the `shift` key
- Release the `A` key
- You will continue moving left

**A-Frame Version:** Whatever is running on the aframe site
**Platform/Device:** Chrome desktop

donmccurdy:
Fixes #1479.

The condition for modifier keys was introduced by 8697452f, with the goal of fixing the issue that it now seems to be causing.

I'm testing in three browsers:

* Chrome (KeyboardEvent.**code**)
* Firefox (KeyboardEvent.**code**)
* Safari (KeyboardEvent.**keyCode**)

Currently, all three exhibit #1479 for all modifier keys. With this change, <kbd>shift</kbd>, <kbd>alt</kbd>, and <kbd>ctrl</kbd> are fixed, and do not interfere with WASD controls.. <kbd>meta</kbd> is unfortunately not fixed, but that's a browser implementation "feature" and pretty much unavoidable with userland code.

### DIFF
--- a/src/components/wasd-controls.js
+++ b/src/components/wasd-controls.js
@@ -197,7 +197,6 @@ module.exports.Component = registerComponent('wasd-controls', {
 
   onKeyUp: function (event) {
     var code;
-    if (!shouldCaptureKeyEvent(event)) { return; }
     code = event.code || KEYCODE_TO_CODE[event.keyCode];
     this.keys[code] = false;
   }
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -119,9 +119,7 @@ module.exports.diff = function (a, b) {
  * @returns {Boolean} Whether the key event should be captured.
  */
 module.exports.shouldCaptureKeyEvent = function (event) {
-  if (event.metaKey) {
-    return false;
-  }
+  if (event.metaKey) { return false; }
   return document.activeElement === document.body;
 };
 
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -119,9 +119,6 @@ module.exports.diff = function (a, b) {
  * @returns {Boolean} Whether the key event should be captured.
  */
 module.exports.shouldCaptureKeyEvent = function (event) {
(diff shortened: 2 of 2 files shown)


donmccurdy:
Actually, since the <kbd>meta</kbd> issue presumably can't be fixed, let's leave that condition in so that #932 remains closed (pressing <kbd>meta</kbd>+<kbd>a</kbd> together should do nothing).

dmarcos:
I see... Thanks for looking at this. What about removing the `shouldCaptureKeyEvent` check from the  `onKeyUp` handler?

https://github.com/aframevr/aframe/blob/master/src/components/wasd-controls.js#L200

I thinks it is what https://github.com/aframevr/aframe/issues/1479#issuecomment-220629019 was suggesting

donmccurdy:
Removing it from `onKeyUp` only, not `onKeyDown`? 

I think the end effect is the same as this option. Preference?

dmarcos:
Yeah. One larger question, Does it make sense to keep `shouldCaptureKey` around?

donmccurdy:
Removed the `shouldCaptureKey` conditional in `onKeyUp`.

I agree with the `document.activeElement` check in `shouldCaptureKey`, since that keeps #1063 closed. The <kbd>meta</kbd> check is borderline, IMHO.

donmccurdy:
@dmarcos which changes is this waiting on? I think all comments are addressed.

dmarcos:
There's one comment to make an if statement a one liner

donmccurdy:
Done.

(Somehow I can't see that comment, though, which is weird)

dmarcos:
Thanks!

PR #2162, opened 2016-12-07
make stats an entity injected into camera so they are visible in and …

machenmusik:
**Description:**
make stats an entity injected into camera so they are visible in and out of VR

**Changes proposed:**
src/components/scene/stats.js

### DIFF
--- a/src/components/scene/stats.js
+++ b/src/components/scene/stats.js
@@ -25,6 +25,7 @@ module.exports.Component = registerComponent('stats', {
 
     // hide the DOM stats panel
     this.statsEl.style = 'display: none !important';
+    this.statsEl.className = 'a-hidden';
 
     // defer most initialization until after camera is injected
     var self = this;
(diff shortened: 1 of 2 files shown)


machenmusik:
to the extent that the current stats component is useful, this should be a nice improvement;
not sure what is going on with the failed check though, advice welcome

machenmusik:
(@ngokevin are there any issues, or has no one had a chance to look at it yet?)

machenmusik:
In the meantime, this is available as `stats-in-vr` component in registry, if we want to include in core, should probably pull in latest changes

ngokevin:
I think this could live as a third-party component for now. Could even be simpler, just reading the rStats value and outputting with text component.

machenmusik:
Agree this can live as a 3rd party component, and that the original PR is stale now... although really something should be done to the built-in stats to make them not entirely useless in VR.

PR #2170, opened 2016-12-10
add auto-enter-vr scene component that will enterVR()

machenmusik:
both with an initial attempt (which may fail, but works on Carmel)
and also on vrdisplayactivate (e.g. when you put the HMD on with Chromium)

### DIFF
--- a/src/components/scene/auto-enter-vr.js
+++ b/src/components/scene/auto-enter-vr.js
@@ -13,21 +13,11 @@ module.exports.Component = registerComponent('auto-enter-vr', {
     if (utils.getUrlParameter('auto-enter-vr') === 'false') { return; }
 
     window.addEventListener('vrdisplayactivate', function () {
-      console.log('vrdisplayactivate, enterVR()');
       scene.enterVR();
     }, false);
 
-    this.el.addEventListener('enter-vr', function () {
-      console.log('enter-vr');
-    }, false);
-
     window.addEventListener('vrdisplaydeactivate', function () {
-      console.log('vrdisplaydeactivate, exitVR()');
       scene.exitVR();
-    }, false);
-
-    this.el.addEventListener('exit-vr', function () {
-      console.log('exit-vr');
     }, false);
 
     // just try to enter... turns out we need to wait for next tick
--- a/src/components/scene/auto-enter-vr.js
+++ b/src/components/scene/auto-enter-vr.js
@@ -5,7 +5,7 @@ var utils = require('../../utils');
  * Automatically enter VR, either upon vrdisplayactivate or immediately if possible.
  */
 module.exports.Component = registerComponent('auto-enter-vr', {
-  schema: {default: true},
+  schema: {default: 'GearVR'},
 
   init: function () {
     var scene = this.el;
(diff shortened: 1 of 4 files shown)


machenmusik:
am investigating how to correctly add this functionality to all scenees by default, so examples using aframe will just work with Carmel and other browsers that allow entering VR without user action and don't emit vrdisplayactivate; for now, you can use this and just say `a-scene auto-enter-vr`

machenmusik:
hmm OK, a component version is currently in a PR to a-painter; pointers to add to registry would be helpful.  The original idea was to detect Carmel and auto-enter VR so that A-Frame sites would work unmodified, which would presumably belong as a PR here, in which case it should not be an external component, correct?  Will modify this PR to be scoped to Carmel only since it has no other way to enter VR...

dmarcos:
There's an `enter-vr-ui` component already. What about adding an `auto` flag to it? https://github.com/aframevr/aframe/blob/master/src/components/scene/vr-mode-ui.js

dmarcos:
We can rename that component to just `enter-vr`

machenmusik:
OK so... according to Justin, the Carmel UA is indistinguishable at the moment, and the gamepad 'Gear VR' doesn't appear to be available (probably the general issue where gamepads aren't enumerable until first used, for security purposes) -- so the way to have it auto-enter VR for Carmel is to have it auto-enter in general.

To add to all scenes by default, so A-Frame scenes just work with Carmel, I added `auto-enter-vr` last in the list of default scene components (rather than first, which didn't work).

I thought that the `enter-vr-ui` component was about showing the button that lets you enter / exit VR, not actually doing it, so that was related but separate UI functionality.  At present, with both enabled, an A-Frame scene will auto-enter VR< and if/when you exit, you will see the normal button to allow re-entry.

src/components/scene/auto-enter-vr.js
```diff
@@ -0,0 +1,53 @@
...
+    if (utils.getUrlParameter('auto-enter-vr') === 'false') { return; }
+
+    // enter VR on vrdisplayactivate (e.g. putting on Rift headset)
+    window.addEventListener('vrdisplayactivate', function () { this.enterVR(); }, false);
+
+    // exit VR on vrdisplaydeactivate (e.g. taking off Rift headset)
+    window.addEventListener('vrdisplaydeactivate', function () { this.exitVR(); }, false);
+
+    // check if we should try to enter VR... turns out we need to wait for next tick
+    var self = this;
```
dmarcos: move this to the top of the function

src/components/scene/auto-enter-vr.js
```diff
@@ -0,0 +1,53 @@
...
+    setTimeout(function () { if (self.shouldAutoEnterVR()) { this.enterVR(); } }, 0);
```
dmarcos: Why is a timeout necessary?

src/components/scene/auto-enter-vr.js
```diff
@@ -0,0 +1,53 @@
...
+
+  update: function () {
+    return this.shouldAutoEnterVR() ? this.enterVR() : this.exitVR();
+  },
+
+  shouldAutoEnterVR: function () {
+    var scene = this.el;
+    var data = this.data;
+    // if false (or string false), we should not auto-enter VR
+    if (data === false || data === 'false') { return false; }
```
dmarcos: the type of the data is a string it should not ever be `false` (boolean)

src/components/scene/auto-enter-vr.js
machenmusik: not exactly sure, but it appears that the renderer etc. needs to be set up for this to succeed, which isn't ready to go at init

dmarcos: It's probably better to do: 
````javascript
sceneEl.addEventListener('renderstart' this.enterVR);
````

File: src/components/scene/auto-enter-vr.js [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+  },
+
+  shouldAutoEnterVR: function () {
+    var scene = this.el;
+    var data = this.data;
+    // if false, we should not auto-enter VR
+    if (data === 'false') { return false; }
+    // if we have a data string to match against display name, try and get it;
+    // if we can't get display name, or it doesn't match, we should not auto-enter VR
+    if (typeof data === 'string') {
```
dmarcos: I think we should change the schema to two flags instead of overload a single one:
I would do:
````javascript
schema: {
  headset: 'GearVR',
  enabled: true
}
````

ngokevin: makes sense

machenmusik: yes but you threw me for a second because the syntax is wrong ;-)

PR #2173, opened 2016-12-12
[docs] Switch event names for vive-controls

Sebkasanzew:
**Description:**
Event names for vive-controls had wrong description

**Changes proposed:**
- Switch trackpad and trigger event names to match correct description


dmarcos:
Thanks!

PR #2132, opened 2016-11-23
Refactor hand controls, oculus touch support

machenmusik:
refactor hand-controls per discussion on https://github.com/aframevr/aframe/pull/2124

PR #2183, opened 2016-12-18
Support Gear VR Touchpad controller 0.4.0

machenmusik:
The `gearvr-controls` component interfaces with the "Gear VR Touchpad"
controller exposed by the Carmel and Samsung Internet VR browsers.  
As that controller has no pose,
the `tracked-controls` component cannot provide its usual
functionality (but an instance is added anyway since other components such 
as `aframe-teleport-controls` use that to find controller entities).  Instead,
the `look-controls` component is added to mimic a 3DOF controller using HMD orientation,
similar to gaze-based cursors.

This component adds button mappings and events, but does not currently provide
a controller model since it is assumed that end users will use this indirectly
through higher level components such as the hand-controls component.  

When used via `hand-controls`, as there is only one "Gear VR Touchpad",  
currently this component will only bind to the right hand, not the left.

### DIFF
--- a/src/components/carmel-gearvr-controls.js
+++ b/src/components/carmel-gearvr-controls.js
@@ -1,0 +1,201 @@
+var registerComponent = require('../core/component').registerComponent;
+var bind = require('../utils/bind');
+var getGamepadsByPrefix = require('../utils/tracked-controls').getGamepadsByPrefix;
+
+var GAMEPAD_ID_PREFIX = 'Gear VR Touchpad';
+
+/**
+ * Vive Controls Component
+ * Interfaces with vive controllers and maps Gamepad events to
+ * common controller buttons: trackpad, trigger, grip, menu and system
+ * It loads a controller model and highlights the pressed buttons
+ */
+module.exports.Component = registerComponent('carmel-gearvr-controls', {
+  schema: {
+    hand: {default: 'left'}
+  },
+
+  // buttonId
+  // 0 - trackpad
+  // 1 - trigger ( intensity value from 0.5 to 1 )
+  // 2 - grip
+  // 3 - menu ( dispatch but better for menu options )
+  // 4 - system ( never dispatched on this layer )
+  mapping: {
+    axis0: 'trackpad',
+    axis1: 'trackpad',
+    button0: 'trackpad'
+  },
+
+  bindMethods: function () {
+    this.onModelLoaded = bind(this.onModelLoaded, this);
+    this.checkIfControllerPresent = bind(this.checkIfControllerPresent, this);
+    this.onGamepadConnected = bind(this.onGamepadConnected, this);
+    this.onGamepadDisconnected = bind(this.onGamepadDisconnected, this);
+  },
+
+  init: function () {
+    this.buttonStates = {};
+    this.previousAxis = [];
+    this.animationActive = 'pointing';
+    this.controllerPresent = false;
+    this.bindMethods();
+    this.getGamepadsByPrefix = getGamepadsByPrefix; // to allow mock
+  },
+
+  addEventListeners: function () {
+    var el = this.el;
+    el.addEventListener('buttonchanged', this.onButtonChanged);
+    el.addEventListener('buttondown', this.onButtonDown);
+    el.addEventListener('buttonup', this.onButtonUp);
+    el.addEventListener('touchstart', this.onButtonTouchStart);
+    el.addEventListener('touchend', this.onButtonTouchEnd);
+    el.addEventListener('model-loaded', this.onModelLoaded);
+  },
+
+  removeEventListeners: function () {
+    var el = this.el;
+    el.removeEventListener('buttonchanged', this.onButtonChanged);
+    el.removeEventListener('buttondown', this.onButtonDown);
+    el.removeEventListener('buttonup', this.onButtonUp);
+    el.removeEventListener('touchstart', this.onButtonTouchStart);
+    el.removeEventListener('touchend', this.onButtonTouchEnd);
+    el.removeEventListener('model-loaded', this.onModelLoaded);
+  },
+
+  checkIfControllerPresent: function () {
+    // var data = this.data;
+    // The 'Gear VR Touchpad' gamepad exposed by Carmel has no pose,
+    // so it won't show up in the tracked-controls system controllers.
+    var gamepads = this.getGamepadsByPrefix(GAMEPAD_ID_PREFIX);
+    var isPresent = gamepads && gamepads.length > 0;
+    // we need up to update the controller state ourselves
+    if (isPresent) { this.controller = gamepads[0]; }
+    if (isPresent === this.controllerPresent) { return; }
+    this.controllerPresent = isPresent;
+    if (isPresent) {
+      this.el.setAttribute('look-controls', '');
+      this.addEventListeners();
+    } else {
+      this.el.removeAttribute('look-controls');
+      this.removeEventListeners();
+    }
+  },
+
+  tick: function () {
+    // The 'Gear VR Touchpad' gamepad exposed by Carmel has no pose,
+    // so it won't show up in the tracked-controls system controllers.
+    // Therefore, we have to do tick processing for the Gear VR Touchpad ourselves.
+    this.checkIfControllerPresent();
+    // we don't need to update the pose since we inject look-controls
+    this.updateButtons();
+  },
+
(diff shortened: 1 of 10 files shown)


src/components/carmel-gearvr-controls.js
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Vive Controls Component
+ * Interfaces with vive controllers and maps Gamepad events to
+ * common controller buttons: trackpad, trigger, grip, menu and system
+ * It loads a controller model and highlights the pressed buttons
+ */
+module.exports.Component = registerComponent('carmel-gearvr-controls', {
+  // since aframe-teleport-controls looks for tracked-controls, add one
+  dependencies: ['tracked-controls'],
```
dmarcos: the gearvr controls is just the trackpad on the headset right? I don't think it fits in the `tracked-controls` models.

src/components/carmel-gearvr-controls.js
```diff
@@ -0,0 +1,200 @@
...
+
+  schema: {
+    hand: {default: 'right'},
```
dmarcos: It feels weird to assign a hand to the trackpad

src/components/carmel-gearvr-controls.js
donmccurdy: +1, but at a higher level, would this make more sense in the registry vs core? I think it's valuable to include, in either case.

Also wondering if the name (and functionality) could be broader than carmel and gearvr. Assuming this is purely using gamepad API, this could be `trackpad-controls` or `dpad-controls`.

machenmusik: right, it is not a `tracked-controls` really, it only has a stub of that because others like `aframe-teleport-controls` are looking for entities with that label.  it makes `hand-controls` allow gaze-based usage on Carmel, so that things like `aframe-teleport-controls` can just reference `hand-controls` and not worry about specific implementation.  if you try https://chenzlabs.github.io/aframe-teleport-controls/mesh/ you can see what I mean (swipe and hold, then release).

machenmusik: @donmccurdy i agree this could be more generic like touchpad-controls if the behavior were consistent across browsers (not just Carmel) and controllers (not just Gear VR Touchpad), but it is not clear that is the case.  happy to rename if that helps.

src/components/carmel-gearvr-controls.js
machenmusik: that is for compatibility with `hand-controls` only

machenmusik:
>The application should detect if there are tracked controllers and if not used an alternative interaction method (e.g: gazed based cursor) that better adapts to the capabilities of the device

Yes, this gets back to some of the original conversation on #2124 and #2132.

These changes allow applications to do that easily by using `hand-controls='right'`, rather than having to explicitly include all of `oculus-touch-controls`, `vive-controls` and `carmel-gearvr-controls` and handle them separately which is tedious and scales poorly (what are they supposed to do with Daydream then)?

And rather than having a well-known component responsible for figuring out which controllers are attached and only injecting the right one(s) -- that was `auto-detect-controllers` -- since you guys specifically rejected that approach in core previously, but then we still have higher level like `hand-controls` in core so people can use "out of the box" rather than having to go hunt everything down in the registry.

machenmusik:
> I have no idea how you're supposed to control your hands with 2 axis. I feel that the touchpad is more appropriate to be paired with a gazed based cursor. 

@dmarcos I just realized, you may not have understood from the description that this is basically what the PR does -- modified the description to make it more obvious that for Carmel GearVR, it is `look-controls` and HMD orientation that is used for controller pose, not the 2D position from the touchpad.

src/components/carmel-gearvr-controls.js
machenmusik: renamed to `gearvr-controls` and added support for Samsung Internet GearVR.  (The test scene I use looks low-res in that browser for some reason, but the functionality works as expected.)

dmarcos:
I think `hand-controls` should just be exclusive to `tracked-controls`. Other method inputs should map to different interaction models that are not hands. A gamepad or the gear VR touchpad should map to a controls schema other than `hand-controls`: a gaze based interaction for instance

machenmusik:
IMO, there's no particularly good reason why `hand-controls` needs to be exclusive.  For zero-DOF controllers like GearVR, the hand position and pose can be emulated from player position and HMD (gaze), and the hand model as visual representation and pointing animation can still be retained.  For 3DOF like Daydream, the hand position can be emulated from player position, and the actual controller orientation can be used.

As previously discussed, I think most of this comes up because `hand-controls` is in core and serves as showcase example as well.  If you'd like a switch to enable/disable emulation, that makes sense to me...

dmarcos:
I want to keep `hand-controls` simple and exclusive to `tracked-controls`. The [tracked-controls](https://aframe.io/examples/showcase/tracked-controllers/) demo (walking and reaching objects out with your hands) is not a very good experience with GearVR or a Daydream control. For experiences that use just pointing at targets I can see a 3DOF controls to be used but for GearVR you probably want something different than emulating a hand with your gaze. You're idea of responsive controls is really worth exploring and probably an excellent addition to the registry. We're trying to keep the core as nimble and simple as we can with pieces that are highly reusable and not very opinionated.

machenmusik:
So as things stand there are no examples of what to do with A-Frame on GearVR (and Daydream),  I agree with your statement that the tracked-controls example really calls out for Touch and/or Vive controllers, but I think https://chenzlabs.github.io/aframe-teleport-controls/mesh/ shows that the hand-controls component can still be useful in other situations.  It sounds as though you are asking those looking to support the most numerous VR platforms by far to go outside A-Frame, rather than also having examples of how to use non-tracked controllers as part of A-Frame.

machenmusik:
Maybe all that needs to change here to accommodate your perspective is that `hand-controls` should not inject `gearvr-controls` -- but this code should still exist in core; if one wants to allow GearVR emulated hand control, then simply do `<a-entity hand-controls="right" gearvr-controls>` -- and if not, `gearvr-controls` is available to be leveraged directly

dmarcos:
Also I'm not sure if  a `gearvr-controls` makes sense. At the end of the day it's just one kind of gamepad. The same as the x-box or a PS gamepad. Maybe a `gamepad-controls` makes more sense. One could register mappings for different gamepads. This could also simplify `tracked-controls` to just handle the pose changes.

machenmusik:
right now the mouse/touchpad emulation is activated for GearVR when using Samsung Internet VR browser, and there is specific hackery to leverage the Carmel gamepad mapping.  The first could be made more generic like the auto-enter discussion, but the second probably shouldn't be.  are you suggesting splitting those into two components?

dmarcos:
> right now the mouse/touchpad emulation is activated for GearVR when using Samsung Internet VR browser, and there is specific hackery to leverage the Carmel gamepad mapping

This is why I push back on this kind of changes. I want to minimize the liabilities in the core. We don't have the resources to pay close attention to each browser, version and device to adapt the code to all the quirks (specially for those in experimental phase). The reason of having the [registry](https://aframe.io/aframe-registry/) is:
1. Enable permission less innovation where multiple solutions to a problem can coexist and compete. 
2. Having a community that helps provide all the functionality that we don't have the resources to develop and maintain.

The core team will make its best to facilitate the conversation and create a system for people to share, discover and use those components.

src/components/gearvr-controls.js
```diff
@@ -0,0 +1,221 @@
...
+var GAMEPAD_ID_PREFIX = 'Gear VR Touchpad';
+
+/**
+ * GearVR Controls Component
+ * Interfaces with Gear VR touchpad.  For Carmel browser, it maps
+ * Gamepad events to common controller button (trackpad) and axes.
+ * For Samsung Internet GearVR browser, it uses mouse and touch events
+ * to simulate appropriate Gamepad behavior.
+ */
+module.exports.Component = registerComponent('gearvr-controls', {
```
dmarcos: I would definitively add this component to the registry

machenmusik:
OK, but even with `gearvr-controls` moved out of core, the other small changes (e.g. use !isOculusTouch rather than isVive) are needed for compatibility.  Would you like me to remove `gearvr-controls` from the PR and have you merge the rest?

machenmusik:
(the `gamepad-controls` comment is interesting though, as it makes the case to include emulated tracking in core to accompany them, for responsive controls more generically supported)

PR #2184, opened 2016-12-18
When updating component attrValue has to be used instead of component.data

dmarcos:
The bug was discovered when doing:
````javascript
el.setAttribute('geometry', {primitive: 'box'});
el.setAttribute('geometry', {primitive: 'sphere'});
````
After the first line the value of el.components.geometry.attrValue was `{primitive: box}`
After the update `{"buffer":true,"primitive":"sphere","skipCache":false,"depth":1,"height":1,"width":1,"segmentsHeight":1,"segmentsWidth":1,"segmentsDepth":1}` causing the default values of the box primitive to be applied to the sphere primitive after update.

### DIFF
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -17,7 +17,12 @@ function parse (value, defaultVec) {
   var vec = {};
 
   if (value && typeof value === 'object') {
-    return vecParseFloat(value);
+    return vecParseFloat({
+      x: value.x || defaultVec && defaultVec.x,
+      y: value.y || defaultVec && defaultVec.y,
+      z: value.z || defaultVec && defaultVec.z,
+      w: value.w || defaultVec && defaultVec.w
+    });
   }
 
   if (typeof value !== 'string' || value === null) {
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -363,6 +363,29 @@ suite('a-entity', function () {
       el.setAttribute('test', 'asym', 1);
       el.setAttribute('test', 'other', 2);
     });
+
+    test('the attribute cache only stores the modified properties', function () {
+      var el = this.el;
+      el.setAttribute('geometry', {primitive: 'box'});
+      assert.deepEqual(el.components.geometry.attrValue, {primitive: 'box'});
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/utils/coordinates.js [hunk 1]
```diff
@@ -17,7 +17,12 @@ function parse (value, defaultVec) {
   var vec = {};
 
   if (value && typeof value === 'object') {
-    return vecParseFloat(value);
+    coordinate = {};
```
ngokevin:
```js
// Fill in missing values for each vector component.
return vecParseFloat({
  x: value.x || defaultVec & defaultVec.x,
  // ...
});
```

tests/core/a-entity.test.js
```diff
@@ -680,6 +680,19 @@ suite('a-entity', function () {
...
+      var el = this.el;
+      el.setAttribute('geometry', {primitive: 'sphere', radius: 10});
+      assert.deepEqual(el.components.geometry.attrValue, {primitive: 'sphere', radius: 10});
+    });
+
+    test('when changing schema only the modified properties are cached', function () {
+      var el = this.el;
+      el.setAttribute('geometry', {primitive: 'sphere', radius: 10});
+      el.setAttribute('geometry', {primitive: 'box'});
+      assert.deepEqual(el.components.geometry.attrValue, {primitive: 'box', radius: 10});
```
ngokevin: Should assert the component data as well with `getAttribute`.

tests/core/a-entity.test.js
```diff
@@ -680,6 +680,19 @@ suite('a-entity', function () {
       el.setAttribute('class', 'pied piper');
       assert.equal(el.getAttribute('class'), 'pied piper');
     });
+
+    test('the attribute cache only stores the modified properties', function () {
+      var el = this.el;
+      el.setAttribute('geometry', {primitive: 'sphere', radius: 10});
```
ngokevin: I would assert the `attrValue` before the change too for clarity. Perhaps start with a box.

tests/core/a-entity.test.js
dmarcos: well, I should probably move this to the `setAttribute` suite

PR #2188, opened 2016-12-19
defend against missing animations

machenmusik:
a recent change to hand-controls models exposed a weakness when animations are missing;
this PR makes hand-controls defensive i.e. the missing animations won't play, but won't break other things


dmarcos:
I think it's fine to let it break. It makes obvious what it's going on. If the animation just don't play you will spent more time figuring out what happened

dmarcos:
Animation are supposed to be defined in the model. The component is designed around the asset. I'm fine to let i break so a bug speaks loudly.

machenmusik:
the problem is that animations were breaking more than just themselves, really model errors should not prevent eventing and pose updates

Issue #2136, opened 2016-11-23
add utils.throttleTick to be called once every X ms


ngokevin:
**Description:**

Common case to only want to call a tick behavior on an interval rather than every frame.

```js
AFRAME.registerComponent('foo', {
  tick: utils.throttleTick(function (t, dt) {
  
  }, 500);
});
```

This would keep track of last time the function ran, and wait 500 ms before calling again.

mikebolt:
The code looks good to me.

You may want to use window.performance.now for a more precise timestamp. We wrote [a polyfill](https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L72) for window.performance.now in tween.js.

Also, this could be useful for tween.js. I might borrow this code.

machenmusik:
(@ngokevin this is done right?)

ngokevin:
Added

PR #2189, opened 2016-12-19
add utils.throttleTick to be called once every X ms

machenmusik:
implement https://github.com/aframevr/aframe/issues/2136

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -155,14 +155,16 @@ Checks if device is a smartphone. Returns a `boolean`.
 
 ### `AFRAME.utils.throttle(function, minimumInterval [, optionalContext])`
 
-Returns throttle function that only fires call after minimum interval,
-in tick time units (milliseconds).
+Returns a throttled function that limits the amount of times it is called
+to at most every `minimumInterval` milliseconds. A context such as `this`
+can be provided to handle function binding for convenience.
 Note, this does not try to make up for any lost time by firing multiple callbacks.
 
 ### `AFRAME.utils.throttleTick(function (t, dt) {...}, minimumInterval [, optionalContext])`
 
-Returns tick throttle function that only fires call after minimum interval,
-in tick time units (milliseconds).
+Returns a throttled tick function that limits the amount of times it is called
+to at most every `minimumInterval` milliseconds. A context such as `this`
+can be provided to handle function binding for convenience.
 Note, this does not try to make up for any lost time by firing multiple callbacks.
 
 ```
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -153,7 +153,13 @@ Checks if device is a smartphone. Returns a `boolean`.
 
 ## Function Utils
 
-### `AFRAME.utils.throttleTick(function (t, dt) {...}, interval)`
+### `AFRAME.utils.throttle(function, minimumInterval [, optionalContext])`
+
+Returns throttle function that only fires call after minimum interval,
+in tick time units (milliseconds).
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+### `AFRAME.utils.throttleTick(function (t, dt) {...}, minimumInterval [, optionalContext])`
 
 Returns tick throttle function that only fires call after minimum interval,
 in tick time units (milliseconds).
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -150,3 +150,18 @@ Checks if device is Gear VR. Returns a `boolean`.
 ### `AFRAME.utils.device.isMobile()`
 
 Checks if device is a smartphone. Returns a `boolean`.
+
+## Tick Utils
+
+### `AFRAME.utils.throttleTick(function (t, dt) {...}, interval)`
+
+Returns tick throttle function that only fires callback after minimum interval.
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+```
+AFRAME.registerComponent('foo', {
+  tick: utils.throttleTick(function (t, dt) {
+  
+  }, 500);
+});
+```
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -166,22 +166,22 @@ in tick time units (milliseconds).
 Note, this does not try to make up for any lost time by firing multiple callbacks.
 
 ```
-...
 AFRAME.registerComponent('foo', {
-  // This function has the same signature as a normal tick handler,
-  // but has been renamed so it is obvious the function is throttled
-  // (for example, when debugging).
-  throttledTick: function (t, dt) {
-    // Called periodically, no more often than every 500 milliseconds
-  },
-
   init: function () {
     // Set up the tick throttling.
     this.tick = AFRAME.utils.throttleTick(
-      // Because the function to throttle is an object method, bind this pointer
-      AFRAME.utils.bind(this.throttledTick, this),
+      // The function to throttle.
+      this.throttledTick,
       // The function to throttle will not be called again sooner than this in milliseconds.
-      500);
+      500,
+      // Because the function to throttle is an object method, bind this pointer.
+      this);
+  },
+
+  // Renamed tick handler to make it clear that it is throttled.
+  // (for example, when debugging).
+  throttledTick: function (t, dt) {
+    // Called at most once every 500 milliseconds
   }
 });
 ```
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -151,17 +151,31 @@ Checks if device is Gear VR. Returns a `boolean`.
 
 Checks if device is a smartphone. Returns a `boolean`.
 
-## Tick Utils
+## Function Utils
 
 ### `AFRAME.utils.throttleTick(function (t, dt) {...}, interval)`
 
-Returns tick throttle function that only fires callback after minimum interval.
+Returns tick throttle function that only fires call after minimum interval,
+in tick time units (milliseconds).
 Note, this does not try to make up for any lost time by firing multiple callbacks.
 
 ```
+...
 AFRAME.registerComponent('foo', {
-  tick: utils.throttleTick(function (t, dt) {
-  
-  }, 500);
+  // This function has the same signature as a normal tick handler,
+  // but has been renamed so it is obvious the function is throttled
+  // (for example, when debugging).
+  throttledTick: function (t, dt) {
+    // Called periodically, no more often than every 500 milliseconds
+  },
+
+  init: function () {
+    // Set up the tick throttling.
+    this.tick = AFRAME.utils.throttleTick(
+      // Because the function to throttle is an object method, bind this pointer
+      AFRAME.utils.bind(this.throttledTick, this),
+      // The function to throttle will not be called again sooner than this in milliseconds.
+      500);
+  }
 });
 ```
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -1,5 +1,6 @@
 var registerSystem = require('../core/system').registerSystem;
 var trackedControlsUtils = require('../utils/tracked-controls');
+var utils = require('../utils');
 
 /**
  * Tracked controls system.
@@ -10,6 +11,7 @@ module.exports.System = registerSystem('tracked-controls', {
     var self = this;
     this.controllers = [];
     this.lastControllersUpdate = 0;
+    this.tick = utils.throttleTick(this.throttledTickHandler.bind(this), 10);
     if (!navigator.getVRDisplays) { return; }
     navigator.getVRDisplays().then(function (displays) {
       if (displays.length > 0) {
@@ -27,9 +29,7 @@ module.exports.System = registerSystem('tracked-controls', {
     }
   },
 
-  tick: function (time) {
-    if (time < this.lastControllersUpdate + 10) { return; }
-    this.lastControllersUpdate = time;
+  throttledTickHandler: function (time, delta) {
     this.updateControllerList();
     this.sceneEl.emit('controllersupdated', { timestamp: time, controllers: this.controllers });
   }
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -11,7 +11,8 @@ module.exports.System = registerSystem('tracked-controls', {
     var self = this;
     this.controllers = [];
     this.lastControllersUpdate = 0;
-    this.tick = utils.throttleTick(this.throttledTickHandler.bind(this), 10);
+    // Throttle the (renamed) tick handler to minimum 10ms interval.
+    this.tick = utils.throttleTick(this.throttledTick, 10, this);
     if (!navigator.getVRDisplays) { return; }
     navigator.getVRDisplays().then(function (displays) {
       if (displays.length > 0) {
@@ -29,7 +30,9 @@ module.exports.System = registerSystem('tracked-controls', {
     }
   },
 
-  throttledTickHandler: function (time, delta) {
+  // This tick handler will be throttled at init time.
+  // This method was renamed only to make it clearer e.g. when debugging
+  throttledTick: function (time, delta) {
     this.updateControllerList();
     this.sceneEl.emit('controllersupdated', { timestamp: time, controllers: this.controllers });
   }
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -30,8 +30,9 @@ module.exports.System = registerSystem('tracked-controls', {
     }
   },
 
-  // This tick handler will be throttled at init time.
-  // This method was renamed only to make it clearer e.g. when debugging
+  /**
+   * Update controller list every 10 millseconds.
+   */
   throttledTick: function (time, delta) {
     this.updateControllerList();
     this.sceneEl.emit('controllersupdated', { timestamp: time, controllers: this.controllers });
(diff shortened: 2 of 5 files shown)


donmccurdy:
Nice, thanks!

@ngokevin how would you feel about just naming this `utils.throttle`? That feels more consistent with common libs like `_.throttle`, and the function also be used for mouse or gamepad events.

dmarcos:
This is a `debounce` function in my little world. I don't know what it's the most common name. And yes, it should be generic and not specific to tick.

dmarcos:
I also believe that the tick should never be debounced but the function of interest that the tick calls

dmarcos:
If you debounce the tick and later change the component you might end up with a strange behaviour and then remember / rediscover that tick was being debounced

dmarcos:
In this case `updateControllerList` should be the pne debouced and documented on the method

donmccurdy:
@dmarcos To me, `debounce` implies that the function won't be invoked immediately, but only after the invocations pause for at least ___ms. I do think `throttle` is the right name (although perhaps there are uses for both functions).

[_.throttle](https://lodash.com/docs/4.17.2#throttle) vs [_.debounce](https://lodash.com/docs/4.17.2#debounce)

dmarcos:
@donmccurdy you're right `throttle` is the correct one

machenmusik:
so... `tick` has a specific signature `function (time, delta)` which is why calling this `throttleTick` makes sense, it won't work generically for functions that don't have compatible signatures or calling patterns.

machenmusik:
if one of you would like to make a quick mod to this that does not use the time values passed by `tick` but generically calls `Date.now()` itself, that may make more sense as a generic throttle...

machenmusik:
> If you debounce the tick and later change the component you might end up with a strange behaviour and then remember / rediscover that tick was being debounced

if you replace tick with function, still generically true.  not sure why you are specifically concerned with tick here

> In this case updateControllerList should be the pne debouced and documented on the method

IMO no, that may be called from other places that should not be throttled

donmccurdy:
Good point about the `time, delta` arguments. One solution would be to use an a-la-carte NPM implementation:

```
utils.throttle = require('lodash.throttle');
```

But, that would add a couple kilobytes to the build. If we don't expect to use the function internally, maybe it's better for developers just to NPM-require the function themselves.

Also seems easy enough to use `Date.now()` instead of relying on those `time, delta` args, and invoking `callback.apply(null, arguments)`.

dmarcos:
Yes, to throttle back we can use `Date.now()` to keep track on when the function was called last. I would not pull 3.6Kb of dependencies for such small piece of functionality. We can implement a simple generic throttle in our utils based on `Date.now()`

docs/core/utils.md
```diff
@@ -150,3 +150,18 @@ Checks if device is Gear VR. Returns a `boolean`.
 ### `AFRAME.utils.device.isMobile()`
 
 Checks if device is a smartphone. Returns a `boolean`.
+
+## Tick Utils
```
ngokevin: `Function Utils` makes sense

docs/core/utils.md
```diff
@@ -150,3 +150,18 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+
+### `AFRAME.utils.throttleTick(function (t, dt) {...}, interval)`
+
+Returns tick throttle function that only fires callback after minimum interval.
```
ngokevin: Note milliseconds

src/systems/tracked-controls.js
```diff
@@ -27,9 +29,7 @@ module.exports.System = registerSystem('tracked-controls', {
     }
   },
 
-  tick: function (time) {
-    if (time < this.lastControllersUpdate + 10) { return; }
-    this.lastControllersUpdate = time;
+  throttledTickHandler: function (time, delta) {
```
ngokevin: This doesn't actually need to be renamed right? In `init`: `this.tick =utils.throttle(utils.bind(this.tick, this))`

Might also be useful to allow the throttleTick do binding for you since we're not using closures: ```utils.throttleTick(this.tick, 10, this);```

src/utils/index.js
```diff
@@ -35,6 +35,24 @@ module.exports.isMobile = function () {
 };
 
 /**
+ * Returns tick throttle function that only fires callback after minimum interval.
+ * Note, this does not try to make up for any lost time by firing multiple callbacks.
+ * @param {callback}   The callback to fire; function (time, deltaTime) {...}
```
ngokevin: params should be `function` and `number`

src/utils/index.js
```diff
@@ -35,6 +35,24 @@ module.exports.isMobile = function () {
...
+ * @param {interval}   Interval between callbacks, in tick time units (milliseconds).
+ * @returns {function} Tick throttling handler function.
+ */
+module.exports.throttleTick = function (callback, interval) {
+  var data = {};
```
ngokevin: Does this need to be an object?

tests/utils/throttleTick.test.js
```diff
@@ -0,0 +1,77 @@
...
+    var dtlen = dts.length;
+    assert.equal(ts.length, dts.length);
+    throttleTickFn(interval);
+    assert.equal(ts.length, tlen + 1);
+    assert.ok(ts[tlen] - ts[tlen - 1] >= interval);
+    assert.equal(dts.length, dtlen + 1);
+    assert.ok(ts[tlen] - ts[tlen - 1] === dts[tlen]);
+  });
+
+  test('utils.throttleTick fires only one callback on multiply late ticks', function () {
```
ngokevin: multiple

tests/utils/throttleTick.test.js
```diff
@@ -0,0 +1,77 @@
+/* global assert, suite, test */
+var throttleTick = require('utils').throttleTick;
+
+suite('utils.throttleTick', function () {
+  var ts;
+  var dts;
+  var interval = 1000;
+  var throttleTickFn = throttleTick(function (t, dt) { ts.push(t); dts.push(dt); }, interval);
+
+  test('utils.throttleTick fires callback on first tick whatever', function () {
```
ngokevin: test names don't have to all start with `utils.throttleTick` since that's the suite name. can be:

```fires first callback on first tick```

File: docs/core/utils.md [hunk 8]
```diff
@@ -150,3 +150,18 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+```
+AFRAME.registerComponent('foo', {
+  tick: utils.throttleTick(function (t, dt) {
```
machenmusik: actually, it would be `// Called periodically, no more often than every 500 milliseconds`
there is no attempt to catch up, align or issue multiple calls to "catch up"

tests/utils/throttleTick.test.js
```diff
@@ -0,0 +1,77 @@
+/* global assert, suite, test */
+var throttleTick = require('utils').throttleTick;
+
+suite('utils.throttleTick', function () {
+  var ts;
```
ngokevin: rather than keeping track with arrays, you can use [sinon test spies](http://sinonjs.org/docs/#sinonspy): `var throttleSpy = this.sinon.spy(utils, 'throttleTick'`)` to assert how many times it was called, and its arguments.

machenmusik: no, throttleTick should only be called once, the actual callback is the one to look for, and this was easier...

tests/utils/throttleTick.test.js
```diff
@@ -0,0 +1,77 @@
...
+  var interval = 1000;
+  var throttleTickFn = throttleTick(function (t, dt) { ts.push(t); dts.push(dt); }, interval);
+
+  test('utils.throttleTick fires callback on first tick whatever', function () {
+    ts = [];
+    dts = [];
+    assert.equal(ts.length, 0);
+    assert.equal(dts.length, 0);
+    throttleTickFn = throttleTick(function (t, dt) { ts.push(t); dts.push(dt); }, interval);
+    throttleTickFn(987634578, 12345);
```
machenmusik: the specific goal was to make sure it works with arbitrary values.  declared variables to make that more clear

docs/core/utils.md
ngokevin: ````Called at most once every 500 millseconds.```

docs/core/utils.md
machenmusik: it is noted in the parameter description.

ngokevin: That isn't part of the docs

src/utils/index.js
machenmusik: `function` would be a reserved word clash... maybe `functionToThrottle`
`number` is less descriptive than `interval`, might be better to call `minimumInterval`

dmarcos: @ngokevin was refering to the types in the docs: 
````
@param {function} callback  The callback to fire; function (time, deltaTime) {...}
@param {number} interval The callback to fire; function (time, deltaTime) {...}
````

File: tests/utils/throttleTick.test.js (hunk 9 above)
Comment (machenmusik) [inline tests/utils/throttleTick.test.js]: I tried this, but function spies don't seem to do the same as method spies (i.e. when you call the function, the spy doesn't get called) -- so I'm giving up on this part for now, feel free to take it up.
``` Spy objects are objects returned from `sinon.spy()`. When spying on existing methods with `sinon.spy(object, method)`, the following properties and methods are also available on `object.method`. ```

Comment (ngokevin) [inline tests/utils/throttleTick.test.js]: OK will try.

**Comment by dmarcos** [comment]
Throttling the tick changes its behavior and will suprise devs in the future since it's not obvious that is behaving differently. I would just throttle those methods that need to be throttled (`updateControllerList ` in this case) down and let tick behave the way is expected. We can do:
````javascript
/**
 * This method is throttled down....
**/
updateControllerList: utils.throttle(function updateControllerList() { ... }, 100);
````

**Comment by machenmusik** [comment]
> updateControllerList: utils.throttle(function updateControllerList() { ... }, 100);

you are assuming, incorrectly, that the only caller to this function is the tick handler.  Consider that other callers may be extremely confused to find that their immediate call attempt (e.g. due to onGamepadConnected / Disconnected) is not made... the throttling should be confined to the intended surface, which is to say the periodic check should happen no more often than X.

BTW, if you only update every 100ms, then your controller pose and state is only updated every 100ms, which is not acceptable from a user standpoint.

**Comment by machenmusik** [comment]
ticks in general are not throttled by this, just this specific tick handler, which seems appropriate.  per your objection above, in systems tracked-controls I call the throttled handler throttledTick so it is obvious when debugging.

**Comment by ngokevin** [comment]
Thinking about it more, a more generic throttle would be easier to use for devs rather than a custom tickThrottle. It gives devs more control on what exactly to throttle. In that case:

```
init: function () {
  this.throttledUpdateControllersList = throttle(this.updateControllersList, 50);
},

updateControllersList: function () {},

tick: function () {
  this.throttledUpdateControllersList();
}
```

**Comment by machenmusik** [comment]
@dmarcos agree but original request was for @ngokevin common case of throttling all periodic handling for a tick handler, using the time passed in to avoid superfluous calls to get redundant time values, which is what the PR addresses.

**Comment by machenmusik** [comment]
(I have some slight concern that generic throttle will be abused and it will be difficult to understand why calls aren't being made when debugging.)

**Comment by dmarcos** [comment]
If we use the convention of naming the throttled method as `methodNameThrottled` there will be less space for confusion

**Comment by machenmusik** [comment]
I kinda like @ngokevin 's thought, but then it is annoying to not be able to see what throttledUpdateControllerList does.  Maybe this is more clear / readable (if it would work)...
```
throttledUpdateControllersList: throttle(function () { this.updateControllersList(); }, 10, this),

tick: function () {
  this.throttledUpdateControllersList();
}
```
But I am not sure that `this` would resolve correctly here.

**Comment by ngokevin** [comment]
I'm not too worried about the conventions and how we use it in `tracked-controls` system, it'll be up to the developer, so just thinking about the API (throttle vs. throttleTick).

I like `utils.throttle` flexibility, but I like not having to call `Date.now()` for each component for each frame. It's much tighter to use the timestamps given by `tick`. Though I haven't run into cases where I've only needed to throttle parts of a `tick`, it's always been all-or-nothing given that components are designed to do one thing and do it well.

**Comment by machenmusik** [comment]
@ngokevin agree, I left throttleTick in there and made a throttle as well.

case in point, the tick handler has to update and emit, so you have to make a throttledTick function that does exactly what is there anyway, and at that point you may as well use the more efficient form that gets time passed in.

File: docs/core/utils.md [hunk 11]
```diff
@@ -150,3 +150,38 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+### `AFRAME.utils.throttleTick(function (t, dt) {...}, minimumInterval [, optionalContext])`
+
+Returns tick throttle function that only fires call after minimum interval,
+in tick time units (milliseconds).
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+```
+...
```
ngokevin: Don't need the ellipsis (or make it a comment `// ...`)

docs/core/utils.md
```diff
@@ -150,3 +150,38 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+AFRAME.registerComponent('foo', {
+  // This function has the same signature as a normal tick handler,
+  // but has been renamed so it is obvious the function is throttled
+  // (for example, when debugging).
+  throttledTick: function (t, dt) {
```
ngokevin: I'd put the `tick` after the `init` to reflect the order of the lifecycle

docs/core/utils.md
```diff
@@ -150,3 +150,38 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+### `AFRAME.utils.throttleTick(function (t, dt) {...}, minimumInterval [, optionalContext])`
+
+Returns tick throttle function that only fires call after minimum interval,
+in tick time units (milliseconds).
+Note, this does not try to make up for any lost time by firing multiple callbacks.
+
+```
+...
+AFRAME.registerComponent('foo', {
+  // This function has the same signature as a normal tick handler,
```
ngokevin: Can shorten `Renamed tick handler to make it clear that it is throttled.`

docs/core/utils.md
```diff
@@ -150,3 +150,38 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+  // but has been renamed so it is obvious the function is throttled
+  // (for example, when debugging).
+  throttledTick: function (t, dt) {
+    // Called periodically, no more often than every 500 milliseconds
```
ngokevin: ```Called at most once every 500 milliseconds.`

docs/core/utils.md
```diff
@@ -150,3 +150,38 @@ Checks if device is Gear VR. Returns a `boolean`.
...
+  },
+
+  init: function () {
+    // Set up the tick throttling.
+    this.tick = AFRAME.utils.throttleTick(
+      // Because the function to throttle is an object method, bind this pointer
```
ngokevin: How come we can't use the `optionalContext` argument?

ngokevin: Would move the comments above this LOC to separate the comment/code

src/systems/tracked-controls.js
```diff
@@ -27,9 +30,9 @@ module.exports.System = registerSystem('tracked-controls', {
     }
   },
 
-  tick: function (time) {
-    if (time < this.lastControllersUpdate + 10) { return; }
-    this.lastControllersUpdate = time;
+  // This tick handler will be throttled at init time.
```
ngokevin: Can use docstring

```
/**
 * Update controller list every 10 millseconds.
 */
```

src/utils/index.js
```diff
@@ -35,6 +35,51 @@ module.exports.isMobile = function () {
 };
 
 /**
+ * Returns throttle function that only fires call after minimum interval.
+ * Note, this does not try to make up for any lost time by firing multiple callbacks.
+ * @param {functionToThrottle} The function to throttle; function (time, deltaTime) {...}
```
ngokevin: the jsdoc params should be JavaScript types

```
@param {function}
@param {number}
@param {object}
@returns {function}
```

docs/core/utils.md
machenmusik: I changed the style to comment each parameter, see if you like it better

machenmusik: because I hadn't added it when I did the docs initially... :-)

mikebolt:
This code looks good to me.

You may want to use a window.performance.now polyfill for more accurate timestamps. We wrote [a polyfill](https://github.com/tweenjs/tween.js/blob/master/src/Tween.js#L72) for window.performance.now in tween.js.

Also, this could be useful for users of tween.js. I might borrow the code and include it in tween.js as a utility.

machenmusik:
@mikebolt thanks, agree - @ngokevin is there a standard way to do in aframe?

should this throttle be published as a (tiny) separate npm module that aframe picks up,
to make it easier to reuse in TWEEN.js etc. ?

ngokevin:
@mikebolt Thanks for the review. We don't need window.performance.now since that provides sub-millisecond updates, which aren't needed in the context of VR.

You could publish the throttle, but would maybe prefer it to be kept in the repo for now.

ngokevin:
Thanks! Glad I don't have to do this anymore:

```
tick: (t, dt) => {
  if (t - this.time < 500) { return; }
  this.time = t;
}

**Comment by ngokevin** [comment]
Only nit (against @dmarcos concern) was we call the method `throttledTick` when it actually isn't throttled. It gets wrapped to `tick` which is the method that is actually throttled. Would have preferred a comment:

```
init: function () {
  this.tick = throttleTick(...);
},

/**
 * Note this method is wrapped to be throttled every 10ms in init().
 */
tick: function () {
  // ...
}
```

ngokevin:
Congrats on 2500th commit

PR #2191, opened 2016-12-19
Support new hand animations

machenmusik:
update to use new hand animations from arturitu via rawgit
TODO: update to aframe CDN URLs once published
TODO: unused poses

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -1,7 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 
-var OCULUS_LEFT_HAND_MODEL_URL = 'https://rawgit.com/arturitu/assets/e698020d6f335fb46a6169f8985b4e12297e157d/controllers/oculus-hands/leftHand.json'; // 'https://cdn.aframe.io/controllers/oculus-hands/leftHand.json';
-var OCULUS_RIGHT_HAND_MODEL_URL = 'https://rawgit.com/arturitu/assets/e698020d6f335fb46a6169f8985b4e12297e157d/controllers/oculus-hands/rightHand.json'; // 'https://cdn.aframe.io/controllers/oculus-hands/rightHand.json';
+var OCULUS_LEFT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/v2/leftHand.json';
+var OCULUS_RIGHT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/v2/rightHand.json';
 
 /**
 *
@@ -156,12 +156,6 @@ module.exports.Component = registerComponent('hand-controls', {
     return controllerId && controllerId.indexOf('Oculus Touch') === 0;
   },
 
-  isOculusTouchController: function () {
-    var trackedControls = this.el.components['tracked-controls'];
-    var controllerId = trackedControls && trackedControls.controller && trackedControls.controller.id;
-    return controllerId && controllerId.indexOf('Oculus Touch') === 0;
-  },
-
   determineGesture: function () {
     var gesture;
     var isGripActive = this.pressedButtons['grip'];
@@ -189,12 +183,12 @@ module.exports.Component = registerComponent('hand-controls', {
   },
 
   gestureAnimationMapping: {
-    '': 'Open',
-    'pointing': 'Point',
-    'pistol': 'Point + Thumb',
-    'fist': 'Fist',
-    'touch': 'Hold',
-    'thumb': 'Thumb Up'
+    default: 'Open',
+    pointing: 'Point',
+    pistol: 'Point + Thumb',
+    fist: 'Fist',
+    touch: 'Hold',
+    thumb: 'Thumb Up'
   },
 
   animateGesture: function (gesture) {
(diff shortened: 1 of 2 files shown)


src/components/hand-controls.js
```diff
@@ -244,9 +253,11 @@ module.exports.Component = registerComponent('hand-controls', {
     if (animationActive) { mesh.play(animationActive, 0); }
 
     // play new animation.
-    mesh.mixer.clipAction(animation).loop = 2200;
-    mesh.mixer.clipAction(animation).clampWhenFinished = true;
-    mesh.mixer.clipAction(animation).timeScale = timeScale;
+    clipAction = mesh.mixer.clipAction(animation);
+    if (!clipAction) { return; }
```
dmarcos: I prefer this to break loudly on a Q/A pass. If the animation is not present then `hand-controls` are not working as expected and I want it to be hard to miss. The hand model and this component will be always in sync.

src/components/hand-controls.js
```diff
@@ -1,7 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 
-var OCULUS_LEFT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/leftHand.json';
-var OCULUS_RIGHT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/rightHand.json';
+var OCULUS_LEFT_HAND_MODEL_URL = 'https://rawgit.com/arturitu/assets/e698020d6f335fb46a6169f8985b4e12297e157d/controllers/oculus-hands/leftHand.json'; // 'https://cdn.aframe.io/controllers/oculus-hands/leftHand.json';
+var OCULUS_RIGHT_HAND_MODEL_URL = 'https://rawgit.com/arturitu/assets/e698020d6f335fb46a6169f8985b4e12297e157d/controllers/oculus-hands/rightHand.json'; // 'https://cdn.aframe.io/controllers/oculus-hands/rightHand.json';
```
dmarcos: Let's point these URLs to the proper CDN so we can merge it.

dmarcos: I'll update the models in the CDN before merge

src/components/hand-controls.js
machenmusik: great for QA, not great for people actually using it.
Maybe spew warnings / errors to console to get the required attention?

src/components/hand-controls.js
```diff
@@ -179,21 +185,23 @@ module.exports.Component = registerComponent('hand-controls', {
   },
 
   gestureAnimationMapping: {
-    'pointing': 'pointing',
-    'pistol': 'pistol',
-    'fist': 'press',
-    'touch': 'touch',
-    'thumb': 'thumb'
+    '': 'Open',
+    'pointing': 'Point',
```
machenmusik: well, the empty string does... is there any harm in leaving them for consistency?

src/components/hand-controls.js
machenmusik: for backwards compatibility, I strongly suggest making a new CDN URL for the new files, to avoid the breakage issue we had earlier today.

if you wouldn't mind updating these once you make the new CDN URLs, that would be great, thx!

src/components/hand-controls.js
dmarcos: The empty string is the default. An empty string as a key feels weird to me.

src/components/hand-controls.js
dmarcos: I don't want to confuse people with multiple versions of the same model for each tiny change we make. I don't think this model has that much use to justify keeping the old one.

src/components/hand-controls.js
machenmusik: right, the default (no gesture) may still have a named animation associated (as it does here)

src/components/hand-controls.js
machenmusik: different perspective, the change makes it more forgiving for users to take as example and replace with their own model -- instead of "why does the controller no longer move or react" it becomes "why do my animations not work" as it should be.  But I do understand your perspective.  Would you prefer the early-return that masks errors to be commented and disabled?

src/components/hand-controls.js
machenmusik: but then you have to coordinate the CDN change with global replacement of code everywhere that references the old version... which seems unlikely to me.  this is just like the difference between cdn.rawgit.com and rawgit.com - when you publish the URLs to production, the assumption is that you won't break things by destructive changes to required CDN assets.

machenmusik: to use a recent example, we changed the URL for oculus-compatible hands rather than overwriting the old ones, which would have broken hand-controls usage across all of a-frame...

dmarcos: The change is really exceptional. I really don't expect assets to change once published.

src/components/hand-controls.js
```diff
@@ -244,9 +253,15 @@ module.exports.Component = registerComponent('hand-controls', {
...
     // play new animation.
-    mesh.mixer.clipAction(animation).loop = 2200;
-    mesh.mixer.clipAction(animation).clampWhenFinished = true;
-    mesh.mixer.clipAction(animation).timeScale = timeScale;
+    clipAction = mesh.mixer.clipAction(animation);
+    // returning when no clipAction will prevent further issues
+    // (e.g. controllers no longer updating pose)
+    // but per https://github.com/aframevr/aframe/pull/2191#discussion_r93121878
+    // the preference is to NOT prevent the issues to catch bugs earlier in QA.
+    // if (!clipAction) { return; }
```
dmarcos: Remove line

src/components/hand-controls.js
machenmusik: They definitely shouldn't (also so you don't have to deal with flushing older versions from all the caches in the Internet).  

If you don't use different URLs, then you have to attempt to update the CDN assets and simultaneously the referencing code (and all uses of it, which seems unreasonable given encouragement to pull versions via npm eather than using direct URL).

You've offered to to update the CDN assets and then update the code, so if that is really what you'd like to do, feel free -- but then we are likely to face spurious issues as we did before reverting the prior hands change.

dmarcos: if you change the url back I can update the CDN and merge this PR simultaneously.

machenmusik: ok, see latest commit

machenmusik:
to avoid spurious conflicts I think you'll want to merge this one before #2192

machenmusik:
may want to do #2192 first to avoid spurious conflicts, as it uses the hand animation names that are then updated as part of this one

dmarcos:
needs rebase

machenmusik:
rebased but not yet double-checked.  still recommend new CDN URLs to avoid breaking npm builds that already pulled down old code...

machenmusik:
ran this last night looked ok on Touch...

machenmusik:
I am guessing this one may be waiting for point release

machenmusik:
tested working with latest experimental Chromium and Oculus Touch;
using Nightly plus Vive, gamepad id has changed from `OpenVR Gamepad` to `OpenVR Controller`

machenmusik:
@dmarcos completely unrelated travis-ci test failure after squash...

machenmusik:
... and this time travis is happy.

This PR is now necessary to accommodate both old and new Vive controller change in gamepad id string.

machenmusik:
29c1364 is now necessary to accommodate both old and new Vive controller change in gamepad id string.

machenmusik:
separating the old-and-new Vive support into
https://github.com/aframevr/aframe/pull/2370

PR #2193, opened 2016-12-20
Fix runtime error in `embedded` component when `vr-mode-ui` component is disabled

RSpace:
**Description:**

It should be possible to use A-Frame embedded without the VR mode UI enabled. After all, embedding often means a heavily customized solution including custom UI to enter and leave VR.

**Changes proposed:**

- Only add and remove `embedded` class from Enter VR element if it exists in the DOM
- Added test for `embedded` component, which tests the component's ability to work both when the `vr-mode-ui` component is enabled and disabled.

### DIFF
--- a/src/components/scene/embedded.js
+++ b/src/components/scene/embedded.js
@@ -12,10 +12,10 @@ module.exports.Component = registerComponent('embedded', {
     var sceneEl = this.el;
     var enterVREl = sceneEl.querySelector('.a-enter-vr');
     if (this.data === true) {
-      enterVREl.classList.add('embedded');
+      if (enterVREl) { enterVREl.classList.add('embedded'); }
       sceneEl.removeFullScreenStyles();
     } else {
-      enterVREl.classList.remove('embedded');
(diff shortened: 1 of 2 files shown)


src/components/scene/embedded.js
```diff
@@ -4,18 +4,16 @@ var registerComponent = require('../../core/component').registerComponent;
...
-  dependencies: ['vr-mode-ui'],
-
   schema: {default: true},
 
   update: function () {
     var sceneEl = this.el;
     var enterVREl = sceneEl.querySelector('.a-enter-vr');
     if (this.data === true) {
-      enterVREl.classList.add('embedded');
+      if (enterVREl) { enterVREl.classList.add('embedded'); }
```
dmarcos: The problem si that it's not guaranteed that the enterVREl element it's going to be ready when the `update` method runs, hence the `vr-mode-ui` dependency

dmarcos: Yeah. If keeping the dependency works for you we keep the two checks

dmarcos:
Thanks!

Issue #2190, opened 2016-12-19
vive controls not tracked in chrome


noerw:
In current chromium the vive controls do not get tracked anymore using the `vive-controls` component:
They are rendered at their correct position at pageload (!), but do not move.

- There is no error in the console.
- It works in Firefox with *WebVR Plus* extension.
- The controls did work for me with aframe 0.3.2.

Did not test whether this also occurs with the `tracked-controllers` component..

- A-Frame Version: `0.4.0`
- Platform/Device: HTC Vive, chromium 53
- [codepen](https://codepen.io/anon/pen/PbVdbB)

dmarcos:
@machenmusik doest it ring a bell?

machenmusik:
a few comments...
- I thought, although I could be wrong, that chromium 53 is too old for webvr etc. ?
- I suspect that the issue may be due to this: because of the multiple controller support, each of the individual components (`oculus-touch-controls` and `vive-controls`) do not inject a default `tracked-controls` instance automatically i.e. do not have `dependencies: ['tracked-controls']` -- as well as a change to default `tracked-controls` values not to auto-select the first OpenVR Gamepad which was causing difficulties with Oculus Touch integration.

machenmusik:
@dmarcos @noerw i was unable to reproduce this, however I note that this is precisely the behavior that occurred when the `hand-controls` models were changed this morning without https://github.com/aframevr/aframe/pull/2188 ... the missing animations caused errors, and position was apparently not being updated. I see that PR is closed, maybe we do actually want it after all...

dmarcos:
I reverted the model from the https://github.com/aframevr/assets this is probably why the problem cannot be reproduced anymore. @noerw can you confirm that everything is working? Please reopen the issue if you can still reproduce the problem.

noerw:
I tested again, and I can still reproduce it.
@machenmusik Your proposed `tracked-controls` fixes did not work.

However, I told you some wrong facts yesterday:
- running chromium 56.0.2902 x64
- the following appears in the console twice. In Firefox directly on pageload, in Chromium once i enter VR mode:
```
components:obj-model:warn Material component properties are ignored when a .MTL is provided (aframe.js:327:8)

THREE.MTLLoader: .setBaseUrl() is deprecated. Use .setTexturePath( path ) for texture path or .setPath( path ) for general base path instead (aframe.js:49259:3)
```

machenmusik:
actually i do believe the model changed.

[...]

PR #2194, opened 2016-12-20
update syntax for vive-controls model for 0.4.0

machenmusik:
Fixes https://github.com/aframevr/aframe/issues/2190


machenmusik:
@dmarcos @ngokevin looks like 0.4.0 model syntax changed and broke prior `vive-controls` model; for those that were using its built-in model, this is an important fix

dmarcos:
Thanks!

PR #2197, opened 2016-12-21
[docs] Fix material properties

jsantell:
For distortion map properties such as `displacementScale`, `displacementBias` and `normalScale`


jsantell:
This affects docs back to v0.3.1 I believe

ngokevin:
Thanks, we can cherry-pick this onto the docs

Issue #2159, opened 2016-12-06
Loading textures for custom shaders in registerShader schema


alfa256:
**Description:**
Right now there is no way to register a src to be loaded as a texture to be sampled by a custom shader.

Something like this or any variation would be nice:

```
    diffusetex: { type: 'src', is: 'uniform' }
    normalmap: { type: 'src', is: 'uniform' }
```

- A-Frame Version: 0.4.0
- Platform/Device: any

This might be also related to #2147

donmccurdy:
To do this, I think we just need the shader registration to recognize an additional type for textures. Specifically THREE.Texture → 't' in THREE.ShaderMaterial uniforms.

machenmusik:
yes, or you can replace the default init/update handling and handle uniforms yourself.  will see if i can post an example

machenmusik:
```
    // this shader uses offset and repeat

    AFRAME.registerShader('offset-repeat', {
        schema: {
            // the texture source (probably a video)
            src: {type: 'map'},
            // texture parameters
            offset: {type: 'vec2', default: {x: 0, y: 0}},
            repeat: {type: 'vec2', default: {x: 1, y: 1}}
        },

        uniforms: {
            // the texture value (once the texture source is loaded, update)
            map: { type: 't', value: null },
            // texture parameters
            offset: {type: 'vec2', value: {x: 0, y: 0}},
            repeat: {type: 'vec2', value: {x: 1, y: 1}}
        },

        init: function(data) {

            this.material = new THREE.ShaderMaterial({

                uniforms: this.uniforms,

                vertexShader: this.vertexShader,

                fragmentShader: this.fragmentShader

            });
        },



        update: function (data) {
            AFRAME.utils.material.updateMap(this, data);
            this.uniforms.map.value = this.material.map;
            this.uniforms.offset.value = data.offset;
            this.uniforms.repeat.value = data.repeat;
        },



        vertexShader: [
          'varying vec2 vUV;',
          'void main(void) {',
          '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
          '  vUV = uv;',
          '}'
        ].join('\n'),

        fragmentShader: [
          'uniform sampler2D map;',
          'uniform vec2 offset;',
          'uniform vec2 repeat;',
          'varying vec2 vUV;',
          'void main() {',
[...]

Issue #2032, opened 2016-10-31
allow configure textures, separate texture component from material


ngokevin:
**Description:**

Lots of properties for THREE.Texture at https://threejs.org/docs/?q=texture#Reference/Textures/Texture , how to expose these in a clean way for each kind of map?

Maybe a texture property type?

```
material="map: [src: #texture; anisotropy: 1.1; flipY: true]"
```

Or a indirect route?

```
material="map: #texture__map"
texture__map="src: #texture; anisotropy: 1.1; flipY: true"
```

ngokevin:
Looking at https://aframe.io/docs/master/components/material.html#built_in_materials and how there are multiple map types trying to expose every texture properties for each one, it seems necessary for a separate texture component. @dmarcos toyed with this before.

bryik:
In Three.js, "renderer.getMaxAnisotropy()" makes it easy to set anisotropy to the max supported value. I wasn't sure how to do this in A-Frame and often set it to 16 (the value for my GPU). Something like ```anisotropy: max;``` or ```anisotropy: auto;``` would be nice.

[Anisotropy example (16x vs default)](https://bl.ocks.org/bryik/a82893253ea55f401af114778bc5453b).

ngokevin:
I think this will be important as currently, we have to create a different property on the material component for each type of map, duplicating a lot of the texture setup. The standard shader has:

- ambientOcclusionMap
  - intensity
  - repeat
  - offset
- displacementMap
  - intensity
  - repeat
  - offset
- normalMap
  - intensity
  - repeat
  - offset
- map (src)
  - intensity
  - repeat
  - offset

Then what if we want to add a new texture property, like configure the wrapping or flip? We'd have to add it to each single one, and in the code, we'd have to duplicate logic.

A separate texture/map component would be cleaner. It could have its own system for caching. You just define a texture component, and reference from the material (or any other) component.

```material="map: [map__texture]" map__texture="src: #textureImg; anisotropy: 1.1; flipY: true"```

And alternative, allow inline definition


```material="map: [src: #textureImg; anisotropy: 1.1; flipY: true]"```

**Comment by donmccurdy** [comment]
I'm not 100% sure whether I like this any better, but for discussion's sake here's a counter-proposal. No integration with the existing `material` component, just a separate `texture-config` component that modifies properties of the object it's attached to, whether there's a `material` component or a model of some kind.

```
<a-entity obj-model="obj: url(...); mtl: url(...);"
[...]

PR #2199, opened 2016-12-21
fix registerShader handling of texture variables

machenmusik:
texture variables with is:'uniform' and type:'map' now work with default init and update handlers

so now you can just do things like this:
```
    AFRAME.registerShader('offset-repeat', {
        schema: {
            // the texture source (probably a video)
            src: {type: 'map', is: 'uniform'},
            // texture parameters
            offset: {type: 'vec2', default: {x: 0, y: 0}, is: 'uniform'},
            repeat: {type: 'vec2', default: {x: 1, y: 1}, is: 'uniform'}
        },

        vertexShader: [
          'varying vec2 vUV;',
          'void main(void) {',
          '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
          '  vUV = uv;',
          '}'
        ].join('\n'),

        fragmentShader: [
          'uniform sampler2D map;',
          'uniform vec2 offset;',
          'uniform vec2 repeat;',
          'varying vec2 vUV;',
          'void main() {',
          '  gl_FragColor = texture2D(map, vec2(vUV.x / repeat.x + offset.x, vUV.y / repeat.y + offset.y));',
          '}'
        ].join('\n')
    });
```


ngokevin:
I'd like to consider https://github.com/aframevr/aframe/issues/2047

machenmusik:
The existing Shader.prototype, with these changes, will handle some of the annoying grunt work of updating values for attributes and uniforms, but will let you do it yourself if you leave out e. g. `is:'uniform'`

IMO this PR has a lot of benefit for those 99.99% of us that are not hardcore shader people (e.g. I don't need any custom stuff anymore for the various shaders I did now that there is basic texture support)

Nothing stops someone from doing their own more direct ShaderMaterial usage with this merged... so for more advanced users, they can always go ahead and do as suggested in #2047, and if they are using values with supported types, they can use the convenience mappings.

machenmusik:
#2047 may simply be an example of how to use more directly, maybe the ones mentioned in Slack, and zero code change

ngokevin:
There's no difference in difficulty between the current Shader API and using ShaderMaterial directly. We are tightly wrapping the API, routing vertexShader + fragmentShader to ShaderMaterial.vertexShader + ShaderMaterial.fragmentShader and mapping the schema to uniforms. It's actually more difficult because you will run into limitations and it's not as documented as three.js's.

It's not a good idea to support two APIs, especially when one is limited and always has to be kept up to date. Using ShaderMaterial is much more solid. You will be able to copy and paste three.js ShaderMaterial examples straight in rather than trying to figure out the relationships to A-Frame, and there is less risk of API regressions.

ngokevin:
The grunt work of updating uniforms could be handled by a util or just boilerplate.

machenmusik:
for the simple shaders I am using, it's actually quite convenient.
e.g. here is the previous trivial example, which contains a lot of useless busywork
```
    // this shader uses offset and repeat
    AFRAME.registerShader('offset-repeat', {
        schema: {
            // the texture source (probably a video)
            src: {type: 'map'},
            // texture parameters
            offset: {type: 'vec2', default: {x: 0, y: 0}},
            repeat: {type: 'vec2', default: {x: 1, y: 1}}
        },

        uniforms: {
            // the texture value (once the texture source is loaded, update)
            map: { type: 't', value: null },
            // texture parameters
            offset: {type: 'vec2', value: {x: 0, y: 0}},
            repeat: {type: 'vec2', value: {x: 1, y: 1}}
        },

        init: function(data) {
            this.material = new THREE.ShaderMaterial({
                uniforms: this.uniforms,
                vertexShader: this.vertexShader,
                fragmentShader: this.fragmentShader
            });
        },



        update: function (data) {
            AFRAME.utils.material.updateMap(this, data);
            this.uniforms.map.value = this.material.map;
            this.uniforms.offset.value = data.offset;
            this.uniforms.repeat.value = data.repeat;
        },



        vertexShader: [
          'varying vec2 vUV;',
          'void main(void) {',
          '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
          '  vUV = uv;',
          '}'
        ].join('\n'),

        fragmentShader: [
          'uniform sampler2D map;',
          'uniform vec2 offset;',
          'uniform vec2 repeat;',
          'varying vec2 vUV;',
          'void main() {',
          '  gl_FragColor = texture2D(map, vec2(vUV.x / repeat.x + offset.x, vUV.y / repeat.y + offset.y));',
          '}'
        ].join('\n')
    });
```
and here it is after the PR changes, which is what I expected:
```
    // this shader uses offset and repeat
    AFRAME.registerShader('offset-repeat', {
        schema: {
            // the texture source (probably a video)
            src: {type: 'map', is: 'uniform'},
            // texture parameters
            offset: {type: 'vec2', default: {x: 0, y: 0}, is: 'uniform'},
            repeat: {type: 'vec2', default: {x: 1, y: 1}, is: 'uniform'}
        },

        vertexShader: [
          'varying vec2 vUV;',
          'void main(void) {',
          '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
          '  vUV = uv;',
          '}'
        ].join('\n'),

        fragmentShader: [
          'uniform sampler2D map;',
          'uniform vec2 offset;',
          'uniform vec2 repeat;',
          'varying vec2 vUV;',
          'void main() {',
          '  gl_FragColor = texture2D(map, vec2(vUV.x / repeat.x + offset.x, vUV.y / repeat.y + offset.y));',
          '}'
        ].join('\n')
    });
```

machenmusik:
utils / boilerplate functions already exist - they are used internally by the default init and update functions
```
  init: function (data) {
    this.attributes = this.initVariables(data, 'attribute');
    this.uniforms = this.initVariables(data, 'uniform');
    this.material = new THREE.ShaderMaterial({
      // attributes: this.attributes,
      uniforms: this.uniforms,
      vertexShader: this.vertexShader,
      fragmentShader: this.fragmentShader
    });
    return this.material;
  },
```
and
```
  update: function (data) {
    this.updateVariables(data, 'attribute');
    this.updateVariables(data, 'uniform');
  },
```
I don't really see a need to remove them, or to further augment them beyond this PR.

The utils handling in this PR to expose the pre-existing texture support for custom shaders is *extremely* useful.

#2047 is probably a simple example of doing those and additionally manipulating your own variables (I don't have anything that needs it, but I think @donmccurdy found one) - zero code change.

ngokevin:
I think I prefer accepting the few additional lines (that could be resolved with utils, it's just updating an object) that comes from A-Frame trying to hide the three.js API. The maintenance of this API doesn't end with adding textures. It might resolve your simple shader, but there are way too many cases out there where you'll just run into a wall (what about `samplerCube`?).

See all of the [uniform types](https://github.com/mrdoob/three.js/wiki/Uniforms-types) we'd need to create A-Frame-to-three.js mappings and property types Every time three.js updates its API, A-Frame will have to answer. I understand it saves a bit of work in some cases, but in others it seems fragile and will cause some frustration.

dmarcos:
@ngokevin I don't think that API will change that much in three.js. At the end they're matching GLSL types and that's settled.

machenmusik:
I don't think that having either what is there now, or this PR, would imply or require full support for all THREE types.  There is a clear pattern to follow if others want to add support for other types (including cubemaps, now that the texture handling is figured out for them and all that is needed is the right entries in the property type mapping table)

machenmusik:
without this PR it is rather tricky to figure out what is going on with the texture mapping (including the oddball cases where the vestiges of the whole src-or-element issue decide they will try and make a copy of your video and you start hearing double)

ngokevin:
The root issue is more code/API to maintain versus the benefit. If we propose an API that is an alternative to an existing one, it shouldn't have holes. We could pile on more code and features, but that increases maintenance versus just pointing to ShaderMaterial and leaving it at that. Every feature has a long-term cost.

I'm not sure what you mean with the texture mapping. You're using the `map` property type either way.

machenmusik:
without this PR, the `map` property type doesn't work correctly ;-)

yeah, clearly it is down to
> more code/API to maintain versus the benefit

I agree with @dmarcos that the surface area of the API is not likely to change radically;
the benefit is being able to do many common custom shaders easily, so IMO it's worth it.

re: #2047 -- 
the example you posed is non-functional and lacks how you would propose literally any parameters be populated.  
```
AFRAME.registerShader('myshader', {  // or registerMaterial to map to three.js API
  schema: {},

  init: function () {
    this.material = new ShaderMaterial({
      // ...
    });
  }
});```
if you are going to leave that as an exercise to the reader, that seems a large step backwards from either what is there now or this PR.  

As things stand right now without this PR, the necessary utility functions are not usable for other purposes besides the built-in `flat` and `standard` shaders that expect `data.src` to become `material.map`

With this PR, that API can easily do simple shaders for common needs ("e.g. how do I offset a texture") as well as less common needs such as 2D image / video manipulation.  The latter especially is of significant benefit.

**Comment by ngokevin** [comment]
OK. I thought about it. Even if we prescribe ShaderMaterial, the same amount of work is needed to support the schema. It would be nice to hide the three.js uniform types and use A-Frame schema types instead. I'll punt this for now, I hope we have volunteers to fill in the rest of the property types (there is a PR open for Matrix4) and support the rest of the ShaderMaterial properties. 

PS: The `flat` and `standard` materials support simple texture offsets and repeats.

File: src/core/propertyTypes.js [hunk 1]
```diff
@@ -72,6 +72,8 @@ function assetParse (value) {
   var el;
   var parsedUrl;
 
+  if (typeof value !== 'string') { return value; }
```
ngokevin: I would add a comment when this would be the case (e.g., a MediaElement?)

src/core/shader.js
```diff
@@ -61,18 +62,19 @@ Shader.prototype = {
...
     var schema = this.schema;
     var schemaKeys = Object.keys(schema);
     schemaKeys.forEach(processSchema);
     function processSchema (key) {
       if (schema[key].is !== type) { return; }
       var varType = propertyToThreeMapping[schema[key].type];
-      var varValue = schema[key].parse(data[key] || schema[key].default);
-      variables[key] = {
+      var varKey = key;
+      // For textures, by convention data.src becomes uniform sampler2D map.
```
ngokevin: I'm not a fan of this magical mapping. If it wants to become a `sampler2D`, it should have a `map` property type.

machenmusik:
OK so we are leaving this open for now pending more property types / ShaderMaterial properties?  if so, would be good to have a list here for review later

ngokevin:
We can file an issue for what's missing. There's also #api on Slack if we want to take it there.

src/core/shader.js
```diff
@@ -96,10 +98,30 @@ Shader.prototype = {
     var schema = this.schema;
     dataKeys.forEach(processData);
     function processData (key) {
+      var varKey = key;
+      var materialKey = '_texture_' + key;
```
ngokevin: What's the prefix for?

src/utils/material.js
```diff
@@ -4,31 +4,42 @@
...
 };
 
 /**
+ * Update `material.map` given `data.src`. For standard and flat shaders.
+ *
+ * @param {object} shader - A-Frame shader instance.
+ * @param {object} data
+ */
+module.exports.updateMap = function (shader, data) {
+  return module.exports.updateMapMaterialFromData('map', 'src', shader, data);
```
ngokevin: This seems too hardcoded, `src` name shouldn't be anything special. Is this only called in two places?

ngokevin:
Related https://github.com/aframevr/aframe/issues/2032 ... the texture loading pieces are a bit of a mess

src/core/shader.js
machenmusik: the `_texture_` prefix is to avoid namespace clashes on the material properties.

src/utils/material.js
machenmusik: that's how the existing flat and standard shaders work, so wanted to accommodate them (and minor modificiations to their shaders)

machenmusik:
re: #2032 -- can the schema code handle properties that are objects/maps (to hold texture parameters) ?

machenmusik:
OK so the codecov thing now is saying
 81.53% (Diff 100%)

ngokevin:
With the extension, in the Diff view, you can see all lines that aren't covered. So I'd see a lot of Red next to the lines that were added, meaning a lot of incoming code that's not unit tested.

machenmusik:
... hampered by, as I understand it, a general lack of tests at all in this area prior to this PR as well?  same thing I ran into with vive-controls earlier.  if there are some minimal criteria (like the semistandard checking) it would be good to add them earlier in the pipeline, it's not entirely reasonable to enforce different standards on external submissions

ngokevin:
Apologies, tests should have been added to these pieces of code previously. Nevertheless, it's a good thing to have. But we can worry about reviewing the PR itself first.

machenmusik:
Did some tests, code coverage looks good now.  OK to merge and open issue for remaining property types?

dmarcos:
I'll look at this tomorrow or friday

machenmusik:
@dmarcos let me know if anything else is needed here

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
+/* global AFRAME */
+/* global assert, process, setup, suite, test, teardown */
+var entityFactory = require('../helpers').entityFactory;
+var componentName = 'shader';
```
ngokevin: this doesn't need to be a var

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
+/* global AFRAME */
```
ngokevin: don't have to separate these two global statements

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
...
+/* global assert, process, setup, suite, test, teardown */
+var entityFactory = require('../helpers').entityFactory;
+var componentName = 'shader';
+
+var VIDEO = 'https://ucarecdn.com/bcece0a8-86ce-460e-856b-40dac4875f15/'; // 'base/tests/assets/test.mp4';
```
ngokevin: does this need to be external, why not just use the local file?

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
...
+
+function isEmpty (map) {
```
ngokevin: You could just do `!Object.keys(map).length` without needing a separate fn that's only called twice

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
...
+  for (var key in map) {
+    return !map.hasOwnProperty(key);
+  }
+  return true;
+}
+
+suite(componentName, function () {
+  setup(function (done) {
+    this.el = entityFactory();
```
ngokevin: i'd wait for it to load.

`el.addEventListener('loaded', function () { done(); });`

tests/core/shader.test.js
```diff
@@ -0,0 +1,261 @@
...
+    done();
+  });
+
+  teardown(function (done) {
+    var el = this.el;
+    process.nextTick(function () {
```
ngokevin: can we clean the test file of all the `process.nextTick()`s and `done()` callbacks? 

they're only needed if something asynchronous happens. you also don't need the `done()` callback if nothing async happens.

src/core/propertyTypes.js
```diff
@@ -72,6 +72,9 @@ function assetParse (value) {
   var el;
   var parsedUrl;
 
+  // If an element was provided (e.g. canvas or video), just return it.
+  if (typeof value !== 'string') { return value; }
```
ngokevin: **Later on** we're going to need to do something about returning either a string or an element. Property types shouldn't be returning two different types of things.

src/core/shader.js
```diff
@@ -61,18 +62,19 @@ Shader.prototype = {
...
       var varType = propertyToThreeMapping[schema[key].type];
-      var varValue = schema[key].parse(data[key] || schema[key].default);
-      variables[key] = {
+      var varKey = key;
+      // For textures, by convention data.src becomes uniform sampler2D map.
+      if (varKey === 'src' && type === 'uniform' && varType === 't') { varKey = 'map'; }
+      variables[varKey] = {
         type: varType,
-        value: self.parseValue(schema[key].type, varValue)
+        value: undefined // let updateVariables handle setting these
```
ngokevin: For comments, i like having two spaces after the statement, with caps + period.

```
undefined  // Let update.
```

src/core/shader.js
```diff
@@ -96,8 +98,32 @@ Shader.prototype = {
     var schema = this.schema;
     dataKeys.forEach(processData);
     function processData (key) {
+      var varKey;
+      var materialKey;
       if (!schema[key] || schema[key].is !== type) { return; }
-      if (variables[key].value === data[key]) { return; }
+      if (schema[key].type === 'map') {
+        // Special handling is needed for textures.
+        // For textures, by convention data.src becomes uniform sampler2D map.
```
ngokevin: We shouldn't this convention. `src` doesn't have any special meaning

src/core/shader.js
```diff
@@ -96,8 +98,32 @@ Shader.prototype = {
...
+          varKey = 'map';
+          materialKey = 'map';
+        } else {
+          varKey = key;
+          materialKey = '_texture_' + key;
+        }
+        // If data unchanged, get out early.
+        if (!variables[varKey] || variables[varKey].value === data[key]) { return; }
+        // We can't actually set the variable correctly until we've loaded the texture.
+        self.el.addEventListener('materialtextureloaded', function (e) {
```
ngokevin: Why do we have to go through the material util and thus the material system for setting a `t` uniform value?

Things are a bit complicated, and I'm looking to refactor:

- Separate texture system/component (somehow)
- Asset property types to always return the same type. e.g., a promise that resolves XHR results

For here to be simple, should we just use the `TextureLoader`?

tests/core/shader.test.js
machenmusik: I know, but i did that as boilerplate so for the next component (since I did a series of them this way) I could clone the pattern and not have to keep changing "shader" with replace-all which is error-prone.

tests/core/shader.test.js
machenmusik: what local file?  I copied the url from a similar pre-existing test...

src/core/propertyTypes.js
machenmusik: ok, but returning string is insufficient.

src/core/shader.js
machenmusik: the two existing shaders - standard and flat - both do this goofy src to map translation.  if there are other examples that don't, this can probably go away

src/core/shader.js
machenmusik: the idea was to do the same thing as the existing shaders (standard and flat), to avoid introducing new and exciting errors.

ngokevin: I have plans on refactoring them. The existing shaders and these shaders have implementations. The texture loader in the material system expects the textures to follow certain properties, such as have offsets, repeats, which are implemented by three.js. If we refactor, we'd rather not have to think about this tangle.

src/core/scene/a-scene.js
```diff
@@ -449,6 +457,25 @@ module.exports = registerElement('a-scene', {
         var effect = this.effect;
         var timeDelta = time - this.time;
 
+        // Experimental versions of Chromium used for WebVR have a known issue
```
ngokevin: Should be part of this PR?

src/core/shader.js
```diff
@@ -96,8 +95,22 @@ Shader.prototype = {
...
+        materialKey = '_texture_' + key;
+        // If data unchanged, get out early.
+        if (!variables[key] || variables[key].value === data[key]) { return; }
+        // We can't actually set the variable correctly until we've loaded the texture.
+        self.el.addEventListener('materialtextureloaded', function (e) {
+          variables[key].value = self.material[materialKey];
+          variables[key].needsUpdate = true;
+        });
+        // Kick off the texture update now that handler is added.
+        utils.material.updateMapMaterialFromData(materialKey, key, self, data);
```
ngokevin: Can we simply and use texture loader?

src/core/shader.js
machenmusik: understood, but do we have to refactor here?  re:
> have offsets, repeats, which are implemented by three.js
they are quasi-standardized parameters that get passed through, 
but the (custom) shaders can freely ignore the values, as sometimes they are irrelevant.

the point of this particular PR is limited in scope, namely to make textures usable by the existing mechanism for custom shaders, without breaking the standard and flat legacy shaders.

src/core/scene/a-scene.js
machenmusik: aargh must have started from a branch that had this, will back it out (second one today)

src/core/shader.js
ngokevin: We don't have to refactor here, but it'll make it simpler and easy to reason later if we try a straight up TextureLoader here rather than jumping into another pile of code.

src/core/shader.js
machenmusik: not sure how/whether safe?  I can confirm this code works, but that testing is all invalid with something else

ngokevin: The unit tests should still be valid, right? The result will still be the same, the texture is set on the material.

machenmusik: sure, but I am not sure what texture loader you are asking to use so I am not sure how to execute, is what I'm saying.

ngokevin: https://threejs.org/docs/#Reference/Loaders/TextureLoader

You pass a URL, and get a texture object back.

```new THREE.TextureLoader().load(url, function (texture) { // ... });```

machenmusik: um... doesn't that break the events contract (which is why I went through utils/material) ?

ngokevin: I'd just emit it after the texture is loaded

machenmusik: it isn't clear to me that will work since not all cases here are just a string url as previously discussed.  if you have some time to try that technique and have video srcObject, canvas, etc. still work, go ahead and take a crack at it, but if not I would like you to consider merging this as-is for now and taking on texture system refactor for 0.5.0

ngokevin: OK, I didn't consider the media elements and canvas. That can be refactored in later.

machenmusik:
thanks!

machenmusik:
does this solve #2159

PR #2200, opened 2016-12-21
add bmfont-text component and a-text primitive to core

machenmusik:
add bmfont-text component and a-text primitive to core from https://github.com/chenzlabs/aframe-bmfont-component


dmarcos:
This is a modification of @fernandojsg component and there's a PR opened there(https://github.com/fernandojsg/aframe-bmfont-component/pull/1). I'll wait for fernando's review first.

machenmusik:
yep, makes sense, the PR to @fernandojsg 's version is mine and incorporated here

machenmusik:
I have merged the output from @fernandojsg component PR review into this PR.

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    side: {default: 'front', oneOf: ['front', 'back', 'double']},
+    transparent: {default: true},
+    alphaTest: {default: 0.5},
+    wrappixels: {type: 'number'}, // if specified, units are bmfont pixels (e.g. DejaVu default is size 32)
+    wrapcount: {type: 'number', default: 40} // units are 0.6035 * font size e.g. about one default font character (monospace DejaVu size 32)
+  },
+
+  init: function () {
+    this.texture = new THREE.Texture();
+    this.texture.anisotropy = 16; // ??
```
dmarcos: What are the question marks?

dmarcos: Can we add a better comment?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+  },
+
+  remove: function () {
+    this.geometry.dispose();
+    this.geometry = null;
+    this.el.removeObject3D('bmfont-text');
+  },
+
+  coerceData: function (data) {
+    // We have to coerce some data to numbers/booleans
```
dmarcos: Can we explain why?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    data = assign({}, data);
+    if (typeof data.lineHeight !== 'undefined') {
```
dmarcos: I prefer to do `data.lineHeigt === undefined`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+      data.lineHeight = parseFloat(data.lineHeight);
+      if (!isFinite(data.lineHeight)) { data.lineHeight = undefined; }
+    }
+    if (typeof data.width !== 'undefined') {
```
dmarcos: I prefer to do `data.width === undefined` it's more compact and consistent with the rest of the code base.

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+        shader = createBasic(data);
+      }
+      this.material = new THREE.RawShaderMaterial(shader);
+    } else {
+      this.material.uniforms.opacity.value = this.data.opacity;
+      this.material.uniforms.color.value.set(this.data.color);
+      this.material.uniforms.map.value = this.texture;
+    }
+
+    if (this.mesh) {
```
dmarcos: one liner

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+      console.error(new TypeError('No font specified for bmfont text!'));
+      return;
+    }
+
+    var geometry = this.geometry;
+    var self = this;
+    this.mesh.visible = false;
+    loadBMFont(this.lookupFont(this.data.fnt), onLoadFont);
+
+    function onLoadFont (err, font) {
```
dmarcos: user `error` instead of `err`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+      if (err) {
+        console.error(new Error('Error loading font ' + self.data.fnt +
```
dmarcos: just do 
````javascript
throw new Error('Error....');
````

File: src/components/bmfont-text.js [hunk 8]
```diff
@@ -0,0 +1,261 @@
...
+          '\nMake sure the path is correct and that it points' +
+          ' to a valid BMFont file (xml, json, fnt).\n' + err.message));
+        return;
+      }
+
+      if (font.pages.length !== 1) {
+        console.error(new Error('Currently only single-page bitmap fonts are supported.'));
```
dmarcos: Just do 
````javascript
throw new Error('...');
````

File: src/components/bmfont-text.js [hunk 9]
```diff
@@ -0,0 +1,261 @@
...
+      }
+      var data = self.coerceData(self.data);
+
+      var src = self.data.fntImage || self.lookupFont(self.data.fnt).replace('.fnt', '.png') || path.dirname(data.fnt) + '/' + font.pages[0];
+      var textRenderWidth = data.wrappixels || (data.wrapcount * 0.6035 * font.info.size);
+      var options = assign({}, data, { font: font, width: textRenderWidth, lineHeight: data.lineHeight || font.common.lineHeight });
+      geometry.update(options);
+      self.mesh.geometry = geometry;
+
+      var obj3d = self.el.object3D;
```
dmarcos: variable at the top and rename to `object3D`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+  });
+}
+
+function threeSideFromString (str) {
+  switch (str) {
+    case 'double': return THREE.DoubleSide;
+    case 'front': return THREE.FrontSide;
+    case 'back': return THREE.BackSide;
+    default:
+      throw new TypeError('unknown side string ' + str);
```
dmarcos: Capitalize first letter of the error message

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    height: {type: 'number'}, // use AFRAME units i.e. meters, not arbitrary numbers... // no default, will be populated at layout
+    align: {type: 'string', default: 'left', oneOf: alignments},
+    letterSpacing: {type: 'number', default: 0},
+    lineHeight: {type: 'number'},  // default to font's lineHeight value
+    fnt: {type: 'string', default: 'default'},
+    fntImage: {type: 'string'}, // default to fnt but with .fnt replaced by .png
+    mode: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
+    color: {type: 'color', default: '#000'},
+    opacity: {type: 'number', default: '1.0'},
+    type: {default: 'SDF', oneOf: ['SDF', 'basic', 'MSDF']},
```
dmarcos: I would rename this to `shader`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    tabSize: {default: 4},
+    anchor: {default: 'center', oneOf: anchors}, // center default to match primitives like plane; if 'align', null or undefined, same as align
+    baseline: {default: 'center', oneOf: baselines},
+    text: {type: 'string'},
+    width: {type: 'number'}, // use AFRAME units i.e. meters, not arbitrary numbers... // default to geometry width, or if not present then DEFAULT_WIDTH
+    height: {type: 'number'}, // use AFRAME units i.e. meters, not arbitrary numbers... // no default, will be populated at layout
+    align: {type: 'string', default: 'left', oneOf: alignments},
+    letterSpacing: {type: 'number', default: 0},
+    lineHeight: {type: 'number'},  // default to font's lineHeight value
+    fnt: {type: 'string', default: 'default'},
```
dmarcos: Can we call it `font`?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    fntImage: {type: 'string'}, // default to fnt but with .fnt replaced by .png
```
dmarcos: Cane we call it `fontImage`?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,261 @@
...
+    mode: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
```
dmarcos: `mode` is a bit ambiguous. Can we come up with a better name?

fernandojsg:
Maybe we could move the discussion to my repo PR and once we got that fixed and working update it? Or just the other way but to avoid commenting in both places

fernandojsg:
In fact the component wasnt yet ready to ship it on aframe.
First we need to fix the functionality for example msdf doesnt work yet but its exposed anyway. Then we should refactor and lint the code.
And finally add some cool examples.

src/components/bmfont-text.js
machenmusik: if we are not worried about compatibility with the prior component, sure

src/components/bmfont-text.js
machenmusik: if we are not worried about compatibility with the prior component, sure

src/components/bmfont-text.js
machenmusik: if we are not worried about compatibility with the prior component, sure
open to suggestions, i don't have a better one at the moment

src/components/bmfont-text.js
machenmusik: if we are not worried about compatibility with the prior component, sure
but here I think the intent was to note the type of font (basic, signed distance field, multichannel signed distance field) so actually `type` is more appropriate than directly referencing `shader` IMO

src/components/bmfont-text.js
machenmusik: I don't know, those were in the original.  i can remove them, but i don't have a better comment -- @fernandojsg any ideas?

src/components/bmfont-text.js
machenmusik: I don't know, those were in the original, hadn't looked

src/components/bmfont-text.js
machenmusik: hmm, there is already a field `font`...

src/components/bmfont-text.js
machenmusik: shouldn't rename until we figure out `fnt` vs. `font`

machenmusik:
IMO, it's better to work issues here where we don't necessarily have to concern ourselves with compatibility with prior uses of the actual component.

Sidebar, it is a nuisance that you can't use the same source between standalone and proposed core... which would make this sort of thing much easier if working components could literally be pulled into core

src/components/bmfont-text.js
donmccurdy: IMO, compatibility with prior component is of no importance. Consistency with CSS conventions is better. I'd call this property `whiteSpace`, from [css spec](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space).

src/extras/primitives/primitives/a-text.js
```diff
@@ -0,0 +1,31 @@
+/* Experimental text primitive.
+ * Issues: color not changing, removeAttribute() not working, mixing primitive with regular entities fails
+ * Color issue relates to: https://github.com/donmccurdy/aframe-extras/blob/master/src/primitives/a-ocean.js#L44
```
donmccurdy: Do we know what is meant by "mixing primitive with regular entities"?

donmccurdy: And, the issue mentioned in `<a-ocean />` should be fixed as of A-Frame v0.4.0.

machenmusik: not sure, but i would assume that means something about problems using a-text primitive with other components in same entity.

src/components/bmfont-text.js
bryik: If it helps, I set anisotropy to 16 in my component because I think it improves the look of large amounts of text particularly when viewed from an angle. The question marks probably relate to a comment I had about setting anisotropy with renderer.getMaxAnisotropy(); I didn't know how to do this in A-Frame so I just left it at 16 (the maximum level for my GPU)

src/extras/primitives/primitives/a-text.js
bryik: The problems with the text primitive do seem to have been fixed with A-Frame v0.4.0.

@donmccurdy with 0.3.0 adding a text primitive to a scene would sometimes break text added the regular way (e.g. `<a-entity bmfont-text="text: Hello World;"></a-entity>` would no longer show up)

src/components/bmfont-text.js
dmarcos: @fernandojsg Why do we need to coerce?

src/components/bmfont-text.js
dmarcos: Can we create a constant at the top and explain why 16?

src/components/bmfont-text.js
dmarcos: We don't care about compatibility with prior components. `type` is not specific enough besides it's also used in the schemas with a different meaning.

src/components/bmfont-text.js
dmarcos: I don't see the field `data.font` used anywhere.

dmarcos:
> Sidebar, it is a nuisance that you can't use the same source between standalone and proposed core... which would make this sort of thing much easier if working components could literally be pulled into core
@machenmusik What does it mean working components? @fernandojsg won't maintain his component anymore and will be now part of the core. The [registry](https://aframe.io/aframe-registry/) paired with the inspector is the way to easily consume / discover components that are not part of the core while separating concerns and liabilities. I want to reduce what gets into core to what we consider is minimal functionality and has to be consistent across a-frame scenes. If functionality keeps growing we won't be able to keep good quality.

machenmusik:
> @machenmusik What does it mean working components?

when you're developing a component, you typically start with it being standalone, and only later does it potentially merit proposal for inclusion into core.  when such proposals are done, it is a shame that there is a bunch of manual editing required to go from standalone component to core PR... e.g. rename component index.js to component.js, relative require statements as opposed to global AFRAME, to THREE or not to THREE :-), component boilerplate doesn't understand docs or tests, etc.

src/components/bmfont-text.js
machenmusik: great suggestion, thanks!

src/components/bmfont-text.js
machenmusik: ok, let's do it :-)

src/extras/primitives/primitives/a-text.js
machenmusik: so are any of these issues still valid, or should the comment be removed?

src/components/bmfont-text.js
machenmusik: looking at this slightly more, I think the data needs to be coerced because it goes right into the text creation in the options merge.

machenmusik: (added a comment but that didn't make this one outdated for some reason)

src/extras/primitives/primitives/a-text.js
bryik: I wrote those comments for my variant of bmfont-text with A-Frame v0.3.0, so I can't say for sure if they're valid for Fernando's bmfont-text with A-Frame v0.4.0. My guess would be no.

dmarcos: Can we check if these comments are still relevant?

machenmusik: (removed)

machenmusik:
squashed the PR edits here.
what else is needed?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,262 @@
...
+var anchors = ['left', 'right', 'center', 'align'];
+var baselines = ['top', 'center', 'bottom'];
+
+var DEFAULT_WIDTH = 1; // 1 matches other AFRAME default widths... 5 matches prior bmfont examples etc.
+
+// @bryik set anisotropy to 16 because I think it improves the look
+// of large amounts of text particularly when viewed from an angle.
+var MAX_ANISOTROPY = 16;
+
+var FONT_BASE_URL = 'https://cdn.rawgit.com/fernandojsg/aframe-bmfont-component/5c88edf40bbc88b336d7db6f040c3ae0d2f65aff/fonts/';
```
dmarcos: Should we move the fonts to the assets repo? https://github.com/aframevr/assets

machenmusik: probably, yes

machenmusik:
i made a crude attempt to cache fnt/png requests in a separate branch, is it better to get baseline in then do later, or try to do all at once?

dmarcos:
@ngokevin @fernandojsg Should we pull the trigger and merge this?

ngokevin:
Can the component name be shortened? `text` might be too prescriptive, what about `bmtext` or `sdftext`?

fernandojsg:
@dmarcos let me have a look at it, I didn't check the last updates for a while, sorry :)
@ngokevin umm as we could have several techniques I wouldn't choose `sdftext`, being a part of the core maybe `text` could be enough? or `text2d` or something non related to the technique used to render it

ngokevin:
I wouldn't mind `text` too much, we'd document that there are other methods. And it seems cleaner.

I haven't had time to take a look either. Will get around, but if I don't, I can always post-review.

machenmusik:
could use some more eyes on the new custom shader (avoid horrible aliasing when text gets too small)

examples/primitives/text/index.html
```diff
@@ -0,0 +1,100 @@
+<html>
```
ngokevin: This could be a `examples/test/text/` example.

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+
+var DEFAULT_WIDTH = 1; // 1 matches other AFRAME default widths... 5 matches prior bmfont examples etc.
+
+// @bryik set anisotropy to 16 because I think it improves the look
+// of large amounts of text particularly when viewed from an angle.
+var MAX_ANISOTROPY = 16;
+
+var FONT_BASE_URL = 'https://cdn.aframe.io/fonts/';
+var fontMap = {
+  'default': FONT_BASE_URL + 'DejaVu-sdf.fnt',
```
ngokevin: Could make these font names all lower case with no dashes to make it less error prone. Then could alphabetize them here

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+  'KelsonSans': FONT_BASE_URL + 'KelsonSans.fnt',
+  'Monoid': FONT_BASE_URL + 'Monoid.fnt',
+  'SourceCodePro': FONT_BASE_URL + 'SourceCodePro.fnt',
+  'mozillavr': FONT_BASE_URL + 'mozillavr.fnt'
+};
+
+var loadedFontPromises = {};
+var loadedTexturePromises = {};
+
+coreShader.registerShader('modified-SDF', {
```
ngokevin: nit: could be `modified-sdf` to not be case sensitive, could add whitelines between the properties of the shader

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+  schema: {
+    map: { type: 'map', is: 'uniform' },
```
ngokevin: no spaces inside object braces

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    '  float alpha = aastep(texColor.a);',
+    '  gl_FragColor = vec4(color, opacity * alpha);',
+    '}'
+  ].join('\n')
+});
+
+module.exports.Component = registerComponent('bmfont-text', {
+  schema: {
+    // scale is now determined by width and wrappixels/wrapcount... scale: {default: 0.003},
+    tabSize: {default: 4},
```
ngokevin: can we alphabetize this prop list

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    anchor: {default: 'center', oneOf: anchors}, // center default to match primitives like plane; if 'align', null or undefined, same as align
```
ngokevin: Can we have all prop comments on top of the props with cap + period. there's also some double inline comments below.

Alternatively, you can add them to a docstring on top of the component (or just in the documentation)

```
**
* Text component.
*
* @param {number} width - Use A-Frame units (meters). Defaults to geometry width or DEFAULT_WIDTH.
*/
```

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    this.updateMaterial(oldData && {shader: oldData.shader, customShader: oldData.customShader});
+  },
+
+  remove: function () {
+    this.geometry.dispose();
+    this.geometry = null;
+    this.el.removeObject3D('bmfont-text');
+  },
+
+  coerceData: function (data) {
```
ngokevin: This method could be a generic function outside the component

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+      var options = assign({}, data, { font: font, width: textRenderWidth, lineHeight: data.lineHeight || font.common.lineHeight });
+      this.geometry.update(options);
+      this.updateLayout(data);
+    }
+    // if shader changed, update
+    this.updateMaterial(oldData && {shader: oldData.shader, customShader: oldData.customShader});
+  },
+
+  remove: function () {
+    this.geometry.dispose();
```
ngokevin: Need to dispose any material or texture?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+        shaderObject.el = this.el;
+        shaderObject.init(data);
+        shaderObject.update(data);
+        shader = shaderObject.material;
+        shader.transparent = data.transparent; // apparently this is not set on either init or update
+        this.material = shader;
+        if (this.mesh) { this.mesh.material = this.material; }
+        return;
+      }
+      if (this.data.shader === 'SDF') {
```
ngokevin: shader types could be case insensitive

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    } else {
+      this.material.uniforms.opacity.value = this.data.opacity;
+      this.material.uniforms.color.value.set(this.data.color);
+      this.material.uniforms.map.value = this.texture;
+    }
+
+    if (this.mesh) { this.mesh.material = this.material; }
+  },
+
+  registerFont: function (key, url) { fontMap[key] = url; },
```
ngokevin: registerFont/lookUpFont/loadBMFontPromise/loadTexturePromise methods can also be generic functions

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    return new Promise(function (resolve, reject) {
+      loadBMFont(src, function (err, font) {
+        if (err) { reject(err); } else { resolve(font); }
+      });
+    });
+  },
+
+  loadTexturePromise: function (src) {
+    return new Promise(function (resolve, reject) {
+      var loader = new THREE.ImageLoader();
```
ngokevin: Could just inline ```new THREE.ImageLoader().load`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+        resolve(image);
+      }, undefined, function () {
+        reject(null);
+      });
+    });
+  },
+
+  updateFont: function () {
+    if (!this.data.font) {
+      console.error(new TypeError('No font specified for bmfont text!'));
```
ngokevin: Do you want to throw this error?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+      return;
+    }
+
+    var geometry = this.geometry;
```
ngokevin: Giant blocks of code can use some comments

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+      self.updateLayout(data);
+    }).catch(function (error) {
+      throw new Error('Error loading font ' + self.data.font +
+          '\nMake sure the path is correct and that it points' +
+          ' to a valid BMFont file (xml, json, fnt).\n' + error.message);
+    });
+  },
+
+  updateLayout: function (data) {
+    var el = this.el;
```
ngokevin: Lots of variables initialized up-front, possible to declare them first and initialize them closer to where they're used?

Alternative initialize + declare them where they're used and use `const`

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    this.mesh.position.y = y * textScale;
+    this.mesh.position.z = 0.001; // put text slightly in front in case there is a plane or other geometry
+    this.mesh.scale.set(textScale, -textScale, textScale);
+    this.geometry.computeBoundingSphere();
+  }
+});
+
+function threeSideFromString (str) {
+  switch (str) {
+    case 'double': return THREE.DoubleSide;
```
ngokevin: can you brace the cases?

`case 'double': { return THREE.DoubleSide; }`

src/extras/primitives/primitives/a-text.js
```diff
@@ -0,0 +1,29 @@
...
+var getMeshMixin = require('../getMeshMixin');
+var registerPrimitive = require('../primitives').registerPrimitive;
+var utils = require('../../../utils/');
+
+registerPrimitive('a-text', utils.extendDeep({}, getMeshMixin(), {
+  defaultComponents: {
+    'bmfont-text': {anchor: 'align', width: 5}
+  },
+  mappings: {
+    text: 'bmfont-text.text',
```
ngokevin: can alphabetize

src/extras/primitives/primitives/a-text.js
```diff
@@ -0,0 +1,29 @@
+/* Experimental text primitive.
+ */
+
```
ngokevin: empty line

src/extras/primitives/primitives/a-text.js
```diff
@@ -0,0 +1,29 @@
...
+var getMeshMixin = require('../getMeshMixin');
```
ngokevin: is the mesh mixin needed?

src/components/bmfont-text.js
dmarcos: can we name if just SDF? What is the modification?

src/components/bmfont-text.js
```diff
@@ -0,0 +1,333 @@
...
+    '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
+    '  vUV = uv;',
+    '}'
+  ].join('\n'),
+  fragmentShader: [
+    'uniform sampler2D map;',
+    'uniform vec3 color;',
+    'uniform float opacity;',
+    'varying vec2 vUV;',
+    '#define ALL_SMOOTH 0.5',
```
dmarcos: can we put defines at the top?

src/components/bmfont-text.js
dmarcos: should not be a case for `modified-sdf` here?

src/components/bmfont-text.js
dmarcos: there should be always a font by default?

dmarcos: maybe just a warning `Font not specified. Falling back to default font XXX`

src/components/bmfont-text.js
machenmusik: modified-sdf is customShader.  I couldn't figure out how to do the validation oneOf with a dynamic shader list and pre-existing values at the same time, or else I would have done as shader

machenmusik:
would `text` be too prone to namespace collisions?

src/components/bmfont-text.js
machenmusik: (addressed, but review didn't pick up)

src/components/bmfont-text.js
machenmusik: added some things in, but not sure they are entirely correct

src/components/bmfont-text.js
machenmusik: I think we don't want to use const because we aren't assuming ES6, will see what I can do

ngokevin: Yeah maybe not for this case for core http://caniuse.com/#search=const

I usually use it in components that are VR-specific though since we have the benefit of targeting experimental browsers

machenmusik: moved stuff around, should be better...

src/extras/primitives/primitives/a-text.js
machenmusik: good question.  was in inherited code, and I believe it's in the A-Frame registerPrimitive sample...

machenmusik:
Could definitely use help on tests, another case of inheriting something with none... and often visual verification is important...

src/extras/primitives/primitives/a-text.js
ngokevin: Yeah, the box example uses the material component so it benefits from the mixin, but we don't need it here.

ngokevin:
Namespace collisions not too much of an issue, Registry components aren't allowed to register names used by core components. I'll just rename mine to text-geometry (or make it part of the geometry component).

I can help with tests in a bit. Thanks!

machenmusik:
if `text` won't cause collision issues great.  it does have a property called text though, so that might be a little strange.  did you want me to change the component name, or were you going to take care of that?

ngokevin:
I think `text.text` is fine. You can change, or I can get to it later.

machenmusik:
gonna be a few houra at minimum on my end, if you have the cycles go ahead

machenmusik:
if ylu folka have better suggestions than using gl_FragCoord.w as a guess of when to drop the alpha test and blend (which makes the distant appearance even worse), let me know.  there are still some cases like extreme off-axis viewing that the custom shader doesn't do much for...

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+  'kelsonsans': FONT_BASE_URL + 'KelsonSans.fnt',
+  'monoid': FONT_BASE_URL + 'Monoid.fnt',
+  'mozillavr': FONT_BASE_URL + 'mozillavr.fnt',
+  'sourcecodepro': FONT_BASE_URL + 'SourceCodePro.fnt'
+};
+
+var loadedFontPromises = {};
+var loadedTexturePromises = {};
+
+coreShader.registerShader('modified-sdf', {
```
fernandojsg: It could be better to take it out to another file like the rest of the shaders and just require it here.

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    font: {type: 'string', default: 'default'},
+    // default to fnt but with .fnt replaced by .png
+    fontImage: {type: 'string'},
+    // no default, will be populated at layout
+    height: {type: 'number'},
+    letterSpacing: {type: 'number', default: 0},
+    // default to font's lineHeight value
+    lineHeight: {type: 'number'},
+    opacity: {type: 'number', default: '1.0'},
+    shader: {default: 'custom', oneOf: ['custom', 'sdf', 'basic', 'msdf']},
```
fernandojsg: Why not including the shader to the list instead of including the new `custom` value and the `customShader` parameter? I believe user won't know the difference and probably it's not going to be that common that users creates custom shaders, and in that case they could always add them to this list.

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    side: {default: 'front', oneOf: ['front', 'back', 'double']},
+    tabSize: {default: 4},
+    text: {type: 'string'},
+    transparent: {default: true},
+    whiteSpace: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
+    // default to geometry width, or if not present then DEFAULT_WIDTH
+    width: {type: 'number'},
+    // units are 0.6035 * font size e.g. about one default font character (monospace DejaVu size 32)
```
fernandojsg: That `0.6035` is an arbitrary value based on the DejaVu font? If so, we should modify that value for each font loaded isn't it?

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    lineHeight: {type: 'number'},
+    opacity: {type: 'number', default: '1.0'},
+    shader: {default: 'custom', oneOf: ['custom', 'sdf', 'basic', 'msdf']},
+    side: {default: 'front', oneOf: ['front', 'back', 'double']},
+    tabSize: {default: 4},
+    text: {type: 'string'},
+    transparent: {default: true},
+    whiteSpace: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
+    // default to geometry width, or if not present then DEFAULT_WIDTH
+    width: {type: 'number'},
```
fernandojsg: I'm not sure about having a value that could specified by the user, could have a default value (DEFAULT_WIDTH) or could be based on the geometry's width. Maybe leaving this one as:
```
width: {type: 'number', default: DEFAULT_WIDTH},
```
And create another boolean parameter `use_geometry_width` or so?

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    var data = coerceData(this.data);
+
+    // decide whether to update font, or just text data
+    if (!oldData || oldData.font !== data.font) {
+      // new font, will also subsequently change data & layout
+      this.updateFont();
+    } else if (this.currentFont) {
+      // new data like change of text string
+      var font = this.currentFont;
+      var textRenderWidth = data.wrappixels || (data.wrapcount * 0.6035 * font.info.size);
```
fernandojsg: The same as before, this `0.6035` is arbitrary right?

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    // Determine wrap pixel count, either as specified or by experimentally determined fudge factor.
+    // (Note that experimentally determined factor will never be correct for variable width fonts.)
+    textRenderWidth = data.wrappixels || (data.wrapcount * 0.6035 * font.info.size);
+    textScale = width / textRenderWidth;
+    // Determine height to use.
+    height = textScale * geometry.layout.height;
+
+    // update geometry dimensions to match layout, if not specified
+    if (elGeo) {
+      if (!elGeo.width) { el.setAttribute('geometry', 'width', width); }
```
fernandojsg: If you have a circle geometry, it won't have a `width` value but this code will set that value and it won't affect that geometry. 
And also this is assuming that the user always wants to modify the geometry (box, plane) to fit the text, but it can't be always true.

src/components/text.js
dmarcos: yep

dmarcos: we really need to call it modified-sdf?

src/components/text.js
dmarcos: We need docs badly for this component. I would also try to reduce the moving parts as much as possible

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    shader: {default: 'custom', oneOf: ['custom', 'sdf', 'basic', 'msdf']},
+    side: {default: 'front', oneOf: ['front', 'back', 'double']},
+    tabSize: {default: 4},
+    text: {type: 'string'},
+    transparent: {default: true},
+    whiteSpace: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
+    // default to geometry width, or if not present then DEFAULT_WIDTH
+    width: {type: 'number'},
+    // units are 0.6035 * font size e.g. about one default font character (monospace DejaVu size 32)
+    wrapcount: {type: 'number', default: 40},
```
dmarcos: camel case `wrapCount`

src/components/text.js
machenmusik: I do think there are use cases for custom shaders (e.g. this hybrid modification, or outlines, or interesting text effects), in which case they should be from the registered shader list, which is why I originally chose an explicitly verifiable "custom" setting for `shader` (essentially a promise you know what you are doing) and then `customShader` uses the existing registered shader name list

src/components/text.js
```diff
@@ -0,0 +1,372 @@
...
+    // if specified, units are bmfont pixels (e.g. DejaVu default is size 32)
+    wrappixels: {type: 'number'}
```
dmarcos: camel case: `wrapPixels`

src/components/text.js
machenmusik: (we could add the modified-sdf shader to the standard list, but I don't see much benefit in doing that at the moment...)

src/components/text.js
dmarcos: this propery is defined twice

src/components/text.js
dmarcos: I would keep it simple and forgo the custom
option to starty width

src/components/text.js
machenmusik: it's the order of precedence that makes things tricky.  if you use a default value in schema, you don't know when that value was explicitly specified, vs. when default was used... was bitten by that in auto-enter-vr.

cases in point:

in the case where you are adding text to a plane or other entity with geometry already, the usual thing is to default to same width (and height) but you also need to be able to specify (e.g. label a huge ground plane centered on origin without using the huge width) -- for the first case not specifying anything but text seems natural, and in second case specifying override width seems natural.  thus the current implementation...

src/components/text.js
machenmusik: probably but not sure what the "right" value to use would be for each; as we are using variable width fonts, there isn't really an exactly correct value.  was hoping that something from font metrics would be usable, but so far no dice.  suggestions?

src/components/text.js
machenmusik: "experimentally determined" to work well with Dejavu default font

src/components/text.js
machenmusik: this override only happens if there isn't already a geometry width -- your point about circle is well taken -- so if you have a box or plane that shouldn't fit the text, it probably has a specified width (and height)

src/components/text.js
machenmusik: (ran into some issue with the primitive when it was camelcased before?)

src/components/text.js
machenmusik: (you would want useGeometryWidth to default to true in first case, but to false in second case, which isn't possible)

src/components/text.js
machenmusik: so then let's make modified-sdf in standard list and make default... did anyone ever figure out if basic or msdf work?  if not, can strike entirely, as all the stock fonts are sdf

src/components/text.js
machenmusik: that was literal naming scheme (it is actually modified sdf shader) feel free to suggest another name.  none of the other shader choices (sdf, basic, msdf) are exposed as registered shaders, but now that we have the recent PRs, agree this one can be added

src/components/text.js
machenmusik: sidebar: is there any point in allowing separate fontImage property really?

machenmusik:
remember the concern about namespace collisions?
```aframe-master.js:71618 The primitive a-text has a mapping collision. The attribute text has the same name as a registered component and has been renamed to text-text```
so renaming in the primitive at least...

**Comment by ngokevin** [comment]
Maybe `<a-text content="Hello">` or `<a-text value="Hello">`? Also, we can talk more about future of primitives in Slack later on.

**Comment by machenmusik** [comment]
already did `<a-text value="hi there">` so we are good there, question is whether to change for `text` as well to make it consistent

**Comment by ngokevin** [comment]
Makes sense to me. `text.value`.

**Comment by machenmusik** [comment]
`text.value` is in there.  we have candidate docs now, functionality LGTM at the moment, any headway on tests?  not sure if there were any from prior incarnations to leverage

**Comment by ngokevin** [comment]
I can help with tests right after I finish my current set of work. I'll put it up next.

File: docs/components/text.md [hunk 44]
```diff
@@ -0,0 +1,91 @@
+---
+title: text
+type: components
+layout: docs
+parent_section: components
+---
+
+The text component renders bitmap and signed distance field font text.
+
+From Fernando Serrano's aframe-bmfont-component with further modifications by Michael Chen,
```
fernandojsg: I believe we should remove our names from here, in a readme as it was originally written makes sense to credit the sources and so on but here we're writing end-user documentation and I don't think it's useful anymore that info.

src/components/text.js
```diff
@@ -0,0 +1,333 @@
...
+    opacity: {type: 'number', default: '1.0'},
+    shader: {default: 'modified-sdf', oneOf: ['modified-sdf', 'sdf', 'basic', 'msdf']},
+    side: {default: 'front', oneOf: ['front', 'back', 'double']},
+    tabSize: {default: 4},
+    transparent: {default: true},
+    value: {type: 'string'},
+    whiteSpace: {default: 'normal', oneOf: ['normal', 'pre', 'nowrap']},
+    // default to geometry width, or if not present then DEFAULT_WIDTH
+    width: {type: 'number'},
+    // units are 0.6035 * font size e.g. about one default font character (monospace DejaVu size 32)
```
fernandojsg: Following the discussion about the arbitrary `0.6035` number. I would remove it from the code as it's just an specific case hand-made computed value used to match the DejaVu size, so if you change the font it doesn't make any sense to keep using it and the values for width will be also arbitrary. 
So better to skip using it and make the code less cryptic.

machenmusik: as it performs a useful function, i think it will be better to try and find a derivation applicable to other fonts, will investigate some more

machenmusik: no longer arbitrary, added some code to do quick computation on font load...

ngokevin:
OK, helping on tests and clean up now

ngokevin:
Would be great to have a section in the doc about other possible text implementatons (geometry, DOM-to-canvas). Usages, advantages, disadvantages, links.

ngokevin:
I've filed https://github.com/mrdoob/three.js/issues/10644 to improve the three.js FileLoader so we don't have to do the whole get or create Promise dance. Else I'm gonna work on decoupling that stuff we have now out of the component to make it easier to read and test.

machenmusik:
re: other implementations - yeah that would probably be useful, although it could probably use a broader perspective if someone has tried them all

machenmusik:
re: moving promise caching to FileLoader - one danger there is loss of granularity (e.g.caching small things like fonts but not always large things like videos that are being progressively streamed) - aframe texture promise cache is separate from this font promise cache, etc.

src/extras/primitives/primitives/a-text.js
```diff
@@ -0,0 +1,28 @@
...
+var registerPrimitive = require('../primitives').registerPrimitive;
+var utils = require('../../../utils/');
+
+registerPrimitive('a-text', utils.extendDeep({}, getMeshMixin(), {
+  defaultComponents: {
+    'text': {anchor: 'align', width: 5}
+  },
+  mappings: {
+    align: 'text.align',
```
ngokevin: These could be automatically populated given the schema. I'd also add a hyphen whenever there's a camelCase (e.g., `font-image`).

machenmusik: sorry, parsing - are you asking to remove the defaults from being specified in the primitive because you want to make them rely on defaults from the mapped components themselves?

ngokevin: Sorry. I meant two things:

1. Populate the mappings given the text component schema. Check the mesh mixin how we do it for the material properties. This allows us to not have to update the primitive every time we update a property name.

2. Use hyphens instead of all lowercase. `font-image`  vs `fontimage`

ngokevin:
Alright! Spent pretty much the entire week looking at this and helping with tests, clean up, docs, and refactors. It looks great. Super thanks @machenmusik! And @fernandojsg @bryik for the long lineage of the text component. :1st_place_medal: 

If we catch anything else, we can address in a smaller follow up PRs.

machenmusik:
awesome!  thanks to everyone for sticking with it! 

maybe we should add showing "Hello World!" to the A-Frame hello world example...

PR #2201, opened 2016-12-21
fix eslint warnings

machenmusik:
Changes to fix eslint warnings that prevent `git commit` and `npm start` from functioning normally


src/utils/device.js
```diff
@@ -42,6 +42,7 @@ module.exports.checkHasPositionalTracking = checkHasPositionalTracking;
 var isMobile = (function () {
   var _isMobile = false;
   (function (a) {
+// eslint-disable-next-line no-useless-escape
```
ngokevin: Can we indent this comment over

PR #2202, opened 2016-12-22
Enhance efficiency of github issue template

mkungla:
**Description:**
As community grows it happens more often where users create issues which really belong to slack or stackoverflow.

**Changes proposed:**
- Perhaps change in github issue template proposed in this PR will increases the likelihood subjects to be posted to places where these belong and therefore ensure quality of issues posted in github.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,3 +1,10 @@
+<!--
+Please if you have a question rather than reporting the issue, please post it on Stackoverflow
+or ask it on A-Frame Slack #questions channel
+https://stackoverflow.com/questions/ask/?tags=aframe
+https://aframe.io/community/#slack
+-->
+
 **Description:**
 
 - A-Frame Version:
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,12 +1,7 @@
-<!--
-Please if you have a question rather than reporting the issue, please post it on Stackoverflow
-or ask it on A-Frame Slack #questions channel
-https://stackoverflow.com/questions/ask/?tags=aframe
-https://aframe.io/community/#slack
--->
+<!-- If you have a support question, please ask at https://stackoverflow.com/questions/ask/?tags=aframe rather than filing an issue. -->
 
 **Description:**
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
For reference, I like the [three.js issue template as well](https://github.com/mrdoob/three.js/issues/new), although perhaps it is worded a bit too strongly.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,10 @@
+<!--
+Please if you have a question rather than reporting the issue, please post it on Stackoverflow
+or ask it on A-Frame Slack #questions channel
+https://stackoverflow.com/questions/ask/?tags=aframe
+https://aframevr.slack.com/messages/questions/
```
donmccurdy: Second link only works if you're already a member of the Slack. Is it still worth including, or will that confuse new people?

dmarcos: Yeah, should we link [here](https://aframe.io/community/#slack) instead?

ngokevin:
Don't know if it gets looked at, but there's a link to the CONTRIBUTING guide 

screen shot 2016-12-22 at 2 33 11 pm

.github/ISSUE_TEMPLATE.md
mkungla: yes that makes sense, will change that

mkungla:
@ngokevin unfortunately new users don't really follow such links like CONTRIBUTING.

ngokevin:
Also haven't had any issues with questions being posted on github issues for past few monhs. It's gone pretty well

mkungla:
I was looking trough open issues today and in first page (posted within month) I already encountered at least 5 subjects which should be  taken into either slack or SO.

mkungla:
If you have better suggestion how to make that doc comment more effective then let's change it, but I see need for preliminary guidelines where user can get information either topic should be posted in github or not.

dmarcos:
@ngokevin Wha do you tink?

dmarcos:
@ngokevin I let you decide if we want to merge this as it is.

mkungla:
@ngokevin your opinion?

ngokevin:
Tweaked it a bit. Thanks!

PR #2208, opened 2016-12-23
Survive invalid tick time

### LINKED ISSUE #2207: hand/tracked controls may freeze because sometimes invalid time is passed into the tick handler
**Description:**
hand/tracked controls may freeze because sometimes invalid time is passed into the tick handler
in debugging, utils.throttleTick is showing that it sometimes receives nonsensical values for time and delta (negative time -86576060... , delta 9626...) 
- A-Frame Version: 0.4.0
- Platform/Device: Windows Chromium
e.g. https://aframe.io/aframe/examples/showcase/tracked-controls/

machenmusik:
Fixes #2207


machenmusik:
thinking more about this, I think the best way to fix this is not in throttleTick, but to actually fix the time value being sent through the system.  Will do another PR along that approach

PR #2209, opened 2016-12-25
Fix null parse test of coordinates

takahirox:
**Description:**

This PR fixes the null parse test of coordinates.
`coordinate.parse(null)`  returns `undefined` if `defaultVec` isn't given.

LMK if not test but `coordinate.parse` impl is wrong, 
(I mean, if it should return `null` for null parse)
I'll close this PR end make another PR to fix the impl.

### DIFF
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -1,7 +1,7 @@
 /* global THREE */
 var extend = require('object-assign');
 // Coordinate string regex. Handles negative, positive, and decimals.
-var regex = /\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*/;
+var regex = /^\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*$/;
 module.exports.regex = regex;
 
 /**
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -1,7 +1,7 @@
 /* global THREE */
 var extend = require('object-assign');
 // Coordinate string regex. Handles negative, positive, and decimals.
-var regex = /^\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*$/;
+var regex = /^\s*((-?\d*\.{0,1}\d+)\s+){2,3}(-?\d*\.{0,1}\d+)\s*$/;
 module.exports.regex = regex;
 
 /**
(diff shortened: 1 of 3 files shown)


mkungla:
I have a feeling that #2210 and #2211 should be here as single PR not individual PR's

takahirox:
Done

tests/utils/coordinates.test.js
```diff
@@ -8,7 +8,7 @@ suite('utils.coordinates', function () {
     });
 
     test('rejects invalid coordinate', function () {
-      assert.ok(coordinates.isCoordinate('1 1 2.5 -3'));
+      assert.ok(!coordinates.isCoordinate('1 1 2.5 -3'));
```
dmarcos: This test was correct. A string with four values "1 1 2.5 -3" is a valid coordinate

src/utils/coordinates.js
```diff
@@ -1,7 +1,7 @@
 /* global THREE */
 var extend = require('object-assign');
 // Coordinate string regex. Handles negative, positive, and decimals.
-var regex = /\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*/;
+var regex = /^\s*(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s+(-?\d*\.{0,1}\d+)\s*$/;
```
dmarcos: This regex should accept a 4th optional value to consider both vec3 and vec4 coordinates

tests/utils/coordinates.test.js
```diff
@@ -24,7 +24,7 @@ suite('utils.coordinates', function () {
     });
 
     test('parses null', function () {
-      assert.equal(coordinates.parse(null), null);
+      assert.equal(coordinates.parse(null), undefined);
```
dmarcos: I believe that returning `null` on `null` input is sensible. `parse` behaves as a pass through if the value is not parsable.

tests/utils/coordinates.test.js
takahirox: So, the test name 'rejects invalid coordinate' is wrong?

tests/utils/coordinates.test.js
takahirox: Then we need to update `parse` impl, not the test code.

https://github.com/aframevr/aframe/blob/master/src/utils/coordinates.js#L28-L30

Probably we first need to make clear the specification.

src/utils/coordinates.js
takahirox: OK, I'll update the regex to accept only three or four numbers.

tests/utils/coordinates.test.js
dmarcos: Yes, this test was incorrect. It was not testing for a rejected bogus coordinate

takahirox: I see. I think we should maintain the code and comments more.

dmarcos:
Thanks!

takahirox:
I haven't updated null parse impl in this PR.
I'm gonna file another issue for it and discuss the specification there.

PR #2210, opened 2016-12-25
Fix isCoordinate

takahirox:
**Description:**

If I'm right, `isCoordinate` test is wrong because
'rejects invalid coordinate' doesn't reject.

This PR fixes this issue.
The root issue was `regex` is missing `^` and `$` and
it has matched four or more space separated numbers.


mkungla:
After adding your commit to #2209 close this PR

Issue #2212, opened 2016-12-25
Bump to Three.js r83


takahirox:
**Description:**

Please let me file "Bump to Three.js r83" because `a-mmd` component I'm working on requires r83.

https://github.com/takahirox/a-mmd

takahirox:
It seems like we need to update sound component for r83

machenmusik:
Something seems wrong with this, I am getting complete failure running tests with npm run test:chrome --

```
Chrome 55.0.2883 (Windows 10 0.0.0) ERROR
  Uncaught TypeError: THREE.FileLoader is not a constructor
  at c:/Users/machenmusik/AppData/Local/Temp/src/core/a-assets.js:7:0 <- C:/Users/machenmusik/AppData/Local/Temp/6a19aa98df29983ee4a1b025a0ab1fa9.browserify:62279

Finished in 0.588 secs / 0 secs

SUMMARY:
√ 0 tests completed
```

PR #2214, opened 2016-12-26
Bump Three.js to r83

takahirox:
**Description:**

This PR updates Three.js to r83.
It passed `npm test` and I confirmed examples keep working.
LMK if there's any other tests I need to check.

#2212

**Changes:**
- Rename `THREE.XHRLoader` with `THREE.FileLoader` in `a-assets`
- Update sound component for `THREE.Audio` which generates `source` when it  starts to play in r83.


dmarcos:
Thanks!

PR #2215, opened 2016-12-26
Remove duplicate bottom property from css

mkungla:
**Changes proposed:**
remove extra bottom property from `.a-enter-vr`


dmarcos:
Thanks!

PR #2216, opened 2016-12-26
Handle camera aspect at component update

radenui:
**Description:**

At camera component update, aspect ration is not managed for embedded scenes: this PR fixes it.

**Changes proposed:**
- handle camera aspect at update => used to be (window.innerWidth / window.innerHeight), now is handled by a scene-like function to get aspect ratio.


dmarcos:
What do you think about recalculating the aspect ratio of the camera whenever the canvas size changes instead of just on camera update?

dmarcos:
Instead of resizing the camera [here](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L295) we could emit an event that the camera component listens to and the aspect ratio is then recalculated

radenui:
I just realized I didn't explain why I needed this here: my problem is not on resize, it's when I change zoom or fov (for instance). At this moment there are no resize event and the aspect ration is still recalculated... currently based on the window size. I wanted to dynamically be able to update camera properties only

radenui:
By the way, your proposal seems good to me as well, it's just not the same use-case

dmarcos:
Why do you need to recalculate the aspect ratio when the zoom or fov change? aspect is not even part of the component schema. The aspect updates should be taken out the `update` function and make them based on canvas resizes, hence the need to introduce a new event on `a-scene` and a new method `updateAspectRatio` in the camera component

radenui:
Well, I do not need to reset the aspect ratio at update time, but *someone* probably does or this code wouldn't be here:
```javascript
camera.aspect = data.aspect || (window.innerWidth / window.innerHeight);
```
Actually `data.aspect` suggests it should be possible to set aspect in camera properties, but it's not present into the schema.

So let me sum up the conversation here:

1 - Either the aspect should be into the camera properties
    - It should be into the schema
    - It definitely should be in the update method

    Or it shouldn't:
    - No point updating the aspect at all at update time

2 - The calculation of the ratio is currently delegated to the scene: it should move to the camera component and triggered by an event from the scene

On the first point, I have no use-case on the top of my head, but I usually like to be able to set components properties where they make sense. And someone probably had this in mind or there wouldn't be a `data.aspect` at all. Not my call...

On the second point: either you want to ba able to set the ratio on the camera component schema, and then it makes sense to handle resize here, either you don't and it's perfectly fine delegating it to the scene as it is now. What do you think ?

dmarcos:
Is there any case where someone would want an aspect ratio that it's different than the viewport? If not then `aspect` should not be part of the schema and the logic should be out of the `update` method. It's better if all the camera updates are handled by the component instead of `a-scene` modifyng it. I think it's cleaner to move the logic back to the `camera` and trigger it by a `canvasresized` event emitted by the scene

dmarcos:
@radenui We're pretty close from having something mergable. My proposed changes are the following:
1. Remove `aspect` from the camera schema since I don't see any use case when one would want an aspect ratio other than the screen.
2. Emit a new event from a-scene `canvasresized`
3. The camera component recalculates the aspect ratio automatically (it's not user configurable) based on the canvas resizes.

EnoahNetzach:
Is this PR still under consideration?

dmarcos:
There are some pending changes described here: https://github.com/aframevr/aframe/pull/2216#issuecomment-301549089

EnoahNetzach:
Yes, what I meant was: is @radenui still willing to finish those?

If not, I could fork this PR and apply the missing changes.

dmarcos:
@radenui are you still actively working on this?

dmarcos:
I'm closing this due to inactivity. We will reopen if there's interest

PR #2192, opened 2016-12-20
Compatibility changes to 0.4.0 for separate gearvr controller component

machenmusik:
Per discussion on #2183, `gearvr-controls` will not be included in core, but will be available as a separate component.  These changes are needed for `hand-controls` to be compatible with `gearvr-controls` and other separate controller components.

PR #2217, opened 2016-12-26
fix timestamps when timeDelta is implausible

machenmusik:
Make sure time doesn't go backwards, and that deltas are reasonable.

Experimental versions of Chromium used for WebVR have a known issue
where the requestAnimationFrame timestamp is off by 1000
(microseconds instead of milliseconds)

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -18,6 +18,14 @@ var isMobile = utils.device.isMobile();
 var registerElement = re.registerElement;
 var warn = utils.debug('core:a-scene:warn');
 
+// Experimental versions of Chromium used for WebVR have a known issue
+// where the requestAnimationFrame timestamp is off by 1000
+// (microseconds instead of milliseconds)
+// This threshold is a time delta between requestAnimationFrame calls
+// that is implausible for correct times in milliseconds,
+// but reasonable for incorrect values in microseconds.
+var FIX_TIMESTAMPS_DELTA_THRESHOLD = 5000;
+
 /**
  * Scene element, holds all entities.
  *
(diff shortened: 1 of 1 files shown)


machenmusik:
This is a more thorough solution to the issue behind #2208

dmarcos:
I would not prefer to introduce work around for browser bugs. Specially when those bugs are known and come from experimental builds

machenmusik:
On the one hand, I understand entirely this sentiment.  (Although personally I prefer this one as it makes all A-Frame `tick` handling immune to stupid time behavior, not the specific off-by-1000 Chromium bug.)

On the other hand, the experimental builds are the only ones with native Oculus support, are commonly used for both Rift and Vive WebVR development, and are currently susceptible to loss of controller updates...  as it is not an A-Frame bug, I am not sure we want this to break loudly in QA vs. writing a `warn` message to logs and continuing on.  If you prefer to switch `tracked-controls` support to use `throttle` rather than `throttleTick`, rather than introducing workarounds into either `throttleTick` or the core itself, I can make a PR for that, but I don't think we should just leave things broken as-is.

dmarcos:
Chromium experimental builds are not the only ones with native Oculus and Vive support. Firefox Nightly also supports both.

machenmusik:
> Firefox Nightly also supports both.

Really?  when I run Nightly, it only supports Oculus via SteamVR, so Touch controllers show up as OpenVR Gamepads and lose many of their unique characteristics...?

machenmusik:
> If you prefer to switch tracked-controls support to use throttle rather than throttleTick, rather than introducing workarounds into either throttleTick or the core itself, I can make a PR for that, but I don't think we should just leave things broken as-is.

I think we are pretty close to done on #2192, if you prefer I can add the change from `throttleTick` to `throttle` in that PR

dmarcos:
Nightly supports Oculus Rift natively but not Touch yet. We're days away from having support for touch. I insist on my preference of not accumulating workarounds in the code base for known bugs in one off experimental builds of a specific browser.

machenmusik:
OK, understood.  I will change `throttleTick` to `throttle` as part of #2192 and close this PR

PR #2219, opened 2016-12-27
Fix coordinates override bug

takahirox:
**Description:**

This PR fixes the coordinates override bug in `coordinates.parse()` that
property value 0 is overridden by defaultVec's when object is given to the method.
Animation pivots/warps examples went wrong because of this bug and this PR fixes them.


takahirox:
Done

dmarcos:
Thanks!

PR #2228, opened 2016-12-30
[docs] Correct Understanding Component-Entity-Systems URL

deprecatedcoder:
Just updating the pointer to the correct one for the Understanding Component-Entity-Systems article.


ngokevin:
thanks!

PR #2229, opened 2016-12-31
Fix deepEqual for array props.

### LINKED ISSUE #2227: v.0.4.0 setAttribute(x, { object }) does not trigger update()  when property type is Array
Using the new setAttribute non-clobbering object update to replace an array property will update the schema data but will not trigger the update method. 

- A-Frame Version: v.0.4.0
- Platform/Device: Chromium WebVR v.55
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Hello, World! - A-Frame</title>
    <meta name="description" content="Hello, World! - A-Frame">
    <script src="https://aframe.io/releases/0.4.0/aframe.js"></script>
    <script>
      AFRAME.registerComponent('update-test', {
        schema: { arr: { default: ['#000', '2', '3'] } },
        init: function() {
          this.target = document.getElementById('dataTarget');
          this.el.sceneEl.addEventListener('loaded', this.send.bind(this));
        },
        send: function() {
          this.target.setAttribute('receiver', {arr: this.data.arr});
          console.log("Should see a second 'receiver updated' log before this line and the cube changed to black color, but the schema data in the target gets changed without triggering update():", 
                      JSON.stringify(this.target.components.receiver.data.arr));
        }
      });
      AFRAME.registerComponent('receiver', {
        schema: { arr: { default: ["#CCC"] } },
        dependencies: ['material'],
        update: function() {
          if(this.data.arr[0]) {
            this.el.setAttribute('material', 'color', this.data.arr[0]);
          }
          console.log('reciever updated', JSON.stringify(this.data.arr));
        }
      });
    </script>
  </head>
  <body>
    <a-scene>
      <a-box id="dataTarget" receiver position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9"></a-box>
      <a-sphere update-test position="0 1.25 -5" radius="1.25" color="#EF2D5E"></a-sphere>
      <a-cylinder position="1 0.75 -3" radius="0.5" height="1.5" color="#FFC65D"></a-cylinder>
      <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4"></a-plane>
      <a-sky color="#ECECEC"></a-sky>
    </a-scene>
  </body>
</html>
```

In this example, I expect to see 'receiver updated' logged two times (once on initialization and again on the `setAttribute` call from `update-test.send()`) and for the cube to turn black. Instead there is no second update log, and the cube stays gray, but the log output from `update-test.send()` shows that `recevier.data` is indeed changed. 

Trying to trace this in my Adit app, it looks like somewhere in the update process the target's schema property gets set by reference to the replacement array object, so the check in `component.updateProperties()` for whether there has been any change in order to trigger update always comes up false.

donmccurdy:
Fixes #2227.

Also, the method mentions supporting nested objects but doesn't appear to. Just a we'll-add-support-when-it's-needed situation?

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -72,12 +72,20 @@ module.exports.clone = function (obj) {
 function deepEqual (a, b) {
   var keysA = Object.keys(a);
   var keysB = Object.keys(b);
+  var valA;
+  var valB;
   var i;
   if (keysA.length !== keysB.length) { return false; }
   // If there are no keys, compare the objects.
   if (keysA.length === 0) { return a === b; }
   for (i = 0; i < keysA.length; ++i) {
-    if (a[keysA[i]] !== b[keysA[i]]) { return false; }
+    valA = a[keysA[i]];
+    valB = b[keysA[i]];
+    if ((Array.isArray(valA) && Array.isArray(valB))) {
+      if (!deepEqual(valA, valB)) { return false; }
+    } else if (valA !== valB) {
+      return false;
+    }
   }
   return true;
 }
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -406,6 +406,19 @@ suite('Component', function () {
       component.updateProperties({color: 'blue'});
       assert.ok(updateStub.calledOnce);
     });
+
+    test('supports array properties', function () {
+      var updateStub = sinon.stub();
+      var TestComponent = registerComponent('dummy', {
+        schema: {list: {default: ['a']}},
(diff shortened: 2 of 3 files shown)


dmarcos:
Yeah, a test suite for `deepEqual` would be very useful

donmccurdy:
Done; note that the last test case is skipped (it otherwise fails).

Issue #2139, opened 2016-11-24
Component naming silently failing


Utopiah:
**Description:**  A component named _namedcomponent_ get registered and works whereas _NamedComponent_ or _namedComponent_ silently fails.

- A-Frame Version: master
- Platform/Device: desktop Firefox stable
- Reproducible Code Snippet or Demo URL [highly encouraged]: http://jsbin.com/jovuwut/edit?html,output

mikebolt:
Add this code:

    var mm = document.getElementById('metameta');
    console.log(mm.attributes);

The resulting NamedNodeMap only contains "namedcomponent". HTML ignores the capitalization of attributes. You can use the capitalized version and register a lower-case component, but not visa-versa.

There is no fix, but the code can and should produce a warning, and this should be documented.

ngokevin:
Warning added.

PR #2233, opened 2017-01-03
Warn on component names that contain uppercase characters

imgntn:
**Description:**

This PR checks for uppercase characters in component names.  Since HTML changes these to lowercase, it's important that users have a warning about how to correctly access their components.

https://github.com/aframevr/aframe/issues/2139

**Changes proposed:**
- Use a regex to test the name for uppercase, warn if this is the case.


dmarcos:
Thanks a lot!

PR #2236, opened 2017-01-03
Update camera.js

karansaulakh:
**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -15,7 +15,7 @@ module.exports.Component = registerComponent('camera', {
     far: {default: 10000},
     fov: {default: 80, min: 0},
     near: {default: 0.005, min: 0},
-    userHeight: {default: 0, min: 0},
+    userHeight: {default: 0.05, min: 0},
     zoom: {default: 1, min: 0}
   },
 


dmarcos:
@karansaulakh Why do you prefer a 0.05 value?

ngokevin:
I forgot why the default user height for the bare camera component is 0 and not 1.6?

dmarcos:
Because a camera can be used for anything not just for a human scale avatar. You might want to use the camera on a plane cockpit or on a car driver seat post. The camera should be scale agnostic and other higher level components should introduce the concept of scale, avatar or player

ngokevin:
The default camera and `<a-camera>` have a 1.6m default, shouldn't the camera component? The camera can be used for anything, but a standing human seems like a reasonable default versus clipping the ground.

dmarcos:
Adding magic offsets is always confusing. I would probably rename `<a-camera>` to `<a-player>` or `<a-avatar>` or something like this. I prefer to keep the camera component agnostic

ngokevin:
Ok, will close

machenmusik:
agree with @dmarcos also see #2265

Issue #2225, opened 2016-12-29
add warning if entity is appended outside of a-scene


jsantell:
**Description:**

It's not possible to add components programmatically to elements that are not yet mounted to the DOM. Discovered in ngokevin/aframe-react#52, if attempting to inject an element that contains components into the DOM, the `this.el.sceneEl` property in the component does not yet exist, causing an error in most components.

- A-Frame Version: 0.4.0
- Platform/Device: Chrome OSX
- Reproducible Code Snippet or Demo URL [highly encouraged]:

```js
var el = document.createElement('a-entity');
el.setAttribute('cursor', '');
document.body.appendChild(el);
```

`Uncaught (in promise) TypeError: Cannot read property 'object3D' of null`

ngokevin:
You should be appending to the scene rather than the document.body. In the aframe-react bug, is that component mounted under the scene?

ngokevin:
Also, yeah we should throw a warning for that

jsantell:
Yes, in the aframe-react bug it is injected within the a-scene entity. Changing the STR to `document.querySelector('a-scene').appendChild(el)` does work in examples I just tried. Maybe it is aframe-react only, but not sure (yet) what in aframe-react could cause this

ngokevin:
OK, repurposing the issue for the warning. Will try to investigate the react side later

jsantell:
Thanks!

imgntn:
seems that the object3D error is related to trying to append the ```cursor``` component -- added the warning but will also open a pull to try to handle the cursor error.

PR #2240, opened 2017-01-04
warn if trying to append to an element without a nearby scene

imgntn:
**Description:**

This PR adds a console warning if you try to append an element without a scene nearby.

Previously, this code 'failed silently' -- your entities need to be children of a scene.
```
document.body.appendChild(el);
```

Now it displays a warning 
```You are attempting to attach an entity outside of the A-Frame scene. Try appending to the "a-scene" element. ```

https://github.com/aframevr/aframe/issues/2225

**Changes proposed:**
- Check is there is a closest scene before loading.

P.S.  Please let me know if there's an earlier place we should be doing this check -- it does end up actually appending the element to the document body, which seems undesirable, but harmless i guess since it doesn't do anything.  If someone missing the warning and sees the element in the DOM, it could be confusing.  Also not sure if there's a performance impact from checking this for all elements (doesn't seem like it.)

### DIFF
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -98,7 +98,6 @@ module.exports = registerElement('a-node', {
                         'outside of the A-Frame scene. ' +
                         'Try appending to the "a-scene" element. '
                       );
-          return;
         }
 
         // Default to waiting for all nodes.
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -93,6 +93,14 @@ module.exports = registerElement('a-node', {
 
         if (this.hasLoaded) { return; }
 
+        if (self.closestScene() === null) {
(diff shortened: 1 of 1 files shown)


imgntn:
yep! -- i just noticed that object3D error is actually coming from the cursor component that is being appended in the issue, so i think that here the warning will be sufficient and i can open another pull to try to exit early on no object3d on the cursor :)

dmarcos:
Thanks!

ngokevin:
I can push another commit, but the earlier place is in attachedCallback, and we have a console.warn utility that will provide where the warning is coming from.

dmarcos:
@ngokevin feel free to revert or open a new PR

imgntn:
thanks -- didn't know about the warn utility, i'll be sure to use it next time.

PR #2241, opened 2017-01-04
Add warnings if cursor and raycaster don't have scenes.

imgntn:
**Description:**

If you try to do something like the below, you'll get knock on errors in raycaster and cursor due to there being no scene element.   I'm still getting used to the codebase so I just added these patches, but perhaps the right way to handle it would be to actually stop the element from appending if there is no nearby scene, instead of just warning.  testing for a scene in many different places seems redundant, but please advise :) 

https://github.com/aframevr/aframe/issues/2225

```
el.setAttribute('cursor','')
document.body.appendChild(el);

```

**Changes proposed:**
- warn if no scene element in raycaster and cursor
- return early so we dont throw an error.
-

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -39,6 +39,10 @@ module.exports.Component = registerComponent('cursor', {
 
   init: function () {
     var cursorEl = this.el;
+    if (cursorEl.sceneEl === null) {
+      console.warn('There is no scene element for the cursor.');
+      return;
+    }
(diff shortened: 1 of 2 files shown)


ngokevin:
Thanks, but we want to do this for all entities/nodes, not just cursor/raycaster: https://github.com/aframevr/aframe/issues/2225

The code to do this is in the ANode.attachedCallback

PR #2243, opened 2017-01-05
It displays a warning if the merge target is not an entity

dmarcos:
Bug unveiled by the inspector https://github.com/aframevr/aframe-inspector/issues/408


ngokevin:
r+

PR #2247, opened 2017-01-05
fix reusing material.src with same selector, fix material.src race condition (fixes #2245, fixes #2246)

### LINKED ISSUE #2245: race condition with reusing material.src
When material.src is reused, possible race conditions may prevent the desired src from being visible.

- A-Frame Version: 0.4.0
- Platform / Device: multiple

1. Find an image URL that takes a while to be loaded.
2. Set up a scene with an entity that loads material.src from that URL (directly via src, or indirectly via selector)
3. After material.src is set, but before the image URL finished loading, change material.src to something else (e.g. a canvas)

Expected result - the canvas will remain visible as material.src, and the image load will be ignored.
Actual result - despite material.src referencing the canvas, the image load will overwrite the canvas and remain visible.  In order to make the canvas visible again, some hijinx is required due to another issue I will file separately

### LINKED ISSUE #2246: reusing material.src with same selector does not work as expected
When material.src is reused with the same selector, usage after the first may not result in texture updates.

- A-Frame Version: 0.4.0
- Platform / Device: multiple

1. Set up a canvas with id, so it can be used as material.src via selector.
2. Set up a scene with an entity that loads material.src from that selector
3. Wait long enough to make sure the canvas is visible.
4. Change material.src to empty/null to remove it.
5. Change what's on the canvas.
6. Set material.src to the canvas via same selector from step 2.

Expected result - see changes on canvas
Actual result - changes NOT seen on canvas.  looks like a check for shader.textureSrc is remembering the original selector, despite material.src having been cleared out prior.

machenmusik:
before executing callback from material.src load, check to make sure we still should (i.e. that the textureSrc still matches); also, clear textureSrc when material.src is cleared to correctly detect subsequent changes


dmarcos:
Is it possible to cover this with a test?

machenmusik:
unless you have some way to make an image load delay for long enough to trigger the race condition during testing, not reliably that I can think of...

src/core/scene/a-scene.js
```diff
@@ -17,6 +17,14 @@ var isMobile = utils.device.isMobile();
 var registerElement = re.registerElement;
 var warn = utils.debug('core:a-scene:warn');
 
+// Experimental versions of Chromium used for WebVR have a known issue
+// where the requestAnimationFrame timestamp is off by 1000
+// (microseconds instead of milliseconds)
+// This threshold is a time delta between requestAnimationFrame calls
+// that is implausible for correct times in milliseconds,
+// but reasonable for incorrect values in microseconds.
+var FIX_TIMESTAMPS_DELTA_THRESHOLD = 5000;
```
dmarcos: This does not belong to this patch. We don't want to introduce workarounds for bugs in experimental builds

src/core/shader.js
```diff
@@ -61,18 +62,19 @@ Shader.prototype = {
...
       var varType = propertyToThreeMapping[schema[key].type];
-      var varValue = schema[key].parse(data[key] || schema[key].default);
-      variables[key] = {
+      var varKey = key;
+      // For textures, by convention data.src becomes uniform sampler2D map.
+      if (varKey === 'src' && type === 'uniform' && varType === 't') { varKey = 'map'; }
+      variables[varKey] = {
         type: varType,
-        value: self.parseValue(schema[key].type, varValue)
+        value: undefined // let updateVariables handle setting these
```
dmarcos: This is confusing. An `initVariable` method that does not initialize the variable.

src/core/scene/a-scene.js
```diff
@@ -449,6 +457,25 @@ module.exports = registerElement('a-scene', {
         var effect = this.effect;
         var timeDelta = time - this.time;
 
+        // Experimental versions of Chromium used for WebVR have a known issue
```
dmarcos: The same for this code. It does not belong to this patch

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+      assert.ok(shader.prototype.fragmentShader);
+      assert.notOk(shader.prototype.uniforms);
+      assert.notOk(shader.prototype.attributes);
+    });
+
+    test('shader instance receives methods and properties', function (done) {
+      var shader = this.shader;
+      var el = this.el;
+      el.setAttribute('material', 'shader:test-shader');
+      process.nextTick(function () {
```
dmarcos: I don't think `process.nextTick` is needed

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+      });
+    });
+
+    test('shader instance called init and update', function (done) {
+      var shader = this.shader;
+      var el = this.el;
+      var initSpy = this.sinon.spy(shader.prototype, 'init');
+      var updateSpy = this.sinon.spy(shader.prototype, 'update');
+      el.setAttribute('material', 'shader:test-shader');
+      process.nextTick(function () {
```
dmarcos: `process.nextTick` is not needed

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+      });
+      if (el.hasLoaded) { done(); }
+      el.addEventListener('loaded', function () {
+        done();
+      });
+    });
+
+    teardown(function (done) {
+      var el = this.el;
+      process.nextTick(function () {
```
dmarcos: let's make this synchronous. `process.nextTick` not needed

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+
+    test('src parameter of type map --> uniform map, not attribute', function (done) {
+      var shader = this.shader;
+      var el = this.el;
+      var initSpy = this.sinon.spy(shader.prototype, 'init');
+      var updateSpy = this.sinon.spy(shader.prototype, 'update');
+      assert.notOk(initSpy.called);
+      assert.notOk(updateSpy.called);
+      el.setAttribute('material', 'shader:test-shader');
+      process.nextTick(function () {
```
dmarcos: `process.nextTick` not needed

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+      // and don't try to assert the uniform values
+      el.addEventListener('materialtextureloaded', function () {
+        var material = el.components.material;
+        var instance = material.shader;
+        // assert.ok(instance.uniforms['map'].value);
+        assert.equal(instance.material.map.image.getAttribute('src'), VIDEO);
+        done();
+      });
+      el.setAttribute('material', 'shader:test-shader; src:' + VIDEO);
+      process.nextTick(function () {
```
dmarcos: remove `process.nextTick`

tests/core/shader.test.js
```diff
@@ -0,0 +1,269 @@
...
+      // and don't try to assert the uniform values
+      el.addEventListener('materialtextureloaded', function () {
+        var material = el.components.material;
+        var instance = material.shader;
+        // assert.ok(instance.uniforms['otherMap'].value);
+        assert.equal(instance.material['_texture_' + 'otherMap'].image.getAttribute('src'), VIDEO);
+        done();
+      });
+      el.setAttribute('material', 'shader:test-shader; otherMap:' + VIDEO);
+      process.nextTick(function () {
```
dmarcos: remove `process.nextTick`

src/core/scene/a-scene.js
machenmusik: oops, I think that crept in from the branch where I started...

machenmusik:
ok I don't know what happened, I think that somehow #2199 got roped into this one... rebased to get rid of all that noise, we can deal with that on #2199 itself

src/core/scene/a-scene.js
machenmusik: oops, I think that crept in from the branch where I started...

tests/core/shader.test.js
machenmusik: will try taking all of these in #2199

src/core/shader.js
machenmusik: not sure how to take this over to #2199... 

that is why the comment is there.  `update` will be triggered, which calls `updateVariables`, which will only take action if the value is changed... so it is important that the value from `init` not be the correct value, else no action will be taken by `updateVariables`

tests/core/shader.test.js
machenmusik: it is needed in this instance, otherwise test fails

tests/core/shader.test.js
machenmusik: it is needed in this instance, otherwise test fails

machenmusik:
changes made to #2199.

is this one good to go now?

src/utils/material.js
```diff
@@ -19,9 +19,11 @@ module.exports.updateMap = function (shader, data) {
 
   // Texture removed.
   if (!material.map) { return; }
+  shader.textureSrc = null;
```
dmarcos: This is weird. I'm thinking a cleaner way to do this

machenmusik: can you say more?  the shader is using .textureSrc as a shadow to avoid redundant changes, and it wasn't being removed / reset when the texture was removed/reset.

machenmusik:
@dmarcos have you come up with an alternative fix, or has this version grown on you?

PR #2250, opened 2017-01-06
Make sure that init/update are not called when invoking flushToDOM on a non loaded entity

dmarcos:
Fixes inspector https://github.com/aframevr/aframe-inspector/issues/409

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -21,18 +21,15 @@ var styleParser = utils.styleParser;
  * of components.
  *
  * @member {object} el - Reference to the entity element.
- * @member {string} attr - Component name exposed as an HTML attribute.
+ * @member {string} attrValue - Value of the corresponding HTML attribute.
  * @member {object} data - Component data populated by parsing the
  *         mapped attribute of the component plus applying defaults and mixins.
  */
-var Component = module.exports.Component = function (el, attr, id) {
+var Component = module.exports.Component = function (el, attrValue, id) {
   this.el = el;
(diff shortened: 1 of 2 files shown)


ngokevin:
Can you explain the remove of the `updateAttrForCache` line?

dmarcos:
It's not removed. It's now part of `updateProperties`. It was redundant before in the constructor since it is also done in `updateProperties` as well

ngokevin:
I'm confused that we're now passing `attr` (presumably the attribute name) into `updateProperties` now when before we were passing an `attrValue`, and the function seems to expect an `attrValue`.

dmarcos:
`this.attrValue` is updated by `updateAttrForCache`. Now everything happens inside `updateProperties`

ngokevin:
OK. It seems like something needs to change then, either the function signature or what we pass into updateProperties:

```js
var Component = module.exports.Component = function (el, attr, id) {
  // ...
  this.updateProperties(attr);  // THIS IS THE ATTRIBUTE NAME.
  // ...
};

Component.prototype = {
  // ...
  /**
   */
  updateProperties: function (value) {
    // ...
    // BUT THIS FUNCTION EXPECTS AN ATTRIBUTE VALUE.
  }
}
```

dmarcos:
The jsdoc for the component constructor was wrong. attr was actually the value of the attribute and not the name. I fixed it.

ngokevin:
r+

Issue #2252, opened 2017-01-07
texture offset broken


machenmusik:
- A-Frame Version: 0.4.0
- Platform / Device: all
- Reproducible Code Snippet or URL:
create an entity with material src that is a texture, and offset.
note that texture offset is not applied.

machenmusik:
fixed by merge of #2253

PR #2253, opened 2017-01-07
Add NPOT texture support, and fix texture offset/repeat behavior

machenmusik:
Add support for NPOT texture settings e.g. material="src:#imgElement; npot:true"
Fix support for texture offset / repeat (broken defaults and broken apply logic)

### DIFF
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -266,6 +266,8 @@ function setTextureProperties (texture, data) {
   var repeat = data.repeat || {x: 1, y: 1};
   var npot = data.npot || false;
 
+  // To support NPOT textures, wrap must be ClampToEdge (not Repeat),
+  // and filters must not use mipmaps (i.e. Nearest or Linear).
   if (npot) {
     texture.wrapS = THREE.ClampToEdgeWrapping;
     texture.wrapT = THREE.ClampToEdgeWrapping;
@@ -274,13 +276,13 @@ function setTextureProperties (texture, data) {
   }
 
   // Don't bother setting repeat if it is 1/1. Power-of-two is required to repeat.
-  if (!(repeat.x === 1 && repeat.y === 1)) {
+  if (repeat.x !== 1 || repeat.y !== 1) {
     texture.wrapS = THREE.RepeatWrapping;
     texture.wrapT = THREE.RepeatWrapping;
     texture.repeat.set(repeat.x, repeat.y);
   }
   // Don't bother setting offset if it is 0/0.
-  if (!(offset.x === 0 && offset.y === 0)) {
+  if (offset.x !== 0 || offset.y !== 0) {
     texture.offset.set(offset.x, offset.y);
   }
 }
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -264,17 +264,25 @@ function loadImageTexture (src, data) {
 function setTextureProperties (texture, data) {
   var offset = data.offset || {x: 0, y: 0};
   var repeat = data.repeat || {x: 1, y: 1};
+  var npot = data.npot || false;
+
+  if (npot) {
+    texture.wrapS = THREE.ClampToEdgeWrapping;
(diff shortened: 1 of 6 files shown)


machenmusik:
This fixes #2252

src/utils/material.js
```diff
@@ -13,7 +13,7 @@ module.exports.updateMap = function (shader, data) {
     if (src === shader.textureSrc) { return; }
     // Texture added or changed.
     shader.textureSrc = src;
-    el.sceneEl.systems.material.loadTexture(src, {src: src, repeat: data.repeat, offset: data.offset}, setMap);
+    el.sceneEl.systems.material.loadTexture(src, {src: src, repeat: data.repeat, offset: data.offset, npot: data.npot}, setMap);
```
dmarcos: `npot` is not defined in the schema.

src/systems/material.js
```diff
@@ -264,17 +264,25 @@ function loadImageTexture (src, data) {
...
-  texture.repeat.set(repeat.x, repeat.y);
+  if (npot) {
+    texture.wrapS = THREE.ClampToEdgeWrapping;
+    texture.wrapT = THREE.ClampToEdgeWrapping;
+    texture.magFilter = THREE.LinearFilter;
+    texture.minFilter = THREE.LinearFilter;
+  }
 
+  // Don't bother setting repeat if it is 1/1. Power-of-two is required to repeat.
+  if (!(repeat.x === 1 && repeat.y === 1)) {
```
dmarcos: These conditions are weird. We prefer early returns to avoid indentation.

src/systems/material.js
```diff
@@ -264,17 +264,25 @@ function loadImageTexture (src, data) {
...
+  var npot = data.npot || false;
 
-  // Don't bother setting repeat if it is 1/1. Power-of-two is required to repeat.
-  if (repeat.x === 1 && repeat.y === 1) { return; }
-
-  texture.wrapS = THREE.RepeatWrapping;
-  texture.wrapT = THREE.RepeatWrapping;
-  texture.repeat.set(repeat.x, repeat.y);
+  if (npot) {
+    texture.wrapS = THREE.ClampToEdgeWrapping;
```
dmarcos: Why `ClampToEdgeWrapping`?

src/utils/material.js
machenmusik: right, and neither are offset or repeat...

src/systems/material.js
machenmusik: Because that is a requirement to be able to use NPOT textures directly.  Clamp not wrap/repeat, and no mipmap filtering.

src/systems/material.js
machenmusik: early returns are the bug, it would leave once it saw default repeat, and ignore offset entirely

dmarcos: `if (repeat.x !==1 || repeat.y !== 1) { .... }`

src/systems/material.js
```diff
@@ -264,17 +264,25 @@ function loadImageTexture (src, data) {
...
+  // Don't bother setting repeat if it is 1/1. Power-of-two is required to repeat.
+  if (!(repeat.x === 1 && repeat.y === 1)) {
+    texture.wrapS = THREE.RepeatWrapping;
+    texture.wrapT = THREE.RepeatWrapping;
+    texture.repeat.set(repeat.x, repeat.y);
+  }
   // Don't bother setting offset if it is 0/0.
-  if (offset.x === 0 && offset.y === 0) { return; }
-  texture.offset.set(offset.x, offset.y);
+  if (!(offset.x === 0 && offset.y === 0)) {
```
dmarcos: `if (offset.x !== 0 || offset.y !== 0) {...}`

src/utils/material.js
machenmusik: added all of them for completeness

dmarcos:
Thanks!

Issue #2259, opened 2017-01-09
Cursor example in docs is not useable 


caseyyee:
In the example provided in the cursor docs (https://aframe.io/docs/0.4.0/components/cursor.html) does not make a useable cursor:

````html
<a-entity cursor="fuse: true; fuseTimeout: 500"
  position="0 0 -1"
  geometry="primitive: ring"
  material="color: black; shader: flat">
</a-entity>
````
![image](https://cloud.githubusercontent.com/assets/1560884/21756763/56dff48c-d5da-11e6-9de8-020e6fcf4d29.png)

The default primitive is too large for a useable cursor.

````html
<a-entity cursor="fuse: true; fuseTimeout: 500"
  position="0 0 -1"
  geometry="primitive: ring; radiusInner: 0.02; radiusOuter: 0.03"
  material="color: black; shader: flat">
</a-entity>
````

Would provide a more useable result:

![image](https://cloud.githubusercontent.com/assets/1560884/21756788/b0b58576-d5da-11e6-9286-14c6d3ed93cf.png)

caseyyee:
A possible solution would be to update the cursor component to include a default or configurable style for the reticule ie:

`cursor="fuse: true; fuseTimeout: 500; style: ring"`

Otherwise #2260 updates the docs so that the geometry would be appropriately sized.

ngokevin:
Thanks. Merged.

`<a-cursor>` provides suitable default shape, just that doc was off.

PR #2260, opened 2017-01-09
Adds useable geometry to cursor #2259

caseyyee:
A possible alternative approach would be to update the cursor component to include a default or configurable style for the reticule ie:

`cursor="fuse: true; fuseTimeout: 500; style: ring"`

Thoughts?


caseyyee:
A possible solution would be to update the cursor component to include a default or configurable style for the reticule ie:

`cursor="fuse: true; fuseTimeout: 500; style: ring"`

Otherwise https://github.com/aframevr/aframe/pull/2260 updates the docs so that the geometry would be appropriately sized.

caseyyee:
Thanks

PR #2263, opened 2017-01-09
Change setBaseUrl to setTexturePath to avoid three.js deprecation warning

darkwing:
I *believe* this accomplishes the same task but don't have enough samples to be 100% sure.


dmarcos:
Thanks!

PR #2264, opened 2017-01-10
Exit fullscreen when on ExitVR()

fernandojsg:
Right now if we call `scene.exitVR()` it will just go back to the standard state if we are on VR but not if we're just in fullscreen mode. This PR introduces a call to `exitFullScreen` on the `exitVR` function.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -216,6 +216,8 @@ module.exports = registerElement('a-scene', {
         // Don't exit VR if not in VR.
         if (!this.is('vr-mode')) { return Promise.resolve('Not in VR.'); }
 
+        exitFullscreen();
+
         if (this.checkHeadsetConnected() || this.isMobile) {
           return this.effect.exitPresent().then(exitVRSuccess, exitVRFailure);
         }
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -184,11 +184,21 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
+        exitFullscreen();
         if (!this.is('vr-mode')) { return Promise.resolve(); }
         if (checkHeadsetConnected() || this.isMobile) {
           return this.effect.exitPresent().then(exitVRSuccess, exitVRFailure);
         }
         exitVRSuccess();
+        function exitFullscreen () {
```
ngokevin: Could move this down to where the enter fullscreen function is

src/core/scene/a-scene.js
```diff
@@ -184,11 +184,21 @@ module.exports = registerElement('a-scene', {
     exitVR: {
       value: function () {
         var self = this;
+        exitFullscreen();
```
ngokevin: We should call this after the not-in-VR check below

ngokevin:
r+wc

src/core/scene/a-scene.js
dmarcos: I think it won't work. Just entering fullscreen does not set the state to `vr-mode`

ngokevin: https://github.com/fernandojsg/aframe/blob/5eb5df7ccdacde78223bc2c914f010277508b4b3/src/core/scene/a-scene.js#L156

src/core/scene/a-scene.js
fernandojsg: sure!

src/core/scene/a-scene.js
fernandojsg: Yep

PR #2269, opened 2017-01-11
fixed syntax error in tutorial, also clarified on location of js code

Benjamin28:
**Description:**
Just wanting to fix some stuff in the tutorial that I got stuck on so that others don't spend so much time.
**Changes proposed:**
- fix syntax error in js listener example by changing : to ,
- add ; to end of line in js listener example
- clarify that js listener must come after <scene> for code to work

### DIFF
--- a/docs/guides/index.md
+++ b/docs/guides/index.md
@@ -527,15 +527,16 @@ Events][animatingonevents] below.
 
 In JavaScript, we grab the box with [`querySelector`][queryselector], use
 [`addEventListener`][addeventlistener], and then [`setAttribute`][setattribute]
-to make the box grow its scale when its hovered over. Note that A-Frame adds
+to make the box grow its scale when its hovered over. The code must appear after
+the scene so that the box has loaded when the script runs. Note that A-Frame adds
 features to `setAttribute` to work with multi-property components. We can pass
 a full `{x, y, z}` object as the second argument.
 
 ```js
 <script>
-  var boxEl = document.querySelector('a-box')
+  var boxEl = document.querySelector('a-box');
   boxEl.addEventListener('mouseenter', function () {
-    boxEl.setAttribute('scale', {x: 2, y: 2: z: 2});
+    boxEl.setAttribute('scale', {x: 2, y: 2, z: 2});
   });
 </script>
 ```


docs/guides/index.md
```diff
@@ -527,15 +527,16 @@ Events][animatingonevents] below.
 
 In JavaScript, we grab the box with [`querySelector`][queryselector], use
 [`addEventListener`][addeventlistener], and then [`setAttribute`][setattribute]
-to make the box grow its scale when its hovered over. Note that A-Frame adds
+to make the box grow its scale when its hovered over. The code must appear after
```
dmarcos: @ngokevin Is this stated anywhere in the docs?

dmarcos: Is the clarification here redundant?

Benjamin28: Fairly certain. Definitely not in that part of the tutorial.

Benjamin28: only one mention of scripts in getting started tutorial before that point.

> We include A-Frame as a script tag in the <head>, pointing to the A-Frame build hosted on a CDN. This has to be included before <a-scene> because A-Frame registers custom HTML elements which must be defined before <a-scene> is attached or else <a-scene> will do nothing.

then from this example

>One way to manually handle the cursor events is to add an event listener with JavaScript just like we would with a normal DOM element. If you aren’t comfortable with JavaScript, you may skip to Animating on Events below.

> In JavaScript, we grab the box with querySelector, use addEventListener, and then setAttribute to make the box grow its scale when its hovered over. Note that A-Frame adds features to setAttribute to work with multi-property components. We can pass a full {x, y, z} object as the second argument.

dmarcos: Can we separate this in 2 PRs. One with the typo that I will quickly merge and the other with the clarification and @ngokevin can evaluate

Benjamin28: Okay, sounds good. Do you want me to separate it?

dmarcos: Yes please. Thanks!

Benjamin28: https://github.com/aframevr/aframe/pull/2270

docs/guides/index.md
```diff
@@ -527,15 +527,16 @@ Events][animatingonevents] below.
...
+the scene so that the box has loaded when the script runs. Note that A-Frame adds
 features to `setAttribute` to work with multi-property components. We can pass
 a full `{x, y, z}` object as the second argument.
 
 ```js
 <script>
-  var boxEl = document.querySelector('a-box')
+  var boxEl = document.querySelector('a-box');
```
dmarcos: We should remove this change from this PR since it has been already merged on the other one

Benjamin28: I think we have to close this PR because I deleted that fork when trying to get a fresh copy of the repo. I'm still a novice with regards to branching, forking etc. I can make another PR.

dmarcos: :) This is a good exercise. I close the PR and you can reopen

PR #2271, opened 2017-01-11
added sentence clarifying where to add javascript

Benjamin28:
**Description:**

**Changes proposed:**
-added one sentence telling user to put their javascript after the scene
-
-


ngokevin:
Don't think the intro is good to get into code details. There is a JavaScript guide we can mention this though for A-Frame, we usually recommend code to live within components which go before the scene. Thanks though!

Issue #2280, opened 2017-01-13
`this.el.setAttribute('mixin' , '');` should clear mixins. Currently tries to do queryselector on empty string


ngokevin:
**Description:**

I want to clear out the mixins component on an entity

[6:08]  
this.el.setAttribute('mixin' , '');

[6:08]  
is throwing an error "'querySelector' on 'Element': 'a-mixin#' is not a valid selector."

[6:09]
any ideas?

ngokevin:
`removeAttribute('mixin')` doesn't also work.

We should also allow `.removeAttribute('mixin', mixinName)`

BhuvanaMeenakshi:
I want to work on this

BhuvanaMeenakshi:
this.el.setAttribute('mixin', this.data.mixin);

Hope this can resolve?

dmarcos:
@BhuvanaMeenakshi Thanks for the interest. Can you open a PR with a test case fixing the issue?

ngokevin:
https://github.com/aframevr/aframe/pull/2291

PR #2282, opened 2017-01-13
patches removeAttribute(`mixin`)

processprocess:
**Description:**
Allows entity.removeAttribute(`mixin`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,31 +34,87 @@
   <a href="https://github.com/aframevr/awesome-aframe">awesome-aframe</a>
 </div>
 
+## Examples
+
+<a href="https://aframe.io/a-painter/?url=https://ucarecdn.com/962b242b-87a9-422c-b730-febdc470f203/">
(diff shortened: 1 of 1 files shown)


ngokevin:
Awesome! The source code lives in `src/`, you can find it in `src/core/a-mixin.js`.

Would you be comfortable writing a unit test for that? `tests/core/a-mixin.js`

takahirox:
Is this change good enough?
I think we should call `unregisterMixin` for all registered mixins for `removeAttribute('mixin')`.

(I was working on #2280 , too)

processprocess:
@takahirox I see what you mean with that. I'm seeing how it can be approached with unregistering all mixins, then removing the mixin attribute.

@ngokevin it looks like a-mixin.js sets up the object initially, then mixin update logic is held in a-node.js. If we want to approach update logic in a-mixin.js, I wonder what the best approach is.

ngokevin:
Needs some rebase and clean up of a bunch of comment and test code

ngokevin:
Commited @ https://github.com/aframevr/aframe/commit/22c9ddb166c004d819f0d7c45f6cbb1e21e1fcf5

PR #2289, opened 2017-01-17
since we have new Oculus Touch model with separate materials, make buttons work

machenmusik:


machenmusik:
I think the appropriate line is 
https://github.com/aframevr/aframe/blob/master/src/components/vive-controls.js#L162
no idea why you're not seeing it trigger for vive, as I don't think anything has changed there... did the model files somehow change during the CDN adjustments?

machenmusik:
well, seeing as how there are no `o ...` lines in 
https://raw.githubusercontent.com/aframevr/assets/gh-pages/controllers/vive/vr-controller-vive.obj
I'd say there's your problem...

I would like to suggest again that once CDN files have been published that are referenced by an aframe version, they should be left intact as-is until all referencing aframe versions have been decommissioned... new versions get new names/paths...

machenmusik:
(taking a quick look...)

dmarcos:
They were moving for me but with no trasition. It flipped from released to pressed positions without intermediate states.

machenmusik:
maybe browser is not setting analog value correctly anymore and only giving you 0 or 1?

machenmusik:
(@dmarcos need any changes before merge?)

src/components/oculus-touch-controls.js
```diff
@@ -223,14 +229,17 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
-    buttonMeshes.thumbstick = controllerObject3D.getObjectByName(leftHand ? 'tooche1 group3 control_surface group2 thumb_stick' : 'tooche group4 control_surface group2 thumb_stick');
-    buttonMeshes.trigger = controllerObject3D.getObjectByName(leftHand ? 'tooche1 group3 trigger' : 'tooche group4 trigger');
-    buttonMeshes['X'] = controllerObject3D.getObjectByName('tooche1 group3 control_surface group2 button2');
-    buttonMeshes['A'] = controllerObject3D.getObjectByName('tooche group4 control_surface group2 button2');
-    buttonMeshes['Y'] = controllerObject3D.getObjectByName('tooche1 group3 control_surface group2 button3');
-    buttonMeshes['B'] = controllerObject3D.getObjectByName('tooche group4 control_surface group2 button3');
-    buttonMeshes.surface = controllerObject3D.getObjectByName(leftHand ? 'tooche1 group3 face control_surface group2' : 'tooche group4 face control_surface group2');
+    buttonMeshes.grip = controllerObject3D.getObjectByName(leftHand ? 'buttonHand_oculus-touch-controller-left.004' : 'buttonHand_oculus-touch-controller-right.005');
```
dmarcos: I think we should rename the button names in the model. They are really complicated compared to the Vive ones https://github.com/aframevr/aframe/blob/master/src/components/vive-controls.js#L174.

src/components/oculus-touch-controls.js
```diff
@@ -223,14 +229,17 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    buttonMeshes.thumbstick = controllerObject3D.getObjectByName(leftHand ? 'stick_oculus-touch-controller-left.007' : 'stick_oculus-touch-controller-right.004');
+    buttonMeshes.trigger = controllerObject3D.getObjectByName(leftHand ? 'buttonTrigger_oculus-touch-controller-left.005' : 'buttonTrigger_oculus-touch-controller-right.006');
+    buttonMeshes['X'] = controllerObject3D.getObjectByName('buttonX_oculus-touch-controller-left.002');
```
dmarcos: I would do `buttonMeshes.X`

src/components/oculus-touch-controls.js
machenmusik: why bother?  no one will ever see them. 
if they were going to be modified for some other reason, I'd agree to throw name changes in...

machenmusik: remember, these files on CDN are already in use, we should not change the existing assets...
if you want to make new clones of them, go ahead

dmarcos: We're talking about moving the models to a separate repo so they're not aframe specific. People will see the names. Anyways, I see them when maintaining the code so that should be enough reason :)

machenmusik: if someone changes names as you like (and hopefully places in new CDN location), I can update this PR to match

dmarcos:
Ok. thanks!

PR #2291, opened 2017-01-17
Accept el.setAttribute('mixin', '')

takahirox:
**Description:**

This PR enables `el.setAttribute('mixin', '')` (#2280).
It clears all mixins of el.

This PR doesn't touch `removeAttribute('mixin')` because
another guy is working on here #2282

### DIFF
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -1045,6 +1045,19 @@ suite('a-entity', function () {
       assert.equal(el.getAttribute('sound__2').src, soundUrl);
       assert.equal(el.getAttribute('sound__2').autoplay, true);
     });
+
+    test('clear mixin', function () {
+      var el = this.el;
+      mixinFactory('material', {material: 'shader: flat'});
+      mixinFactory('position', {position: '1 2 3'});
+      el.setAttribute('mixin', 'material position');
+      el.setAttribute('material', 'color: red');
+      assert.shallowDeepEqual(el.getAttribute('material'), {shader: 'flat', color: 'red'});
+      assert.shallowDeepEqual(el.getAttribute('position'), {x: 1, y: 2, z: 3});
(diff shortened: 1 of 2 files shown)


ngokevin:
Thanks!

tests/core/a-entity.test.js
```diff
@@ -1045,6 +1045,19 @@ suite('a-entity', function () {
       assert.equal(el.getAttribute('sound__2').src, soundUrl);
       assert.equal(el.getAttribute('sound__2').autoplay, true);
     });
+
+    test('clear mixin', function () {
+      var el = this.el;
+      mixinFactory('material', {material: 'shader: flat'});
+      mixinFactory('position', {position: '1 2 3'});
+      el.setAttribute('mixin', 'material position');
```
dmarcos: can we check for `el.mixinEls` too? `assert.equal(el.mixinEls.length, 2)` before and `assert.equal(el.mixinEls.length, 0)` after

takahirox: That's a good idea. I will.

dmarcos: thanks!

takahirox:
BTW, I'm thinking to replace `new(old)Mixins.split(' ')` with `new(old)Mixins.trim().split(/\s+/)`.
The latter can accept two or more spaces, tabs, and new lines.
It'd be useful for users.

    (example in the console)
    -> var str = ' abc def\tghi\n jkl  mn  '
    <- undefined
    -> str
    <- " abc def	ghi
         jkl  mn  "
    -> str.split(' ')
    <- ["", "abc", "def	ghi↵", "jkl", "", "mn", "", ""]
    -> str.trim().split(/\s+/)
    <- ["abc", "def", "ghi", "jkl", "mn"]

What do you guys think?

dmarcos:
sure!

PR #2292, opened 2017-01-17
Add el.mixinEls.length check for mixin clear test

takahirox:
**Description:**

This PR adds `el.mixinEls.length` check for mixin clear test to get the test stricter.

Refer to #2291


dmarcos:
Thanks!

PR #2297, opened 2017-01-19
Alphatest addition

msj121:
By allowing an alphatest, transparencies in models loaded will now work better.

I personally believe this should be an option able to be set in the schema... Perhaps this should only be added in once schema has been updated and changed to allow for multiple variables without breaking backwards compatibility? I do think single variable schemas should be removed as they limit functionality WAY too much for the future, unless there is a way to set one of the multiple variables as the default for single variables.

As far as I can tell this pull request of tested code should not be added to the repo yet as alphatest might cause undesired consequences like hiding a material that has no texture set to it because the file wasn't found etc.... I do think its a good example of possibilities for model loading and the necessity to fine tune things, as models are beastly complicated to get right.

### DIFF
--- a/src/components/collada-model.js
+++ b/src/components/collada-model.js
@@ -21,6 +21,11 @@ module.exports.Component = registerComponent('collada-model', {
 
     this.loader.load(src, function (colladaModel) {
       self.model = colladaModel.scene;
+      self.model.traverse(function(child){
+        if(typeof child.material !== "undefined"){
+          child.material.alphaTest = 0.5;
+        }
+      });
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'collada', model: self.model});
     });


ngokevin:
Possibly simpler to just document how to modify a COLLADA model post-load using the event rather than building it into the component? We'll probably move COLLADA to a community component as we get glTF in.

msj121:
I think this is a property for the model loading process itself. It seems inefficient to load the model, then to loop through all materials after an event.

That being said, I have noticed with some collada models that they simply don't load properly without this setting being adjustable, it is not about adjusting a model to me, but loading it properly in the first place. 

For instance, sprites in models tend to have transparent backgrounds (so flowers in a model), but these show up as artifacts without using this setting. By default I think it should be turned on, but at the very least I think it should be exposed.

ngokevin:
Can you address the linting?

ngokevin:
I was considering this closely. I will close because:

1. It doesn't scale, we can't cover all the material properties that can be loaded COLLADA, or all the cases where you'd "post-process" a model after loading. 
2. The workflow is expected, the same as in three.js, you load a model, and then write some code to configure as you need. This case, you can use `model-loaded` event.
3. COLLADA is on a deprecation path.

If there's any flashing, you can `visible="false"` until you are done with everything.

I did consider banning single-prop components, but instead in the future, we'll be more diligent about using that, and use it for simple cases like pos/rot/sca/visible. Thanks!

msj121:
Fair enough. It will be the same issue with GLTF, I do understand that covering all cases would create a large library. I do think that this specific property having a default might actually reduce bugs for model loading quite a bit.

Perhaps something I will try to take up with ThreeJS themselves though they do not focus too much on model loading from the bug reports I saw.

PR #2310, opened 2017-01-24
add warning if component declared after the scene in HTML (fixes #2307)

### LINKED ISSUE #2307: Add warning when a component is statically registered in the HTML *after* <a-scene>
**Description:**

Throw a warning if a developer does this. Saying components should be registered before <a-scene> is declared. Suggest to put component script tags in the <head>

```html
<body>
  <a-scene></a-scene>
  <script src="my-component.js"></script>
</body>
```

```html
<head>
  <script src="my-component.js"></script>
</head>
<body>
  <a-scene></a-scene>
</body>
```

ngokevin:
**Description:**

Store attached scene in an array.

When registering a component, compare document position of the script to the scene.

Also check if the scene hasn't loaded because we only care about static registration. If components are fetched and injected while the scene is already loaded, then that should be OK as further uses of the component will work.

**Changes proposed:**
- Cache that regex.
- Tweak case insensitive warning.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -1,5 +1,6 @@
-/* global HTMLElement */
+/* global HTMLElement, Node */
 var schema = require('./schema');
+var scenes = require('./scene/scenes');
 var systems = require('./system');
 var utils = require('../utils/');
 
@@ -11,6 +12,10 @@ var isSingleProp = schema.isSingleProperty;
 var stringifyProperties = schema.stringifyProperties;
 var stringifyProperty = schema.stringifyProperty;
 var styleParser = utils.styleParser;
+var warn = utils.debug('core:component:warn');
+
+var aframeScript = document.currentScript;
+var upperCaseRegExp = new RegExp('[A-Z]+');
 
 /**
  * Component class definition.
@@ -261,6 +266,11 @@ Component.prototype = {
   }
 };
 
+// For testing.
+if (window.debug) {
+  var registrationOrderWarnings = module.exports.registrationOrderWarnings = {};
+}
+
 /**
  * Registers a component to A-Frame.
  *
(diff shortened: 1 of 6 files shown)


src/core/component.js
```diff
@@ -11,6 +11,10 @@ var isSingleProp = schema.isSingleProperty;
 var stringifyProperties = schema.stringifyProperties;
 var stringifyProperty = schema.stringifyProperty;
 var styleParser = utils.styleParser;
+var warn = utils.debug('core:component:warn');
+
+var aframeScript = document.currentScript;
+var upperCaseRe = new RegExp('[A-Z]+');
```
dmarcos: `upperCaseRegExp` is probably more clear

src/core/component.js
```diff
@@ -251,15 +255,28 @@ Component.prototype = {
  */
 module.exports.registerComponent = function (name, definition) {
   var NewComponent;
+  var prevSibling;  // Previous sibling of current script.
```
dmarcos: `previousSibling`? Let's avoid acronyms and contractions unless they're conventional wisdom

src/core/component.js
```diff
@@ -251,15 +255,28 @@ Component.prototype = {
...
   var proto = {};
 
-  var testForUpperCase = new RegExp('[A-Z]+');
+  // Warning if component is statically registered after the scene.
+  if (document.currentScript && document.currentScript !== aframeScript) {
```
dmarcos: Maybe for future proof. Can we do something like this
````javascript
var sceneEls = document.querySelectorAll('a-scene');
var i;
var sceneEl;
for (i = 0; i < sceneEls.length; i++) {
     sceneEl = sceneEls[i];
     if (sceneEl.hasLoaded) { continue; }
    if (currentScriptEl.compareDocumentPosition(sceneEl) !== 4) { console.warn('script should follow...'); }
}
````

Comment (dmarcos) [inline src/core/component.js]: The code above needs to be tested. Specially the bit mask condition of compareDocumentPosition

**Comment by ngokevin** [comment]
Updated with tests.

I keep track of the active scenes in an array `src/core/scene/scenes`. Then I just have to loop over that array without needing to traverse DOM. Also allows for not hard-coding `a-scene` (i.e., for `ar-scene`).

File: src/core/component.js (hunk 3 above)
Comment (ngokevin) [inline src/core/component.js]: Pretty much did `require('./scene/scenes').forEach(checkPosition)` so we don't have to do querySelector.

File: src/core/component.js [hunk 4]
```diff
@@ -261,6 +266,11 @@ Component.prototype = {
   }
 };
 
+// For testing.
+if (window.debug) {
+  var _registrationOrderWarnings = module.exports._registrationOrderWarnings = {};
```
dmarcos: can we avoid the underscore?

src/core/component.js
```diff
@@ -272,13 +282,27 @@ module.exports.registerComponent = function (name, definition) {
   var NewComponent;
   var proto = {};
 
-  var testForUpperCase = new RegExp('[A-Z]+');
+  // Warning if component is statically registered after the scene.
+  if (document.currentScript && document.currentScript !== aframeScript) {
+    scenes.forEach(function checkPosition (sceneEl) {
```
dmarcos: why do we need to store multiple scenes?

ngokevin: It's what you suggested. I can change it to store a single scene if we want.

src/core/component.js
ngokevin: I can remove, but don't want to indicate internal or test-only variables?

dmarcos: It's not a pattern and gives a false sense of safety

src/core/component.js
dmarcos: Ok

src/core/component.js
ngokevin: ok

src/core/component.js
ngokevin: Alright, I'll change to store single scene if that's what you're OKing.

dmarcos: we can keep multiple scenes. that was the meaning of the ok :)

ngokevin:
Alright, everything addressed, going to squeeze it into the tag.

PR #2311, opened 2017-01-24
[docs] add more notes to guides about using a local server

kennardconsulting:
I understand this is mentioned briefly in `getting-started.md`, but it's done as kind of an option? It implies you can either 'Grab the Boilerplate' _or_ 'Include the JS Build' (from CDN). And indeed the latter will work up to a point, but it breaks at this step.

### DIFF
--- a/docs/guides/index.md
+++ b/docs/guides/index.md
@@ -266,8 +266,8 @@ image, or a 360&deg; video. For example, a dark gray background would be:
 
 ## Applying an Image Texture
 
-> From this point on, you must be serving your HTML over a Web server (such as a local
-> [Mongoose](https://www.cesanta.com/products/binary)) for image textures to load.
+> Make sure you're [serving your HTML using a local server](../introduction/getting-started.md#using-a-local-server)
+> for textures to load properly.
 
 We can apply an image texture to the box with an image, video, or `<canvas>`
 using the `src` attribute, just like we would with a normal `<img>` element.
--- a/docs/guides/index.md
+++ b/docs/guides/index.md
@@ -266,6 +266,9 @@ image, or a 360&deg; video. For example, a dark gray background would be:
 
 ## Applying an Image Texture
 
+> From this point on, you must be serving your HTML over a Web server (such as a local
+> [Mongoose](https://www.cesanta.com/products/binary)) for image textures to load.
+
 We can apply an image texture to the box with an image, video, or `<canvas>`
 using the `src` attribute, just like we would with a normal `<img>` element.
 We also should remove the `color="red"` that we set so that the color doesn't
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -41,20 +41,7 @@ We can grab the boilerplate in one of two ways:
 
 <a class="btn btn-download" href="https://github.com/aframevr/aframe-boilerplate/archive/master.zip" download="aframe-boilerplate.zip">Download .ZIP<span></span></a>
 
-We should develop with a local server so that files are properly served. We can either:
-
-- Download the [Mongoose](https://www.cesanta.com/products/binary) application
-  and open it from the same directory as your HTML file.
-- Or use Node and npm to start the local server with `npm install && npm run start`.
-
-[angle]: https://www.npmjs.com/package/angle
-
-If you have npm, you can get started with scene template right from the command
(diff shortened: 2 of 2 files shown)


ngokevin:
I'd would clarify it more in the Getting Started section. If there's a spot that doesn't imply people should use a local server, we should address it there, at the root.

kennardconsulting:
Sure I can do that. However can we have both?

The reason I say is that, at the moment, there isn't a 'flow' that goes from 'Getting Started > Guides'. At the moment the flow (which is a little confusing) goes:

aframe.io > Get Started > Device & Platform Support > Best Practices > FAQ > The End

To get to the guides, you then have to backtrack to question 3 in the FAQ and click the 'guides' link?

ngokevin:
Perhaps move Getting Started to be the first guide?

kennardconsulting:
I think that would be better. Maybe it should go:

aframe.io > Introduction >  Device & Platform Support > Get Started > Guides

Really the 'Best Practices' and 'FAQ' are not something a newbie needs to read?

kennardconsulting:
Okay made the changes to Getting Started as requested.

Not sure how to change the 'flow' of the pages though (i.e. make 'Devices & Performance > Getting Started')? Is that outside the GitHub project?

ngokevin:
I'm not too sure about that. Right now the Introduction section serves as high-level overview and introductory reading. The Guides are more hands on. I think that works for now. It certainly wouldn't hurt if beginners read through Best Practices and FAQ before getting their hands dirty.

To move Getting Started, you move it to the guides/ folder. The metadata at the top of each file determines the order and sections. Do `npm run docs` to get a preview.

kennardconsulting:
I think there's a separate discussion here. Do you want to merge this PR and I'll start a new issue?

* At the moment, the first time the doc mentions 'asset management' or 'entities' or 'lights' is in Best Practices. They have had no introduction before that. So even if it wouldn't hurt beginners to read Best Practices and the FAQ, they won't necessarily be able to understand them.

* At the moment, the Getting Started section ultimately leads to the FAQ, the 3rd question of which is 'how can I get started'. This is a bit circular :)

docs/guides/index.md
```diff
@@ -266,6 +266,9 @@ image, or a 360&deg; video. For example, a dark gray background would be:
 
 ## Applying an Image Texture
 
+> From this point on, you must be serving your HTML over a Web server (such as a local
```
ngokevin: I'd just put something like `Make sure you're [serving your HTML using a local server](../introduction/getting-started.md#using-a-local-server) for textures to load properly.`

docs/introduction/getting-started.md
```diff
@@ -41,20 +41,7 @@ We can grab the boilerplate in one of two ways:
...
-
-[angle]: https://www.npmjs.com/package/angle
-
-If you have npm, you can get started with scene template right from the command
-line with [`angle`][angle], a command line interface for A-Frame:
-
-```sh
-npm install -g angle && angle initscene
-```
+In either case, it is important to [deploy the boilerplate from a local server](#deploy-local) so that assets load correctly.
```
ngokevin: I would use the term `serve` rather than `deploy` everywhere here

docs/introduction/getting-started.md
```diff
@@ -91,6 +80,25 @@ Then we can just require A-Frame from our app, perhaps built with Browserify or
 require('aframe');
 ```
 
+<a id="deploy-local"></a>
```
ngokevin: This is done automatically. You can use `#using-a-local-server` if you rename to `Using a Local Server`

docs/introduction/getting-started.md
```diff
@@ -91,6 +80,25 @@ Then we can just require A-Frame from our app, perhaps built with Browserify or
...
+
+## Deploy from Local Server
+
+We should develop with a local server so that files are properly served. We can either:
+
+- Download the [Mongoose](https://www.cesanta.com/products/binary) application
+  and open it from the same directory as your HTML file.
+- Or use Node and npm to start the local server with `npm install && npm run start`.
```
ngokevin: nit: Don't need `Or` since it's bullet pointed

kennardconsulting:
All done. Please review.

ngokevin:
thanks!

PR #2314, opened 2017-01-25
Attaches event listeners to the controllers controls regardless of th…

dmarcos:
**DO NOT MERGE WITHOUT TESTING ON HARDWARE**

The code can be reviewed but it needs to be tested with hardware to make sure it works. I'm far from my beloved VR headsets now. If someone can test then we can merge. 

This is a requirement from the motion capture components. When reproducing we want the controls to listen to events even if there's no real tracked controls present.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -65,9 +65,6 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.onModelLoaded = bind(this.onModelLoaded, this);
     this.onControllersUpdate = bind(this.onControllersUpdate, this);
     this.checkIfControllerPresent = bind(this.checkIfControllerPresent, this);
-    this.removeControllersUpdateListener = bind(this.removeControllersUpdateListener, this);
-    this.onGamepadConnected = bind(this.onGamepadConnected, this);
-    this.onGamepadDisconnected = bind(this.onGamepadDisconnected, this);
   },
 
   init: function () {
@@ -94,6 +91,9 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     el.addEventListener('touchstart', this.onButtonTouchStart);
     el.addEventListener('touchend', this.onButtonTouchEnd);
     el.addEventListener('model-loaded', this.onModelLoaded);
+    el.sceneEl.addEventListener('controllersupdated', this.onControllersUpdate, false);
+    window.addEventListener('gamepadconnected', this.checkIfControllerPresent, false);
+    window.addEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);
   },
 
   removeEventListeners: function () {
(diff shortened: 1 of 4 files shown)


machenmusik:
should the motion capture components be hooked up to the utils and systems portions so the event replay makes it actually appear as though the controllers are present and returning the appropriate pose data?  I thought we intentionally abstracted tracked-controls etc. from the actual gamepad API for precisely this purpose...

dmarcos:
The capture components won't be part of the core so I prefer them to use the same public API as any others. This change is innocuous for the normal use of the vive/oculus components. We can revisit if it ends up being a problem.

machenmusik:
contrary view - the utils and system functions are part of core, allowing
the gamepad api to be wrapped/replaced so that existing APIs can be used
seems better to me than creating the opportunity for dueling systems trying
to fire the same events.  and if it can be done with the core API
navigator.getGamepads() etc. then even software that would not be aware of
the emulation would obey...

machenmusik:
you may only need to modify / wrap `trackedControlsUtils.getGamepadsByPrefix()` ...

dmarcos:
Monkey patching is difficult to understand and debug in case of problems. it should be used very sparsely. If events name collisions is a problem we can revisit. For the moment I lean toward the simplest solution.

machenmusik:
if this doesn't cause problems when using multiple (e.g. both oculus touch and vive), that may work, but for more complex interactions you may need the actual gamepad api values, not just derived events, which is why I mention an alternate strategy of providing mock data and events which at the moment is fed by the tracked-controls utils functions.  (rather than monkey patching, exposing API to allow injection... also used for testing anyway, IIRC the controls classes had to provide local overrides for getGamepadsByPrefix for that purpose)

ngokevin:
Tested with hand-controls + vive-controls on Vive.

@mchen RE: your idea. I think just relying on events and event details are simpler than wrapping or stubbing the whole GamePad API. 

r+ from me

machenmusik:
```
mchen [8:07 PM] 
re: recorder and tracked controller events - the worry I have is that if the app uses both aframe controls and also gamepad api (e.g. arcade), recorder replays may not work well / correctly... and that if during recording playback the gamepad api input isn't turned off, chaos may ensue

[8:08]  
even if you ignore the former, you will probably want to address the latter

ngokevin [8:08 PM] 
you mean you’re not using tracked-controls?

mchen [8:08 PM] 
both tracked-controls (via oculus-touch-controls and vive-controls) and actual gamepad API -- to support actual gamepads, Ooculus remote, etc.

[8:10]  
even if i wasn't, it's not clear that playback would only ever occur without HMD and controllers present/active...

ngokevin [8:10 PM] 
what are some things our recorder would not pick up from your arcade?

[8:10]  
because the gamepad API doesn’t emit events right? you have to poll it for button states

mchen [8:12 PM] 
right.  so there is code that looks at gamepad states, across all of them, and does the right translation into keys for JSMESS emulator to see

[8:13]  
if you are recording tracked controller events only, would fail completely for this case
```

Issue #2306, opened 2017-01-23
Tilt does´t work in ios touching screen


goltra:
I have an ionic2 app with videosphere and in ios  touching vertically the screen I can´t get tilt effect (If I move my device it work). It doesn´t work in android neither. I try it in browser and works ok.

- A-Frame Version: 0.4.0
- Platform / Device: Ios 10 / Iphone 6S
- Reproducible Code Snippet or URL:

dmarcos:
What do you mean by `tilt` efect? You say if you move the device it works?

goltra:
I mean, you can move videosphere touching your screen, can you? ... well, if you move right and left the vide move but if you move your finger up and down the video doesn´t move (tilt effect)... I hope I have explaind

dmarcos:
Is there any error in console? does ionic expose the device motion APIs?

goltra:
In the screenshot I have attached you can see warnings and error I have. In fact, last warning appear just when I touch the screen.. 
If I move the device vertical and horizontal movement works ... 

![aframe-error](https://cloud.githubusercontent.com/assets/1634767/22260149/5a538fc6-e268-11e6-8894-ed311266a40f.jpg)

goltra:
@dmarcos you can try accessing with your mobile browser to https://aframe.io/examples/. You can check what I meaning

ngokevin:
> If you move your finger up and down the video 

`look-controls` does not support vertical panning because it'll throw off the orientation axis. If you want to design for non-VR, you'd need to create your own controls to modify the camera.

goltra:
@ngokevin I understand if you are in vr mode tilt doesn't work but if you aren't , in my opinion, it should work ... however could you make me an example about "play" with camera and vertical panning?

Thanks

goltra:
I have implemented this and created a pr here
https://github.com/aframevr/aframe/pull/2315

Please take a look at this.

PR #2315, opened 2017-01-26
allow vertical touch-movement if you are not in vr-mode

goltra:
**Description:**
Added ability to move vertically touching screen in not vr-mode

https://github.com/aframevr/aframe/issues/2306

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -275,12 +275,24 @@ module.exports.Component = registerComponent('look-controls', {
 
   onTouchMove: function (e) {
     var deltaY;
+    var deltaX;
+    var sceneEl = this.el.sceneEl;
     var yawObject = this.yawObject;
+    var pitchObject = this.pitchObject;
+
     if (!this.touchStarted) { return; }
     deltaY = 2 * Math.PI * (e.touches[0].pageX - this.touchStart.x) /
             this.el.sceneEl.canvas.clientWidth;
-    // Limits touch orientaion to to yaw (y axis)
+    
+    // Limits touch orientaion to to yaw (y axis) only if 
+    // vr mode is actived
     yawObject.rotation.y -= deltaY * 0.5;
+	if(!sceneEl.is('vr-mode')){
+    	deltaX = 2 * Math.PI * (e.touches[0].pageY - this.touchStart.y) /
+            this.el.sceneEl.canvas.clientHeight;
+		pitchObject.rotation.x -= deltaX * 0.5;
+	}
+
     this.touchStart = {
       x: e.touches[0].pageX,
       y: e.touches[0].pageY


dmarcos:
Thanks! Not allowing vertical touch is a deliberate decision because it becomes very confusing for de user in combination with device motion. You quickly lose the reference of where is up/down and you can only recover by reloading the page

morandd:
I just bumped up against this issue as well, and I see it has been raised several times in the github discussion.

It's a bit confusing for us A-Frame users because the "look-controls" component covers 3 quite different interaction styles: desktop with a mouse, mobile with multitouch, and VR/magic window mode. 

Why not allow pitch, within definable min and max angles, in touchmove in look-controls? This makes sense for my use case, and from the previous queries about this, it seems I'm not alone. Alternatively, another way to do this could be to add a toggle to look-controls so that ptich/yaw are controlled by single touch in 2d mode, and by deviceorientation in vr mode.

Overall, it would be really nice if it were easier to switch between camera control schemes between default, touchscreen device, and VR/magic window. For example on the deskop or mobile device, maybe you start with orbit controls, but then when you enter VR with a headset/cardboard/"magic window" mode, you switch to free-look controls.  Currently we users have to write our own components to juggle cameras, but I think having some defaults baked into AFrame would be quite helpful. Over in the orbit-controls I submitted a PR with to add a "vrCamera" attribute, which lets the user specify a different camera element ID that gets activated on enter-vr, then switch back to orbit on exit-vr. Until (...) this is accepted I have written my own component to do this juggle.

Ok, sorry for a bit of a soliloquy/jeremiad. AFrame is awesome! Just trying to offer some constructive contributions -- and I'm happy to make some PRs to manifest this if that's helpful.

ngokevin:
It's difficult to support every single use case. We prefer that the standard implementation remain focused on a default preset. This is better as increased complexity in code will increase maintenance in the long run and slow releases. Components can be built or customized that can be updated or published asynchronously from A-Frame. I think an okay compromise that at least it's possible to achieve what you need via components.

morandd:
Yep, understood about wanting to keep it simple. However I would say the current state of affairs having to juggle cameras is not so user friendly. 

My situation is I think a pretty common one:
I want to show off a 3d object on a web site. It should work on the desktop, on a phone/tablet, and with a VR headsset or cardboard. Some scenes I want to show using orbit controls, and some scenes I want to use a free look camera.

I thought this would work out of the box so to say.

Here is the logic to switch cameras to accomplish this:

```
Scenes using the orbit-controls paradigm
	On Desktop:
		start: use orbit controls
		enter-vr: save orbit cam position. Switch active cam to look-controls. 
		exit-vr: save look-control position. Restore orbit-cam position. Switch active to orbit cam
	On mobile:
		start: use orbit-controls
		enter-vr: save orbit cam position. Switch active cam to look-controls. 
		exit-vr: save look-control position. Restore orbit-cam position. Switch active to orbit cam

Scenes using the look-controls paradigm
	On Desktop:
		start: use look-controls
		enter-vr: Remain in look-controls mode
		exit-vr: Reset camera pitch to 0'. Remain in look-controls mode
	On mobile:
		start: use multitouch-look-controls*
		enter-vr: switch to regular look-controls
		exit-vr: switch to multitouch-look-controls

*multitouch-look-controls adds two gestures: single finger up/down drag to pitch, and pinch to zoom/dolly.
```

If I were to share my "controller controller" code for the above,  any suggestions where best I should contribute it? For new users (like me) there are already plenty of camera controller components floating about, so it would be nice to contribute this code / readme / blog entry to an established codebase.

ngokevin:
I understand there are many use cases, but we try to set the direction of the project towards true WebVR. You can add it to https://github.com/aframevr/awesome-aframe, publish it and we'll feature it on the blog, or contribute a FAQ entry on how to customize controls. Thanks!

dmarcos:
The decision to lock finger panning to the `x` axis is what we observed delivers the best experience for the majority of applications when combined with gyro orientation that all tablets and phones have. If you allow for vertical panning the user quickly loses the notion of up / down and it gets disorienting. The good thing about a-frame is that it's very easy to override the defaults if your application does not fit within those design decisions. We cannot cover every single use case on A-Frame core because as @ngokevin said we would not be able to guarantee good quality for all the possible combinations of controls moving forward. We want aframe core to be useful with sane defaults but still contain the complexity of the code base. We've been focusing the efforts on making very easy for anyone to modify and extend the default behavior.

PR #2328, opened 2017-01-30
Updated VREffect to allow scale camera

fernandojsg:
As per discussed on https://github.com/mrdoob/three.js/pull/10667
I included exactly the same version from three.js, there're some changes with spaces on functions and so on but I didn't want to change them back to semistandard' style as we don't lint this file and I believe it's better to keep exaclty the same file as three.js.
Basically the lines changes by me are from 310 to 315 to prevent decompose two times the same matrix and use the world scale for the camera to scale the IPD.

### DIFF
--- a/vendor/VREffect.js
+++ b/vendor/VREffect.js
@@ -9,7 +9,7 @@
  *
  */
 
-THREE.VREffect = function ( renderer, onError ) {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

PR #2331, opened 2017-01-30
enable non-recursive raycaster (fixes #2329)

### LINKED ISSUE #2329: feat(raycaster): Raycaster doesn't trigger objects when recursive is set to false (refs: v0.4.0)
### Environment:
- A-Frame Version: 0.4.0 (refs: 87c934aae6c85ac05af48e3c569db584197fb946)
- Platform / Device: Tested on Chrome
### Editor : 
https://ide.c9.io/maissani/testvr
### Application: 
https://testvr-maissani.c9users.io
### Files: 
https://preview.c9users.io/maissani/testvr
ps: I introduced a console.log line 59310 in aframe-v4.js to reveal intersection value behavior

### Expected Behavior and Issue
Raycaster with recursive true currently working properly. 
Raycaster with recursive false is not working properly: it doesn't trigger on Objects. 
It seems that 
```JavaScript
intersections = this.raycaster.intersectObjects(this.objects, data.recursive)
```
Doesn't  return any intersection when set to false
This bug is probably related to Three.js or not

A simple workAround tho fix this is to set to true whatever ( and then rollback when it's fixed on three.js), and add a conditional check to get only the parrent object (only if he is the origin of intersection )

machenmusik:
fixes #2329

### DIFF
--- a/examples/test/raycaster/simple.html
+++ b/examples/test/raycaster/simple.html
@@ -7,28 +7,33 @@
     <script src="../../../dist/aframe-master.js"></script>
   </head>
   <body>
+
+<script>
+
+  AFRAME.registerComponent('raycast-blab', {
+    init: function () {
+      this.el.addEventListener('raycaster-intersected', function (evt) {
+        var el = evt.detail.target;
+	// May get two intersection events per tick; same element, different faces.
+        console.log('raycaster-intersected ' + el.outerHTML);
+        el.setAttribute('material', 'color', '#7f7');
+      });
+
+      this.el.addEventListener('raycaster-intersected-cleared', function (evt) {
+        var el = evt.detail.target;
+	// May get two intersection events per tick; same element, different faces.
+        console.log('raycaster-intersected-cleared ' + el.outerHTML);
+        el.setAttribute('material', 'color', '#f77');
+      });
+    }
+  });
+
+</script>
+
     <a-scene>
       <a-entity camera look-controls wasd-controls raycaster="recursive: false; interval: 1000"></a-entity>
+      <a-entity geometry="primitive: box" material="color: #000" position="0 0 -2" raycast-blab></a-entity>
     </a-scene>
 
-    <script>
-      var el = document.createElement('a-entity');
-      el.setAttribute('geometry', 'primitive: box');
-      el.setAttribute('material', 'color: #000');
-      el.setAttribute('position', '0 0 -2');
-      document.querySelector('a-scene').appendChild(el);
-
-      document.querySelector('a-entity[geometry]').addEventListener('raycaster-intersected', function (evt) {
-	// getting two intersection events per tick; same element, different faces...
-        console.log('raycaster-intersected ' + evt.detail.el.outerHTML);
-        this.setAttribute('material', 'color: #7f7');
-      });
-
-      document.querySelector('a-entity[geometry]').addEventListener('raycaster-intersected-cleared', function (evt) {
-	// getting two intersection events per tick; same element, different faces...
-        console.log('raycaster-intersected-cleared ' + evt.detail.el.outerHTML);
-        this.setAttribute('material', 'color: #f77');
-      });
-    </script>
   </body>
 </html>
--- a/examples/test/raycaster/simple.html
+++ b/examples/test/raycaster/simple.html
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Raycaster</title>
+    <meta name="description" content="Raycaster - A-Frame">
+    <script src="../../../dist/aframe-master.js"></script>
+  </head>
+  <body>
+    <a-scene>
+      <a-entity camera look-controls wasd-controls raycaster="recursive: false; interval: 1000"></a-entity>
+    </a-scene>
+
+    <script>
+      var el = document.createElement('a-entity');
+      el.setAttribute('geometry', 'primitive: box');
+      el.setAttribute('material', 'color: #000');
+      el.setAttribute('position', '0 0 -2');
+      document.querySelector('a-scene').appendChild(el);
+
+      document.querySelector('a-entity[geometry]').addEventListener('raycaster-intersected', function (evt) {
+	// getting two intersection events per tick; same element, different faces...
+        console.log('raycaster-intersected ' + evt.detail.el.outerHTML);
+        this.setAttribute('material', 'color: #7f7');
+      });
+
+      document.querySelector('a-entity[geometry]').addEventListener('raycaster-intersected-cleared', function (evt) {
+	// getting two intersection events per tick; same element, different faces...
+        console.log('raycaster-intersected-cleared ' + evt.detail.el.outerHTML);
+        this.setAttribute('material', 'color: #f77');
+      });
+    </script>
+  </body>
+</html>
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -67,11 +67,13 @@ module.exports.Component = registerComponent('raycaster', {
    */
   refreshOnceChildLoaded: function (evt) {
     var self = this;
-    var el = evt.detail.el;
-    if (!el) { return; }
-    if (el.hasLoaded) { this.refreshObjects(); } else {
-      el.addEventListener('loaded', function nowRefresh (evt) {
-        el.removeEventListener('loaded', nowRefresh);
+    var childEl = evt.detail.el;
+    if (!childEl) { return; }
+    if (childEl.hasLoaded) {
+      this.refreshObjects();
+    } else {
+      childEl.addEventListener('loaded', function nowRefresh (evt) {
+        childEl.removeEventListener('loaded', nowRefresh);
         self.refreshObjects();
       });
     }
@@ -84,13 +86,13 @@ module.exports.Component = registerComponent('raycaster', {
     var children;
     var data = this.data;
     var i;
-    var len;
     var objects;
     var objectsAreEls = data.objects ? this.el.sceneEl.querySelectorAll(data.objects) : null;
 
     // Push meshes onto list of objects to intersect.
     if (objectsAreEls) {
-      for (objects = [], i = 0, len = objectsAreEls.length; i < len; i++) {
+      objects = [];
+      for (i = 0; i < objectsAreEls.length; i++) {
         objects.push(objectsAreEls[i].object3D);
       }
     } else {
@@ -98,7 +100,8 @@ module.exports.Component = registerComponent('raycaster', {
       objects = this.el.sceneEl.object3D.children;
     }
 
-    for (this.objects = [], i = 0, len = objects.length; i < len; i++) {
+    this.objects = [];
+    for (i = 0; i < objects.length; i++) {
       // A-Frame wraps everything (e.g. in a Group) so we want children.
       children = objects[i].children;
 
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -84,32 +84,28 @@ module.exports.Component = registerComponent('raycaster', {
     var children;
     var data = this.data;
     var i;
-    var j;
-    var object;
+    var len;
     var objects;
-    var objectsAreEls = !!data.objects;
+    var objectsAreEls = data.objects ? this.el.sceneEl.querySelectorAll(data.objects) : null;
 
     // Push meshes onto list of objects to intersect.
-    // If objects not defined, intersect with everything.
     if (objectsAreEls) {
-      objects = this.el.sceneEl.querySelectorAll(data.objects);
+      for (objects = [], i = 0, len = objectsAreEls.length; i < len; i++) {
+        objects.push(objectsAreEls[i].object3D);
+      }
     } else {
+      // If objects not defined, intersect with everything.
       objects = this.el.sceneEl.object3D.children;
     }
 
-    this.objects = [];
-    for (i = 0; i < objects.length; i++) {
-      // If we were given elements, get the object3D.
-      object = objectsAreEls ? objects[i].object3D : objects[i];
+    for (this.objects = [], i = 0, len = objects.length; i < len; i++) {
       // A-Frame wraps everything (e.g. in a Group) so we want children.
-      children = object.children;
-      if (children && children.length) {
-        // Add the object3D's children so non-recursive raycasting will work correctly.
-        for (j = 0; j < children.length; j++) { this.objects.push(children[j]); }
-      } else {
-        // If there aren't any children, then until a refresh after geometry loads,
-        // raycast won't see this object... but that should happen automatically.
-      }
+      children = objects[i].children;
+
+      // Add the object3D's children so non-recursive raycasting will work correctly.
+      // If there aren't any children, then until a refresh after geometry loads,
+      // raycast won't see this object... but that should happen automatically.
+      if (children) { this.objects.push.apply(this.objects, children); }
     }
   },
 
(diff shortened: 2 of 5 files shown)


dmarcos:
We need to fix the tests

machenmusik:
this is turning out to be more insidious than expected...  still working on it

i can see now why the implementation did what it did, in that once the change makes entities require geometry in order to intersect, a slew of the tests break because they never actually cared, and doing it correctly is not as simple as one would expect.

machenmusik:
looks like there is an issue with refreshObjects() happening before geometry is attached (and thus there are no children of the entity's object3D to add yet).  if there is some better way to understand when refreshObjects() can be invoked correctly, that would be most welcome...

machenmusik:
ok now I am a little baffled, as `npm run test:chrome` works locally with all non-skipped tests passing, but apparently something in travis-ci changes, and things fail.

ngokevin:
Probably just some race condition or timing issue.

machenmusik:
latest PR incorporates this...
```ngokevin [8:00 PM] 
i suggest in the raycaster component, we do a refresh on:

[8:00]  
1. `sceneEl.addEventListener('loaded')`

[8:00]  
2. `sceneEl.addEventListener('child-attached')` + `loaded` on that attached child```

**Comment by machenmusik** [comment]
... and travis-ci didn't like it (although `npm run test:chrome` did), so reverting.

**Comment by machenmusik** [comment]
So I think we'd rather go back to the approach in 37ffadb, but certain tests fail (although not `npm run test:chrome`) and it is rather tedious to keep throwing things at travis-ci via GitHub... is it better to take the current changes which pass tests, although less elegant?

**Comment by maissani** [comment]
Any update about this fix ?

**Comment by machenmusik** [comment]
all the tests work, the latest version of PR uses strategy suggested,
maybe it got a bit buried in the push to 0.5.0...

I believe the raycaster interval fix was merged into 0.5.0 separately,
but the remainder is sitting ready and waiting

**Comment by machenmusik** [comment]
@dmarcos @ngokevin now that 0.5.0 rush is over, are we ok to take this?

**Comment by ngokevin** [comment]
What did we think about just removing the `recursive` property altogether and being `recursive` by default? I remember something about hard to figure out what element exactly was intersected in a big tree?

**Comment by machenmusik** [comment]
it can be significantly more expensive to do recursive raycast where non-recursive suffices as well, but perhaps @maissani and others can comment

**Comment by cvan** [comment]
@machenmusik did you still want to get this merged?

**Comment by machenmusik** [comment]
if it still works, sure... @maissani @donmccurdy is this PR still useful to you?

**Comment by donmccurdy** [comment]
I've forgotten the context here a bit, but as far as I remember: With `raycaster="recursive:false`, the raycaster tests against each entity's root Object3D, and will never detect anything. That's no good; we should either fix that or remove the option.

1. If we can simplify this PR a bit, great. I'll add proposed changes, if we choose to move forward, in separate comments.
2. Removing the `recursive` option would be OK, too. For users who understand the performance profile and scene graph enough to care, setting up a custom component with a THREE.Raycaster is not that difficult.

File: src/components/raycaster.js [hunk 1]
```diff
@@ -60,25 +63,54 @@ module.exports.Component = registerComponent('raycaster', {
...
+      // A-Frame wraps everything (e.g. in a Group) so we want children.
+      children = object.children;
+      if (children && children.length) {
+        // Add the object3D's children so non-recursive raycasting will work correctly.
+        for (j = 0; j < children.length; j++) { this.objects.push(children[j]); }
+      } else {
+        // If there aren't any children, then until a refresh after geometry loads,
+        // raycast won't see this object... but that should happen automatically.
+      }
+    }
```
donmccurdy: With the change above, this block reduces to:

```js
for (i = 0; i < objects.length; i++) {
  var children = objects[i].children;
  if (children) {
    this.objects.push.apply(this.objects, children);
  }
}
```

src/components/raycaster.js
```diff
@@ -60,25 +63,54 @@ module.exports.Component = registerComponent('raycaster', {
...
+  refreshOnceChildLoaded: function (evt) {
+    var self = this;
+    var el = evt.detail.el;
+    if (!el) { return; }
+    if (el.hasLoaded) { this.refreshObjects(); } else {
+      el.addEventListener('loaded', function nowRefresh (evt) {
+        el.removeEventListener('loaded', nowRefresh);
+        self.refreshObjects();
+      });
+    }
```
machenmusik: wow this was from a long time ago... refreshing my memory as well...
I think this was supposed to be the last one, but listener should be on sceneEl not el

machenmusik: looking at it a little more closely now, shouldn't this happen when children are attached somewhere in the scene?
`this.el.sceneEl.addEventListener('child-attached', this.refreshOnceChildLoaded);`

src/components/raycaster.js
```diff
@@ -60,25 +63,54 @@ module.exports.Component = registerComponent('raycaster', {
...
-    if (data.objects) {
-      objectEls = this.el.sceneEl.querySelectorAll(data.objects);
-      this.objects = [];
-      for (i = 0; i < objectEls.length; i++) {
-        this.objects.push(objectEls[i].object3D);
-      }
-      return;
+    // If objects not defined, intersect with everything.
+    if (objectsAreEls) {
+      objects = this.el.sceneEl.querySelectorAll(data.objects);
```
machenmusik: had been assuming JIT would hoist the length call, but you're right that it can be explicit.

src/components/raycaster.js
machenmusik: (presumably this wants the explicit length hoisting described above)

machenmusik:
(also rebased against something more recent)

src/components/raycaster.js
machenmusik: I think I originally had some changes that would defer this work until next actual raycast (to coalesce / avoid spurious invocations) - but that seems to have gone, should we bring it back?

src/components/raycaster.js
donmccurdy: Actually not here, on the explicit length hoisting. With normal arrays, doing `i = 0; i < array.length; i++` is completely fine, clean, and fast.

But if you have a [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) instead of an array (the result of `querySelectorAll()` is a NodeList, not an array) then calling `.length` is slower, as [described here](http://stackoverflow.com/a/25402066/1314762).

donmccurdy: Heh, ok it's more complicated than I thought, there are "live" and "static" NodeLists, and the result of querySelectorAll() is the latter, which means it can probably be hoisted anyway. Live NodeLists are more trouble. Feel free to ignore my concern and I'll just patch it if I actually run into a noticeable performance issue. :)

machenmusik: ok, that's how I thought it worked, but I already proactively patched since code is still quite readable that way

machenmusik:
@donmccurdy @dmarcos @ngokevin anything else on this one?

src/components/raycaster.js
```diff
@@ -60,25 +63,50 @@ module.exports.Component = registerComponent('raycaster', {
   },
 
   /**
+   * Update list of objects to test for intersection once child is loaded.
+   */
+  refreshOnceChildLoaded: function (evt) {
+    var self = this;
+    var el = evt.detail.el;
```
ngokevin: I'd use `childEl` to not confuse with the raycaster.

src/components/raycaster.js
```diff
@@ -60,25 +63,50 @@ module.exports.Component = registerComponent('raycaster', {
...
+    if (!el) { return; }
+    if (el.hasLoaded) { this.refreshObjects(); } else {
```
ngokevin: Let's put `else` blocks on their own level.

```
if (clause) {
  // ...
} else {
  // ...

}
```

src/components/raycaster.js
```diff
@@ -60,25 +63,50 @@ module.exports.Component = registerComponent('raycaster', {
...
    * Update list of objects to test for intersection.
    */
   refreshObjects: function () {
+    var children;
     var data = this.data;
     var i;
-    var objectEls;
+    var len;
+    var objects;
+    var objectsAreEls = data.objects ? this.el.sceneEl.querySelectorAll(data.objects) : null;
```
ngokevin: Why the logic change here? Previously, we had:

```
if (data.objects) {
  // Run query selector.
} else {
  // Intersect with everything.
}
```

src/components/raycaster.js
```diff
@@ -60,25 +63,50 @@ module.exports.Component = registerComponent('raycaster', {
...
 
     // Push meshes onto list of objects to intersect.
-    if (data.objects) {
-      objectEls = this.el.sceneEl.querySelectorAll(data.objects);
-      this.objects = [];
-      for (i = 0; i < objectEls.length; i++) {
-        this.objects.push(objectEls[i].object3D);
+    if (objectsAreEls) {
+      for (objects = [], i = 0, len = objectsAreEls.length; i < len; i++) {
```
ngokevin: The `objects = []` assignment in the loop is a bit hard to follow. We also don't need `len` variable for a for loop `for (i = 0; i < objectEls.length; i++)`.

src/components/raycaster.js
```diff
@@ -60,25 +63,50 @@ module.exports.Component = registerComponent('raycaster', {
...
       }
-      return;
+    } else {
+      // If objects not defined, intersect with everything.
+      objects = this.el.sceneEl.object3D.children;
     }
 
-    // If objects not defined, intersect with everything.
-    this.objects = this.el.sceneEl.object3D.children;
+    for (this.objects = [], i = 0, len = objects.length; i < len; i++) {
```
ngokevin: Isn't this only necessary for non-recursive raycasting?

tests/components/raycaster.test.js
```diff
@@ -31,30 +31,46 @@ suite('raycaster', function () {
...
     });
 
-    test('defaults to intersecting all objects', function () {
+    test('defaults to intersecting all objects', function (done) {
       var el = this.el;
       var el2 = document.createElement('a-entity');
       var el3 = document.createElement('a-entity');
+      var i;
       var objects;
+      // add some geometry so raycast will actually work, and wait for it to be loaded
```
ngokevin: Usually capitalize + period all comments

tests/components/raycaster.test.js
```diff
@@ -75,16 +91,31 @@ suite('raycaster', function () {
   });
 
   suite('refreshObjects', function () {
+    setup(function createRaycasterAndTarget (done) {
+      // Define camera and light before tests to avoid injection.
+      this.el.sceneEl.appendChild(document.createElement('a-camera'));
+      var waitForMe = document.createElement('a-light');
+      waitForMe.addEventListener('loaded', function finishSetup () { setTimeout(function () { done(); }, 0); });
```
ngokevin: Don't need to specify the `ms` to 0 for setTimeout

tests/components/raycaster.test.js
```diff
@@ -75,16 +91,31 @@ suite('raycaster', function () {
...
       var numObjects = el.components.raycaster.objects.length;
       var sceneEl = this.el.sceneEl;
-      sceneEl.addEventListener('child-attached', doAssert);
+      // add some geometry so raycast will actually work
+      newEl.setAttribute('geometry', 'primitive: box');
+      sceneEl.addEventListener('child-attached', eventuallyDoAssert);
       sceneEl.appendChild(newEl);
+
+      function eventuallyDoAssert () {
+        if (newEl.hasLoaded) { doAssert(); } else { newEl.addEventListener('loaded', doAssert); }
```
ngokevin: Here as well:

```
if (clause) { 

} else {

}
```

tests/components/raycaster.test.js
```diff
@@ -120,12 +156,105 @@ suite('raycaster', function () {
       });
 
       targetEl.setAttribute('geometry', 'primitive: box; depth: 1; height: 1; width: 1;');
-      targetEl.setAttribute('material', '');
       targetEl.setAttribute('position', '0 0 -1');
-      targetEl.addEventListener('loaded', function finishSetup () {
+      el.sceneEl.appendChild(targetEl);
+      // `npm run test:forefox` needs the timeout for the tests to succeed.
```
ngokevin: Probably not worth mentioning, appendChild is asynchronous so it's a good idea to have always.

examples/test/raycaster/simple.html
```diff
@@ -0,0 +1,34 @@
...
+    <meta name="description" content="Raycaster - A-Frame">
+    <script src="../../../dist/aframe-master.js"></script>
+  </head>
+  <body>
+    <a-scene>
+      <a-entity camera look-controls wasd-controls raycaster="recursive: false; interval: 1000"></a-entity>
+    </a-scene>
+
+    <script>
+      var el = document.createElement('a-entity');
```
ngokevin: Within an ad-hoc component would be cool to contain the logic

examples/test/raycaster/simple.html
```diff
@@ -0,0 +1,34 @@
...
+      el.setAttribute('geometry', 'primitive: box');
+      el.setAttribute('material', 'color: #000');
+      el.setAttribute('position', '0 0 -2');
+      document.querySelector('a-scene').appendChild(el);
+
+      document.querySelector('a-entity[geometry]').addEventListener('raycaster-intersected', function (evt) {
```
ngokevin: query selector needed? this is `el`, right?

examples/test/raycaster/simple.html
```diff
@@ -0,0 +1,34 @@
...
+	// getting two intersection events per tick; same element, different faces...
```
ngokevin: What does this comment mean? A bug, or what we expect?

src/components/raycaster.js
machenmusik: see @donmccurdy recommendations above -- basically, this.objects should have only object3D when done, so the first part is exactly that, and the second part grabs the child object3D values because the top-level group one is useless.

src/components/raycaster.js
machenmusik: len came from @donmccurdy request - recanted after already done - be easier if y'all would agree with each other first ;-)

src/components/raycaster.js
machenmusik: see @donmccurdy recommendations above -- basically, this.objects should have only object3D when done, so we always do this second step now

src/components/raycaster.js
ngokevin: I see, sounds fine. We can move the assignment out though

examples/test/raycaster/simple.html
machenmusik: I think the behavior is permitted.

src/components/raycaster.js
donmccurdy: Right, my preference would be to give `objects` a consistent type `Array<Object3D>` as early as possible, rather than checking type in later loops.

And yes — this entire loop is only necessary for non-recursive raycasting. Maybe saves the raycaster some very trivial amount of work, too.

src/components/raycaster.js
donmccurdy: Haha yeah sorry.  My worry about `objectsAreEls.length` performance turned out to be unnecessary. This loop can just use normal syntax. :)

src/components/raycaster.js
donmccurdy: nit: `objectsAreEls` sounds like a boolean to me. Maybe call it `objectEls`?

donmccurdy:
If we choose to keep the `recursive:false` option around, then I'm happy with this PR after the remaining comments.

But I don't have a strong opinion about whether that option should exist, one way or another.

machenmusik:
@dmarcos @ngokevin - @donmccurdy and I think this one is done, did we want to merge or close?

ngokevin:
Thanks!

PR #2333, opened 2017-01-31
Add gltf-model component and system.

### LINKED ISSUE #2261: Consider putting gltf-model into core?
Discussing how best to maintain `gltf-model` with @xirvr; perhaps either (1) merge it into aframe-extras.loaders, or (2) merge it into core.

Advantages of having `gltf-model` in core:

- glTF is (IMO) the most promising asset format for the web. If we want to see that gain traction, I think we need to be willing to treat it as a default early on (even if the popularity isn't there yet).

Disadvantages:

- Animation should be treated as format-agnostic, using THREE.AnimationMixer. Currently the `animation-mixer` component lives in aframe-extras, and it's not clear that we want more animation-related code moving into core. We could ensure that core/gltf-model exposes its animations in a way that is compatible with aframe-extras/animation-mixer.

Further discussion of animation: https://github.com/xirvr/aframe-gltf/issues/9

***

Proposed syntax for any THREE.AnimationMixer-compatible format, including glTF, with animation:

```html
<a-entity foo-model="src: homestar.foo"
          animation-mixer="clip: run"></a-entity>
```

donmccurdy:
Fixes #2261.

**Description:**

Adds `gltf-model` component and system, based on @xirvr's [aframe-gltf implementation](https://github.com/xirvr/aframe-gltf).

**Changes proposed:**
- Include THREE.GLTFLoader.
- Currently excludes animations, but for Three.js r83+ an implementation might look very much like [`aframe-extras.animation-mixer`](https://github.com/donmccurdy/aframe-extras/blob/master/src/loaders/animation-mixer.js).
- ~~Adds example model (which is 11MB, so we can pick another from [gltf-test](https://github.com/cx20/gltf-test) if we prefer).~~


src/components/gltf-model.js
```diff
@@ -0,0 +1,37 @@
...
+  update: function () {
+    var self = this;
+    var el = this.el;
+    var src = this.data;
+
+    if (!src) { return; }
+
+    this.remove();
+
+    this.loader.load(src, function (gltfModel) {
```
dmarcos: Let's name the anonymous function for easier debugging

ngokevin:
We can host the sample models at https://github.com/aframevr/assets. Just commit it there, and it'll be available via ```cdn.aframe.io```

donmccurdy:
Updated.

docs/primitives/a-gltf-model.md
```diff
@@ -0,0 +1,33 @@
+---
+title: <a-collada-model>
```
ngokevin: collada

src/components/gltf-model.js
```diff
@@ -0,0 +1,37 @@
+var registerComponent = require('../core/component').registerComponent;
+var THREE = require('../lib/three');
+
+/**
+ * glTF model loader.
+ */
+module.exports.Component = registerComponent('gltf-model', {
+  schema: {type: 'model'},
```
ngokevin: Should this be multiprop to allow room for future props?

src/extras/primitives/primitives/a-gltf-model.js
```diff
@@ -0,0 +1,9 @@
+var getMeshMixin = require('../getMeshMixin');
+var registerPrimitive = require('../primitives').registerPrimitive;
+var utils = require('../../../utils/');
+
+registerPrimitive('a-gltf-model', utils.extendDeep({}, getMeshMixin(), {
```
ngokevin: Does the material component have an effect on `gltf` mesh? If not, we can remove `getMeshMixin`.

src/systems/gltf-model.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports.System = registerSystem('gltf-model', {
+  init: function () {
+    this.models = [];
+  },
+
+  /**
+   * Updates shaders for all glTF models in the system.
+   */
+  tick: function () {
+    var sceneEl = this.sceneEl;
```
ngokevin: we could test this code too

tests/components/gltf-model.test.js
```diff
@@ -0,0 +1,58 @@
+/* global assert, process, setup, suite, test */
+var entityFactory = require('../helpers').entityFactory;
+
+var SRC = '/base/tests/assets/box/Box.gltf';
+
+suite.only('gltf-model', function () {
```
ngokevin: got an `only` stuck on here (need to lint that, i do this a lot).

I've been following this template for tests now (https://github.com/aframevr/angle/blob/master/templates/component/tests/index.test.js#L6) ... closured variables for `el` and store the `component instance. saves a lot of proxy variables and grabbing the component.

docs/components/gltf-model.md
```diff
@@ -0,0 +1,61 @@
...
+
+We can load a glTF model by pointing to an asset that specifies the `src` to a glTF file.
+
+```html
+<a-scene>
+  <a-assets>
+    <a-asset-item id="tree" src="/path/to/tree.gltf"></a-asset-item>
+  </a-assets>
+
+  <a-entity collada-model="#tree"></a-entity>
```
ngokevin: collada

docs/components/gltf-model.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Loading Inline
+
+We can also load a glTF model by specifying the path directly within `url()`. Note this is less performant than going through the asset management system.
+
+```html
+<a-entity gltf-model="url(/path/to/tree.gltf)"></a-entity>
+```
+
+## More Resources
```
ngokevin: Can we add sections on playing animations until we get an API?

docs/components/gltf-model.md
```diff
@@ -0,0 +1,61 @@
+---
+title: gltf-model
+type: components
+layout: docs
+parent_section: components
+---
+
+[glTF][about-gltf] is an open project by Khronos providing a common, extensible format for 3D assets that is both efficient and highly interoperable with modern web technologies.
```
ngokevin: Could describe give the advantages and purpose of glTF over other common formats (obj, collada). The .psd vs .jpg analogy is a good one

docs/components/gltf-model.md
```diff
@@ -0,0 +1,61 @@
...
+
+Converters:
+
+- [FBX → glTF][fbx-converter] - coming soon.
+- [COLLADA → glTF][collada-converter].
+- [OBJ → glTF][obj-converter].
+
+See the [official glTF specification][spec] for available features, community resources, and ways to contribute.
+
+[fbx-converter]: http://gltf.autodesk.io/
```
ngokevin: I like putting link definitions above the paragraph its used to easier keep track

examples/test/model/index.html
```diff
@@ -11,13 +11,16 @@
       <a-assets>
```
ngokevin: Would be cool to add text labels under each model for the model format. And the scene could use a better background `<a-sky color="#FAFAFA">`

ngokevin: Could also add the different format types in the page title & meta description `Models (glTF, COLLADA, OBJ)`

src/components/gltf-model.js
donmccurdy: I slightly prefer multiprop, but can't actually think of any others we'd need. Maybe crossorigin. I'm expecting animation to be a separate component, but if not that would definitely need some properties. I've used `src: {type: 'asset'}` in most of my loaders so far i think.

ngokevin: Alright, can just leave it as is

xirvr:
I don't have any code comments, but wanted to say thanks for this PR and glTF FTW :)

examples/test/model/index.html
donmccurdy: scope creep alert 😉 

Added the meta title/description, but I'm getting errors trying to do `<a-text value="foo"></a-text>":

```
THREE.WebGLShader: gl.getShaderInfoLog() fragment WARNING: 0:102: '
' : extension directive should occur before any non-preprocessor tokens 
 1: precision highp float;
2: precision highp int;
3: #define SHADER_NAME ShaderMaterial
4: #define GAMMA_FACTOR 2
5: #define NUM_CLIPPING_PLANES 0
6: #define UNION_CLIPPING_PLANES 0
7: uniform mat4 viewMatrix;
...
```

docs/components/gltf-model.md
donmccurdy: Yes, but the answer is "write a component that looks very much like `animation-mixer`, so IMO it would be better to leave that for another PR.

docs/components/gltf-model.md
donmccurdy: Done.

src/extras/primitives/primitives/a-gltf-model.js
donmccurdy: Done (for both gltf and collada, for consistency)

examples/test/model/index.html
ngokevin: Is that on latest? We pushed some text patches switching to raw shader material.

machenmusik: In the latest and greatest master, that warning has been eliminated by using RawShaderMaterial for the sdf and msdf text shaders.

src/systems/gltf-model.js
donmccurdy: Done.

examples/test/model/index.html
donmccurdy: Rebase did the trick. ✅

docs/components/gltf-model.md
donmccurdy: Done.

donmccurdy:
I believe this is caught up now.

tests/systems/gltf-model.test.js
```diff
@@ -0,0 +1,37 @@
+/* global process, setup, suite, test, sinon, THREE */
+var entityFactory = require('../helpers').entityFactory;
+
+suite.only('glTF system', function () {
```
ngokevin: only

ngokevin:
looks good to me!

ngokevin:
great looks totally fantastic

PR #2334, opened 2017-01-31
ensure all components don't update if data has not changed (fixes #2322)

### LINKED ISSUE #2322: Rotation component continuously updated on stillstanding camera entity
The rotation component is continuously updated on a static camera entity (possible performance implications?).

- A-Frame Version: 0.4.0
- Platform / Device: Mac / Firefox 51.0.1

[https://gist.github.com/i3games/80636bb3bf2b2aa7d09569013d1df123](https://gist.github.com/i3games/80636bb3bf2b2aa7d09569013d1df123)

check console.log

The new values for pitch, yaw and roll are the same as the old values (0,0,0).

ngokevin:
**Description:**

**Changes proposed:**
- Do a equality check even if data is single-prop.
- Save an equality check on component initializations.
- Make utils.deepEqual handle simple values and nested objects.
- Unit tests for deepEquals and cases where componentChanged event should or should not fire.

### DIFF
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -155,8 +155,10 @@ suite('Component', function () {
       el.addEventListener('loaded', function () {
         el.setAttribute('dummy', '');
         data = el.getAttribute('dummy');
-        assert.shallowDeepEqual(data.direction, TestComponent.prototype.schema.direction.default);
-        assert.shallowDeepEqual(data.position, TestComponent.prototype.schema.position.default);
+        assert.shallowDeepEqual(data.direction,
+                                TestComponent.prototype.schema.direction.default);
+        assert.shallowDeepEqual(data.position,
+                                TestComponent.prototype.schema.position.default);
         assert.notEqual(data.direction, data.position);
         done();
       });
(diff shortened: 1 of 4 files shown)


tests/core/component.test.js
```diff
@@ -180,6 +182,105 @@ suite('Component', function () {
...
+    test('emits componentchanged for value', function (done) {
+      var el = entityFactory();
+      el.addEventListener('componentchanged', function (evt) {
+        if (evt.detail.name !== 'visible') { return; }
+        assert.shallowDeepEqual(evt.detail.oldData, true);
+        assert.shallowDeepEqual(evt.detail.newData, false);
+        assert.equal(evt.detail.name, 'visible');
+        done();
+      });
+      setTimeout(() => {
```
dmarcos: Why do we need a timeout? Can we move `var el = entityFactory();` to a setup method? https://github.com/aframevr/aframe/blob/master/tests/core/a-entity.test.js#L27

ngokevin:
- Timeout seems to be needed because the event listener doesn't seem to be registered in time when we immediately try to invoke it.

```js
addEventListener('componentchanged')  // Not called
setAttribute

// ...

addEventListener('componentchanged')  // Called
setTimeout(() => {
  setAttribute
});
```

- Moved entity setup to setup method.

PR #2335, opened 2017-01-31
Moves variable from closures to the component instance so multiple lo…

dmarcos:
…ok-controls don't collide

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -20,7 +20,11 @@ module.exports.Component = registerComponent('look-controls', {
   init: function () {
     var sceneEl = this.el.sceneEl;
 
+    // Aux variables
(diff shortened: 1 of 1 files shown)


ngokevin:
Tested r+

Would later like https://github.com/aframevr/aframe/pull/1853, but will need to rebase after this.

ngokevin:
Is this an issue with other components that utilize closures?

dmarcos:
yes, if people are going to use multiple instances. I found this bug with the motion controls work. I had to cameras with look-controls and weird things started to happen because they were stepping on each others toes

dmarcos:
We merge this and we can look at other controls

ngokevin:
So tracked-controls could have been buggy due to this?

PR #2336, opened 2017-02-01
Hint ES5 and ES6 syntax.

donmccurdy:
ES6 included for tests, to prevent linter warnings about template strings. Could also override `.jshintrc` under the tests/ directory.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -43,8 +43,8 @@
     "boss"          : false,     // true: Tolerate assignments where comparisons would be expected
     "debug"         : false,     // true: Allow debugger statements e.g. browser breakpoints.
     "eqnull"        : false,     // true: Tolerate use of `== null`
-    "es5"           : false,     // true: Allow ES5 syntax (ex: getters and setters)
-    "esnext"        : false,     // true: Allow ES.next (ES6) syntax (ex: `const`)
+    "es5"           : true,      // true: Allow ES5 syntax (ex: getters and setters)
+    "esnext"        : true,      // true: Allow ES.next (ES6) syntax (ex: `const`)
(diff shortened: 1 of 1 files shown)


dmarcos:
Is this to allow ES5 / ES6?

donmccurdy:
No intended effect on actual source code, or on the precommit lint for that matter. My editor shows warnings in some of the tests, because they use template strings, and this would silence that.

ngokevin:
Is the .jshintrc file being used? `semistandard` has its own linting rule set.

donmccurdy:
It's comparable to an `.editorconfig`. Used nowhere within the project, but can be detected by Sublime, Atom, etc. with linters installed.

ngokevin:
Sounds fine to me. Doesn't affect the actual CI linting, and it's mainly for the tests. The CI linting doesn't complain about ES6 either.

donmccurdy:
There is a semistandard sublime plugin too apparently; i can try that and see how it goes.

ngokevin:
Just saw we could override. I guess if we don't want ES6 in the actual codebase yet, we could just set it on the `tests/` dir

dmarcos:
Can we just remove template strings from the tests?

ngokevin:
They're pretty convenient, just need to tweak the linting setup a bit for editors.

dmarcos:
What about running tests in a browser that does not support them? I just don't want ES5 / ES6 and transpilers to cripple into the code base.

ngokevin:
The unit tests run in modern browsers that'll support the features natively, we don't plan on running our basic CI on 2 year old browsers. So we'll never be adding transpilation or Babel.

donmccurdy:
Meh.

PR #2341, opened 2017-02-02
Update component.md

stfairy:
**Description:**

The component name should be changed from "foo" to "bar" to match the one defined in the JavaScript code.

**Changes proposed:**
- Change the component name from "foo" to "bar".


ngokevin:
Good catch!

PR #2342, opened 2017-02-02
test chrome in travis as well

ngokevin:


dmarcos:
Travis does not like this

ngokevin:
Was still playing with getting Chrome on the Travis instance, it passes now!

PR #2350, opened 2017-02-04
Add shadow support.

### LINKED ISSUE #651: Support for shadows
<a href="https://github.com/caseyyee"></img></a> **Issue by [caseyyee](https://github.com/caseyyee)**
_Wednesday Nov 18, 2015 at 19:02 GMT_
_Originally opened as https://github.com/aframevr/aframe-core/issues/470_

---

It doesn't seem that lighting casts any shadows.  Would be great to have this.

donmccurdy:
**Description:**

Fixes #651.

Migrates [shadows from aframe-extras](https://github.com/donmccurdy/aframe-extras/tree/master/src/shadows), with a few changes to the API. Example includes shadows on the "Hello World" example, but we should probably come up with something more interesting.

Some of the shadow properties are non-default; this has worked better so far (from various bugs filed against extras) but perhaps we want to be consistent with THREE defaults instead? In any case, tweaks to the shadow camera properties are going to be par for the course on any serious scene.

**Changes proposed:**
- Add `shadow` component, which recursively enables cast/receive on child geometry. Note that it is recursive and on by default.
- Add shadow properties to the `light` component, for supported types.
- Add shadow setup to the `light` system. The shadowMap is enabled automatically, and not exposed in schema, but the shadow type can be set with `<a-scene light="shadowMapType: pcfsoft">`.

/cc @fernandojsg @takahirox

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -20,7 +20,21 @@ module.exports.Component = registerComponent('light', {
     intensity: {default: 1.0, min: 0, if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     penumbra: {default: 0, min: 0, max: 1, if: {type: ['spot']}},
     type: {default: 'directional', oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']},
-    target: {type: 'selector', if: {type: ['spot', 'directional']}}
+    target: {type: 'selector', if: {type: ['spot', 'directional']}},
+
+    // Shadows.
+    castShadow: {default: false, if: {type: ['point', 'spot', 'directional']}},
+    shadowBias: {default: 0, if: {castShadow: true}},
+    shadowCameraFar: {default: 500, if: {castShadow: true}},
+    shadowCameraFov: {default: 50, if: {castShadow: true}},
+    shadowCameraNear: {default: 0.5, if: {castShadow: true}},
+    shadowCameraTop: {default: 5, if: {castShadow: true}},
+    shadowCameraRight: {default: 5, if: {castShadow: true}},
+    shadowCameraBottom: {default: -5, if: {castShadow: true}},
+    shadowCameraLeft: {default: -5, if: {castShadow: true}},
+    shadowCameraVisible: {default: false, if: {castShadow: true}},
+    shadowMapHeight: {default: 512, if: {castShadow: true}},
+    shadowMapWidth: {default: 512, if: {castShadow: true}}
   },
 
   /**
@@ -44,6 +58,7 @@ module.exports.Component = registerComponent('light', {
 
     // Existing light.
     if (light && !('type' in diffData)) {
+      var shadowsLoaded = false;
       // Light type has not changed. Update light.
       Object.keys(diffData).forEach(function (key) {
         var value = data[key];
@@ -81,6 +96,24 @@ module.exports.Component = registerComponent('light', {
             break;
           }
 
+          case 'castShadow':
+          case 'shadowBias':
+          case 'shadowCameraFar':
+          case 'shadowCameraFov':
+          case 'shadowCameraNear':
+          case 'shadowCameraTop':
+          case 'shadowCameraRight':
+          case 'shadowCameraBottom':
+          case 'shadowCameraLeft':
+          case 'shadowCameraVisible':
+          case 'shadowMapHeight':
+          case 'shadowMapWidth':
+            if (!shadowsLoaded) {
+              self.updateShadow();
+              shadowsLoaded = true;
+            }
+            break;
+
           default: {
             light[key] = value;
           }
@@ -91,6 +124,7 @@ module.exports.Component = registerComponent('light', {
 
     // No light yet or light type has changed. Create and add light.
     this.setLight(this.data);
+    this.updateShadow();
   },
 
   setLight: function (data) {
@@ -116,6 +150,47 @@ module.exports.Component = registerComponent('light', {
         el.getObject3D('light-target').position.set(0, 0, -1);
       }
     }
+  },
+
+  /**
+   * Updates shadow-related properties on the current light.
+   */
+  updateShadow: function () {
+    var el = this.el;
+    var data = this.data;
+    var light = this.light;
+
+    light.castShadow = data.castShadow;
+
+    // Shadow camera helper.
+    var cameraHelper = el.getObject3D('cameraHelper');
+    if (data.shadowCameraVisible && !cameraHelper) {
+      el.setObject3D('cameraHelper', new THREE.CameraHelper(light.shadow.camera));
+    } else if (!data.shadowCameraVisible && cameraHelper) {
+      el.removeObject3D('cameraHelper');
+    }
+
+    if (!data.castShadow) { return light; }
+
+    // Shadow appearance.
+    light.shadow.bias = data.shadowBias;
+    light.shadow.mapSize.height = data.shadowMapHeight;
+    light.shadow.mapSize.width = data.shadowMapWidth;
+
+    // Shadow camera.
+    light.shadow.camera.near = data.shadowCameraNear;
+    light.shadow.camera.far = data.shadowCameraFar;
+    if (light.shadow.camera instanceof THREE.OrthographicCamera) {
+      light.shadow.camera.top = data.shadowCameraTop;
+      light.shadow.camera.right = data.shadowCameraRight;
+      light.shadow.camera.bottom = data.shadowCameraBottom;
+      light.shadow.camera.left = data.shadowCameraLeft;
+    } else {
+      light.shadow.camera.fov = data.shadowCameraFov;
+    }
+    light.shadow.camera.updateProjectionMatrix();
+
+    if (cameraHelper) { cameraHelper.update(); }
   },
 
   /**
@@ -188,6 +263,10 @@ module.exports.Component = registerComponent('light', {
    * Remove light on remove (callback).
    */
   remove: function () {
-    this.el.removeObject3D('light');
+    var el = this.el;
+    el.removeObject3D('light');
+    if (el.getObject3D('cameraHelper')) {
+      el.removeObject3D('cameraHelper');
+    }
   }
 });
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -26,7 +26,7 @@ module.exports.Component = registerComponent('light', {
     castShadow: {default: false, if: {type: ['point', 'spot', 'directional']}},
     shadowBias: {default: 0, if: {castShadow: true}},
     shadowCameraFar: {default: 500, if: {castShadow: true}},
-    shadowCameraFov: {default: 50, if: {castShadow: true}},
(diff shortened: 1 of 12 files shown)


src/components/light.js
```diff
@@ -81,6 +95,21 @@ module.exports.Component = registerComponent('light', {
...
+          case 'shadowCameraNear':
+          case 'shadowCameraTop':
+          case 'shadowCameraRight':
+          case 'shadowCameraBottom':
+          case 'shadowCameraLeft':
+          case 'shadowDarkness':
+          case 'shadowMapHeight':
+          case 'shadowMapWidth':
+            this.updateShadow();
+            break;
```
donmccurdy: Hm this gets executed repeatedly. I'll try to think of a better way to do this part.

takahirox:
`renderSingleSided` and `renderReverseSided` I requested aren't supported yet?

https://github.com/donmccurdy/aframe-extras/issues/117

donmccurdy:
Oops I forget. Added!

takahirox:
Thanks!

src/components/shadow.js
```diff
@@ -0,0 +1,43 @@
...
+    this.onMeshChanged = bind(this.update, this);
+    this.el.addEventListener('model-loaded', this.onMeshChanged);
+  },
+
+  update: function () {
+    var data = this.data;
+    this.el.object3D.traverse(function (node) {
+      if (node instanceof THREE.Mesh) {
+        node.castShadow = data.cast;
+        node.receiveShadow = data.receive;
```
takahirox: I think we need to update material here and in remove() like this
to handle set/removeAttribute('shadow') executed after scene is rendered.

    var material = node.material;
    if (material.isMultlMaterial === true) {
        for (var i = 0, il = material.materials.length; i < il; i++) {
            material.materials[i].needsUpdate = true;
        }
    } else {
        material.needsUpdate = true;
    }

Reference:
Run the following commands one by one in the console at
https://threejs.org/examples/webgl_animation_skinning_morph.html

    mesh.receiveShadow = false
    mesh.material.materials[0].needsUpdate = true;
    mesh.receiveShadow = true
    mesh.material.materials[0].needsUpdate = true;

takahirox:
Just curious to know if we can dynamically enable/disable shadow after scene is rendered once?

donmccurdy:
> Just curious to know if we can dynamically enable/disable shadow after scene is rendered once?

Programmatically, yes: `sceneEl.systems.light.setShadowMapEnabled(false)`. Systems don't support updates right now, so there's no pure-HTML way.

takahirox:
I see, supporting update is our next step :D

src/systems/light.js
```diff
@@ -73,5 +98,17 @@ module.exports.System = registerSystem('light', {
...
+  /**
+   * Enables/disables the renderer shadow map.
+   * @param {boolean} enabled
+   */
+  setShadowMapEnabled: function (enabled) {
+    var renderer = this.sceneEl.renderer;
+    this.shadowMapEnabled = enabled;
+    if (renderer) {
+      renderer.shadowMap.enabled = enabled;
+    }
```
takahirox: You need to update materials in a scene to dynamically handle shadow on/off 
after scene is rendered once if you switch `shadowMap.enabled`.

Reference:
Run the following commands one by one on the console at
https://threejs.org/examples/webgl_animation_skinning_morph.html

    renderer.shadowMap.enabled = false;
    mesh.material.materials[0].needsUpdate = true;
    renderer.shadowMap.enabled = true;
    mesh.material.materials[0].needsUpdate = true;

takahirox:
OK, I see.

I inserted some comments inline to support dynamic updates just for your information.

src/components/light.js
```diff
@@ -20,7 +20,21 @@ module.exports.Component = registerComponent('light', {
     intensity: {default: 1.0, min: 0, if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     penumbra: {default: 0, min: 0, max: 1, if: {type: ['spot']}},
     type: {default: 'directional', oneOf: ['ambient', 'directional', 'hemisphere', 'point', 'spot']},
-    target: {type: 'selector', if: {type: ['spot', 'directional']}}
+    target: {type: 'selector', if: {type: ['spot', 'directional']}},
+
+    // Shadows.
```
dmarcos: Should we make the light a component with dynamic schema if a the property `castShadows: true`?

src/components/shadow.js
```diff
@@ -0,0 +1,43 @@
...
+    cast: {default: true},
+    receive: {default: true}
+  },
+
+  init: function () {
+    this.onMeshChanged = bind(this.update, this);
+    this.el.addEventListener('model-loaded', this.onMeshChanged);
+  },
+
+  update: function () {
```
dmarcos: What happens if we add entities dynamically to the scene? Should we listen for the `loaded` event?

dmarcos: disregard. I thought this was the system

dmarcos: but still. A mesh could be added to the entity without being a model. Should we listen for `'object3dset' to update the shadow flags?

donmccurdy: `object3dset` sounds like a more reliable option, sure.

src/components/light.js
donmccurdy: Either way IMO. Preference?

ngokevin:
Any tips on tuning the shadows?

**Single Point Light + Torus Knot**

Point light in the center of the knot. Boxy shadow shapes.

screen shot 2017-02-06 at 6 56 00 pm

**Directional Light + Point Light**

Trying to animate the point light, spinning around the scene. Point light casts odd shape for the cylinder at an odd angle. Do I have to tweak `light.angle`?

screen shot 2017-02-06 at 6 53 06 pm

src/systems/light.js
```diff
@@ -14,13 +21,33 @@ var DEFAULT_LIGHT_ATTR = 'data-aframe-default-light';
  * @param {bool} userDefinedLights - Whether user lighting is defined.
  */
 module.exports.System = registerSystem('light', {
+  schema: {
+    shadowMapType: {default: 'pcf', oneOf: ['basic', 'pcf', 'pcfsoft']},
```
ngokevin: This seems like it should be a shadow system to match the shadow component?

```shadow="type: basic"```

ngokevin:
I continued playing around. Got it looking a bit better, but doesn't look as smooth as the [three.js shadow + point light example](https://threejs.org/examples/#webgl_shadowmap_pointlight), even though I've cloned all the parameters.

https://github.com/ngokevin/aframe/commit/958fdfc2b2cfe37508951dbc439f88ad81dc4cab

screen shot 2017-02-07 at 3 30 08 pm

I also added some meshes (sphere, box, cylinder), the shadows seem clipped and in a weird spot despite playing the shadow camera top/near/left/right and map width/height

src/systems/light.js
donmccurdy: Semantically nice. So then shadows would be enabled on the renderer if any `shadow` components are present, not based on whether `light="castShadow: true"` is set?

donmccurdy:
Re: tuning, I think scene dimensions are very much a contributor to how your camera parameters should be set. Not really sure how to optimize that. I've just pushed a `shadowCameraVisible` option that is helpful for debugging.

Directional lights will perform a fair bit better than point lights, as well.

src/systems/light.js
donmccurdy: I don't see this in most shadow demos/docs, is there a particular type of material this effects?

src/components/shadow.js
donmccurdy: (see below — is this just for THREE.MultiMaterial?)

src/systems/light.js
donmccurdy: Done.

src/components/shadow.js
donmccurdy: Done.

src/systems/light.js
takahirox: This is a general material issue.

This is necessary if we dynamically turn on/off shadow after scene is rendered once.
So it's kinda rare situation and maybe it isn't written in any documents.

Shadow related GLSL code is kinda hard coded depending on shadow parameters.
So we need to set `material.needsUpdate` `true` to update GLSL code
if we dynamically switch `shadowMap.enable` on/off.

Probably they'll help you understand.
https://github.com/mrdoob/three.js/blob/93bfc79da1758d4891f27f809598f836728b009b/src/renderers/webgl/WebGLPrograms.js#L190
https://github.com/mrdoob/three.js/blob/ddd8c2b1eb8a92a344b261c2125d7a2997248309/src/renderers/webgl/WebGLProgram.js#L358

src/components/shadow.js
takahirox: No. Not MultiMaterial specific, but general material specification.

src/systems/light.js
donmccurdy: Thanks, makes sense. Updated!

src/components/light.js
donmccurdy: Done.

ngokevin:
Still playing with the shadows. The camera helper really helps, I saw the camera was totally not where I expected it to be. Should we start adding baseline tests and docs?

src/systems/light.js
takahirox: Cool! And I found this's written in the Three.js document.

https://threejs.org/docs/#Manual/Next_Steps/How_to_update_things

donmccurdy:
Tests added.

ngokevin:
Was trying to get rid of some shadow clipping and debugging why my shadow camera was not lined up with my light, but playing with the shadow parameters from the Inspector doesn't have an effect. Are the updates working?

ngokevin:
Will save this for 0.5.1 until I get a good handle on it and document the tuning process.

donmccurdy:
Sounds good. I've fixed the updates — there were a couple more hoops to jump through for updating the shadow camera and cameraHelper. Also removed `shadowDarkness`, which isn't supported anymore.

donmccurdy:
Added a first draft of documentation. I'm not attached to the structure really: there's no *Systems* category in the docs, and most of the properties are on the `light` component, so that's where I put most of the documentation. Perhaps a post in the *Guides* section would be more appropriate?

We should probably also link to something explaining light baking, but I'm just learning that process myself.

donmccurdy:
Rebased, and reverted temporary changes to the hello-world example.

Any more to do here? Should an example be part of this PR?

ngokevin:
Thanks, this is my next priority. I have that example I was doing to copy the threejs example, if you have any cool ones we can add that too

ngokevin:
We can cherry-pick this example for now: https://github.com/ngokevin/aframe/commit/958fdfc2b2cfe37508951dbc439f88ad81dc4cab

I had *much* better and expected results setting `shadowCameraFov` to 90. With the help of the Inspector.

ngokevin:
I added an example, changed defaultCameraFov to 90 for better point lights.

Dynamic schemas we could do in a follow up.

PR #2345, opened 2017-02-02
Point polyfill to private fork to fix low res issue on phones + problem with webview in android (fixes #1940, #2117)

dmarcos:
…em with webview in android

The private fork cherry-picks a commit from a rejected PR (https://github.com/googlevr/webvr-polyfill/pull/175/commits/afd03b53c1830fb1073783b79112494deb9c0a0d) and makes `dpdb,json` local to the repo so we can easily add devices. 

@andymartinwork (from a-frame slack channel) confirmed that the webview issue is gone. We need more testing on the low res issue on more devices. We have to engage people to tell us if they're still having problems.

PR #2352, opened 2017-02-06
fix iOS version test

### LINKED ISSUE #2008: ios 10.1 blurry in stereo/vr mode
**Description:**

Something to do with the webvr polyfill buffer scale again?

nullcode:
**Description:**
Fixes #2008 

The blurry effect from #2008 was actually caused by incorrectly setting BUFFER_SCALE to 0.5 even on iOS 10.

**Changes proposed:**
- The workaround was created to fix webvr-polyfill/issues/102, however they PRed a different solution for it since, so this workaround in aframe might not be needed at all (haven't tested).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,7 +23,7 @@ window.WebVRConfig = window.WebVRConfig || {
 
 // Workaround for iOS Safari canvas sizing issues in stereo (webvr-polyfill/issues/102).
 // Should be fixed in iOS 10.
-if (/(iphone|ipod|ipad).*os.*(7|8|9)/i.test(navigator.userAgent)) {
+if (/(iphone|ipod|ipad).*os.(7|8|9)/i.test(navigator.userAgent)) {
   window.WebVRConfig.BUFFER_SCALE = 1 / window.devicePixelRatio;
 }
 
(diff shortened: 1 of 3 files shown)


dmarcos:
Nice! Thanks for investigating. I'm going to do some testing on my side

giordyb:
Just tried master + this patch on an iPhone 7 Plus w/ iOS 10 and it does indeed fix the issue for me! Finally! thank you @nullcode

ngokevin:
Seems the workaround is still needed, despite that issue being closed:

> Ok, this is now busted if BUFFER_SCALE is not 0.5. Seems like another iOS bug, which gets resolved if you set the size of the canvas to the same thing. - boris

ngokevin:
Rebasing the PR, and added tests. Will merge once it's green.

## Before

img_4573

## After

img_4574

ngokevin:
I actually don't see a difference on my iPhone 6, 10.1.1, even with https://github.com/aframevr/aframe/pull/2345 ... perhaps it's more discernable on a Plus. I'll let @dmarcos take a final look, or @giordyb can you post before/after screenshots?

dmarcos:
It does trick for me:
## Before
img_3332
## After
img_3331

PR #2361, opened 2017-02-08
[docs] changed wrong default values according to a-curvedimage.js

Stefie:
Default values taken from https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives/a-curvedimage.js

**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Thanks!

PR #2377, opened 2017-02-13
update Examples > Primitives > Models for glTF (and document on Examples index)

cvan:
**Description:**

glTF example updates (per https://twitter.com/superhoge/status/830993015577546752).

**Changes proposed:**
- Previously, only https://aframe.io/aframe/examples/test/model/ had an example of using a `glTF` model (using `<a-entity gltf-model>`)
- Updated https://aframe.io/aframe/examples/primitives/models/ to show `<a-gltf-model>`
- Updated https://aframe.io/aframe/examples/ to include text about model formats (so the examples are more easily searchable)

### DIFF
--- a/examples/boilerplate/3d-model/index.html
+++ b/examples/boilerplate/3d-model/index.html
@@ -2,8 +2,8 @@
 <html>
   <head>
     <meta charset="utf-8">
-    <title>3D Model</title>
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks!

PR #2382, opened 2017-02-13
Add check for tablet userAgents and consider them as mobile devices (…

### LINKED ISSUE #2309: Nexus 9 (and other tablets) entering in VR Mode and not using Gyroscope for movement by default
**Description:**

- A-Frame Version: V4
- Platform / Device: Nexus 9
- Reproducible Code Snippet or URL: http://codepen.io/tinchoforever/pen/ZLyEbj
This on a tablet (nexus 7 or 9) or Ipad should be 
1) Enabling by default devicemotion camera 
2) on vr mode, go full screen. 
On Aframe 2.0, 1) works and 2) says your device doesn't support webvr mode. I have been checking Google WebVr Poly and it works there perfectly. My guess is that this is something inside how Aframe decides to use VR Mode and Gyroscope Control of the Camera by default.

dmarcos:
…fix #2309)

**Description:**

**Changes proposed:**
-
-
-


ngokevin:
r+

ngokevin:
> Amazing, this fix the use of gyroscope but going to full screen mode is not possible.

ngokevin:
Will just merge this to at least get gyroscope working then.

PR #2383, opened 2017-02-14
Allow default lights to be manually disabled.

### LINKED ISSUE #2376: Allow default lights to be disabled.
It would be nice to have the option of disabling default lights, without providing my own. For example, lighting and shadows can be baked into textures.

Proposed syntax:

```html
<a-scene light="enableDefaultLights: false">

  <!-- ... -->

</a-scene>
```

donmccurdy:
Fixes #2376.

### DIFF
--- a/tests/systems/light.test.js
+++ b/tests/systems/light.test.js
@@ -59,8 +59,8 @@ suite('light system', function () {
 
     assert.notOk(document.querySelectorAll('[light]').length);
 
-    sceneEl.addEventListener('loaded', function () {
-      el.sceneEl.systems.light.setupDefaultLights();
+    el.sceneEl.systems.light.setupDefaultLights();
+    process.nextTick(function () {
       assert.notOk(document.querySelectorAll('[' + DEFAULT_LIGHT_ATTR + ']').length);
       done();
     });
--- a/tests/systems/light.test.js
+++ b/tests/systems/light.test.js
@@ -52,16 +52,18 @@ suite('light system', function () {
 
   test('it does not add default lights to scene if they are disabled', function (done) {
     var el = this.el;
-    var lightEl = document.createElement('a-entity');
-    lightEl.setAttribute('light', {defaultLightsEnabled: false});
+    var sceneEl = el.sceneEl;
+
+    // Systems cannot yet be updated via setAttribute().
+    sceneEl.systems.light.data.defaultLightsEnabled = false;
 
     assert.notOk(document.querySelectorAll('[light]').length);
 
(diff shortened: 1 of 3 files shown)


tests/systems/light.test.js
```diff
@@ -50,6 +50,20 @@ suite('light system', function () {
     el.appendChild(lightEl);
   });
 
+  test('it does not add default lights to scene if they are disabled', function (done) {
+    var el = this.el;
+    var lightEl = document.createElement('a-entity');
+    lightEl.setAttribute('light', {defaultLightsEnabled: false});
```
ngokevin: Shouldn't this be set on the scene and not an entity?

tests/systems/light.test.js
```diff
@@ -50,6 +50,20 @@ suite('light system', function () {
...
+
+    assert.notOk(document.querySelectorAll('[light]').length);
+
+    lightEl.addEventListener('loaded', function () {
```
ngokevin: I'd wait until `scene.loaded`

tests/systems/light.test.js
donmccurdy: Oops, too much copy/paste. Fixed.

tests/systems/light.test.js
```diff
@@ -50,6 +50,22 @@ suite('light system', function () {
...
+  test('it does not add default lights to scene if they are disabled', function (done) {
+    var el = this.el;
+    var sceneEl = el.sceneEl;
+
+    // Systems cannot yet be updated via setAttribute().
+    sceneEl.systems.light.data.defaultLightsEnabled = false;
+
+    assert.notOk(document.querySelectorAll('[light]').length);
+
+    sceneEl.addEventListener('loaded', function () {
```
ngokevin: Looks like we don't need to listen if we're going to call the method directly.

tests/systems/light.test.js
```diff
@@ -50,6 +50,22 @@ suite('light system', function () {
...
+      el.sceneEl.systems.light.setupDefaultLights();
+      assert.notOk(document.querySelectorAll('[' + DEFAULT_LIGHT_ATTR + ']').length);
```
ngokevin: I'd add a setTimeout since the method appends a DOM node.

tests/systems/light.test.js
donmccurdy: Done.

tests/systems/light.test.js
donmccurdy: Done.

Issue #2321, opened 2017-01-27
[docs] flesh out `embedded` component documentation


ngokevin:
**Description:**

Right now it's only one paragraph.

- We should embed an actual scene within that page to show it off.
- Add some CSS to the example of styling the scene (e.g, height + width).
- Note that only one scene can exist on a page at a time.
- Also mention I-Frames as an alternative as long as `allowvr` attribute is set, and preferably same origin.

PR #2384, opened 2017-02-14
Adding an example, tips and limitations

Utopiah:
As per #2321.

Note that the example hasn't been tested on the documentation server due to CSP limitations.

**Description:** Adding an example, tips and limitations

**Changes proposed:**
- clarified documentation with examples with CSS customization

### DIFF
--- a/docs/components/embedded.md
+++ b/docs/components/embedded.md
@@ -15,3 +15,46 @@ UI smaller.
 ```html
 <a-scene embedded></a-scene>
 ```
+
+<script src="/aframe/dist/aframe-master.min.js"></script>
+<div id="myEmbeddedScene">
+<a-scene embedded>
+  <a-entity position="0 0 3.8"><a-camera></a-camera></a-entity>
+  <a-sphere position="0 1.25 -1" radius="1.25" color="#EF2D5E"></a-sphere>
+  <a-box position="-1 0.5 1" rotation="0 45 0" width="1" height="1" depth="1"  color="#4CC3D9">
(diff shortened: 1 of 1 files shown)


docs/components/embedded.md
```diff
@@ -15,3 +15,46 @@ UI smaller.
...
+Amongst other things you can:
+* use CSS z-index to put element above or under the scene
+* use a transparent `<a-sky>` to overlay on top of an existing page, background and HTML elements includued
+* use HTML elements to control entities based on interactions with the content on your page
+* use `look-controls="enabled:false"` on the camera to prevent the user from freely looking in the scene
+
+
+## Known limitations
+* only one a-scene can exist on a page. Alternatively you can use `<iframe>` with ```allowvr="yes"```, preferably from the same origin, and display several scenes on the same page.
+* because of limited fullscreen capabilities of mobile browsers proper positionning on mobile can be problematic. It should be tested thoroughly. As a partial solution you can put the embedded scene at the top of the page.
```
dmarcos: Can you be more specific? :) I would like to know the problem you're refering. Maybe fixable?

Utopiah: I ended up with http://output.jsbin.com/fefehe/ either I leave it on top or ... entering VR doesn't behave as expected (i.e. quasi full screen stereoscopic). Trying "tricks" with hide/showing elements on WebKit didn't help.

If you have an example with content then embedding then more content that behaves as expected when entering VR I'd like to try. Example of what failed:

- <div>random text
- <a-scene>AFrame content
- <div>random text

forcing to do
- <a-scene>AFrame content
- <div>random text
- <div>random text

dmarcos:
Thanks

PR #2389, opened 2017-02-14
weird example scene

Utopiah:
The example scene has a texture that isn't available and have entities positioned on the camera. Newcomers might wonder what's "wrong".

**Description:** fixed the simple scene

**Changes proposed:**
- remove the texture
- positioned entities properly


ngokevin:
@a-frobot docs-v0.5.0

PR #2393, opened 2017-02-15
fix case where removing texture (null was not equal to empty string)

### LINKED ISSUE #2388: Removing or setting a null for SRC in a material doesn't works
**Description:** 
If we've an entity with a material set and a `src` loaded, if we try to clear the `src` value by removing the `src`attribute itself or by setting it to null or "", it will clear the value but it will still keep the texture binded to the entity

- A-Frame Version: v0.5.0
- Platform / Device: 
- Reproducible Code Snippet or URL: http://codepen.io/fernandojsg/pen/NdooBz

machenmusik:
Fixes #2388

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -11,25 +11,26 @@ module.exports.updateMapMaterialFromData = function (materialName, dataName, sha
   var el = shader.el;
   var material = shader.material;
   var src = data[dataName];
+
   // Because a single material / shader may have multiple textures,
   // we need to remember the source value for this data property
   // to avoid redundant operations which can be expensive otherwise
   // (e.g. video texture loads).
-  var shadowSrcName = '_texture_' + dataName;
+  if (!shader.materialSrcs) { shader.materialSrcs = {}; }
 
   if (!src) {
     // Forget the prior material src.
-    shader[shadowSrcName] = null;
+    delete shader.materialSrcs[materialName];
     // Remove the texture.
     setMap(null);
     return;
   }
 
   // Don't process if material src hasn't changed.
-  if (src === shader[shadowSrcName]) { return; }
+  if (src === shader.materialSrcs[materialName]) { return; }
 
   // Remember the new src for this texture (there may be multiple).
-  shader[shadowSrcName] = src;
+  shader.materialSrcs[materialName] = src;
 
   // If the new material src is already a texture, just use it.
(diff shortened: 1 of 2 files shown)


machenmusik:
easiest way to test - run local instance with this change via `npm start`, and load @fernandojsg's codepen on same machine, changing aframe script src to `http://localhost:9000/build/aframe.js`

ngokevin:
Can we add unit tests? Both for `setAttribute('material', 'src', '')` and `removeAttribute('material', 'src')`.

machenmusik:
ok, LGTM now...

machenmusik:
@dmarcos take a look at latest? should be much more comprehensible

machenmusik:
specifically, latest added comments try to better explain why having a single remembrance of last texture load attempt is insufficient, per discussion on #2402

src/utils/material.js
```diff
@@ -10,24 +11,42 @@ module.exports.updateMapMaterialFromData = function (materialName, dataName, sha
...
 
-  if (src) {
-    if (src === shader[shadowSrcName]) { return; }
-    // Texture added or changed.
-    shader[shadowSrcName] = src;
-    if (src instanceof THREE.Texture) { setMap(src); return; }
-    el.sceneEl.systems.material.loadTexture(src, {src: src, repeat: data.repeat, offset: data.offset, npot: data.npot}, setMap);
+  if (!src) {
+    // Forget the prior material src.
+    shader[shadowSrcName] = null;
```
dmarcos: Would not be better to store all the material src within an object and not hanging directly from the shader with a strange `__texture__` key? Something like `shader.materialsSrc = []`

machenmusik: I thought that the developers can arbitrarily define and use anything in the shader property namespace, so was trying to avoid collisions by using unlikely-to-collide naming, though I could be wrong.  Maybe it is better to make a map with name that is unlikely to collide as you suggest, although as it is internal, I thought usual convention was to use underscore prefix.

dmarcos: you can then append it to the `shader.material` object. Safer than a strange `key`

machenmusik: While I was in there, really we don't care about the data property name, we care about the material property name, which is what we use to actually set the texture.

machenmusik:
LGTM... @dmarcos ?

dmarcos:
Thanks!

PR #2402, opened 2017-02-16
Make sure that when an empty string  is passed to the `src` property …

### LINKED ISSUE #2388: Removing or setting a null for SRC in a material doesn't works

dmarcos:
…of the material the texture is removed (fix #2388)

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -8,26 +8,28 @@ var THREE = require('../lib/three');
  */
 module.exports.updateMapMaterialFromData = function (materialName, dataName, shader, data) {
   var el = shader.el;
+  var self = this;
   var material = shader.material;
   var src = data[dataName];
   var shadowSrcName = '_texture_' + dataName;
 
   if (src) {
+    // Return if material has not changed
     if (src === shader[shadowSrcName]) { return; }
-    // Texture added or changed.
-    shader[shadowSrcName] = src;
+    this.lastTextureRequestSrc = src;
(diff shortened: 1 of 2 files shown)


dmarcos:
@machenmusik @ngokevin alternative to https://github.com/aframevr/aframe/pull/2393

machenmusik:
the reason to have the check inside setMap is to handle the case where the property has changed in between the original call and the texture load that results in setMap (i.e. if you asked a texture to be loaded then changed it before the load completes, don't apply the texture in setMap)

removing it will cause a regression and hard to trace errors...

dmarcos:
@machenmusik what about now? A bit more clear?

machenmusik:
it seems that the check to avoid redundant texture load operations wouldn't work then, also not sure it would work correctly for cases whether multiple textures being loaded for same material/shader.

what in #2393 annoys you?

dmarcos:
This code is overly complicated and difficult to understand and maintain. I'm looking for opportunities to improve it.

machenmusik:
@dmarcos take a look at latest #2393 ?  should be much more comprehensible

Issue #2413, opened 2017-02-18
Expose `axismove` events as `trackpadmoved` on vive-controls


dmarcos:

machenmusik:
This is done, right?  Although there seem to be some spurious events firing per discussion on #2423

Issue #2379, opened 2017-02-13
Controller events being emitted double (e.g., vive-controls `gripup`)


ngokevin:
**Description:**

1. Add an event listener to a controller (`gripup`) in the `showcase/tracked-controls` example.
2. Add a console.log in the event listener.
3. Open the scene and press the button. 

See the log shows up twice.

- A-Frame Version: 0.5.0
- Platform / Device: Vive / Windows / FF Nightly

machenmusik:
is this due to https://github.com/aframevr/aframe/pull/2314#issuecomment-275812279 ?
specifically, if the event listeners for both `oculus-touch-controls` and `vive-controls` are active regardless of whether they are present...

ngokevin:
Probably. Need to poke around to think of the best solution. @dmarcos would know best. Maybe introduce a property to force event listening that will be disabled by default, but enabled by replayers. 

What I don't think will work is trying to detect whether a different controller exists, that might be messy.

machenmusik:
i think if you're not going to emulate the gamepad api (which would naturally emulate presence) then you need to inject something into the appropriate isControllerPresent to only attach when actively emulating, rather than leaving multiple instances of controls components with active listeners.  brings back to mind some of the earlier debates LOL.  note that if you have real vive and emulate touch or other supported controls, you would still have problems in all likelihood

machenmusik:
@dmarcos ?  perhaps you can add an `emulated` property that controls the event listener behavior when not present, as @ngokevin suggests

ngokevin:
https://github.com/aframevr/aframe/pull/2556

PR #2415, opened 2017-02-18
expose axismove vive controls issue 2413

cmvanb:
**Description:**
Exposes 'axismove' events from tracked-controls component in vive-controls as 'trackpadmoved'.

https://github.com/aframevr/aframe/issues/2413

**Changes proposed:**
- add onAxisMoved event listener function, which exposes the underlying event

### DIFF
--- a/src/components/vive-controls.js
+++ b/src/components/vive-controls.js
@@ -55,6 +55,7 @@ module.exports.Component = registerComponent('vive-controls', {
     this.onButtonUp = function (evt) { self.onButtonEvent(evt.detail.id, 'up'); };
     this.onButtonTouchStart = function (evt) { self.onButtonEvent(evt.detail.id, 'touchstart'); };
     this.onButtonTouchEnd = function (evt) { self.onButtonEvent(evt.detail.id, 'touchend'); };
+    this.onAxisMoved = bind(this.onAxisMoved, this);
     this.controllerPresent = false;
     this.everGotGamepadEvent = false;
     this.lastControllerCheck = 0;
@@ -180,7 +181,7 @@ module.exports.Component = registerComponent('vive-controls', {
     if (evt.detail.axis[0] === 0 && evt.detail.axis[1] === 0) {
       return;
     }
-    this.el.emit('trackpadmoved', evt);
+    this.el.emit('trackpadmoved', evt.detail);
   },
 
   onButtonEvent: function (id, evtName) {
--- a/src/components/vive-controls.js
+++ b/src/components/vive-controls.js
@@ -178,10 +178,8 @@ module.exports.Component = registerComponent('vive-controls', {
   },
 
   onAxisMoved: function (evt) {
-    if (evt.detail.axis[0] === 0 && evt.detail.axis[1] === 0) {
-      return;
-    }
-    this.el.emit('trackpadmoved', evt.detail);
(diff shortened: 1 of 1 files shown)


src/components/vive-controls.js
```diff
@@ -174,6 +177,13 @@ module.exports.Component = registerComponent('vive-controls', {
     controllerObject3D.position.set(0, -0.015, 0.04);
   },
 
+  onAxisMoved: function (evt) {
+    if (evt.detail.axis[0] === 0 && evt.detail.axis[1] === 0) {
+      return;
+    }
+    this.el.emit('trackpadmoved', evt.detail);
```
dmarcos: Do we want to map the axis to `x` and `y` coordinates for better ergonomics? Something like:
````javascript
this.el.emit('trackpadmoved', { x: evt,detail.axis[0], y: evt.detail.axis[1] });
````

File: src/components/vive-controls.js [hunk 2]
```diff
@@ -174,6 +177,13 @@ module.exports.Component = registerComponent('vive-controls', {
     controllerObject3D.position.set(0, -0.015, 0.04);
   },
 
+  onAxisMoved: function (evt) {
+    if (evt.detail.axis[0] === 0 && evt.detail.axis[1] === 0) {
```
dmarcos: Can we do this a one liner? if (evt.detail.axis[0] === 0 && evt.detail.axis[1] === 0) { return; }`

dmarcos:
Thanks! Almost done!

machenmusik:
really we should do this for any controller button with axes, right?

dmarcos:
@machenmusik yes

machenmusik:
i think the right thing to do is to add button id to the axismove event in tracked-controls, then axismove can be handled similar to button events and automatically use mapping (vive-controls, oculus-touch-controls and derivatives)

machenmusik:
i am kinda indifferent to turning axis values into a map, the nice thing about array representation is it supports 1D, 2D and in theory 3D

src/components/vive-controls.js
cmvanb: Will do!

src/components/vive-controls.js
cmvanb: Good idea, will do.

cmvanb:
Updated PR per review comments.

machenmusik:
should event be trackpadmove, along the lines of mousedown/up/move?

machenmusik:
@dmarcos #2379 may make this tricky to support correctly on touch as well, have you folka made any progress in fixing?

dmarcos:
I know that there's a few incosistencies but I always push in reviews to use the past participle as event names. The DOM event names are totally inconsistent

dmarcos:
Yeah! I don't like the harcoded relationship axis0 / axis1 to `trackpad` either. We should use the mapping object somehow. @machenmusik @cmvanp I'm merging this patch but feel free to come up with an improvement

dmarcos:
@cmvanb Thanks and congrats for your first patch!

machenmusik:
i think there may be confusion issues due to #2379 (@dmarcos since you changed listeners to be attached always, an entity that has both vive and touch controls will emit both vive and touch events on axismove) but following the pattern it should be straightforward to adopt similar for touch (thumbstickmoved).  might be worth generalizing all that to gamepads though since it has nothing to do with tracked controls pose

PR #2417, opened 2017-02-19
Expose animations on root gltf object.

donmccurdy:
The GLTFLoader response is formatted as:

```
{
  animations: {Array<THREE.AnimationClip>},
  cameras: {Array<THREE.Camera>},
  scene: {Scene},
  scenes: {Array<THREE.Scene>}
}
```

I'm not confident enough about the API choice to propose an animation component yet, but we should at least expose the underlying animation objects so that users can write their own playback. This PR takes the approach of THREE.ObjectLoader, attaching animations to the root object. By contrast, THREE.JSONLoader attaches the animations to Geometry objects (which doesn't make sense in this case). Another option would be emitting the animations with the `model-loaded` event details, but not attaching them anywhere.

An absolute-minimum animation component, compatible with this, would be:

```js
AFRAME.registerComponent('play-all-animations', {
  init: function () {
    this.el.addEventListener('model-loaded', function (e) {
      var model = e.detail.model;
      var clips = model.animations;

      if (!clips) { return; }

      this.remove();

      var mixer = this.mixer = new THREE.AnimationMixer(model);
      clips.forEach(function (clip) {
        mixer.clipAction(clip).play();
      });
    }.bind(this));
  },
  remove: function () {
    if (this.mixer) { this.mixer.stopAllAction(); }
  },
  tick: function (t, dt) {
    if (this.mixer && dt) { this.mixer.update(dt / 1000); }
  }
});
```

/cc @caseyyee

### DIFF
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -23,6 +23,7 @@ module.exports.Component = registerComponent('gltf-model', {
 
     this.loader.load(src, function gltfLoaded (gltfModel) {
       self.model = gltfModel.scene;
+      self.model.animations = gltfModel.animations;
       self.system.registerModel(self.model);
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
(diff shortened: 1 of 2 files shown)


dmarcos:
It looks good to me. Is there anyway to add a test?

caseyyee:
@donmccurdy this appears to work well :)  Thanks!

donmccurdy:
Sounds good, I should be able to add a test this weekend.

donmccurdy:
Updated.

caseyyee:
Thanks!

PR #2418, opened 2017-02-19
expose constants on default export

kellyselden:
For example, make it less error prone for your app to stay in sync with upstream changes to `DEFAULT_CAMERA_HEIGHT`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -80,6 +80,7 @@ module.exports = window.AFRAME = {
   ANode: ANode,
   AScene: AScene,
   components: components,
+  constants: require('./constants'),
   geometries: require('./core/geometry').geometries,
   registerComponent: registerComponent,
   registerElement: require('./core/a-register-element').registerElement,


dmarcos:
Constants are supposed to be internal to core. I would not like to expose them unless there's a good reason to. @kellyselden What's exactly your use case? Can you be more specific on what *keeping your app in sycn with upstream changes* mean?

kellyselden:
My app is updating the position of a camera, and needs to keep the `DEFAULT_CAMERA_HEIGHT` around to offset the Y param when setting, because the camera is auto adding 1.6 to the Y.

machenmusik:
Are you looking for this value?
`document.querySelector('[camera]').components.camera.data.userHeight`

machenmusik:
Perhaps you can set the camera user height property to zero to avoid the offset?

http://codepen.io/machenmusik/pen/QdewNZ

kellyselden:
>Are you looking for this value?
document.querySelector('[camera]').components.camera.data.userHeight

It would be nice to have the value before the camera is rendered.

> Perhaps you can set the camera user height property to zero to avoid the offset?

I want to use the default user height.

My flow is as follows:

```hbs
<a-camera position={{position}}></a-camera> // y = 0
```

```js
let position = document.querySelector('[camera]').getAttribute('position'); // y = 1.6
```

```hbs
<a-camera position={{position}}></a-camera> // y = 1.6
```

```js
let position = document.querySelector('[camera]').getAttribute('position'); // y = 3.2
```

dmarcos:
@kellyselden You can access the default value before initialization like this: `AFRAME.components.camera.schema.userHeight.default`

kellyselden:
> You can access the default value before initialization like this: AFRAME.components.camera.schema.userHeight.default

@dmarcos I saw that value when sniffing, but it only ever shows 0 for me. Even when on desktop and without a camera or scene.

machenmusik:
given that the default user height is not coming from VR or HMD setup, but is instead a hardcoded value, perhaps to avoid the offset issue, you can handle user height yourself and don't let the framework code mess things up for you.

ngokevin:
Can you use `cameraEl.getAttribute('camera').userHeight` to get the value?

kellyselden:
> Can you use cameraEl.getAttribute('camera').userHeight to get the value?

@ngokevin mentioned up here https://github.com/aframevr/aframe/pull/2418#issuecomment-281202407

That is the right value, but it would be very convenient to have that value after I import aframe, but before I start rendering the scene/generating html.

machenmusik:
it looks like because of the way handlebars is being used, you'd like to remove (subtract) the user height offset, but that rather than simply setting userHeight to zero so that no offset is applied, you'd like to be able to access the default user height value.  Assuming that's the case, and you would therefore use @dmarcos suggestion except that you see the wrong value at present, perhaps that should be filed as a bug so that the behavior can be corrected for you, rather than asking for internal constants to be exposed?

machenmusik:
interesting: 
- the default camera userHeight is, in fact, zero.
- if no camera is defined, and one is therefore injected into the scene, it gets a userHeight of 1.6, which is the value of an unexposed constant.
- because the userHeight is only set to non-default value upon injection, it is not available at time of `loaded` event, however it is available at `renderstart`

```
<script src="https://aframe.io/releases/0.5.0/aframe.min.js"></script>
<a-scene debug>
  <a-sky color="#3CF"></a-sky>
  <a-text align="center" position="0 1 -2" value="Hi there."></a-text>
</a-scene>
<script>
  document.querySelector('a-scene').addEventListener('renderstart', function (evt) {
    document.querySelector('a-text').setAttribute('value', 'Default camera userHeight:\n' + evt.detail.target.camera.el.components.camera.data.userHeight);
  });
</script>
```

kellyselden:
@machenmusik You are right, there may be a bug somewhere in here that I am trying to work around, but I'm so new to Aframe so I just hacked around it for the time being.

kellyselden:
I think what I observed was if you initially render this

`<a-camera position="0 0 0">`

position Y evaluates to 1.6. Then if you change it after render using handlebars (ember.js might remove the element and re-add it) to this

`<a-camera position="5 0 6">`

position Y evaluates to 0 now, because after initial render, Aframe stops autopositioning the Y param.

(my workaround code if anyone is interested)
https://github.com/kellyselden/ember-a-frame/blob/v0.0.2/tests/dummy/app/components/my-camera.js#L9
https://github.com/kellyselden/ember-a-frame/blob/v0.0.2/tests/dummy/app/components/my-camera.js#L19-L20
https://github.com/kellyselden/ember-a-frame/blob/v0.0.2/tests/dummy/app/components/my-camera.js#L46-L52

machenmusik:
A simpler approach may be to use something similar to the previous snippet to get the default userHeight from the injected camera, then modify said camera by setting userHeight to 0 so that offsets will no longer be applied, and to whatever subsequent Y values you want.  But I'm not sure that is necessary since the default userHeight is hardcoded to 1.6, so you don't really need to wait for renderstart since you know what you are going to get back.

kellyselden:
> But I'm not sure that is necessary since the default userHeight is hardcoded to 1.6, so you don't really need to wait for renderstart since you know what you are going to get back.

Now we are getting back to the original issue. I know the hardcoded value is 1.6, but to make my app less brittle, I'd like to get the value from the framework in case it ever gets tweaked.

machenmusik:
OK, then go ahead and wait for renderstart... e.g.
http://s.codepen.io/machenmusik/debug/GWpxzY/PNkvYLBoZnJA

kellyselden:
@machenmusik https://github.com/aframevr/aframe/pull/2418#issuecomment-282888041

> That is the right value, but it would be very convenient to have that value after I import aframe, but before I start rendering the scene/generating html.

I feel like we are going in circles. I've laid out the use case why it would be nice to have the value as a constant and not a runtime calculated value. The remaining question is whether it is worth it for you all to expose it.

ngokevin:
Sorry for the circles. I think given there's no good way to grab the value, we make constants public, but be careful about what we place in the constants folder. We can have a constants index.js that exposes public things. If something needs to be private, then constants live within their own modules or directly require constant files.

kellyselden:
@ngokevin It was mentioned above that `AFRAME.components.camera.schema.userHeight.default` should have it, but it is only ever zero in my testing, when it seems like it should be 1.6. Maybe there's bug there where it is not 1.6 when it is supposed to be? If so, that would be a good alternative to a public constants section.

ngokevin:
0 is the default for the camera component, but 1.6 is the default for the default camera.

machenmusik:
@kellyselden - @ngokevin statement matches my observations above https://github.com/aframevr/aframe/pull/2418#issuecomment-282892629

machenmusik:
it seems that you want two things to happen:

1 - let aframe pick the default camera height

2 - let your code set the height to the value it wants

and that due to your current methodology, you would need a private constant due to the default camera having a non-zero userHeight offset value.

the codepen above is an alternative method that still satisfies 1 and 2, but removes the requirement for the private constant to be exposed to your code, because the camera userHeight gets set to zero and thus you no longer need to offset the desired Y value by the userHeight.

without understanding more about your specific use case, it isn't clear why the alternate method does not suffice.

kellyselden:
Closing for now. @machenmusik's workaround is better than my workaround. It makes waiting for first render not as bad as before. Will reopen if any new info arises.

PR #2423, opened 2017-02-20
Emulated oculus and vive controls

### LINKED ISSUE #2379: Controller events being emitted double (e.g., vive-controls `gripup`)

machenmusik:
(Per request, this PR no longer extends #2413 / #2415 to `oculus-touch-controls` as `thumbstickmoved`)

Should fix #2379.

### DIFF
--- a/docs/components/oculus-touch-controls.md
+++ b/docs/components/oculus-touch-controls.md
@@ -22,6 +22,7 @@ mappings, events, and a Touch controller model.
 
 | Property             | Description                                        | Default Value        |
 |----------------------|----------------------------------------------------|----------------------|
+| emulated             | Whether to emulate (treat as present regardless).  | false                |
 | hand                 | The hand that will be tracked (i.e., right, left). | left                 |
 | model                | Whether the Touch controller model is loaded.      | true                 |
 | rotationOffset       | Offset to apply to model rotation.                 | 0                    |
--- a/docs/components/oculus-touch-controls.md
+++ b/docs/components/oculus-touch-controls.md
@@ -27,6 +27,9 @@ mappings, events, and a Touch controller model.
 | model                | Whether the Touch controller model is loaded.      | true                 |
 | rotationOffset       | Offset to apply to model rotation.                 | 0                    |
 
+The emulated property is rarely needed, but is provided for use cases that must force controller
+event listeners to be added despite no controllers actually being present, e.g. motion / event capture replays.
+
 ## Events
 
 | Event Name           | Description                    |
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -25,10 +25,11 @@ var EMULATED_TOUCH_THRESHOLD = 0.001;
 module.exports.Component = registerComponent('oculus-touch-controls', {
   schema: {
     hand: {default: 'left'},
+    emulated: {default: false},
     buttonColor: {type: 'color', default: '#999'},          // Off-white.
     buttonTouchColor: {type: 'color', default: '#8AB'},
     buttonHighlightColor: {type: 'color', default: '#2DF'}, // Light blue.
-    model: { default: true },
+    model: {default: true},
     rotationOffset: {default: 0} // no default offset; -999 is sentinel value to auto-determine based on hand
   },
 
@@ -65,6 +66,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.onModelLoaded = bind(this.onModelLoaded, this);
     this.onControllersUpdate = bind(this.onControllersUpdate, this);
     this.checkIfControllerPresent = bind(this.checkIfControllerPresent, this);
+    this.onGamepadConnectionEvent = bind(this.onGamepadConnectionEvent, this);
   },
 
   init: function () {
@@ -83,6 +85,12 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.isControllerPresent = isControllerPresent; // to allow mock
   },
 
+  update: function (oldData) {
+    if ((oldData.emulated !== undefined) && (this.data.emulated !== oldData.emulated)) {
+      this.checkIfControllersPresent();
+    }
+  },
+
   addEventListeners: function () {
     var el = this.el;
     el.addEventListener('buttonchanged', this.onButtonChanged);
@@ -91,9 +99,6 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     el.addEventListener('touchstart', this.onButtonTouchStart);
     el.addEventListener('touchend', this.onButtonTouchEnd);
     el.addEventListener('model-loaded', this.onModelLoaded);
-    el.sceneEl.addEventListener('controllersupdated', this.onControllersUpdate, false);
-    window.addEventListener('gamepadconnected', this.checkIfControllerPresent, false);
-    window.addEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);
   },
 
   removeEventListeners: function () {
@@ -104,26 +109,38 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     el.removeEventListener('touchstart', this.onButtonTouchStart);
     el.removeEventListener('touchend', this.onButtonTouchEnd);
     el.removeEventListener('model-loaded', this.onModelLoaded);
-    el.sceneEl.removeEventListener('controllersupdated', this.onControllersUpdate, false);
-    window.removeEventListener('gamepadconnected', this.checkIfControllerPresent, false);
-    window.removeEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);
   },
 
   checkIfControllerPresent: function () {
     var data = this.data;
     var isPresent = this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, { hand: data.hand });
-    if (isPresent === this.controllerPresent) { return; }
+    if ((isPresent || data.emulated) === this.controllerPresent) { return; }
     this.controllerPresent = isPresent;
     if (isPresent) { this.injectTrackedControls(); } // inject track-controls
+    if (isPresent || data.emulated) { this.addEventListeners(); } else { this.removeEventListeners(); }
+  },
+
+  onGamepadConnectionEvent: function (evt) {
+    this.everGotGamepadEvent = true;
+    // Due to an apparent bug in FF Nightly
+    // where only one gamepadconnected / disconnected event is fired,
+    // which makes it difficult to handle in individual controller entities,
+    // we no longer remove the controllersupdate listener as a result.
+    this.checkIfControllerPresent();
   },
 
   play: function () {
     this.checkIfControllerPresent();
-    this.addEventListeners();
+    this.addControllersUpdateListener();
+    window.addEventListener('gamepadconnected', this.onGamepadConnectionEvent, false);
+    window.addEventListener('gamepaddisconnected', this.onGamepadConnectionEvent, false);
   },
 
   pause: function () {
     this.removeEventListeners();
+    this.removeControllersUpdateListener();
+    window.removeEventListener('gamepadconnected', this.onGamepadConnectionEvent, false);
+    window.removeEventListener('gamepaddisconnected', this.onGamepadConnectionEvent, false);
   },
 
   updateControllerModel: function () {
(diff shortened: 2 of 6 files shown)


machenmusik:
Verified on both Vive (via SteamVR/OpenVR support) and Touch (direct Oculus support) with latest experimental Chromium.

cmvanb:
Looks like a nice upgrade. After #2415 was merged, I noticed that the trigger on the HTC Vive also fires onAxisMove (z axis) events. Does this PR handle the trigger axis as 'triggermoved' or as 'trackpadmoved'?

machenmusik:
trigger is not an axis, it is a button with a value associated -- or at least that is the case with Chromium.  per discussion with @dmarcos in Slack, we should definitely avoid browser-specific ids and button/axis mappings if possible, bleah.

machenmusik:
`axismove` events should have been fired before #2415 as well, meaning that you could easily handle `axismove` yourself if need be.  but for buttons with values, as trigger should be (and grip for Touch as well), it looks like no specific event was fired when the value changed, only the more generic 'buttonchanged' that is also fired on up/down/touchstart/touchend.

cmvanb:
Ok, so 'buttonchanged' still needs to be fired correctly for this specific case (htc vive trigger), with the correct value?

machenmusik:
`buttonchanged` should be firing now, see https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls.js#L151

oculus-touch-controls and vive-controls provide mappings to friendly names for up/down/touchstart/touchend, but looks like they don't handle changed.  that can be added, but would like to settle the button/axis disparity between Chromium and Nightly first to avoid redoing work.  @dmarcos any more word from Gecko folks?

machenmusik:
@cmvanb if you have time in next few days, can you confirm which environments (browser / HMD+controllers) receive events such as `triggerchanged` with the latest PR?

cmvanb:
I can test FF nightly + Vive for you. I will also try Chromium + Vive, although Chromium didn't work last time I tried. Will do that tonight or tomorrow. I don't have an Oculus to test with.

machenmusik:
(FYI, there is a brand new experimental chromium that resolves previous issues with latest steamvr)

cmvanb:
I tested the branch with vive + FF nightly 54.0a1 (2017-02-23) (64-bit).

When clicking the trackpad, 'trackpadchanged' event is fired (multiple times/frames) as expected.
When sliding a finger over the trackpad, 'trackpadmoved' event (multiple times/frames) is fired as expected. The x/y values are correct.
When pulling the trigger down (fast), several events are fired in this order:
'trackpadmoved'
'trackpadmoved'
'triggerdown'
'triggerchanged'
'trackpadmoved'
'triggerchanged'

On releasing the trigger:
'trackpadmoved'
'triggerup'
'triggerchanged'
'trackpadmoved'
'triggerchanged'

It seems like the 'trackpadmoved' event is still being fired for the trigger. Otherwise, the values seem correct to me.

PS: Chromium still isn't working for me. I'm using the version made available here: https://webvr.info/get-chrome/

machenmusik:
thanks @cmvanb !

I believe there is currently a fundamental difference in how the latest experimental Chromium and Nightly model Vive and Touch expose controllers with respect to buttons and axes.

Chromium represents trigger (Vive and Touch) as a button that has an analog value, and not as an axis... same thing for grip on Touch.  The only axes are trackpad (Vive) and thumbstick (touch).

Nightly appears to be representing buttons that have analog values (like trigger) as axes, for example the new Oculus Touch support was exposing controllers with 3 buttons and 4 axes in Nightly, rather than the 6 buttons and 2 axes in Chromium,  This causes problems since `axismove` events are currently fired by `tracked-controls` when trigger changes, with no indication of which axes changed.

@cmvanb can you confirm whether the `triggerchanged` button state provides an analog value using Nightly?  If not, then even the button events are inconsistent, so suppressing spurious `trackpadmoved` events would not suffice.

IMO it is highly undesirable to have browser-specific gamepad API mappings for the same controllers on the same machine.  @cvan @kearwood @toji - would appreciate your thoughts on how best to resolve discrepancies.

src/components/oculus-touch-controls.js
```diff
@@ -81,6 +88,13 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.previousButtonValues = {};
     this.bindMethods();
     this.isControllerPresent = isControllerPresent; // to allow mock
+    this.getGamepadsByPrefix = getGamepadsByPrefix; // to allow mock
+  },
+
+  update: function (oldData) {
+    if ((oldData.emulated !== undefined) && (this.data.emulated !== oldData.emulated)) {
+      this.checkIfControllersPresent();
```
ngokevin: This LOC isn't getting covered by the tests. Maybe something wrong with the condition?

I think just `if (this.data.emulated !== oldData.emulated)` alone should work

ngokevin:
How difficult to split the `emulated` changes from the `axismove` additions?

src/components/oculus-touch-controls.js
machenmusik: the original reason for that was to avoid flagging the first update call (right after init) as needing the controllers check, since play() will do it.

ngokevin: I see. Maybe `emulated in oldData` + a comment is a minor improvement.

machenmusik: in order for that line to be covered, the entity would have to have emulated be true right at initial definition, and the tests use a setup function where no attributes are provided initially.

machenmusik:
Easiest way to "split" the axismove handling is to simply comment out the onAxisMoved handler, in which case the code won't fire.  but remember there is already a different version of axismove committed into master for vive only, not for oculus, so we'd be disabling that vive-only functionality by doing so

machenmusik:
so rather than split out the axismove handling, I added axis change propagation, which allows culling of spurious moves.

machenmusik:
@cmvanb if you get the change, can you check to see whether the latest version of this PR prevents the spurious `trackpadmoved` events you were seeing before?

docs/components/oculus-touch-controls.md
```diff
@@ -34,34 +35,43 @@ mappings, events, and a Touch controller model.
 | triggerup            | Trigger released.              |
 | triggertouchstart    | Trigger touched.               |
 | triggertouchend      | Trigger no longer touched.     |
+| triggerchanged       | Trigger changed.               |
```
cvan: FYI: Firefox ships with an implementation of experimental Gamepad events: `gamepadbuttondown`, `gamepadbuttonup`, and `gamepadaxismove` (fired on `window`).

If you're curious, might want to check these links out:

* https://github.com/w3c/gamepad/issues/4
* https://github.com/w3c/gamepad/issues/56
* https://github.com/w3c/gamepad/pull/15

src/components/oculus-touch-controls.js
```diff
@@ -139,26 +174,38 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
-  injectTrackedControls: function () {
+  injectTrackedControls: function (gamepad) {
     var el = this.el;
     var data = this.data;
     var isRightHand = data.hand === 'right';
-
-    // since each hand is named differently, avoid enumeration
+    // Inject with specific gamepad id, if provided. This works around a temporary issue
+    // where Chromium uses `Oculus Touch (Right)` but Firefox uses `Oculus Touch (right)`.
     el.setAttribute('tracked-controls', {
```
cvan: might want to check out these Gamepad librarys I wrote; maybe there are some parts you can reuse.

* https://github.com/MozVR/oculus/blob/5aa680f/assets/js/gamepad.js
* https://github.com/MozVR/oculus/blob/e73ea2c/assets/js/common.js#L107
* https://github.com/MozVR/gamepad-plus
* https://github.com/MozVR/gamepad-plus/blob/b3cc035/demo.js

src/components/vive-controls.js
```diff
@@ -202,8 +202,8 @@ module.exports.Component = registerComponent('vive-controls', {
     Object.keys(axesMapping).map(function (key) {
       var value = axesMapping[key];
       var detail = {};
```
cvan: where is `detail` used? perhaps do `var detail = evt.detail || {};`?

src/components/oculus-touch-controls.js
```diff
@@ -288,8 +288,8 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     Object.keys(axesMapping).map(function (key) {
       var value = axesMapping[key];
       var detail = {};
-      var changed = false;
-      value.map(function (axisNumber) { changed |= evt.detail.changed[axisNumber]; });
+      var changed = !evt.detail.changed;
+      if (!changed) { value.map(function (axisNumber) { changed |= evt.detail.changed[axisNumber]; }); }
```
cvan: you can use `forEach` instead of `map`

src/components/vive-controls.js
machenmusik: no, `detail` is used to build the translated/remapped event detail.

cvan: ah, I didn't see it in the diff - my bad

docs/components/oculus-touch-controls.md
machenmusik: good to know, thanks.  this code needs to support other browsers that currently lack those, let's revisit when more widely available?

src/components/oculus-touch-controls.js
machenmusik: yeah, I have a similar not-just-tracked-controls set for aframe-arcade, but that's beyond this PR's scope, and last we discussed, core team preference was to keep non-tracked controllers (gearvr, daydream, oculus remote, xbox controller, etc) out of core.  maybe file new issue to support controllers included with VR equipment including 0DOF (emulated pose and orientation) and 3DOF (emulated pose, true orientation)?

machenmusik:
@dmarcos relayed discussion with mozillans where different Oculus Touch button/axis mapping is not likely to change soon (unlike id string).  Can I urge Mozillans to reconsider?  At least make analog buttons like trigger and grip generate actual button events, rather than treating them like sliders... otherwise even simple trigger button down/up will become unreliable

dmarcos:
I don't understand at first sight what the emulated property does. If I don't get it quickly that I wrote the original controls many other people will get confused too.

machenmusik:
See https://github.com/aframevr/aframe/pull/2446#issuecomment-283754949

ngokevin:
`emulated` won't be used often, almost purely for motion capture replaying. Where you want to force controller event listeners to be added despite no controllers being present. Currently, we add controller event listeners regardless which causes events to be double-emitted.

machenmusik:
@ngokevin you might actually be better off also capturing emitted events from `oculus-touch-controls` and `vive-controls` (perhaps daydream and gearvr, not sure about leap hands) as well as those from tracked-controls, at which point you don't even need emulated true, you just need the app to be paying attention.

ngokevin:
@dmarcos Any thoughts? Issue is when we made controls always add event listeners, controller events would double emit for hand-controls since both Oculus + Vive event handlers are attached. The proposal is an `emulated` property which controls when we force event listeners to be added. Let us know if you have other proposals (e.g., naming `forceEventListeners`, `forcePresent`), but the use case is for emulating + motion capture.

machenmusik:
should we deprecate this in favor of #2446?

ngokevin:
It's more easy to review if things are in isolation, so we'll try not to clump non-cohesive changes together in one PR. @dmarcos still needs to take a look and understand, but the tl;dr is that we forced event listeners to be added regardless of Gamepad present in order for motion capture, but that causes doubly-emitted events (vive + oculus controls). The proposed `emulated` property will only add *all* event listeners if specifically enabled.

machenmusik:
Right.  if you merge this PR, #2446 will simplify itself :-)

ngokevin:
Would you mind breaking out the `axismove` changes to another PR?

machenmusik:
so which things between this and #2446 want to be separate PRs?  there is some overlap so may be nontrivial, and i am tied up for a little while

i think we definitely want the fix for #2448 that @caseyyee confirmed via cherry-pick, that one is easy, but maybe not entirely done with Nightly Rift+Touch without #2446 

i think we need to fix the double-event issue introduced unless we move back to `auto-detect-controllers` paradigm, and IIRC the axismove changes here are needed now to prevent spurious events from Nightly having control elements be both buttons and axes, so a bit hesitant to separate temporarily only to have them merged back together shortly

machenmusik:
w.r.t. separating PRs: I think these delineations would work

- cherry-picked #2510 - already done

- axismove and nightly touch support - need new mapping structure for both

- emulated

may get a chance to look at that late this evening.

ngokevin:
After discussing in Slack, it seems we'll revert https://github.com/aframevr/aframe/pull/2314 for another solution, so we don't need `emulated`. Thanks!

PR #2427, opened 2017-02-23
Add correct vrDisplay.isConnected check to checkHeadsetConnected()

### LINKED ISSUE #2426: checkHeadsetConnected () doesn't use vrDisplay.isConnected, which breaks Microsoft Edge
For privacy reasons, Microsoft Edge does not expose the orientation of the HMD before requestPresent(). This breaks checkHeadsetConnected():

function checkHeadsetConnected () {
  var orientation;
  controls.update();
  orientation = dolly.quaternion;
  if (orientation._x !== 0 || orientation._y !== 0 || orientation._z !== 0) {
    return true;
  }
  return false;
}

The fix here is to use vrDisplay.isConnected() (if available)

- A-Frame Version: 0.50.0
- Platform / Device: Microsoft Holographic Devices

frankolivier:
Fixes #2426

Microsoft Edge doesn't report HMD orientation before requestPresent (for  privacy reasons) which causes checkHeadsetConnected() to always return false; on Microsoft Edge, the 'Enter VR' button falls back to the fullscreen experience, and the content is not rendered in the HMD.

Fix it is to add a vrDisplay.isConnected check (if available) to get the correct answer from hardware.
-
-


src/utils/device.js
```diff
@@ -7,6 +7,8 @@ var controls = new THREE.VRControls(dolly);
  * orientation is available
  */
 function checkHeadsetConnected () {
+  var vrDisplay = controls.getVRDisplay();
+  if (vrDisplay && vrDisplay.isConnected) { return vrDisplay.isConnected; }
```
dmarcos: Can we do? `if (vrDisplay && vrDisplay.isConnected) { return true; }`

patrickkettner: can't `isConnected` be false, though?

dmarcos: Sure! We can do that

ngokevin: Can we change to `'isConnected' in vrDisplay` and it looks good.

frankolivier:
done

ngokevin:
Thanks!

PR #2428, opened 2017-02-23
Fix camera height when re-entering VR (#2394)

### LINKED ISSUE #2394: Camera Y position off entering VR second time on Oculus Rift in Chrome
**Description:**

1.  Go to an A-Frame example with a floor. https://ngokevin.github.io/kframe/components/text/examples/vaporwave/ is a good example.
2. Go into VR with the Rift. Observe how floor is at expected level.
3. Go out of VR and back into VR. Observe how floor is now at head level.

Looking at `document.querySelector('a-scene').camera.el.getAttribute('position')` clearly shows that the camera has a Y position at the correct head height the first time I enter VR, but that it has a Y position just above or below 0 the second time I enter VR.

I have to reload the page before I again get the floor positioned correctly, and again it's only the first time I enter VR it has the correct position.

I'm pretty sure this is an A-Frame bug and not a WebVR API or polyfill bug, as https://webvr.info/samples/XX-vr-controllers.html doesn't have the problem I describe. I rarely see this problem with the Vive, but it has happened a few times.

- A-Frame Version: 0.5.0
- Platform / Device: Windows 10, Chromium VR Feb 1 2017 build, Oculus Rift CV 1 with touch

machenmusik:
This PR attempts to resolve #2394, which happens repeatedly on Rift and infrequently on Vive.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -36,6 +36,9 @@ module.exports.Component = registerComponent('look-controls', {
     } else {
       enableGrabCursor();
     }
+
+    // Reset previous HMD position when we exit VR.
+    sceneEl.addEventListener('exit-vr', this.onExitVR);
   },
 
   update: function (oldData) {
@@ -75,6 +78,7 @@ module.exports.Component = registerComponent('look-controls', {
     this.onTouchStart = bind(this.onTouchStart, this);
     this.onTouchMove = bind(this.onTouchMove, this);
     this.onTouchEnd = bind(this.onTouchEnd, this);
+    this.onExitVR = bind(this.onExitVR, this);
   },
 
   setupMouseControls: function () {
@@ -285,6 +289,10 @@ module.exports.Component = registerComponent('look-controls', {
 
   onTouchEnd: function () {
     this.touchStarted = false;
+  },
+
+  onExitVR: function () {
+    this.previousHMDPosition.set(0, 0, 0);
   }
 });
 
--- a/tests/components/look-controls.test.js
+++ b/tests/components/look-controls.test.js
@@ -12,6 +12,19 @@ suite('look-controls', function () {
     });
   });
 
+  suite('exit-vr', function () {
+    test('reset previous HMD position upon exit-vr event', function (done) {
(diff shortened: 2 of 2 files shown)


machenmusik:
@RSpace trying this blind as I can't test until weekend, if you have a chance to try it please comment with your findings

dmarcos:
The camera component restores the camera pose when exiting VR. https://github.com/aframevr/aframe/blob/master/src/components/camera.js#L163
What about handling there?

machenmusik:
@dmarcos see further discussion in #2394 -- the re-entry problem is avoided by not doing removeHeightOffset/addHeightOffset (which is what happens when you set userHeight to 0) or restoreCameraPose (which currently cannot be avoided).  A first workaround would be to also decline to change position/rotation in restoreCameraPose when userHeight is 0, but that does not seem to be the core problem...

RSpace:
I can confirm that this PR fixes the issue. I no longer get the wrong height when entering VR the second and additional times on the Oculus Rift. I have also tested on other devices and seen no regressions:

* Oculus Rift, Win 10, Chromium VR Feb 17 build: OK 
* HTC Vive, Win 10, Chromium VR Feb 17 build: OK
* Gear VR, Android 6, Chrome 56: OK
* Cardboard, Android 6, Chrome 56: OK

Thanks a lot for finding and fixing the issue @machenmusik ! I definitely recommend merging this PR before next release.

dmarcos:
@RSpace thanks for testing. I need to spend some time understanding the problem a bit better to see what is the best solution.

machenmusik:
@dmarcos if this fixed it (yay!) the problem was that the delta application of pose that `look-controls` does was not being properly reset when exiting VR, which only manifests the second and subsequent times VR is entered.

machenmusik:
@RSpace thanks so much for the effort to verify especially regression check!

machenmusik:
@dmarcos have you had a chance to consider?  happy to say it definitely works.
http://codepen.io/machenmusik/pen/ZeGxZe

dmarcos:
Thanks!

PR #2433, opened 2017-02-24
Fix edge-case 'isPlaying()' TypeError in src/core/a-entity.js

wosevision:
"Uncaught (in promise) TypeError: Cannot read property 'isPlaying' of null" originally reported in #1483 still happens in rare cases. Upgrading to 0.5.0 does not fix this issue.

- Added check for self.parentEl before calling self.parentEl.isPlaying() to guard against edge cases where parent is null after dynamic DOM shuffling


wosevision:
@ngokevin Hmm, that sounds reasonable – okay, I'll try to come up with a Codepen with the simplest possible reproduction. It may prove tricky as it is _quite_ an edge case (would probably only show up in limited scenarios, i.e. integrations) but I can definitely confirm there is still at least one more scenario where `self.parentEl` ends up null.

Apologies for the half-baked pull request!

ngokevin:
Np, we originally did what you did, add a check, but if it's manifesting again we need to see what's really up

ngokevin:
Duping https://github.com/aframevr/aframe/pull/2521, will write a test to investigate

PR #2434, opened 2017-02-26
Document and test unique array component property type behavior

wmurphyrd:
**Description:**

The behavior of `setAttribute('component', {...}) changed starting with v0.4.0 such that array property types behave uniquely. 

1. Arrays are assigned by reference, so later changes to the variable that was passed in `setAttribute` will also be reflected in the component data
2. Updating an array property by passing it in an object to `setAttribute` will never trigger the component's `update` method (because it is assigned by reference [here](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L707) before the check for changes to call update [here](https://github.com/aframevr/aframe/blob/dec48d73bbe0d0189ae203eb12431ffc5521e0a2/src/core/component.js#L221))

See full discussion in #2227. 

**Changes proposed:**

- Document current behavior
- Add test cases to detect future changes in behavior

I also added documentation of the clobber update flag, but I can remove this if its not something you want in the docs.

### DIFF
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -338,25 +338,28 @@ suite('a-entity', function () {
     });
 
     test('partial updates of array properties assign by reference', function () {
+      // Arrays are assigned by reference and mutable.
       var sourceArray = [1, 2, 3];
-      registerComponent('array-comp', {
-        schema: { arr: { type: 'array' } }
+      registerComponent('test', {
+        schema: {array: {type: 'array'}}
       });
-      this.el.setAttribute('array-comp', { arr: sourceArray });
-      assert.strictEqual(this.el.getAttribute('array-comp').arr, sourceArray);
+      this.el.setAttribute('test', {array: sourceArray});
+      assert.strictEqual(this.el.getAttribute('test').array, sourceArray);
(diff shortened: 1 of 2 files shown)


tests/core/a-entity.test.js
```diff
@@ -337,6 +337,29 @@ suite('a-entity', function () {
...
 
+    test('partial updates of array properties assign by reference', function () {
+      var sourceArray = [1, 2, 3];
+      registerComponent('array-comp', {
+        schema: { arr: { type: 'array' } }
+      });
+      this.el.setAttribute('array-comp', { arr: sourceArray });
+      assert.strictEqual(this.el.getAttribute('array-comp').arr, sourceArray);
+    });
+    test('partial updates of array-type properties do not trigger update', function () {
```
ngokevin: I see the issue now...I would expect added or removed elements from the array to trigger an update.

ngokevin: Never mind, I think it makes sense now.

ngokevin:
I updated the docs some more because I forgot to do that when we changed the clobber logic for setAttribute, and did some style nits to the tests. Thanks!

Issue #2378, opened 2017-02-13
Future problem of a-asset-item and gltf-model combination


takahirox:
**Description:**

I've run into this problem while I was trying to bump Three.js to r84.

In r84 `THREE.GLTFLoader` expects that glTF file is loaded as arraybuffer.
(In r83 it loads as text)

https://github.com/mrdoob/three.js/blob/r84/examples/js/loaders/GLTFLoader.js#L28

But `a-asset-item` loads data as DOMString(text) in advance and keeps it in `THREE.Cache`.

`THREE.FileLoader` returns cached data if cache hits.

I mean, `THREE.FileLoader` returns cached text even though
`THREE.GLTFLoader` sets `THREE.FileLoader.setResponseType('arraybuffer')`
if `a-asset-item` loads in advance.

Actually `gltf-model.test` fails because of this issue with Three.js r84.

I have some ideas in my mind to solve this problem.

1. We don't use `a-asset-item` for `gltf-model`. Easy but inconvenient.
2. Disable `THREE.Cache`. Easy but worse performance.
3. Enable arraybuffer `a-asset-item` (#2112). (Add responseType prop?)
4. `a-asset-item` checks the extension of file and loads as arraybuffer if it's .gltf or .glb

I prefer 3.
Which one do you guys like or any other better ideas?

/cc @donmccurdy

ngokevin:
3. sounds good to me.

donmccurdy:
[Context for the change in r84](https://github.com/mrdoob/three.js/pull/10500#issuecomment-270354703).

I also like (3). Should responseType be set automatically, or do we just document that it needs to be done for glTF? 

GLTFLoader has an internal cache, sort of, so disabling THREE.Cache for `gltf-model` only might be a 5th option.

takahirox:
IMO, it'd be nice to have 3. and 4.

We provide a way to set responseType in `a-asset-item` like this

    <a-asset-item id="model" responseType="arraybuffer" src="model.gltf"></a-asset-item>

It'd be worth for A-Frame non-core components which need to load data as non-text.
(`a-mmd` I've making is actually one of them)
They document the data type.

And also `a-asset-item` automatically sets appropriate responseType for the data type which
A-Frame core supports. (arraybuffer for .gltf and .glb)

takahirox:
I'm working on updating `a-asset-item` and have two questions.

1. Why does `a-asset-item` call `THREE.Cache.files[src] = responseText;` 

[...]

PR #2442, opened 2017-03-01
Array buffer asset item

takahirox:
**Description:**

This PR adds `responseType` attribute to `a-asset-item`.
It's necessary for `THREE.GLTFLoader` in r84. #2378
And it'll be helpful for user components which need `arraybuffer` data.

It has a limitation.
If we attempt to load the same data as different `responseType`s,
only the `responseType` of the first attempt is applied
because it loads from cache  if exists.

I think it won't cause any problems because
there's no situation that we wanna load the same data with different `responseType`s.
(If exists, let me know.)

**Changes proposed:**
- Add `responseType` attribute to `a-asset-item`. 
- Add `responseType` tests

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -98,9 +98,7 @@ registerElement('a-asset-item', {
       value: function () {
         var self = this;
         var src = this.getAttribute('src');
-        var responseType = this.getAttribute('responseType');
-        if (responseType === null) responseType = 'text';
-        fileLoader.setResponseType(responseType);
+        fileLoader.setResponseType(this.getAttribute('response-type' || 'text'));
         fileLoader.load(src, function handleOnLoad (textResponse) {
           self.data = textResponse;
           /*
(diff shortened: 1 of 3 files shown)


src/core/a-assets.js
```diff
@@ -98,6 +98,9 @@ registerElement('a-asset-item', {
       value: function () {
         var self = this;
         var src = this.getAttribute('src');
+        var responseType = this.getAttribute('responseType');
+        if (responseType === null) responseType = 'text';
```
ngokevin: `var responseType = this.getAttribute('response-type') || 'text';`

html attribute case insensitive

takahirox: Oops, thanks! I'll update.

ngokevin:
I've added docs and made the quick update. Thanks!

takahirox:
Thanks for the update!

I have a question.

You changed the order of

    THREE.Cache.remove(XHR_SRC);
    var assetItem = document.createElement('a-asset-item');

Is there any special reasons?
And if so, we should change the order here, too?

https://github.com/aframevr/aframe/blob/3c074bdde332f3d28649eb9bbf3c3e207107fe9d/tests/core/a-assets.test.js#L276

PR #2446, opened 2017-03-02
Nightly oculus touch

### LINKED ISSUE #2379: Controller events being emitted double (e.g., vive-controls `gripup`)

machenmusik:
This PR adds support to #2513 for the different Oculus Touch controller mapping exhibited by the current Firefox Nightly build(s).  Note that Nightly is currently NOT exposing capacitive touch events as Chromium does; therefore, some `hand-controls` poses are more cumbersome since buttons need to be fully actuated rather than merely touched.

Per discussion, this PR no longer includes #2423 in its entirety, as another approach has been agreed to fix #2379.

### DIFF
--- a/docs/components/oculus-touch-controls.md
+++ b/docs/components/oculus-touch-controls.md
@@ -45,26 +45,26 @@ mappings, events, and a Touch controller model.
 | griptouchstart       | Grip button touched.           |
 | griptouchend         | Grip button no longer touched. |
 | gripchanged          | Grip button changed.           |
-| Adown                | A button pressed.              |
-| Aup                  | A button released.             |
-| Atouchstart          | A button touched.              |
-| Atouchend            | A button no longer touched.    |
-| Achanged             | A button changed.              |
-| Bdown                | B button pressed.              |
-| Bup                  | B button released.             |
-| Btouchstart          | B button touched.              |
-| Btouchend            | B button no longer touched.    |
-| Bchanged             | B button changed.              |
-| Xdown                | X button pressed.              |
-| Xup                  | X button released.             |
-| Xtouchstart          | X button touched.              |
-| Xtouchend            | X button no longer touched.    |
-| Xchanged             | X button changed.              |
-| Ydown                | Y button pressed.              |
-| Yup                  | Y button released.             |
-| Ytouchstart          | Y button touched.              |
-| Ytouchend            | Y button no longer touched.    |
-| Ychanged             | Y button changed.              |
+| abuttondown          | A button pressed.              |
+| abuttonup            | A button released.             |
+| abuttontouchstart    | A button touched.              |
+| abuttontouchend      | A button no longer touched.    |
+| abuttonchanged       | A button changed.              |
+| bbuttondown          | B button pressed.              |
+| bbuttonup            | B button released.             |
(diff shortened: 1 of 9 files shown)


dmarcos:
I don't understand why `emulated` is for and why it's needed. If I cannot understand it quickly by reading the docs myself that I wrote the original controls :) I bet there will be a lot of other people that will be confused.

machenmusik:
:)
Are you asking for docs change to provide a fuller explanation, or code change, or something else?  

The reason why `emulated` is needed is because per some of our original discussions on how to support multiple types of controllers, the strategy employed is to speculatively pre-attach the various components that could be used (`oculus-touch-controls` and `vive-controls`), and they are supposed to do nothing if their respective controllers are not present.  When #2314 committed changes to always attach event listeners, that broke the do-nothing contract, and so with both `oculus-touch-controls` and `vive-controls` attached, both components reacted to events that something like `tracked-controls` would generate, leading to e.g. double-event bug #2379.  If applications perform non-idempotent operations as a result of getting tracked controller events, the effects can be more significant.

To fix #2379, the do-nothing-by-default contract is restored, but then emulation / replaying needs some other mechanism to indicate that the events should be acted upon even though the physical controller is not present.  That is the `emulated` flag.  

Although, actually it turns out that for replay examples like https://ngokevin.github.io/kframe/scenes/aincraft/?avatar-recording=recording.json 
even #2314 was not really enough to allow correct functionality, because it is `tracked-controls` events that are actually being replayed, and `tracked-controls` was only injected when the controllers were physically present.  (in the example, you would not see the hand pose change.) -- you need to explicitly have `tracked-controls` present (but it must not match any actual controller, to avoid collisions in events between actual and replay), and if either #2446 or #2423 is included, then you need to add `emulated:true` to whichever one of `oculus-touch-controls` or `vive-controls` you'd like the replay to use.

machenmusik:
(@ngokevin said this: https://github.com/aframevr/aframe/pull/2423#issuecomment-283795941)

machenmusik:
I liked @ngokevin 's commentary so added something along those lines to the docs, to @dmarcos 's point.

Anything more you'd like to do with this?  I think we can deprecate #2423 in favor of this one.

machenmusik:
@dmarcos per discussion on #2513...

> Note that Nightly is currently NOT exposing capacitive touch events as Chromium does; therefore, some hand-controls poses are more cumbersome since buttons need to be fully actuated rather than merely touched.

It appears that Nightly is currently exposing trigger as axis ONLY, so in order to generate typical button events (down/up, even if we ignore touch), there is no choice but to add some emulation code (see https://github.com/aframevr/aframe/pull/2446/commits/b2f803edb08a7e80886fe67749057d2ed39f400c)

machenmusik:
update: latest nightly builds appear broken, however the 3-22 build has oculus touch mapping as 6 buttons and 2 axes (albeit without touch) - as that is same as chromium, may be able to simplify the mapping again.  closing this for now

PR #2454, opened 2017-03-04
prevents an infinite loop with the init method calls setAttribute on the component being initialized

dmarcos:

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -34,6 +34,7 @@ var Component = module.exports.Component = function (el, attrValue, id) {
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
+  this.el.components[this.attrName] = this;
   this.updateProperties(attrValue);
 };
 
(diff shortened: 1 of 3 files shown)


ngokevin:
What use case to allow this and mask the error? Should we add an error message?

src/core/component.js
```diff
@@ -34,6 +34,7 @@ var Component = module.exports.Component = function (el, attrValue, id) {
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
+  this.el.components[this.attrName] = this;
```
ngokevin: Why the move here? It seems the entity should be managing the lifecycle of the component when we can, not both ways.

dmarcos: If you call `setAttribute` in the init routine the new call will trigger a new component instantiation nested in the first one because everything happens in the constructor. Setting the component on the entity early in the constructor prevents the nested initializations when calling set attribute.

dmarcos: Still you want nested serAttribute calls to update the attrValue

PR #2455, opened 2017-03-04
antialias by default on desktop too

ngokevin:
**Description:**

Anti-aliasing is desirable for VR. If we're targeting room scale, enabling it makes sense as the default.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -35,7 +35,7 @@ var warn = utils.debug('core:a-scene:warn');
  * @member {object} systems - Registered instantiated systems.
  * @member {number} time
  */
-module.exports = registerElement('a-scene', {
+module.exports.AScene = registerElement('a-scene', {
   prototype: Object.create(AEntity.prototype, {
     defaultComponents: {
       value: {
@@ -340,13 +340,9 @@ module.exports = registerElement('a-scene', {
 
     setupRenderer: {
       value: function () {
-        var canvas = this.canvas;
-        // Set at startup. To enable/disable antialias
-        // at runttime we would have to recreate the whole context
-        var antialias = this.getAttribute('antialias') === 'true';
         var renderer = this.renderer = new THREE.WebGLRenderer({
-          canvas: canvas,
-          antialias: antialias || window.hasNativeWebVRImplementation,
+          canvas: this.canvas,
+          antialias: shouldAntiAlias(this),
           alpha: true
         });
         renderer.setPixelRatio(window.devicePixelRatio);
@@ -517,3 +513,20 @@ function exitFullscreen () {
     document.webkitExitFullscreen();
   }
 }
+
+/**
+ * Determines if renderer anti-aliasing should be enabled.
+ * Enabled by default if has native WebVR or is desktop.
+ *
+ * @returns {bool}
+ */
+function shouldAntiAlias (sceneEl) {
+  // Explicitly set.
+  if (sceneEl.getAttribute('antialias') !== null) {
+    return sceneEl.getAttribute('antialias') === 'true';
+  }
+
+  // Default not AA for mobile.
+  return !sceneEl.isMobile;
+}
+module.exports.shouldAntiAlias = shouldAntiAlias;  // For testing.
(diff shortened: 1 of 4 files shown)


src/core/scene/a-scene.js
```diff
@@ -344,10 +344,10 @@ module.exports = registerElement('a-scene', {
         var canvas = this.canvas;
         // Set at startup. To enable/disable antialias
         // at runttime we would have to recreate the whole context
-        var antialias = this.getAttribute('antialias') === 'true';
+        var antialias = this.getAttribute('antialias') !== 'false';
```
dmarcos: We can enable antialias by default only on desktop

ngokevin:
Updated, with tests. Logic is (explicitly enabled or disabled) or (has native webvr or is not mobile)

ngokevin:
@dmarcos r?

src/core/scene/a-scene.js
```diff
@@ -518,3 +514,19 @@ function exitFullscreen () {
...
+
+/**
+ * Determines if renderer anti-aliasing should be enabled.
+ * Enabled by default if has native WebVR or is desktop.
+ *
+ * @returns {bool}
+ */
+function shouldAntiAlias (sceneEl) {
+  // Explicitly set.
+  if (sceneEl.getAttribute('antialias') === 'true') { return true; }
```
dmarcos: Can we do ` if (sceneEl.getAttribute('antialias') !== null) { return sceneEl.getAttribute('antialias') === 'true'; }`

src/core/scene/a-scene.js
```diff
@@ -518,3 +514,19 @@ function exitFullscreen () {
...
+  if (sceneEl.getAttribute('antialias') === 'false') { return false; }
+
+  // Has WebVR or is desktop.
+  return window.hasNativeWebVRImplementation || !sceneEl.isMobile;
```
cvan: this will match Daydream, FYI

cvan: this okay?

src/core/scene/a-scene.js
```diff
@@ -341,13 +341,9 @@ module.exports = registerElement('a-scene', {
...
       value: function () {
-        var canvas = this.canvas;
-        // Set at startup. To enable/disable antialias
-        // at runttime we would have to recreate the whole context
-        var antialias = this.getAttribute('antialias') === 'true';
         var renderer = this.renderer = new THREE.WebGLRenderer({
-          canvas: canvas,
-          antialias: antialias || window.hasNativeWebVRImplementation,
+          canvas: this.canvas,
+          antialias: shouldAntiAlias(this),
```
cvan: could this instead by exposed as an `antialias` component on `<a-scene>`? that way folks can force enable/disable it if they want.

src/core/scene/a-scene.js
```diff
@@ -518,3 +514,19 @@ function exitFullscreen () {
     document.webkitExitFullscreen();
   }
 }
+
+/**
+ * Determines if renderer anti-aliasing should be enabled.
+ * Enabled by default if has native WebVR or is desktop.
```
cvan: can this be mentioned in the Docs? (especially if you go the `antialias` component route)

src/core/scene/a-scene.js
dmarcos: antialias cannot be disabled once the context has been created. You would have to create a new one and resubmit all textures and buffers (http://stackoverflow.com/questions/27554969/dynamically-turn-on-off-antialiasing-and-shadows-in-webglrenderer).

ngokevin: There's a code quality issue to make a `renderer` component to be configurable, `antialias` would be part of that.

dmarcos: I still don't think it's could be modifiable at runtime. You would invalidate all the components that have submitted textures / buffers to the GPU.

dmarcos: I pretty elaborated machinery could be implemented to make antialias dynamic but that's something you don't want to do in 99.9999999% of the cases. We can revisit in the future if there's a very good reason

dmarcos:
Thanks!

src/core/scene/a-scene.js
ngokevin: Yeah, I agree, it cannot be configurable at runtime. I just mean configurable at init time.

PR #2456, opened 2017-03-05
Bump Three.js to r84

takahirox:
**Description:**

This PR bumps Three.js to r84.

Because of `THREE.GLTFLoader` specification in r84 or newer,
`.gltf` and `.glb` files need to be loaded as `arraybuffer`.

I let `a-asset-item` automatically set `arraubuffer` response-type attribute for them.

I think this automation is necessary because of
- Usability
- A-Frame compatibility
- Not confusing people who know .gltf is json based text data

FYI, `THREE.GLTFLoader` in r84
- (partially) supports 1.1
- can't load 2.0 (in r85, it (partially) supports 2.0)

**Changes proposed:**
- Update Three.js to r84
- `a-asset-item` detects glTF files from their extensions and automatically sets `arraybuffer` response-type attribute.


src/core/a-assets.js
```diff
@@ -98,7 +98,26 @@ registerElement('a-asset-item', {
...
-        fileLoader.setResponseType(this.getAttribute('response-type') || 'text');
+        var responseType = this.getAttribute('response-type');
+        /*
+          If response-type isn't set, we load files as text
+          (default XMLHttpRequest.responseType).
+          But we load .gltf and .glb files as arraybuffer
+          because of THREE.GLTFLoader specification.
+        */
+        if (!responseType) {
+          var extension = '';
```
ngokevin: I'd make this code a private exported `function inferResponseType (src)` and test that individually.

takahirox: Thanks for the comment.
What does 'test on its own' mean in this case?

ngokevin: We can test the function by calling it directly rather than creating a-asset-item

takahirox: That's very useful. I will update.

takahirox:
Updated.

PR #2462, opened 2017-03-06
Accept glTF file which does not include default scene

takahirox:
**Description:**

This PR lets `gltf-model` accept glTF file which doesn't include default scene parameter.

    this.loader.load(src, function gltfLoaded (gltfModel) {
      self.model = gltfModel.scene;

`gltfModel.scene` (default scene) is optional in glTF specification.
If it isn't specified in the data file, let's use `gltfModel.scenes[0]` instead.
This follows Three.js glTF loader example.

https://github.com/mrdoob/three.js/blob/75384cfe6341b823e96cbec673c4ad92389b20cf/examples/webgl_loader_gltf.html#L231

**Changes proposed:**
- if default scene isn't specified in the glTF file, use scenes[0] instead.

### DIFF
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -22,7 +22,7 @@ module.exports.Component = registerComponent('gltf-model', {
     this.remove();
 
     this.loader.load(src, function gltfLoaded (gltfModel) {
-      self.model = gltfModel.scene;
+      self.model = gltfModel.scene || gltfModel.scenes[0];
       self.model.animations = gltfModel.animations;
       self.system.registerModel(self.model);
       el.setObject3D('mesh', self.model);
--- a/tests/assets/box/Box_no_default_scene.gltf
+++ b/tests/assets/box/Box_no_default_scene.gltf
@@ -1,0 +1,249 @@
+{
+    "accessors": {
+        "accessor_21": {
+            "bufferView": "bufferView_29",
+            "byteOffset": 0,
+            "byteStride": 0,
+            "componentType": 5123,
+            "count": 36,
+            "type": "SCALAR"
+        },
+        "accessor_23": {
+            "bufferView": "bufferView_30",
+            "byteOffset": 0,
+            "byteStride": 12,
+            "componentType": 5126,
+            "count": 24,
+            "max": [
+                0.5,
+                0.5,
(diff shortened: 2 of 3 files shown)


src/components/gltf-model.js
```diff
@@ -22,7 +22,7 @@ module.exports.Component = registerComponent('gltf-model', {
     this.remove();
 
     this.loader.load(src, function gltfLoaded (gltfModel) {
-      self.model = gltfModel.scene;
+      self.model = gltfModel.scene || gltfModel.scenes[0];
```
ngokevin: Wouldn't this always return `gltfModel.scene`?

takahirox: You mean, it'd be in `THREE.GLTFLoader`?
That's a good idea.
I'll talk to Three.js folks.

But at least with Three.js r84, we need this out of `THREE.GLTFLoader`.

ngokevin: If `gltfModel.scenes[0]` exists, then `gltfModel.scene || gltfModel.scenes[0]` will always be `gltfModel.scene`.

takahirox: `gltfModel.scene` can be one of `gltfModel.scenes[0~gltfModel.scenes.length-1]`,
or `undefined` if default scene isn't specified.

default scene indicates the index in `gltfModel.scenes`.
`gltfModell.scene` = `gltfModel.scenes[defaultScene];`.
But default scene is optional in glTF specification.
If default scene isn't specified in the glTF data file,
`gltfModel.scene` will be `undefined`.

So `gltfModel.scene` can be `undefined` even though gltfModex.scenes[0] exists.

donmccurdy: Yes, in practice I haven't found multiple scenes in a glTF yet, but there's no requirement that `scene` === `scenes[0]`, or that `scene` even be provided.

The change looks good; we could also add a `scene:0` property to the `gltf-model` schema to select one, if that becomes a thing.

ngokevin: Sorry I misread. Didn't notice one was singular and one was plural.

ngokevin:
Can we add a unit test for this?

takahirox:
Added

PR #2471, opened 2017-03-08
Fixes #2470 Added timer

### LINKED ISSUE #2470: Consider using THREE.Clock or performance.now in render(time)
**Description:** 
Currently we use `requestAnimationFrame`'s timestamp as the time to be used on `scene.render`, to update the `TWEEN` and the scene itself.
The following code (https://github.com/aframevr/aframe/blob/master/vendor/VREffect.js#L186-L194) is used to detect which `requestAnimationFrame` we should call, if we've one HMD connected it will call the `vrDisplay`'s one. 

```javascript
this.requestAnimationFrame = function( f ) {
		if ( vrDisplay !== undefined ) {
			return vrDisplay.requestAnimationFrame( f );
		} else {
			return window.requestAnimationFrame( f );
		}
	};
```

The problem is that even if you have a HMD connected, it takes a while to init it so for a few frames it will call `window.requestAnimationFrame` and after that `vrDisplay.requestAnimationFrame`.
`window.requestAnimationFrame` will returns the timestamp since the page was **loaded**, but 
`vrDisplay.requestAnimationFrame` on firefox nightly returns the timestamp relative to the **first load of the browser** not the current page (@kearwood could correct me if I'm wrong here).
So that generates that jump in the timestamp that we see in nightly:
image
We can see that:
- 0 is generated by https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L392
- Then we have 3 frames with 3xxx generated by the `windows.RAF`
- Finally `vrDisplay.RAF` jumps in and sends the, much higher, timestamp as I had the browser opened for a while.

So we have two problems:
- The jumps between these values generates incorrect deltas and can generate weird issues (Like the one we've seen in TWEEN: https://github.com/aframevr/aframe/issues/2175)
- The timestamp is typically used to control the timeline of the experience, and we're using an incorrect value for that.

So my proposal is to use `THREE.Clock` or `DATE / performance.now` directly on the render loop to be consistent across browsers and independent from the `RAF` being used.

fernandojsg:
Introducing a timer to fix the problems with the `requestAnimationFrame`

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -469,7 +469,6 @@ module.exports = registerElement('a-scene', {
         var effect = this.effect;
         var delta = this.clock.getDelta() * 1000;
         this.time = this.clock.elapsedTime * 1000;
-
         if (this.isPlaying) { this.tick(this.time, delta); }
         this.animationFrameID = effect.requestAnimationFrame(this.render);
         effect.render(this.object3D, this.camera);
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -389,7 +389,8 @@ module.exports = registerElement('a-scene', {
               if (window.performance) {
                 window.performance.mark('render-started');
               }
-              sceneEl.render(0);
+              sceneEl.clock = new THREE.Clock();
+              sceneEl.render();
               sceneEl.renderStarted = true;
               sceneEl.emit('renderstart');
             }
@@ -441,7 +442,6 @@ module.exports = registerElement('a-scene', {
     tick: {
       value: function (time, timeDelta) {
         var systems = this.systems;
-
         // Animations.
         TWEEN.update(time);
         // Components.
@@ -465,15 +465,14 @@ module.exports = registerElement('a-scene', {
      * Renders with request animation frame.
      */
     render: {
-      value: function (time) {
+      value: function () {
         var effect = this.effect;
-        var timeDelta = time - this.time;
+        var delta = this.clock.getDelta() * 1000;
+        this.time = this.clock.elapsedTime * 1000;
 
-        if (this.isPlaying) { this.tick(time, timeDelta); }
-
+        if (this.isPlaying) { this.tick(this.time, delta); }
         this.animationFrameID = effect.requestAnimationFrame(this.render);
         effect.render(this.object3D, this.camera);
-        this.time = time;
       },
       writable: true
     }
--- a/tests/core/scene/a-scene.test.js
+++ b/tests/core/scene/a-scene.test.js
@@ -409,6 +409,20 @@ helpers.getSkipCISuite()('a-scene (with renderer)', function () {
     sinon.assert.called(Component.tick);
     sinon.assert.calledWith(Component.tick, scene.time);
   });
+
+  test('clock', function () {
+    var scene = this.el;
+
+    assert.isAbove(scene.time, 0);
+    var prevTime = scene.time;
+    assert.ok(scene.time, scene.clock.elapsedTime);
+    for (var i = 0; i < 10; i++) {
+      scene.render();
+      assert.isAbove(scene.time, prevTime);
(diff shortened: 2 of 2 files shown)


src/utils/index.js
```diff
@@ -7,6 +7,7 @@ var objectAssign = require('object-assign');
 
 var warn = debug('utils:warn');
 
+module.exports.Timer = require('./timer');
```
ngokevin: Could keep this alphabetical

src/core/scene/a-scene.js
```diff
@@ -465,15 +466,15 @@ module.exports = registerElement('a-scene', {
...
-        var timeDelta = time - this.time;
+        this.timer.update();
+        var time = this.timer.elapsedTime;
+        var timeDelta = this.timer.deltaTime;
 
         if (this.isPlaying) { this.tick(time, timeDelta); }
-
         this.animationFrameID = effect.requestAnimationFrame(this.render);
         effect.render(this.object3D, this.camera);
-        this.time = time;
```
ngokevin: I'd keep the time property on the scene.

src/core/scene/a-scene.js
```diff
@@ -465,15 +466,15 @@ module.exports = registerElement('a-scene', {
      * Renders with request animation frame.
      */
     render: {
-      value: function (time) {
+      value: function () {
         var effect = this.effect;
-        var timeDelta = time - this.time;
+        this.timer.update();
+        var time = this.timer.elapsedTime;
```
ngokevin: Don't need intermediate vars, only used once.

src/utils/timer.js
```diff
@@ -0,0 +1,22 @@
...
+  this.deltaTime = 0;
+}
+
+Object.assign(Timer.prototype, {
+  start: function () {
+    this.startTime = this.previousTime = window.performance.now();
+    this.running = true;
+  },
+
+  update: function () {
```
ngokevin: Unit tests

src/utils/index.js
fernandojsg: done, I didn't noticed O:)

src/core/scene/a-scene.js
fernandojsg: :+1:

src/core/scene/a-scene.js
fernandojsg: Umm do you want it even if it's accesible with `this.timer.ellapsedTime`?

src/utils/index.js
dmarcos: Can we use [THREE.Clock](https://threejs.org/docs/#Reference/Core/Clock)?

fernandojsg: @dmarcos this was my first try, but the current implementation it's quite ugly (https://github.com/mrdoob/three.js/blob/master/src/core/Clock.js) you need to call `getElapsedTime()` and `getDelta()` but you'll get different values as it will call `performance.now()` again. Or you call `getDelta()` and then access `elapsedTime` that doesn't seems very coherent.
Also it works with seconds instead of ms, so we need to multiply everything by 1000.
I would better stick for now with our implementation and probably open a PR in three.js but I guess it's widely used so it will break many things if we change the implementation a lot.

src/core/scene/a-scene.js
ngokevin: Yeah, it's documented, might be being used. And if browsers ever fix the behavior, it's possible we might revert?

fernandojsg: Yep, good point

ngokevin:
r+ from me

fernandojsg:
Test added, I'm using right now `assert.isAbove(scene.time, 0);` because of the "bug" we were talking on slack related to https://github.com/mrdoob/three.js/pull/10962.
I'll change that in the future as soon as it will get merged there and we'll upgrade too.

dmarcos:
Thanks!

dmarcos:
waiting for tests

PR #2474, opened 2017-03-09
 Allow "perspective" screenshots to have width/height properties apply

b-d:
**Description:**
This allows customizing the size of the "perspective" screenshots taken by specifying the width and height e.g. <a-scene screenshot="width: 1024; height: 1024">. Before this only applied to "equirectangular" screenshots, and "perspective" just took the current dimensions of the scene canvas. 

**Changes proposed:**
copy line 166 to line 151


ngokevin:
Thanks!

PR #2475, opened 2017-03-09
Fix isCoordinate by successfully matching numbers with e-notation

seanlerner:
When I rotate an entity it sometimes comes back with an e-notation number. For example, the number coming back was:

`-71.99999999999999 62.074139506678954 -1.5056867448590336e-22`

This number isn't recognized by isCoordinate and my entity would disappear.

I've modified the regex pattern to successfully match a number with e-notation.


ngokevin:
Thanks!

seanlerner:
Your welcome :)

PR #2479, opened 2017-03-10
changed isCoordinate to areCoordinates

### LINKED ISSUE #2450: Should isCoordinate be renamed areCoordinates?
I was confused for a moment when inspecting `AFRAME.utils.coordinates.isCoordinate` in the debugger. Seems it should be renamed (or aliased) to `areCoordinates`.

https://www.quora.com/Are-coordinates-plural-or-singular

dye784:
**Description:**
This PR fixes #2450. Just a minor change in the function name. The function isCoordinate should be plural. Modified to be areCoordinates.

**Changes proposed:**
isCoordinate was modified to be areCoordinates

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -14,12 +14,12 @@ A-Frame's utility modules are public through `AFRAME.utils`.
 
 Module for handling vec3 and vec4 types.
 
-### `.areCoordinates (value)`
+### `.isCoordinates (value)`
 
 Tests whether a string is a vec3.
 
 ```js
-AFRAME.utils.coordinates.areCoordinates('1 2 3')
+AFRAME.utils.coordinates.isCoordinates('1 2 3')
 // >> true
 ```
 
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -14,12 +14,12 @@ A-Frame's utility modules are public through `AFRAME.utils`.
 
 Module for handling vec3 and vec4 types.
 
-### `.isCoordinate (value)`
+### `.areCoordinates (value)`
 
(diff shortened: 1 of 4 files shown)


dmarcos:
In other parts of the code base we have checks `isXXX`. I prefer to keep function name consistency over grammatical correctness

dye784:
I see what you are saying. Would it not be better to be consistent with good grammar when naming functions though? I suspect although the other functions that are named `isXXX` are named grammatically correct. I feel like that would be more intuitive for other users that are using A-Frame when they are building their own apps to have function names to be grammatically correct. 😄

kellyselden:
@dmarcos I would argue that developer ergonomics should be considered as well. When I first saw `isCoordinate` on the `AFRAME` global, I had to second guess and look up the docs/implementation. I didn't know if it was testing a single number or a string of numbers. Eliminating that confusion would be worth it IMO.

dmarcos:
Name consistency is about developer ergonomics. Thinking about it is actually even gramatically correct. `isCoordinate` actually means the variable is of the type `coordinate`

kellyselden:
Can you link to the type definition? I can't find it in the source.

dmarcos:
I would just change `isCoordinate` to `isCoordinates` since the function is to determine if a the passed string contains coordinates

kellyselden:
@dmarcos The more I think about it you are right, `areCoordinates` sounds like it might take an array as a param or multiple params to test them all.

dye784:
@kellyselden Good point. It does sound like you would input more than one set of coordinates if you have `areCoordinates`.

dmarcos:
Thanks!

PR #2481, opened 2017-03-11
add .editorconfig

kellyselden:
mainly to have editors enforce the 2 space indents.


ngokevin:
I think we'll pass on this since most editors require a specific EditorConfig plugin for this to work, not quite agnostic enough. If there's a more general way to tell editors to follow the linting rules, we can check that out.

Issue #2031, opened 2016-10-31
Compatibility with node


impronunciable:
**Description:**

- A-Frame Version: 0.3.2
- Platform/Device: node.js

I wanted to make an aframe demo with react using [next.js](https://github.com/zeit/next.js) doing isomorphic rendering. Importing `aframe` won't work since node does not recognize the window global.

It would be great to adapt aframe to work with node so it can be used on isomorphic apps

ngokevin:
Workaround might be to do a super stub of the `window` with `global.window = { ... }`. 

aframe-react actually only needs `AFRAME.components`, so yeah, need to find a way to decouple for Node purposes.

ngokevin:
Solution might be to have a `browser.js` entry point and a separate `index.js` entry point which contains Node-friendly requires. The `browser.js` entry point can export what `index.js` does. Point `package.json` `main` to `index.js` and `package.json` `browser` field to whatever is in `dist/`.

ltfschoen:
@ngokevin @impronunciable I have created a repo [https://github.com/ltfschoen/react-fluxible-test](https://github.com/ltfschoen/react-fluxible-test) that shows how to incorporate A-Frame React into a Fluxible app (i.e. uses isomorphic architecture with server-side rendering)

kellyselden:
#2484 should get you almost all the way there.

ngokevin:
#2484

PR #2484, opened 2017-03-11
add node compatibility and regression test

kellyselden:
After #2476 and #2477 were merged, it is now possible to run aframe in node. This is a regression test to preserve this functionality as aframe grows, and to iterate upon to make it run better in node. This is a partial solution to #2031, but there will most likely need to be more work.

I tried to put the test in `tests/` first, but it made the globbing [here](https://github.com/aframevr/aframe/blob/3b9d8f419d45f3167d0053208cab6fb557dd8b22/tests/karma.conf.js#L24) difficult because `tests/!(node)/**/*.test.js` kept ignoring the `tests/__init.test.js`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ env:
   matrix:
     - CI_ACTION="npm run test:firefox -- --single-run"
     - CI_ACTION="npm run test:chrome -- --single-run"
+    - CI_ACTION="npm run test:node"
     - CI_ACTION="npm run lint"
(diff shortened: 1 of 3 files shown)


kellyselden:
@ngokevin Renamed to "/tests/node/test.js". I figured it would clearer to call it test to know what it is doing. I can still rename it "index.js" if you desire.

ngokevin:
Thanks!

PR #2486, opened 2017-03-11
use more npm pre scripts instead of chaining

kellyselden:


dmarcos:
@ngokevin what do you think of the `pre` scripts? I kind of prefer to see everything explicitly defined in the command. Those pre things are kind of magical and easy to miss when trying to understand how things work

ngokevin:
I agree, I'd rather see it explicitly because they're easy to miss. Thanks though!

PR #2490, opened 2017-03-13
Fix sound

fernandojsg:
**Description:** There're several problems with the current sound implementation and in nightly we found that it's a real nightmare to play them correctly, sometimes they play, sometimes you get an warn saying that you need to increase the pool, and sometimes you just don't know what happended but nothing is sounding.

**Changes proposed:**
- Added `positional (true/false)` parameter to define if you want the sound to be positional or not.
- Added `loaded` variable to define if the buffer is already loaded or not.
- Using the `loaded` variable the `playSound` detect if it's still not loaded and show the correct warning.
- Added `mustPlay` variable. This variable will be used in case that you called `playSound` but the buffer is still not loaded. As soon as the buffer will be loaded, it will play the sound.
- Disable `autoplay` for all the buffers, the `autoplay` property for the component will be handled by ourselves on loaded and calling the `playSound`. Otherwise it will play all the buffer just after loaded and then will try to call `playSound` again.


src/components/sound.js
```diff
@@ -57,16 +59,19 @@ module.exports.Component = registerComponent('sound', {
...
     }
-
     // All sound values set. Load in `src`.
     if (srcChanged) {
       var self = this;
       this.audioLoader.load(data.src, function (buffer) {
         self.pool.children.forEach(function (sound) {
           sound.setBuffer(buffer);
         });
+        self.loaded = true;
```
dmarcos: what if src changes? We have to set `this.loaded` to false before calling `audioLoader.load`?

src/components/sound.js
```diff
@@ -57,16 +59,19 @@ module.exports.Component = registerComponent('sound', {
...
+
         // Remove this key from cache, otherwise we can't play it again
         THREE.Cache.remove(data.src);
+
+        if (self.data.autoplay || self.mustPlay) { self.playSound(); }
```
dmarcos: What about `this.isSoundPlaying`

dmarcos: instead of `this.mustPlay` ?

dmarcos:
Can we add tests?

src/components/sound.js
fernandojsg: Yep you're right

src/components/sound.js
fernandojsg: For me `isSoundPlaying` feels like the sound is already playing not that we've asked the sound to start playing and it was delayed until the buffer is loaded

fernandojsg:
@dmarcos tests added!

dmarcos:
Can you remove the dist from the PR?

fernandojsg:
@dmarcos done!

PR #2492, opened 2017-03-15
don't load css in node environment

kellyselden:
Now that #2484 has landed, we can begin to iterate on the node experience and make it better. This removes the need to mock `require` to ignore css requiring in node (which is a bundler-only feature).

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -96,3 +96,17 @@ module.exports.isLandscape = function () {
 module.exports.isIOSOlderThan10 = function (userAgent) {
   return /(iphone|ipod|ipad).*os.(7|8|9)/i.test(userAgent);
 };
+
+/**
+ * Check if running in a browser or spoofed browser (bundler).
+ * We need to check a node api that isn't mocked on either side.
+ * `require` and `module.exports` are mocked in browser by bundlers.
+ * `window` is mocked in node.
+ * `process` is also mocked by browserify, but has custom properties.
+ */
+module.exports.isBrowserEnvironment = !!(!process || process.browser);
+
+/**
+ * Check if running in node on the server.
+ */
+module.exports.isNodeEnvironment = !module.exports.isBrowserEnvironment;
--- a/tests/node/test.js
+++ b/tests/node/test.js
@@ -4,23 +4,8 @@
 const path = require('path');
 const assert = require('chai').assert;
 const jsdom = require('jsdom');
-const Module = require('module');
-const sinon = require('sinon');
-
-const _require = Module.prototype.require;
 
 suite('node acceptance tests', function () {
-  let sandbox;
-
-  suiteSetup(function () {
-    sandbox = sinon.sandbox.create();
-    sandbox.stub(Module.prototype, 'require', function () {
-      if (arguments[0].indexOf('.css') === -1) {
-        return _require.apply(this, arguments);
-      }
-    });
-  });
-
   setup(function () {
     let _window = global.window = jsdom.jsdom().defaultView;
     global.navigator = _window.navigator;
@@ -43,13 +28,25 @@ suite('node acceptance tests', function () {
     delete global.WebVRConfig;
   });
 
-  suiteTeardown(function () {
-    sandbox.restore();
-  });
-
   test('can run in node', function () {
     let aframe = require(path.join(process.cwd(), 'src'));
 
     assert.ok(aframe.version);
+  });
+
+  suite('environment', function () {
+    let aframe;
+
+    setup(function () {
+      aframe = require(path.join(process.cwd(), 'src'));
+    });
+
+    test('isNodeEnvironment is true for node', function () {
+      assert.strictEqual(aframe.utils.device.isNodeEnvironment, true);
+    });
+
+    test('isBrowserEnvironment is false for node', function () {
+      assert.strictEqual(aframe.utils.device.isBrowserEnvironment, false);
+    });
   });
 });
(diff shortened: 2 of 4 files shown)


src/utils/device.js
```diff
@@ -96,3 +96,21 @@ module.exports.isLandscape = function () {
 module.exports.isIOSOlderThan10 = function (userAgent) {
   return /(iphone|ipod|ipad).*os.(7|8|9)/i.test(userAgent);
 };
+
+/**
+ * Check if running in a browser.
+ */
+function isBrowser () {
```
ngokevin: Does this need to be a function? I'd do maybe:

```module.exports.isBrowserEnvironment = !process;```

```module.exports.isNodeEnvironment = !!process;```

Comment (ngokevin) [inline src/utils/device.js]: We can add a unit test for these in Karma.

File: src/utils/device.js [hunk 2]
```diff
@@ -96,3 +96,21 @@ module.exports.isLandscape = function () {
...
+  // * `require` and `module.exports` are mocked in browser by bundlers
+  // * `window` is mocked in node
+  return !process;
+}
+module.exports.isBrowser = isBrowser;
+
+/**
+ * Check if running in node on the server.
+ */
+module.exports.isNode = function () {
```
ngokevin: We can also add a unit test for these in our Node tests

src/utils/device.js
kellyselden: I was matching the convention in the file `isIOS ()` and `isGearVR ()` are functions when I don't think they need to be either.

ngokevin: Yeah, I guess they don't need to be. Maybe later we can make those properties.

ngokevin:
These are always run though since we call them from index.js and a-scene.js. The disadvantage is that we are calling them multiple times.

kellyselden:
I changed to value exports instead of functions.

When adding tests. I found out that not only is `process` mocked by browserify/karma, all node globals are mocked (found while debugging chrome while karma was running). This meant I had to get a little more explicit with environment detection.

src/utils/device.js
```diff
@@ -96,3 +96,30 @@ module.exports.isLandscape = function () {
 module.exports.isIOSOlderThan10 = function (userAgent) {
   return /(iphone|ipod|ipad).*os.(7|8|9)/i.test(userAgent);
 };
+
+/**
+ * Check if running in a browser.
+ */
+// we need to check a node api that isn't mocked on either side
```
kellyselden: I'm not sure if I did the commenting right. I was trying to separate the public comments that go in the api docs, from private implementation comments.

ngokevin: I think you can just use normal comments throughout for these all properties. We don't actually generate docs from this, it's mostly all developer comments, but in JSDoc form. (with preference to capitalized / ending in period comments)

src/utils/device.js
```diff
@@ -96,3 +96,30 @@ module.exports.isLandscape = function () {
...
+ * Check if running the bundler.
+ */
+// `process` is also mocked by browserify, but has custom properties
+var isBundlerEnvironment = !!(process && process.browser);
+module.exports.isBundlerEnvironment = isBundlerEnvironment;
+
+/**
+ * Check if running in a browser or spoofed browser.
+ */
+var isBrowserLikeEnvironment = isBrowserEnvironment || isBundlerEnvironment;
```
ngokevin: Can we narrow it to two?

- `isBrowserEnvironment` - true if the environment is even browser-like.
- `isNodeEnvironment` - Not browser-like.

And then we can export straight up in two lines:

```
module.exports.isBrowserEnvironment = !process || !!(process && process.browser);

module.exports.isNodeEnvironment = !!module.exports.isBrowserEnvironment
```

tests/node/test.js
```diff
@@ -4,23 +4,8 @@
 const path = require('path');
 const expect = require('chai').expect;
 const jsdom = require('jsdom');
-const Module = require('module');
-const sinon = require('sinon');
-
-const _require = Module.prototype.require;
 
 describe('node acceptance tests', function () {
```
ngokevin: can we run mocha in tdd mode to match the other tests?

tests/utils/device.test.js
```diff
@@ -36,3 +36,21 @@ suite('isIOSOlderThan10', function () {
     assert.notOk(device.isIOSOlderThan10(v12));
   });
 });
+
+suite('environment', function () {
+  test('isNodeEnvironment is false for browser tests', function () {
+    assert.strictEqual(device.isNodeEnvironment, false);
```
ngokevin: could use `assert.ok` or `assert.notOk`

tests/node/test.js
kellyselden: Added in #2496 and will rebase.

tests/utils/device.test.js
kellyselden: I would like to test the correct type as well. I don't want to accidently return a truthy/falsy object (which I almost did: `process && process.browser` was returning `undefined` instead of `false`).

src/utils/device.js
kellyselden: Will do.

tests/utils/device.test.js
ngokevin: OK

ngokevin:
Can rebase this now

kellyselden:
rebased and updated.

ngokevin:
Thanks! Love the test code deletions.

Issue #2359, opened 2017-02-08
div.a-enter-vr is blocking cursor clicks and grabbing on desktop


clavis-magna:
div.a-enter-vr is extending across the width of the window and blocking cursor clicks / cursor grabbing events passing to the scene

- A-Frame Version: 0.4.0 (and also tested on master: 0.4.0 (Date 27-01-2017, Commit #dd9e073))
- Platform / Device: desktop mac/pc Chrome and firefox
- Reproducible Code Snippet or URL: reproducible on https://aframe.io/examples/

![div-a-enter-vr](https://cloud.githubusercontent.com/assets/2109125/22719174/49caa39c-edf8-11e6-9f72-4b7b21899e59.png)

clavis-magna:
overriding the style with an auto width has fixed it in the projects I am working on, but with limited testing across browsers thus far.

`<style>.a-enter-vr {width: auto !important}></style>`

dye784:
I'll take this. I think it could be fixed just by removing the `width` and `height` from `.a-enter-vr`.

PR #2493, opened 2017-03-15
removed height and width from a-enter-vr css class

dye784:
**Description:**
This PR fixes issue #2359. `div.a-enter-vr` extends across the window blocking cursor clicks / cursor grabbing events passing to the scene.

**Changes proposed:**
Changed css of `a-enter-vr`. Removed `height` and `width`. It seems like the `div` is only used for the button.


dye784:
Oops. Will fixed that. Thanks @ngokevin.

ngokevin:
Tested. Thanks!

PR #2496, opened 2017-03-16
use mocha tdd in node tests

kellyselden:
Requested [here](https://github.com/aframevr/aframe/pull/2492#discussion_r106457070)

### DIFF
--- a/tests/node/test.js
+++ b/tests/node/test.js
@@ -2,17 +2,17 @@
 'use strict';
 
 const path = require('path');
-const expect = require('chai').expect;
+const assert = require('chai').assert;
 const jsdom = require('jsdom');
 const Module = require('module');
 const sinon = require('sinon');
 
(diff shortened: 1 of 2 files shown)


tests/node/test.js
```diff
@@ -2,17 +2,17 @@
...
 const sinon = require('sinon');
 
 const _require = Module.prototype.require;
 
-describe('node acceptance tests', function () {
+suite('node acceptance tests', function () {
   let sandbox;
 
-  before(function () {
+  suiteSetup(function () {
```
ngokevin: Does `setup` by itself work like in the other tests?

kellyselden: Should be. I used [this](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/interfaces/tdd.js#L40-L43) as reference.

PR #2498, opened 2017-03-17
fix name isBrowserEnvironment

kellyselden:
I missed this in renaming #2496


ngokevin:
(this never happened)

Issue #2494, opened 2017-03-16
a-plane segments


Disorrder:
I can't understand is there segments settings on the `a-plane`?
[a-plane](https://aframe.io/docs/0.5.0/primitives/a-plane.html#attributes_segments_width) primitive doc
[plane](https://aframe.io/docs/0.5.0/components/geometry.html#plane) geometry doc
Well, Three.js [PlaneGeometry](https://threejs.org/docs/?q=plane#Reference/Geometries/PlaneGeometry) has `widthSegments` and `heightSegments` and them works as expected

- A-Frame Version: 0.5.0
- Platform / Device: Google Chrome
- Reproducible Code Snippet or URL: https://codepen.io/Disorrder/pen/WpZqEJ

**UPD**: yep, there is [no support](https://github.com/aframevr/aframe/blob/master/src/geometries/plane.js#L13) for this attrs. Please, add them :)

ngokevin:
Sounds good to me. Labeled if anyone wants to grab it.

PR #2499, opened 2017-03-17
fix plane geometry

Disorrder:
**Description:**
In order to fix issue #2494 
[THREE.PlaneGeometry](https://threejs.org/docs/?q=plane#Reference/Geometries/PlaneGeometry) docs

**Changes proposed:**
- add segments


ngokevin:
Thanks!

Issue #2412, opened 2017-02-17
processPropertyDefinition() should normalise the default value


delapuente:
If the type is explicitly given or inferred, the normalisation procedure of a schema property should also normalise its default value making it pass through the `parse` function. This would allow using the attribute string representation as the default value (see bug #2410)

- A-Frame Version: master
- Platform / Device: all
- Reproducible Code Snippet or URL:

```html
<script>
  AFRAME.registerComponent('test', {
    schema: { v: {type: 'vec2', default: '1 1'} }
  });
</script>
<a-scene>
  <a-entity test></a-entity>
</a-scene>
<script>
  document.addEventListener('DOMContentLoaded', () => {
    var e = document.querySelector('a-entity');
    var def = e.components.test.schema.v.default;
    var shouldBeTrue =
      JSON.stringify(def) === JSON.stringify(e.getAttribute('test').v);
    console.log('Are equal?', shouldBeTrue);
  });
</script>
```

dye784:
Cool. Will work on this one.

ngokevin:
We now do a warning https://github.com/aframevr/aframe/issues/2511 if the default value doesn't match the type rather than trying to force parse. Thanks Damon!

PR #2500, opened 2017-03-17
processPropertyDefinition now parses default string values

dye784:
**Description:**
This fixes issue #2412. The code(shown below) from the issue now returns true. If a type `vec2` or `vec3` is given, the normalization procedure should run `parse` and change it into an object with the correct coordinates.

```
<script>
  AFRAME.registerComponent('test', {
    schema: { v: {type: 'vec2', default: '1 1'} }
  });
</script>
<a-scene>
  <a-entity test></a-entity>
</a-scene>
<script>
  document.addEventListener('DOMContentLoaded', () => {
    var e = document.querySelector('a-entity');
    var def = e.components.test.schema.v.default;
    var shouldBeTrue =
      JSON.stringify(def) === JSON.stringify(e.getAttribute('test').v);
    console.log('Are equal?', shouldBeTrue);
  });
</script>
``` 

**Changes proposed:**

`processPropertyDefinition` now runs `parse` on the default string values such as `'1 1'` into `{x: 1, y: 1}` and `'2 2 2'` into `{x: 2, y: 2, z: 2 }`. The function `parse` was labeled `parseCoordinates` to be more clear and avoid conflicting names in the file. Not sure if that was necessary though.

An error is given if the string does not have 2 or 3 values separated by an empty space. For example, `type: 'vec2` and `default: 1 1 1` would run `warn('Default value for type vec2 can only be 2 coordinates')`.

### DIFF
--- a/src/core/schema.js
+++ b/src/core/schema.js
@@ -1,6 +1,7 @@
 var debug = require('../utils/debug');
 var propertyTypes = require('./propertyTypes').propertyTypes;
 var warn = debug('core:schema:warn');
+var parseCoordinates = require('../utils/coordinates').parse;
 
 /**
  * A schema is classified as a schema for a single property if:
@@ -75,6 +76,18 @@ function processPropertyDefinition (propDefinition) {
   // Fill in default value.
   if (!('default' in propDefinition)) {
     propDefinition.default = propType.default;
+  } else if (typeName === 'vec2' && typeof defaultVal === 'string') {
+    // Parse default string val into object
+    if (defaultVal.split(' ').length !== 2) {
+      warn('Default value for type vec2 can only be 2 coordinates');
+    }
+    propDefinition.default = parseCoordinates(defaultVal);
+  } else if (typeName === 'vec3' && typeof defaultVal === 'string') {
+    // Parse default string val into object
+    if (defaultVal.split(' ').length !== 3) {
+      warn('Default value for type vec3 can only be 3 coordinates');
+    }
+    propDefinition.default = parseCoordinates(defaultVal);
   }
 
   return propDefinition;
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -166,7 +166,7 @@ suite('Component', function () {
       });
     });
 
-    test.only('vec2 and vec3 default string values are converted to coordinate objects', function (done) {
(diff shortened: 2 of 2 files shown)


ngokevin:
Thanks! I didn't think about this before, but there's no way to tell if the parser should be run in general on a default value. We don't know if the default value supplied is pre-parsed or parsed. The PR works for `vec2`/`vec3`, but that's a bit too ad-hoc.

Perhaps the proper solution is to enforce the default values be already parsed, and not allow, for example, strings to be passed as default values for `vec3`s?

dye784:
Yeah it was definitely hacky. I agree, probably the better solution is to have the default values be parsed correctly from the beginning. So would it be better if I made a utility function that checks if the `defaultVal` contains the correct values and then run it in `processPropertDefinition` if `typeName === 'vec2'` or if `typeName === 'vec3'`?

ngokevin:
It's guaranteed to affect more than just `vec2` and `vec3`. I think proper solution is to enforce post-parsed default values in schemas.

dmarcos:
We can have a `validateSchema` method that performs several checks when a component is registered and throws some warnings / errors in console if things do comply like having 
````javascript
v: {type: 'vec2', default: 'ohYeah'}
````

**Comment by dye784** [comment]
Yeah. What are the different types and their default values? Would have to have to perform checks on all of them.

**Comment by ngokevin** [comment]
Sorry, after reading the original issue more, I think it's OK to normalize some default values where we can. It's not often we access `schema.propertyName.default` manually anyways.

File: src/core/schema.js [hunk 1]
```diff
@@ -75,6 +76,18 @@ function processPropertyDefinition (propDefinition) {
   // Fill in default value.
   if (!('default' in propDefinition)) {
     propDefinition.default = propType.default;
+  } else if (typeName === 'vec2' && typeof defaultVal === 'string') {
```
ngokevin: Can we move all of these checks into its own function?

dye784: Yeah I'm going to make a function that does this. I'm make it so that if the type is substantiated the default value has to be the same as well. For example if `type` is `number` `default` should also be a `number`. Sound good?

ngokevin:
Closing in favor of https://github.com/aframevr/aframe/pull/2511

PR #2502, opened 2017-03-18
Fix deepEqual when comparing non Object instantiated objects

dmarcos:
**Description:**

We end up in an infinite loop when a property of the type `selector` changes value. Two HTML elements are passed through the `deepEqual` function that runs ad infinitum since each entity references the components that reference the entity itself, that reference the components, that reference the entity...

**Changes proposed:**
Deep equal only compares deeply if the passed objects are plain JS objects.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -125,10 +125,13 @@ function deepEqual (a, b) {
   var keysB;
   var valA;
   var valB;
+  var areBothObjects = a && b && a.constructor === Object && b.constructor === Object;
+  var areBothArrays = a && b && a.constructor === Array && b.constructor === Array;
 
-  // If not objects, compare as values.
-  if (typeof a !== 'object' || typeof b !== 'object' ||
-      a === null || b === null) { return a === b; }
+  // If not objects or arrays, compare as values.
+  if (!(areBothObjects || areBothArrays) || a === null || b === null) {
+    return a === b;
+  }
 
   // Different number of keys, not equal.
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -125,10 +125,13 @@ function deepEqual (a, b) {
   var keysB;
   var valA;
   var valB;
+  var areBothObjects = a && b && a.constructor === Object && b.constructor === Object;
```
ngokevin: Should we evaluate these inline so that the if expression can lazy evaluate and possibly not have to do `a && b` check twice?

tests/utils/objects.test.js
```diff
@@ -1,5 +1,7 @@
 /* global assert, suite, test */
+// var entityFactory = require('../helpers').entityFactory;
```
ngokevin: Can keep here

tests/utils/objects.test.js
```diff
@@ -127,5 +129,15 @@ suite('utils.objects', function () {
       var objA = {x: 0, y: 0, z: 0, self: objA};
       assert.ok(deepEqual(objA, objA));
     });
+
+    test('avoid deep equal of object that are not instantiated' +
+         'with the Object constructor in order to avoid infinite loops', function (done) {
+      var el1 = entityFactory();
```
ngokevin: Can we just do `deepEqual(document.createElement('a-entity'), document.createElement('a-entity'))`

PR #2503, opened 2017-03-18
Fixed 404'ing image links

joeyfb:
Thumbnail images we're not available as PNGs, only as JPGs.

**Description:**
Fixing broken images for users running documentation.

**Changes proposed:**
- Changing thumbnail image links to .jpg


dmarcos:
Thanks!

PR #2505, opened 2017-03-20
Emit event when oculus or vive controllers are connected

fernandojsg:
I didn't include any extra information as it can be accessed directly by the target

### DIFF
--- a/examples/showcase/tracked-controls/index.html
+++ b/examples/showcase/tracked-controls/index.html
@@ -22,8 +22,8 @@
                  material="color: #F2E646;"></a-mixin>
       </a-assets>
       <!-- Hands -->
-      <a-entity id="left" oculus-touch-controls="hand: left" vive-controls="hand: left" aabb-collider="objects: .cube;" grab></a-entity>
-      <a-entity id="right" oculus-touch-controls="hand: right" vive-controls="hand: right" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="left" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="right" aabb-collider="objects: .cube;" grab></a-entity>
 
       <!-- A-Frame cubes -->
       <a-entity position="0 0 -0.5">
@@ -57,13 +57,5 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
     </a-scene>
-    <script>
-      document.getElementById('left').addEventListener('controllerconnected', function(evt){
-        console.log('left', evt);
-      })
-      document.getElementById('right').addEventListener('controllerconnected', function(evt){
-        console.log('right',evt);
-      });
-    </script>
   </body>
 </html>
--- a/examples/showcase/tracked-controls/index.html
+++ b/examples/showcase/tracked-controls/index.html
@@ -22,8 +22,8 @@
                  material="color: #F2E646;"></a-mixin>
       </a-assets>
       <!-- Hands -->
-      <a-entity hand-controls="left" aabb-collider="objects: .cube;" grab></a-entity>
-      <a-entity hand-controls="right" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity id="left" oculus-touch-controls="hand: left" vive-controls="hand: left" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity id="right" oculus-touch-controls="hand: right" vive-controls="hand: right" aabb-collider="objects: .cube;" grab></a-entity>
 
       <!-- A-Frame cubes -->
       <a-entity position="0 0 -0.5">
@@ -57,5 +57,13 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
     </a-scene>
+    <script>
+      document.getElementById('left').addEventListener('controllerconnected', function(evt){
+        console.log('left', evt);
+      })
+      document.getElementById('right').addEventListener('controllerconnected', function(evt){
+        console.log('right',evt);
+      });
+    </script>
   </body>
 </html>
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -109,7 +109,6 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   pause: function () {
-    this.removeControllersUpdateListener();
     this.removeEventListeners();
     this.removeControllersUpdateListener();
     // Note that due to gamepadconnected event propagation issues, we don't rely on events.
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -100,7 +100,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     var isPresent;
 
     // Find which controller matches both prefix and hand.
-    isPresent = this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, {
+    isPresent = this.isControllerPresent(this, GAMEPAD_ID_PREFIX, {
       hand: data.hand
     });
 
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -70,7 +70,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.bindMethods();
 
     this.emitIfAxesChanged = controllerUtils.emitIfAxesChanged;   // Allow mock.
-    this.isControllerPresent = controllerUtils.isControllerPresent;  // Allow mock.
+    this.checkControllerPresentAndSetup = controllerUtils.checkControllerPresentAndSetup;  // Allow mock.
   },
 
   addEventListeners: function () {
@@ -96,25 +96,9 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   checkIfControllerPresent: function () {
-    var data = this.data;
-    var isPresent;
-
-    // Find which controller matches both prefix and hand.
-    isPresent = this.isControllerPresent(this, GAMEPAD_ID_PREFIX, {
-      hand: data.hand
+    this.checkControllerPresentAndSetup(this, GAMEPAD_ID_PREFIX, {
+      hand: this.data.hand
     });
-
-    // Nothing changed, no need to do anything.
-    if (isPresent === this.controllerPresent) { return; }
-
-    // Update controller presence.
-    this.controllerPresent = isPresent;
-    if (isPresent) {
-      this.injectTrackedControls();
-      this.addEventListeners();
-    } else {
-      this.removeEventListeners();
-    }
   },
 
   play: function () {
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -27,16 +27,19 @@ module.exports.getGamepadsByPrefix = function (idPrefix) {
 /**
  * Enumerate controllers (as built by system tick, e.g. that have pose) and check if they match parameters.
  *
- * @param {object} sceneEl - the scene element.
+ * @param {object} component - the tracked controls component.
  * @param {object} idPrefix - prefix to match in gamepad id, if any.
  * @param {object} queryObject - map of values to match (hand; index among controllers with idPrefix)
  */
-module.exports.isControllerPresent = function (sceneEl, idPrefix, queryObject) {
+module.exports.isControllerPresent = function (component, idPrefix, queryObject) {
   var isPresent = false;
   var index = 0;
   var gamepad;
   var isPrefixMatch;
   var gamepads;
+  var el = component.el;
+  var sceneEl = el.sceneEl;
+
   var trackedControlsSystem = sceneEl && sceneEl.systems['tracked-controls'];
   if (!trackedControlsSystem) { return isPresent; }
   gamepads = trackedControlsSystem.controllers;
@@ -58,6 +61,10 @@ module.exports.isControllerPresent = function (sceneEl, idPrefix, queryObject) {
     }
     if (isPresent) { break; }
     if (isPrefixMatch) { index++; } // update count of gamepads with idPrefix
+  }
+
+  if (isPresent) {
+    el.emit('controllerconnected', {name: component.name, component: component});
   }
   return isPresent;
 };
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -25,20 +25,47 @@ module.exports.getGamepadsByPrefix = function (idPrefix) {
 };
 
 /**
+ * Check if the controller match the parameters and inject the tracked-controls component
+ * and add eventlistener, otherwise it will just remove the listener.
+ * It will also generate a controllerconnected or controllerdisconnected.
+ *
+ * @param {object} component - the tracked controls component.
+ * @param {object} idPrefix - prefix to match in gamepad id, if any.
+ * @param {object} queryObject - map of values to match (hand; index among controllers with idPrefix)
+ */
+module.exports.checkControllerPresentAndSetup = function (component, idPrefix, queryObject) {
+  var el = component.el;
+  var isPresent = isControllerPresent(component, idPrefix, queryObject);
+
+  // Nothing changed, no need to do anything.
+  if (isPresent === component.controllerPresent) { return isPresent; }
+  component.controllerPresent = isPresent;
+
+  // Update controller presence.
+  if (isPresent) {
+    component.injectTrackedControls();
+    component.addEventListeners();
+    el.emit('controllerconnected', {name: component.name, component: component});
+  } else {
+    component.removeEventListeners();
+    el.emit('controllerdisconnected', {name: component.name, component: component});
+  }
+};
+
+/**
  * Enumerate controllers (as built by system tick, e.g. that have pose) and check if they match parameters.
  *
  * @param {object} component - the tracked controls component.
  * @param {object} idPrefix - prefix to match in gamepad id, if any.
  * @param {object} queryObject - map of values to match (hand; index among controllers with idPrefix)
  */
-module.exports.isControllerPresent = function (component, idPrefix, queryObject) {
+function isControllerPresent (component, idPrefix, queryObject) {
   var isPresent = false;
   var index = 0;
   var gamepad;
   var isPrefixMatch;
   var gamepads;
-  var el = component.el;
-  var sceneEl = el.sceneEl;
+  var sceneEl = component.el.sceneEl;
 
   var trackedControlsSystem = sceneEl && sceneEl.systems['tracked-controls'];
   if (!trackedControlsSystem) { return isPresent; }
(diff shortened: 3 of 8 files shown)


src/components/oculus-touch-controls.js
```diff
@@ -150,6 +150,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       controller: 0,
       rotationOffset: data.rotationOffset !== -999 ? data.rotationOffset : isRightHand ? -90 : 90
     });
+    el.emit('controller-connected', {type: 'oculus-touch-controller'});
```
dmarcos: no dashes `controllerconnected` is the convention

machenmusik:
(1) shouldn't this actually be done in the tracked-controls system that finds new controllers?
(2) shouldn't the type be the gamepad ID, if not the gamepad state itself?  I am not entirely understanding what this provides in addition to ongamepadconnected events, maybe it is simply the abstraction?

machenmusik:
would also like to understand the intended behavior in light of #2423 and #2446 e.g. what should be done if anything for emulated situations?

donmccurdy:
+1 for using gamepad ID rather than a new string. And if a new string is needed, we should be consistent on `-controls` or `-controller` here.

machenmusik:
if the general idea is to have the components re-emit their own version of `gamepadconnected` when they are going to inject, don't we also want something for `gamepaddisconnected`, and make it independent of 3DoF vs. 6DoF for example daydream controller and upcoming gearvr controller?  (maybe even gearvr touchpad to the extent folks continue to use that with Carmel?)

fernandojsg:
I needed this feature to detect when a gamepad is connected and determine (without need to redo most of the work done already by the core components) if it's an oculus or a vive controller. 
In my particular case I needed to rotate the geometry of the gun to fit the controllers orientation differences. So I needed an event as the model could be loaded before the gamepad was already loaded.
I used `-controls` and `-controllers` to match the current components `vive-controls` and `oculus-touch-controllers`. I agree that we should define a consistent name for both, so probably we should start changin the name of these components? And even using singular instead of plural in their names?
I decided to send just the minimum amount of info because I could find the rest of info by looking at the target's componet, but f you feel that the ID could be useful too for sure we could include.
I'm open to any other proposal that could works for the same purposes.

machenmusik:
@fernandojsg I see, so by `type` you really meant source component i.e. `oculus-touch-controls` or `vive-controls` which is not immediately obvious from the injected `tracked-controls`-- if we go back to the original decision to have all the different types of possible components attached, I think the pattern you are looking for is actually the `auto-detect-controllers` approach and then the particular injected component would be available through the standard eventing for new component addition, rather than only ambiguous `tracked-controls`.  Given the current decision there, it seems as though the injected `tracked-controls` should have some indication of source component, so you can tell that the parent of the injected `tracked-controls` is vive not oculus.  Does that sound like what you are looking for?

machenmusik:
> standard eventing for new component addition

Looks like there may be stale documentation
https://aframe.io/docs/0.5.0/core/entity.html#events_componentinit
https://aframe.io/docs/0.5.0/core/entity.html#events2_componentinitialized

If the name & id of the source/parent component doing the injection were available as part of as well as injected name/id/data, then this pattern would be generalizable to listening for and filtering `componentinitialized` which would be a nice solution.  Thoughts?

src/components/oculus-touch-controls.js
dmarcos: let's move these events to `tracked-controls`

machenmusik: won't work as there is no context provided to distinguish which component injected the tracked-controls.  if that is provided to `tracked-controls` when injected, then I agree e.g. by expanding schema with something like `source` that is name of injecting component `oculus-touch-controls`, `vive-controls`, `daydream-controller` (which maybe should have been -controls), `gearvr-controls` and then providing `source` in the `controllerconnected` data

machenmusik: if we were to expand the `tracked-controls` schema, would the data be available when `componentinitialized` is fired for `tracked-controls` ?  or is it only available in subsequent `update` ?

dmarcos: we can add the `el` reference to the details of the event

machenmusik: the case @fernandojsg is same element with both oculus and vive components, need to know which injected; really the component reference, not the element

dmarcos: this is why I asked about moving the event to `tracked-controls`. `tracked-controls` will pass the Gamepad ID and `el` as details of the event. @fernandojsg would that work for you?

machenmusik: I think the idea was to avoid the ambiguity of having to know correspondence from gamepad.id to controller component.  Another approach is to listen for `componentinitialized` of `tracked-controls`, and look at `.controllerPresent` on the controller components -- but then not sure how to handle emulation (separate discussion)

machenmusik:
Maybe no changes are required?
Have a look at `which-controller-present` in this CodePen...
http://codepen.io/machenmusik/pen/qroKWN

dmarcos:
@fernandojsg would the example that @machenmusik points out in the codepen work for you?

fernandojsg:
the @machenmusik proposal probably could works in most of the cases but I feel we're adding some extra steps that are not really necessary: adding a new component `which-controller-present` (should it be included in the core, or should we link from the aframe registry?), include that component in all the `tracked-controls` entities you want to check.
I feel it's the same if we'll create a `is-model-loaded` component and will emit a event when a `json-model` is loaded, and we should include it on every model we've loaded.
Maybe users could have also strange behaviour if this component is initialized after the tracked-controls is ready so it won't catch the events.
Probably a mix between the two approach could be to include that code inside the `tracked-controls` to remove my proposal code in each controls and emit these events directly from the tracked-controls.

fernandojsg:
I've updated my PR with the component code of @machenmusik included directly on the `tracked-controls` . So we don't need that extra component and it's cleaner also as we don't need to remember to include that component. What do you think?

src/components/tracked-controls.js
```diff
@@ -35,6 +36,27 @@ module.exports.Component = registerComponent('tracked-controls', {
...
+  },
+
+  onComponentInitialized: function (evt) {
+    if (evt.detail.name !== 'tracked-controls') { return; }
+    var self = this;
+    Object.keys(this.el.components).forEach(function (name) {
+      const component = self.el.components[name];
+      if (component.controllerPresent) {
+        // Emit controllerpresent event.
+        self.el.emit('controllerpresent', {name: name, component: component});
```
dmarcos: The problem with this is that if the controller is not there when the component is initialized you won't be notified

fernandojsg: I believe this is the case when you include directly `tracked-controls` instead of `vive-controls` or `oculus-controls` the specific controls inject tracked controls when the controllers is already available, but if you just inject `tracked-controls` by yourself you won't get 100% of the cases.

dmarcos: A controller could become available after the component / application is initialized

fernandojsg: Yep, so what do to? Going back to the original approach to inject it when the controller is being injected? https://github.com/aframevr/aframe/blob/master/src/components/vive-controls.js#L94

dmarcos: `vive-controls` and `oculus-touch-controls` listen to `gamepadconnected`.  I think everything should happen through `tracked-controls`. `oculus-touch-controls` and `vive-controls` should set `tracked-controls` on init and wait for an event when the controller becomes available.

dmarcos: Listening to `gamepadconnected` breaks the abstraction that `tracked-controls` is supposed to offer

machenmusik: So the general problem is this:

Assume some tracked controllers would be available through Gamepad API.
But they are not available until actually used, due to security restrictions.

Current `oculus-touch-controls` and `vive-controls` usage is top-down declarative (declare the components to assert they are allowed to be present, and components check whether really present) so `tracked-controls` is only injected upon detection by appropriate component; unless they want to be constantly checking controller list for new entries all the time, they would need to listen to new-controller event which is `gamepadconnected`.

During original discussions, the core team preferred top-down declarative to bottom-up (e.g. `auto-detect-controllers` at the time due to minimal point-in-time concerns over scalability of pattern, and overall fit with other A-Frame component usage.

If we want to preserve top-down declarative, but attempt to abstract away completely from Gamepad API with `tracked-controls`, then the `tracked-controls` system needs to be active regardless of whether any component instances exist, and needs to wrap or transform any data the Gamepad API would provide, including connected/disconnected events.

machenmusik: tl;dr - maybe we need `controllerconnected` and `controllerdisconnected` to come from `tracked-controls` system, but then you need to have a `tracked-controls` instance to emit/receive it before you know there are any to be used!

dmarcos: yeah that's a start. we can listen for a `controllerconnected` event on the `sceneEl` emitted by the `tracked-controls` system instead of the window's `gamepadconnected`. Then I wonder if there's a way for the system to automatically set the `tracked-controls` component on the entity when the controller becomes available. `vive-controls` and `oculus-touch-controls` would register to that mechanism

dmarcos: Something like `sceneEl.systems['tracked-controls'].subscribeToController(el, data, callback);`

machenmusik: so you are 90% of the way to `auto-detect-controllers` approach now :-)
where `auto-detect-controllers` was the component stand-in that would listen for new controller available, and inject whatever was appropriate.

A little strange to have `tracked-controls` bottoms-up and `oculus-touch-controls`, `vive-controls` etc. top-down though, rather than having `auto-detect-controllers` that would listen for stage events and inject the higher-level components (oculus, vive) not just the lower-level (tracked-controls)

can we agree on names for the scene events
(closely matching wrapped Gamepad API events)
`controllerconnected`
`controllerdisconnected`
?

I'd suggest that we maintain `controllerpresent` as the name of the event for the higher-level component that actually picks up.

machenmusik: oh, you don't want to expose `controllerconnected` and `controllerdisconnected` more generally, but keep that pseudo-private to `tracked-controls` system?

machenmusik: >  Then I wonder if there's a way for the system to automatically set the tracked-controls component on the entity when the controller becomes available.
You won't know which entity, unless you have a component attached to it.  (that is one job `auto-detect-controllers` was originally doing)

dmarcos: You know the entity because you pass it when you suscbribe to it. Look at the signature:
`sceneEl.systems['tracked-controls'].subscribeToController(el, data, callback);`

dmarcos: controllerconnected / controllerdisconnected sounds good to me

machenmusik: tl;dr - we seem agreed to wrap `gamepadconnected`/`gamepaddisconnected` within tracked-controls system.  it is not clear to me whether the component to attach to entities that want to represent the controllers is agreed to still be `oculus-touch-controls` and `vive-controls` top-down declarative (and in response they do their detection and injection of `tracked-controls` the same as it is done now) -- but using Occam's razor, for now that seems simplest

dmarcos: I probably leave it to `vive-controls` `oculus-touch-controls` for the moment. Not sure if adding `tracked-controls` automatically would be to much magic.

dmarcos: I imagine an entity wanting to subscribe to the controller but not use a `tracked-controls` component

machenmusik: > You know the entity because you pass it when you suscbribe to it.

if you have something like that line of code in `oculus-touch-controls` and `vive-controls` :
`sceneEl.systems['tracked-controls'].subscribeToController(el, data, callback);`
what are you thinking goes in data?

I am thinking it may be better to provide id and gamepad/controller reference in `controllerconnected` stage events and let the listener decide whether to inject whatever component(s) like `tracked-controls` rather than trying to come up with a data schema with matching filter code

dmarcos: We can sprinkle the magic later if we see fit.

machenmusik: So generally, if you want to do something with a newly connected controller,
`this.el.sceneEl.addEventListener('controllerconnected', this.onControllerConnected.bind(this))`

But this begs the question of whether we want to do that for any gamepads, not just tracked controllers.

dmarcos: we can then have `trackedcontrollerconnected` \ `trackedcontrollerdisconnected`. We can have a separate `gamepad` component?

machenmusik: well, tracked controller is really a controller with pose, so IMO we should stick with `controllerconnected` and `controllerdisconnected` (that would also fire for gamepads) -- oculus, vive, daydream can do their usual checking for ID prefix and pose... other (future) gamepad controls could attach to same event mechanism

machenmusik: to @fernandojsg 's original question though... which component should actually fire `controllerpresent` (which is when a higher-level component recognizes the actual presence of its controller, not just when a controller connected event is seen) -- the higher-level component after it injects (original PR), or by `tracked-controls` as per the latest here?

dmarcos: that's the goal of `sceneEl.systems['tracked-controls'].subscribeToController(el, data, callback);` you subscribe to the particular controller you're looking for so you don't have to manually check from a generic event

machenmusik: so the a-blast use case is kinda like `hand-controls` usage -- I think it's called `shoot-controls` -- but it needs to specifically know which higher-level controller component (`oculus-touch-controls` and `vive-controls`) to adjust model pose, etc.

I don't follow how `oculus-touch-controls` and `vive-controls` doing the subscription helps `shoot-controls`, and that subscription wouldn't help `shoot-controls` since it only knows `oculus-touch-controls` and `vive-controls`, it doesn't know the details of how they recognize what controller/gamepad.

machenmusik:
> I've updated my PR with the component code of @machenmusik included directly on the tracked-controls . So we don't need that extra component and it's cleaner also as we don't need to remember to include that component. What do you think?

So where did we end up with this one after all?  Looking at it again, I think @fernandojsg proposal would work, and if other third-party components obey the `controllerPresent` convention, they could participate as well.

> I believe this is the case when you include directly tracked-controls instead of vive-controls or oculus-controls the specific controls inject tracked controls when the controllers is already available, but if you just inject tracked-controls by yourself you won't get 100% of the cases.

True but if you inject `tracked-controls` yourself (rather than letting `oculus-touch-controls`, `vive-controls, `daydream-controls` etc. do their detection) then you should know what you are doing and not need the signal event to tell you.

> I imagine an entity wanting to subscribe to the controller but not use a tracked-controls component

For `oculus-touch-controls`, `vive-controls, `daydream-controls` this is not possible as they rely upon `tracked-controls` for button / axis / pose update (and recording/replay takes full advantage of that fact).

dmarcos:
Let's make `tracked-controls` properties `id`, `prefixId` and `controller` multi valuated to avoid collisions so the controller detection moves down from `oculus-touch-controls`, `vive-controls` to `tracked-controls`. We can then emit an event in `updateGamepad` when one of the controllers we're looking for is detected (https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls.js#L77)

machenmusik:
I'm not sure how that actually helps the original use case @fernandojsg was trying to solve (need to know whether oculus-touch-controls or vive-controls is actually connected) -- the upper level enjoys the abstraction of not having to know the raw gamepad id strings, but only the higher-level component abstractions,..

dmarcos:
I was trying to solve duplicated code and emitting the event in want sweep. There's an easy way to emit the higher level event by passing the component to `isControllPresent` mixining in the function on the component (https://github.com/aframevr/aframe/blob/master/src/utils/tracked-controls.js#L34)

machenmusik:
my recollection is that some components need to do their own filtering using getGamepadsByPrefix
... and also we only want the event to fire when things change, if we fire when isControllerPresent is true it will spam multiple times every tick.

machenmusik:
thinking about problems separately,

to solve "how do i know which component injected this tracked-controls" just add a string to schema that higher level components fill in with their name when injecting.  then you can read it even if you missed the event.

to fire an event when you see a new controller would ideally be gamepadconnected/disconnected from browser, but in practice i have not seen that work correctly so systems component would probably have to fire - perhaps on scene as we were discussing previously.  

but putting the two together is slightly trickier as for fernando's use case, you only want the event AFTER you can read which component injected.  So maybe fire event from tracked-controls schema update on that property.

fernandojsg:
My main concern with moving the logic from `vive/oculus-controls` to `tracked-controls` is that it could grow up a lot as more controllers appears isn't it? Or maybe we could refactor after a while with something like `registerController` to specify the name and how it should detect it :)

So to follow with the PR with @dmarcos proposal, we should move `checkIfControllerPresent` and `injectTrackedControls` from the  `vive/oculus-controls` to `tracked-controls` right? and the rest code should remain on each component.

fernandojsg:
And so the `trackedcontrollerconnected` event will be emitted in the `injectTrackedControls` on  `tracked-controls` similar as I did on the same function but on each component.

dmarcos:
Let's do something easier. The `isControllPresent` signature will receive the component and it will manage the `controllerPresent` flag. The event wil be emitted only when the controller shows up passing the component name in the detail. We have to make `oculus-touch-controls` use `isControllerPresent` again

machenmusik:
the less invasive change makes sense to me, bigger refactor if any can happen post-0.6.0

fernandojsg:
I've modified the `isControllerPresent` to accept a `component` parameter instead a `sceneEl`. And this function will emit the event if present.
I've modified (just while we're discussing this PR, nothing definitely) the trackedcontrols example to add a listener for these events on each hand and log them. But it just keep printing forever.
Time to sleep here, but if someone have any idea, it's more than welcome ;)

machenmusik:
I think the event needs to fire only when this.controllerPresent changes state, not just when isControllerPresent would return true? and really the event should probably only fire after tracked-controls has been injected, so that anything from tracked-controls may be accessible?

machenmusik:
@dmarcos had suggested that controllerpresent handling be moved into tracked-controls (utils?) - maybe checkIfControllerPresent should move to utils in order to accomplish, and take a callback function to fire if changed (after which the event could be generated) where that callback may differ across higher-level components

fernandojsg:
I've refactored the code and moved all the inject and add/remove eventlistener to a new function in `utils/tracked-controls` `checkControllerPresentAndSetup` (Any suggestion for a better name maybe?) and it's working as expected.

The test are still not passing, as in `checkControllerPresentAndSetup` I'm calling `isControllerPresent` that we were mocked previously as they were included on the component itself, but now I'm calling it directly from the utils module and I can't think on an easy way to mock it so we could return true or false so it could inject correctly the listener and the component. Any idea?

machenmusik:
@fernandojsg if you look at the vive-tracker PR, you can see a change to the way tests are mocked -- just change the system controllers list to what you need it to look like

fernandojsg:
@machenmusik thanks! I've used your implementation and it's working as expected 👍

examples/showcase/tracked-controls/index.html
```diff
@@ -57,5 +57,13 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
     </a-scene>
+    <script>
```
dmarcos: We can get rid of these logs

src/utils/tracked-controls.js
```diff
@@ -25,42 +25,76 @@ module.exports.getGamepadsByPrefix = function (idPrefix) {
...
-  for (var i = 0; i < gamepads.length; ++i) {
-    gamepad = gamepads[i];
-    isPrefixMatch = (!idPrefix || idPrefix === '' || gamepad.id.indexOf(idPrefix) === 0);
-    isPresent = isPrefixMatch;
-    if (isPresent && queryObject.hand) {
-      isPresent = (gamepad.hand || DEFAULT_HANDEDNESS) === queryObject.hand;
-    }
-    if (isPresent && queryObject.index) {
-      isPresent = index === queryObject.index; // need to use count of gamepads with idPrefix
+  if (gamepads) {
```
dmarcos: We can return early and save one level of indentation

fernandojsg: Done

dmarcos:
Thank you! That's x2 bonus points for a PR that removes more code than adds.

fernandojsg:
I've a question, Do we need to call `addControllersUpdateListener` on `play` even if the controller is not present?

machenmusik:
i believe that due to issues with gsmepadconnected/disconnected propagation, that is still needed to reliably detect controllers, but would be happy to be proven wrong

fernandojsg:
@machenmusik ok thanks, I've just keep them by now to keep the consistency between all the controllers components, we could address that later if needed.

@dmarcos fixed the issue with the `remove/add listeners` on `play/pause`

dmarcos:
Thanks!

fernandojsg:
@machenmusik I used false as I saw that we're initializing `controllerPresent = false` in all the `init` for each controller's component and that's why I used that value as default.

PR #2510, opened 2017-03-21
also apply userHeight to controller when no stage parameters (fixes #2448)

### LINKED ISSUE #2448: Camera Y position incorrect in some cases while in VR
mchen [12:59 PM] 
interesting default camera Y height behavior with current master...
http://codepen.io/machenmusik/pen/JWXGap

With Rift and latest Chromium, when using the auto-enter/exit VR, user height is correct both outside VR (default user height 1.6) and in VR (true HMD height), first time and every time.  That consistency is provided by #2428.

With Rift and latest Chromium, when manually clicking the enter VR button but not wearing the HMD, user height is correct outside VR (default user height 1.6) but wrong inside VR (e,g, camera placed higher than true HMD height once you put it on).

With Rift and latest Nightly, it appears that vrdisplayactivate is not working and therefore neither is auto-enter VR (i.e. you put on the Rift, see nothing, and still have to click the button).

With Rift and latest Nightly, when manually clicking the enter VR button but not wearing the HMD, user height is always 1.6 since it just goes fullscreen and doesn't try to enter VR at all.  In fact, it won't try to enter VR at all unless you reload the page.

With Rift and latest Nightly, when manually clicking the enter VR button while wearing the HMD, user height is correct outside VR (default user height 1.6) but wrong inside VR (e,g, camera placed lower than true HMD height once you put it on, apparently by 1.6 the default user height), first time and every time.  That consistency is provided by #2428.

So two apparent bugs:
(1) Nightly vrdisplayactivate is not working.
(2) Manually entering VR via UI button seems to have some issue with getting initial HMD position/height, and since applied as deltas by `look-controls`, that stays wrong while in VR.

machenmusik:
to fix Nightly issue with null stageParameters,
add camera userHeight if null stageParameters;
to keep tests working, only getVRDisplays when VR entered.

(NOTE: tried to move getVRDisplays from system into component where dolly is,
but got weird effect where right hand was applying offset but right hand wasn't,
so there may still be some strange race condition somewhere)

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -90,8 +90,7 @@ module.exports.Component = registerComponent('tracked-controls', {
           dolly.applyMatrix(standingMatrix);
         } else {
           // Apply default camera height
-          var userHeight = this.el.sceneEl.camera.el.components.camera.data.userHeight;
-          dolly.position.y += userHeight;
+          dolly.position.y += el.sceneEl.camera.el.getAttribute('camera').userHeight;
           dolly.updateMatrix();
         }
       }
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -84,9 +84,16 @@ module.exports.Component = registerComponent('tracked-controls', {
       dolly.updateMatrix();
 
       // Apply transforms.
-      if (vrDisplay && vrDisplay.stageParameters) {
-        standingMatrix.fromArray(vrDisplay.stageParameters.sittingToStandingTransform);
-        dolly.applyMatrix(standingMatrix);
+      if (vrDisplay) {
+        if (vrDisplay.stageParameters) {
+          standingMatrix.fromArray(vrDisplay.stageParameters.sittingToStandingTransform);
+          dolly.applyMatrix(standingMatrix);
+        } else {
+          // Apply default camera height
+          var userHeight = this.el.sceneEl.camera.el.components.camera.data.userHeight;
+          dolly.position.y += userHeight;
+          dolly.updateMatrix();
+        }
       }
 
       // Decompose.
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -13,13 +13,12 @@ module.exports.System = registerSystem('tracked-controls', {
     this.lastControllersUpdate = 0;
     // Throttle the (renamed) tick handler to minimum 10ms interval.
     this.tick = utils.throttle(this.throttledTick, 10, this);
-    if (navigator.getVRDisplays) {
-      this.sceneEl.addEventListener('enter-vr', function () {
-        navigator.getVRDisplays().then(function (displays) {
-          if (displays.length) { self.vrDisplay = displays[0]; }
-        });
+    if (!navigator.getVRDisplays) { return; }
+    this.sceneEl.addEventListener('enter-vr', function () {
+      navigator.getVRDisplays().then(function (displays) {
+        if (displays.length) { self.vrDisplay = displays[0]; }
       });
-    }
+    });
   },
 
   updateControllerList: function () {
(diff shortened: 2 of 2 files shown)


src/systems/tracked-controls.js
```diff
@@ -13,12 +13,13 @@ module.exports.System = registerSystem('tracked-controls', {
...
     // Throttle the (renamed) tick handler to minimum 10ms interval.
     this.tick = utils.throttle(this.throttledTick, 10, this);
-    if (!navigator.getVRDisplays) { return; }
-    navigator.getVRDisplays().then(function (displays) {
-      if (displays.length > 0) {
-        self.vrDisplay = displays[0];
-      }
-    });
+    if (navigator.getVRDisplays) {
+      this.sceneEl.addEventListener('enter-vr', function () {
```
ngokevin: How come we need to enter VR before we grab the display here?

src/systems/tracked-controls.js
```diff
@@ -13,12 +13,13 @@ module.exports.System = registerSystem('tracked-controls', {
     this.lastControllersUpdate = 0;
     // Throttle the (renamed) tick handler to minimum 10ms interval.
     this.tick = utils.throttle(this.throttledTick, 10, this);
-    if (!navigator.getVRDisplays) { return; }
-    navigator.getVRDisplays().then(function (displays) {
-      if (displays.length > 0) {
-        self.vrDisplay = displays[0];
-      }
-    });
+    if (navigator.getVRDisplays) {
```
dmarcos: you can still return early here `  if (!navigator.getVRDisplays) { return; }`

src/components/tracked-controls.js
```diff
@@ -84,9 +84,16 @@ module.exports.Component = registerComponent('tracked-controls', {
...
-        standingMatrix.fromArray(vrDisplay.stageParameters.sittingToStandingTransform);
-        dolly.applyMatrix(standingMatrix);
+      if (vrDisplay) {
+        if (vrDisplay.stageParameters) {
+          standingMatrix.fromArray(vrDisplay.stageParameters.sittingToStandingTransform);
+          dolly.applyMatrix(standingMatrix);
+        } else {
+          // Apply default camera height
+          var userHeight = this.el.sceneEl.camera.el.components.camera.data.userHeight;
+          dolly.position.y += userHeight;
```
ngokevin: Could just add directly `dolly.position.y += el.sceneEl.camera.el.getAttribute('camera').userHeight;`

src/systems/tracked-controls.js
machenmusik: an example: tests all fail otherwise as they are not expecting any offset to be applied.

src/components/tracked-controls.js
machenmusik: originally had direct usage, but then couldn't see what value was being used, which was useful.
is there a preference to use getAttribute('camera') rather than components.camera.data?

ngokevin: I'd go through `getAttribute` API

caseyyee:
Works well.  Tested in both Firefox Nightly and Daydream for 3DOF support. 

Prob want to squish the commits.

Thanks for the patches + reviews!

machenmusik:
(@ngokevin you can squash at merge time via Github, correct?)

ngokevin:
Yeah. I can add a unit test too.

src/systems/tracked-controls.js
dmarcos: should navigator.getVRDisplays be mocked during tests instead of modyfing the code to fulfill test needs?

machenmusik: I don't know.  for rift/vive, the tracked controllers currently don't make sense outside vr.  but more generally, no reason not to have tracked controllers unassociated with VR HMD, Should those get a user height offset just because the Rift and Vive controllers do while in VR?

PR #2511, opened 2017-03-22
Validate schema's default values

dye784:
**Description:**
What originally was intended to fix issue #2412 with PR #2500 grew into this PR. The intent of this is to enforce properly parsed default values based on a defined type instead of parsing the `default` into something acceptable.

**Changes proposed:**
Created `validate.js` which contains two functions: `isValidDefaultCoordinate` and `isValidDefaultValue`. `isValidDefaultValue` checks that the type and the `typeof` default value match. It calls `isValidDefaultCoordinate` to check `vec2`, `vec3` and `vec4`. `processPropertyDefinition` throws a warning if `isValidDefaultValue` returns `false`.


dye784:
Noticed something is happening on default that creates something with 
```
schema: {
  type: 'number',
  default: '1.0'
}
```
And that's giving a warning because of the functions I built.

src/utils/validate.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Checks if Valid default coordinates
+ * @param {unknown} possibleCoordinates
```
dmarcos: Can we give to this file a more precise name? `validate-schema.js`, `validate-type.js`

dye784: Good point. Thanks.

src/utils/validate-schema.js
```diff
@@ -0,0 +1,61 @@
+/**
```
ngokevin: I'd actually just put all of this code within the schema module.

src/utils/validate-schema.js
```diff
@@ -0,0 +1,61 @@
...
+ * Checks if Valid default coordinates
+ * @param {unknown} possibleCoordinates
+ * @param {number} dimensions - 2 for 2D Vector or 3 for 3D vector
+ * @returns {boolean} A boolean determining if coordinates are parsed correctly.
+ */
+function isValidDefaultCoordinate (possibleCoordinates, dimensions) {
+  if (typeof possibleCoordinates !== 'object' || possibleCoordinates === null) {
```
ngokevin: Not sure whether `null` should be invalid.

src/utils/validate-schema.js
```diff
@@ -0,0 +1,61 @@
...
+    return false;
+  } else if (Object.keys(possibleCoordinates).length !== dimensions) {
+    return false;
+  } else {
+    if (dimensions === 2 && (possibleCoordinates.x === 0 || possibleCoordinates.x) && (possibleCoordinates.y === 0 || possibleCoordinates.y)) {
```
ngokevin: I think we could condense this code more. We only need to check X/Y/Z/W each once.

src/utils/validate-schema.js
```diff
@@ -0,0 +1,61 @@
...
+module.exports.isValidDefaultCoordinate = isValidDefaultCoordinate;
+
+/**
+ * Validates the default values in a schema
+ * @param {string} type - type in a prop
+ * @param {unknown} defaultVal - default in a prop
+ * @returns {boolean} A boolean determining if defaults are valid.
+ */
+function isValidDefaultValue (type, defaultVal) {
+  if (type === 'audio' && typeof defaultVal !== 'string') return false;
```
ngokevin: Braces around the if code `{` `}`

src/utils/validate-schema.js
```diff
@@ -0,0 +1,61 @@
...
+}
+module.exports.isValidDefaultCoordinate = isValidDefaultCoordinate;
+
+/**
+ * Validates the default values in a schema
+ * @param {string} type - type in a prop
+ * @param {unknown} defaultVal - default in a prop
+ * @returns {boolean} A boolean determining if defaults are valid.
+ */
+function isValidDefaultValue (type, defaultVal) {
```
ngokevin: Can we unit test these? (part of the `schema.test.js`)

dye784: Yeah. Was going to unit test it afterwards since it's quite a bit of tests to write. But I'll get on it.

src/utils/validate-schema.js
dye784: Gotcha.

src/utils/validate-schema.js
dye784: Hmm...that does make sense. Since currently if you do not define a default value, when you register a component it will register a default value at `{ x:0, y:0, z:0 }` If you don't want that to happen assigning it to `null` makes sense. I'll adjust it.

src/utils/validate-schema.js
dye784: Totally, I did it a dumb way. 😂

dye784:
I've made all the changes you suggested @ngokevin. What do you think? Should I write some connection tests?

PR #2512, opened 2017-03-22
unit test controller + userHeight if no stageParam, fix karma version, don't share 3D variables between controller entities

ngokevin:
**Description:**

**Changes proposed:**
- Added unit tests for last PR which applies user height to controllers of no stage parameters.
- Fix Karma version because 1.5.0 doesn't work with console.log
- Don't use closure for private variables in tracked-controllers component because that will get shared and polluted between multiple instances of the controller. Prevents potential bugs and stopped some of my tests from failing.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -1,6 +1,9 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 
+var ZERO_ORIENTATION = [0, 0, 0, 1];
+var ZERO_POSITION = [0, 0, 0];
+
 /**
  * Tracked controls component.
  * Wrap the gamepad API for pose and button states.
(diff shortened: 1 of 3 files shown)


src/components/tracked-controls.js
```diff
@@ -54,69 +63,65 @@ module.exports.Component = registerComponent('tracked-controls', {
...
+    var pose;
+    var position;
+    var standingMatrix = this.standingMatrix;
+    var vrDisplay = this.system.vrDisplay;
+
+    if (!controller) { return; }
+
+    // Compose pose from Gamepad.
+    pose = controller.pose;
+    orientation = pose.orientation || [0, 0, 0, 1];
```
dmarcos: Can we have this as constants to avoid instantiating an array on each frame?

src/components/tracked-controls.js
```diff
@@ -54,69 +63,65 @@ module.exports.Component = registerComponent('tracked-controls', {
...
+    position = pose.position || [0, 0, 0];
```
dmarcos: same here. A constant?

ngokevin:
done

PR #2513, opened 2017-03-22
Axismove for touch and vive

machenmusik:
Per discussion on #2423, this PR separates the axismove changes and omits the emulated property.

This PR extends #2413 / #2415 to oculus-touch-controls as thumbstickmoved, and also addresses spurious axismove events discussed in #2423.

### DIFF
--- a/docs/components/oculus-touch-controls.md
+++ b/docs/components/oculus-touch-controls.md
@@ -34,34 +34,43 @@ mappings, events, and a Touch controller model.
 | triggerup            | Trigger released.              |
 | triggertouchstart    | Trigger touched.               |
 | triggertouchend      | Trigger no longer touched.     |
+| triggerchanged       | Trigger changed.               |
 | thumbstickdown       | Thumbstick pressed.            |
 | thumbstickup         | Thumbstick released.           |
 | thumbsticktouchstart | Thumbstick touched.            |
 | thumbsticktouchend   | Thumbstick no longer touched.  |
+| thumbstickchanged    | Thumbstick changed.            |
 | gripdown             | Grip button pressed.           |
 | gripup               | Grip button released.          |
 | griptouchstart       | Grip button touched.           |
 | griptouchend         | Grip button no longer touched. |
-| Adown                | A button pressed.              |
-| Aup                  | A button released.             |
-| Atouchstart          | A button touched.              |
-| Atouchend            | A button no longer touched.    |
-| Bdown                | B button pressed.              |
-| Bup                  | B button released.             |
-| Btouchstart          | B button touched.              |
-| Btouchend            | B button no longer touched.    |
-| Xdown                | X button pressed.              |
-| Xup                  | X button released.             |
-| Xtouchstart          | X button touched.              |
-| Xtouchend            | X button no longer touched.    |
-| Ydown                | Y button pressed.              |
-| Yup                  | Y button released.             |
-| Ytouchstart          | Y button touched.              |
-| Ytouchend            | Y button no longer touched.    |
+| gripchanged          | Grip button changed.           |
+| abuttondown          | A button pressed.              |
+| abuttonup            | A button released.             |
+| abuttontouchstart    | A button touched.              |
+| abuttontouchend      | A button no longer touched.    |
+| abuttonchanged       | A button changed.              |
+| bbuttondown          | B button pressed.              |
+| bbuttonup            | B button released.             |
+| bbuttontouchstart    | B button touched.              |
+| bbuttontouchend      | B button no longer touched.    |
+| bbuttonchanged       | B button changed.              |
+| xbuttondown          | X button pressed.              |
+| xbuttonup            | X button released.             |
+| xbuttontouchstart    | X button touched.              |
+| xbuttontouchend      | X button no longer touched.    |
+| xbuttonchanged       | X button changed.              |
+| ybuttondown          | Y button pressed.              |
+| ybuttonup            | Y button released.             |
+| ybuttontouchstart    | Y button touched.              |
+| ybuttontouchend      | Y button no longer touched.    |
+| ybuttonchanged       | Y button changed.              |
 | menudown             | Menu button pressed.           |
 | menuup               | Menu button released.          |
+| menuchanged          | Menu button changed.           |
 | systemdown           | System button pressed.         |
 | systemup             | System button released.        |
+| systemchanged        | System button changed.         |
 
 ## Assets
 
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -2,8 +2,6 @@ var registerComponent = require('../core/component').registerComponent;
 var bind = require('../utils/bind');
 var isControllerPresent = require('../utils/tracked-controls').isControllerPresent;
 
-var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
-
 var DAYDREAM_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/google/';
 var DAYDREAM_CONTROLLER_MODEL_OBJ_URL = DAYDREAM_CONTROLLER_MODEL_BASE_URL + 'vr_controller_daydream.obj';
 var DAYDREAM_CONTROLLER_MODEL_OBJ_MTL = DAYDREAM_CONTROLLER_MODEL_BASE_URL + 'vr_controller_daydream.mtl';
@@ -18,7 +16,7 @@ var GAMEPAD_ID_PREFIX = 'Daydream Controller';
  */
 module.exports.Component = registerComponent('daydream-controls', {
   schema: {
-    hand: {default: DEFAULT_HANDEDNESS}, // This informs the degenerate arm model.
+    hand: {default: ''}, // This informs the degenerate arm model.
     buttonColor: {type: 'color', default: '#000000'},
     buttonTouchedColor: {type: 'color', default: '#777777'},
     buttonHighlightColor: {type: 'color', default: '#FFFFFF'},
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -29,20 +29,21 @@ module.exports.Component = registerComponent('daydream-controls', {
   // 1 - menu ( never dispatched on this layer )
   // 2 - system ( never dispatched on this layer )
   mapping: {
-    axis0: 'trackpad',
-    axis1: 'trackpad',
-    button0: 'trackpad',
-    button1: 'menu',
-    button2: 'system'
+    axes: {'trackpad': [0, 1]},
+    buttons: ['trackpad', 'menu', 'system']
   },
+
+  // Use these labels for detail on axis events such as thumbstickmoved.
+  // e.g. for thumbstickmoved detail, the first axis returned is labeled x, and the second is labeled y.
+  axisLabels: ['x', 'y', 'z', 'w'],
 
   bindMethods: function () {
     this.onModelLoaded = bind(this.onModelLoaded, this);
     this.onControllersUpdate = bind(this.onControllersUpdate, this);
     this.checkIfControllerPresent = bind(this.checkIfControllerPresent, this);
     this.removeControllersUpdateListener = bind(this.removeControllersUpdateListener, this);
-    this.onGamepadConnected = bind(this.onGamepadConnected, this);
-    this.onGamepadDisconnected = bind(this.onGamepadDisconnected, this);
+    this.onAxisMoved = bind(this.onAxisMoved, this);
+    this.onGamepadConnectionEvent = bind(this.onGamepadConnectionEvent, this);
   },
 
   init: function () {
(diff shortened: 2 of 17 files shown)


machenmusik:
@cmvanb - we decided to close #2423 in favor of this PR, which should have the same benefits

docs/components/oculus-touch-controls.md
```diff
@@ -34,34 +34,43 @@ mappings, events, and a Touch controller model.
...
 | gripdown             | Grip button pressed.           |
 | gripup               | Grip button released.          |
 | griptouchstart       | Grip button touched.           |
 | griptouchend         | Grip button no longer touched. |
+| gripchanged          | Grip button changed.           |
 | Adown                | A button pressed.              |
 | Aup                  | A button released.             |
 | Atouchstart          | A button touched.              |
 | Atouchend            | A button no longer touched.    |
+| Achanged             | A button changed.              |
```
dmarcos: can we make all the event names lower case?

machenmusik: the ones that correspond to buttons A, B, X, and Y are uppercase for clarity because that is the way the button labels are always used, IMO don't think we should lowercase them?  worried about namespace collisions too.  Note the existing support is the same way...

dmarcos:
Our event names don't have uppercase. Can we keep them lower cased?

dmarcos:
we could do `abuttonchanged`

machenmusik:
`buttonchanged` is a slightly different but used event; personally not in favor of `abuttonup`, `abuttonchanged`, `bbuttonchanged`, etc. due to being too close to the separately used `buttonup`, `buttonchanged`, etc. -- but it is of course technically possible by changing mappings here and also in #2446.

machenmusik:
@dmarcos @ngokevin any more thoughts on this one?  hoping to get #2446 in as well, but this one first.

interested in outside perspectives on backward compatibility (whether the event names change is problematic for existing usage) -- I think right now most things use trigger / thumbstick / grip which aren't changing.

src/components/oculus-touch-controls.js
```diff
@@ -50,8 +51,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       button5: 'surface'
     },
     'right': {
-      axis0: 'thumbstick',
-      axis1: 'thumbstick',
+      axes: {'thumbstick': [0, 1]},
```
dmarcos: I prefer to have `axis + index` instead of an array for consistency with buttons

machenmusik: makes processing harder e.g. you can't just do buttons.forEach ... i would argue we should change buttons to array for consistency instead.

machenmusik: so like axes, buttons would be a keyed map of either singletons or arrays (if aliased)

buttons: { thumbstick: 0, ... }

more readable IMO

docs/components/oculus-touch-controls.md
machenmusik: done and also propagated into #2446

src/components/oculus-touch-controls.js
```diff
@@ -139,55 +156,76 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
-    if (button === 'trigger' && buttonMeshes.trigger) {
-      buttonMeshes.trigger.rotation.x = -analogValue * (Math.PI / 24);
+    if (!button) { return; }
+
+    if (button === 'trigger' || button === 'grip') {
+      // At the moment, if trigger or grip,
+      // touch events aren't happening (touched is stuck true);
+      // synthesize touch events from very low analog values.
+      analogValue = evt.detail.state.value;
+      isPreviousValueEmulatedTouch = this.isEmulatedTouchEvent(this.previousButtonValues[button]);
```
dmarcos: Touch is going to be supported soon in Firefox. I would prefer to wait for the implementation to land rather hand handling the quirks that we are going to remove soon.

machenmusik: Touch is supported in Nightly, see #2446

dmarcos: Is this touch emulation a good idea from the UX perspective? Can we move it to a separate PR?

machenmusik: that code is not new. it's already in.  whether it is still needed depends in part on the Nightly button/axis mapping, which we should probably discuss in #2446

src/components/oculus-touch-controls.js
```diff
@@ -103,27 +98,49 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
-    el.sceneEl.removeEventListener('controllersupdated', this.onControllersUpdate, false);
-    window.removeEventListener('gamepadconnected', this.checkIfControllerPresent, false);
-    window.removeEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);
   },
 
   checkIfControllerPresent: function () {
     var data = this.data;
-    var isPresent = this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, { hand: data.hand });
+    var isPresent = false;
+    var whichControllers;
```
dmarcos: You can do in one line `var whichControllers = ...`

src/components/oculus-touch-controls.js
machenmusik: update: the 2017-03-22 build of Nightly makes Touch mapping compatible with Chromium, so #2446 is no longer required.  Also, Nightly provides analog value for all buttons, so we do in fact need this emulated touch code (e.g. to make hand-controls work correctly) since Nightly does not otherwise provide capacitive touch events.

dmarcos: I asked our Firefox devs what's the ETA for the touch events. This will be coming soon so I would try to get rid of the hack and wait for Firefox to catch up.

machenmusik: according to @ngokevin there are cases where analog values have very small nonzero values, which may still need functionality similar/identical to this...

dmarcos:
I'm a bit lost with all this PRs. What does this one includes?

machenmusik:
axismove events similar to #2413 / #2415 for oculus touch (prior commits were vive specific, which frankly we should really avoid where possible in future...)

machenmusik:
```
This PR extends #2413 / #2415 to oculus-touch-controls as thumbstickmoved, and also addresses spurious axismove events discussed in #2423.
```

src/components/daydream-controls.js
```diff
@@ -89,40 +90,33 @@ module.exports.Component = registerComponent('daydream-controls', {
...
   },
 
   play: function () {
     this.checkIfControllerPresent();
-    window.addEventListener('gamepadconnected', this.onGamepadConnected, false);
-    window.addEventListener('gamepaddisconnected', this.onGamepadDisconnected, false);
     this.addControllersUpdateListener();
-    this.addEventListeners();
+    window.addEventListener('gamepadconnected', this.onGamepadConnectionEvent, false);
+    window.addEventListener('gamepaddisconnected', this.onGamepadConnectionEvent, false);
```
dmarcos: We can do `window.addEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);`

src/components/daydream-controls.js
```diff
@@ -31,20 +31,21 @@ module.exports.Component = registerComponent('daydream-controls', {
...
+
   bindMethods: function () {
     this.onModelLoaded = bind(this.onModelLoaded, this);
     this.onControllersUpdate = bind(this.onControllersUpdate, this);
     this.checkIfControllerPresent = bind(this.checkIfControllerPresent, this);
     this.removeControllersUpdateListener = bind(this.removeControllersUpdateListener, this);
-    this.onGamepadConnected = bind(this.onGamepadConnected, this);
-    this.onGamepadDisconnected = bind(this.onGamepadDisconnected, this);
+    this.onAxisMoved = bind(this.onAxisMoved, this);
+    this.onGamepadConnectionEvent = bind(this.onGamepadConnectionEvent, this);
```
dmarcos: We can get rid of this since we can do `window.addEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);`

src/components/daydream-controls.js
```diff
@@ -89,40 +90,33 @@ module.exports.Component = registerComponent('daydream-controls', {
...
-    // this.removeControllersUpdateListener();
-    this.checkIfControllerPresent();
-  },
-
-  onGamepadDisconnected: function (evt) {
-    // for now, don't disable controller update listening, due to
-    // apparent issue with FF Nightly only sending one event and seeing one controller;
-    // this.everGotGamepadEvent = true;
-    // this.removeControllersUpdateListener();
+  onGamepadConnectionEvent: function (evt) {
```
dmarcos: We can get rid of this method

src/components/daydream-controls.js
```diff
@@ -89,40 +90,33 @@ module.exports.Component = registerComponent('daydream-controls', {
...
+    this.everGotGamepadEvent = true;
```
dmarcos: This flag is never checked

src/components/oculus-touch-controls.js
```diff
@@ -103,50 +98,49 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
-    this.checkIfControllerPresent();
+    } else { this.removeEventListeners(); }
   },
 
-  onGamepadDisconnected: function (evt) {
-    // for now, don't disable controller update listening, due to
-    // apparent issue with FF Nightly only sending one event and seeing one controller;
-    // this.everGotGamepadEvent = true;
-    // this.removeControllersUpdateListener();
+  onGamepadConnectionEvent: function (evt) {
```
dmarcos: We can get rid of this method and do `window.addEventListener('gamepaddisconnected', this.checkIfControllerPresent, false);`

src/components/daydream-controls.js
machenmusik: you can see why in the comments, but yeah we can remove them for now

machenmusik: actually it is checked in onControllersUpdate (which last I recall is actually fired every tick).  so the idea is that if you got gamepad connect/disconnect events, you don't always need to be checking to see when it shows up

src/components/daydream-controls.js
machenmusik: see below comment on the everGotGamepadEvent flag

src/components/daydream-controls.js
machenmusik: see below comment on the everGotGamepadEvent flag

src/components/daydream-controls.js
machenmusik: see below comment on the everGotGamepadEvent flag

src/components/oculus-touch-controls.js
machenmusik: I think the question is whether we should restore the check in onControllersUpdate (don't look for touch controllers every tick the controllers list is updated), which would probably be more efficient

machenmusik:
summing up recent comments - in daydream-controls, everGotGamepadEvent is in fact used in onControllersUpdate, which is more efficient.  oculus-touch-controls was not and so was checking for controller presence every tick, so changed it to stop doing that if everGotGamepadEvent (like the other controllers).

src/components/oculus-touch-controls.js
machenmusik: I have added the check in onControllersUpdate to match the other controller types.

src/components/daydream-controls.js
dmarcos: We don't have to set the flag to `false` on `gamepaddisconnected`?

machenmusik: no because we expect gamepadconnected next time since we're getting events, right?  so we don't need to poll every tick

src/components/oculus-touch-controls.js
ngokevin: We filed bugs about making the threshold configurable in Firefox to address the small values.

dmarcos: I think we can get rid of all the emulated touch logic since Firefox is already supporting touch events.

dmarcos: On oculus touch controllers I believe the touch flag should be independent of the button value since they have a capacitive surface.

dmarcos:
I've tried this PR with Vive and Oculus touch and I observed a couple of problems:
1. The analog buttons (triggers in Vive an Oculus) return duplicated touchstart / touchend events on a single press.
2. The tracked controllers example only shows one hand with Vive while it works with both hands on Oculus

machenmusik:
> The analog buttons (triggers in Vive an Oculus) return duplicated touchstart / touchend events on a single press.

Hmm, if that's true it sounds as though real events are working correctly and we can remove the emulation stuff altogether?  Otherwise I would expect only some doubling... will look at suppressing actual when we are doing emulation.

> The tracked controllers example only shows one hand with Vive while it works with both hands on Oculus

That is the usual symptom of problems with gamepadconnected/disconnected issues.  Are you sure you tried the latest version including `d8ae5cc`?  When we were working on things this afternoon (notably the support-vive-tracker-dist codepen) I saw both hands with Nightly and Vive just fine...

I am thinking that given the apparently unreliable behavior of gamepadconnected/disconnected propagation, maybe those handlers should be removed altogether again for now (as you suggested in comments), and we should send (throttled) events to cause periodic check from system level.  If we don't throttle, we'll be in the scale-unfriendly pattern of each candidate controller component redoing its own version of checkIfControllerPresent every tick -- which is soon to be two hands times four controller components, and maybe five if you count vive-tracker.

dmarcos:
What is exactly the problem with `gamepadconnected` / `gamepaddisconnected` ? I want to know how to report it as a bug for Firefox.

machenmusik:
With vive trackers, i saw only one gamepadconnected fire on one of the vive-tracker components, in a scene with two actual trackers and two separate entities, each with one instance of vive-tracker set to filter for the right index.  Since only one instance received it, that was the one that realized it was present and decided to show up.

gamepadconnected doesn't know anything about index, so for that to work, the expectation is that the gamepadconnected event would propagate to all listening components, so they can do their own checks against the system controllers list.  If the two tracker connections were coalesced into a single gamepadconnected event, things would still work if that event fired when both were available via gamepad API, and if all the listening components actually received it.

can't use the codepen I had before as example since I committed the changes back to checking on controllersupdate.

src/components/oculus-touch-controls.js
```diff
@@ -254,6 +224,25 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    var self = this;
+    var axesMapping = this.mapping[this.data.hand].axes;
+    // In theory, it might be better to use mapping from axis to control.
+    // In practice, it is not clear whether the additional overhead is worthwhile,
+    // and if we did grouping of axes, we really need de-duplication there.
+    Object.keys(axesMapping).forEach(function (key) {
+      var value = axesMapping[key];
+      var detail = {};
+      var changed = !evt.detail.changed;
+      if (!changed) { value.forEach(function (axisNumber) { changed |= evt.detail.changed[axisNumber]; }); }
```
dmarcos: Can we we do:
````javascript
if (!changed) {
.. } else {
...
}
````

Comment (machenmusik) [inline src/components/oculus-touch-controls.js]: yes to the indentation, no to the else clause (because the forEach may change the value of `changed` and it needs to be re-evaluated)

**Comment by machenmusik** [comment]
so I think we're good now, but you're probably gonna want to have GitHub squash when merging

File: src/components/oculus-touch-controls.js [hunk 11]
```diff
@@ -234,7 +234,9 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       var value = axesMapping[key];
       var detail = {};
       var changed = !evt.detail.changed;
-      if (!changed) { value.forEach(function (axisNumber) { changed |= evt.detail.changed[axisNumber]; }); }
+      if (!changed) {
```
dmarcos: What about this?
````javascript
var axisChanged = false;
if (!evt.detail.changed) { return }
value.forEach(function (axisNumber) { 
  axisChanged |= evt.detail.changed[axisNumber]; 
  detail[self.axisLabels[axisNumber]] = evt.detail.axis[axisNumber];
}); 
if (!axisChanged) { return; } 
self.el.emit(key + 'moved', detail);
````

machenmusik: it has the side effect of making more garbage in detail even if not changed, which is less desirable

machenmusik:
Since #2545 was merged first, I've incorporated the salient aspects of #2588 here now per discussion.

dmarcos:
Thanks!

Issue #2497, opened 2017-03-16
Expose `material.alphaTest` and `material.depthWrite` properties


ngokevin:
**Description:**

To configure depth testing.

In `src/components/material.js`.

http://stackoverflow.com/questions/42799420/why-is-the-opacity-different-depending-on-the-camera-angle/42822927?noredirect=1#comment72797756_42822927

takahirox:
Let me take this.

takahirox:
I think we should set `material.needsUpdate = true` in `updateMaterial()` if `alphaTest` value is updated, correct?

alfa256:
:+1:  Was going to post this.

PR #2516, opened 2017-03-23
Expose material.alphaTest and material.depthWrite properties

takahirox:
**Description:**

#2497

**Changes proposed:**
- Expose `material.alphaTest` and `material.depthWrite` properties
- Sets `material.needsUpdate` `true` if `material.alphaTest` is updated (Three.js manner)


dmarcos:
Thanks!

PR #2521, opened 2017-03-24
Check for presence of parent element

rxl881:
**Description:**

In some cases a component can be created and unmounted before it has fully finished loading. This leads to errors when the entityLoadCallback tries to access the isPlaying property of the parent element (as it no longer exists).

screen shot 2017-03-24 at 11 23 43
screen shot 2017-03-24 at 11 24 04

**Changes proposed:**
- Check for existence of entity parent

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -295,7 +295,7 @@ var proto = Object.create(ANode.prototype, {
 
       ANode.prototype.load.call(this, function entityLoadCallback () {
         self.updateComponents();
-        if (self.isScene || self.parentEl.isPlaying) { self.play(); }
+        if (self.isScene || (self.parentEl && self.parentEl.isPlaying)) { self.play(); }
       });
     },
     writable: window.debug
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -291,7 +291,7 @@ var proto = Object.create(ANode.prototype, {
     value: function () {
       var self = this;
 
-      if (this.hasLoaded) { return; }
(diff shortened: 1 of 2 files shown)


src/core/a-entity.js
```diff
@@ -295,7 +295,7 @@ var proto = Object.create(ANode.prototype, {
 
       ANode.prototype.load.call(this, function entityLoadCallback () {
         self.updateComponents();
-        if (self.isScene || self.parentEl.isPlaying) { self.play(); }
+        if (self.isScene || (self.parentEl && self.parentEl.isPlaying)) { self.play(); }
```
ngokevin: You think it'd be possible to reproduce this in a unit test or JS consistently? Like load and unload a component, and the error previously threw every time?

dmarcos:
Yes, this PR could be hiding the symptoms of a deeper issue. In principle the `load` method should only fire when the component is attached to the DOM and hence has a parent element

dmarcos:
We need a test case to see in what conditions the problem manifests

ngokevin:
Yeah, a unit test that fails before this PR, and passes after this PR would be perfect. We had a PR for this previously, but wanted to nail it down.

ngokevin:
@rxl881 Can I see your code that reproduces this?

ngokevin:
OK, I think I can reproduce this in a way.

ngokevin:
Found the test cases, tweaked it to check parentEl earlier. Thanks!

rxl881:
@ngokevin - Sorry for lack of response (I was swamped with work stuff for a bit). Thanks very much for adding the test / fixing.

PR #2525, opened 2017-03-28
Show an error message when the inspector URL could not be loaded

belen-albeza:
**Description:**

I was giving an A-Frame talk and wanted to show the Inspector… Could not make it because I didn't have a working network connection. At that time, I didn't know that the Inspector was being pulled on the fly, and the "Loading…" message was being shown indefinitely, no errors on the console, and I found it confusing.

This shows an error message in the "loading" magenta button when the injected `<script>` tag fires an `error` event.

**Changes proposed:**

- Add an `onerror` event handler that shows an error message in place of "Loading…"


ngokevin:
Thanks!

PR #2526, opened 2017-03-28
hand-controls support for Nightly with Oculus Touch

machenmusik:
This PR includes #2513 in its entirety (is intended to be committed after #2513 is merged) and modifies `hand-controls` to accommodate Nightly support for Oculus Touch by using button press as well as touch for all gesture determinations.

As of 2017-03-22, Nightly has adopted button and axis mapping that is compatible with Chromium, so #2446 is no longer necessary; however, it does not provide the capacitive touch events that Chromium does, so this cherry-pick is needed for full `hand-controls` gesture support.

Per discussion, #2513 modifies the event names generated by Oculus Touch, which is why that PR is required.


machenmusik:
if Nightly supports touched property on gamepads, then we may not need this one anymore,,,

Issue #2366, opened 2017-02-09
document Entity.appendChild asyncrony in "Using JavaScript and DOM APIs"


caseyyee:
- A-Frame Version: 4.1
- Platform / Device: windows
- Reproducible Code Snippet or URL:

```js
var scene = document.querySelector('a-scene');
var entity = document.createElement('a-entity');
scene.appendChild(entity);
entity.setAttribute('position', { x: 0, y: 1, z: 2 });

// returns undefined
console.log(entity.getAttribute('position'));

setTimeout(function() {
  // returns Object {x: 0, y: 1, z: 2}
  console.log(entity.getAttribute('position'));
}, 0);
```

Here's the pen http://codepen.io/caseyyee/pen/MJPamO

ngokevin:
`appendChild` is async so we need to wait for the entity to attach and load.

This is not an issue if code is executed using components as recommended: https://aframe.io/docs/0.4.0/guides/using-javascript-and-dom-apis.html#where-to-place-javascript-code-for-a-frame

A note about this can be added here: https://aframe.io/docs/0.4.0/guides/using-javascript-and-dom-apis.html#adding-an-entity-with-appendchild

dmarcos:
You could also do:
````javascript
var entity = document.createElement('a-entity');
scene.appendChild(entity);
entity.setAttribute('position', { x: 0, y: 1, z: 2 });
entity.addEventListener('loaded', function() {
  console.log(entity.getAttribute('position'));
});
````

**Comment by caseyyee** [comment]
@dmarcos @ngokevin As a note, I was setting up entities in a component and none of the attributes were available in `init` or `play` callbacks either.

So you would have to listen for the loaded event within the component itself to avoid this in the same way?

**Comment by ngokevin** [comment]
If the component depends on other components, you can set a dependency. Although we should maybe prioritize pos/rot/scale components. Try:

```js
dependencies: ['position'],
```

caseyyee:
Alright thanks.   Last thing I will say is that this makes it quite inconvenient when dealing with collections of entities and have to work with them immediately.   Especially for cases where your generating content procedurally.

dmarcos:
position, rotation and scale are `defaultComponents`. They should already initialize before than any other component.

ngokevin:
Such is life when dealing with the DOM. Though the concept of systems is for exactly this, managing collections of entities. Possibly:

```js
AFRAME.registerSystem('foo', {
  handleEntityAdded: function (entity) {
[...]

PR #2529, opened 2017-03-29
add documentation note on the asynchrony of appendChild

belen-albeza:
**Description:**

Fix for issue #2366 

**Changes proposed:**
- Added a note to to the documentation about how `appendChild` is async, and a possible workaround.

### DIFF
--- a/docs/guides/using-javascript-and-dom-apis.md
+++ b/docs/guides/using-javascript-and-dom-apis.md
@@ -198,6 +198,21 @@ var entityEl = document.createElement('a-entity');
 sceneEl.appendChild(entityEl);
 ```
 
+Note that `.appendChild()` is an **asynchronous** operation, and until its appending to the DOM has been completed, its attributes cannot be queried (i.e. `.getAttribute()` will return `undefined`). If you need to query an attribute on an entity that has just been appended, you can listen to the `loaded` event on that entity:
+
+```js
+var sceneEl = document.querySelector('a-scene');
+var entityEl = document.createElement('a-entity');
+entityEl.setAttribute('position', {x: 0, y: 1, z: 2});
+sceneEl.appendChild(entityEl);
+
+// calling entity.getAttribute would return undefined now, so we
+// listen to the loaded event instead
+entityEl.addEventListener('loaded', function () {
+    console.log(entityEl.getAttribute('position'));
+});
+```
+
 ### Removing an Entity with `.remove()`
 
 To remove an entity from the DOM and thus from the scene, we call


ngokevin:
Needs a rebase, unlucky someone just touched that same file!

docs/guides/using-javascript-and-dom-apis.md
```diff
@@ -198,6 +198,21 @@ var entityEl = document.createElement('a-entity');
...
+```js
+var sceneEl = document.querySelector('a-scene');
+var entityEl = document.createElement('a-entity');
+entityEl.setAttribute('position', {x: 0, y: 1, z: 2});
+sceneEl.appendChild(entityEl);
+
+// calling entity.getAttribute would return undefined now, so we
+// listen to the loaded event instead
+entityEl.addEventListener('loaded', function () {
+    console.log(entityEl.getAttribute('position'));
```
ngokevin: Could use two space indent for consistency

docs/guides/using-javascript-and-dom-apis.md
```diff
@@ -198,6 +198,21 @@ var entityEl = document.createElement('a-entity');
...
 
+Note that `.appendChild()` is an **asynchronous** operation, and until its appending to the DOM has been completed, its attributes cannot be queried (i.e. `.getAttribute()` will return `undefined`). If you need to query an attribute on an entity that has just been appended, you can listen to the `loaded` event on that entity:
+
+```js
+var sceneEl = document.querySelector('a-scene');
+var entityEl = document.createElement('a-entity');
+entityEl.setAttribute('position', {x: 0, y: 1, z: 2});
+sceneEl.appendChild(entityEl);
+
+// calling entity.getAttribute would return undefined now, so we
```
ngokevin: nit: capitalize + period the comment

docs/guides/using-javascript-and-dom-apis.md
```diff
@@ -198,6 +198,21 @@ var entityEl = document.createElement('a-entity');
 sceneEl.appendChild(entityEl);
 ```
 
+Note that `.appendChild()` is an **asynchronous** operation, and until its appending to the DOM has been completed, its attributes cannot be queried (i.e. `.getAttribute()` will return `undefined`). If you need to query an attribute on an entity that has just been appended, you can listen to the `loaded` event on that entity:
```
ngokevin: I'd also recommend a component if you need to do an operation on a component.

```
AFRAME.registerComponent('do-something-after-loaded', {
  init: function () {
    // Do something.
  }
}); 

// ...
entityEl.setAttribute('do-something', '');
sceneEl.appendChild(entityEl);
```

ngokevin:
Thanks! Rebased and committed at https://github.com/aframevr/aframe/commits/master

PR #2537, opened 2017-03-31
Avoid touchstart collision

machenmusik:
Due to unfortunate name collision, add empty touches array to avoid Daydream error in TouchPanner.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -227,7 +227,8 @@ module.exports.Component = registerComponent('tracked-controls', {
     if (buttonState.touched === previousButtonState.touched) { return false; }
 
     evtName = buttonState.touched ? 'start' : 'end';
-    this.el.emit('touch' + evtName, {id: id, state: buttonState});
+    // Due to unfortunate name collision, add empty touches array to avoid Daydream error.
+    this.el.emit('touch' + evtName, {id: id, state: buttonState, touches: []});
     previousButtonState.touched = buttonState.touched;
     return true;
   },
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -1,5 +1,3 @@
-/* global CustomEvent */
-
 var registerComponent = require('../core/component').registerComponent;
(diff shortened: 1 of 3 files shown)


src/components/tracked-controls.js
```diff
@@ -227,7 +229,10 @@ module.exports.Component = registerComponent('tracked-controls', {
     if (buttonState.touched === previousButtonState.touched) { return false; }
 
     evtName = buttonState.touched ? 'start' : 'end';
-    this.el.emit('touch' + evtName, {id: id, state: buttonState});
+    // Due to unfortunate name collision, add empty touches array to avoid Daydream error.
+    // this.el.emit doesn't allow changes to event outside detail...
+    // this.el.emit('touch' + evtName, {id: id, state: buttonState, touches: []});
+    this.el.dispatchEvent(new CustomEvent('touch' + evtName, {touches: [], detail: {id: id, state: buttonState}}));
```
dmarcos: We should use `this.el.emit`

dmarcos: we could also rename to `touchstarted` / `touchended`

machenmusik: you can't use emit, since it does not let you provide event data other than detail.

machenmusik: touchstart and touchend seem appropriate actually...

TouchPanner does not protect against unexpected input, but it is probably not the only one.  it might make sense to add empty touches to touchend as well, but that is not currently breaking anything

machenmusik: i suppose a fourth argument could be added to emit, or alternate behavior for the third argument to accept more than bubbles?

machenmusik: Tests expect `emit` to be called and thus currently fail.
Correct functionality was confirmed on Daydream including with #2538.
Extend `emit`, or change tests?

dmarcos: We can add a fourth argument `extraData` to `emit`.

dmarcos:
Thanks!

PR #2540, opened 2017-04-01
Throw an error if HTML file is opened with `file://` protocol

### LINKED ISSUE #2533: Throw an error if HTML file is opened with `file://` protocol vs a local server or hosted
**Description:**

Common issue for newbie web developers is they try to open the HTML file by simply opening it and serving it through `file://` protocol. Doesn't work well if there are any assets or file fetching in the scene. A-Frame should throw a loud error in the console.

nikolaiwarner:
**Description:**

Since assets do not load well when the html page is not served from a server, let the user know in the form of a warning.

- Fixes #2533

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,12 @@ if (document.currentScript && document.currentScript.parentNode !== document.hea
        'HTML.');
 }
 
+if (window.location.protocol === 'file:') {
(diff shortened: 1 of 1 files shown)


ngokevin:
thanks!

PR #2544, opened 2017-04-04
fix duplicate asset requests (fixes #2472)

### LINKED ISSUE #2472: Multiple downloads of a-sky image
Looking at the network console (in FF Nightly & Canary), the [boilerplate example for panorama](https://aframe.io/aframe/examples/boilerplate/panorama/) downloads the a-sky source image twice.

It appears this SO may be related:
http://stackoverflow.com/questions/38978890/mozilla-a-frame-loads-images-twice

When using a-assets instead, I'm seeing a 3rd download of the same image.
```
<a-scene>
    <a-assets>
        <img id="sky" crossorigin="anonymous" src="https://aframe.io/aframe/examples/boilerplate/panorama/puydesancy.jpg">
    </a-assets>
    <a-sky src="#sky"></a-sky>
</a-scene>
```

aframe-network1
aframe-network2

ngokevin:
**Description:**

One image was causing three requests:

1. ``  in `<a-assets>`.
2. Again if `crossorigin` wasn't set and A-Frame set it for you + re-requested.
3. three.js TextureLoader

**Changes proposed:**
- Pass `` element through all the way down to the Texture creation process so that three.js TextureLoader doesn't get called. All the way down, we handle string (inline URLs) vs. element (`<a-assets>`).
- Store `` element in THREE.Cache if we use `<a-assets>` so three.js doesn't ever re-fetch.
- Start enforcing `crossorigin` on images because else A-Frame will have to make a second request with the `crossorigin` flag set if it's not set.
- Set `crossOrigin` on the TextureLoader just in case, we had some reports of it not working for inline URLs
- [x] tests


machenmusik:
(sounds like images will work more like canvas and video now)

PR #2548, opened 2017-04-05
system update (fixes #1623)

### LINKED ISSUE #1623: Systems don't support update()
**Description:**
While systems now have schemas and get configured in init() they don't support update() and so can't be reconfigured dynamically.
- A-Frame Version:
- Platform/Device:

ngokevin:
**Description:**

Cribbed a piece of @wizgrav https://github.com/aframevr/aframe/pull/1564

**Changes proposed:**
- System update handler, called on `init` and on data updates with `oldData` just like Components.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -120,10 +120,8 @@ module.exports = registerElement('a-scene', {
 
     initSystem: {
       value: function (name) {
-        var system;
         if (this.systems[name]) { return; }
-        system = this.systems[name] = new systems[name](this);
-        system.init();
+        this.systems[name] = new systems[name](this);
       }
     },
 
@@ -290,7 +288,7 @@ module.exports = registerElement('a-scene', {
     },
 
     /**
-     * Wraps Entity.setAttribute to take into account for systems.
+     * Wrap Entity.setAttribute to take into account for systems.
      * If system exists, then skip component initialization checks and do a normal
      * setAttribute.
      */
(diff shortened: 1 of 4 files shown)


src/core/scene/a-scene.js
```diff
@@ -124,6 +124,7 @@ module.exports = registerElement('a-scene', {
         if (this.systems[name]) { return; }
         system = this.systems[name] = new systems[name](this);
         system.init();
+        system.update({});
```
dmarcos: We call `update` in the [component constructor](https://github.com/aframevr/aframe/blob/master/src/core/component.js#L38). Should we do the same for systems?

src/core/system.js
```diff
@@ -57,6 +49,28 @@ System.prototype = {
   init: function () { /* no-op */ },
 
   /**
+   * Update handler. Called during scene attribute updates.
+   * Systems can use this to dynamically update their state.
+   */
+  update: function (oldData) { /* no-op */ },
+
+  updateProperties: function (rawData) {
```
dmarcos: The `updateProperties` method of a [component](https://github.com/aframevr/aframe/blob/master/src/core/component.js#L232) calls `update`. Should we do the same for consistency?

ngokevin:
Updated.

src/core/system.js
```diff
@@ -35,13 +38,9 @@ var System = module.exports.System = function (sceneEl) {
...
 
   // Process system configuration.
-  if (!Object.keys(schema).length) { return; }
-  rawData = HTMLElement.prototype.getAttribute.call(sceneEl, this.name);
-  if (isSingleProp(schema)) {
-    this.data = parseProperty(rawData, schema);
-    return;
-  }
-  this.data = parseProperties(styleParser.parse(rawData) || {}, schema, false, this.name);
+  this.buildData();
```
dmarcos: Can we call `updateProperties` here?

ngokevin: `updateProperties` doesn't call `init()`, and `init()` needs the data when it's run. So need to run them separately.

PR #2550, opened 2017-04-05
use debug fork with no timestamp/diffing

ngokevin:
**Description:**

We use the console library `debug`. On each log, it appends a millisecond diff from the previous log. If an error/warning/log repeats, then the browser won't be able to collapse all the logs into one line in the console, spamming it. 

I've removed the diff logic on my fork, shaving some bytes in the process. If you do `var warn = AFRAME.utils.debug('hey:warn'); while (1) { warn('hey'); }` in the console, the browser will now know it's repeated.

**Changes proposed:**
- Use `ngokevin/debug#noTimestamp` (https://github.com/ngokevin/debug/tree/noTimestamp) (https://github.com/ngokevin/debug/commit/ef5f8e66d49ce8bc64c6f282c15f8b7164409e3a)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "license": "MIT",
   "dependencies": {
     "browserify-css": "^0.8.2",
-    "debug": "^2.2.0",
+    "debug": "ngokevin/debug#noTimestamp",
     "deep-assign": "^2.0.0",
     "document-register-element": "dmarcos/document-register-element#8ccc532b7",
     "envify": "^3.4.1",


dmarcos:
Should we move the fork to the `aframevr` org like the [webvr-polyfill](https://github.com/aframevr/webvr-polyfill)

ngokevin:
I'd just keep it because we won't need to maintain it in the future or need to show it. Just pointing to a tiny patch I put on top of it.

PR #2553, opened 2017-04-05
Reverse mouse drag in Mobile (fix #2547)

### LINKED ISSUE #2547: Request: look-controls reverseTouchDrag
Currently, you can use the following to invert the scroll direction of a camera on desktop:
`<a-entity camera look-controls="reverseMouseDrag: true"></a-entity>`

This, however, does not work on mobile; it would be useful to have a `reverseTouchDrag` option for mobile.

dmarcos:


src/components/look-controls.js
```diff
@@ -147,11 +147,19 @@ module.exports.Component = registerComponent('look-controls', {
     hmdEuler.setFromQuaternion(hmdQuaternion, 'YXZ');
     if (isMobile) {
       // In mobile we allow camera rotation with touch events and sensors
-      rotation = {
-        x: radToDeg(hmdEuler.x) + radToDeg(pitchObject.rotation.x),
-        y: radToDeg(hmdEuler.y) + radToDeg(yawObject.rotation.y),
-        z: radToDeg(hmdEuler.z)
-      };
+      if (this.data.reverseMouseDrag) {
```
ngokevin: It seems for mobile / touch, there might only be one expected direction since you're physically dragging it (drag the environment, not the camera). Since you wouldn't touch-drag while in first-person/VR.

dmarcos: Well. It depends on the user and the content. Not sure it that's a fair assumption. You might want to also let the developer make the choice to have consistency across desktop / mobile.

ngokevin: If so, maybe we need `reverseTouchDrag` so the mouse + touch directions aren't tied?

dmarcos:
Closing this. We have not seen much more demand.

PR #2545, opened 2017-04-04
Support GearVR 3DOF controller

machenmusik:
This PR, which includes #2538 in its entirety (and is intended to be merged afterward), should add support for the new 3DOF Gear VR Controller as proposed at https://webvr.slack.com/?redir=%2Farchives%2FC0EEECMA5%2Fp1490839724160554

Note that at time of this writing, this PR has not yet been tested on a supporting browser with supported hardware.  Once such testing has been completed, this comment will be edited appropriately.

PR #2561, opened 2017-04-07
sets controller handness automatically.

caseyyee:
Hand selection is automatic as per: https://github.com/aframevr/aframe/commit/cb1399eeb3df1f94d7216371d3794e0770791832#diff-a772568948cd05296a498a049c2b78adR87

So there is no need to specify a hand.

### DIFF
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -2,8 +2,6 @@ var registerComponent = require('../core/component').registerComponent;
 var bind = require('../utils/bind');
 var isControllerPresent = require('../utils/tracked-controls').isControllerPresent;
 
-var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
-
 var DAYDREAM_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/google/';
 var DAYDREAM_CONTROLLER_MODEL_OBJ_URL = DAYDREAM_CONTROLLER_MODEL_BASE_URL + 'vr_controller_daydream.obj';
 var DAYDREAM_CONTROLLER_MODEL_OBJ_MTL = DAYDREAM_CONTROLLER_MODEL_BASE_URL + 'vr_controller_daydream.mtl';
@@ -18,7 +16,7 @@ var GAMEPAD_ID_PREFIX = 'Daydream Controller';
  */
 module.exports.Component = registerComponent('daydream-controls', {
   schema: {
-    hand: {default: DEFAULT_HANDEDNESS}, // This informs the degenerate arm model.
+    hand: {default: ''}, // This informs the degenerate arm model.
     buttonColor: {type: 'color', default: '#000000'},
     buttonTouchedColor: {type: 'color', default: '#777777'},
     buttonHighlightColor: {type: 'color', default: '#FFFFFF'},
--- a/docs/components/daydream-controls.md
+++ b/docs/components/daydream-controls.md
@@ -15,8 +15,14 @@ and/or pressed buttons (trackpad).
 ## Example
 
 ```html
-<a-entity daydream-controls="hand: left"></a-entity>
-<a-entity daydream-controls="hand: right"></a-entity>
+<!-- Match Daydream controller if present, regardless of hand. -->
+<a-entity daydream-controls></a-entity>
+```
+
+```html
+<!-- Match Daydream controller if present and for specified hand. -->
+ <a-entity gearvr-controls="hand: left"></a-entity>
+ <a-entity gearvr-controls="hand: right"></a-entity>
 ```
 
 ## Value
@@ -26,7 +32,7 @@ and/or pressed buttons (trackpad).
 | buttonColor          | Button colors when not pressed.                    | #000000 |
 | buttonTouchedColor   | Button colors when touched.                        | #777777 |
 | buttonHighlightColor | Button colors when pressed and active.             | #FFFFFF |
-| hand                 | The hand that will be tracked (i.e., right, left). | right   |
+| hand                 | Set hand that will be tracked (i.e., right, left). |         |
 | model                | Whether the Daydream controller model is loaded.   | true    |
 | rotationOffset       | Offset to apply to model rotation.                 | 0       |
 
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -1,3 +1,5 @@
+var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
+
 /**
  * Return enumerated gamepads matching id prefix.
  *
@@ -48,7 +50,7 @@ module.exports.isControllerPresent = function (sceneEl, idPrefix, queryObject) {
     isPrefixMatch = (!idPrefix || idPrefix === '' || gamepad.id.indexOf(idPrefix) === 0);
     isPresent = isPrefixMatch;
     if (isPresent && queryObject.hand) {
-      isPresent = gamepad.hand === queryObject.hand;
+      isPresent = (gamepad.hand || DEFAULT_HANDEDNESS) === queryObject.hand;
     }
     if (isPresent && queryObject.index) {
       isPresent = index === queryObject.index; // need to use count of gamepads with idPrefix


machenmusik:
you still need the hand filter, for cases where it gets passed down (e.g. hand-controls="left" should only match daydream-controls if left hand) -- did you want to make it so default hand is empty that would be accepted for any hand?

ngokevin:
I think already introducing daydream-controls into hand-controls is causing issues on how things should actually be or makes us consider cases that shouldn't happen (e.g., hand-controls doesn't make sense for Daydream). What if we removed `daydream-controls` from `hand-controls`, it would clear up the issues with this PR?

machenmusik:
it's not just hand-controls.  shoot-controls for a-blast is handed.  in a-paint, paint-controls (I think) is handed.  goes back to the whole top-down declarative vs. bottom-up assembly discussion.  because of the original paradigm and examples, most apps out there declare what they want to see like hand-controls, and deal with what ends up getting populated.  if you want to be able to use daydream-controls without filtering on hand, that is easy, but that is kinda the opposite of what this PR does (removing filtering and forcing incompatibility)

dmarcos:
I think I would remove `daydream-controls` from `hand-controls` since it implements an interaction model that does not translate well to 3DOF. I would create a separate example (laser pointer kind of interaction technique) to illustrate 3DOF controllers

caseyyee:
+1 on removing from hand-controls.

I do understand why you would want to try to include it into hand-controls, but I think the controls are sufficiently different in 3dof that it can't really be made to work.

As far as hand filtering.  I will refer to comments made in slack:

image

caseyyee:
As far is this PR goes.  We'll open changes to remove hand-controls as a separate PR, and use this one to address the daydream-controls hand filtering changes.

ngokevin:
r+ from me

machenmusik:
the commit only removes from docs, if you want to allow default to accept either hand, I think that change needs to be made to `src/components/daydream-controls.js` right?

caseyyee:
OK, just talked this through with @machenmusik:

Make default "either" and if that's the value, don't use the hand as filter criteria.

Will add to this PR code changes (not just docs).

caseyyee:
@machenmusik r?

src/components/daydream-controls.js
```diff
@@ -83,7 +83,9 @@ module.exports.Component = registerComponent('daydream-controls', {
   },
 
   checkIfControllerPresent: function () {
-    var isPresent = this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, {hand: this.data.hand});
+    var isPresent = this.hands.reduce(function (present, hand) {
+      return present && this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, {hand: hand});
+    }.bind(this));
```
machenmusik: a little different from prior discussion; this will fail to indicate presence if gamepad doesn't include hand property, rather than not filtering if default value is provided... which is presumably why you needed to go into tracked-controls?

src/components/tracked-controls.js
```diff
@@ -21,6 +21,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     id: {type: 'string', default: ''},
     idPrefix: {type: 'string', default: ''},
     rotationOffset: {default: 0},
+    hand: {type: 'string'},
```
machenmusik: @dmarcos specifically wanted to remove hand from tracked-controls, so was surprised to see you add it back in.  I had actually renamed it to defaultHand before @dmarcos asked to remove it entirely and just use right DEFAULT_HANDEDNESS constant.  so I am not sure we want this, per that prior discussion?

src/components/tracked-controls.js
```diff
@@ -84,7 +85,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     var controllerQuaternion = this.controllerQuaternion;
     var controllerEuler = this.controllerEuler;
     var deltaControllerPosition = this.deltaControllerPosition;
-    var hand = (controller ? controller.hand : undefined) || DEFAULT_HANDEDNESS;
+    var hand = this.data.hand || (controller ? controller.hand : undefined) || DEFAULT_HANDEDNESS;
```
machenmusik: is the intent here to allow the arm model to use something different than gamepad hand property when present?  or to allow specifying default hand to use when gamepad doesn't provide?  if the latter, I had actually renamed it to defaultHand before @dmarcos asked to remove it entirely and just use right DEFAULT_HANDEDNESS constant. so I am not sure we want this, per that prior discussion?

src/components/tracked-controls.js
dmarcos: Yes, I think we should remove it all together. This should not be a choice make by the application

src/components/tracked-controls.js
dmarcos: Yep. this should not be part of the component API. Let's keep the const approach we had before

src/components/daydream-controls.js
```diff
@@ -18,14 +16,16 @@ var GAMEPAD_ID_PREFIX = 'Daydream Controller';
  */
 module.exports.Component = registerComponent('daydream-controls', {
   schema: {
-    hand: {default: DEFAULT_HANDEDNESS}, // This informs the degenerate arm model.
+    hand: {default: ''}, // This informs the degenerate arm model.
```
dmarcos: We can get rid of this property. The handness will be determined either by the gamepad property or the default const. Not up to the application to decide

src/components/daydream-controls.js
```diff
@@ -18,14 +16,16 @@ var GAMEPAD_ID_PREFIX = 'Daydream Controller';
...
     buttonColor: {type: 'color', default: '#000000'},
     buttonTouchedColor: {type: 'color', default: '#777777'},
     buttonHighlightColor: {type: 'color', default: '#FFFFFF'},
     model: {default: true},
     rotationOffset: {default: 0} // use -999 as sentinel value to auto-determine based on hand
   },
 
+  hands: ['left', 'right'],
```
dmarcos: We don't need this

src/components/daydream-controls.js
machenmusik: so we have had some discussion here.  if you remove the hand property that is used to filter so it only matches a particular hand,  then it is not possible to specify which handedness model should be used correctly.  i thought we had generally agreed to keep the filter property, but to default it to a value that means don't filter and simply accept whatever is there ("either", or blank).

src/components/daydream-controls.js
machenmusik: agree, if asked to filter by hand, should pass hand as filter for isControllerPresent

src/components/tracked-controls.js
caseyyee: Ah OK.   I didn't quite understand what you were getting at.  Yeah, I'm OK with removing any application handling of hand data.

src/components/daydream-controls.js
caseyyee: @machenmusik Yes that's right.  OK, maybe the approach here is to get tracked-controls to work without the hand property.

src/components/tracked-controls.js
caseyyee: Yes, the idea was that you could manually override any controller.hand defined property.  But I'm OK with removing that as per other comment about not giving the application a option to configure.

caseyyee:
Let me clarify my intent as I found the existing behaviour a bit confusing in use and I don't think I have been entirely clear on what the issues are:

1.  Default handedness breaks for left handed users:
```html
<a-entity daydream-controls></a-entity>
```
This on it's own defaults to the `DEFAULT_HANDEDNESS` which is set to `right`.   This causes an issue when you use the `daydream-controls` as a _left_ handed user (and configured as _left_ handed user from daydream).  `isControllerPresent` doesn't find the controller in this case, and the controller just seems broken/non-functional.

The expectation is that the daydream-controls would correctly map the `controller.hand` and apply the correct arm model to the appropriate hand my daydream is configured to.

2. Specifying both hands is confusing (solution to problem 1):
```html
<a-entity daydream-controls="hand: left">
<a-entity daydream-controls="hand: right">
```
But this causes confusion to anyone using the controls since daydream, is of course a single controller.

This can also be confusing to handle at a application level if you need to configure content to one hand or another (since the reach is different for left and right hands).

The solution:

By specifying just the `daydream-controls`, it should default to the `controller.hand` property to set handedness.  This will make it work for both left and right handed users without specifying two elements.

We should also expose the hand through the component that is in use, so that applications can adjust accordingly based on hand.

Also, a little more thinking and comments in this PR, I agree that letting applications select a arm model manually is not important.

machenmusik:
The existing daydream-controls implementation is based on oculus / vive / tracked-controls core components (most of which predates my involvement, so others may need to weigh in on earlier context).  Those have all leveraged the same structure: 

tracked-controls handles the real interfacing with gamepad API, and has schema values that let the app filter which controller the component should bind to and represent.  (I have generally described that that the top-down approach, where the user specifies what they want, and if present the components do appropriate bindings.)

higher-level components (e.g. oculus-touch-controls and vive-controls ) are allowed and indeed encouraged to be multiply specified, and only those actually present are bound using tracked-controls.

The approach you are suggesting (remove the hand property, have the component use what it provides to you) is what I generally describe as bottom-up approach -- which was previously rejected when proposed w.r.t. oculus and vive, specifying multiple options explicitly was deemed preferable.

In your proposal, not only would you remove the hand property, you'd also want to expose the hand through the component in use -- which is actually what the hand property provides for?

> 2, Specifying both hands is confusing (solution to problem 1):
> `<a-entity daydream-controls="hand: left"> <a-entity daydream-controls="hand: right">`

Actually, I believe we proposed that to allow either hand to match depending on what was actually exposed via gamepad API, the usage should be what you stated was preferable:

> `<a-entity daydream-controls>`

by keeping the hand property, but having it default to a value (either blank, or "either") that would simply omit the hand filter given to isControllerPresent.

In that case, the natural behavior is as described as preferable, but also a specific hand filtered binding can still be specified.

machenmusik:
Using a specific example to illustrate...

Let's say that by necessity and design, I have a separate model for left hand and right hand, and I'd like to use declarative specification to accomplish the following:

1. If I don't have a Daydream controller available, don't show either hand model.

2. If I do have a Daydream controller available, use the correct hand model based on whether the system decided that the Daydream controller represented left or right hand.

If we remove the hand property, there is no way to do this with `<a-entity daydream-controls obj-model=...>`.

If we keep the hand property to use as filter, then this would be done the same way as with oculus or vive:
`<a-entity daydream-controls="hand:left" obj-model=(details for left hand model)><a-entity>`
`<a-entity daydream-controls="hand:right" obj-model=(details for right hand model)><a-entity>`
and only one of the entities would actually get the controller.

src/components/daydream-controls.js
```diff
@@ -83,7 +81,14 @@ module.exports.Component = registerComponent('daydream-controls', {
   },
 
   checkIfControllerPresent: function () {
-    var isPresent = this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, {hand: this.data.hand});
+    var handedness = this.data.hand === '' ? ['left', 'right'] : [this.data.hand];
```
machenmusik: it doesn't seem right to me to go through isControllerPresent twice.  if you want to do your own processing, perhaps you should use getGamepadsByPrefix once instead?

src/components/daydream-controls.js
```diff
@@ -83,7 +81,14 @@ module.exports.Component = registerComponent('daydream-controls', {
...
+    handedness = handedness.filter(function (hand) {
+      return this.isControllerPresent(this.el.sceneEl, GAMEPAD_ID_PREFIX, {hand: hand});
+    }.bind(this));
+    var isPresent = handedness.length > 0;
+    if (handedness.length > 0) {
+      this.el.setAttribute('daydream-controls', { hand: handedness[0] });
+    }
```
machenmusik: (this will fail if there is one of each hand, but I don't think that is a real-world use case for Daydream at the moment.)

this will fail to ever allow a match if the daydream controller is ever missing hand property, not sure if that ever happens as real-world case at system level, but might happen in browser with broken support; my suggestion / request is to make it such that if no hand filter is requested, no hand filter is applied at all, just take what's there if anything (rather than trying with both hand filters separately and forcing one to apply)

src/components/daydream-controls.js
caseyyee: Nope.  I realize this, I haven't had time to address this yet.  As per our discussion, I think that `isControllerPresent` needs to still return true with empty hand, which it doesn't right now.

machenmusik: you can take a look at latest gearvr-controls PR changes, which work even for zero DOF with no hand if you specify that those controllers should be allowed (and so should also work for 3DOF that has no hand)

caseyyee: Awesome!  Doing that now.  Thanks

src/components/daydream-controls.js
dmarcos: Sorry I don't understand the problem with the model. In daydream you can only have one controller and the app cannot make any assumption about the hand since it's a user preference. The model you use should be hand agnostic. I would remove this property from the schema and reintroduce it later if there's a legitimate use case.

machenmusik: that is entirely not the case.  the system will have the daydream controller show up with hand 'left' or 'right', so of course apps can use a left-handed or right-handed model as they see fit?!?

machenmusik: (i.e. just because there is only one controller doesn't mean apps have to use hand-agnostic, symmetrical models, because the system is designed to tell you which hand that one controller represents, and we should reflect that when requested)

caseyyee:
made consistent with https://github.com/aframevr/aframe/pull/2545

src/components/daydream-controls.js
dmarcos: I still don't follow. Handness is a user preference and not something set by the application, hence it should not be an attribute of the component. We can emit an event whenever the controller is detected that passes the handedness for whoever needs that information from the outside.

dmarcos: I've been thinking about `daydream` a bit more. Until daydream supports multiple controllers we should make this component a `singleton` somehow. I propose that in the `play` method we do:
var daydreamEl = this.sceneEl.querySelector(['daydream-controls']);
````javascript
if (daydreamEl && daydreamEl.components['daydream-controls']) { 
  console.warn('Detected multiple instances of daydream-controls component. Only one controller is currently supported');
  return;
}
````

machenmusik: handedness - what you suggest is not how the other controllers work.  you say what you want matched and the right tracked controls is injected if present.  same here, except you can leave hand blank meaning either should match.

machenmusik: it doesn't matter how mny are specified s long as only one is matched, and even then i don't agree in all cases that only one should be allowed, since you may have more than one object that should be controlled by the daydream controller, just like oculus or vive or soon gearvr.

dmarcos:
Thanks everybody!

PR #2565, opened 2017-04-09
object property type (fixes #2555)

### LINKED ISSUE #2555: object property type
**Description:**

I had a game state component where I wanted to just store a plain JavaScript object in the game state, via the schema. Don't need any of the HTML parsing pieces, but we could have that too.

```
schema: {
  foo: {type: 'object'}
}

// ... setAttribute('component', 'foo', {bar: 10})
```

Possible HTML form just in case, but won't be used often:

```<a-entity component="foo: {bar: 10};>```

### PR DESCRIPTION
**Pull request by ngokevin**
**Description:**

Useful for like initial game state that have properties that are objects. Or allow accepting general JSON.

More useful when used with setAttribute to let the component accept and pass through objects:

```js
el.setAttribute('comp', 'objProp', {some: 'object', foo: {bar: 10}});
el.setAttribute('comp', 'objProp', jsonString);
```

Can do with HTML, but not something I'd use often.

```html
<a-entity comp="objProp: {foo: \"bar\", qaz: 10}">
```

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -13,6 +13,7 @@ registerPropertyType('asset', '', assetParse);
 registerPropertyType('boolean', false, boolParse);
 registerPropertyType('color', '#FFF', defaultParse, defaultStringify);
 registerPropertyType('int', 0, intParse);
+registerPropertyType('object', '', objectParse, JSON.stringify);
 registerPropertyType('number', 0, numberParse);
 registerPropertyType('map', '', assetParse);
 registerPropertyType('model', '', assetParse);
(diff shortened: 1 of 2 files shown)


dmarcos:
The problem with this is that yo will have to clone the object if you want to avoid surprises because you are coupling the internal state of the object with one that it's externally managed. People might also expect events or the component to be updated if the object changes. I see this as a can of worms. I'm not sure if it should be part of the core. We have not see much demand for it so far

dmarcos:
I see this to be easily abused and we bypasses all the schema goodnes we have

ngokevin:
OK, I can pass in inline parsers for now.

PR #2567, opened 2017-04-11
Robustness tweaks for tracked controls

machenmusik:
Robustness tweaks for tracked-controls
- if matching hand filter supplied to isControllerPresent, and gamepad provides no hand, use DEFAULT_HANDEDNESS for matching purposes
- defend against missing pose
- allow system-level specification of id prefixes that should always be included when present in controller list, e.g. if pose is sometimes missing

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -109,7 +109,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     // Scale offset by user height.
     deltaControllerPosition.multiplyScalar(userHeight);
     // Apply controller X and Y rotation (tilting up/down/left/right is usually moving the arm)
-    if (pose.orientation) {
+    if (pose && pose.orientation) {
       controllerQuaternion.fromArray(pose.orientation);
     } else {
       controllerQuaternion.copy(headObject3D.quaternion);
@@ -145,12 +145,12 @@ module.exports.Component = registerComponent('tracked-controls', {
     // Compose pose from Gamepad.
     pose = controller.pose;
     // If no orientation, use camera.
-    if (pose.orientation) {
+    if (pose && pose.orientation) {
       dolly.quaternion.fromArray(pose.orientation);
     } else {
       dolly.quaternion.copy(headObject3D.quaternion);
     }
-    if (pose.position) {
+    if (pose && pose.position) {
       dolly.position.fromArray(pose.position);
     } else {
       // The controller is not 6DOF, so apply arm model.
@@ -160,7 +160,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     dolly.updateMatrix();
 
     // Apply transforms, if 6DOF and in VR.
-    if (pose.position && vrDisplay) {
+    if (pose && pose.position && vrDisplay) {
       if (vrDisplay.stageParameters) {
         standingMatrix.fromArray(vrDisplay.stageParameters.sittingToStandingTransform);
         dolly.applyMatrix(standingMatrix);
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -7,6 +7,7 @@ var utils = require('../utils');
  * It maintains a list with the available tracked controllers
  */
 module.exports.System = registerSystem('tracked-controls', {
+  schema: {alwaysAllowIdPrefixes: {type: 'array'}},
   init: function () {
     var self = this;
     this.controllers = [];
@@ -26,7 +27,12 @@ module.exports.System = registerSystem('tracked-controls', {
     var gamepads = trackedControlsUtils.getGamepadsByPrefix('');
     for (var i = 0; i < gamepads.length; i++) {
       var gamepad = gamepads[i];
-      if (gamepad && gamepad.pose) { controllers.push(gamepad); }
+      if (gamepad && gamepad.pose) { controllers.push(gamepad); } else
+      if (gamepad && this.data.alwaysAllowIdPrefixes) {
+        this.data.alwaysAllowIdPrefixes.forEach(function (idPrefix) {
+          if (gamepad.id.indexOf(idPrefix) === 0) { controllers.push(gamepad); }
+        });
+      }
     }
   },
 
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -1,3 +1,4 @@
+var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
 /**
  * Return enumerated gamepads matching id prefix.
  *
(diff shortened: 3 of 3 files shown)


src/components/tracked-controls.js
```diff
@@ -145,12 +145,12 @@ module.exports.Component = registerComponent('tracked-controls', {
     // Compose pose from Gamepad.
     pose = controller.pose;
     // If no orientation, use camera.
-    if (pose.orientation) {
+    if (pose && pose.orientation) {
```
dmarcos: Is it possible / correct for a pose to not have orientation? We may want to let it break to identify underlying issues instead of graceful handle the bogus case.

src/systems/tracked-controls.js
```diff
@@ -7,6 +7,7 @@ var utils = require('../utils');
  * It maintains a list with the available tracked controllers
  */
 module.exports.System = registerSystem('tracked-controls', {
+  schema: {alwaysAllowIdPrefixes: {type: 'array'}},
```
dmarcos: What is this for?

src/components/tracked-controls.js
machenmusik: it would be possible to have a controller with position but not orientation, so guard for absence of either

src/systems/tracked-controls.js
machenmusik: originally for controllers that may have initial instability in gamepad api responses (classic example is when vive controllers first turn on, hand may be indeterminate, pose may be ill defined until confirmed within tracking volume)

dmarcos: Is this a problem in both Chromium and Firefox, the WebVR API implementation, the spec all of the above? I think the gamepad API should always return a pose if it corresponds to a tracked device. Do you have a consistent way to reproduce the issue. I can open a bug on Firefox / Gecko

dmarcos: I've been told that Firefox should already have flags to check for the existence of rotation/position via a flag on the Gamepad handler: `gamepad.hasOrientation / gamepad.hasPosition`

dmarcos: We should use those instead of the `alwaysAllowIdPrefixes` property

dmarcos: I've been told that `gamepad.pose` should be always available and we can check for `gamepad.pose.hasOrientation` or `gamepad.pose.hasPosition`

dmarcos: It this is not the case then there's a bug in the browser :)

machenmusik: according to https://developer.mozilla.org/en-US/docs/Web/API/Gamepad
pose doesn't even exist, so i think some defensiveness may be appropriate

machenmusik: sidebar: i think our current list of browsers that ideally should be supported is...

Firefox/Nightly (desktop; do you plan to support mobile?)
Chrome/Chromium (desktop and at least Daydream mobile)
Oculus Browser (Carmel is deprecated)
Samsung Internet

agree?  so Firefox behavior alone may not suffice, both Chromium and Firefox is probably good enough in most cases since at present Oculus and Samsung browsers generally follow...

machenmusik: (i have not been able to reproduce consistently, to answer prior question)

src/components/tracked-controls.js
machenmusik: the additional guard is to handle missing / null pose

mkeblx: I don't think can have position w/o orientation, pretty sure removing 'hasOrientation' from WebVR spec for that reason

src/systems/tracked-controls.js
dmarcos: the `pose` is not part of the gamepad spec. It's an extension specific to VR. All tracked controllers should have a pose attribute even if that pose is not valid yet. for that we can check the hasOrientation / hasPosition flags

mkeblx: See https://bugs.chromium.org/p/chromium/issues/detail?id=706561 which removes `hasOrientation` among other things. For `VRDisplay`, but will do likely do same for Gamepad Extensions so do need to rely on that flag? At least `hasPosition` false case might be better

machenmusik: so unless we think there will be an orientation and/or position even when hasXXX is false (if they even continue to exist) why would we not stick with looking at the actual variable we intend to use?  it seems clear to me in these last few comments that there is a loose convention but not actual specification, so I think it makes sense to be slightly defensive... at least until there is an actual spec agreed and implemented by browsers we intent to support (see above proposal, and I would also add Edge if actual HMDs were available)

src/components/tracked-controls.js
machenmusik: from slack...

```example: it is perfectly legal for the user to have a rift and paired touches, but the touches are not currently powered up.  ideally that might appear as gamepad present with usual touch id, but connected false and (i would argue) undefined or null pose.

so that the app can show a message "please turn on your touch controllers"```

File: src/systems/tracked-controls.js (hunk 2 above)
Comment (dmarcos) [inline src/systems/tracked-controls.js]: There is a spec https://w3c.github.io/gamepad/extensions.html#gamepadpose-interface

Comment (dmarcos) [inline src/systems/tracked-controls.js]: I want to minimize the amount of hacks and work with browser vendors to get these things right

Comment (machenmusik) [inline src/systems/tracked-controls.js]: reading that extension, it says that in some cases position and orientation may/should be null, so the guards seem appropriate?

File: src/components/tracked-controls.js (hunk 1 above)
Comment (machenmusik) [inline src/components/tracked-controls.js]: reading that extension, it says that in some cases position and orientation may/should be null, also pose itself, so the guards seem appropriate?

File: src/systems/tracked-controls.js (hunk 2 above)
Comment (dmarcos) [inline src/systems/tracked-controls.js]: The guards are ok but I think we can get rid of the attribute since if the gamepad is pose capable it should always be available

File: src/components/tracked-controls.js [hunk 3]
```diff
@@ -160,7 +160,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     dolly.updateMatrix();
 
     // Apply transforms, if 6DOF and in VR.
-    if (pose.position && vrDisplay) {
+    if (pose && pose.position && vrDisplay) {
```
dmarcos: pose should be always available

dmarcos:
After checking the spec I don't think we need any of the guards in this PR since if a Gamepad is pose capable the pose attribute should always be available.

machenmusik:
the line in question:
`If the hardware cannot supply any pose values, MUST be set to null.`
You and I read that differently.
For example, if the controller is paired but currently disconnected, then the hardware cannot supply any values, and pose must be null.  Even where gamepad must have worked at one point, this can happen if your batteries run out.  The guards are to ensure that A-Frame survives cases where pose is null even if only temporarily, rather than throwing errors that prevent correct processing.

machenmusik:
if 6DOF controllers leave the tracking volume / frustrum,
would they have null pose?
or would.they have orientation (from accelerometers), but null position?

dmarcos:
I think it means that the `pose` attribute is always present if the hardware is `capable` of providing poses (not all gamepads provide poses) regardless of the state of the controller.

machenmusik:
would be good to understand what browsers actually do including previously mentioned exceptional cases.
if all browsers interpret as you say, i don't know how some of these conditions are detectable (e.g. if there must always be a pose, then what pose values should be provided while disconnected?)

dmarcos:
If pose is not defined for a tracked gamepas in any of the browsers we should file a bug against.

PR #2568, opened 2017-04-11
[WIP] fix hand controls blend character animations (fixes #2445)

### LINKED ISSUE #2445: Animation of hand-controls not playing
Some times it plays but it does not play the animation in reverse. Sometimes they get stuck in a pose. Probable causes: Models update, three.js update (maybe the animation API has changed)...

ngokevin:
**Description:**

Fix all hand animations to work. Including cross-fade gestures.

**Changes proposed:**
- Change `.animateGesture` to pass in current *and* last gesture. If there was a gesture and now there is *no* gesture, then configure `.playAnimation` to reverse the last gesture (`.playAnimation('fist', 'fist', true)`). If we're going from gesture to gesture then it'd pass in `.playAnimation('fist', 'pointing', false)`.
- Fix `.playAnimation` to play the animation in reverse by setting `clipAction.loop = THREE.PingPong` and setting `clipAction.repetitions` to 0.
- [x] Crossfade from gesture-to-gesture without returning to the default pose. **TODO**. Currently, the mixer will stop all current clip actions which restores to open pose an *then* it'll go to the next clip. Need to crossfade it smoothly. **UPDATE**: I was able to crossfade smoothly by stopping all action, setting a low weight on the previous gesture, playing both previous/current clip action, then calling clip actipn's `crossFadeTo`.

**Clean ups:**
- Remove the animation mappings which was causing another layer of indirection of strings to strings. Just refer to the model's animation strings.
- Move model URLs into an object keyed by hand, removes LOC from `update()`
- Add more comments
- Move some stuff from a method to a function that doesn't need to be a method

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -1,27 +1,58 @@
+/* global THREE */
 var registerComponent = require('../core/component').registerComponent;
 
-var OCULUS_LEFT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/v2/leftHand.json';
-var OCULUS_RIGHT_HAND_MODEL_URL = 'https://cdn.aframe.io/controllers/oculus-hands/v2/rightHand.json';
+// Found at https://github.com/aframevr/assets.
+var MODEL_URLS = {
+  left: 'https://cdn.aframe.io/controllers/oculus-hands/v2/leftHand.json',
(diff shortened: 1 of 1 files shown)


ngokevin:
Now ready for review. All gestures now animation.

Few magic numbers in there, but I can make variables for those.

Issue #2535, opened 2017-03-30
request: triangle primitive


vrimaginar:
this is a request for a new primitive, the triangle. It can be used easily in scenes with houses in it.

ngokevin:
Thanks. There are ways to do triangles currently (using the existing geometry and reducing the segments), but I guess a triangle would look like passing three vertices.

ngokevin:
Fixed!

PR #2573, opened 2017-04-11
add triangle geometry (#2535)

andyli:
This adds a triangle geometry as discussed in #2535.
It is specified by 3 vertices, `vertexA`, `vertexB`, and `vertexC`.

If there is nothing need to amend regarding to property names, default values etc., I will send a PR to add docs.


ngokevin:
Most looks good, do you think it'd be better to center the triangle by default?

andyli:
Yes, I think it make sense since the other geometries are centered by default.
I have just added a commit for that.

ngokevin:
Thanks! We can add docs, just a section in the `geometry.md` file works fine

machenmusik:
vertexA/B/C seems a little verbose.  i take it there is some problem preventing an array of vec3?  ig so, maybe we should change that - I am thinking something like vertices or verts as vec3 array would be better, and that would give use plumbing to support strips and other more advanced primitives?

andyli:
@machenmusik 
I've considered using an array of vec3 also, but I'm not sure. JS array doesn't have fixed length, so either we need additional check on array length during input, or have additional specification on how to handle longer/shorter array. I think using 3 (or even 4) vec3 variables instead of using array is still acceptable.

Building strip or other geometries of course is a different case, and I agree array would be more suitable when the vertex input amount is varied or very large.

Let me know if I've missed anything. :)

machenmusik:
for just basic triangle this is probably fine.  my worry is that people will try to make thousands of these because they can't do strips and then wonder why performance isn't what it could be.  so i naturally jump to syntax that also would work for strips etc.

andyli:
> make thousands of these because they can't do strips

People probably will, as well as thousands of boxes and spheres too ;)
I guess that means we have to add strip and other more advanced geometries...

For now, I will leave it as `vertexA`, `vertexB`, and `vertexC`. Documentation is added. Also found out my initial implementation has no UVs, so added some code to compute a sensible set.

andyli:
Rebased on top of master to trigger a TravisCI build.

andyli:
Bump ;)

ngokevin:
Thanks for the wait. I've updated the PR to not have to recreate the quaternion and vectors on each instance. Waiting for tests to pass then will mergw!

PR #2576, opened 2017-04-12
Allow class selector

machenmusik:
Using selectors for images / videos / canvas only works when the selector is by id (starts with `#`) although the selector syntax is more general.  This PR allows class selector (e.g. `.main-video`) to work as well.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -127,7 +127,7 @@ function deepEqual (a, b) {
   var valB;
 
   // If not objects or arrays, compare as values.
-  if (a === null || b === null ||
+  if (a === null || b === null || a === undefined || b === undefined ||
       !(a && b && (a.constructor === Object && b.constructor === Object) ||
                   (a.constructor === Array && b.constructor === Array))) {
     return a === b;
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -80,7 +80,7 @@ function assetParse (value) {
   if (parsedUrl) { return parsedUrl[1]; }
 
   // ID.
-  if (value.charAt(0) === '#') {
+  if ('#.'.indexOf(value.charAt(0)) >= 0) {
     el = selectorParse(value);
     if (el) {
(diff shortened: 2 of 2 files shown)


machenmusik:
working example (using for with built version of PR)
http://codepen.io/machenmusik/pen/aWbpYb

src/utils/index.js
```diff
@@ -127,7 +127,7 @@ function deepEqual (a, b) {
   var valB;
 
   // If not objects or arrays, compare as values.
-  if (a === null || b === null ||
+  if (a === null || b === null || a === undefined || b === undefined ||
       !(a && b && (a.constructor === Object && b.constructor === Object) ||
                   (a.constructor === Array && b.constructor === Array))) {
     return a === b;
```
donmccurdy: This whole block now simplifies to:

```js
if (!a || !b
    || (a.constructor === Object && b.constructor === Object)
    || (a.constructor === Array && b.constructor === Array))) {
  return a === b;
}
```

ngokevin: boolean wizardry

ngokevin:
Should we allow this? Can we prescribe that assets should have IDs because they are unique? If you have more than one of the class, then it's ambiguous which asset.

dmarcos:
Yes, I think assets should be identified uniquely by id.

machenmusik:
unfortunately, some assets don't work by id -- e.g. `<video id='1'>` will fail to be accessible via selector `#123`, Firefox / Nightly will throw errors.

machenmusik:
the standard querySelector behavior is to allow class and only return first (which may be only).  seems to me that aframe should respect browser implementation decision to allow and just pass it through

dmarcos:
I would say that `id=1` is a bad id and also ill formed. ids have to start with a letter:
https://www.w3.org/TR/html4/types.html#type-id

machenmusik:
abandoning this approach per discussion, in favor of better documented id selector or string requirement (and using getElementByID to resolve id)

PR #2580, opened 2017-04-13
Add gyro

jason90929:
if gyro is disabled(default is enabled), user is only drag to rotate camera, user can't use mobile auto move, like this video demo
it is used for mobile device
https://youtu.be/Yysc7OlDbzw


dmarcos:
Thanks! Can you remove the changes from the dist directory?

jason90929:
alright, I'll resend a pull request

PR #2583, opened 2017-04-13
Support for vive-tracker

machenmusik:
Support for vive-tracker.

Handles quirk in Nightly where gamepad id strings are same for both controllers and trackers (either all `OpenVR Gamepad` or `OpenVR Tracker`), but hand properties are correct.

At the moment, applies a rotation correction to what gamepad API returns to make it behave more like Vive controller / natural expectations -- specifically, the rotation correction makes it so that the controller pose aligns with the corrected pose if you take the tracker, put it on top of the touchpad, and rotate so the LED-bearing protrusion points toward the strap / bottom of the handle -- but this may be removed pending discussions.

The rotation correction can be avoided by explicitly specifying zero rotation, e.g. `<a-entity vive-tracker="index:0; rotation:0 0 0">`

Uses tracker model from https://github.com/aframevr/assets/pull/22 with corrected pose orientation.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -1,5 +1,6 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var DEG2RAD = require('../lib/three').Math.DEG2RAD;
 var DEFAULT_CAMERA_HEIGHT = require('../constants').DEFAULT_CAMERA_HEIGHT;
 
 var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
@@ -20,9 +21,11 @@ var FOREARM = {x: 0, y: 0, z: -0.175};
 module.exports.Component = registerComponent('tracked-controls', {
   schema: {
     controller: {default: 0},
+    hand: {type: 'string', default: 'any-or-none'},
     id: {type: 'string', default: ''},
     idPrefix: {type: 'string', default: ''},
     rotationOffset: {default: 0},
+    rotation: {type: 'vec3'},
     // Arm model parameters when not 6DoF.
     armModel: {default: true},
     headElement: {type: 'selector'}
@@ -31,6 +34,7 @@ module.exports.Component = registerComponent('tracked-controls', {
   init: function () {
     this.axis = [0, 0, 0];
     this.buttonStates = {};
+    this.rotationOffset = new THREE.Quaternion();
 
     this.dolly = new THREE.Object3D();
     this.controllerEuler = new THREE.Euler();
@@ -44,6 +48,13 @@ module.exports.Component = registerComponent('tracked-controls', {
 
     this.previousControllerPosition = new THREE.Vector3();
     this.updateGamepad();
+  },
+
+  update: function (oldData) {
+    var data = this.data;
+    if (data.rotation === oldData.rotation) { return; }
+    this.controllerEuler.set(DEG2RAD * data.rotation.x, DEG2RAD * data.rotation.y, DEG2RAD * data.rotation.z);
+    this.rotationOffset.setFromEuler(this.controllerEuler);
   },
 
   tick: function (time, delta) {
@@ -80,8 +91,10 @@ module.exports.Component = registerComponent('tracked-controls', {
 
     // Hand IDs: 0 is right, 1 is left.
     for (i = 0; i < controllers.length; i++) {
-      if ((data.idPrefix && controllers[i].id.indexOf(data.idPrefix) === 0) ||
-          (!data.idPrefix && controllers[i].id === data.id)) {
+      if (((data.idPrefix && controllers[i].id.indexOf(data.idPrefix) === 0) ||
+           (!data.idPrefix && controllers[i].id === data.id)) &&
+          ((data.hand === 'any-or-none') ||
+           ((controllers[i].hand === 'undefined' ? DEFAULT_HANDEDNESS : controllers[i].hand) === (data.hand === 'none' ? '' : data.hand)))) {
         matchCount++;
         if (matchCount - 1 === data.controller) {
           this.controller = controllers[i];
@@ -150,6 +163,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     var controller = this.controller;
     var controllerEuler = this.controllerEuler;
     var controllerPosition = this.controllerPosition;
+    var controllerQuaternion = this.controllerQuaternion;
     var currentPosition;
     var deltaControllerPosition = this.deltaControllerPosition;
     var dolly = this.dolly;
@@ -196,7 +210,9 @@ module.exports.Component = registerComponent('tracked-controls', {
     }
 
     // Decompose.
-    controllerEuler.setFromRotationMatrix(dolly.matrix);
+    controllerQuaternion.setFromRotationMatrix(dolly.matrix);
+    controllerQuaternion.multiply(this.rotationOffset);
+    controllerEuler.setFromQuaternion(controllerQuaternion);
     controllerPosition.setFromMatrixPosition(dolly.matrix);
 
     // Apply rotation (as absolute, with rotation offset).
--- a/docs/components/vive-tracker.md
+++ b/docs/components/vive-tracker.md
@@ -1,0 +1,58 @@
+---
+title: vive-tracker
+type: components
+layout: docs
+parent_section: components
+---
+
+[trackedcontrols]: ./tracked-controls.md
+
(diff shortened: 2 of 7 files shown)


machenmusik:
per separate discussion, rather than using getGamepadsByPrefix, need to use controllers list from tracked-controls system

machenmusik:
(what is this new failing node test about?)

HAZARDU5:
Would be keen to see this as we've got a few demos we'd like to try building utilizing the Vive trackers.

machenmusik:
@kfarr can correct me if I am wrong but at present there seems to be one functional and one test issue. 

The functional issue is that the observed position (with the default orientation correction) appears to have a small Z offset when compared to collocated entities and the SteamVR mode, which may be a combination of the rotation and the natural pivot position of the tracker model.

The test issue is that somehow only the node.js tests fail with a rather cryptic message.

But the PR should otherwise be functional for experimentation with Nightly.

kfarr:
Yes this is definitely functional as is. The offset is not a big deal and honestly the controllers also exhibit "drift" between Firefox and the SteamVR interface, so I'm not sure it's anything specifically related to the trackers. (You can see this for yourself by pressing the system menu button rapidly on the touch controllers to switch between the Firefox output and the SteamVR system menu.)

cvan:
@machenmusik just wanted to check because I didn't see any labels, do you need a review on this?

dmarcos:
@cvan it's marked as Work in Progress ([WIP]) in the title of the PR

cvan:
yeah, just wanted to see if he needed a hand

machenmusik:
@cvan I think it's actually good except the weird failing node test (and maybe if we want to fix the default pose z-offset, but said to not be a big deal)

machenmusik:
@cvan test now passes, so removing [WIP] from title... what more did we want to do this one before merging?  maybe a good example or two (which is where the Z-offset from what SteamVR shows is annoying, since otherwise we just need the most basic usage for that)

machenmusik:
in looking more closely at the Z-offset:

go to this codepen:
https://codepen.io/machenmusik/pen/XRJmWB

if you look at the dummy tracker and marker, 
```  <!-- nothing wrong with the model, this looks right -->
  <a-entity position="0 1 -1" obj-model="obj:https://cdn.aframe.io/controllers/vive/vr_tracker-vive.obj; mtl:https://cdn.aframe.io/controllers/vive/vr_tracker-vive.mtl">
   <a-entity tracker-marker="?"></a-entity>
  </a-entity>```
the model is aligned correctly with the center of the marker and rays
(no position from vive-tracker and tracked-controls at all)

if you look at the first real tracker (labeled 0),
which gets its pose from vive-tracker,
and then bring over the vive wand and touch the marker's center as a sanity check,
you can see that the entity itself has inherited the right pose 
(as shown by the marker and rays, agreeing with SteamVR)

but the tracker model doesn't appear in the same place as either the marker or rays (or SteamVR) -- that's the Z offset issue.

if you have a second tracker and look at the marker labeled 1,
you can see that the Z-offset issue is the same even without the rotation correction
(which is needed to make things match SteamVR, BTW)

thoughts / suggestions?

**Comment by dmarcos** [comment]
I have to spend time looking so it won't make it for 0.6.0

**Comment by machenmusik** [comment]
No problems there.  If the Nightly bug with wrong gamepad id is gone, I may be able to (squash and) simplify changes to make less invasive?  Which might have better chance of merge regardless of when.  Or did we want tracked-controls to understand how to filter by hand anyway, and those changes should remain?

**Comment by machenmusik** [comment]
duh.  i just realized that the Z offset is probably whatever adjustments were being made with old vive controller model.  pushed change to PR, but can't check actual outcome -- can anyone check https://codepen.io/machenmusik/pen/XRJmWB to see if offset is gone?  @kfarr @dmarcos?

**Comment by machenmusik** [comment]
if the offset is gone, maybe we can land in 0.6.0 after all (well, from a functional standpoint anyway, still maybe no time for @dmarcos @ngokevin to look)

File: src/components/tracked-controls.js [hunk 1]
```diff
@@ -19,7 +19,7 @@ var FOREARM = {x: 0, y: 0, z: -0.175}; // vector from eyes to elbow (divided by
 module.exports.Component = registerComponent('tracked-controls', {
   schema: {
     controller: {default: 0},
-    hand: {default: undefined},
+    hand: {type: 'string', default: 'any-or-none'},
```
cvan: nice, thanks - do the Docs need to be updated?

machenmusik:
confirmed, offset is gone (internally; some very slight difference from SteamVR).

machenmusik:
@cvan I think we can remove the work-in-progress label now (I would but I can't)

machenmusik:
(I may end up doing a simpler version that does not handle the aforementioned Nightly quirk, which may simplify things and avoid some conflicts with latest master from other merged changes.)

src/components/tracked-controls.js
machenmusik: should be now

machenmusik:
Ok so revised to better fit against latest master.  Any issues?  @kfarr is this better now w.r.t. the mystery Z offset we used to encounter?

machenmusik:
(May need to rebase and fix a bit after current round of optimizations in master)

machenmusik:
Rebased, and starkly basic but fully functional example added and tested against Firefox release 55 (64-bit).

machenmusik:
@ngokevin why am I getting this test failure? seems that a recent commit may have poisoned the well...
```
[error]: docs/introduction/index.md
    Link definition does not exist: [Mozilla][mozilla]
```

machenmusik:
I think this one is in good shape now, if we feel comfortable enough merging, then can use `rotation` to address #2965 at least temporarily

src/components/tracked-controls.js
```diff
@@ -80,8 +91,10 @@ module.exports.Component = registerComponent('tracked-controls', {
 
     // Hand IDs: 0 is right, 1 is left.
     for (i = 0; i < controllers.length; i++) {
-      if ((data.idPrefix && controllers[i].id.indexOf(data.idPrefix) === 0) ||
-          (!data.idPrefix && controllers[i].id === data.id)) {
+      if (((data.idPrefix && controllers[i].id.indexOf(data.idPrefix) === 0) ||
+           (!data.idPrefix && controllers[i].id === data.id)) &&
+          ((data.hand === 'any-or-none') ||
+           ((controllers[i].hand === 'undefined' ? DEFAULT_HANDEDNESS : controllers[i].hand) === (data.hand === 'none' ? '' : data.hand)))) {
```
leweaver: Is this a typo or intentional? In src/utils/tracked-controls.js  you are testing against: `controllers[i].hand === undefined`

machenmusik:
the vive-tracker controller object hand property is undefined (which makes
sense because the tracker has no handedness)

machenmusik:
@dmarcos you were advancing an alternate with your spectator camera work, should this be closed?

cvan:
@machenmusik: do you mean @dmarcos’ PR #3152?

machenmusik:
No I don't think it's a PR yet

NellWaliczek:
If there's a discussion happening around spectator cameras, I'd love to read it :)

machenmusik:
@NellWaliczek I haven't seen what @dmarcos is cooking up yet.  A Glitch version of an old codepen I did is here https://glitch.com/edit/#!/aframe-spectator-camera

cvan:
is this blocked on @dmarcos’ work? if so, is there a branch we can test against? excited to try this out - thank you

machenmusik:
Was working perfectly, @kfarr and I were playing with it... not blocked AFAIK, but was sitting around working for months and never got merged, then @dmarcos said he was going to redo entirely as part of spectator camera work so I gave up on this one.

machenmusik:
Rather than fix the new conflict, I'd like to close this because @dmarcos has said that a separate PR will be done along with some spectator camera functionality.  But looking back here, I don't see either the new PR yet (and looks like @NellWaliczek is interested here too) or @dmarcos explicitly confirming...

Can core team confirm this is deprecated now so I can close?  Or close it yourselves?
Or if that is deferred for some reason, and we do actually want to merge this, let me know

ngokevin:
Yeah, @dmarcos has a patch. We'll have him to make a WIP PR so it's visible. I've seen his PR work with spectator camera before so at least it's not imaginary.

dmarcos:
Yeah I had a PR. I will share it short after the VREffect / VRControls PR lands. Thanks for the patience.

PR #2585, opened 2017-04-14
add Gyro

jason90929:
if gyro is disabled(default is enabled), user is only drag to rotate camera, user can't use mobile auto move, like this video demo
it is used for mobile device
https://youtu.be/Yysc7OlDbzw

you can use it and test it on mobile
```
<a-camera gyro-enabled="false"></a-camera>
```

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -14,6 +14,7 @@ module.exports.Component = registerComponent('look-controls', {
     enabled: {default: true},
     hmdEnabled: {default: true},
     reverseMouseDrag: {default: false},
+    gyroEnabled: {default: true},
     standing: {default: true}
   },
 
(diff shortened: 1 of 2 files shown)


dmarcos:
What happens if the gyro is reenabled? I think probably the orientation would have to be reset so it plays well with mouse panning.

ngokevin:
I would leave this out of the default set and recommend tweaking your own controls (taking look-controls and removing the VRControls bits).

ngokevin:
I propose to close ^

dmarcos:
Yeah! The more toggles we have in the core components the more combinations we have to keep track of and QA. It's not hard for someone to copy and customize the controls.

andreyrd:
> It's not hard for someone to copy and customize the controls.

It is hard if you can't use webpack / similar build systems. I have to either set up a custom build of aframe or rewrite and paste in the dependencies for `look-controls.js`.

IMO `gyroEnabled` is a simple enough toggle that I feel it should be included, especially when toggles for `hmdEnabled` and `touchEnabled` exist. It does not need to be on the primitive, but should at least be on the component.

andreyrd:
That said, it seems like `hmdEnabled` doesn't actually do anything? From what I understand by looking through the code for `look-controls`, gryo data is written to `hmdEuler`, so it seems logical that if `hmdEnabled == false`, it should ignore it.

At the moment all that `hmdEnabled` seems to do is reset pitch and yaw, but only once in `update()`:
```js
    // Reset pitch and yaw if disabling HMD.
    if (oldData && !data.hmdEnabled && !oldData.hmdEnabled) {
      this.pitchObject.rotation.set(0, 0, 0);
      this.yawObject.rotation.set(0, 0, 0);
    }
```

PR #2588, opened 2017-04-14
Support gearvr-controls 3DOF like #2545, after 2513

machenmusik:
This PR is to support the new Gear VR Controller, like #2545, but including changes to bring it in line with the new axismove changes in #2513.  As such, it includes both #2513 and #2545 in their entirety.


machenmusik:
I don't think we need this anymore, the salient aspects are now merged into #2513

PR #2590, opened 2017-04-14
Add Relative section to scale doc

andreasplesch:
**Description:**

Adds a Relative scale section
to be consistent with the rotation and position component docs.

**Changes proposed:**
- mention concisely that scales nest and are part of the transform hierarchy


andreasplesch:
Thanks. Feel free to rephrase, of course.

Issue #1416, opened 2016-04-28
HLS m3u8 video stream does not work with a-videosphere Safari 9 and iOS 9 mobile Safari


liuxiaoyue:
hello！ I met the same problem， same issues 1021 (https://github.com/aframevr/aframe/issues/1021),
i think, this is not a limitation of iOS,i find a example that can play, but, Internal testing products，so i can't public sorry! I try to modify aframe texture, use THREE.Texture, end up in faliure! Can be modified？ thanks！

dmarcos:
iOS might be able to play m3u8. The problem is using that video stream in the context of a WebGL texture. Please reopen If you can provide a working example.

liuxiaoyue:
This issue has been resolved，thanks！

invaderhd:
@liuxiaoyue did you find a solution? Can you share it

radenui:
Seems like this issue has been resoved (or worked around) in other contexts: 
- For [THREE](https://github.com/mrdoob/three.js/issues/9754#issuecomment-268186791)
- For [GoogleVr](https://github.com/googlevr/vrview/pull/105)

I couldn't figure out yet how to build the required custom material to handle such a video format in a-frame, but I think the hard work has been done already.

dmarcos:
@radenui thanks for letting us know

machenmusik:
so how does one know when bgra fix is required?  

wondering whether it makes more sense to enable detection and remediation in core/default shaders (so if similar problem arises for different elements, can set bgra property or similar)

radenui:
Seems like there are actually 2 problems:
- FlipY  on OSX and iOS (fixed by `vUV = vec2( uv.x, 1.0 - uv.y );`),
- Channel R and B mixed on iOS (fixed by `gl_FragColor = texture2D( texture, vUV  ).bgra;`) [issue comment here](https://github.com/mrdoob/three.js/issues/9754#issuecomment-266309206)

machenmusik:
are you sure? `FlipY disabling in my tests was required for MacOs Safari 10 also.`
https://github.com/mrdoob/three.js/issues/9754#issuecomment-273515995

machenmusik:
(need to understand also if this is just for video, or also for image, canvas, etc.)

robinnorth:
[...]

PR #2597, opened 2017-04-18
make accommodations necessary for iOS 10 HLS

machenmusik:
Provides core support needed for iOS 10 HLS playback discussed in #1416.

Note that this leaves video flipped in Y dimension and also with B and R color channels reversed.
It is expected that corrections can be applied using custom shaders.
(Accommodating those corrections using the built-in shaders is outside the scope of this PR.)

### DIFF
--- a/src/shaders/ios10hls.js
+++ b/src/shaders/ios10hls.js
@@ -1,7 +1,8 @@
 var registerShader = require('../core/shader').registerShader;
 
 /**
- * Custom shader for iOS 10 HLS.
+ * Custom shader for iOS 10 HTTP Live Streaming (HLS).
+ * For more information on HLS, see https://datatracker.ietf.org/doc/draft-pantos-http-live-streaming/
  */
 module.exports.Shader = registerShader('ios10hls', {
   schema: {
--- a/src/shaders/ios10hls.js
+++ b/src/shaders/ios10hls.js
@@ -1,0 +1,32 @@
+var registerShader = require('../core/shader').registerShader;
+
+/**
+ * Custom shader for iOS 10 HLS.
+ */
+module.exports.Shader = registerShader('ios10hls', {
+  schema: {
+    src: {type: 'map', is: 'uniform'},
+    opacity: {type: 'number', is: 'uniform', default: 1}
+  },
+
+  vertexShader: [
+    'varying vec2 vUV;',
+    'void main(void) {',
+    '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
+    '  vUV = uv;',
+    '}'
+  ].join('\n'),
+
+  fragmentShader: [
+    'uniform sampler2D src;',
+    'uniform float opacity;',
+    'varying vec2 vUV;',
+    'void main() {',
+    '  vec2 offset = vec2(0, 0);',
+    '  vec2 repeat = vec2(1, 1);',
+    '  vec4 color = texture2D(src, vec2(vUV.x / repeat.x + offset.x, (1.0 - vUV.y) / repeat.y + offset.y)).bgra;',
+    '  gl_FragColor = vec4(color.rgb, opacity);',
+    '}'
+  ].join('\n')
+});
+
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -206,8 +206,8 @@ module.exports.System = registerSystem('material', {
 });
 
 function isHLS (videoEl) {
-  if (videoEl.type === 'application/x-mpegurl') return true;
-  if (videoEl.src && videoEl.src.indexOf('.m3u8') > 0) return true;
+  if (videoEl.type.toLowerCase() === 'application/x-mpegurl') { return true; }
+  if (videoEl.src && videoEl.src.toLowerCase().indexOf('.m3u8') > 0) { return true; }
   return false;
 }
 
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -144,6 +144,16 @@ module.exports.System = registerSystem('material', {
     texture.minFilter = THREE.LinearFilter;
     setTextureProperties(texture, data);
 
+    // if we're on iOS, and the video is HLS, we currently need to do some hacks
+    if (utils.device.isIOS() && isHLS(videoEl)) {
+      // really it's BGRA, so this needs correction in shader
+      texture.format = THREE.RGBAFormat;
+      texture.needsCorrectionBGRA = true;
+      // apparently this is needed for HLS, so this needs correction in shader
+      texture.flipY = false;
+      texture.needsCorrectionFlipY = true;
+    }
+
     // Cache as promise to be consistent with image texture caching.
     videoTextureResult = {texture: texture, videoEl: videoEl};
     textureCache[hash] = Promise.resolve(videoTextureResult);
@@ -194,6 +204,12 @@ module.exports.System = registerSystem('material', {
     });
   }
 });
+
+function isHLS (videoEl) {
+  if (videoEl.type === 'application/x-mpegurl') return true;
+  if (videoEl.src && videoEl.src.indexOf('.m3u8') > 0) return true;
+  return false;
+}
 
 /**
  * Calculates consistent hash from a video element using its attributes.
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -1,6 +1,7 @@
 var registerSystem = require('../core/system').registerSystem;
 var THREE = require('../lib/three');
 var utils = require('../utils/');
+var isHLS = require('../utils/material').isHLS;
 
 var debug = utils.debug;
 var error = debug('components:texture:error');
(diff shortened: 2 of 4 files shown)


machenmusik:
(Due to the current Mobile Safari CORS issue, this is a bit of a PITA to diagnose / address / test as the A-Frame page and the HLS content must have the same domain.)

dmarcos:
So with PR the video at least plays but flipped and with wrong color? :)

dmarcos:
Is that better than the video not playing at all?

machenmusik:
> Is that better than the video not playing at all?
Without this change, there is no way to make it work at all.
With this change, it is actually simple to use custom shader to fix flip and channel order.

So yes, this change is much better than the video not playing at all... see discussion in #1416

dmarcos:
Can we add add a custom shader that it's set automatically when playing on iOS?

machenmusik:
shader added but I haven't got it being set automatically yet

machenmusik:
ok, the switching could probably use a little better guard (e.g. auto-switch if it's a shader like standard or flat we know won't work) if you wanted to touch up the PR

src/systems/material.js
```diff
@@ -144,6 +144,16 @@ module.exports.System = registerSystem('material', {
     texture.minFilter = THREE.LinearFilter;
     setTextureProperties(texture, data);
 
+    // if we're on iOS, and the video is HLS, we currently need to do some hacks
+    if (utils.device.isIOS() && isHLS(videoEl)) {
+      // really it's BGRA, so this needs correction in shader
+      texture.format = THREE.RGBAFormat;
+      texture.needsCorrectionBGRA = true;
```
dmarcos: Can we add this to the `this` instead of the texture?

src/systems/material.js
```diff
@@ -144,6 +144,16 @@ module.exports.System = registerSystem('material', {
...
+      // apparently this is needed for HLS, so this needs correction in shader
+      texture.flipY = false;
+      texture.needsCorrectionFlipY = true;
```
dmarcos: Can we add this to the this instead of the texture?

machenmusik: (see above)

src/systems/material.js
machenmusik: I don't think so - `this` is the system, not a given video texture, right?  Complicating things is the fact that iOS 10 MP4 playback does not need any of these workarounds, so needs to be done on a per-video-texture basis.

machenmusik:
due to CORS issues with Mobile Safari, in order to test this, your HTML page must be in the same exact domain as the IDevice-compatible video content (m3u8 manifests and ts / aac files),  Subdomains will not work at present.

cvan:
> due to CORS issues with Mobile Safari, in order to test this, your HTML page must be in the same exact domain as the IDevice-compatible video content (m3u8 manifests and ts / aac files), Subdomains will not work at present.

can you point me to a sample video you're using as a test case?

src/systems/material.js
```diff
@@ -195,6 +205,12 @@ module.exports.System = registerSystem('material', {
   }
 });
 
+function isHLS (videoEl) {
+  if (videoEl.type === 'application/x-mpegurl') return true;
```
cvan: do MIME types get lowercased automatically when calling the `type` getter? 

I notice in [Apple's docs](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/StreamingMediaGuide/DeployingHTTPLiveStreaming/DeployingHTTPLiveStreaming.html), it appears as `application/x-mpegURL`.

- also, do you want to check `vnd.apple.mpegURL` as well?
- and what about `video/MP2T` (`.ts`)?

src/systems/material.js
```diff
@@ -195,6 +205,12 @@ module.exports.System = registerSystem('material', {
...
+  if (videoEl.src && videoEl.src.indexOf('.m3u8') > 0) return true;
```
cvan: I would check `videoEl.src.toLowerCase()`

src/shaders/ios10hls.js
```diff
@@ -0,0 +1,32 @@
+var registerShader = require('../core/shader').registerShader;
+
+/**
+ * Custom shader for iOS 10 HLS.
```
cvan: can you add a comment with a URL to a doc/spec for this API?

cvan:
@machenmusik nice work on supporting this 👍 if you can point me to some test case URLs, I'll try this out on my end.

src/systems/material.js
machenmusik: i've only seen lowercase, but good question.

not sure on your other two points

src/shaders/ios10hls.js
machenmusik: sorry which API?

machenmusik:
i think i used hls urls from the prior threejs tickets on the topic

maybe should just convert the aframe 360 video from MP4 to iPhone-safe HLS?

src/systems/material.js
dmarcos: Can we check for `toLowerCase` and add braces to the return?
````javascript
if (videoEl.src && videoEl.src.toLowerCase().indexOf('.m3u8') > 0) { return true; }
````

File: src/shaders/ios10hls.js (hunk 5 above)
Comment (cvan) [inline src/shaders/ios10hls.js]: some Apple docs (or Google's Web Fundamentals docs) about what HLS is

**Comment by cvan** [comment]
> maybe should just convert the aframe 360 video from MP4 to iPhone-safe HLS?

how about including both as `<source>`s?

**Comment by machenmusik** [comment]
if including both as `<source>` works, that's fine although it complicates the example since you can't do with inline `src` in `a-videosphere` for example -- I was just trying to say that if we want HLS source, should probably make a version of the recently-iDevice-friendly MP4 example that is HLS format

**Comment by cvan** [comment]
awesome, thanks for making the updates 👍 btw, is there a simple test or two you can add?

**Comment by wjhtinger** [comment]
How about play the live 360 stream from webrtc?

**Comment by cvan** [comment]
See [my comment](https://github.com/aframevr/aframe/issues/2663#issuecomment-300538257) in issue #2663. If @machenmusik has other insights, would be glad to know and discuss. Thanks, everyone 👍

**Comment by dmarcos** [comment]
@machenmusik I want to merge this but I would like to see it working first. Do you have any example I can try?

**Comment by machenmusik** [comment]
so per discussion with @dmarcos, it seems that recent changes may have broken this somehow.., may need to do a little more investigation before merging this one :-/

**Comment by machenmusik** [comment]
found the problem :-)

**Comment by dmarcos** [comment]
Thanks!

File: src/utils/material.js [hunk 6]
```diff
@@ -127,3 +131,9 @@ function handleTextureEvents (el, texture) {
   });
 }
 module.exports.handleTextureEvents = handleTextureEvents;
+
+module.exports.isHLS = function (videoEl) {
+  if (videoEl.type && videoEl.type.toLowerCase() in ['application/x-mpegurl', 'application/vnd.apple.mpegurl']) { return true; }
```
cvan: arrays don't work like this. you've got to use `includes` or `indexOf` (which has better compatibility across the board). test for this?

src/utils/material.js
```diff
@@ -127,3 +131,9 @@ function handleTextureEvents (el, texture) {
...
+  if (videoEl.src && videoEl.src.toLowerCase().indexOf('.m3u8') > 0) { return true; }
```
cvan: what about checking for `.ts` too?

src/systems/material.js
```diff
@@ -144,6 +145,16 @@ module.exports.System = registerSystem('material', {
     texture.minFilter = THREE.LinearFilter;
     setTextureProperties(texture, data);
 
+    // if we're on iOS, and the video is HLS, we currently need to do some hacks
+    if (utils.device.isIOS() && isHLS(videoEl)) {
```
cvan: have you tested this on [Edge](https://webvr.rocks/microsoft_edge) (which supports WebVR in release channel)? Edge [supports HLS](http://caniuse.com/#feat=http-live-streaming) too; it may or may not have the same bugs.

machenmusik: it does not.

src/utils/material.js
machenmusik: .ts is not HLS, it is not clear that either .ts playback works or that it needs the workaround (MP4 does not)

if you'd like to add .ts, and it needs this workaround, perhaps you'd like to do in separate PR as this one is merged now?

src/utils/material.js
machenmusik: maybe it only works based on src.   i do owe some tests (as subsequent PR) so will fix there as needed

src/systems/material.js
cvan: it does not have the bugs, or it does not have proper HLS support?

src/utils/material.js
cvan: it doesn't work for me in Chrome nor iOS - you sure?

screenshot of Chrome

[here's a test case](http://jsbin.com/qakosuhude) I tested on iOS.

machenmusik: i think the test page i was using was matching on src (masking broken type check) - will fix when i add isHLS tests as mentioned above

machenmusik: see #2668

src/systems/material.js
machenmusik: it does not appear to have the bugs.

src/utils/material.js
cvan: I'll test it some time to see - I just need to do some investigation there

src/utils/material.js
cvan: awesome! thanks

PR #2602, opened 2017-04-20
Check capabilities instead of dolly

dmarcos:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Works. When I start Oculus Rift in FF Nightly without tracking, user height is no longer applied.

Tests need updating to stub capabilities.

ngokevin:
r+

PR #2607, opened 2017-04-21
make checkIfControllerPresent and addEventListeners calls consistent …

### LINKED ISSUE #2606: triggerup event is emitted twice
**Description:**

Both `oculus-touch-controls` and `hand-controls` will emit `triggerup`.

machenmusik:
…across controllers

should fix #2606


dmarcos:
Thanks!

PR #2610, opened 2017-04-21
fix videoPlayback html attributes

esterlus:
**Description:**

Video tags assume that autoplay/controls should be set if absent from video elements.
This seems to deviate from the expected standard as described here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#Attributes

`webkit-playsinline` is superseded by `playsinline` and should be set as well

**Changes proposed:**
- adhere autoplay and controls attributes if not set (do not automatically assume autoplay / controls is intended)
- check video tag correctly and set playinline where webkit-playsinline is set


ngokevin:
Thanks!

Issue #2554, opened 2017-04-05
Array property type default share the same reference


ngokevin:
**Description:**

```
schema: {
  foo: {type: 'array'},
  bar: {type: 'array'}
}
```

The two arrays share the same reference. If you push to `bar`, it will also be seen as pushed to `foo`.

It's fixed if I specify my own array:

```
schema: {
  foo: {type: 'array', default: []},
  bar: {type: 'array', default: []}
}
```

- A-Frame Version: 0.5.0

takahirox:
Could I try to fix this?

ngokevin:
Yeah!

PR #2615, opened 2017-04-24
Fix shared array reference

takahirox:
**Description:**

This PR fixes the issue "Array property type default share the same reference" #2554

**Changes proposed:**
- clone default value with `.slice()` if it's an array in `parseProperty()` of `schema.js`

### DIFF
--- a/src/core/schema.js
+++ b/src/core/schema.js
@@ -138,6 +138,7 @@ function parseProperty (value, propDefinition) {
   // Use default value if value is falsy.
   if (value === undefined || value === null || value === '') {
     value = propDefinition.default;
+    if (Array.isArray(value)) { value = value.slice(); }
   }
   // Invoke property type parser.
   return propDefinition.parse(value, propDefinition.default);
--- a/tests/core/schema.test.js
+++ b/tests/core/schema.test.js
@@ -119,6 +119,16 @@ suite('schema', function () {
         scale: {x: 0, y: 0, z: 0}
       });
     });
+
+    test('does not share the reference to an array without default values specification', function () {
+      var schema = processSchema({
+        foo: {type: 'array'},
+        bar: {type: 'array'}
+      });
+      var parsed = parseProperties({}, schema);
+
+      assert.ok(parsed.foo !== parsed.bar);
+    });
   });
 
   suite('processPropertyDefinition', function () {


takahirox:
Ah, wait.
Maybe I think I should do that in arrayParse() of propertyTypes.js.
I'll update...

src/core/schema.js
```diff
@@ -138,6 +138,7 @@ function parseProperty (value, propDefinition) {
   // Use default value if value is falsy.
   if (value === undefined || value === null || value === '') {
     value = propDefinition.default;
+    if (Array.isArray(value)) { value = value.slice(); }
```
ngokevin: Do we need to do something similar for objects as well?

dmarcos: What do you mean by objects? The only types we have that can receive an object are `vec2`, `vec3` and `vec4` and those are already copied when parsed: https://github.com/aframevr/aframe/blob/master/src/utils/coordinates.js#L17

ngokevin: In case I define my own property type and set the default value as an object. I often create "schemaless" components that just accept flexible data.

```
default: {},
parse: function (val) {
  // ... Parse to object or pass through object.
}
```

dmarcos: if you do that is the parse function that should make sure that the object is copied.

dmarcos: Copying objects is inefficient and should be a deliberate decision. I think sharing objects is a more sensible default

ngokevin: To describe the situation that could cause unexpectedness (similar to when I found out components were messing with each other's arrays):

```
AFRAME.registerComponent('foo', {
  schema: {
    someData: { 
      default: {value: 0}, parse: function () { // ... }
    }
  }

  init: function () {
    this.data.someData.value += 1;
  }
});

<a-entity id="a" foo></a-entity>
<a-entity id="b" foo>/a-entity>
```

Both instances of `foo` are sharing the same object and polluting each other's data sets. Entity A and Entity B would be both surprised that somehow their values are now `2` when I'd expect `1`.

This is also inconsistent that if `foo` did define data, it would no longer share the object. Only if they use the default value, it'd be copied once per initialization (in most of my cases in creating schemaless components, it'd be copying an empty object).

dmarcos: Sure but your custom `parse` function can copy the objects

ngokevin: That's more of a weird workaround. It's not really parsing, I'd have to run a check if it's exactly the default object to figure to copy it, and I only need it to be done once if it's using the default value and not on every parse. I propose to handle it because the case is unexpected.

The object copy won't be hit that often if we handle the logic here. Only once per property initialization, and only if it's usually the default value. Plus, it's usually an empty object.

dmarcos: Can we have an `object` type in the property types that is shallow? It allows just one level where the attributes are primitive types based on the default value? Should we deal with this in a separate issue / PR?

ngokevin: I don't really have a need for that. I might have a component that generates entities. Or I sometimes use it to create schemaless components. One level is often limiting.

```
AFRAME.registerComponent('creator', {
  schema: {
    geometry: myCustomObjPropertyType,
    material: myCustomObjPropertyType
  }
});
```

ngokevin: Will merge though and figure it out later.

Issue #2629, opened 2017-04-25
BUG: in registerPrimitive, "mappings" is case-sensitive


morandd:
The keys in the "mappings" struct in registerprimitive are case-sensitive and must be lower case to work.
This is totally non-obvious. 

Could you either add a mention about this in the doc page 
  https://aframe.io/docs/0.5.0/primitives/#register-a-primitive

or resolve this in primitives.js.  I think that the DOM is returning the element attribute names using either lower case or whichever case the user happened to use, but there are a number of instances like "mappings[attr]" which is a case-sensitive lookup between the DOM attribute and the mapping name.

There is something about camelCase in addComponentMapping but it's not clear to me what this is.

- A-Frame Version:  0.5.0
- Platform / Device: all
- Reproducible Code Snippet or URL:

AFRAME.registerPrimitive('a-myNewComponent', {
	defaultComponents: {
		myNewComponent: {}
	},
	mappings: {
		width: 		'myNewComponent.width', // fine
		defaultColor: 'myNewComponent.defaultColor', // Should work, but fails
		lineopacity:	'myNewComponent.lineOpacity' // works if I do this
	}
})

ngokevin:
We can document it. Note all HTML attributes are lower case on the Web.

morandd:
thanks!

ngokevin:
fixed

PR #2631, opened 2017-04-26
warn about camelCase mappings in registerPrimitive

morandd:
Added a warning in registerPrimitive when the user has provided mappings keys that are not all lower case. Since the DOM reports all element attributes in lower case (regardless of how the user typed them in the HTML source), the mappings keys must also be lower case.

This issue was discussed here
  https://github.com/aframevr/aframe/issues/2629
A separate PR was previously accepted to update the documentation to mention that mappings are case-sensitive.

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/extras/primitives/primitives.js
+++ b/src/extras/primitives/primitives.js
@@ -45,6 +45,7 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
           var self = this;
           Object.keys(mappings).forEach(function resolveCollision (key) {
             var newAttribute;
+            if (key != key.toLowerCase()) console.warn('Mapping keys should be specified in lower case. This mapping may not be recognized');
             if (components[key]) {
               newAttribute = mappings[key].replace('.', '-');
               mappings[newAttribute] = mappings[key];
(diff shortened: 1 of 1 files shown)


src/extras/primitives/primitives.js
```diff
@@ -45,6 +45,7 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
           var self = this;
           Object.keys(mappings).forEach(function resolveCollision (key) {
             var newAttribute;
+            if (key != key.toLowerCase()) console.warn('Mapping keys should be specified in lower case. This mapping may not be recognized');
```
ngokevin: You can use `warn` https://github.com/morandd/aframe/blob/9e0f76f70927cbb9d1936f2dcd59f7d06654586d/src/extras/primitives/primitives.js#L9

ngokevin: `!==`

ngokevin: Use `braces` for `if` statements

morandd:
switched from console.warn to warn() and added braces around if. I hope it's OK (but please take a quick look - I'm not such a genius with javascript)

ngokevin:
Thanks. No problem, just style / minor stuff.

ngokevin:
Still need to adjust one more thing. Switch `!=` to `!==`. 

https://travis-ci.org/aframevr/aframe/jobs/225977249

ngokevin:
thanks!

Issue #2633, opened 2017-04-26
Disabling look-control doesn't remove grab cursor 


a-barbieri:
**Description:**

- A-Frame Version: 0.5.0 (Date 10-02-2017, Commit #110055d)
- Platform / Device: OSX 10.12 Chrome 57
- Reproducible Code Snippet or URL: https://codepen.io/deva/pen/eWgYeo

In my application I would like to remove the *look controls* and set some custom behaviours.

I removed controls from the camera as display below but I can still see the grab-cursor while clicking/dragging the scene.

```
<a-camera look-controls-enabled="false" wasd-controls-enabled="false">
```

ngokevin:
In `src/components/look-controls.js`, the code should check to see if `this.data.enabled` before adding the class on the document body for the cursor. Help wanted!

gurumukhi:
Created PR : https://github.com/aframevr/aframe/pull/2637

gurumukhi:
Created new PR - https://github.com/aframevr/aframe/pull/2642/files

ngokevin:
https://github.com/aframevr/aframe/pull/2642

PR #2637, opened 2017-04-27
Adding check before adding grabbing class to body

gurumukhi:
Checking look-controls-enabled before adding a-grabbing class to body.

Ref - https://github.com/aframevr/aframe/issues/2633

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -256,7 +256,9 @@ module.exports.Component = registerComponent('look-controls', {
   onMouseDown: function (event) {
     this.mouseDown = true;
     this.previousMouseEvent = event;
-    document.body.classList.add('a-grabbing');
+    if(this.data.enabled){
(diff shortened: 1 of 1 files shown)


mkeblx:
```js
if (!this.data.enabled) { return; }
```  
 at beginning might be a little cleaner. A related issue that can tackle is that there is still the open hand grab-cursor in `init` being added even if not enabled which don't want so would need to only do if `enabled`.

gurumukhi:
Created new PR - https://github.com/aframevr/aframe/pull/2642/files

mkeblx:
Cool. 
One more thing.. the deferred `sceneEl.addEventListener('render-target-loaded', enableGrabCursor);` also needs to be only be called if enabled so could put the check around both if & else.

ngokevin:
Dupe https://github.com/aframevr/aframe/pull/2642

PR #2642, opened 2017-04-29
Checking look-controls-enabled before grabbing

gurumukhi:
Checking look-controls-enabled before adding grabbing classes.

Ref - #2633

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -18,6 +18,7 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   init: function () {
+    var self = this;
     var sceneEl = this.el.sceneEl;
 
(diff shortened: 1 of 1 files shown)


src/components/look-controls.js
```diff
@@ -33,7 +33,7 @@ module.exports.Component = registerComponent('look-controls', {
     function enableGrabCursor () { sceneEl.canvas.classList.add('a-grab-cursor'); }
     if (!sceneEl.canvas) {
       sceneEl.addEventListener('render-target-loaded', enableGrabCursor);
-    } else {
+    } else if (this.data.enabled) {
```
ngokevin: This check should moved into `enableGrabCursor`

PR #2645, opened 2017-05-02
Change script type to html in 360 gallery sample

elifer5000:
When script type is nunjucks, a <SERVERNAME>/$%7Bthumb%7D not found is logged, and the thumbnail is not loaded.
Changing the text type to `html` solves this problem.

[Not working fiddle](https://jsfiddle.net/a0tuLa22/)
[Working fiddle](https://jsfiddle.net/okmnjako/1/)


elifer5000:
There's a typo in the original commit, it should say `html` and not `text`

elifer5000:
There's a typo in the original commit, it should say `html` and not `text`

Issue #2635, opened 2017-04-27
a-grabbing class applied on 360 video on mouse mouse.


aminasyeda:
I have a problem with the 'a-grabbing' class on a 360 video. 

Steps to reproduce: 
Right click on a 360 video. Click 'Inspect'. Without doing a mouse-click get the mouse back to the video. The video rotates with just the mouse mouse (with no grabbing/dragging of the mouse). One of the other ways to reproduce this issue is to click repeatedly on a video embedded in an iframe, drag the iframe and release the mouse on the video. Apparently, the 'a-grabbing' class gets added to the body element in such cases. 
The behavior is the same even if I try to remove the grabbing class by adding a mouseMove event listener. Please advise.

Please see the behavior on : https://aframe.io/examples/showcase/videosphere/

- A-Frame Version: 0.5.0 (Date 10-02-2017, Commit #bf8b8f9)
- Platform / Device: Windows/ Linux/ Chrome

mkeblx:
Something like making `mouseDown` method only handle primary button should be all that's needed.. `if (event.button !== 0) { return; }`

BhuvanaMeenakshi:
I want to work on this

mkeblx:
BhuvanaMeenakshi this should be fixed in #2647

PR #2647, opened 2017-05-03
look-controls enable/disable lifecycle fix

mkeblx:
**Description:**
As look-controls cursor setup only done in init would not work correctly if
changed during runtime.

**Changes proposed:**
-Modify to do on setup + change of enablement state.
-Also, fix for #2635

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -18,7 +18,6 @@ module.exports.Component = registerComponent('look-controls', {
   },
 
   init: function () {
-    var self = this;
     var sceneEl = this.el.sceneEl;
 
     // Aux variables
@@ -30,16 +29,7 @@ module.exports.Component = registerComponent('look-controls', {
     this.setupHMDControls();
     this.bindMethods();
 
-    // Enable grab cursor class on canvas.
-    function enableGrabCursor () {
-      if (!self.data.enabled) { return; }
-      sceneEl.canvas.classList.add('a-grab-cursor');
-    }
-    if (!sceneEl.canvas) {
-      sceneEl.addEventListener('render-target-loaded', enableGrabCursor);
-    } else {
-      enableGrabCursor();
-    }
+    this.setEnabled(this.data.enabled);
 
     // Reset previous HMD position when we exit VR.
     sceneEl.addEventListener('exit-vr', this.onExitVR);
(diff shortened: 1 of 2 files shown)


src/components/look-controls.js
```diff
@@ -59,6 +52,24 @@ module.exports.Component = registerComponent('look-controls', {
     this.updatePosition();
   },
 
+  setEnabled: function (enabled) {
+    var sceneEl = this.el.sceneEl;
+
+    function enableGrabCursor () {
+      sceneEl.canvas.classList.add('a-grab-cursor');
```
cvan: IMO this should be a `data-` attribute instead of a class name

src/components/look-controls.js
```diff
@@ -59,6 +52,24 @@ module.exports.Component = registerComponent('look-controls', {
...
+    }
+    function disableGrabCursor () {
+      sceneEl.canvas.classList.remove('a-grab-cursor');
+    }
+
+    if (!sceneEl.canvas) {
+      sceneEl.addEventListener('render-target-loaded',
+        enabled ? enableGrabCursor : disableGrabCursor);
+    } else {
+      enabled ? enableGrabCursor() : disableGrabCursor();
```
cvan: I'd just use `if` statements

src/components/look-controls.js
```diff
@@ -259,6 +270,8 @@ module.exports.Component = registerComponent('look-controls', {
 
   onMouseDown: function (event) {
     if (!this.data.enabled) { return; }
+    // only handle primary button
```
cvan: `// Handle only primary button.`

src/components/look-controls.js
```diff
@@ -259,6 +270,8 @@ module.exports.Component = registerComponent('look-controls', {
...
+    if (event.button !== 0) { return; }
```
cvan: is this universally true for all platforms?

src/components/look-controls.js
dmarcos: yeah let's do an if statement

src/components/look-controls.js
mkeblx: I think so.
> 0: Main button pressed, usually the left button or the un-initialized state

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button

src/components/look-controls.js
mkeblx: Ok

src/components/look-controls.js
mkeblx: I could do that & mod the CSS. Lmk if that is indeed what want

dmarcos:
Thanks!

PR #2648, opened 2017-05-03
[docs] Fix parseInt usage

dunatotatos:
If used in a naive way, ParseInt returns weird results like [3, NaN, NaN]. See for example http://stackoverflow.com/questions/262427/javascript-arraymap-and-parseint

Remove a ; which was a syntax error.


ngokevin:
thanks!

Issue #1271, opened 2016-03-25
Video materials should display poster attribute


jarmitage:
`<video>` assets can contain a `poster` attribute specifying an image to be displayed before the video has loaded, or before it has been played.

When adding a video as a material `src`, it would be beneficial if the material could look for and display the  `poster` image. This should allow for the `src` attribute to be left empty when the video is not playing, so that memory is not used to load a video that is not being played.

The current workaround is to use an additional `<img>` asset containing the `poster` image, and to dynamically switch the material `src` to the `<video>` `src` when needed.

ngokevin:
This can be application behavior. Use a video element in `<a-assets>`, listen to its events, and show an image texture on an entity in the meantime

cvan:
I'd like to work on this. Especially when videos take a while to load (which is pretty much always), having a black screen while the video loads (or having to manually apply an image texture yourself), it's just not ideal default behaviour.

The HTML5 `<video>` tag already supports the [`poster` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster) (not just used on default HTML5 players, but it's used by YouTube/Vimeo players too).

So, like you have `<video poster="poster.png" src="poster.mp4">`, A-Frame devs could just similarly do `<a-videosphere poster="poster.png" src="poster.mp4">`.

Would be fantastic!

ngokevin:
This can be handled application level. Show image, when video ready, toggle the visible.

PR #2657, opened 2017-05-05
fix `<video autoplay>` playback (when audio is present) on mobile for both iOS + Android

cvan:


src/core/a-assets.js
```diff
@@ -163,18 +164,32 @@ function mediaElementLoaded (el) {
...
       }
     }
+
+    // Respect WebKit's user-gesture requirement for auto-playback of
+    // video containing audio. For more info:
+    // https://webkit.org/blog/6784/new-video-policies-for-ios/
+    window.addEventListener('click', handleFirstClick);
+    function handleFirstClick () {
+      try {
+        if (el.paused) {
```
dmarcos: one liner

src/core/a-assets.js
```diff
@@ -221,22 +236,24 @@ function setCrossOrigin (mediaEl) {
...
- * Extract domain out of URL.
+ * Extract domain (without port) out of URL.
  *
  * @param {string} url
  * @returns {string}
  */
 function extractDomain (url) {
-  // Find and remove protocol (e.g., http, ftp, etc.) to get domain.
-  var domain = url.indexOf('://') > -1 ? url.split('/')[2] : url.split('/')[0];
+  if ('URL' in window) {
```
dmarcos: one liner

src/core/a-assets.js
cvan: updated, thanks

machenmusik:
so unless i misunderstand, this will attempt to fix a mobile issue with autoplay by forcing play on first click, but without really understanding that the video should be playing so may actually start videos that should remain paused?

machenmusik:
also the fix does not appear to be constrained to apply only to mobile / cases known to require it... is that safe?

src/core/a-assets.js
```diff
@@ -163,18 +164,30 @@ function mediaElementLoaded (el) {
...
       }
     }
+
+    // Respect WebKit's user-gesture requirement for auto-playback of
+    // video containing audio. For more info:
+    // https://webkit.org/blog/6784/new-video-policies-for-ios/
+    window.addEventListener('click', handleFirstClick);
+    function handleFirstClick () {
+      try {
+        if (el.paused) { el.play(); }
```
dmarcos: Should we check if autoplay is set on the video element?

cvan: it's already being checked [above](https://github.com/aframevr/aframe/pull/2657/files#diff-565a65bb6612d5571efb7439e5423720R139)

cvan: I haven't tested the differences between all the browsers, but to be on the safe side, I think this should be checking `el.autoplay` instead of just `el.getAttribute('autoplay')`, but I think it's fine

cvan:
btw, after I get this one out of the way, I have one final PR: a patch (client-side only too) that fixes the cross-origin video playback issues on iOS (e.g., for videos hosted on CDN) 🎉 @machenmusik: I know you've been dabbling in that realm quite a bit.

cvan:
ok, just made these two changes (weren't exactly necessary, but figured it'd clean things up):

* https://github.com/aframevr/aframe/pull/2657/files#diff-565a65bb6612d5571efb7439e5423720R183
* https://github.com/aframevr/aframe/pull/2657/files#diff-565a65bb6612d5571efb7439e5423720R148

tested on Android and iOS, and all appears good

machenmusik:
should probably test on rift/vive with a video that shouldn't autoplay to ensure do no harm.
maybe fire an event or two so "tap to play" ux can be displayed/dismissed?

machenmusik:
apologies that I am not currently in a position to test on iOS (had borrowed a device but returned it)

cvan:
> maybe fire an event or two so "tap to play" ux can be displayed/dismissed?

I'm just handling `autoplay`, yeah. I don't want to add video controls for non-autoplay video; that would make too many assumptions about the content IMO. and, FWIW, desktop already doesn't have video controls/toggles. [this component](https://github.com/oscarmarinmiro/aframe-video-controls) I think is a good place to start (until browsers start shipping with their own consistent video 360° controls).

machenmusik:
right, I didn't mean actually implementing any UX for it in core, just the events that would be needed to do so (since otherwise it's kinda cryptic, and if you actually have to present UX without the events in core, you have to do yourself anyway, negating the benefits in moving handling into core).  I basically do this now at higher level (when you call play, set a timeout to check if it actually did anything, if not present some ux and attach self-removing handler)

cvan:
@machenmusik Thanks, that I agree with. I had a long response, but it got lost in my tabs. Basically it was to the effect of this:

Chrome for Android and Safari on iOS & desktop have the same initial requirements now: `<video autoplay>` with an audio source or `videoEl.muted === false` require an initial user-gesture requirement for playback.

That's my only intention of addressing in this PR (and in a follow-up PR: the [infamous CORS Safari bug](https://bugs.webkit.org/show_bug.cgi?id=135379)).

<hr>

The [`aframe-video-controls` A-Frame component](https://github.com/oscarmarinmiro/aframe-video-controls) is a good place to prototype what "controls" look like, beyond the obvious `<video autoplay>` case to handle.

FWIW, Samsung's got built-in 360° video controls (built proprietarily to start, to get something out the door). It looks like the next browser likely to ship 360° video controls (but this time as a "standards-compliant solution") is the Oculus Browser/WebVR team (in coordination with React VR); there will be no manual touch user-gesture requirement (`vrdisplayactivate` will just be triggered). And, I'm not aware of Google's status of adding support in Chrome, but it wouldn't be a stretch to imagine Google's working on 360° video controls now or in the near future. Lastly, I think I can speak for the folks at Mozilla when I say Firefox and Servo would like to add in-chrome  360° video controls, but it's currently not on our list of prioritized bugs (as of now).

Thanks for the feedback. Hope that answers some questions too.

cvan:
@dmarcos given my comment above, this look good to you?

machenmusik:
@cvan per conversation, does this emit an event when installing the tap/click to play handler?  standard html video events can be used to detect play, but maybe not installing the handler

cvan:
> @cvan per conversation, does this emit an event when installing the tap/click to play handler? standard html video events can be used to detect play, but maybe not installing the handler

not _installing_, but upon playing, yes. do you think there ought to be an event? and, if so, should it be emitted on the `<video>` element in `<a-assets>` _and_ fired on the `<a-entity>`? (if a video URL is provided instead of a `src` to `<video>`, A-Frame creates its own `<video>` behind the scene, so the event wouldn't be incredibly useful in that case.)

I can add an event. what are the use cases you're thinking for that? (besides just good measure?)

machenmusik:
The use case is literally to present some UX (say, head locked a-text "Tap/click to play") when the code added in this PR determines that it is required.  That UX should only be visible when said click handler is required to be installed (which my usual code at this point does isMobile() to help filter out, IIRC).  

If you're not going to constrain this behavior to mobile (or specific UA behavior as you detailed in comments) I do wonder if some testing/experimentation is needed in case non-mobile cases have a sufficient delay before playing that it looks like the not-going-to-play mobile case (e.g. if the video takes a long time to buffer initially).

cvan:
> The use case is literally to present some UX (say, head locked a-text "Tap/click to play") when the code added in this PR determines that it is required. That UX should only be visible when said click handler is required to be installed (which my usual code at this point does isMobile() to help filter out, IIRC).

I was thinking more to handle this with a poster 360° image/skybox (à la `<video poster="example.png" src="example.mp4"></video>`): #1271

I'd like to get video playback to work first before adding a UI, as any defaults – even if they're necessary – can prove to be quite intrusive for users. IMO, it's best to start with no controls but get playback working out of the box, and then add controls shortly thereafter. It's harder to remove a UI (and it could complicate users' experiences if the UI changes a lot, `<video>` 360° controls get standardised, etc.) than it is to add a UI.

> If you're not going to constrain this behavior to mobile (or specific UA behavior as you detailed in comments) I do wonder if some testing/experimentation is needed in case non-mobile cases have a sufficient delay before playing that it looks like the not-going-to-play mobile case (e.g. if the video takes a long time to buffer initially).

In my follow-up PR for iOS, I do sniffing for Safari (for the WebKit CORS bug). This is a good point you bring up. At the very least, I think it'd probably make sense for me to sniff for Safari _or_ Chrome for Android; that would get `autoplay` video playback working for the majority of the use cases.

And with a `poster` image (#1271), as soon as a user clicks anywhere on the page (including clicking the `Enter VR` button), a 360° video would play. It's really great 👍

Thanks for the feedback; this helped a ton. I really appreciate it.

machenmusik:
re: poster, sometimes the ux that should be displayed for click-to-play is a property of the page/site and not the video, so maybe less tied to the poster that would normally be attached to the video.

I agree that we don't need to include ux here, but as we are modifying the usual event handling to do something smarter for click-to-play, I do think we should emit a click-to-play event on the A-Frame entity, or the scene as an alternative, as the A-Frame event handling markup doesn't understand the video element really.  Then authors can do whatever event handling for their UX with typical A-Frame semantics.

dmarcos:
I just tried this on an iphone 7 Plus and I cannot get the video to play. I loaded the `tests\videoshpere` example and video does not play when tapping on VR mode. Can anyone get the video to play?

cvan:
> I just tried this on an iphone 7 Plus and I cannot get the video to play. I loaded the `tests\videoshpere` example and video does not play when tapping on VR mode. Can anyone get the video to play?

if you copy a video file from the Uploadcare URLs and move them to the `examples/` directory, you can see it working.

if you use Safari on macOS to Remotely Debug your iPhone, you'll see the Console is littered with CORS errors. that's the "infamous CORS bug" I referred to in https://github.com/aframevr/aframe/pull/2657#issuecomment-300490769 and https://github.com/aframevr/aframe/pull/2657#issuecomment-300164119

I have a branch locally, the second PR of these mobile video issues, which addresses that - using a workaround trick with Blobs.

<hr>

we could have an `.mp4` video sample checked in to `examples/` directory of this repo. as long as we aren't adding more than one, I think we should be fine re: the repo size. (plus, it'd make it easier for folks doing offline development, if they forgot to download the videos first.)

so if you download https://ucarecdn.com/fadab25d-0b3a-45f7-8ef5-85318e92a261/city.mp4 and save it to `examples/boilerplate/360-video/city.mp4` and update the video path in `examples/boilerplate/360-video/index.html` to point `city.mp4`, you should see it all working.

let me know if you're still having issues or have questions. thanks 👍

machenmusik:
the CORS "fix" is unlikely to work with HLS video, so if it is going to be introduced into core can we please either verify HLS survives it, or place appropriate video type guards?

cvan:
> the CORS "fix" is unlikely to work with HLS video, so if it is going to be introduced into core can we please either verify HLS survives it, or place appropriate video type guards?

yep, I'll test it with HLS and let you know - thanks 👍

dmarcos:
@cvan I think this is not too far from being mergable. There are some pending questions when you have the time: https://github.com/aframevr/aframe/pull/2657#issuecomment-306945378

machenmusik:
Also type guards to avoid breaking other video... https://github.com/aframevr/aframe/pull/2657#issuecomment-301069075
Some concern the blob technique will not work for large videos.
What size videos are you testing with?

cvan:
> What size videos are you testing with?

- https://ucarecdn.com/25f36ea9-579e-4914-85ed-e131dd9e75d1/walking_the_dog_360_video_short.mp4
- https://ucarecdn.com/16cbcc72-ef7a-44f5-be04-3988d0ce9c6c/nice_dive_with_wild_dolphins.mp4 (fetched from https://www.youtube.com/watch?v=zUzLvybAGKM)
- https://ucarecdn.com/fadab25d-0b3a-45f7-8ef5-85318e92a261/city.mp4

@machenmusik: which videos are you testing with? so I can test large videos and real-world HLS/DASH examples?

machenmusik:
@cvan given the potential CORS issues with content, am wondering whether we should convert the iPhone friendly version of stock MP4 into HLS (MP2TS) so that an A-Frame demo showing iOS HLS would actually work right... which would also be the source you need to test here

I think the original three.js ticket for iOS HLS had some streams, I remember seeing an example in there with an ultralight but don't recall the specific name (much less the URL)

cvan:
@machenmusik do you have a URL of a video example?

machenmusik:
@cvan let me try to quickly make HLS from @Utopiah's https://github.com/aframevr/aframe/issues/1846#issuecomment-309201624 ...

machenmusik:
Is this one dead?

cvan:
@machenmusik video policies need to be handled quite uniquely such that user activation is respected properly based on the audio/video media being played (or autoplayed).

if I address the outstanding review comments, is there any chance that this would be merged? I ask based on the feedback on related PRs and issues concerning video playback (not only for Safari on desktop and iOS, but also for Android - both for Randall's new GeckoViewVR APK, existing Chrome-based browsers).

machenmusik:
well, we got #3176 merged successfully, but #2830 didn't make it, and there seems to be understandable negative feedback on accommodating browser specific quirks in core.  Safari tech preview should finally fix their ancient CORS problem, so that particular aspect shouldn't be required going forward at least.

@dmarcos @ngokevin @donmccurdy thoughts?

machenmusik:
sidebar: @cvan let me know what you think of the example in #3210 that has multiple sources and HLS, along with some crude click-to-play scripting per the above

ngokevin:
This was marked `needs answers` for https://github.com/aframevr/aframe/pull/2657#issuecomment-306945378

ngokevin:
Closing for age and in favor of handling this explicitly for now, and have developer wire a click listener to play media. I also encourage browsers become more friendly with VR content. The PR is focused on a single-video experience, but has the effect if there are multiple videos / sound, will play them all at once.

PR #2668, opened 2017-05-13
Add iOS HLS tests

machenmusik:
Add tests and do minor refactoring per discussion on #2597

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -151,13 +151,6 @@ Checks if device is Gear VR. Returns a `boolean`.
 
 Checks if device is a smartphone. Returns a `boolean`.
 
-## `AFRAME.utils.material`
-
-### `AFRAME.utils.material.isHLS (src, type)`
-
-Given video element src and type, guess whether stream is HLS.
-This is used internally to make special accommodations for HLS playback on iOS 10+ using Mobile Safari.
-
 ## Function Utils
 
 ### `AFRAME.utils.throttle (function, interval [, optionalContext])`
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -151,6 +151,13 @@ Checks if device is Gear VR. Returns a `boolean`.
 
 Checks if device is a smartphone. Returns a `boolean`.
 
+## `AFRAME.utils.material`
+
+### `AFRAME.utils.material.isHLS (src, type)`
+
+Given video element src and type, guess whether stream is HLS.
+This is used internally to make special accommodations for HLS playback on iOS 10+ using Mobile Safari.
+
 ## Function Utils
 
 ### `AFRAME.utils.throttle (function, interval [, optionalContext])`
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -146,7 +146,7 @@ module.exports.System = registerSystem('material', {
     setTextureProperties(texture, data);
 
     // if we're on iOS, and the video is HLS, we currently need to do some hacks
-    if (utils.device.isIOS() && isHLS(videoEl.getAttribute('src'), videoEl.getAttribute('type'))) {
+    if (utils.device.isIOS() && isHLS(videoEl.src || videoEl.getAttribute('src'), videoEl.type || videoEl.getAttribute('type'))) {
       // really it's BGRA, so this needs correction in shader
       texture.format = THREE.RGBAFormat;
       texture.needsCorrectionBGRA = true;
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -146,7 +146,7 @@ module.exports.System = registerSystem('material', {
     setTextureProperties(texture, data);
 
     // if we're on iOS, and the video is HLS, we currently need to do some hacks
-    if (utils.device.isIOS() && isHLS(videoEl.src || videoEl.getAttribute('src'), videoEl.type || videoEl.getAttribute('type'))) {
+    if (this.sceneEl.isIOS && isHLS(videoEl.src || videoEl.getAttribute('src'), videoEl.type || videoEl.getAttribute('type'))) {
       // really it's BGRA, so this needs correction in shader
       texture.format = THREE.RGBAFormat;
       texture.needsCorrectionBGRA = true;
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -146,7 +146,7 @@ module.exports.System = registerSystem('material', {
     setTextureProperties(texture, data);
 
     // if we're on iOS, and the video is HLS, we currently need to do some hacks
-    if (utils.device.isIOS() && isHLS(videoEl)) {
+    if (utils.device.isIOS() && isHLS(videoEl.getAttribute('src'), videoEl.getAttribute('type'))) {
       // really it's BGRA, so this needs correction in shader
       texture.format = THREE.RGBAFormat;
       texture.needsCorrectionBGRA = true;
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -1,5 +1,7 @@
 var THREE = require('../lib/three');
 
+var HLS_MIMETYPES = ['application/x-mpegurl', 'application/vnd.apple.mpegurl'];
+
 /**
  * Update `material` texture property (usually but not always `map`)
  * from `data` property (usually but not always `src`)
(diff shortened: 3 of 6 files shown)


machenmusik:
Doing the tests was a bit ugly, but they are now fully functional; in particular, there is not a mechanism to mock things like `AFRAME.utils.device.isIOS()` without the ugly hack of temporarily modifying `AFRAME.utils.device`  i.e. the `scene.isIOS` flag discussed in #2597 to make the tests easy does not exist.

machenmusik:
tests pass, but actual phone tests do not now -- don't merge yet

ngokevin:
I meant In the system, if you change the utils.device.isIOS call to check the scene value, it should be easily mockable by setting the scene.isIOS.

machenmusik:
aaargh, for some reason even though shader substitution test works on firefox ad chrome on my machine, it fails on firefox in the cloud. :-/ gonna have to disable those tests for now as I don't have more time to spend on that

src/systems/material.js
```diff
@@ -146,7 +146,7 @@ module.exports.System = registerSystem('material', {
     setTextureProperties(texture, data);
 
     // if we're on iOS, and the video is HLS, we currently need to do some hacks
-    if (utils.device.isIOS() && isHLS(videoEl)) {
+    if (utils.device.isIOS() && isHLS(videoEl.src || videoEl.getAttribute('src'), videoEl.type || videoEl.getAttribute('type'))) {
```
ngokevin: Just to confirm, you tried changing `utils.device.isIOS()` here to `this.el.sceneEl.isIOS`, and in the test, try mocking `sceneEl.isIOS = true` before the assertions?

docs/core/utils.md
```diff
@@ -151,6 +151,13 @@ Checks if device is Gear VR. Returns a `boolean`.
 
 Checks if device is a smartphone. Returns a `boolean`.
 
+## `AFRAME.utils.material`
+
+### `AFRAME.utils.material.isHLS (src, type)`
```
ngokevin: I don't think this has to be public API

machenmusik: it did to be accessible in tests...?

src/systems/material.js
machenmusik: no, i thought you were saying the substitution had already been done, so only changing the flag was necessary.  if the function has to be monkey patched, not sure it is any clearer to do it as you suggest rather than directly manipulating to return desired value

machenmusik: here the check for the property and if none the getAttribute value was unexpectedly required for both synthetic tests and actual real-world usage to work correctly.

docs/core/utils.md
ngokevin: Public API meaning documented

src/systems/material.js
ngokevin: Use sceneEl.iIOS here so it doesn't have to run the user agent parse again, and you can stub it in the tests.

docs/core/utils.md
machenmusik: ah, I understand you now.

machenmusik:
@ngokevin good now?

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
...
+      var shader = this.shader;
+      var el = this.el;
+      var initSpy = this.sinon.spy(shader.prototype, 'init');
+      var updateSpy = this.sinon.spy(shader.prototype, 'update');
+      assert.notOk(initSpy.called);
+      assert.notOk(updateSpy.called);
+
+      // Mock iOS.  NOTE: this doesn't work... el.sceneEl.isIOS = true;
+      var realIsIOS = AFRAME.utils.device.isIOS;
+      AFRAME.utils.device.isIOS = function () { return true; };
```
ngokevin: `this.sinon.stub(AFRAME.utils.device.isIOS, () => true)` might work, if we use sinon, the teardown will automatically restore it.

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
...
+          // Undo mock of iOS.
+          AFRAME.utils.device.isIOS = realIsIOS;
+
+          // Verify shader was substituted.
+          assert.equal(material.data.shader, 'ios10hls');
+
+          done();
+        });
+      });
+      el.setAttribute('material', {shader: 'testShader', src: videoEl});
```
ngokevin: Do we need to specify `testShader`?

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
...
+      var material = el.components.material;
+      var instance = material.shader;
+      assert.ok(instance);
```
ngokevin: I don't think we need to assert these, should keep focused on what we want to assert (e.g., the shader is switched)

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
...
+      assert.ok(initSpy.calledOnce);
+      assert.ok(updateSpy.calledOnce);
+      // The value won't be assigned until the texture loads.
+      assert.ok(instance.uniforms['src']);
+      assert.notOk(instance.attributes && (instance.attributes['map'] ||
+                                           instance.attributes['src']));
+    });
+
+    // Skip since this fails Travis CI Firefox run, even though it works on Chrome, and with local Firefox.
+    test.skip('iOS HLS video uses appropriate shader (setAttribute)', function (done) {
```
ngokevin: This is a large test, hard to spot what's different from the previous one

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
                                            instance.attributes['src']));
     });
 
+    // Skip since this fails Travis CI Firefox run, even though it works on Chrome, and with local Firefox.
```
machenmusik: no, but it is irrelevant since the actual substitution code won't be ever really be fired using Firefox, and I don't have the time to spare to investigate that

tests/core/shader.test.js
```diff
@@ -152,6 +152,114 @@ suite('shader', function () {
...
+    test.skip('iOS HLS video uses appropriate shader', function (done) {
+      var shader = this.shader;
+      var el = this.el;
+      var initSpy = this.sinon.spy(shader.prototype, 'init');
+      var updateSpy = this.sinon.spy(shader.prototype, 'update');
+      assert.notOk(initSpy.called);
+      assert.notOk(updateSpy.called);
+
+      // Mock iOS.  NOTE: this doesn't work... el.sceneEl.isIOS = true;
```
machenmusik: it may work now (I didn't clean up the skipped tests per the last changes) but I don't have the time to spare to investigate that

tests/core/shader.test.js
machenmusik: perhaps, but I don't have the time to spare to investigate that

tests/core/shader.test.js
machenmusik: I don't know, the original test on which this was based did so kept it, I don't have the time to spare to investigate that

tests/core/shader.test.js
machenmusik: the original test on which this was based did so kept it

tests/core/shader.test.js
machenmusik: agreed, maybe we should just remove the skipped tests for now?

ngokevin: I think one of the two tests should suffice, I can help try to simplify.

src/utils/material.js
```diff
@@ -118,11 +120,16 @@ function handleTextureEvents (el, texture) {
 
   // Video events.
   if (!texture.image || texture.image.tagName !== 'VIDEO') { return; }
+
   texture.image.addEventListener('loadeddata', function emitVideoTextureLoadedDataAll () {
     // Check to see if we need to use iOS 10 HLS shader.
     if (texture.needsCorrectionBGRA && texture.needsCorrectionFlipY) {
```
machenmusik: One thing I was considering is passing `needsCorrectionBGRA` and `needsCorrectionFlipY` into shaders in general when true, so someone with a custom shader would know when corrections are necessary.  Thoughts?

machenmusik:
@ngokevin if you have the time to tidy up whichever of the skipped tests you prefer and get rid of the other one, that should wrap this one up

ngokevin:
Will do right now, thanks!

ngokevin:
Cleaned up at https://github.com/aframevr/aframe/commit/a2d641a8c2d5bf50226be8c4c5ecc198cbb7c72f

- The isIOS check worked for me
- Rather than waiting for the video to load, I just mock the event
- Remove extraneous assertions and other setup meant for other tests

machenmusik:
nice @ngokevin, thank you!

PR #2674, opened 2017-05-18
Avoid string.split() in animation update functions

MortimerGoro:
Avoid memory allocation produced by string.split() in setAttribute function, which is called in the RAF after each TWEEN update animation


dmarcos:
There are a failing tests we have look at

PR #2678, opened 2017-05-18
laser-controls, line, raycaster origin/direction/showLine, cursor downEvents/upEvents (implements, fixes #2646)

### LINKED ISSUE #2646: integrate controller-cursor component with cursor component
**Description:**

Provide out-of-the-box way for attaching laser cursor to controllers, mapping the events, creating a line.

https://github.com/bryik/aframe-controller-cursor-component

Possible API brainstorm:

```
<a-camera>
  <a-entity cursor></a-entity>
</a-camera>

<a-entity hand-controls="right" cursor"></a-entity>
```

The `cursor` component can perhaps detect when it's attached to a controller or not (`tracked-controls`), or whether its parent has a camera.

ngokevin:
**Changes Proposed:**

- `laser-controls` component. Dropping `<a-entity laser-controls>` will display the default controller models for Daydream, GearVR, Vive, or Oculus Touch depending on which is connected. Then configure cursor, raycaster to create an interactive laser. Laser/gaze adapts OK across 0DOF/3DOF/6DOF for those that want to support across.
- Line component from @fernandojsg
- Allow configuration of raycaster component through `origin` and `direction` vec3s. Primarily to tilt the laser down for Oculus Touch which isn't symmetrical or centered on on the Z-axis.
- Allow raycaster component to draw a line with `showLine: true` starting from the raycaster origin and extending to its length. laser-controls turns this on by default
- Raycaster line adjusts its length depending on the distance to the intersection point if any

laser

**TODO:**

- [x] Unit tests
- [x] Test/configure laser for daydream-controls
- [x] Test/configure laser for oculus-touch-controls
- [x] split the crawling cursor changes to a new PR
- [x] Truncate the laser at the intersection point
- [x] Move laser/line logic to raycaster since it will be based off of raycaster origin/direction/intersection 
- [ ] update docs


src/components/cursor.js
```diff
@@ -12,9 +14,9 @@ var EVENTS = {
 };
 
 var STATES = {
-  FUSING: 'cursor-fusing',
-  HOVERING: 'cursor-hovering',
-  HOVERED: 'cursor-hovered'
+  FUSING: 'fusing',
```
dmarcos: I'm glad to see the prefix gone

src/components/laser-controls.js
```diff
@@ -0,0 +1,44 @@
...
+    var data = this.data;
+    var el = this.el;
+
+    el.setAttribute('daydream-controls', {hand: data.hand});
+    el.setAttribute('gearvr-controls', {hand: data.hand});
+    el.setAttribute('oculus-touch-controls', {hand: data.hand});
+    el.setAttribute('vive-controls', {hand: data.hand});
+
+    el.addEventListener('controllerconnected', function (evt) {
+      el.setAttribute('cursor', utils.extend({
```
dmarcos: You're probably aware but we still need the logic here to position the laser based on the controller

dmarcos:
I like the initial approach so far. @dmarcos approves

ngokevin:
Still WIP. Pulled in @fernandojsg's line component. The cursor component uses the configurable raycaster origin/direction to draw the line. Currently working on lining up the lasers. I think only Oculus Touch needs to be modified, the rest look centered.

screen shot 2017-05-18 at 6 18 22 pm

machenmusik:
> laser-controls is more a preset for the shipped controls, what use cases are you thinking for other components using it?

for example when i was adding gearvr-controls, if laser-controls had existed i would have wanted to use it.

firstly, laser-controls should be made to work for a new controller new-controls with <a-entity laser-controls="hand:right" new-controls="hand:right"> as long as new-controls follows the event naming convention of the others like vive-controls, daydream-controls, gearvr-controls.  Maybe just a default config...

secondly, it seems to me that the list of controller components in laser-controls is kinda arbitrary, and my personal preference is to make the list of controller components to add a configurable array that defaults to the ones in core, as you see with the components in my codepen above.  that way it is easy to add controllers you want, or remove ones you don't.

ngokevin:
If you see `laser-controls`, it just configures controls based on the controls that A-Frame ships for convenience. It's not a general purpose API. There's not really benefit to running your own controls through it because you are free to configure the `cursor` and `raycaster` yourself, which isn't much work. The benefit `laser-controls` provides is not having to configure every controller that A-Frame has.

Yeah, that is something maybe we can do separately, provide targets for the `downEvent`/`upEvent`. By default, it can be the `canvas`. For desktop cases, mouse down still does a click, yeah.

machenmusik:
> The benefit laser-controls provides is not having to configure every controller that A-Frame has.

Right, and that's still true when you want your new controller to be accommodated and behave the same way, which is what I was trying to provide by way of example.  I do think that if folks are going to use laser-controls with custom model then you'll need to pass through some other data like model false, maybe rotation/rotationOffset...

ngokevin:
What behavior does `laser-controls` provide that you want to use? It's just default configuration for the default controllers. If you have your own custom controls, you configure it yourself: `<a-entity custom-controls cursor="laserEnabled: true">` (or inject `cursor` from `custom-controls`.)

machenmusik:
If you want other core controllers to work but just want to add support for your new controller, the natural thing to expect, like it was for hand-controls, is to just explicitly add your new component right alongside laser-controls.

It's a good bit of syntax convenience, especially letting the core component maintain the list of core controller components itself - and really the only requirements of a new controller component to interoperate should be to generate the right events (e.g. `triggerdown` `triggerup` `controllerconnected`) -- and to update pose (orientation, and position either from arm model or explicitly if 6DOF) which comes for free if based on `tracked-controls`.  Use cases include the way we accommodated daydream-controller and gearvr-controls before incorporating into core.

examples/test/laser-controls/alignment.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Laser Controls - Alignment</title>
+    <meta name="description" content="Raycaster - A-Frame">
```
cvan: couldn't tell - is this description intentional?

examples/test/laser-controls/components/intersect-color-change.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+AFRAME.registerComponent('intersect-color-change', {
+  init: function () {
+    var el = this.el;
+    var material = el.getAttribute('material');
+    var initialColor = material.color;
+    var initialOpacity = material.opacity;
+
+    el.addEventListener('mousedown', function (evt) {
+      el.setAttribute('material', 'color', '#EF2D5E');
```
cvan: can this be a default?

examples/test/laser-controls/components/intersect-color-change.js
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    el.addEventListener('mouseup', function (evt) {
+      el.setAttribute('material', 'color', initialColor);
+    });
+
+    el.addEventListener('mouseenter', function () {
+      el.setAttribute('material', 'opacity', 1.0);
```
cvan: this as well?

src/components/cursor.js
cvan: this a breaking change, no? perhaps a deprecation first?

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
   dependencies: ['raycaster'],
 
   schema: {
+    downEvents: {default: []},
     fuse: {default: utils.device.isMobile()},
-    fuseTimeout: {default: 1500, min: 0}
+    fuseTimeout: {default: 1500, min: 0},
+    laserAlwaysShow: {default: true},
+    laserColor: {default: '#74BEC1'},
+    laserEnabled: {default: false},
```
cvan: how does this interact with `visible="false"`?

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
   dependencies: ['raycaster'],
 
   schema: {
+    downEvents: {default: []},
     fuse: {default: utils.device.isMobile()},
-    fuseTimeout: {default: 1500, min: 0}
+    fuseTimeout: {default: 1500, min: 0},
+    laserAlwaysShow: {default: true},
```
cvan: document these new properties [above](https://github.com/aframevr/aframe/pull/2678/files#diff-e9782f61decda1df23d89fcb932be260L22)?

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
...
+      canvas.addEventListener('mouseup', this.onCursorUp);
     }
+
+    data.downEvents.forEach(function (downEvent) {
+      el.addEventListener(downEvent, self.onCursorDown);
+    });
+    data.upEvents.forEach(function (upEvent) {
+      el.addEventListener(upEvent, self.onCursorUp);
+    });
+    el.addEventListener('raycaster-intersection', this.onIntersection);
```
cvan: do you think it'll be confusing for folks if the event is named `cursor` but the event names are prefixed with `raycaster-`?

examples/test/laser-controls/alignment.html
```diff
@@ -0,0 +1,30 @@
...
+      <a-entity position="-1 1.5 -2">
+        <a-entity oculus-touch-controls cursor="laserEnabled: true" raycaster="far: 2; objects: null; origin: 0 0 0.1; direction: 0 -0.4 -1"></a-entity>
+      </a-entity>
+
+      <a-entity position="0 1.5 -2">
+        <a-entity daydream-controls cursor="laserEnabled: true" raycaster="objects: null"></a-entity>
+      </a-entity>
+
+      <a-entity position="1 1.5 -2">
+        <a-entity gearvr-controls cursor="laserEnabled: true" raycaster="objects: null; origin: 0 0.005 0"></a-entity>
```
cvan: I notice there's `objects: null` quite a bit; perhaps the [default](https://github.com/aframevr/aframe/pull/2678/files#diff-d0074fac66c607708186057c046f5dbeR25) should change based on `cursor="laserEnabled: true"`?

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
   dependencies: ['raycaster'],
 
   schema: {
+    downEvents: {default: []},
```
cvan: can this be conditionally set to be `trackpaddown` and `triggerdown`? otherwise, that's quite a bit of manual work for devs

src/components/cursor.js
cvan: I don't see where `laserAlwaysShow` is used.

just thinking out loud, but what if – borrowing from CSS' `border` property – there were these properties?

- laserColor
- laserWidth
- laserStyle

if `laserWidth <= 0`, then there's no laser. `laserStyle` could indicate whether it's visible or not.

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
...
+        color: data.laserColor,
+        start: raycaster.ray.origin,
+        end: endVec3.copy(raycaster.ray.origin).add(
+          directionVec3.copy(raycaster.ray.direction).multiplyScalar(rayLength)
+        ),
+        width: data.laserWidth
+      });
+    }
+
+    if (data.laserEnabled && oldData.laserEnabled) { el.removeAttribute('line'); }
```
cvan: forgive me if I'm wrong, but won't this incorrectly remove the [`line` just set above](https://github.com/aframevr/aframe/pull/2678/files#diff-e9782f61decda1df23d89fcb932be260R73) (if `laserEnabled` is truthy on this component `update`)?

src/components/cursor.js
```diff
@@ -33,60 +38,118 @@ module.exports.Component = registerComponent('cursor', {
...
+  },
+
+  play: function () {
+    this.addEventListeners();
+  },
+
+  pause: function () {
+    this.removeEventListeners();
```
cvan: do you think `pause` should also remove the `line` attribute?

dmarcos:
Since we are going to have a proper 3DOF / 6DOF responsive example now I would get rid of the gearVR  and daydream support in `hand-controls` and make it a 6DOF only type of controls: https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L159

examples/test/laser-controls/alignment.html
donmccurdy: Is there a laser-related reason that `objects:null` is set, or is that just for this demo?

examples/test/laser-controls/index.html
```diff
@@ -0,0 +1,43 @@
...
+                             intersect-color-change></a-mixin>
+      </a-assets>
+
+      <a-entity geometry="primitive: plane; height: 50; width: 50"
+                material="color: #E2F3F5; shader: flat"
+                position="0 -1.8 0"
+                rotation="-90 0 0"></a-entity>
+
+      <a-entity id="leftHand" laser-controls="hand: left" raycaster="objects: [mixin~='box'], [mixin~='sphere']"></a-entity>
+      <a-entity id="rightHand" laser-controls="hand: right" cursor="laserAlwaysShow: false" raycaster="objects: [mixin~='box'], [mixin~='sphere']"></a-entity>
```
donmccurdy: The `laserAlwaysShow` property doesn't seem to be defined anywhere

src/components/cursor.js
```diff
@@ -33,62 +39,140 @@ module.exports.Component = registerComponent('cursor', {
...
-    canvas.addEventListener('mouseup', this.onMouseUp);
-    el.addEventListener('raycaster-intersection', this.onIntersection);
-    el.addEventListener('raycaster-intersection-cleared', this.onIntersectionCleared);
+    if (data.projectionEnabled) {
+      this.injectProjectionEl();
+      // If projection enabled, speed up interval so mesh position updates are smooth.
+      raycasterData = el.getDOMAttribute('raycaster');
+      if (!raycasterData || !('interval' in raycasterData)) {
+        el.setAttribute('raycaster', 'interval', 15);
+      }
```
donmccurdy: This seems out of scope for the `cursor` component... IMO it could be in `laser-controls`, or just suggested in docs.

examples/test/laser-controls/index.html
ngokevin: It's still a WIP, nothing is cleaned up.

src/components/cursor.js
ngokevin: This is for projecting a mesh on the intersection point, not the laser. Without it, it gets very choppy.

ngokevin:
There's not much code to draw the line that needs a separate component. There is a separate line component that handles most of it. I'm moving it to the raycaster not because of clutter, but because the line uses raycaster data to draw the line. If the raycaster gets updated, it's easier to update the line from there.

I've separately squished in https://github.com/jujunjun110/aframe-crawling-cursor as a feature as well.

machenmusik:
On the one hand crawling cursor handles the depth disparity problem that typical fixed depth gaze cursors have, and having that out of the box would be great.  On the other hand using the demo, the entire cursor entity/object is subject to the normal at the intersection point which at least for the demo causes a lot of flickering / unstable appearance when one looks at edges, which I find objectionable.  (I.e. if the shape is a ring or 2d polygon, it doesn't seem to draw the outline of the shape interpolated between the intersection poonts from a ray projected through each vertex, but rather pivots the shape around center point according to single point intersection)

machenmusik:
Going back to some earlier comments it makes me wonder whether it would be worth refactoring raycaster into shared pools of candidate objects (updated once per pool not once per raycaster) and multiple raycasters referencing - then it would be less inefficient to try an interpolated crawling polygon that actually fit itself to the surface

ngokevin:
Currently stuck on getting positionally-offset raycasters to work. I want to lower the origin of the raycaster for Oculus Touch controls. I take an offset Vector3 in object space and convert it to world space to update the raycaster origin. But it seems that this doesn't have too much of an effect.

Do three.js raycasters cast the ray using the `direction` relative to the `origin`, or is the `direction` in absolute terms?

donmccurdy:
> Do three.js raycasters cast the ray using the direction relative to the origin, or is the direction in absolute terms?

Should be absolute, since you pass the raycaster a point and not a node.

src/components/cursor.js
dmarcos: I'm not 100% sure about this since it sets a precedent of a pattern that won't be available for all the events. If we do this for `downEvents` and `upEvents` why not also for `enter` \ `leave` and why not for all the other components that emit events. this is basically a declarative way to do 
```javascript
el.addEventListener('mousedown', function () { .... emit('xxx') }
```

ngokevin: These are events to listen to for controller inputs, not for emitting events.

ngokevin:
Got it working yesterday, here's the GIF. `THREE.Raycaster.direction` is absolute, not relative to the `THREE.Raycaster.origin` which is weird. But I just apply the origin to the direction, and it works.

lightsaber

src/components/cursor.js
ngokevin: See https://github.com/aframevr/aframe/pull/2678/files#diff-3f66fed14508fae90dc58b66fac99025R36 for usage

PR #2679, opened 2017-05-19
Skip type checking if an object is passed through setAttribute twice

dmarcos:
I'm trying to find ways to optimize `setAttribute` / `getAttribute` calls specially when they're called once per frame like for example in a component's `tick` method.

My first attempt is to disable type checking when `setAttribute` is called twice with the same object. Types are checked the first time and then the object is considered trusted. This would allow to construct your `tick` method in the following way:

````javascript
  tick: (function () {
      var newRotation = {};
      return function () {
      var el = this.el;
      var currentRotation = el.getAttribute('rotation');
      newRotation .x = currentRotation.x + 0.1;
      newRotation.y = currentRotation.y + 0.1;
      newRotation.z = currentRotation.z + 0.1;
      el.setAttribute('rotation', newRotation );
    }
  })()
````
This aligns with the good practice of avoiding object allocations inside your render loop. This optimization yields a 30%-40% percent FPS improvement with respect to master when rotating 2500 cubes using a tick like the above one.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -511,9 +511,10 @@ var proto = Object.create(ANode.prototype, {
    *
    * @param {string} attr - Component name.
    * @param {object} attrValue - The value of the DOM attribute.
+   * @param {bollean} clobber - if the new attrValue will ccompletely replace the previous properties
    */
   updateComponent: {
-    value: function (attr, attrValue) {
+    value: function (attr, attrValue, clobber) {
       var component = this.components[attr];
       var isDefault = attr in this.defaultComponents;
       if (component) {
@@ -522,7 +523,7 @@ var proto = Object.create(ANode.prototype, {
           return;
         }
         // Component already initialized. Update component.
-        component.updateProperties(attrValue);
+        component.updateProperties(attrValue, clobber);
         return;
       }
       // Component not yet initialized. Initialize component.
@@ -682,6 +683,8 @@ var proto = Object.create(ANode.prototype, {
    */
   setAttribute: {
     value: function (attrName, arg1, arg2) {
+      var arg1Type = typeof arg1;
+      var clobber;
       var componentName;
       var isDebugMode;
 
@@ -690,12 +693,11 @@ var proto = Object.create(ANode.prototype, {
 
       // Determine which type of setAttribute to call based on the types of the arguments.
       if (COMPONENTS[componentName]) {
-        if (typeof arg1 === 'string' && typeof arg2 !== 'undefined') {
+        if (arg1Type === 'string' && typeof arg2 !== 'undefined') {
           singlePropertyUpdate(this, attrName, arg1, arg2);
-        } else if (typeof arg1 === 'object' && arg2 === true) {
-          multiPropertyClobber(this, attrName, arg1);
         } else {
-          componentUpdate(this, attrName, arg1);
+          clobber = arg1Type !== 'object' || (arg1Type === 'object' && arg2 === true);
+          this.updateComponent(attrName, arg1, clobber);
         }
 
         // In debug mode, write component data up to the DOM.
@@ -712,30 +714,6 @@ var proto = Object.create(ANode.prototype, {
        */
       function singlePropertyUpdate (el, componentName, propName, propertyValue) {
         el.updateComponentProperty(componentName, propName, propertyValue);
-      }
-
-      /**
-       * Just update multiple component properties at once for a multi-property component.
-       * >> setAttribute('foo', {bar: 'baz'})
-       */
-      function componentUpdate (el, componentName, propValue) {
-        var component = el.components[componentName];
-        if (component && typeof propValue === 'object') {
-          // Extend existing component attribute value.
-          el.updateComponent(
-            componentName,
-            utils.extendDeep(utils.extendDeep({}, component.attrValue), propValue));
-        } else {
-          el.updateComponent(componentName, propValue);
-        }
-      }
-
-      /**
-       * Pass in complete data set for a multi-property component.
-       * >> setAttribute('foo', {bar: 'baz'}, true)
-       */
-      function multiPropertyClobber (el, componentName, propObject) {
-        el.updateComponent(componentName, propObject);
       }
 
       /**
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -35,6 +35,8 @@ var Component = module.exports.Component = function (el, attrValue, id) {
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
   this.el.components[this.attrName] = this;
+  // Last value passed to updateProperties
+  this.previousAttrValue = undefined;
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by ngokevin** [comment]
Doesn't the closure pattern cause objects to get shared across different instances of the same type of component? We've ran into issues with that before. Sometimes I wish for a closure vs. prototype definition of components.

File: src/core/component.js [hunk 1]
```diff
@@ -390,7 +398,7 @@ module.exports.registerComponent = function (name, definition) {
  * @param {boolean} silent - Suppress warning messages.
  * @return {object} The component data
  */
-function buildData (el, name, attrName, schema, elData, silent) {
+function buildData (el, name, attrName, schema, elData, oldElData, silent, skipTypeChecking) {
```
ngokevin: We should just plop this back into the component definition, too many arguments.

dmarcos:
You should not use the closure pattern to store state of the component. In this case is just an auxiliary variable to make a temporary computation so it's fine.

fernandojsg:
30-40% improvements here in the 5k cubes scene

examples/performance/cubes/setup.js
```diff
@@ -0,0 +1,46 @@
+var URL_PARAMS = getUrlParams();
```
ngokevin: Might be nice to encapsulate this in a `main` component `<a-scene main>`

src/core/a-entity.js
```diff
@@ -513,7 +513,7 @@ var proto = Object.create(ANode.prototype, {
    * @param {object} attrValue - The value of the DOM attribute.
    */
   updateComponent: {
-    value: function (attr, attrValue) {
+    value: function (attr, attrValue, clobber) {
```
ngokevin: I'd doc-string this to note `.setAttribute` passes this in

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
    * @param {string} attrValue - HTML attribute value.
    *        If undefined, use the cached attribute value and continue updating properties.
    */
-  updateProperties: function (attrValue) {
+  updateProperties: function (attrValue, clobber) {
     var el = this.el;
     var isSinglePropSchema = isSingleProp(this.schema);
     var oldData = extendProperties({}, this.data, isSinglePropSchema);
+    var skipTypeChecking = typeof this.previousAttrValue === 'object' && attrValue === this.previousAttrValue;
```
ngokevin: I'd comment the conditions of type checking (If the passed data is an object and the object has not changed)

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
...
+    var currentAttrValue = !clobber && this.attrValue;
```
ngokevin: What's `!clobber && this.attrValue` mean? Can you add a comment

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
...
+    this.previousAttrValue = attrValue;
```
ngokevin: We can define this in the constructor as well `this.previousAttrValue = undefined`

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
...
 
-    if (attrValue !== undefined) { this.updateCachedAttrValue(attrValue); }
+    if (!el.hasLoaded) {
```
ngokevin: Can you comment this clause?

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
...
-    if (!el.hasLoaded) { return; }
+    attrValue = this.parseAttrValueForCache(attrValue);
 
     if (this.updateSchema) {
-      this.updateSchema(buildData(el, this.name, this.attrName, this.schema, this.attrValue,
-                                  true));
+      this.updateSchema(this.buildData(attrValue, currentAttrValue, true));
     }
-    this.data = buildData(el, this.name, this.attrName, this.schema, this.attrValue);
+    this.data = this.buildData(attrValue, currentAttrValue, false, skipTypeChecking);
```
ngokevin: We could have `buildData`'s third argument be an options object so we can be explicit vs passing in bools

```
buildData(attrValue, currentAttrValue, {skipTypeChecking: true, silent: false})
```

src/core/component.js
```diff
@@ -200,20 +200,27 @@ Component.prototype = {
...
+
+    if (attrValue !== undefined) { this.updateCachedAttrValue(attrValue); }
```
ngokevin: Curious when would it be undefined?

src/core/component.js
```diff
@@ -281,6 +288,73 @@ Component.prototype = {
...
+   *
+   * Finally coerce the data to the types of the defaults.
+   *
+   * @param {object} elData - Element current data.
+   * @param {object} oldElData - Element previous data.
+   * @param {boolean} silent - Suppress warning messages.
+   * @param {boolean} skipTypeChecking - Skip type checking and cohercion.
+   * @return {object} The component data
+   */
+  buildData: function (elData, oldElData, silent, skipTypeChecking) {
```
ngokevin: nit: half the code calls it elData and half calls it attrValue, should we normalize?

tests/core/a-entity.test.js
```diff
@@ -1216,6 +1213,19 @@ suite('a-entity component lifecycle management', function () {
     sinon.assert.calledOnce(TestComponent.update);
   });
 
+  test('does not check types if the same object is passed again', function () {
+    var el = this.el;
+    var componentData;
+    var attrValue = {a: 3};
```
ngokevin: `foo` is a bit better than `a`

src/core/a-entity.js
```diff
@@ -715,30 +716,6 @@ var proto = Object.create(ANode.prototype, {
       }
 
       /**
-       * Just update multiple component properties at once for a multi-property component.
```
ngokevin: The logic here and below are handled by which pieces now?

src/core/component.js
dmarcos: when the component is defined via mixins for example. there's not an attrValue on the entity

src/core/a-entity.js
dmarcos: There was a lot of redundancy in that logic. The flag `clobber` basically replaces all of it

src/core/component.js
dmarcos: it's maybe premature if we only have two flags? It's also not a pattern we have in other parts of the code

ngokevin: Not concerned about the number of flags, but being explicit, because calling a function `build(value, oldValue, true, false)`, hard to know what `true`/`false` does in that order.

But maybe not so we don't create another object.

dmarcos: `buildData` is not part of the public API and only used in two places in the whole code base so the chances of misuse are small. And also what you said. It's called only in two places but invoked on every component update so better to avoid new object instantations.

Issue #2673, opened 2017-05-18
firefox grab cursor stuck to grabbing


ngokevin:
**Description:**

A typo in the CSS reported by Liubov Lozhkina in Slack (@liubov). `-moz-grabbing` -> `-moz-grab`.

ngokevin:
https://github.com/aframevr/aframe/pull/2680

PR #2680, opened 2017-05-19
Change all -moz-grabbing to -moz-grab

khanh2003:
**Description:**
Addressing #2673

### DIFF
--- a/src/style/aframe.css
+++ b/src/style/aframe.css
@@ -35,14 +35,14 @@
 
 .a-canvas.a-grab-cursor:hover {
   cursor: grab;
-  cursor: -moz-grabbing;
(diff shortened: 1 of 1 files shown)


src/style/aframe.css
```diff
@@ -35,7 +35,7 @@
 
 .a-canvas.a-grab-cursor:hover {
   cursor: grab;
```
cvan: as it's the ideal, non-vendor-prefixed one, I would move `cursor: grab;` to be the last property-value style

khanh2003: ok then

ngokevin:
Thanks!

PR #2681, opened 2017-05-19
Add a note about color type regarding color animations

eng1neer:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Will work with animation component, perhaps better to mention `<a-animation>` or put it in the animation docs, but will just pull in for now. Thanks!

PR #2683, opened 2017-05-19
Throttle componentchanged event firing

dmarcos:
**Description:**
Second Optimization to try to optimize `setAttribute` calls. emitting events is expensive and if `setAttribute` is call on a `tick` method we don't need to fire the `componentchanged` event 60 or 90 times per second. This patch throttles down the `componentchanged` event firing to once every 200ms.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -31,11 +31,20 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
  *         mapped attribute of the component plus applying defaults and mixins.
  */
 var Component = module.exports.Component = function (el, attrValue, id) {
+  var self = this;
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
   this.el.components[this.attrName] = this;
   this.updateProperties(attrValue);
+  this.throttledEmitComponentChanged = utils.throttle(function emitComponentChanged (oldData) {
+    el.emit('componentchanged', {
+      id: self.id,
+      name: self.name,
+      newData: self.data,
+      oldData: oldData
+    }, false);
+  }, 200);
 };
 
 Component.prototype = {
@@ -240,12 +249,8 @@ Component.prototype = {
 
       // Update component.
       this.update(oldData);
-      el.emit('componentchanged', {
-        id: this.id,
-        name: this.name,
-        newData: this.getData(),
-        oldData: oldData
-      }, false);
+      // Limit event to fire once every 200ms.
+      this.throttledEmitComponentChanged(oldData);
     }
   },
 


src/core/component.js
```diff
@@ -31,11 +31,20 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
...
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
   this.el.components[this.attrName] = this;
   this.updateProperties(attrValue);
+  this.throttledEmitComponentChanged = utils.throttle(function (oldData) {
+    el.emit('componentchanged', {
+      id: self.id,
+      name: self.name,
+      newData: self.getData(),
```
ngokevin: Should we just return the data as is, and tell people to be careful to modify it? getData creates a copy each time.

src/core/component.js
```diff
@@ -31,11 +31,20 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
...
+      oldData: oldData
+    }, true);
+  }, 200);
```
ngokevin: How'd you choose 200? For 90FPS, that's 1 out of 20 frames. Is that alright if code relying on this event could be 20 frames behind in data?

src/core/component.js
```diff
@@ -240,12 +249,8 @@ Component.prototype = {
 
       // Update component.
       this.update(oldData);
-      el.emit('componentchanged', {
-        id: this.id,
-        name: this.name,
-        newData: this.getData(),
-        oldData: oldData
-      }, false);
+      // It limits the event firing to one every 200ms.
```
ngokevin: Nit: `Limit event to fire every 200ms.`

src/core/component.js
```diff
@@ -31,11 +31,20 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
  *         mapped attribute of the component plus applying defaults and mixins.
  */
 var Component = module.exports.Component = function (el, attrValue, id) {
+  var self = this;
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
   this.el.components[this.attrName] = this;
   this.updateProperties(attrValue);
+  this.throttledEmitComponentChanged = utils.throttle(function (oldData) {
```
ngokevin: Can name the function, useful for profiling

src/core/component.js
```diff
@@ -31,11 +31,20 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
...
+    el.emit('componentchanged', {
+      id: self.id,
+      name: self.name,
+      newData: self.getData(),
+      oldData: oldData
+    }, true);
```
ngokevin: Maybe we shouldn't bubble this

src/core/component.js
dmarcos: In time sensitive applications is better to poll instead of relying on events. If you want fresh data I would call `getAttribute` on the event handler.

dmarcos: 200 is arbirtrary yes. 5 times per second is actually plenty. I was tempted to limit at 2 per second. If you really want to have finer control I would just call `getAttribute` on a `tick` method instead of relying on th `componentchanged` event

ngokevin: Cool, I'd add a comment explaining that

PR #2686, opened 2017-05-20
Dispose of unused textures when material is unregistered.

### LINKED ISSUE #2166: Consider Calling "dispose()" on Texture when removing
**Description:**
When removing the src-Attribute of the material-component or when removing the entire material component from an entity, the texture remains in memory.

This causes memory leaks (especially on mobile) when dealing with a lot and/or large textures. THREE.js will remove a texture from memory when "dispose()" is called on it.

Like this:
this.el.object3DMap.mesh.material.map.dispose();

Example:

http://lab.immersiveweb.ch/aframe/

Click on the left box: Removes Material (Texture count in stats remains the same)
Click on the right box: Disposes Texture and then removes Material (Texture count in stats drops)

Should we consider adding this behaviour into the Material-Component?

- A-Frame Version: 0.3.2
- Platform/Device: All
- Reproducible Code Snippet or Demo URL [highly encouraged]: http://lab.immersiveweb.ch/aframe/

donmccurdy:
Fixes #2166.

### DIFF
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -3,6 +3,7 @@ var THREE = require('../lib/three');
 var utils = require('../utils/');
 var isHLS = require('../utils/material').isHLS;
 
+var bind = utils.bind;
 var debug = utils.debug;
 var error = debug('components:texture:error');
 var TextureLoader = new THREE.TextureLoader();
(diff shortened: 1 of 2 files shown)


ngokevin:
Perfect!

PR #2694, opened 2017-05-23
have raycaster listen to object3dset event

ngokevin:
**Description:**

Raycaster would error or not work if an object3d was not loaded by the test it gathered the objects. Listen for `object3dset` event and simple push to the `objects` array for intersection testing. I think I still get the error, but at least the raycaster works.

- [ ] tests

Seeing if it fixes for @bryik

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -54,12 +54,21 @@ module.exports.Component = registerComponent('raycaster', {
   update: function () {
     var data = this.data;
     var raycaster = this.raycaster;
+    var self = this;
 
     // Set raycaster properties.
     raycaster.far = data.far;
     raycaster.near = data.near;
 
     this.refreshObjects();
+
+    // Pick up any `Object3D`s that get set later (e.g., after model loading).
+    this.objects.forEach(object => {
+      if (!object.el) { return; }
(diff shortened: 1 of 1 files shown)


src/components/raycaster.js
```diff
@@ -54,12 +54,21 @@ module.exports.Component = registerComponent('raycaster', {
...
     raycaster.far = data.far;
     raycaster.near = data.near;
 
     this.refreshObjects();
+
+    // Pick up any `Object3D`s that get set later (e.g., after model loading).
+    this.objects.forEach(object => {
+      if (!object.el) { return; }
+      object.el.addEventListener('object3dset', function (evt) {
+        self.objects.push(evt.detail.object);
```
machenmusik: This is only correct to do if the object's element still satisfies the gathering criteria by the time the model finishes loading, right?  for example, the object has not changed to a non-selectable css class while the object was loading

machenmusik: (because nothing ever removes the listener if the object becomes non-selectable)

ngokevin:
The error was because obj-model set material as a dependency which created an empty BufferGeometry which three.js didn't like.

PR #2704, opened 2017-05-25
reduce memory in raycaster tick

ngokevin:
**Description:**

**Changes proposed:**
- Reuse a single array rather than copying with `.slice`.
- Reuse a single array rather than using `.map`.
- Don't copy intersectedEls in event emit.
- [x] Tests pass

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -30,6 +30,7 @@ module.exports.Component = registerComponent('raycaster', {
     this.intersectedEls = [];
     this.objects = null;
     this.prevCheckTime = undefined;
+    this.prevIntersectedEls = [];
     this.raycaster = new THREE.Raycaster();
     this.updateOriginDirection();
     this.refreshObjects = bind(this.refreshObjects, this);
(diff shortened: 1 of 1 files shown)


dmarcos:
Are there any measurable results that can be seen?

src/components/raycaster.js
```diff
@@ -143,9 +145,10 @@ module.exports.Component = registerComponent('raycaster', {
     });
 
     // Update intersectedEls.
-    intersectedEls = this.intersectedEls = intersections.map(function getEl (intersection) {
-      return intersection.object.el;
-    });
+    intersectedEls.length = intersections.length;
```
dmarcos: Do we know the effects of writting the length propery of an array? Are elements disposed if the new length is smaller than the old one?

ngokevin: The intent was to avoid creating new arrays, not about disposing elements. The elements will still be there, but at least we reuse the same array.

ngokevin:
I'm having trouble measuring any noticable improvements, but the changes seem at least better for best practice.

Issue #2614, opened 2017-04-21
Enter VR button disappears after exit VR, in Samsung & Oculus browsers


mkeblx:
Can observe it is momentarily shown before being hidden.

- A-Frame Version: 0.5 / master
- Platform / Device: Oculus Browser - Gear VR
- Reproducible Code Snippet or URL:
https://aframe-master.glitch.me

cvan:
@machenmusik, @caseyyee, or @mkeblx (!) – someone who has a functional Oculus Browser device: can you take a looksie?

(I'm waiting on a new Samsung Gear VR case)

mkeblx:
Will do.

cvan:
thanks, @mkeblx. I'll debug it once I get my new Samsung Gear VR; if you (or someone else) can get to it before then, sweet!

cutec-chris:
How is it possible to debug this ? 
Does entering VR work for someone ? For me the Oculus Browser only shows an Window with the Scene.

weddingdj:
I tested aframe.io in the new Gear VR 2017 and the VR enter button disappears even before I can click on it to enter VR. The site loads, the button is visible for a brief moment, then it disappears. Mabye it has something to do with the mobile version / css of aframe.io

machenmusik:
I wonder if the old code to auto-enter VR (by necessity) for Carmel Developer Preview is trying and failing with Oculus Browser.  But previously for me pages were loading in 2D with enter-VR button visible just fine with Oculus Browser.  Did this just change recently?

dmarcos:
Can we test with 0.4.0?

machenmusik:
On the assumption that most will be using Oculus browser now, maybe we should remove auto-enter-vr from default scene components, and those that need can add manually? @dmarcos if you like that approach I can do quick PR

weddingdj:
I investigated and it looks like if the exit-vr event is not triggered.

weddingdj:
When loading a scene:
![screen shot 2017-06-07 at 22 18 57](https://user-images.githubusercontent.com/3025886/26909579-b58af476-4bcf-11e7-8c33-a91c1b310402.png)

If I touch the touch pad of the controller:
![screen shot 2017-06-07 at 22 20 27](https://user-images.githubusercontent.com/3025886/26909587-c2d22c9e-4bcf-11e7-981c-3fdf270ba266.png)

No output when entering or leaving VR mode.

dmarcos:
[...]

Issue #2659, opened 2017-05-08
GearVR Oculus Browser - Hides VR Button, but fails to enter VR


shaunmnemonic:
**Description:**

When loading in the Oculus browser the 'EnterVR Goggles' get hidden because the `auto-enter.js` file tries to auto-enter-vr, but requires a user interaction and the goggles don't re-appear. 

- A-Frame Version: 0.5.0
- Platform / Device: GearVR, Oculus Browser
- Reproducible Code Snippet or URL: Any of the Aframe.io Examples

mkeblx:
This doesn't occur on master.

dmarcos:
@shaunmnemonic Please reoopen if you can reproduce on master

PR #2707, opened 2017-05-26
no longer auto-enter-vr

machenmusik:
Now that Oculus Browser has a user interface both out of VR and inside VR, and effectively deprecates the previous Carmel Developer Preview which only worked in VR (and thus required auto-entering VR to be usable), remove the `auto-enter-vr` component from the default list of scene components.  To keep the prior behavior, authors need only specify the component explicitly e.g. `a-scene auto-enter-vr`

Hopefully this will fix issues encountered such as https://github.com/aframevr/aframe/issues/2614#issuecomment-304334866

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -43,8 +43,7 @@ module.exports.AScene = registerElement('a-scene', {
         'inspector': '',
         'keyboard-shortcuts': '',
         'screenshot': '',
-        'vr-mode-ui': '',
-        'auto-enter-vr': ''
+        'vr-mode-ui': ''
       }
     },
 


dmarcos:
I don't have a GearVR around. Can anyone verify that this fix the issue?

machenmusik:
I am a little confused over whether this is still a problem or not given https://github.com/aframevr/aframe/issues/2659 versus the comment referenced above.  Wondering if recent changes in oculus browser caused regression...

machenmusik:
Using the anime UI example from a-frobot examples link, with the latest version 3.3.2 of Oculus Browser, the example behaves as expected without this change.

machenmusik:
Closing for now as we don't seem to need it at present

machenmusik:
Reopening temporarily in case it is needed for #2614

ngokevin:
I don't think it's needed for #2614. #2614 is because A-Frame is not handling GearVR back button (nor should it, specifically).

ngokevin:
Closing for now, can re-open if there's a need for this, but I don't know the details.

PR #2713, opened 2017-05-30
check if component is initialized before removing

ngokevin:
**Description:**

Was running into errors in unit test if you detached an entity while one of its components was still initializing. Its data would be `null` because it had not yet built data, and `.removeComponent` would be called. Geometry would then be calling `system.unuseGeometry(data)` where `data` was null, causing an error.

**Changes proposed:**
- Wait for `componentinitialized` in `removeComponent` if not initialized.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -436,6 +436,16 @@ var proto = Object.create(ANode.prototype, {
       component = this.components[name];
       if (!component) { return; }
 
+      // Wait for component to initialize.
+      if (!component.initialized) {
+        this.addEventListener('componentinitialized', function tryRemoveLater (evt) {
+          if (evt.detail.name !== name) { return; }
+          this.removeComponent(name);
+          this.removeEventListener('componentinitialized', tryRemoveLater);
+        });
+        return;
+      }
+
       component.pause();
       component.remove();
       delete this.components[name];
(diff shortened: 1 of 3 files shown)


src/core/a-entity.js
```diff
@@ -435,14 +435,28 @@ var proto = Object.create(ANode.prototype, {
       if (isDefault || isMixedIn) { return; }
 
       component = this.components[name];
+
+      if (!component) { return; }
+
+      // Wait for component to initialize.
+      if (!component.initialized) {
```
dmarcos: What about doing just?
````javascript
 if (!component || !component.initialized) { return; }
````

Comment (ngokevin) [inline src/core/a-entity.js]: Then the `removeComponent` call will never go through. A warning might be simpler though.

**Comment by cvan** [comment]
@ngokevin: sorry, I think Waffle automatically added that on my behalf

File: src/core/a-entity.js (hunk 1 above)
Comment (ngokevin) [inline src/core/a-entity.js]: Maybe it's best to try to make sure that component gets cleaned up. Might cause leaks or unexpected behavior if we let it be orphaned in the case entities are being attached/detached before the component initializes?

Comment (dmarcos) [inline src/core/a-entity.js]: Asynchrony might bite you. What happens if you do:
````javascript
myMethod () {
this.setAttribute('my-component', { });
this.removeAttribute('my-component');
this.setAttribute('my-component', {});
}````
I believe the component will be removed if the component was not initialized in the first call
```

ngokevin: Initialization is already synchronous. This is only in the case when you do appendChild/removeChild because that's inherently asynchronous.

PR #2715, opened 2017-06-01
Fixed issue where light fails to set already loaded target

RSpace:
**Description:**

By switching out models and lights dynamically in a scene, I have exposed an issue in the light component where a null reference exception occurs when creating a new spot or directional light where the target is already loaded.

`setLight` calls `getLight` which calls `onSetTarget`. `onSetTarget` expects `this.light` to be present, which works as long as the target is not loaded, since an event listener is set up in `getLight`, giving `this.light` time to get set in `setLight`. When the target is not loaded, however, `onSetTarget` is called immediately and `this.light` is undefined.

**Changes proposed:**
- Added tests to expose the issue
- Call `onSetTarget` with the new light to set the target on, instead of relying on `this.light`
- ~~Continue to also set target on `this.light` if present (this was necessary to get all tests passing, not sure why)~~

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -88,9 +88,9 @@ module.exports.Component = registerComponent('light', {
             } else {
               // Target specified, set target to entity's `object3D` when it is loaded.
               if (value.hasLoaded) {
-                self.onSetTarget(value);
+                self.onSetTarget(value, light);
               } else {
-                value.addEventListener('loaded', bind(self.onSetTarget, self, value));
+                value.addEventListener('loaded', bind(self.onSetTarget, self, value, light));
               }
             }
             break;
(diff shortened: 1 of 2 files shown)


src/components/light.js
```diff
@@ -255,8 +255,9 @@ module.exports.Component = registerComponent('light', {
     }
   },
 
-  onSetTarget: function (targetEl) {
-    this.light.target = targetEl.object3D;
+  onSetTarget: function (targetEl, light) {
+    if (this.light) { this.light.target = targetEl.object3D; }
```
dmarcos: Can we just rely on the `light` parameter? Do we need to check both this.light and light?
````javascript
light.target = targrtEl.object3D
````

RSpace: @dmarcos Right, I made a simplification and all tests still pass.

dmarcos:
Thanks!

PR #2716, opened 2017-06-02
Added pinch-control in the look-controls component

tbalouet:
**Description:**
enables the user to pinch in and out to move forward/backward on mobile

**Changes proposed:**
- Add a dolly to modify position on mobile
- Enable 2 fingers touch for pinch
- Get fingers distance on touchmove to figure pinch in/out and move on z axis accordingly

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -107,6 +107,7 @@ module.exports.Component = registerComponent('look-controls', {
     // The canvas where the scene is painted
     this.mouseDown = false;
     this.pitchObject = new THREE.Object3D();
+    this.dollyObject = new THREE.Object3D();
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks for submitting. This should be a separate component outside of the core. It handles position so it wouldn't have been good for look-controls anyways which handles rotation. wasd-controls handles position, so your component could be used alongside that. `<a-entity camera look-controls wasd-controls pinch-controls>`. I hope it's fine to close; in the future, you can file an issue beforehand so we don't waste your time :)

tbalouet:
Ok, thanks for letting me know! I thought it was a nice added feature but I should've checked with you beforehand indeed. Hope I haven't wasted your time too much. Thanks!

PR #2720, opened 2017-06-03
#2280 fix

BhuvanaMeenakshi:
Description:

this.el.setAttribute('mixin' , '');
is throwing an error "'querySelector' on 'Element': 'a-mixin#' is not a valid selector."

Changes proposed:
Adding another attribute can eliminate the error

this.el.setAttribute('mixin', this.data.mixin);


dmarcos:
You should submit a PR from the latest master and only including the commit or commits with the proposed changes. As It is we cannot see what changes you're trying to make

ngokevin:
This is still from the wrong branch. You need to submit to master branch not docs-v0.5.0

Issue #2721, opened 2017-06-03
Ambient occlusion maps require second UV channel


donmccurdy:
In the process of testing baked occlusion/roughness/metallic, remembered there's a gotcha with three.js's AO implementation: the occlusion map requires a 2nd set of UVs.

Here's a sample model: [metal-rough-cube.zip](https://github.com/aframevr/aframe/files/1050005/metal-rough-cube.zip)

```html
<a-entity obj-model="obj: #metal-rough-cube"
          material="src: ./metal-rough-cube/metalRoughCube_DefaultMaterial_BaseColor.png;
                    normalMap: ./metal-rough-cube/metalRoughCube_DefaultMaterial_Normal.png;
                    roughnessMap: ./metal-rough-cube/metalRoughCube_DefaultMaterial_OcclusionRoughnessMetallic.png;
                    metalnessMap: ./metal-rough-cube/metalRoughCube_DefaultMaterial_OcclusionRoughnessMetallic.png;
                    ambientOcclusionMap: ./metal-rough-cube/metalRoughCube_DefaultMaterial_OcclusionRoughnessMetallic.png"></a-entity>
```

The AO doesn't show up (easier to see this if you only have ambient lighting).

I can "fix" that with this line:

```js
var geometry = mesh.geometry;
geometry.addAttribute( 'uv2', new THREE.BufferAttribute( geometry.attributes.uv.array, 2 ) );
```

That's a requirement I wouldn't expect most users to know about... Do we have any demos using AO through the `material` component? Not sure what we should do here. Some options:

1. Document this in the `material` component docs.
2. Try to detect this in the `obj-model` component and duplicate the UVs.
3. ~~Detect this in the `material` component and duplicate the UVs.~~
4. Remove `material.ambientOcclusionMap` and let the `*-model` components deal with it.

With THREE.GLTF**2**Loader we automatically duplicate UVs, so this is a non-issue. THREE.GLTFLoader, currently used by A-Frame, doesn't support PBR or AO.

takahirox:
> Try to detect this in the obj-model component and duplicate the UVs.
> Detect this in the material component and duplicate the UVs.

I'm wondering if Three.js can do and opened an issue there https://github.com/mrdoob/three.js/issues/11672

Even if Three.js does that in r87, we need an workaround til then tho.

In the options Don suggested, I don't really like letting users manually do this so vote for 2 or 3.

takahirox:
Note from the Three.js thread

[...]

PR #2722, opened 2017-06-03
[material] Add metalnessMap and roughnessMap.

### LINKED ISSUE #2709: Support Metalness and Roughness from Texture
The physical based rendering of A-Frame is limited because these values can only come from integers. Popular physical based pipelines like Substance Designer generate texture maps for diffuse, roughness, metalness, and normal.

donmccurdy:
Fixes #2709.

But, maybe this needs more discussion. See #2721. 

1) The occlusion in my sample model doesn't actually work until you duplicate the UV channel.
2) The metalness and roughness in my sample aren't visible, but definitely were applied to `mesh.material`. That _probably_ just means I exported the texture wrong (I suck at Substance Painter) but it would be nice if someone could verify with a known-correct texture.

All in all, mixed feelings about whether these more advanced textures should be in the `material` component or just left to three.js loaders. Documenting good export processes for well-defined model formats is hard enough. 🙃

[metal-rough-cube.zip](https://github.com/aframevr/aframe/files/1050019/metal-rough-cube.zip)

### DIFF
--- a/src/shaders/standard.js
+++ b/src/shaders/standard.js
@@ -26,7 +26,11 @@ module.exports.Shader = registerShader('standard', {
 
     fog: {default: true},
     height: {default: 256},
+
     metalness: {default: 0.0, min: 0.0, max: 1.0},
+    metalnessMap: {type: 'map'},
+    metalnessTextureOffset: {type: 'vec2'},
+    metalnessTextureRepeat: {type: 'vec2', default: {x: 1, y: 1}},
 
     normalMap: {type: 'map'},
     normalScale: {type: 'vec2', default: {x: 1, y: 1}},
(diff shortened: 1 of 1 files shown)


dmarcos:
Could we have at least a working sample as part of the aframe examples?

ngokevin:
@cvan Why are labels being created?

donmccurdy:
> Could we have at least a working sample as part of the aframe examples?

If we plan to merge this, sure. But mostly I just wanted to add context to the discussion in #2709. Do we want to support this?

dmarcos:
I don't have a strong opinion. Have we seen much demand besides the people involved in #2709?

donmccurdy:
No, none actually.

donmccurdy:
Will close this, and re-open if we come to another conclusion in the issue.

donmccurdy:
Re-opening. I think it may be better and less surprising to just match the available parameters of MeshStandardMaterial and make this available for users working with OBJs.

donmccurdy:
> Could we have at least a working sample as part of the aframe examples?

I'm inclined to say this feature should be available but not advertised in its own example.

dmarcos:
@donmccurdy What do we do with this?

donmccurdy:
I'd vote to merge it, for parity with MeshStandardMaterial. Users could create an OBJ in any program, texture it in Substance Painter, and wire up their metal/rough textures with the `material` component. So it's a plausible workflow, but not one I really want to recommend or advertise with an example unless people are using it successfully.

dmarcos:
Thanks!

PR #2726, opened 2017-06-05
fix setAttribute updates clobbering previously set component data (fixes #2725)

### LINKED ISSUE #2725: component DOM data is getting clobbered on the second getAttribute
**Description:**

Most likely from the patch that simplified setAttribute logic.

1. Have a component with DOM data `<a-entity geometry="primitive: box; width: 5">`
2. `setAttribute('geometry', {depth: 10, height: 10})`
3. `setAttribute('geometry', {depth: 20, height: 20})`
4. Width is now reset back to 1

Has something to do with the whole `attrValue` and `previousAttrValue` caching.

ngokevin:
**Description:**

setAttribute was refactored in https://github.com/aframevr/aframe/commit/1a45e04 to make things look simpler. But logic to preserve previous component data was wiped out in the process.

**Changes proposed:**
- Revert setAttribute workflow back to being a bit more explicit. Restore the merging of the previous component data in the current data set. But this type, don't create new objects to preserve type checking.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -511,7 +511,7 @@ var proto = Object.create(ANode.prototype, {
    *
    * @param {string} attr - Component name.
    * @param {object} attrValue - The value of the DOM attribute.
-   * @param {bollean} clobber - if the new attrValue will ccompletely replace the previous properties
+   * @param {boolean} clobber - If new attrValue completely to replace previous properties.
    */
   updateComponent: {
     value: function (attr, attrValue, clobber) {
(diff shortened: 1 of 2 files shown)


dmarcos:
I think I had considered this case. the previous attribute is preserved on `component.attrValue` and should be honored and extended if the `clobber` flag is not `true`:

https://github.com/aframevr/aframe/blob/master/src/core/component.js#L339

Your patch is doing something similar but at the entity level. Is it redundant?

ngokevin:
I'll poke to see why that wasn't working.

ngokevin:
`attrValue` is getting blown away with each update. Before wasn't a problem since it was at entity-level. Will see if I can resolve that, but it feels hairy.

PR #2727, opened 2017-06-05
have setAttribute wait for existing DOM data to init component

ngokevin:
**Description:**

(reopened to update GitHub after last outage)

Consider `<a-entity foo bar="a: 10">` where `foo` does `setAttribute('bar', {b: 20})`. The resulting data should be `{a: 10, b: 20}`, but the data ends up `{b: 20}` because the `setAttribute` takes over the initialization process and the DOM data gets ignored.

Ran into this when I had `raycaster` set a `line` component with `start` and `end`. I was trying to set the `color` and `width` via HTML, but that data was wiped out. `<a-entity raycaster="showLine: true" line="color: red; width: 0.05">`.

This would have also helped cases where we were setting `dependencies` to satisfy this lifecycle. But that would initialize the component in cases we might not want to (e.g., `raycaster` might not always want a line). Before we had `obj-model` set a dependency on `material` to use its data *if* defined. We ended up having to remove that.

**Changes proposed:**
- In setAttribute, check if the component is defined in the DOM and has not yet been initialized. If so, then initialize it immediately using DOM data before applying new data.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -510,14 +510,15 @@ var proto = Object.create(ANode.prototype, {
    * When initializing, we set the component on `this.components`.
    *
    * @param {string} attr - Component name.
-   * @param {object} attrValue - The value of the DOM attribute.
-   * @param {bollean} clobber - if the new attrValue will ccompletely replace the previous properties
+   * @param {object} attrValue - Value of the DOM attribute.
+   * @param {boolean} clobber - If new attrValue completely replaces previous properties.
    */
   updateComponent: {
     value: function (attr, attrValue, clobber) {
       var component = this.components[attr];
       var isDefault = attr in this.defaultComponents;
       if (component) {
+        // Remove component.
         if (attrValue === null && !isDefault) {
           this.removeComponent(attr);
           return;
@@ -526,6 +527,7 @@ var proto = Object.create(ANode.prototype, {
         component.updateProperties(attrValue, clobber);
         return;
       }
+
       // Component not yet initialized. Initialize component.
       this.initComponent(attr, attrValue, false);
     }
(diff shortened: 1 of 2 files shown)


dmarcos:
I see the level of a asynchronous a potential problem in the future. Is there any way to make it synchronous?  is not calling `updateComponent` enough?

src/core/a-entity.js
```diff
@@ -686,28 +688,43 @@ var proto = Object.create(ANode.prototype, {
...
+        normalSetAttribute(this, attrName, arg1);
+        return;
+      }
 
-        // In debug mode, write component data up to the DOM.
-        isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
-        if (isDebugMode) { this.components[attrName].flushToDOM(); }
+      // Wait for compoent to be initialized with DOM data.
+      if (!this.components[attrName] && this.hasAttribute(attrName)) {
+        this.addEventListener('componentinitialized', function waitForInit (evt) {
```
dmarcos: Can we do?
```javascript
this.updateComponent('attrName');
this.setAttribute(attrName, arg1);
```

dmarcos: This would force the initialization of the pending component and then we can set it.

ngokevin: That sort of makes the lifecycle unpredictable. I'd be more comfortable letting things initialize from the normal `attachedCallback` flow (`attachedCallback` -> `load` -> `updateComponents`), and then try to set. Clear order of `init` everything first, then run `update`s.

Forcing an update would change the order of things, and thinking about stuff like `dependencies`.

ngokevin: Might work. I can try. `dependencies` doesn't matter since `initComponent` will set them up. In the case another component depends on it, it's fine to `init` it a little early.

ngokevin: OK, updated.

src/core/a-entity.js
```diff
@@ -686,28 +688,37 @@ var proto = Object.create(ANode.prototype, {
...
-        // In debug mode, write component data up to the DOM.
-        isDebugMode = this.sceneEl && this.sceneEl.getAttribute('debug');
-        if (isDebugMode) { this.components[attrName].flushToDOM(); }
+      // Not a component.
+      if (!COMPONENTS[componentName]) {
+        normalSetAttribute(this, attrName, arg1);
         return;
+      }
+
+      // Initialize component first if not yet initialized.
```
ngokevin: All the real changes are in this clause. The rest was collapsing the indent with an early return.

PR #2729, opened 2017-06-05
don't clobber previous component attrValue (fixes #2725)

### LINKED ISSUE #2725: component DOM data is getting clobbered on the second getAttribute

ngokevin:
**Description:**

Previous attempt at https://github.com/aframevr/aframe/pull/2726

There was previously logic at the Entity level for updates to merge the existing properties into the new data set. That logic was removed since more data-building logic now lives in the component. But there was no logic to merge the existing properties when it came to the cached `attrValue`.

In the future, I think it might be nice to rename that to `rawData`.

**Changes proposed:**
- When updating `this.attrValue`, merge in the previous data if not clobbering.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -156,12 +156,25 @@ Component.prototype = {
   /**
    * Update the cache of the pre-parsed attribute value.
    *
-   * @param {string} value - HTML attribute value.
+   * @param {string} value - New data.
+   * @param {boolean } clobber - Whether to wipe out and replace previous data.
    */
-  updateCachedAttrValue: function (value) {
-    var isSinglePropSchema = isSingleProp(this.schema);
+  updateCachedAttrValue: function (value, clobber) {
     var attrValue = this.parseAttrValueForCache(value);
+    var isSinglePropSchema = isSingleProp(this.schema);
+    var property;
+
     if (value === undefined) { return; }
+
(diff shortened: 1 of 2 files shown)


dmarcos:
Good stuff!

gregfagan:
Thanks for this fix! I was taking a look at it a few days back, and I noticed that `a-entity.updateComponentProperty` still performs property merging for the case of single property updates:

https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L551

This should be duplicated work now; instead of `extend`ing the previous `attrValue` object, `updateComponentProperty` could just call into `updateComponent` (which it already does) with a new simple object, `{[property]: propertyValue}`.

This is most likely a small amount of overhead, but it is something to think about if this area gets touched again.

ngokevin:
Yeah, perhaps, want to try to remove it?

gregfagan:
Sure thing! Check out https://github.com/aframevr/aframe/pull/2743

PR #2734, opened 2017-06-07
Set this.throttledEmitComponentChanged before calling updateProperties

dmarcos:
If the `update` method calls `setAttribute` on the same component the `componentChanged` is emitted but the method is not properly set when `updateProperties` is called in the component constructor.

### DIFF
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -585,6 +585,19 @@ suite('Component', function () {
       component.updateProperties({list: ['b']});
       sinon.assert.calledOnce(updateStub);
     });
+
+    test('emit componentchanged when update calls setAttribute', function (done) {
+      var TestComponent = registerComponent('dummy', {
+        schema: {color: {default: 'red'}},
+        update: function () { this.el.setAttribute('dummy', 'color', 'blue'); }
+      });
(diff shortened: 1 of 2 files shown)


tests/core/component.test.js
```diff
@@ -585,6 +585,19 @@ suite('Component', function () {
       component.updateProperties({list: ['b']});
       sinon.assert.calledOnce(updateStub);
     });
+
+    test('emit componentchanged when update calls setAttribute', function (done) {
+      var TestComponent = registerComponent('dummy', {
```
ngokevin: Can name this component `test` so it gets automatically cleaned up by the test suite.

tests/core/component.test.js
```diff
@@ -585,6 +585,19 @@ suite('Component', function () {
...
+        schema: {color: {default: 'red'}},
+        update: function () { this.el.setAttribute('dummy', 'color', 'blue'); }
+      });
+      this.el.addEventListener('componentchanged', function (evt) {
+        assert.equal('blue', evt.detail.newData.color);
```
ngokevin: can swap the two arguments to be consistent `assert (variable, value)`

ngokevin:
r+wc

Issue #2711, opened 2017-05-30
Make GearVR controller arm parameters exposed so they can be modified


UXVirtual:
When the ` gearvr-controls` component is used and tested on actual hardware, the controller model rotates around a point in space outside the models geometry. I haven't tested with another model yet to determine if it's an issue with the GearVR controller model itself or if it's an a-frame bug or polyfill issue.

- A-Frame Version: master (70875d8fef84ac754b465d506dfe64ad62e68f29)
- Platform / Device: Samsung Galaxy S7 Edge / GearVR 2017 Controller / Oculus Internet Browser
- Reproducible Code Snippet or URL: https://hazardu5.github.io/aframe-workshop/examples/19/finish

machenmusik:
> the controller model rotates around a point in space outside the models geometry.
Since the Gear VR Controller is 3DOF, its positional information is approximated using a simplistic arm model, which you can see at https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls.js.  By specific request during PR discussion, the parameters of the arm model were removed from the component schema, so unfortunately it is not as easy as one would like to tweak or turn off.

machenmusik:
@HAZARDU5 so I believe the behavior you are describing is the correct functionality of the arm model for 3DOF and not a bug.  Are you making a request to make the arm model parameters accessible (or capability to turn off the arm model) ?  If so, we should probably change the issue title.

HAZARDU5:
Yeah I guess so - could we expose the arm model parameters?

machenmusik:
@dmarcos thoughts re: exposing controls for 3DOF arm model?  goes back to prior PR discussions a bit, I know you and @ngokevin preferred keeping private, but unless they will be maintained as part of core, may be better to open up or perhaps componentize

HAZARDU5:
Customizing the arm model could be useful in instances where you want the controller's gyro sensors to directly control an object in a scene and have the controller rotate around a point at 0,0,0 inside the model. E.g. I use the laser pointer to take a model toy off a shelf then it switches to a zoomed in view of the model that follows the camera and allows the toy to follow the rotation of the controller 1:1. Or for another example, the user can rotate the controller to tilt a labyrinth style maze where a ball, affected by physics, must be guided to the end of the maze. It'd be great to be able to do this without having to do some serious hacks to correct for the arm model offsets during transformations.

dmarcos:
[...]

PR #2746, opened 2017-06-08
add armModel and tests for daydream-controls and gearvr-controls

machenmusik:
Addresses #2711 by allowing one to disable the 3DOF arm model for `tracked-controls`, `daydream-controls` and `gearvr-controls` using the property `armModel`.


machenmusik:
Also, as it turns out, neither `daydream-controls` nor `gearvr-controls` was handling `buttonchanged` events and emitting e.g. `trackpadchanged`, `triggerchanged` -- however those are not commonly used so probably no one noticed...

dmarcos:
Nice! Thank you!

PR #2747, opened 2017-06-08
eliminate need to mock HTMLElement in node

kellyselden:


kellyselden:
It makes it easier to run in Node, because you have to provide these mocked globals before requiring aframe. For example: https://github.com/kellyselden/ember-aframe/blob/master/index.js#L15-L37

PR #2748, opened 2017-06-09
setAttribute: partial update with strings [fix #2651]

### LINKED ISSUE #2651: Doing `.setAttribute(component, 'foo: bar')` may clobber properties
**Description:**

http://stackoverflow.com/questions/43788105/issue-with-aframe-where-setattribute-reset-previous-proprieties

When passing a string as the second argument for a multi-prop component, A-Frame may be clobbering/resetting existing properties. Update `Entity.setAttribute` to consider this case.

- A-Frame Version: 0.5.0
- Platform / Device:
- Reproducible Code Snippet or URL: http://codepen.io/deva/pen/QvMeom

gregfagan:
* Enable the entity setAttribute API to accept partial updates of multi-property components when using a CSS-style property string
* Enable the use of the `clobber` flag under same scenario
* Remove single-use helper functions in a-entity for updating a single property of multi-property components
* Add unit tests and update a test where behavior changed

Details:

Previously, it was only possible to do partial updates when passing objects. Extending this behavior (and the clobber flag) to CSS-style strings requires differentiating those strings from simple property names, as a string in the second parameter could be either one.

This is handled by calling out to `utils.styleParser` and checking the return type, which is redundant to work done inside the component. This inefficiency could be reduced perhaps by reworking slightly the interface between the entity and component to pass the parsed result down; or a different mechanism of differentiation could be devised.

However, using strings in either capacity (CSS-style or the single property update form) is inherently a less efficient convenience for the object form, and their use should be discouraged when performance matters (e.g. `tick`).

The `clobber` flag (third parameter to setAttribute) behavior in this scenaro has now changed: previously, any time the second parameter was anything other than an object, `clobber` was treated internally as true. Now, CSS-style strings are treated the same as objects, where it defaults to false but is configurable by parameter.

This means that `el.setAttribute(‘myComponent’, ‘’)` previously would reset the component to its default values (because it was clobbering and not passing any new values). To get the same behavior as before, use `el.setAttribute(‘myComponent’, ‘’, true)`. One unit test was updated to reflect this change.

Finally, `entity.updateComponentProperty` has very little logic, is not a public API, and was only called by `setAttribute`; so it was simply merged in and its unit tests removed.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -535,21 +535,6 @@ var proto = Object.create(ANode.prototype, {
   },
 
   /**
-   * Updates one property of the component
-   *
-   * @param {string} name - Component name
-   * @param {string} property - Component property name
-   * @param {any} propertyValue - New property value
-   */
-  updateComponentProperty: {
-    value: function (name, property, propertyValue) {
-      var attrValue = {};
-      attrValue[property] = propertyValue;
-      this.updateComponent(name, attrValue);
-    }
-  },
-
-  /**
    * If `attr` is a component name, detach the component from the entity.
    *
    * If `propertyName` is given, reset the component property value to its default.
@@ -676,13 +661,14 @@ var proto = Object.create(ANode.prototype, {
    * 4. Set a value for a single-property component, mixin, or normal HTML attribute.
    *
    * @param {string} attrName - Component or attribute name.
-   * @param {string|object} arg1 - Can be a property name or object of properties.
-   * @param {string|bool} arg2 - Can be a value, or boolean indicating whether to update or
-   *   replace.
+   * @param {*} arg1 - Can be a value, property name, CSS-style property string, or
+   *   object of properties.
+   * @param {*|bool} arg2 - If arg1 is a property name, this should be a value. Otherwise,
+   *   it is a boolean indicating whether to clobber previous values (defaults to false).
    */
   setAttribute: {
     value: function (attrName, arg1, arg2) {
-      var arg1Type = typeof arg1;
+      var newAttrValue;
       var clobber;
       var componentName;
       var delimiterIndex;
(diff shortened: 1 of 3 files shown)


src/core/a-entity.js
```diff
@@ -704,29 +690,33 @@ var proto = Object.create(ANode.prototype, {
...
 
-      // Determine which type of setAttribute to call based on the types of the arguments.
-      if (arg1Type === 'string' && typeof arg2 !== 'undefined') {
-        singlePropertyUpdate(this, attrName, arg1, arg2);
+      // Determine new attributes from the arguments
+      if (
+        typeof arg2 !== 'undefined' &&
+        typeof arg1 === 'string' &&
+        arg1.length > 0 &&
+        (typeof utils.styleParser.parse(arg1)) === 'string'
```
gregfagan: Thoughts for alternatives here? Should the work be done to `component.updateProperties` to allow this parsed value to be passed directly (and if so, would it make more sense to go ahead and use `component.parse`)?

src/core/a-entity.js
```diff
@@ -704,29 +690,33 @@ var proto = Object.create(ANode.prototype, {
                              HTMLElement.prototype.getAttribute.call(this, attrName));
       }
 
-      // Determine which type of setAttribute to call based on the types of the arguments.
-      if (arg1Type === 'string' && typeof arg2 !== 'undefined') {
-        singlePropertyUpdate(this, attrName, arg1, arg2);
+      // Determine new attributes from the arguments
+      if (
+        typeof arg2 !== 'undefined' &&
```
dmarcos: Can we move this to the line above?

src/core/a-entity.js
dmarcos: This adds quit a bit of overhead on each `setAttribute` call. Can we somehow invert the logic (is the condition for the else branch less expensive) so all these conditions are not run unless necessary?

gregfagan: I ordered the clauses of the `if` to short circuit as favorably as possible, so that the overhead of the final clause is only encountered in the case where it is unavoidable: you've passed 3 arguments, and the 2nd is a non-empty string.

Because the single property update case has the most restrictions on its arguments, I think it is probably more elegant to test for.

If you *are* passing 3 arguments with the 2nd a string, the overhead must exist unless we move the logic of differentiation down to the component (where the parsed data can be reused) or redesign the API to not be ambiguous for strings (e.g. not accept a property name in the 2nd parameter at all).

src/core/a-entity.js
gregfagan: Do I understand correctly that your preference is to combine lines 694 and 695? As in:

```
if (typeof arg2 !== 'undefined' &&
  typeof arg1 === 'string' &&
  arg1.length > 0 &&
  (typeof utils.styleParser.parse(arg1)) === 'string'
) {
```

or maybe:

```
if (typeof arg2 !== 'undefined' &&
  typeof arg1 === 'string' &&
  arg1.length > 0 &&
  (typeof utils.styleParser.parse(arg1)) === 'string') {
```

Or either only with all clauses of the `if` left aligned?

dmarcos: I meant alligning the first clause with the parenthesis like the second form you described:
````javascript
if (typeof arg2 !== 'undefined' &&
    typeof arg1 === 'string' &&
    arg1.length > 0 &&
    (typeof utils.styleParser.parse(arg1)) === 'string') { 
    ...
 }
````

src/core/a-entity.js
dmarcos: Cool, thanks for putting some thought on it.

dmarcos:
Just a small nit and this is ready to go

https://github.com/aframevr/aframe/pull/2748/files#r121260759

gregfagan:
Thanks for the feedback @dmarcos. I went ahead and amended your suggestions into the commit and then rebased it onto master.

dmarcos:
Great! Thanks for sticking with it. Not many people get their hands dirty with these parts of the code. Thank you very much!

PR #2751, opened 2017-06-09
handle vrdisplaypresentchange when triggering enter/exit VR outside aframe (fixes #2614)

### LINKED ISSUE #2614: Enter VR button disappears after exit VR, in Samsung & Oculus browsers

ngokevin:
**Description:**

>  A user agent MUST dispatch this event type to indicate that a VRDisplay has begun or ended VR presentation. This event should not fire on subsequent calls to requestPresent() after the VRDisplay has already begun VR presentation.

Previously, we only handled exitVR on pressing the escape key. Thanks to @digitec for pointing in the right direction.

**Changes proposed:**
- handle onvrdisplaypresentchange
- [x] tested
- [x] unit tests

### DIFF
--- a/src/components/scene/auto-enter-vr.js
+++ b/src/components/scene/auto-enter-vr.js
@@ -2,9 +2,9 @@ var registerComponent = require('../../core/component').registerComponent;
 var utils = require('../../utils');
 
 /**
- * Automatically enter VR, either upon vrdisplayactivate (e.g. putting on Rift headset)
+ * Automatically enter VR either upon `vrdisplayactivate` (e.g. putting on Rift headset)
  * or immediately (if possible) if display name contains data string.
- * The default data string is 'GearVR' for Carmel browser which only does VR.
+ * Default data string is `GearVR` for Carmel browser which only does VR.
  */
 module.exports.Component = registerComponent('auto-enter-vr', {
   schema: {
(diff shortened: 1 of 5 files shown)


ngokevin:
@mkeblx Does Oculus Browser implement `onvrdisplaydeactivate`? That'd be the ideal way for A-Frame to know to "exit" VR. Otherwise, I don't think we should be hacking in listening for GearVR back button.

src/core/scene/a-scene.js
```diff
@@ -74,6 +74,10 @@ module.exports.AScene = registerElement('a-scene', {
         });
         this.addFullScreenStyles();
         initPostMessageAPI(this);
+
+        // Handler to exit VR (e.g., Oculus Browser back button).
+        this.onVRPresentChange = bind(this.onVRPresentChange, this);
+        window.addEventListener('onvrdisplaypresentchange', this.onVRPresentChange);
```
cvan: need to drop the `on`; the event is called `vrdisplaypresentchange`

src/core/scene/a-scene.js
```diff
@@ -265,6 +273,17 @@ module.exports.AScene = registerElement('a-scene', {
     },
 
     /**
+     * Handle exiting VR through other means than `<ESC>` key. For example, GearVR back button
+     * on Oculus Browser.
+     */
+    onVRPresentChange: {
+      value: function () {
+        if (!this.is('vr-mode')) { return; }
```
cvan: can we have `vr-mode` set if `vrdisplaypresentchange` fires upon entering VR? if an A-Frame site wants to enter VR mode using the WebVR API directly, currently devs have to explicitly call `$('a-scene').enterVR()`; simply calling `navigator.getVRDisplays().then(displays => displays[0].requestPresent([{source: $('canvas[data-aframe-canvas]')}])` won't work.

this is actually necessary for properly entering VR mode upon page navigation (without the `auto-enter-vr` component, which _forces_ entering VR mode, not responding to the user's already being in VR mode).

there are cases where sites would want to handle entering VR mode by using the WebVR API directly, without needing to special case for A-Frame scenes vs. three.js scenes vs. others. examples: the [WebVR Agent](https://github.com/webvrrocks/webvr-agent), a JS user script/bookmarklet, a WebExtension/Chrome browser extension, etc.

this will require a couple changes to the `enterVR`. specifically, instead of calling [`this.effect.requestPresent(…)`](https://github.com/ngokevin/aframe/blob/c01a63ba/src/core/scene/a-scene.js#L197), we'll want to just check if `if (checkHeadsetConnected || this.isMobile)` and call `enterVRSuccess()`.

src/core/scene/a-scene.js
```diff
@@ -265,6 +273,17 @@ module.exports.AScene = registerElement('a-scene', {
     },
 
     /**
+     * Handle exiting VR through other means than `<ESC>` key. For example, GearVR back button
+     * on Oculus Browser.
+     */
+    onVRPresentChange: {
+      value: function () {
```
cvan: this should check `if (evt.display && evt.display.getLayers && evt.display.getLayers().indexOf(self.canvas) === -1) { return; }` so that VR mode isn't toggled for the wrong presentation event (examples: multiple A-Frame scenes on a page, or if there are two VR displays connected but the active A-Frame scene isn't the content being presented)

ngokevin:
Thanks, updated. Didn't do the canvas checking here, maybe another PR, but enter/exit works on Oculus Browser now.

ngokevin:
One sec...fixing last second introduced issue

src/components/scene/auto-enter-vr.js
```diff
@@ -2,9 +2,9 @@ var registerComponent = require('../../core/component').registerComponent;
 var utils = require('../../utils');
 
 /**
- * Automatically enter VR, either upon vrdisplayactivate (e.g. putting on Rift headset)
+ * Automatically enter VR either upon `vrdisplayactivate` (e.g. putting on Rift headset)
  * or immediately (if possible) if display name contains data string.
- * The default data string is 'GearVR' for Carmel browser which only does VR.
+ * Default data string is `GearVR` for Carmel browser which only does VR.
  */
 module.exports.Component = registerComponent('auto-enter-vr', {
```
ngokevin: I don't have an entirely good grasp on the component, seems the useful thing we can keep and move to `<a-scene>` module is the display activate handlers. @machenmusik?

src/core/scene/a-scene.js
```diff
@@ -176,24 +183,27 @@ module.exports.AScene = registerElement('a-scene', {
...
+        if (!fromExternal && this.checkHeadsetConnected() || this.isMobile) {
           return this.effect.requestPresent().then(enterVRSuccess, enterVRFailure);
         }
+
         enterVRSuccess();
         return Promise.resolve();
 
         function enterVRSuccess () {
           self.addState('vr-mode');
-          self.emit('enter-vr', event);
```
ngokevin: I think before this was `undefined`, but I can rig it to pass something useful.

ngokevin: The event before was just the Enter VR button click event. I won't get to passing in a display in this PR.

src/components/scene/auto-enter-vr.js
machenmusik: See #2707 - now that Oculus Browser is widely available, I think that auto-enter-vr should not be added to scene by default anymore - if developing for Carmel, easy enough to add back in

ngokevin:
Fixing tests

src/core/scene/a-scene.js
```diff
@@ -109,16 +109,26 @@ module.exports.AScene = registerElement('a-scene', {
 
         // Add to scene index.
         scenes.push(this);
+
+        // Handler to exit VR (e.g., Oculus Browser back button).
+        this.onVRPresentChangeBound = bind(this.onVRPresentChange, this);
```
dmarcos: We don't need to rename and append `bound` to the method name.

ngokevin: Some leftovers from when I was debugging :P

PR #2752, opened 2017-06-10
eliminate need to mock document in node

kellyselden:
similar to #2747


ngokevin:
We use `document.createElement` everywhere. It doesn't matter for Node because we don't expect it to be executing any runtime code, right?

kellyselden:
Correct. I believe the main use case for aframe in node is to build the objects, not (yet) DOM manipulation.

PR #2755, opened 2017-06-10
eliminate need to mock navigator in node

kellyselden:
similar to #2747 and #2752.

This is the last piece. After this, using aframe in node will only require a `global.window` mock.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,7 +24,7 @@ if (window.location.protocol === 'file:') {
 window.Promise = window.Promise || require('promise-polyfill');
 
 // Check before the polyfill runs.
-window.hasNativeWebVRImplementation = !!navigator.getVRDisplays || !!navigator.getVRDevices;
+window.hasNativeWebVRImplementation = !!window.navigator.getVRDisplays || !!window.navigator.getVRDevices;
 
 window.WebVRConfig = window.WebVRConfig || {
   BUFFER_SCALE: 1,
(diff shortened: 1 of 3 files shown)


vendor/VRControls.js
```diff
@@ -35,9 +35,9 @@ THREE.VRControls = function ( object, onError ) {
 
 	}
 
-	if ( navigator.getVRDisplays ) {
+	if ( window.navigator.getVRDisplays ) {
```
ngokevin: Is it needed in VRControls? This should only be called at render time not start time

kellyselden: Without that change, I get this stack trace:

```
  1) node acceptance tests can run in node:
     ReferenceError: navigator is not defined
      at new THREE.VRControls (vendor\VRControls.js:38:7)
      at Object.<anonymous> (src\utils\device.js:3:16)
      at require (internal/module.js:11:18)
      at Object.<anonymous> (src\utils\index.js:5:14)
      at require (internal/module.js:11:18)
      at Object.<anonymous> (src\index.js:1:75)
      at require (internal/module.js:11:18)
      at Context.<anonymous> (tests\node\test.js:18:18)
```

So it may currently be incorrectly run at require time?

ngokevin: Because VRControls is a third-party three.js module, I'd go to utils.device and change it so that those modules aren't instantiated until they're needed.

```js
var controls;

someUtilFunction () {
  if (!controls) { instantiateControls(); }  // Lazy-instantiate VRControls to not break in Node.
}
```

kellyselden: Added in https://github.com/aframevr/aframe/pull/2756

ngokevin: Can update this now

kellyselden: updated

ngokevin:
Thanks!

ngokevin:
Maybe would be good to add `Requiring A-Frame in Node` section at like https://aframe.io/docs/0.5.0/core/globals.html

PR #2756, opened 2017-06-11
delay VRControls initialization for node

kellyselden:
In response to https://github.com/aframevr/aframe/pull/2755#discussion_r121292464.

This makes running in node more happy and eliminates the need for this change https://github.com/aframevr/aframe/pull/2755/files#diff-97c6b758b49a87147586d2d87795312aR38.

cc @ngokevin

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -1,12 +1,23 @@
 var THREE = require('../lib/three');
 var dolly = new THREE.Object3D();
-var controls = new THREE.VRControls(dolly);
+
+var getControls = (function () {
+  var controls;
+  return function () {
+    if (!controls) {
+      // Lazy-instantiate VRControls to not break in Node.
+      controls = new THREE.VRControls(dolly);
+    }
+    return controls;
+  };
+})();
 
 /**
  * Determine if a headset is connected by checking if the orientation is available.
  */
 function checkHeadsetConnected () {
   var orientation;
+  var controls = getControls();
   var vrDisplay = controls.getVRDisplay();
 
   // If `isConnected` is available, just use that.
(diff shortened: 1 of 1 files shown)


src/utils/device.js
```diff
@@ -1,12 +1,21 @@
 var THREE = require('../lib/three');
 var dolly = new THREE.Object3D();
-var controls = new THREE.VRControls(dolly);
+
+var controls;
```
ngokevin: Maybe to hide the controls variable:

```js
var getControls = (function () {
  var controls;
  return function getControls () {
    if (!controls) { 
      controls = new THREE.VRControls(dolly);
    }
    return controls;
  };
})();
```

kellyselden: good call

kellyselden: updated

dmarcos:
This PR has regressed VR mode on desktop. Neither Oculus nor Vive enter VR anymore. Delaying the initialization of VRControls introduce timing issues. Note that VRControls is also asynchronous and we have to somehow handle that asynchrony. The timing was right and the extra delay that this PR introduces affects the initialization timing. This PR is not critical for 0.6.0. I'm going to revert and we revisit later in a follow up PR. @kellyselden feel free to reopen. Make sure that Vive / Oculus enter VR mode.

kellyselden:
Sounds good. I'll take a look.

ngokevin:
Perhaps try checking if it's a node environment and not initializing it if so?

kellyselden:
Also, if you can think of a test that would have caught this regression, I would be happy to include/write it.

PR #2757, opened 2017-06-11
update sinon to 2.x

kellyselden:
I was getting a peerdep error without this update: https://travis-ci.org/kellyselden/aframe/jobs/241841600#L384-L386

I figure we aren't getting it yet here because we have a primed npm cache on travis.


ngokevin:
Is there a way we can downgrade Sinon and its peers back? Sinon 2.x has a lot of deprecations that we don't have much time to handle or learn the new API.

kellyselden:
Yep, I should be able to submit that quick.

PR #2767, opened 2017-06-13
have primitive handle mappings with non-default components

ngokevin:
**Description:**

Ran into this. I no longer needed this, but a good fix anyways.


dmarcos:
Test failing in travis

ngokevin:
Updated, fixed FF test

PR #2772, opened 2017-06-14
fix vive-controls button colors

ngokevin:
**Description:**

When Vive button was pressed, it would do:

- handle buttondown -> change to highlight color 
- handle buttonup -> change to default coor
- handle touchend -> change to highlight color

Vive buttons don't have capacitive touch, remove touch-related handlers.

**Changes proposed:**
- Remove touch related handlers from Vive controls. PR has lots of extra changes, but removing those few lines of code is all that's needed to fix the original issue.
- Clean code, tests, add tests
- [x] Tested on Vive/Windows/FF Nightly

### DIFF
--- a/src/components/vive-controls.js
+++ b/src/components/vive-controls.js
@@ -1,7 +1,9 @@
 var registerComponent = require('../core/component').registerComponent;
-var bind = require('../utils/bind');
-var checkControllerPresentAndSetup = require('../utils/tracked-controls').checkControllerPresentAndSetup;
-var emitIfAxesChanged = require('../utils/tracked-controls').emitIfAxesChanged;
+var utils = require('../utils/');
+
+var bind = utils.bind;
+var checkControllerPresentAndSetup = utils.trackedControls.checkControllerPresentAndSetup;
+var emitIfAxesChanged = utils.trackedControls.emitIfAxesChanged;
 
 var VIVE_CONTROLLER_MODEL_OBJ_URL = 'https://cdn.aframe.io/controllers/vive/vr_controller_vive.obj';
 var VIVE_CONTROLLER_MODEL_OBJ_MTL = 'https://cdn.aframe.io/controllers/vive/vr_controller_vive.mtl';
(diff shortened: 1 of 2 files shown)


src/components/vive-controls.js
```diff
@@ -1,75 +1,93 @@
...
     var self = this;
     this.animationActive = 'pointing';
+    this.checkControllerPresentAndSetup = checkControllerPresentAndSetup;  // To allow mock.
+    this.controllerPresent = false;
+    this.emitIfAxesChanged = emitIfAxesChanged;  // To allow mock.
+    this.lastControllerCheck = 0;
     this.onButtonChanged = bind(this.onButtonChanged, this);
     this.onButtonDown = function (evt) { self.onButtonEvent(evt.detail.id, 'down'); };
     this.onButtonUp = function (evt) { self.onButtonEvent(evt.detail.id, 'up'); };
-    this.onButtonTouchStart = function (evt) { self.onButtonEvent(evt.detail.id, 'touchstart'); };
```
dmarcos: the touchpad is capacitive. Are those events not captured anymore with this PR?

ngokevin: Hmm, I can adjust to make this continue to emit the `trackpadtouchend` events...but I just realized maybe the deeper issue is that doing triggerdown/triggerup causes a touchend event from tracked-controls.

dmarcos: I think that's not a bug. I think those touchstart / touchend are legit. I think @machenmusik observed that for the first time and probably related to this https://bugzilla.mozilla.org/show_bug.cgi?id=1353523. A slight press on the trigger is triggered as a touchstart event.

ngokevin: For me, it takes more than a light touch. But I updated the PR to ignore the touch events for buttons that don't have touch. `trackpadtouchstart` and vice-versa now continue to emit.

machenmusik: Coming to this a little late, but IIRC for Vive, the touch start / end events were intentional for analog buttons (trigger) as well as real touch sensitive buttons (trackpad) but are not useful for digital buttons (grip, menu).

PR #2784, opened 2017-06-16
remove auto-enter-vr (fixes #2781)

### LINKED ISSUE #2781: Enter VR button not appearing anymore (Oculus browser), impossible to enter VR
**Description:**

- A-Frame Version: latest master
- Platform / Device: Gear VR, Oculus browser
- Reproducible Code Snippet or URL: http://aframe-master.glitch.me/

This regression happened after fixing #2614 
The enter VR button briefly appears, then disappears.

ngokevin:
**Description:**

The auto enter VR code was somehow causing the enter vr button to not show up. Not sure what auto-enter-vr does now, doesn't seem needed anymore. I moved the vrdisplayactivate code to the scene.

I also found out Oculus Browser has a prompt I didn't see yesterday. Upon entering VR, I see a bottom bar that gives options such as 360, 180, top/bottom, 3D. I have to dismiss it, and then try again, and then it kicks into VR. Not sure what's happening there.

**Changes proposed:**
- Remove auto-enter-vr
- Add back a test I accidentally deleted before
- Move vrdisplayactivate code to scene module


src/core/scene/a-scene.js
```diff
@@ -113,6 +115,12 @@ module.exports.AScene = registerElement('a-scene', {
         // Handler to exit VR (e.g., Oculus Browser back button).
         this.onVRPresentChangeBound = bind(this.onVRPresentChange, this);
         window.addEventListener('vrdisplaypresentchange', this.onVRPresentChangeBound);
+
+        // Enter VR on `vrdisplayactivate` (e.g. putting on Rift headset).
```
dmarcos: This logic will go into the `link` will be on the link system but we can keep it for now

machenmusik:
I suspect that when auto-enter-vr tried but failed since not initiated by user gesture, something did not clean up correctly from the failure and so the enter-VR button remained hidden... By removing auto-enter-vr, the issue is masked, but may be exposed again if other errors cause entering VR to fail.

machenmusik:
So we definitely don't need #2707 now since this did that and more.

PR #2787, opened 2017-06-16
removed position

dirkk0:
The position attribute implies that this component alone can do positional audio, which is not true (as of 0.5)
To have positional audio, you need to attach this to another entity.

**Description:**

**Changes proposed:**
-
-
-


ngokevin:
This should work, we should find out why it isn't.

https://github.com/mrdoob/three.js/blob/master/src/audio/PositionalAudio.js#L82

ngokevin:
Thanks, we can take a look for a real solution. I'll test it out now.

PR #2790, opened 2017-06-16
[docs]

Mimaaa:
**Description:**
Fix typo on the [building a basic scene page](https://aframe.io/docs/0.5.0/guides/building-a-basic-scene.html#tweaking-lighting) under the tweaking light title.

**Changes proposed:**
- from `<a-light>s` to `<a-lights>`

### DIFF
--- a/docs/guides/building-a-basic-scene.md
+++ b/docs/guides/building-a-basic-scene.md
@@ -351,7 +351,7 @@ repeat in the Y direction (commonly referred to as U and V for textures).
 
 [light]: ../primitives/a-light.md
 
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks, but it was correct before. Multiple `<a-light>`s. `<a-lights>` is not valid.

Mimaaa:
Alright, my bad. Still sounds kinda weird. Shouldn't it be `<a-light>` without the `s`?

ngokevin:
I wanted to make it plural without having to say `with <a-light> elements`

Mimaaa:
Ah, that's why. 🆒

PR #2791, opened 2017-06-16
rolling back sinon and pinning its peers

kellyselden:
https://github.com/aframevr/aframe/pull/2757#issuecomment-309135066

The peers broke semver by requiring sinon 2, but sinon 2 has too much deprecation noise.

cc @ngokevin


ngokevin:
Did this already at https://github.com/aframevr/aframe/commit/0f3ff992ae71819e0048f4f04fc36083688a1a10 but hmm builds are erroring.

ngokevin:
Maybe we have to lock it down without the `^`?

kellyselden:
Yea a revert won't cut it, that will just put it back into a broken state. This PR should be what you want.

ngokevin:
OK, I see, can you rebase?

kellyselden:
rebased

ngokevin:
Thanks!

andreasplesch:
Should this go into the angle template as well ? I just had what seemed to be same/similar issue using angle initcomponent, the npm install.

PR #2795, opened 2017-06-18
[doc] geometry.md, register custom geometry

andreasplesch:
**Description:**

Make custom geometry example code functional (it wasn't):
THREE.Geometry.vertices needs to be a flat array.
BoundingBox and Normals need to calculated.
The example vertices were colinear (!), replace with triangle

https://glitch.com/edit/#!/abounding-minute?path=index.html:18:8
for code and live preview

**Changes proposed:**

- see commit
-
-


ngokevin:
thanks!

PR #2805, opened 2017-06-21
Docs v0.5.0

vrimaginar:
[docs] update cursor component documentation

- Change 'cursor-fusing' argument in 'Adding visual feedback' to 'fusing'.


ngokevin:
I think instead we should just revert it back to `cursor-fusing` in the code. Not worth breaking the few examples with animated cursors. Also the PR should be opened against master branch.

PR #2810, opened 2017-06-22
Modified extend to take into account primitive and absent values

### LINKED ISSUE #2809: `mixin` attribute does not merge non-object single-property components properly
The line in charge of [extending components coming from mixins](https://github.com/aframevr/aframe/blob/master/src/extras/primitives/primitives.js#L81) will return an invalid value when the mixin component is a non-object single-property value such as an string:

```js
data[name] = utils.extendDeep(data[name] || {}, mixinComponents[name]);
```

This is caused by how [`deep-assign`](https://www.npmjs.com/package/deep-assign) works.

Try the following snippet:

```js
const deepAssign = require('deep-assign');
undefined
deepAssign({}, '[camera]');
{ '0': '[',
  '1': 'c',
  '2': 'a',
  '3': 'm',
  '4': 'e',
  '5': 'r',
  '6': 'a',
  '7': ']' }
deepAssign({}, 1);
{}
```

delapuente:
Fix #2809

With the addition of the `extend()` function, the extension mechanism can consider the cases in which the component is absent or is of a primitive type:

1. If both `base` and `extension` are absent, `extend()` will return `undefined`.
2. Else, If `base` or `extension` is absent, `extend()` will returns a copy of the other.
3. Else, If none of them is absent and both are pure JS objects, the second will extend the first as usual.
4. Finally, the `extension` is preferred so a copy of it is returned.

### DIFF
--- a/src/extras/primitives/primitives.js
+++ b/src/extras/primitives/primitives.js
@@ -101,11 +101,11 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
 
           return data;
 
+          /**
+           * For the base to be extensible, both objects must be pure JavaScript objects.
+           * The function assumes that base is undefined, or null or a pure object.
+           */
           function extend (base, extension) {
-            // XXX: For the base to be extensible, both objects must be pure
-            // JavaScript objects. The function assumes that base is
-            // undefined, or null or a pure JavaScript object.
-
             if (isUndefined(base)) {
               return copy(extension);
             }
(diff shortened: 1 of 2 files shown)


ngokevin:
Can you post example code which this PR might address? Will help clarify the case, thanks!

dmarcos:
I would add a test with the case that this PR fixes

ngokevin:
added the tests

Issue #2808, opened 2017-06-22
Changing a-text value from javascript


postlund:
**Description:**
I'm trying to change the text in an a-text dynamically with setAttribute from Javascript.  The example linked below simply just changes the text to another hardcoded text (compared to initial value) each tick. It works as I expect it too in Chrome ("my text" is shown), but not in Firefox or Safari ("text" is shown). I cannot really see what's wrong here, especially since it works in Chrome. Any ideas?

- A-Frame Version: 0.5.0
- Platform / Device: Firefox/Safari, macOS
- Reproducible Code Snippet or URL: https://pastebin.com/aWytFC2k

ngokevin:
Fixed

PR #2814, opened 2017-06-23
fix case if text data is updated while font is loading (#2808)

ngokevin:
**Description:**

What was happening:

1. Text component initializes.
2. Text component sets to load font.
3. Tick function tries to change text value, but font not loaded so nothing happens.
4. Font loads.
5. Tick function tries to change text value, but since the data did not change from step 3, nothing ever happens no matter how many ticks.

**Changes proposed:**
- Reference `self` for the component rather than referencing old data in case it changes.

### DIFF
--- a/src/components/text.js
+++ b/src/components/text.js
@@ -106,7 +106,7 @@ module.exports.Component = registerComponent('text', {
 
     // Update geometry and layout.
     if (font) {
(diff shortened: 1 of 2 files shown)


dmarcos:
Any test we can add?

ngokevin:
Added

PR #2816, opened 2017-06-23
fix OBJ and MTL URLs in gearvr-controls docs

machenmusik:
@brianchirls noticed that the URLs to the OBJ and MTL in `gearvr-controls` docs were wrong


ngokevin:
thanks!

PR #2818, opened 2017-06-23
Switch back to using webvr-polyfill@0.9.35

jsantell:
**Description:**

Switches back to webvr-polyfill master @ 0.9.35 now that googlevr/webvr-polyfill#267 has landed


dmarcos:
Thanks!

PR #2820, opened 2017-06-25
when selecting element by id in selectorParse, use getElementById instead of querySelector for better performance

eminaz:
when selecting element by id in selectorParse, use getElementById instead of querySelector for better performance (https://jsperf.com/getelementbyid-vs-queryselector), and this way it also support id that starts with a digit (querySelector doesn’t support that)

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -122,6 +122,9 @@ function numberParse (value) {
 function selectorParse (value) {
   if (!value) { return null; }
   if (typeof value !== 'string') { return value; }
+  if(value[0] === '#') {
+    return document.getElementById(value.substring(1));
+  }
   return document.querySelector(value);
 }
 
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks! Almost ready to go. There's a broken test to fix

eminaz:
Fixed.

src/core/propertyTypes.js
```diff
@@ -122,6 +122,10 @@ function numberParse (value) {
 function selectorParse (value) {
   if (!value) { return null; }
   if (typeof value !== 'string') { return value; }
+  if (value[0] === '#' && !/[.,[: ]/.test(value)) {
```
ngokevin: Do we need the second check?

eminaz: that's to check if the query is '#someId.someClass div[name="b"]' etc.

eminaz: I've updated the regex

ngokevin:
I've updated to fix tests, added tests for selectors that were not accounted for, fixed the regex.

ngokevin:
thanks much

PR #2831, opened 2017-06-30
Fixing 2 issues with GearVR.

Artyom17:
**Description:**

1) EnterVR button didn't work at all
2) EnterVR button worked only once, since exitVR from external cause wasn't handled.

**Changes proposed:**
- Remove check if isGearVR from utils.isMobile (already done in 0.6.0)
- Add vrdisplaypresentchange callback to a-scene and exit VR correctly on external cause (already done in 0.6.0)

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -105,6 +105,11 @@ module.exports = registerElement('a-scene', {
 
         window.addEventListener('load', resize);
         window.addEventListener('resize', resize);
+
+        // Handler to exit VR (e.g., Oculus Browser back button).
+        this.onVRPresentChangeBound = bind(this.onVRPresentChange, this);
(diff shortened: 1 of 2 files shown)


dmarcos:
This PR is made against the docs branch. It should be against master

dmarcos:
We will be releasing 0.6.1 later this week. If this PR is submitted against master the fix will shipped with it. We then have to recommend people to upgrade. Fixing on 0.4.0 and 0.5.0 won't help much since we would have to cut a 0.4.1 and 0.5.1 releases that devs would have to upgrade as well. We just do bug fix releases of the latest public version. 0.6.0 at the moment

ngokevin:
Yeah, should be against the doc branch, but we don't cherry-pick fixes to older versions. Newer versions exist to improve older ones.

PR #2832, opened 2017-06-30
Fixing 2 issues with GearVR

Artyom17:
**Description:**

1) EnterVR button didn't work at all
2) EnterVR button worked only once, since exitVR from external cause wasn't handled.

**Changes proposed:**
- Remove check if isGearVR from utils.isMobile (already done in 0.6.0)
- Add vrdisplaypresentchange callback to a-scene and exit VR correctly on external cause (already done in 0.6.0)


ngokevin:
0.6.0 fixes this, and the docs branch is just for cherry-picking documentation fixes.

PR #2836, opened 2017-07-03
fix deprecation direction

kellyselden:
The github tag release notes and any other places should be updated too.


dmarcos:
Thanks!

PR #2839, opened 2017-07-03
[docs] Add some details on model troubleshooting.

donmccurdy:
Moves "check the console" advice to the top of the section, and gives examples of absolute/relative URLs.


dmarcos:
🙏 Thanks!

PR #2840, opened 2017-07-04
Store oldData after building component data. undefined passed to upda…

### LINKED ISSUE #2829: Entity.setAttribute('position') Not Always Effective
While developing a new site, I noticed my interface was mysteriously broken this morning (6/30/2017). I had not pushed any code changes, but it's pulling in aframe.io/aframe/dist/aframe-master.js which _did change_ last night.

The code that no longer worked is supposed to manage UI elements by automatically readjusting the position of buttons (simple plane geometries) whenever a button is added/removed. It does this by looping all elements (each button is `control`):
```
pos = control.getAttribute('position');
if (pos) {
  pos.x = offset_left;
  control.setAttribute('position', pos);
}
```

I checked `control.getAttribute('position')` and they were all correct, but none of the positions were actually changed in the scene. I tried manually performing `control.setAttribute('position', pos)` via the developer console and noticed something odd: it took **two** calls to `setAttribute` to actually get the element to move. So my updated UI loop (which works again, hooray!):
```
pos = control.getAttribute('position');
if (pos) {
  pos.x = offset_left;
  // For some reason, this now needs to be called twice.
  control.setAttribute('position', pos);
  control.setAttribute('position', pos);
}
```

- A-Frame Version: 0.6.0 (Date 30-06-2017, Commit #13ced44)
- Platform / Device: Firefox and Chrome on Windows 10 (latest stable versions of each)
To reproduce, go to https://aframe.io/aframe/examples/boilerplate/hello-world/ and execute this script via the console:
```
var pos = document.querySelectorAll('a-box')[0].getAttribute('position');
pos.x -= 1;
document.querySelectorAll('a-box')[0].setAttribute('position', pos);
// The box does **not** update position, until you run the following line (uncomment before running):
// document.querySelectorAll('a-box')[0].setAttribute('position', pos); // Now it works!
```

dmarcos:
…te on component initialization (fix #2829)

The following call to `getAttribute` returns the data object of the component (this happens due to one of the perf optimization to reduce unnecessary object copies) that gets modified and then passed to the `setAttribute` method. 
````javascript
var position = el.getAttribute('position');
position += 1;
el.setAttribute('position', position);
````
the oldData object was constructed before building the data object resulting in the oldData to actually contain the current data. The condition that checks if the component should be updated (oldData !== data) was failing because the objects were identical.

Two fundamental changes in this PR:
1. The oldData object is built after the data object is constructed to contain the actual previous data.
2. The update method receives undefined on first call when the component initializes. The core components have been modified to comply.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -37,6 +37,8 @@ var Component = module.exports.Component = function (el, attrValue, id) {
   this.attrName = this.name + (id ? '__' + id : '');
   this.initialized = false;
   this.el.components[this.attrName] = this;
+  // Store component data from previous update call.
+  this.oldData = undefined;
   // Last value passed to updateProperties.
   this.previousAttrValue = undefined;
   this.throttledEmitComponentChanged = utils.throttle(function emitComponentChange (oldData) {
@@ -228,8 +230,8 @@ Component.prototype = {
   updateProperties: function (attrValue, clobber) {
     var el = this.el;
     var isSinglePropSchema;
-    var oldData;
     var skipTypeChecking;
+    var oldData = this.oldData;
 
     // Just cache the attribute if the entity has not loaded
     // Components are not initialized until the entity has loaded
@@ -239,10 +241,8 @@ Component.prototype = {
     }
 
     isSinglePropSchema = isSingleProp(this.schema);
-    // Copy old data since this.data is going to be recreated.
-    oldData = extendProperties({}, this.data, isSinglePropSchema);
     // Disable type checking if the passed attribute is an object and has not changed.
-    skipTypeChecking = typeof this.previousAttrValue === 'object' &&
+    skipTypeChecking = attrValue !== null && typeof this.previousAttrValue === 'object' &&
                        attrValue === this.previousAttrValue;
     // Cache previously passed attribute to decide if we skip type checking.
     this.previousAttrValue = attrValue;
@@ -263,9 +263,14 @@ Component.prototype = {
       this.init();
       this.initialized = true;
       delete el.initializingComponents[this.name];
-      // Play the component if the entity is playing.
+      // We pass empty object to multiple property schemas and single property schemas that parse to objects like position, rotation, scale
+      // undefined is passed to the rest of types.
+      oldData = (!isSinglePropSchema || typeof parseProperty(undefined, this.schema) === 'object') ? {} : undefined;
       this.update(oldData);
+      // Play the component if the entity is playing.
       if (el.isPlaying) { this.play(); }
+      // Store current data as previous data for future updates.
+      this.oldData = extendProperties({}, this.data, isSinglePropSchema);
       el.emit('componentinitialized', {
         id: this.id,
         name: this.name,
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by ngokevin** [comment]
Can we avoid making oldData undefined on the first call? It's extra work to check and opens up for potential for lots of bugs in components (e.g., not accounting it could be both undefined or an object). Can we separate the internal old data versus the public one?

**Comment by dmarcos** [comment]
Old Data being undefined it what makes sense to me? What do you return on first call, when the component has been just initialized?

**Comment by ngokevin** [comment]
Empty data makes just as much sense. And it's backwards incompatible and puts less type checking effort and less potential bugs for component authors.

What do we return where?

**Comment by dmarcos** [comment]
What about components that don't have an object as data type squema like visible for instance?

**Comment by ngokevin** [comment]
undefined works for better single-prop schemas. I think that was the current behavior.

**Comment by dmarcos** [comment]
Ok, I'll revert back to passing `{}` on multi prop schemas and `undefined` on single property

File: src/core/component.js [hunk 1]
```diff
@@ -228,8 +228,8 @@ Component.prototype = {
   updateProperties: function (attrValue, clobber) {
     var el = this.el;
     var isSinglePropSchema;
-    var oldData;
     var skipTypeChecking;
+    var oldData = this.oldData;
```
ngokevin: Can initialize this in the component constructor as undefined.

src/core/component.js
```diff
@@ -263,18 +261,23 @@ Component.prototype = {
       this.init();
       this.initialized = true;
       delete el.initializingComponents[this.name];
-      // Play the component if the entity is playing.
+      // We pass empty object to multiple property schemas and single property schemas that parse to objects like position, rotation, scale
+      // undefined is passed to the
```
ngokevin: This comment stopped suddenly

src/core/component.js
```diff
@@ -263,18 +261,23 @@ Component.prototype = {
...
+      oldData = (!isSinglePropSchema || typeof parseProperty(undefined, this.schema) === 'object') ? {} : undefined;
       this.update(oldData);
+      // Play the component if the entity is playing.
       if (el.isPlaying) { this.play(); }
+      this.oldData = extendProperties({}, this.data, isSinglePropSchema);
```
ngokevin: `// Store current data as previous data for future updates.`

src/core/component.js
```diff
@@ -263,18 +261,23 @@ Component.prototype = {
...
         id: this.id,
         name: this.name,
         data: this.getData()
       }, false);
     } else {
       // Don't update if properties haven't changed
-      if (!skipTypeChecking && utils.deepEqual(oldData, this.data)) { return; }
+      if (!skipTypeChecking && utils.deepEqual(this.oldData, this.data)) { return; }
 
+      this.oldData = extendProperties({}, this.data, isSinglePropSchema);
```
ngokevin: `// Store current data as previous data for future updates.`

src/core/component.js
```diff
@@ -346,7 +349,7 @@ Component.prototype = {
 
     // 1. Default values (lowest precendence).
     if (isSinglePropSchema) {
-      data = schema.default;
+      data = typeof schema.default === 'object' ? utils.extend({}, schema.default) : schema.default;
```
ngokevin: `Clone default value if object so components don't share object.`

ngokevin:
Can also update the PR description for future reference.

nwoltman:
Is there going to be a 0.6.1 patch release for this bug fix?

ngokevin:
yeah

PR #2850, opened 2017-07-07
Update a-plane.md

StormNetwork:
**Description:**

**Changes proposed:**
-
-
-


StormNetwork:
Change the "material" to "geometry".

ngokevin:
We also need to change the default values to `1`.

PR #2852, opened 2017-07-08
pivot should respond to entityEl.setAttribute

Leopoldthecoder:
**Description:**
Currently the `pivot` component doesn't respond to entityEl.setAttribute because it has no `update` handler.

**Changes proposed:**

- Replacing `init` with `update` to respond to changes
- Taking `oldData` into calculation to yield a desired `pivot` value

Plus, why is `pivot` not in the documentation? Am I missing anything or is it not ready for us to use? Thanks.

### DIFF
--- a/src/extras/components/pivot.js
+++ b/src/extras/components/pivot.js
@@ -12,8 +12,14 @@ registerComponent('pivot', {
 
   schema: {type: 'vec3'},
 
-  init: function () {
-    var data = this.data;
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks. `pivot` was previously something someone wanted to reduce the amount of entity nesting for animations. We created some examples using it (e.g., the A-Frame logo). We don't really want it in the core, but are keeping it for those examples still around,

PR #2854, opened 2017-07-08
Replace 'replace' dependency.

donmccurdy:
It's unmaintained, and has an insecure dependency on an older version of `minimatch`.


dmarcos:
Thanks!

PR #2855, opened 2017-07-09
Use attribute for animation color as "from" if "from" is not defined

### LINKED ISSUE #2842: a-animation for color requires from attribute
**Description:**

The <a-animation> element with color as animated attribute does not use the parent element's color as from value. It is necessary to provide an explicit from value.

- A-Frame Version: 0.5.0, 0.6.0
- Platform / Device: all

https://glitch.com/edit/#!/almond-taker?path=public/index.html:10:40
https://almond-taker.glitch.me/

The box color should interpolate from blue to red but resets to black as initial color at the start of the animation.

takahirox:
**Description:**

This PR lets animation color use attribute as `from` if `from` is not defined, fixes #2842

**Changes proposed:**
- Use attribute as `from` if `from` is not defined
- Add an animation color test, "from is not defined".

### DIFF
--- a/src/core/a-animation.js
+++ b/src/core/a-animation.js
@@ -462,7 +462,7 @@ function getAnimationValues (el, attribute, dataFrom, dataTo, currentValue) {
    *   Then converts to hex for the setAttribute
    */
   function getForColorComponent () {
-    from = new THREE.Color(dataFrom);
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

PR #2858, opened 2017-07-10
`reload` should restart scene components

### LINKED ISSUE #2239: scene reload() only shows a white page
**the reload() method for a scene only shows a blank white page:**

- A-Frame Version: 4.0
- Platform / Device: tried on safari & chrome on macOS and windows 
with any scene open, just run document.querySelector('a-scene').reload(); in the developer console.

rockwalrus:
**Description:**
`AScene.reload` doesn't currently reset attributes and components on the scene element, which is different behavior from every other element in the scene.

**Changes proposed:**
While we can't destroy and re-create the entire scene element like we do for every other element, we can:
-  remove the current components
- delete the current attributes
- re-add the original attributes

This patch fixes #2239.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -300,6 +300,8 @@ var proto = Object.create(ANode.prototype, {
       if (this.hasLoaded || !this.parentEl) { return; }
 
       ANode.prototype.load.call(this, function entityLoadCallback () {
+        /* if you change this callback, update a-scene.reload appropriately. */
+
         // Check if entity was detached while it was waiting to load.
         if (!self.parentEl) { return; }
 
@@ -423,7 +425,7 @@ var proto = Object.create(ANode.prototype, {
   },
 
   removeComponent: {
-    value: function (name) {
+    value: function (name, force) {
       var component;
       var isDefault;
       var isMixedIn;
@@ -431,7 +433,7 @@ var proto = Object.create(ANode.prototype, {
       // Don't remove default or mixed-in components.
       isDefault = name in this.defaultComponents;
       isMixedIn = isComponentMixedIn(name, this.mixinEls);
-      if (isDefault || isMixedIn) { return; }
+      if ((isDefault || isMixedIn) && !force) { return; }
 
       component = this.components[name];
       if (!component) { return; }
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "snazzy": "^5.0.0",
     "too-wordy": "ngokevin/too-wordy",
     "uglifyjs": "^2.4.10",
+    "watchify": "^3.9.0",
     "write-good": "^0.9.1"
   },
   "link": true,
--- a/src/components/scene/canvas.js
+++ b/src/components/scene/canvas.js
@@ -39,5 +39,8 @@ module.exports.Component = registerComponent('canvas', {
       document.activeElement.blur();
       document.body.focus();
     }
+  },
+  remove: function () {
+    this.el.canvas = undefined;
   }
 });
--- a/src/components/scene/vr-mode-ui.js
+++ b/src/components/scene/vr-mode-ui.js
@@ -66,7 +66,7 @@ module.exports.Component = registerComponent('vr-mode-ui', {
 
   remove: function () {
     [this.enterVREl, this.orientationModalEl].forEach(function (uiElement) {
-      if (uiElement) {
+      if (uiElement && uiElement.parentNode) {
         uiElement.parentNode.removeChild(uiElement);
       }
     });
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -1,4 +1,4 @@
-/* global Promise, screen */
+/* global Promise, screen, HTMLElement */
 var initMetaTags = require('./metaTags').inject;
 var initWakelock = require('./wakelock');
 var re = require('../a-register-element');
@@ -66,7 +66,20 @@ module.exports.AScene = registerElement('a-scene', {
         this.hasLoaded = false;
         this.isPlaying = false;
         this.originalHTML = this.innerHTML;
+        this.originalAttributes = [];
+
+        var attributes = this.attributes;
+
+        for (var i = 0; i < attributes.length; ++i) {
+          var attribute = attributes[i];
+          this.originalAttributes.push({
+            name: attribute.name,
+            value: attribute.value
+          });
+        }
+
         this.renderTarget = null;
+
         this.addEventListener('render-target-loaded', function () {
           this.setupRenderer();
           this.resize();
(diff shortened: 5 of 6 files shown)


src/core/a-entity.js
```diff
@@ -423,15 +425,15 @@ var proto = Object.create(ANode.prototype, {
   },
 
   removeComponent: {
-    value: function (name) {
+    value: function (name, force) {
```
dmarcos: Why do you need a `force` flag?

ngokevin:
Thinking further, why do we need a `.reload()` feature? I'm sort of in favor of removing this altogether because it will get messy trying to maintain the feature and address all edge cases of resetting a scene to a clean slate.

src/core/a-entity.js
rockwalrus: `removeComponent` on a default component is currently a no-op. I didn't want to re-implement all the `removeComponent` logic for `reload`, so I just added the flag. I agree that it's kind of counter-intuitive that you can't remove default components, so it might be a better approach to make `removeComponent` unconditionally remove components, and change the callers to not call `removeComponent` for default components if that is the desired behavior.

dmarcos:
I'm fine with deprecating. I checked with @fernandojsg and we don't use it in the inspector. @rockwalrus how did you hit the issue? what's your use case?

rockwalrus:
Our use case for `reload` is for demoing; we can use it to quickly reset the demo to the initial state without having to leave VR and hit the reload button or implement a more complicated solution. We don't have much state in our application, so it isn't hard to store all of it in A-Frame components.

dmarcos:
@rockwalrus Firefox Nightly now implements the `vrdisplayactivate` that it's fired on reload and the scene should reenter VR automatically so you would not need  to call `reload`. Anything that prevents you demoing on Firefox Nightly?

rockwalrus:
`vrdisplayactivate` would be sufficent for our use case. Unfortunately, our target devices are mobile phones running Chrome for Android with Cardboard viewers and it would be quite the challenge to convince our customers to install a Firefox nightly on every device they want to demo with.

One case for `reload` that `vrdisplayactivate` wouldn't handle is having multiple embedded scenes on one page where you want to be able to reset them, perhaps in a programming tutorial.

ngokevin:
I think it's fine to reload the page and enter VR programmatically (this can be done on mobile when using the polyfill). Everything done in this current PR could be done without being in the core as well.

dmarcos:
Multiple embedded scenes is not supported at the moment and it's not something I think we will do in the short / mid term. There are a lot of performance implications and we have to first focus on get good overall performance on a single scene :) 

You said that your scenes don't have much state store on them. It will be more performant to reset the component properties to the original values rather than recreating the whole scene (specially on mobile) as `reload does`

rockwalrus:
@dmarcos We also dynamically add and remove elements, which makes the way `reload` works a little more attractive.

dmarcos:
@rockwalrus you could add a class to all your dynamic entities `<a-scene class="myElement">` and select them all with `document.querySelectorAll('.myElement')` and remove them. Still more performant that reloading the scene, specially if you care about cardboard / mobile

dmarcos:
In your case I would create a `reset` component for the scene that restores the state and removes all the dynamic entities.

rockwalrus:
@ngokevin I do like the idea of moving `reload` to a separate component. If `removeComponent` gets changed to always remove the component and gets documented it should be possible to implement entirely with public API.

src/core/a-entity.js
dmarcos: Why do you need to remove `default` components? Cannot you store and restore the initial values? The flag breaks the component precedence model and It will make things confusing and introduce weird edge cases and bugs.

dmarcos: An example is that an entity without `position`, `rotation` or `scale` it does not make sense so those cannot be removed.

rockwalrus: `canvas` and `look-controls` are default components on `<a-scene>` that need resetting during `reload`. Removing them and re-adding them seemed the best way to simulate the initial load and the closest to what happens to components on other elements.

dmarcos: An `a-scene` without a `canvas` component does not make sense either. That's why is a default component and we don't allow removal. I would have a `restore` component that can save the original state of all the components you care about and restore the values via `setAttribute` instead of removing / adding components. The logic is not very complicated and it will yield better performance. The default components are there for a reason and a mechanism to workaround the mechanism introduces a lot of inconsistencies and potential for misuse.

rockwalrus:
@ngokevin Come to think of it, I'm not sure there are any hooks a separate component could use to grab the original HTML before it gets modified, so that might need to be added, too.

rockwalrus:
I see three paths forward for `reload`. Here are my thoughts about them:

# Remove `reload` and leave no upgrade path
There's no sense in leaving `reload` in the codebase if it's not going to be fixed. It currently isn't useful at all.

#2239 has four participants, so there is some interest in this functionality. While it's certainly possible to work around the lack of `reload()`, it takes more effort than a single method call.

Reload functionality really needs to be universal in order to work well.  Without core support or a clear de-facto standard to follow, it's likely that some components won't support any given approach.

# Remove `reload` and provide an out-of-core component to handle it
@ngokevin is right that it probably wouldn't take too much effort to move the current implementation into a component. That will require some support from core.  I see three possible ways to make that happen.
 
## Core can expose enough raw functionality to make reloading possible
@dmarcos gives good arguments against this approach. "Here there be dragons" methods don't seem to be the A-Frame way.


## Core can expose something like a `resetComponents` method
This would be safer, but at the price of keeping a fair amount of the logic in core.

## Core can provide a private interface to the component
It's hard to imagine a way to do this which wouldn't be brittle and ugly. 

# Fix `reload`
Moving `reload` out of core doesn't really decrease complexity, just shuffles it around. Keeping it in the core means that changes can be made without having to worry about how they affect the core/component boundary. More aggressive optimizations to `reload`, like keeping `<a-assets>` or the injected `<canvas>` around would be impossible to do cleanly as a component.

src/core/a-entity.js
rockwalrus: The problem I was solving was that after a call to `reload`, `canvas` and `look-controls` were still relying on the injected `<canvas>` element that was no longer in the tree. The most maintainable approach I saw was to follow the pattern that the rest of `reload` follows and simulate how the DOM was originally constructed as closely as possible. That meant resetting  `<a-scene>`'s `components` to empty and reinitializing everything.   I tried to make the process as close to the original initialization as possible so that the component level wouldn't need to know the difference between initializing and reinitializing.

So while in general removing `<canvas>` doesn't make sense, it actually did in this one weird case.

dmarcos:
To determine if `reload` is worth keeping I prefer to analyze the use cases that need the feature. In the case described above (restoring the initial state of a scene) reloading the whole scene is not the ideal more performant way to proceed and it should be solved differently.

rockwalrus:
Yeah, I do think you're misunderstanding our use case a bit. Our customer wants everything to be reset exactly the way it is on initial load, which is why nuking the DOM and restoring it to its initial state like `reload` does is so attractive. Otherwise we'd have to keep track of every individual attribute that could ever change, including entity `position` and `rotation`, and remember to add new attributes to that list if we ever start changing them.

It's entirely possible that trying to change just the things that need to be changed might be faster. Without benchmarks I wouldn't be able to say. There is an overhead to determining what things need to be added/changed/removed and individually making those changes, and `reload` runs pretty quickly. At any rate,`reload` already works much faster than a user can strap a Cardboard viewer to their face, so our customer is satisfied with the performance.

dmarcos:
Your use case to reload a page is already covered by the WebVR API since reloading a page and remaining in VR is a considered scenario. Your problem is with devices where there's no WebVR implementation (Cardboard like devices). Are you using a cardboard with a button? You could hook enter VR to a press on the button after reload to be able to enter VR (fullscreen on polyfilled devices)

rockwalrus:
That would work for us, but sadly, no buttons on the viewers. That does give me an idea, though. Tomorrow I'll look into seeing if we can get it working with a web app manifest with `display: fullscreen`.

rockwalrus:
A web app manifest doesn't help. `display: fullscreen` actually breaks the polyfill.

rockwalrus:
FYI: after a refactoring of our code, having `reload` isn't as vital to us.

ngokevin:
OK, I would close, look into deprecating `.reload()`, and leave as application-level / third-party code.

PR #2861, opened 2017-07-11
Add a mouse based cursor / raycaster and applies to the link traversal example

dmarcos:
* It adds a mouse cursor mode to the `cursor` component similar to https://github.com/mayognaise/aframe-mouse-cursor-component
* It modifies the `raycaster` component to accommodate the mouse cursor use case
* It makes the link traversal example usable on desktop
* It adds link traversal example to the index.html 

[Demo](http://swimminglessonsformodernlife.com/aframe/examples/showcase/link-traversal/mountains.html)

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -80,6 +80,7 @@ AFRAME.registerComponent('cursor-listener', {
 | downEvents  | Array of additional events on the entity to *listen* to for triggering `mousedown` (e.g., `triggerdown` for vive-controls).  | []                               |
 | fuse        | Whether cursor is fuse-based.                                                                                              | false on desktop, true on mobile |
 | fuseTimeout | How long to wait (in milliseconds) before triggering a fuse-based click event.                                             | 1500                             |
+| rayOrigin     | Where the intersection ray is cast from (i.e.,entity or mouse) | entity
 | upEvents    | Array of additional events on the entity to *listen* to for triggering `mouseup` (e.g., `trackpadup` for daydream-controls). | []                               |
 
 To further customize the cursor component, we configure the cursor's dependency
--- a/docs/components/raycaster.md
+++ b/docs/components/raycaster.md
@@ -58,6 +58,8 @@ AFRAME.registerComponent('collider-check', {
 | origin    | Vector3 coordinate of where the ray should originate from relative to the entity's origin.                                                            | 0, 0, 0       |
 | recursive | Checks all children of objects if set. Else only checks intersections with root objects.                                                              | true          |
 | showLine  | Whether or not to display the raycaster visually with the [line component][line].                                                                                         | false          |
+| useWorldCoordinates  | Whether the raycaster origin and direction properties are specified in world coordinates.
+.                                                                                         | false          |
 
 ## Events
 
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -1,3 +1,4 @@
+/* global THREE */
 var registerComponent = require('../core/component').registerComponent;
 var utils = require('../utils/');
 
@@ -37,7 +38,8 @@ module.exports.Component = registerComponent('cursor', {
     downEvents: {default: []},
     fuse: {default: utils.device.isMobile()},
     fuseTimeout: {default: 1500, min: 0},
-    upEvents: {default: []}
+    upEvents: {default: []},
+    rayOrigin: {default: 'entity', oneOf: ['mouse', 'entity']}
   },
 
   init: function () {
@@ -51,6 +53,12 @@ module.exports.Component = registerComponent('cursor', {
     this.onCursorUp = bind(this.onCursorUp, this);
     this.onIntersection = bind(this.onIntersection, this);
     this.onIntersectionCleared = bind(this.onIntersectionCleared, this);
+    this.onMouseMove = bind(this.onMouseMove, this);
+  },
+
+  update: function (oldData) {
+    if (this.data.rayOrigin === oldData.rayOrigin) { return; }
+    this.updateMouseEventListeners();
   },
 
   play: function () {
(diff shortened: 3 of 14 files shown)


docs/components/cursor.md
```diff
@@ -80,6 +80,7 @@ AFRAME.registerComponent('cursor-listener', {
 | downEvents  | Array of additional events on the entity to *listen* to for triggering `mousedown` (e.g., `triggerdown` for vive-controls).  | []                               |
 | fuse        | Whether cursor is fuse-based.                                                                                              | false on desktop, true on mobile |
 | fuseTimeout | How long to wait (in milliseconds) before triggering a fuse-based click event.                                             | 1500                             |
+| mode     | Camera (gaze) or mouse controlled. | gaze
```
ngokevin: I'd just make it a boolean based for mouse. `mouseMode: true|false`. The cursor can also be used for controllers as well, not just for gaze.

dmarcos: but you also might want mouse and not gaze. Also the raycaster operates differently for each mode: raycaster from the entity or from the mouse cursor on the screen. If you want gaze + mouse you can add two cursors to the same or different entities. In the link traversal examples the cursor is added to the `a-scene`

ngokevin: Don't know if my comment was understood. `mode: gaze` doesn't really do anything (and the non-mouse cursor is not just for gaze). What we want to represent is either the cursor has mouse capabilities or it doesn't, represented as a boolean `mouseMode: true` or `mouseMode: false`. This would be consistent to how it's represented in the code as well.

dmarcos: The problem is that gaze and mouse are mutually exclusive because the raycaster configuration is different (hence the mode property). A cursor as it is, cannot operate in both gaze and mouse modes. A `mouseMode` flag is confusing because it conveys an additional functionality on top of the normal behavior that can be toggled. We could make the `raycaster` a multiple component and the cursor component could add two of them: One for gaze and the other for mouse. This might introduce confusion since configuring the raycaster by the user would not be as simple as:
```javascript
el.setAttribute('raycaster', 'objects', '.clickable')
```
Two different raycasters would have to be configured instead:
```javascript
 el.setAttribute('raycaster'__gaze, 'objects', '.clickable');
 el.setAttribute('raycaster'__mouse, 'objects', '.clickable');
```
An alternative would be to add an `objects` property to the `cursor` component so the user does not configure the raycaster anymore

ngokevin: A boolean `mouseMode` property properly represents the mutual exclusiveness, that the entire component is flipped to a mouse mode or it's not. "Modes" usually mean an entire configuration, not "additional functionality".

"Gaze mode" does not make sense as the cursor can be used with controllers as well, and in the component code, "gaze mode" doesn't trigger any logic. It's only when setting to mouse mode does any different code run.

How I could see this component being used (although I'm not too amped on including a mouse mode in the first place):

```
<a-scene>
  <a-mixin id="raycaster" raycaster="objects: .clickable"></a-mixin>

  <a-entity mixin="raycaster" cursor="mouseMode: true"></a-entity>

  <a-camera>
    <a-entity mixin="raycaster" cursor></a-entity>
  </a-camera>

  <a-entity mixin="raycaster" laser-controls</a-entity>
</a-scene>
```

ngokevin: I think the `mode` sort of makes sense, but I don't really like having to provide strings, that `gaze` isn't the only other way to use the cursor, and that `gaze` is not represented in the code as doing anything.

dmarcos: Maybe we have different expectations on what a flag means. When seeing `mouseMode: true` enabled many people will interpret it as `I can use the mouse cursor as well to select things` in addition to the normal behavior. A `mode` word indicates... modality :smile: We can replace `gaze` with `screen` \ `entity` to indicate where the cursor performs the raycasting from. A cleaner solution is probably adding two raycasters and, an objects property and a  `mouseMode` flag (false by default) so the modes are not mutually exclusive. The API would be then:
`<a-entity cursor="mouseMode: true; objects: clickable">` 
The cursor would operate in both modes simultaneously and use two raycasters: `raycaster__mouse` and `raycaster__entity`

ngokevin: Thinking about it, yeah that makes sense as long as it's not named `gaze`. I need to think about operating in both simultaneously, it would make it less work for developers and possibly more performant (one cursor vs. two cursors), although it's more magic and complex.

dmarcos:
I replaced `mode` with `rayOrigin` with two options `entity` / `mouse`

ngokevin:
So not going to try the idea of having a cursor work with both simultaneously?

dmarcos:
It needs more thought and probably not a priority right now.

src/components/raycaster.js
```diff
@@ -27,7 +27,8 @@ module.exports.Component = registerComponent('raycaster', {
     objects: {default: ''},
     origin: {type: 'vec3'},
     recursive: {default: true},
-    showLine: {default: false}
+    showLine: {default: false},
+    worldCoordinates: {default: false}
```
ngokevin: I'd name this `isWorldSpace` (or `areWorldCoordinates`) to not sound like it takes coordinates.

src/components/cursor.js
```diff
@@ -118,8 +125,34 @@ module.exports.Component = registerComponent('cursor', {
...
+    window.addEventListener('mousemove', this.onMouseMove, false);
+    el.setAttribute('raycaster', 'worldCoordinates', true);
   },
 
+  onMouseMove: (function () {
+    var mouse = new THREE.Vector2();
+    var origin = new THREE.Vector3();
+    var direction = new THREE.Vector3();
+    return function (evt) {
+      var camera = this.el.sceneEl.camera;
```
ngokevin: Need a test for this block

src/components/cursor.js
```diff
@@ -118,8 +125,34 @@ module.exports.Component = registerComponent('cursor', {
...
+      camera.parent.updateMatrixWorld();
+      camera.updateMatrixWorld();
+      mouse.x = (evt.clientX / window.innerWidth) * 2 - 1;
+      mouse.y = -(evt.clientY / window.innerHeight) * 2 + 1;
+      origin.setFromMatrixPosition(camera.matrixWorld);
+      direction.set(mouse.x, mouse.y, 0.5).unproject(camera).sub(origin).normalize();
+      this.el.setAttribute('raycaster', {origin: origin, direction: direction});
```
ngokevin: Reuse an object for this

src/components/cursor.js
```diff
@@ -51,6 +53,11 @@ module.exports.Component = registerComponent('cursor', {
     this.onCursorUp = bind(this.onCursorUp, this);
     this.onIntersection = bind(this.onIntersection, this);
     this.onIntersectionCleared = bind(this.onIntersectionCleared, this);
+    this.onMouseMove = bind(this.onMouseMove, this);
+  },
+
+  update: function () {
+    this.updateMouseEventListeners();
```
ngokevin: I'd only call this if `rayOrigin` changes

docs/components/raycaster.md
```diff
@@ -58,6 +58,7 @@ AFRAME.registerComponent('collider-check', {
 | origin    | Vector3 coordinate of where the ray should originate from relative to the entity's origin.                                                            | 0, 0, 0       |
 | recursive | Checks all children of objects if set. Else only checks intersections with root objects.                                                              | true          |
 | showLine  | Whether or not to display the raycaster visually with the [line component][line].                                                                                         | false          |
+| worldCoordinates  | Determine if origin and direction are given in local or world coordinates.                                                                                         | false          |
```
ngokevin: `Whether the raycaster origin and direction properties are specified in world space.`

docs/components/cursor.md
```diff
@@ -80,6 +80,7 @@ AFRAME.registerComponent('cursor-listener', {
 | downEvents  | Array of additional events on the entity to *listen* to for triggering `mousedown` (e.g., `triggerdown` for vive-controls).  | []                               |
 | fuse        | Whether cursor is fuse-based.                                                                                              | false on desktop, true on mobile |
 | fuseTimeout | How long to wait (in milliseconds) before triggering a fuse-based click event.                                             | 1500                             |
+| rayOrigin     | if the intersection ray is cast from the entity or the mouse | entity
```
ngokevin: `Where the intersection ray is cast from (i.e., `entity` or `mouse`)`

tests/components/cursor.test.js
```diff
@@ -76,6 +76,16 @@ suite('cursor', function () {
         done();
       });
     });
+
+    suite('update', function () {
+      test('attach mousemove event listener on mouse mode', function () {
```
ngokevin: I'd also test removing mousemove event listeners

stephanedemotte:
I can change the rayOrigin from `entity` to `mouse` dynamically, but when i try to change from `mouse` to `entity` nothing append :/

dmarcos:
@stephanedemotte can you open an issue?

stephanedemotte:
done ;)

stephanedemotte:
@dmarcos do you have time to check the example ? #2929
Need to know if it's a bug or my code :/

Thank you !

PR #2871, opened 2017-07-15
[doc] a-text.md, raycasting

andreasplesch:
**Description:**
address raycasting
**Changes proposed:**
- add a limitations sections with a short explanation.

### DIFF
--- a/docs/primitives/a-text.md
+++ b/docs/primitives/a-text.md
@@ -41,3 +41,7 @@ Wraps the [text component][text].
 | wrap-count     | text.wrapCount     |
 | wrap-pixels    | text.wrapPixels    |
 | z-offset       | text.zOffset       |
+
+## Limitations
+
+The generated text is not suitable for cursor or raycaster detection. Please add a geometry component.
--- a/docs/primitives/a-text.md
+++ b/docs/primitives/a-text.md
@@ -14,10 +14,6 @@ Wraps the [text component][text].
 
 ```html
 <a-text value="Hello, World!"></a-text>
-
-<!-- for interactivity add geometry -->
-
-<a-text value="click me!" geometry="primitive:plane"></a-text>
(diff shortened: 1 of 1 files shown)


dmarcos:
We have a cursor example (https://github.com/aframevr/aframe/blob/master/examples/test/cursor/index.html). Do we add some text there to illustrate how to do it?

machenmusik:
I think the `a-plane` labeled with text in https://github.com/aframevr/aframe/blob/master/examples/test/text/scenarios.html would work with cursor, right?  If so, maybe we can point at something like that as example.

machenmusik:
Yeah, IMO the core text component should be thought of more as a specific type of material / decoration and so to be interacted with, the entity needs a geometry (e.g. plane or box) that defines its area / volume for interaction...?

andreasplesch:
A simplistic example on the page may suffice ?

```html
<a-text value="Hello, World!"></a-text>

<!-- for cursor detection add geometry -->

<a-text value="click me!" geometry="primitive:plane"></a-text>

```

andreasplesch:
The wording is friendlier now although I think it is preferable to say explicitly that text itself does not work with cursor or raycaster to avoid any confusion.

docs/primitives/a-text.md
```diff
@@ -14,6 +14,10 @@ Wraps the [text component][text].
 
 ```html
 <a-text value="Hello, World!"></a-text>
+
+<!-- for interactivity add geometry -->
```
ngokevin: I'd add this example below to the limitations, and then:

```<a-text value="Now interactable" geometry="primitive: plane"</a-text>```

andreasplesch: Good idea. Not sure if interactable is a proper word but it is better than interactive.

PR #2872, opened 2017-07-15
Update index.md

mmullen200:
**Description:**

I have proposed some small fixes for typos and idiom. I hope this helps!

**Changes proposed:**
The changes should be clear in the file. You'll see that I rearranged one sentence to be: "A-Frame aims to define fully immersive interactive VR experiences that go beyond basic 360&deg; content, making full use of positional tracking and controllers." I know that's a fairly large change, but the previous version of the sentence had a subject-verb agreement problem, basically saying "interactive experiences that GOES beyond..." My fix is the best solution I could find to fix the issue.

Line 111: It looks to me that the ".js" I deleted is just stray text at the end of the sentence. I hope I'm reading that right.

Line 132: "built on the notion of HTML" Is that really what we want to say here? Is there a better way to express this idea?


dmarcos:
Thank you! 👏

mmullen200:
My pleasure!

PR #2873, opened 2017-07-15
Wait for entities to load when the a-node has not been yet initialized (fix #2802)

### LINKED ISSUE #2802: Nested entities initialize out of order in scene without a-assets
**Description:**

I believe components can assume during their initialization that components on child nodes have already been initialized due to the inside-out initialization order. This assumption does not hold up in a scene with no `<a-assets>` declared. 

- A-Frame Version: 0.5.0
- Platform / Device: Win10, Chome, Chromium, Firefox, & Nightly

This scene errors when a component on `<a-locomotor>` expects an object to be returned from `getAttribute` on a child node component but gets a string instead: 

```html
<head>
  <title>Most Basic Super-Hands Example</title>
  <script src="https://aframe.io/releases/0.5.0/aframe.min.js"></script>
  <script src="//cdn.rawgit.com/donmccurdy/aframe-extras/v3.3.0/dist/aframe-extras.min.js"></script>
  <script src="https://rawgit.com/wmurphyrd/aframe-super-hands-component/master/dist/super-hands.js"></script>
</head>

<body>
  <a-scene>
    <a-locomotor>
      <!-- Make sure your super-hands entities also have controller and collider components -->
      <a-entity hand-controls="left" super-hands sphere-collider="objects: a-box"></a-entity>
      <a-entity hand-controls="right" super-hands sphere-collider="objects: a-box"></a-entity>
    </a-locomotor>
    <!-- hover & drag-drop won't have any obvious effect without some additional event handlers or components. See the examples page for more -->
    <a-box hoverable grabbable stretchable drag-droppable
    color="blue" position="0 0 -1"></a-box>
  </a-scene>
</body>
```

This scene, identical except for the addition of an empty `<a-assets>` loads successfully:

```html
<head>
  <title>Most Basic Super-Hands Example</title>
  <script src="https://aframe.io/releases/0.5.0/aframe.min.js"></script>
  <script src="//cdn.rawgit.com/donmccurdy/aframe-extras/v3.3.0/dist/aframe-extras.min.js"></script>
  <script src="https://rawgit.com/wmurphyrd/aframe-super-hands-component/master/dist/super-hands.js"></script>
</head>

<body>
  <a-scene>
    <a-assets></a-assets>
    <a-locomotor>
      <!-- Make sure your super-hands entities also have controller and collider components -->
      <a-entity hand-controls="left" super-hands sphere-collider="objects: a-box"></a-entity>
      <a-entity hand-controls="right" super-hands sphere-collider="objects: a-box"></a-entity>
    </a-locomotor>
    <!-- hover & drag-drop won't have any obvious effect without some additional event handlers or components. See the examples page for more -->
    <a-box hoverable grabbable stretchable drag-droppable
    color="blue" position="0 0 -1"></a-box>
  </a-scene>
</body>
```

This can definitely be solved in the component by not making the assumption, but I thought it was worth documenting for you.

dmarcos:
It checks for the tagName to be `a-node` , `a-entity` or a registered primitive.

### DIFF
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -1,5 +1,6 @@
 /* global MutationObserver */
 var registerElement = require('./a-register-element').registerElement;
+var isNode = require('./a-register-element').isNode;
 var utils = require('../utils/');
 
 var bind = utils.bind;
@@ -101,8 +102,7 @@ module.exports = registerElement('a-node', {
         if (this.hasLoaded) { return; }
 
         // Default to waiting for all nodes.
-        childFilter = childFilter || function (el) { return el.isNode; };
-
+        childFilter = childFilter || isNode;
         // Wait for children to load (if any), then load.
         children = this.getChildren();
         childrenLoaded = children.filter(childFilter).map(function (child) {
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -26,9 +26,11 @@ var TestComponent = {
 suite('a-entity', function () {
   setup(function (done) {
     var el = this.el = entityFactory();
-    el.addEventListener('loaded', function () {
+    var onLoaded = function () {
       done();
-    });
+      el.removeEventListener('loaded', onLoaded);
+    };
+    el.addEventListener('loaded', onLoaded);
   });
 
   teardown(function () {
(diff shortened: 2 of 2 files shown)


dmarcos:
We should make sure there are no other `a-nodes` the parent should be waiting for? `a-animation`, `a-asseet-item`... Not sure if `a-assets` need some rethinking to make sure we block rendering on them

src/core/a-node.js
```diff
@@ -97,11 +97,16 @@ module.exports = registerElement('a-node', {
...
         var childrenLoaded;
         var self = this;
+        var primitives = AFRAME.primitives.primitives;
 
         if (this.hasLoaded) { return; }
 
         // Default to waiting for all nodes.
-        childFilter = childFilter || function (el) { return el.isNode; };
+        childFilter = childFilter || function (el) {
+          var tagName = el.tagName.toLowerCase();
```
ngokevin: The prototype is not available?

src/core/a-node.js
```diff
@@ -97,11 +97,16 @@ module.exports = registerElement('a-node', {
...
+          var isEntity = tagName === 'a-entity' || tagName === 'a-node' || primitives[tagName] !== undefined;
+          return el.isNode || isEntity;
```
ngokevin: This function should check isNode the very first. If not, only the execute the lowerCase and other checks

dmarcos:
The prototype of the children elements is not set by the time `createdCallback` and `attachCallback` are called on the parent. I'm using the existing method `isNode` now that checks for the registered tags (https://github.com/aframevr/aframe/blob/master/src/core/a-register-element.js#L32). I think this should be sufficient.

tests/core/a-entity.test.js
```diff
@@ -641,6 +643,52 @@ suite('a-entity', function () {
         done();
       });
     });
+
+    test('wait for all the node children that are not yet nodes to load', function (done) {
```
ngokevin: These tests pass for me without the changes. Do these reproduce the issue?

dmarcos:
I had run the tests with breakpoints that introduce changes in the timing. I added extra conditions to the tests so they're checking what they're supposed to test: The parent waits for the children nodes to load when the `isNode` flag is not set yet (because `createdCallback` has not been yet invoked)

ngokevin:
Need to remove the `.only`.

ngokevin:
r+wc nice, i forgot we did that `isNode` function

PR #2876, opened 2017-07-17
Replace BlendCharacter dep with ObjectLoader.

### LINKED ISSUE #2616: No BlendCharacter in Three.js r85
**Description:**

`THREE.BlendCharacter` has been removed from Three.js r85.

So maybe we need to rewrite our `blend-character-model.js` in reference to  [Three.js webgl_animation_skinning_blending example](https://github.com/mrdoob/three.js/blob/dev/examples/webgl_animation_skinning_blending.html) or make our `THREE.BlendCharacter`.

donmccurdy:
Fixes #2616.

BlendCharacter was removed in the upcoming version of three.js. Unless someone has the original model I don't see any way to convert this to glTF, so instead I'm just using ObjectLoader and AnimationMixer directly.


src/components/hand-controls.js
```diff
@@ -83,6 +85,14 @@ module.exports.Component = registerComponent('hand-controls', {
     this.removeEventListeners();
   },
 
+  tick: function (t, dt) {
```
dmarcos: For consistency. Can we call t -> time and dt -> delta?

donmccurdy: 👍 done

src/components/hand-controls.js
```diff
@@ -143,7 +153,7 @@ module.exports.Component = registerComponent('hand-controls', {
    * Update handler. More like the `init` handler since the only property is the hand, and
    * that won't be changing much.
    */
-  update: function () {
+  update: function (prevHand) {
```
dmarcos: Can we expand to `previousHand` too?

donmccurdy: Done

dmarcos:
Sweet! 🍰 Thanks!

PR #2880, opened 2017-07-18
re-add controller events after unpausing scene (fixes #2879)

### LINKED ISSUE #2879: controller buttons/events don't work after pausing and unpausing scene
**Description:**

Blocks Inspector + Motion Capture work.

1. Enter VR. See controllers work when pressing buttons.
2. Open Inspector. Exit Inspector.
3. Enter VR again. Controller buttons do not work

- A-Frame Version: master/0.6.0
- Platform / Device: windows/vive
https://aframe.io/examples/showcase/tracked-controllers/

ngokevin:
**Description:**

The controller code is somewhat awkward since each component needs to follow a certain API and implements various methods, but this way works of making sure controller events get re-added on unpause.

**Changes proposed:**
- Set a state variable for checking whether events are on or not. Then the tracked-controls util can determine whether it needs to re-add events on unpause.
- [x] adjust tests

### DIFF
--- a/src/components/vive-controls.js
+++ b/src/components/vive-controls.js
@@ -94,6 +94,7 @@ module.exports.Component = registerComponent('vive-controls', {
     el.addEventListener('touchstart', this.onButtonTouchStart);
     el.addEventListener('model-loaded', this.onModelLoaded);
     el.addEventListener('axismove', this.onAxisMoved);
+    this.controlllerEventsActive = true;
   },
 
   removeEventListeners: function () {
@@ -105,6 +106,7 @@ module.exports.Component = registerComponent('vive-controls', {
     el.removeEventListener('touchstart', this.onButtonTouchStart);
     el.removeEventListener('model-loaded', this.onModelLoaded);
     el.removeEventListener('axismove', this.onAxisMoved);
+    this.controlllerEventsActive = false;
   },
 
   /**
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -72,6 +72,7 @@ module.exports.Component = registerComponent('daydream-controls', {
     el.addEventListener('touchend', this.onButtonTouchEnd);
     el.addEventListener('model-loaded', this.onModelLoaded);
     el.addEventListener('axismove', this.onAxisMoved);
+    this.controllerEventsActive = true;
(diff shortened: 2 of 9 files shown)


src/components/vive-controls.js
```diff
@@ -105,6 +106,7 @@ module.exports.Component = registerComponent('vive-controls', {
     el.removeEventListener('touchstart', this.onButtonTouchStart);
     el.removeEventListener('model-loaded', this.onModelLoaded);
     el.removeEventListener('axismove', this.onAxisMoved);
+    this.controlllerEventsActive = false;
```
dmarcos: Do we need a new flag? Can we just rely on the `isPaused` flag of the component?

src/components/vive-controls.js
```diff
@@ -94,6 +94,7 @@ module.exports.Component = registerComponent('vive-controls', {
     el.addEventListener('touchstart', this.onButtonTouchStart);
     el.addEventListener('model-loaded', this.onModelLoaded);
     el.addEventListener('axismove', this.onAxisMoved);
+    this.controlllerEventsActive = true;
```
dmarcos: I see that `this.removeEventListeners` is called on `pause` but `this.addEventListeners` is not called in play. We should probably call it there.

src/components/vive-controls.js
dmarcos: I think that removing / adding event listeners in the controller specific controls is not strictly necessary if `tracked-controls` pauses / plays as expected.

ngokevin: We need to call `addEventListener` in the `play` handler only when coming out of a pause. We don't want to call it directly because `utils.tracked-controls` will already call it and we don't want to call it on the first initialization.

ngokevin: Removing and adding controller specific event listeners is needed because we don't want to be adding Oculus Touch triggerdown when Oculus Touch controllers not connected, which would result in double-emit if we're using a Vive controller.

src/components/vive-controls.js
ngokevin: It is called in `play` via the tracked-controls util.

dmarcos:
Nice! 🎁

PR #2884, opened 2017-07-18
have default light cast shadows by default

ngokevin:
**Description:**

Doesn't hurt to set `castShadow` on the default light. I think it will only be used if any objects have the shadow component.

I also set the shadows on the Hello, World example.

screen shot 2017-07-18 at 4 45 22 pm

### DIFF
--- a/examples/boilerplate/hello-world/index.html
+++ b/examples/boilerplate/hello-world/index.html
@@ -8,10 +8,10 @@
   </head>
   <body>
     <a-scene>
-      <a-box position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9"></a-box>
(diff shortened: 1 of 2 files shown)


dmarcos:
@fernandojsg what do you think?

ngokevin:
This doesn't hurt if there aren't any shadow-casting objects. With this, then the only step to add shadows is just to add `shadow` attribute to entities. Rather than having to create your own light

fernandojsg:
I'm not sure how three.js handles a casting light without objects that cast or receive, but I don't think it's a big deal, probably it will just create an empty shadowmap or so, but I don't believe it will affect the performance.

I believe it could be nice to have it activated by default as in simple scenes it will give an overall nice quality and you don't need to add the light by yourself. and in more complex scenes probably you'll tweak the lights anyway so you don't care so much about the default ones.

PR #2893, opened 2017-07-21
Removing irrelevant text

ShaguftaMethwani:
**Description:**
Removing irrelevant text
**Changes proposed:**
- Removing text that might have come from copy-pasting the text from `rotation`
-
-


ngokevin:
thanks!

PR #2896, opened 2017-07-21
Expose emissive and emissiveIntensity Three.js properties to the standard material

joshmarinacci:
**Description:**

The standard Three.js material has settings to make a material be emissive, meaning it can project light even when the rest of the scene is dark. Currently the emissive property is not exposed inside AFrame. This patch adds the property to the standard Aframe material.

**Changes proposed:**
-
-
-


ngokevin:
Can you add this to the docs as well, thanks!

joshmarinacci:
I've added the docs.

PR #2901, opened 2017-07-26
add vertexColors to base material properties

ngokevin:
**Description:**

Allow for vertex or face coloring.

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -47,19 +47,20 @@ Here is an example of using an example custom material:
 The material component has some base properties. More properties are available
 depending on the material type applied.
 
-| Property    | Description                                                                                                                                       | Default Value |
-|-------------|---------------------------------------------------------------------------------------------------------------------------------------------------|---------------|
-| alphaTest   | Alpha test threshold for transparency.                                                                                                            | 0             |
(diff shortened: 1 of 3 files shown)


docs/components/material.md
```diff
@@ -47,19 +47,20 @@ Here is an example of using an example custom material:
...
+| opacity      | Extent of transparency. If the `transparent` property is not `true`, then the material will remain opaque and `opacity` will only affect color.   | 1.0           |
+| repeat       | Texture repeat to be used.                                                                                                                        | {x: 1, y: 1}  |
+| shader       | Which material to use. Defaults to the [standard material][standard]. Can be set to the [flat material][flat] or to a registered custom material. | standard      |
+| side         | Which sides of the mesh to render. Can be one of `front`, `back`, or `double`.                                                                    | front         |
+| transparent  | Whether material is transparent. Transparent entities are rendered after non-transparent entities.                                                | false         |
+| vertexColors | Whether to use vertex or face colors to shade the material. Can be one of `none`, `vertex`, or `face`.                                            | none          |
```
fernandojsg: I don't like the name either but I agree that if we're consistent with three.js we should keep using it https://github.com/mrdoob/three.js/blob/dev/src/materials/Material.js#L27 :\

PR #2902, opened 2017-07-26
added Google Blocks in Where can I find assets?

Utopiah:
**Description:** added Google Blocks in Where can I find assets?

**Changes proposed:**
- added links to Google Blocks on top since it's the largest easiest to use library already (using more complex assets with textures for example from TurboSquid usually brings newcomers to a world of pain)


dmarcos:
Thanks!

PR #2903, opened 2017-07-26
link component releasted, using past tense

Utopiah:
**Description:** link component releasted, using past tense

**Changes proposed:**
- link component for link traversal was (instead of will be) released with A-Frame 0.6.1


dmarcos:
Thanks!

PR #2904, opened 2017-07-26
Add the packaging metadata to build the aframe snap

elopio:
**Description:**
This package will let you publish the latest aframe in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

**Changes proposed:**
- Add the snapcraft.yaml file

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: aframe
+version: master
+summary: A web framework for building virtual reality experiences
+description: |
+  Make WebVR with HTML and Entity-Component
+  Works on Vive, Rift, Daydream, GearVR, desktop
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  aframe:
+    command: sh -c \"cd $SNAP/lib/node_modules/aframe && node ./scripts/budo -t envify\"
+    daemon: simple
+    plugs: [network, network-bind]
+
+parts:
+  aframe:
+    source: .
+    plugin: nodejs
+    node-packages: [budo]
+    install: cp -R examples $SNAPCRAFT_PART_INSTALL/lib/node_modules/aframe


elopio:
Hello!

Here at ubuntu we are working in a new packaging and delivery system. If you haven't heard about snaps before, here's a lot of info: https://snapcraft.io/

Obviously, aframe is a building block for other applications, delivering a packaged version of aframe itself is not so useful. But I think having this package in the Ubuntu store will be very nice. First, it will be a showcase for all the Ubuntu users, who then can jump into developing their own scenes. And second, the store has an edge channel that's very useful for crowdtesting. You can automate the delivery of every change made to the master branch, put it in edge, and ask your early adopters and the community of ubuntu testers to give it a try. All of this can be done with a few clicks in https://build.snapcraft.io.

To test this in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/aframe
    $ cd aframe
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous
    $ sensible-browser localhost:9000

(--dangerous because the snap you build doesn't come signed from the store. Once you publish it, your users will just have to do sudo snap install aframe).

If you have any questions, please let me know.

pura vida.

ngokevin:
thanks for the pr! I don't think this should be part of the main repo since it's a for a third party packaging tool for a fairly niche store. This would belong best perhaps as a tutorial since people would have to modify the metadata anyways.

ngokevin:
Closing for now, but we can keep the conversation going

elopio:
@ngokevin well, it depends on how you define niche store. There are millions of ubuntu users. They will all be able to see the aframe showcase snap in the software center, and install it with one click.

We are working on making this extremely low-maintenance. After getting the snapcraft.yaml, you just have to go to https://build.snapcraft.io once, and then it will take care of building and publishing new versions for you. So, I think that giving it a try will require just a little effort, and then you can see the stats and the feedback you receive from the users to decide if you want to make a stable release in the store and keep maintaining it.

Of course, I understand if you decide not to have a snap at this moment. If you change your mind, just ping me and I'll be back to give you a hand.

The tutorial sounds like a very good idea. I will be learning more about aframe next month (that's why I found it, and I made this PR to experiment a little with it). So, I will write the tutorial, about how to deliver a app made with aframe, after I finish my first scene. electron-builder already supports snaps, so it might be fun to combine both :)

Thanks for the review!
pura vida.

PR #2906, opened 2017-07-27
Fixes line update bug and add example

fernandojsg:
I've just fixed the `line` component as the `isEqualVec3` was wrong and it was doing exactly the opposite, so updating the vertices of non-changed position and it wasn't updating the modified vertices.

I've added also a simple test example with:
- `start` and `end` parameter
- just `start`
- just `end`
- dynamic changing `color`
- dynamic modifying the `end` position.

I've also changed the default color to white `#fff` as it's the one we used in most of our demos for raycast selector.


examples/test/lines/index.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
```
ngokevin: this would be better as a unit test

ngokevin:
Thanks, I was running into this earlier.

examples/test/lines/index.html
fernandojsg: Ok, I'll do that

src/components/line.js
```diff
@@ -5,7 +5,7 @@ module.exports.Component = registerComponent('line', {
   schema: {
     start: {type: 'vec3', default: {x: 0, y: 0, z: 0}},
     end: {type: 'vec3', default: {x: 0, y: 0, z: 0}},
-    color: {type: 'color', default: '#74BEC1'},
+    color: {type: 'color', default: '#fff'},
```
ngokevin: The default color is used by the raycaster and laser controls, I wouldn't make it white since it'd be hard to see.

fernandojsg: @ngokevin ok, should I just commit directly to master reverting that change?

tests/components/line.test.js
```diff
@@ -0,0 +1,80 @@
+/* global assert, process, setup, suite, test */
+var entityFactory = require('../helpers').entityFactory;
+
+suite('line', function () {
+  setup(function (done) {
+    var el = this.el = entityFactory();
```
ngokevin: Not a big deal, but for future tests, I'd move `var el;` up a level and use that so we don't need to do `var el = this.el`. And also create a `var component;` and store the instance in there so we don't have to do `el.components.X`

fernandojsg: Ok, I'll just do a quick PR fixing both issues

src/components/line.js
ngokevin: yah sure

PR #2907, opened 2017-07-27
Refactoring line.test and revert default color

fernandojsg:
As discussed in https://github.com/aframevr/aframe/pull/2906
/cc @ngokevin


ngokevin:
thanks!

PR #2910, opened 2017-07-28
Remove --save flag from yarn cmd in readme

mkungla:


ngokevin:
i'd just remove the `--save` piece and leave everything else

mkungla:
updated

PR #2911, opened 2017-07-29
[docs] Describe use and limitations of AO maps.

### LINKED ISSUE #2721: Ambient occlusion maps require second UV channel

donmccurdy:
Fixes #2721.

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -87,23 +87,23 @@ These properties are available on top of the base material properties.
 
 | Property                      | Description                                                                                                                                     | Default Value |
 |-------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|---------------|
-| ambientOcclusionMap           | Ambient occlusion map. Used to add shadows to the mesh. Can either be a selector to an `<img>` an inline URL.                                   | None          |
+| ambientOcclusionMap           | Ambient occlusion map. Used to add shadows to the mesh. Can either be a selector to an `<img>`, or an inline URL. Requires 2nd set of UVs (see below). | None          |
 | ambientOcclusionMapIntensity  | The intensity of the ambient occlusion map, a number between 0 and 1.                                                                           | 1             |
 | ambientOcclusionTextureRepeat | How many times the ambient occlusion texture repeats in the X and Y direction.                                                                  | 1 1           |
 | ambientOcclusionTextureOffset | How the ambient occlusion texture is offset in the x y direction.                                                                               | 0 0           |
 | color                         | Base diffuse color.                                                                                                                             | #fff          |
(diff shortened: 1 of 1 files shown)


colinfizgig:
Question:  Are the second UV maps used for Lightmaps and AO maps limited to duplicating the first UV set on the object?  or can we set up a second UV set in a 3d application and use that instead?  This could enable things like lightmappers for Aframe etc to work.  Currently it seems that the gltf importer duplicates the first UV set automatically, which precludes using different UV maps that optimize texture sizing and layout on 3d objects.  The use case for allowing custom uv sets would be that you could take a repeating brick texture and tile it while having a Lightmap for the entire scene instead of individual lightmaps for each object.

donmccurdy:
The glTF loader will only duplicate the UVs if there is no second UV set, so yes— you can create a custom second set of UVs in your software instead. The second UVs can only be used for AO/lightmap, though.

PR #2912, opened 2017-07-29
Adding position offset for 3DoF controllers

realitylab-ruben:
**Description:**
I've only started working with the Gear controller and AFrame a few days ago, and the first thing I noticed was that the GearVR Controller wasn't exactly working as I expected. Instead of turning around the camera like most applications do, the controller was just rotating in place. 

While this makes sense, I felt that it was a bit confusing and had to work around the component to get the controller to show up in its right place. I think that the GearVR controller should be in a default position just in front of the camera and rotate around the viewer instead of just staying in one spot.

I haven't tested it yet, but I'm assuming that the Daydream has the exact same issue.

**Changes proposed:**
- Add a ```positionOffset``` vector3 value to schema of the ```daydream-controls``` and ```gearvr-controls``` components.
- A child entity will be generated with its position set to the value in the schema
- This child entity will contain the controller model. Because the ```tracked-controls``` will be its parent, the controller will automatically spin around the camera.
- If no position offset is given, the default will be set to ```0.1 -0.1 -0.275```. X will be set to ```-0.1``` if the ```hand``` attribute is set to ```left```.

The PR currently already had the changes mentioned above implemented for the ```gearvr-controls```. Before putting more effort into creating a proper example and changing the ```daydream-controls``` component as well, I first would like to get some feedback before moving on.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -183,6 +183,7 @@
     <h2>Showcase</h2>
 
     <ul class="links">
+      <li><a href="showcase/3dof-controllers/">3DOF Controllers</a></li>
       <li><a href="showcase/anime-UI/">Anime UI</a></li>
       <li><a href="showcase/composite/">Composite</a></li>
       <li><a href="showcase/curved-mockups/">Curved Mockups</a></li>
--- a/examples/showcase/3dof-controllers/components/geometry-changer.js
+++ b/examples/showcase/3dof-controllers/components/geometry-changer.js
@@ -1,0 +1,15 @@
+/* global AFRAME */
+AFRAME.registerComponent('geometry-changer', {
+  init: function () {
+    this.el.addEventListener('raycaster-intersected', this.onIntersected.bind(this));
+    this.el.addEventListener('raycaster-intersected-cleared', this.onIntersectedCleared.bind(this));
+  },
+
+  onIntersected: function () {
+    this.el.setAttribute('color', 'red');
+  },
+
+  onIntersectedCleared: function () {
+    this.el.setAttribute('color', 'blue');
+  }
+});
--- a/examples/showcase/3dof-controllers/index.html
+++ b/examples/showcase/3dof-controllers/index.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Tracked Controls</title>
+    <meta name="description" content="Tracked Controls – A-Frame">
+    <script src="../../../dist/aframe-master.js"></script>
+    <script src="./components/geometry-changer.js"></script>
+</head>
+<body>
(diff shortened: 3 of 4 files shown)


dmarcos:
Is not that misleading? The controller is 3DOF and there wont't be 1:1 matching between the hands and the controller. It might be ok for some interactions. Not sure if the offset should be part of the schema since you could achieve the same effect by nesting the controllers inside an entitiy

dmarcos:
Is not that misleading? The controller is 3DOF and there wont't be 1:1 matching between the hands and the controller. It might be ok for some interactions. Not sure if the offset should be part of the schema since you could achieve the same effect by nesting the controllers inside an entitiy

realitylab-ruben:
It would indeed, but I would argue that this would make the model part of the component fairly useless. I can't think of any particular example where the user would want the controller to float in a static point in space where it will only rotate at a single point.

It's also a mechanic that I've, so far, experienced in all Gear-related applications. The only example I could find on short notice is in the [official Samsung Gear VR tutorial](https://www.youtube.com/watch?v=gghYoUlhyVU#t=1m42). In general, the controller is rendered away from the user and rotating around him to give a sense of where he is holding it.

The SDK for both [Daydream](https://developers.google.com/vr/unity/reference/class/gvr-arm-model) and [Gear](https://developer.oculus.com/documentation/unity/latest/concepts/unity-ovrinput/#unity-ovrinput-d7803e81) offer a similar functionality out of the box with their respective Unity SDKs.

What I would, however, admit is that this solution now only works to facilitate the model position and nothing more. If you'd have to add a raycaster, you would need to put the raycaster in another  child entity to the entity holding the gearvr-controls. [I did something similar in my example.](https://github.com/aframevr/aframe/pull/2912/files#diff-321ec347dea00ae9da476ecb4f4ca732R17)

My personal problem is that both ```daydream-controls``` and ```gear-controls``` don't use the same practices out of the box that the official Unity/Unreal SDKs do, and that's what I'm trying to solve here. Maybe I have jumped the gun a bit by trying to fix this in the AFrame repository and I should first see about turning this into a standalone component.

machenmusik:
There should be an arm model in place for both gearvr-controls and daydream-controls, unless you have disabled it.  Are you proposing a different arm model?  Or is the arm model not doing the right thing for your use case so you were proposing a different strategy?

dmarcos:
Yeah. The arm model should be the one applying the offset

realitylab-ruben:
Hmm, that's weird, because I'm not seeing it in the Samsung Gear. I've only just now noticed that there is code for an arm model, but it's not working in my own projects, at least not on the Samsung Gear.

This was my mistake then, my apologies. I'll close this pull request and open up an issue. Thanks for your feedback.

Issue #2849, opened 2017-07-06
Safari iOS: On fullscreen get "undefined is not an object (evaluating 'e.display.isPresenting')"


studiouniverse:
On enter fullscreen (when tapping fullscreen-goggle-button) get "undefined is not an object (evaluating 'e.display.isPresenting')" when using Safari iOS.

The VR continues to work, but rotating the phone back to portrait does not clear the Aframe dual-vision VR mode until page refresh.

```
aframe-master.min.js:303:4802
a-scene.js:309
TypeError: undefined is not an object (evaluating 'e.display.isPresenting')
```

- A-Frame Version: 0.6.0
- Platform / Device: iPhone 7 iOS 10.3.1 Safari
- Reproducible Code Snippet or URL: http://aframe.io (first demo that comes up), also tested on personal code using https://aframe.io/releases/0.6.0/aframe.min.js

jaredpike:
The console error seems unrelated to the issue. But to fix it, line 309 in `src/core/scene/a-scene.js` should be `evt.detail.display.isPresenting`.

dmarcos:
Closing because no longer reproduces in newer versions of A-Frame / iOS.

PR #2920, opened 2017-08-01
fix missing detail property in a-scene onVRPresentChange

hapticdata:
**Description:** `onVRPresentChange` throws error that `evt.display` is undefined. The `display` property exists under `CustomEvent#detail` property and not directly on `CustomEvent`.

**Changes proposed:**
- look for the `isPresenting` at `evt.detail.display.isPresenting`

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -305,8 +305,10 @@ module.exports.AScene = registerElement('a-scene', {
      */
     onVRPresentChange: {
       value: function (evt) {
+        // Polyfill places display inside the detail property
+        var display = evt.display || evt.detail.display;
         // Entering VR.
-        if (evt.detail.display.isPresenting) {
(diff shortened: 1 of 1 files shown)


ngokevin:
Which platform/browser? Sure it's not on the event iself? https://w3c.github.io/webvr/spec/1.1/#interface-vrdisplayevent

hapticdata:
ah, right. I forgot [I ran into this on webvr-ui](https://github.com/googlevr/webvr-ui/blob/master/src/webvr-manager.js#L213-L219) as well.

When using the WebVR API it is directly on `event.display` when using a polyfill it is placed inside the `CustomEvent`'s `detail` property.

Pull request updated to look for either.

hapticdata:
this fixes issue #2849

ngokevin:
thanks!

PR #2921, opened 2017-08-01
On Daydream devices, use native HMD pose orientation rather than sensor fusion

jsantell:
**Description:**

Due to the `isMobile` check occuring before the `vr-mode` check, when Daydream is in VR mode, as it's also a mobile device, sensor fusion controls take precedence. Is there scenario I'm not considering in this branching logic?

**Changes proposed:**

If in VR mode with HMD data, use that -- fall back to sensor fusion on mobile after.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -165,14 +165,21 @@ module.exports.Component = registerComponent('look-controls', {
     hmdQuaternion = hmdQuaternion.copy(this.dolly.quaternion);
     hmdEuler.setFromQuaternion(hmdQuaternion, 'YXZ');
 
-    if (sceneEl.isMobile) {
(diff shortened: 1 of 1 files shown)


dmarcos:
Nice, thanks! I'm far from my headsets this week. I will do a sanity check as soon as I can next week and will merge.

dmarcos:
@jsantell while your logic is probably more appropriate on Daydream (because you disable panning using the finger) I think the result should be identical (both logic old and new use native HMD orientation given by hmdEuler). Do you see any difference in your Daydream device when entering proper VR mode (webvr in Chrome Daydream is hidden behind origin trials....) ?

cvan:
this seems pretty close. @jsantell: friendly ping?

jsantell:
This was originally for Aframe+ WebARonAR[Core|kit] support, and some falsy assumptions on my end I think. Closing for now!

Issue #2782, opened 2017-06-15
A-frame Random page reloads/refresh


CallumMitchellRV:
I've been working on a VR game within A-Frame. When I started testing on devices (iPhone 7/iPhone 7 Plus running Safari, and Samsung Galaxy S6 running Chrome), I began to notice the page would randomly refresh, which is an issue for us, as it restarts the game.

This is happening with the demos on their website, as well as our self-hosted game.

Has anyone experienced this issue, and/or know what the issue or the solution is?

- A-Frame Version: 0.5.0
- Platform / Device: iPhone 7/7 Plus
- Reproducible Code Snippet or URL: [https://aframe.io/examples/hello-metaverse/]

dmarcos:
It restarts randomly, while interacting with the scene, in 2D mode, in VR mode?

CallumMitchellRV:
It hasn't happened once on a desktop (am I correct in saying that 2D is desktop). We've only had this page refresh issue on devices, most regularly in VR mode. I believed it was an issue on our end, until we started testing on a-frame.io examples, and it happened there too.

dmarcos:
I can't reproduce on my iPhone 7 Plus with https://aframe.io/examples/hello-metaverse/.
@CallumMitchellRV Do you think we can close this? Is it still a problem for you?

CallumMitchellRV:
We're still very much having this issue on multiple devices on a-frame's own projects. Admittedly, on the iPhone 7 Plus it's less regular than the iPhone 7, sometimes 5 minutes, sometimes 1 minutes. I recommend leaving your phone in the VR mode until it happens. If you notice that i's no longer in VR mode, a page refresh has occurred. 

I'm very much in need of a solution or an explanation to this, so I'm not prepared to close this issue at this point.

dmarcos:
@CallumMitchellRV Can you try with master branch and see if it still reproduces?

jonnyharper:
Hi, I'm having the same issue. On the basic Hello Web VR example, if I just leave my iPhone sat for a few minutes, I'm booted out of VR mode and the attached logs in console.

![screen shot 2017-06-16 at 09 16 04](https://user-images.githubusercontent.com/2477827/27218114-b1f4f248-5274-11e7-86c4-97fbc68d7006.png)

CallumMitchellRV:
[...]

PR #2925, opened 2017-08-02
fix page-reloads on iOS (#2782)

klaasman:
**Description:**
On iOS devices, the application seems to randomly reload. This has been pinpointed to a reload after 30 seconds of inactivity while in VR mode, caused by the iOS wakelock feature. As seen in other wakelock-implementations and issues, the `window.location = window.location` creates this issue.

**Changes proposed:**
Update the wakelock implementation to match other implementation which don't cause this reload-bug.

One thing to mention is the wakelock does not work on iOS (11), however I think it's better to have an application which will sleep after X seconds, instead of an application which will reload every X seconds.

### DIFF
--- a/vendor/wakelock/wakelock.js
+++ b/vendor/wakelock/wakelock.js
@@ -47,9 +47,9 @@ function iOSWakeLock() {
   this.request = function() {
     if (!timer) {
       timer = setInterval(function() {
-        window.location = window.location;
+        window.location.href = '/';
         setTimeout(window.stop, 0);
-      }, 30000);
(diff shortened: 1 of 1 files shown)


dmarcos:
Have you tried on Android?

klaasman:
No, but the touched lines belong to the `iOSWakeLock()` – it shouldn't have any impact on Android.

klaasman:
Or I'm missing something, in that case: enlighten me :-)

dmarcos:
Thanks! 🥇

PR #2926, opened 2017-08-02
[docs] add example for utils throttle

dirkk0:
demo here:
http://curious-electric.com/aframe/throttle/

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -156,11 +156,12 @@ Checks if device is a smartphone. Returns a `boolean`.
 
 ### `AFRAME.utils.throttle (function, minimumInterval [, optionalContext])`
 
+[lodash]: https://lodash.com/docs/#throttle
+
 Returns a throttled function that is called at most once every
 `minimumInterval` milliseconds. A context such as `this` can be provided to
-handle function binding for convenience.
-
-`throttle` should be used on non-tick functions. `throttleTick` should be used for tick functions.
+handle function binding for convenience. The same as [lodash's
+`throttle`][lodash].
 
 ```js
 AFRAME.registerComponent('foo', {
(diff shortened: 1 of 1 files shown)


docs/core/utils.md
```diff
@@ -154,13 +154,27 @@ Checks if device is a smartphone. Returns a `boolean`.
...
 
-### `AFRAME.utils.throttle (function, interval [, optionalContext])`
+### `AFRAME.utils.throttle (function, minimumInterval [, optionalContext])`
 
 Returns a throttled function that is called at most once every
 `minimumInterval` milliseconds. A context such as `this` can be provided to
 handle function binding for convenience.
 
-### `AFRAME.utils.throttleTick (function (t, dt) {...}, interval [, optionalContext])`
+
```
ngokevin: big empty line here

docs/core/utils.md
```diff
@@ -154,13 +154,27 @@ Checks if device is a smartphone. Returns a `boolean`.
...
+```js
+AFRAME.registerComponent('foo', {
```
ngokevin: `throttle` should be used on non-tick functions. `throttleTick` should be used for `tick` functions.

ngokevin:
It's not quite correct, I would not use `throttle` when `throttleTick` should be used. Throttling `tick` shouldn't be explained in the `throttle` section. I don't think `throttle` needs to an example, `throttleTick` could have one though.

dirkk0:
I don't quite understand? What would I use if I need to fire something
every 1/10 second and still need the regular tick function?
As far as I understood (and I tested it) with throttleTick the regular tick
gets replaced, at least in the given example.

dirkk0:
ok, hopefully this works.

PR #2935, opened 2017-08-06
option to not hide stats while in vr mode

ngokevin:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
I wanted to leave the default behavior, mostly because of mobile right now.

ngokevin:
Ah, I see. Yeah, the stats are pretty inefficient which can make the scene slower by itself. Right now, I just attached the fps counter to the controller which works enough.

PR #2936, opened 2017-08-06
remove inline fns for crit code paths in tracked-controls/raycaster/component for gc

ngokevin:
**Description:**

**Changes proposed:**
- In tracked-controls/raycaster `tick` and component `buildData`, don't create inline functions (`forEach`, `filter`) nor use callbacks. Use simple `for` loop to try to reduce references that will need to get cleaned up by garbage collector.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -144,74 +144,87 @@ module.exports.Component = registerComponent('raycaster', {
   /**
    * Check for intersections and cleared intersections on an interval.
    */
-  tick: function (time) {
-    var el = this.el;
-    var data = this.data;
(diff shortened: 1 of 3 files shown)


dmarcos:
I believe what you're doing is correct by my understanding on how lexical scoping and closures work in JS. I have not proof though that iterators are measurably worse since the JS interpreter I'm sure does optimizations under the hood to reduce or eliminate potential overhead. I think we can merge this because at worse it will be innocuous and we have good test coverage to detect regressions. We're also one month out from release and we will likely catch any bugs that the unit tests might miss.

PR #2939, opened 2017-08-07
Prevent cloning objects into previousData that are not plain objects

dmarcos:
`a-saturday-night` uses `selector` properties that change at run time. This triggered `Converted circular structure to JSON` errors when trying to clone into previousData A-Frame entity elements that hold references to other A-Frame entities via components .We should only be cloning plain objects.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -336,21 +336,23 @@ Component.prototype = {
     var isSinglePropSchema = isSingleProp(schema);
     var mixinEls = this.el.mixinEls;
     var previousData;
-
     // 1. Default values (lowest precendence).
     if (isSinglePropSchema) {
-      // Clone default value if object so components don't share object.
-      data = typeof schema.default === 'object' ? utils.clone(schema.default) : schema.default;
+      // Clone default value if plain object so components don't share the same object
+      // that might be modified by the user.
+      data = schema.default.constructor === Object ? utils.clone(schema.default) : schema.default;
     } else {
       // Preserve previously set properties if clobber not enabled.
       previousData = !clobber && this.attrValue;
-      // Clone default value if object so components don't share object
-      data = typeof previousData === 'object' ? utils.clone(previousData) : {};
+      // Clone previous data to prevent sharing references with attrValue that might be
+      // modified by the user.
+      data = typeof previousData === 'object' ? cloneData(previousData) : {};
 
       // Apply defaults.
       for (i = 0, keys = Object.keys(schema), keysLength = keys.length; i < keysLength; i++) {
         defaultValue = schema[keys[i]].default;
         if (data[keys[i]] !== undefined) { continue; }
+        // Clone default value if object so components don't share object
         data[keys[i]] = defaultValue && defaultValue.constructor === Object
           ? utils.clone(defaultValue)
           : defaultValue;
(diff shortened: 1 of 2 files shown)


ngokevin:
Can you paste pseudocode that describes "`selector` properties being changed at runtime"?

src/core/component.js
```diff
@@ -354,21 +354,23 @@ Component.prototype = {
...
     var mixinEls = this.el.mixinEls;
     var previousData;
-
     // 1. Default values (lowest precendence).
     if (isSinglePropSchema) {
-      // Clone default value if object so components don't share object.
-      data = typeof schema.default === 'object' ? utils.clone(schema.default) : schema.default;
+      // Clone default value if plain object so components don't share the same object
+      // that might be modified by the user.
+      data = typeof schema.default === 'object' && schema.default.constructor === Object ? utils.clone(schema.default) : schema.default;
```
ngokevin: Do we need both `typeof ` and `.constructor` checks?

src/core/component.js
```diff
@@ -487,6 +489,24 @@ module.exports.registerComponent = function (name, definition) {
...
+* Clones only the properties that are plain objects while
+* keeping a reference for the rest.
+*
+* @param data - Component data to clone.
+* @returns Cloned data.
+*/
+function cloneData (data) {
+  var clone = {};
+  var parsedProperty;
+  for (var key in data) {
```
ngokevin: declare vars up top.

i wonder if we should start to use `for (key in obj)` for every loop over an object where we know it's a plain JS object. e.g., change some of the code in the `buildData` to also use `for...in`

dmarcos:
An example is: `el.setAttribute('avatar', 'el', '#bear')` A selector changes at runtime and the previous is cloned as oposed to being defined statically in the HTML

cvan:
@fernandojsg is this true that this blocks https://github.com/aframevr/a-saturday-night/pull/118#issuecomment-320808995 ?

fernandojsg:
@cvan yes, it was described on my PR https://github.com/aframevr/a-saturday-night/pull/117 (I closed it because @caseyyee and I were doing practically the same)

ngokevin:
Speaking towards the technicals, I start to often use strings rather than `selector` property type. The `selector` property time has given us issues like these, keeping a DOM object in state. And my problems were I often wanted a reference to the raw selector string in case I wanted to re-run it. Still useful to have, but if you ever have issues with `selector` property type, fall back to string.

dmarcos:
I addressed the review comments and fixed the test. Ok to merge?

src/core/component.js
```diff
@@ -336,21 +336,23 @@ Component.prototype = {
...
+      // that might be modified by the user.
+      data = schema.default.constructor === Object ? utils.clone(schema.default) : schema.default;
     } else {
       // Preserve previously set properties if clobber not enabled.
       previousData = !clobber && this.attrValue;
-      // Clone default value if object so components don't share object
-      data = typeof previousData === 'object' ? utils.clone(previousData) : {};
+      // Clone previous data to prevent sharing references with attrValue that might be
+      // modified by the user.
+      data = typeof previousData === 'object' ? cloneData(previousData) : {};
```
ngokevin: I believe you can use `utils.extend({}, previousData)`. The problem with `utils.clone` is it uses JSON to do it.

ngokevin:
r+wc

src/core/component.js
dmarcos: `utils.extend` copies references. I still want to clone the plain object properties

PR #2940, opened 2017-08-07
add basic perf demo with controller/raycaster and in-vr stat counter

ngokevin:
**Description:**

I wanted a base example to do performance profiles on with some meshes, controllers, and a couple raycasters. I'm seeing some garbage collection hangs, try to clear that up within A-Frame.

**Changes proposed:**
- Add example with in-vr FPS counter attached to basic controller.
- For desktop testing

### DIFF
--- a/examples/performance/in-vr/components/stats-in-vr.js
+++ b/examples/performance/in-vr/components/stats-in-vr.js
@@ -1,0 +1,38 @@
+/* global AFRAME */
+AFRAME.registerComponent('stats-in-vr', {
+  schema: {
+    // Number of frames for moving average.
+    frames: {default: 30}
+  },
+
+  init: function () {
+    this.renderTimes = [];
+  },
+
+  tick: function (t, dt) {
+    var color;
+    var fps;
+    var renderTimes = this.renderTimes;
+
+    renderTimes.push(dt);
+    if (renderTimes.length > this.data.frames) { renderTimes.shift(); }
+
+    fps = renderTimes.length / (renderTimes.reduce(sum) / 1000);
+
+    color = 'green';
+    if (fps < 80) {
+      color = 'red';
+    } else if (fps < 85) {
+      color = 'orange';
+    } else if (fps < 89) {
+      color = 'yellow';
+    }
+
+    if (color) { this.el.setAttribute('text', 'color', color); }
+    this.el.setAttribute('text', 'value', fps.toFixed(0));
+  }
+});
+
+function sum (a, b) {
+  return a + b;
+}
--- a/examples/index.html
+++ b/examples/index.html
@@ -8,33 +8,27 @@
       html {
         font-size: 14px;
       }
-
       body {
         background: rgb(239,45,94);
         color: #fff;
         font: normal 500 1.2rem/1.2 Inconsolata, Andale Mono, Courier New, monospace;
         padding: 1rem 2rem;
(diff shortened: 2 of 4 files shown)


ngokevin:
Gonna try one that doesn't depend on rStats which seems like a performance hog.

ngokevin:
Updated using `dt` to measure FPS with a moving average.

examples/performance/in-vr/components/stats-in-vr.js
```diff
@@ -0,0 +1,42 @@
...
+    color = 'green';
+    if (fps < 80) {
+      color = 'red';
+    } else if (fps < 85) {
+      color = 'orange';
+    } else if (fps < 89) {
+      color = 'yellow';
+    }
+
+    if (color) {
```
dmarcos: one liner since vertical space in your screen is at a premium

examples/performance/in-vr/components/stats-in-vr.js
```diff
@@ -0,0 +1,42 @@
...
+    this.renderTimes = [];
+  },
+
+  tick: function (t, dt) {
+    var color;
+    var fps;
+    var renderTimes = this.renderTimes;
+
+    renderTimes.push(dt);
+    if (renderTimes.length > this.data.frames) {
```
dmarcos: one liner

ngokevin:
updated

examples/performance/in-vr/components/stats-in-vr.js
```diff
@@ -0,0 +1,38 @@
...
+    var fps;
+    var renderTimes = this.renderTimes;
+
+    renderTimes.push(dt);
+    if (renderTimes.length > this.data.frames) { renderTimes.shift(); }
+
+    fps = renderTimes.length / (renderTimes.reduce(sum) / 1000);
+
+    color = 'green';
+    if (fps < 80) {
```
dmarcos: Not sure if these ranges are useful. I think 8x is still an acceptable value since the asynchronous re projection of the different runtimes will fill the gaps. I bet that the stats will be most of the time `yellow` or `orange`. What about:
fps > 85 green
fps 70-85 yellow
fps 60-70 orange
fps > 60 red

ngokevin: These are moving averages, so if the last 45 frames gave an average of < 80, it's bad. These ranges are from when I was trying it out in VR and feeling what feels bad. When the average was below 80, it felt pretty bad. I only felt good if it was stable at 90. Even 89 indicated there was a frame drop or two somewhere.

ngokevin: I'm going to publish this component separately soon anyways. Useful in other projects

ngokevin: Oh, and I think if these ranges represented stable FPS, your ranges would be good, but they take into account frame drops.

machenmusik:
remember, target FPS changes per platform (desktop generally 90fps in-VR but some may be lower, and generally capped 60fps out-of-VR; mobile generally 60fps, if that, either way)

ngokevin:
Only concerned bout desktop VR with controllers here

machenmusik:
well, ok, but from your description it seems as though it would work just fine for mobile except for your picking the number 90 ;-)

Issue #2938, opened 2017-08-07
Object coordinates in embedded mode


Alxblsk:
- A-Frame Version: 0.6.1
- Platform / Device: Desktop/Mobile
- Reproducible Code Snippet or URL: https://codepen.io/Alxblsk/pen/NvpPPv

Hello all,

I've started learning A-Frame and discovered that in embedded mode clicking on a primitive may not always perform correct in cases, when a-frame container is not fully visible or container is smaller  than browser height. Please take a look on my snippet. With that margin, you can click on the empty space above primitives, but not at the bottom of them. It looks weird.

This is video I've recorded to demonstrate an issue: https://monosnap.com/file/bLZhB5MOz4ChoNy0KwY9M3r4pQzTJI

calrk:
I've noticed this on the mouse component before, and I believe it has to do with these lines (153-154 components/cursor.js)
```
mouse.x = (evt.clientX / window.innerWidth) * 2 - 1;
mouse.y = -(evt.clientY / window.innerHeight) * 2 + 1;
```

I fixed it in the component I used by using the bounding client of the canvas to calculate the values
```
const bounds = this.el.sceneEl.canvas.getBoundingClientRect();
const width = bounds.width;
const height = bounds.height;
const left = evt.clientX - bounds.left;
const top = evt.clientY - bounds.top;

mouse.x = (left / width) * 2 - 1;
mouse.y = - (top / height) * 2 + 1;
```

ngokevin:
Thanks! PR appreciated to use canvas size isntead in `cursor.js`. Wonder if `getBoundingClientRect` is performant to be called on every tick though.

calrk:
Made the pull request here:
https://github.com/aframevr/aframe/pull/2942

calrk:
@Alxblsk I think the issue you are having is part of the component you are using rather than a-frame, but for the same reason.
https://github.com/jesstelford/aframe-click-drag-component/blob/master/src/index.js#L132

Alxblsk:
@calrk This is an example I've found in Internet, to be faster. In my own example, I do not use `aframe-click-drag-component` but still see the same issue. I'll attach another one demo soon.

ngokevin:
Fixed using debounced window.resize handler + canvas bounding rect. https://github.com/aframevr/aframe/pull/2942

PR #2942, opened 2017-08-08
Calculating mouse position based on canvas bounding rectangle for emb…

calrk:
…edded scenes

**Description:**
For this issue:
https://github.com/aframevr/aframe/issues/2938
Not sure about the performance impacts, and whether this code should only execute for if scenes have the 'embedded' attribute

**Changes proposed:**
- Using sceneEl.canvas.getBoundingClientRect() to properly calculate the mouse position relative to the canvas

### DIFF
--- a/examples/test/embedded-cursor/index.html
+++ b/examples/test/embedded-cursor/index.html
@@ -1,0 +1,165 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Cursor</title>
+    <meta name="description" content="Cursor - A-Frame">
+    <script src="../../../dist/aframe-master.js"></script>
+    <style type="text/css">
+      a-scene{
+        width: 50vw;
+        height: 50vh;
+        margin-top: 25vh;
+        margin-left: 25vw;
+      }
+    </style>
+  </head>
+  <body>
+    <a-scene embedded="true" cursor="rayOrigin: mouse">
+      <a-assets>
+        <audio id="blip1" src="../../showcase/anime-UI/audio/321103__nsstudios__blip1.wav"></audio>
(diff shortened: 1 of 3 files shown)


ngokevin:
This will be a bit more involved, but I think we want to cache the bounds (e.g., in `this.canvasBounds`, and then on window `resize` event in a throttled handler (`utils.throttle`), we want recalculate those bounds.

calrk:
Is this what you meant by using the throttledTick and window resize event together?

ngokevin:
Close, I meant just `utils.throttle` the handler, not `throttleTick`. In case someone is dragging to resize the window, many resize events would fire.

calrk:
I think this is what you mean, copying the style of 'throttledEmitComponentChanged' in component.js.
I made an example which I have also uploaded [here](http://games.clarklavery.com/aframe-sample/index.html) to test it out by mousing over the central cube. The problem with using the `throttle` wrapper is that if the user does a lot of small quick drags, it won't update to the final result. Doing a single resize works fine though.

The Library drag component library that was being used has the same issue. 
https://github.com/jesstelford/aframe-click-drag-component/blob/master/src/index.js#L132

ngokevin:
OK, I see. We want to use `debounce` instead. We need to add a utility function for that. In `utils/index.js` next to `throttle`, we can add:

```js
module.exports.debounce = function (func, wait, immediate) {
	var timeout;
	return function() {
		var context = this, args = arguments;
		var later = function() {
			timeout = null;
			if (!immediate) func.apply(context, args);
		};
		var callNow = immediate && !timeout;
		clearTimeout(timeout);
		timeout = setTimeout(later, wait);
		if (callNow) func.apply(context, args);
	};
};
```

And use `utils.debounce` instead. Thanks!

calrk:
Ok I've added the debounce function in and it works better now.

ngokevin:
Thanks!

PR #2943, opened 2017-08-08
optimize tracked controls tick, discovery, and various utils

ngokevin:
**Description:**

There was some messy/dirty/redundant code. Tested with Vive controllers.

**Changes proposed:**
- Limit the tracked-controls tick to every 500ms. Before it was every 10ms which does not help much since it usually takes more time to render a single frame. Pretty much, it was getting called on every frame. I don't think this polling needs to be called so often.
- Don't emit an event and allocate an object every single controller poll. Only emit if the number of controllers change. Before this was emitting every tick, and then possibly handled by 4 controller components at once, for each hand. Save 2 event emissions, 2 object allocations, and 8 event handler calls per tick.
- Remove `getGamepadsByPrefix` util. This was only being used in the tracked controls tick, and tracked controls never used the prefix feature. We call the `navigator.gamePads` now directly in the controls tick. Previously, we were creating a string and array on each tick, and then re-looping over that array and copying it to another array.
- Clean up axismove event abstraction, removing callbacks, and object/array creations.
- Don't need to call `updateControllersList` in the `isControllerPresent` util. This didn't really do anything since controllers were picked up through the `controllersupdate` event rather than within here.
- Various cleanups include removing unused properties (`axisLabels`), fixing 95-char lengths, and cleaner comments.

### DIFF
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -29,13 +29,9 @@ module.exports.Component = registerComponent('daydream-controls', {
   // 1 - menu (never dispatched on this layer)
   // 2 - system (never dispatched on this layer)
   mapping: {
-    axes: {'trackpad': [0, 1]},
+    axes: {trackpad: [0, 1]},
     buttons: ['trackpad', 'menu', 'system']
   },
-
-  // Use these labels for detail on axis events such as thumbstickmoved.
-  // e.g. for thumbstickmoved detail, the first axis returned is labeled x, and the second is labeled y.
-  axisLabels: ['x', 'y', 'z', 'w'],
 
   bindMethods: function () {
     this.onModelLoaded = bind(this.onModelLoaded, this);
@@ -59,8 +55,8 @@ module.exports.Component = registerComponent('daydream-controls', {
     this.everGotGamepadEvent = false;
     this.lastControllerCheck = 0;
     this.bindMethods();
-    this.checkControllerPresentAndSetup = checkControllerPresentAndSetup; // to allow mock
-    this.emitIfAxesChanged = emitIfAxesChanged; // to allow mock
+    this.checkControllerPresentAndSetup = checkControllerPresentAndSetup;  // To allow mock.
+    this.emitIfAxesChanged = emitIfAxesChanged;  // To allow mock.
   },
 
   addEventListeners: function () {
@@ -117,7 +113,12 @@ module.exports.Component = registerComponent('daydream-controls', {
   injectTrackedControls: function () {
     var el = this.el;
     var data = this.data;
-    el.setAttribute('tracked-controls', {idPrefix: GAMEPAD_ID_PREFIX, hand: data.hand, rotationOffset: data.rotationOffset, armModel: data.armModel});
+    el.setAttribute('tracked-controls', {
+      armModel: data.armModel,
+      hand: data.hand,
+      idPrefix: GAMEPAD_ID_PREFIX,
+      rotationOffset: data.rotationOffset
+    });
     if (!this.data.model) { return; }
     this.el.setAttribute('obj-model', {
       obj: DAYDREAM_CONTROLLER_MODEL_OBJ_URL,
@@ -137,8 +138,6 @@ module.exports.Component = registerComponent('daydream-controls', {
     if (!this.everGotGamepadEvent) { this.checkIfControllerPresent(); }
   },
 
-  // No need for onButtonChanged, since Daydream controller has no analog buttons.
-
   onModelLoaded: function (evt) {
     var controllerObject3D = evt.detail.model;
     var buttonMeshes;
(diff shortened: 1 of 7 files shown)


src/components/gearvr-controls.js
```diff
@@ -51,17 +47,25 @@ module.exports.Component = registerComponent('gearvr-controls', {
     var self = this;
     this.animationActive = 'pointing';
     this.onButtonChanged = bind(this.onButtonChanged, this);
-    this.onButtonDown = function (evt) { self.onButtonEvent(evt.detail.id, 'down'); };
-    this.onButtonUp = function (evt) { self.onButtonEvent(evt.detail.id, 'up'); };
-    this.onButtonTouchStart = function (evt) { self.onButtonEvent(evt.detail.id, 'touchstart'); };
-    this.onButtonTouchEnd = function (evt) { self.onButtonEvent(evt.detail.id, 'touchend'); };
+    this.onButtonDown = function (evt) {
```
dmarcos: Can we keep the one liners. It makes the code harder to parse?

ngokevin: Some were going over a character limit causing them to wrap. Want me to revert still?

dmarcos: I think it's ok going over the character limit in same cases to avoid contortions and breaking style consistency

ngokevin: Changed it

src/components/daydream-controls.js
```diff
@@ -50,17 +46,25 @@ module.exports.Component = registerComponent('daydream-controls', {
...
-    this.onButtonUp = function (evt) { self.onButtonEvent(evt.detail.id, 'up'); };
-    this.onButtonTouchStart = function (evt) { self.onButtonEvent(evt.detail.id, 'touchstart'); };
-    this.onButtonTouchEnd = function (evt) { self.onButtonEvent(evt.detail.id, 'touchend'); };
+    this.onButtonDown = function (evt) {
+      self.onButtonEvent(evt.detail.id, 'down');
+    };
+    this.onButtonUp = function (evt) {
+      self.onButtonEvent(evt.detail.id, 'up');
+    };
+    this.onButtonTouchStart = function (evt) {
```
dmarcos: one liner

src/components/daydream-controls.js
```diff
@@ -50,17 +46,25 @@ module.exports.Component = registerComponent('daydream-controls', {
...
+      self.onButtonEvent(evt.detail.id, 'touchstart');
```
dmarcos: one liner

src/components/daydream-controls.js
```diff
@@ -50,17 +46,25 @@ module.exports.Component = registerComponent('daydream-controls', {
...
+    };
+    this.onButtonTouchEnd = function (evt) {
```
dmarcos: one liner

src/components/daydream-controls.js
```diff
@@ -137,21 +146,24 @@ module.exports.Component = registerComponent('daydream-controls', {
...
   onModelLoaded: function (evt) {
     var controllerObject3D = evt.detail.model;
     var buttonMeshes;
     if (!this.data.model) { return; }
     buttonMeshes = this.buttonMeshes = {};
-    buttonMeshes.menu = controllerObject3D.getObjectByName('AppButton_AppButton_Cylinder.004');
-    buttonMeshes.system = controllerObject3D.getObjectByName('HomeButton_HomeButton_Cylinder.005');
-    buttonMeshes.trackpad = controllerObject3D.getObjectByName('TouchPad_TouchPad_Cylinder.003');
-    // Offset pivot point
+    buttonMeshes.menu = controllerObject3D.getObjectByName(
```
dmarcos: If needed I think it looks better to break the whole expression to the line below instead of just an argument

src/components/daydream-controls.js
ngokevin: yeah i can catch them all if you just note once

ngokevin: though i missed this file xD

src/components/daydream-controls.js
ngokevin: i'll just restore to one liners

fernandojsg:
I believe it could be nice to have the tick ms as a parameter, for some applications 500ms it could be way too much

ngokevin:
It just detects new controllers right? Half a second isn't much

fernandojsg:
@ngokevin my fault, I was thinking that the tick was in updating the position/rotation, not just on updating the list, I've checked the coded already O:)

ngokevin:
Updated to save another couple object allocations and to skip type check per tracked controls per tick.

machenmusik:
Has this been checked to ensure that controller pose update rate is unaffected in-VR?  I know things have changed a lot since we started these

machenmusik:
Ok so I think the assumption is that the browser continues to make pose changes to the same controller object, which is still updated unthrottled by tracked-controls component tick, so unless some browser has done wacky things in its implementation, LGTM.  Only quirk is if controller goes away, and we won't know it for up to 500ms; need to ensure that dying controller object state doesn't break things before detection.

ngokevin:
Yeah, it'd be pretty noticeable if controllers only moved once every 500ms. The tracked-controls component handles the pose update. The system only handles detection for connected/disconnected so it's fine.

I'll double-check what happens on disconnect.

PR #2945, opened 2017-08-08
optimize wasd-controls (add early return, skip type checking)

ngokevin:
**Description:**

Shouldn't try to run any WASD controls code if we're not using WASD (e.g., in real VR).

**Changes proposed:**
- Add early return to `wasd-controls` tick to not run `updateVelocity` if no keys have been pressed and no velocity to decay. Keep better track of what relevant keys have been pressed to be able to detect this.
- Re-use object for updating `position`.

### DIFF
--- a/src/components/wasd-controls.js
+++ b/src/components/wasd-controls.js
@@ -8,6 +8,10 @@ var shouldCaptureKeyEvent = utils.shouldCaptureKeyEvent;
 
 var CLAMP_VELOCITY = 0.00001;
 var MAX_DELTA = 0.2;
+var KEYS = [
+  'KeyW', 'KeyA', 'KeyS', 'KeyD',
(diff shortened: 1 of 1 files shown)


src/components/wasd-controls.js
```diff
@@ -42,27 +47,29 @@ module.exports.Component = registerComponent('wasd-controls', {
...
 
-    // Get velocity.
+    // Update velocity.
+    delta = delta / 1000;
     this.updateVelocity(delta);
+
     if (!velocity[data.adAxis] && !velocity[data.wsAxis]) { return; }
 
     // Get movement vector and translate position.
+    currentPosition = el.object3D.position;
```
dmarcos: I would keep using `getAttribute` to keep using the public API instead of coupling with the internals. I don't think there's any difference.

PR #2950, opened 2017-08-08
optimize component change/initialize events

ngokevin:
**Description:**

Don't pass data/oldData into the component events, only the component signature (a one time allocation). Leave it up to the application to fetch the old/current data if needed.

**Changes proposed:**
- Save an object allocation every `componentchanged` (i.e., every component instance, every 200ms).
- Save two object allocations every `componentinitialized` (including the `getData` clone).

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -35,19 +35,17 @@ var Component = module.exports.Component = function (el, attrValue, id) {
   this.el = el;
   this.id = id;
   this.attrName = this.name + (id ? '__' + id : '');
+  this.evtDetail = {id: this.id, name: this.name};
   this.initialized = false;
   this.el.components[this.attrName] = this;
+
   // Store component data from previous update call.
   this.oldData = undefined;
+
   // Last value passed to updateProperties.
   this.previousAttrValue = undefined;
-  this.throttledEmitComponentChanged = utils.throttle(function emitComponentChange (oldData) {
-    el.emit('componentchanged', {
-      id: self.id,
-      name: self.name,
-      newData: self.data,
-      oldData: oldData
-    }, false);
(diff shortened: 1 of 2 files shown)


src/core/component.js
```diff
@@ -41,13 +41,9 @@ var Component = module.exports.Component = function (el, attrValue, id) {
...
   // Last value passed to updateProperties.
   this.previousAttrValue = undefined;
-  this.throttledEmitComponentChanged = utils.throttle(function emitComponentChange (oldData) {
-    el.emit('componentchanged', {
-      id: self.id,
-      name: self.name,
-      newData: self.data,
-      oldData: oldData
-    }, false);
+  this.signature = {id: this.id, name: this.name};
```
dmarcos: Is `evtDetail` a better name?

ngokevin: ok

src/core/component.js
```diff
@@ -35,19 +35,17 @@ var Component = module.exports.Component = function (el, attrValue, id) {
...
   this.previousAttrValue = undefined;
-  this.throttledEmitComponentChanged = utils.throttle(function emitComponentChange (oldData) {
-    el.emit('componentchanged', {
-      id: self.id,
-      name: self.name,
-      newData: self.data,
-      oldData: oldData
-    }, false);
+  this.throttledEmitComponentChanged = utils.throttle(function emitChange () {
+    el.emit('componentchanged', self.evtDetail, false);
```
dmarcos: we are not passing `oldData` anymore as detail?

dmarcos: the same for newData?

ngokevin: I think applications can fetch that themselves if needed? I suppose we could update the object without harm.

PR #2951, opened 2017-08-09
object pooling in component core, optimize style parser to reuse object

ngokevin:
**Description:**

Rather than create a new object each time to create a new `oldData`, just have two objects representing `oldData` and `data`. Also when updating, rather than create a new `data` object, have a `newData` object that is used for holding the new data which then gets copied into `data`. So in total, only use three objects: `oldData`, `data`, `newData`.

**Changes proposed:**
- Adapt `style-attr` implementation to be able to pass in an object for reuse instead of creating a new one each time. Also remove lots of callbacks, `Object.keys()` array creations from their code.  Also tweaked to do prettier stringifies. We manage this forked dependency now. 
- Use object pooling in core component code to reduce object allocations. Recycle them when component is removed.
  - 1. Remove object allocation on each update for creating object for `oldData`.
  - 2. Remove object allocation on initialization for initial `oldData`.
  - 3. Remove object allocation on each update during `buildData` for default value copying.  
- Add some helper properties to component (`isSingleProp`, `isSinglePropObject`, `isObjectBased`)
- Update our `styleParser` utils a bit to not allocate object + array + use callbacks to convert keys to camelCase.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -16,6 +16,7 @@ var warn = utils.debug('core:component:warn');
 
 var aframeScript = document.currentScript;
 var upperCaseRegExp = new RegExp('[A-Z]+');
+var objectPool = utils.objectPool.createPool();
 
 /**
  * Component class definition.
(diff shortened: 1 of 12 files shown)


src/core/component.js
```diff
@@ -138,7 +146,7 @@ Component.prototype = {
   stringify: function (data) {
     var schema = this.schema;
     if (typeof data === 'string') { return data; }
-    if (isSingleProp(schema)) { return stringifyProperty(data, schema); }
+    if (this.isSingleProp) { return stringifyProperty(data, schema); }
```
dmarcos: Full name `isSinglePropertySchema`?

ngokevin:
Spent a day to get this updated with passing tests. Optimizing the component core, string parsing, and adding in object pooling practices.

ngokevin:
@dmarcos r?

ngokevin:
Updated to not delete the keys when recycling an object, but just set every key to `undefined`. This means object pooling best used when objects share a common schema, in case we iterate over keys. So I create different object pools for each type of component, since their object keys will be the same. I also update schema parser to ignore undefined keys.

## Before

screen shot 2017-10-03 at 7 16 43 pm

## After

**Improvements:** See the `updateProperties` and `updateCachedAttrValue` no longer on the list of frequent memory allocators.

screen shot 2017-10-03 at 7 17 57 pm

ngokevin:
Updated. Still need to polish/comment a bit more (and try to remove the clones for default values). But up and working again.

dmarcos:
Tests fail

ngokevin:
Moved to https://github.com/aframevr/aframe/pull/3772

PR #2954, opened 2017-08-10
5x faster utils.bind function

ngokevin:
**Description:**

https://jsperf.com/bind1-vs-bind2

We only care about binding the `this`/context. The additional argument binding we don't need, else just use vanilla bind for that.

**Changes proposed:**
- Simpler and fast bind function.

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -1,4 +1,3 @@
-var bind = require('../utils/bind');
 var diff = require('../utils').diff;
 var debug = require('../utils/debug');
 var registerComponent = require('../core/component').registerComponent;
@@ -90,7 +89,7 @@ module.exports.Component = registerComponent('light', {
               if (value.hasLoaded) {
                 self.onSetTarget(value, light);
               } else {
-                value.addEventListener('loaded', bind(self.onSetTarget, self, value, light));
+                value.addEventListener('loaded', self.onSetTarget.bind(self, value, light));
               }
             }
             break;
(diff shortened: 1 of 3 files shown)


dmarcos:
What the `jsperf` seems to test is the binding a function no the execution cost of the bound function right?

dmarcos:
Why not using the regular `Function.prototype.bind`?

src/components/light.js
```diff
@@ -90,7 +89,7 @@ module.exports.Component = registerComponent('light', {
               if (value.hasLoaded) {
                 self.onSetTarget(value, light);
               } else {
-                value.addEventListener('loaded', bind(self.onSetTarget, self, value, light));
+                value.addEventListener('loaded', self.onSetTarget.bind(self, value, light));
```
dmarcos: you are using the `Function.prototype.bind` not the one in utils

dmarcos: Is it intentional?

src/utils/bind.js
```diff
@@ -1,15 +1,11 @@
...
  */
-module.exports = function bind (fn, ctx/* , arg1, arg2 */) {
-  return (function (prependedArgs) {
-    return function bound () {
-      // Concat the bound function arguments with those passed to original bind
-      var args = prependedArgs.concat(Array.prototype.slice.call(arguments, 0));
-      return fn.apply(ctx, args);
-    };
-  })(Array.prototype.slice.call(arguments, 2));
+module.exports = function bind (fn, ctx) {
```
dmarcos: I would try to use the regular `Function.prototype.bind`. Not sure that this has any benefit.

ngokevin:
I'll re-profile to see if it's still an issue. Previously, a lot of our initialization time was spent on binding functions (not calling them). On Chrome, vanilla bind is the fastest. On Firefox, this PR bind is the fastest. I updated the JS perf to include vanilla.

dmarcos:
IIRC the perf issue we saw was when calling the bound function rather than initializing

ngokevin:
Will close to probably deprecate `utils.bind` then and use vanilla.

Issue #2961, opened 2017-08-11
Bug noticed in 0.6.0 while completing "Building a 360° Image Gallery" guide.


rttomlinson:
**Description:**
I noticed a bug while going through the "Building a 360° Image Gallery" guide.
As of 0.6.0, if you mousedown outside of a box, drag over the box, drag outside of the box, and then mouseup outside of the box. The box will respond as expected per the "event_set" components. That is, scale up to "1.2 1.2 1" on mousenter and scale back down to "1 1 1" mouseleave. However, if the mousedown event occurs while already hovered over the box, the box will scale down to "1 1 1" as expected, but when the cursor is dragged outside the box and mouseup occurs, the box will scale back up to "1.2 1.2 1". We expect the box to stay at "1 1 1". This does not appear to be an issue in v. 0.5.0.

My next step is to see if the issue is related to the event_set component and 0.6.0 or with 0.6.0 itself.

- A-Frame Version: 0.6.0
- Platform / Device: Glitch, but only Cloud9(Linux)
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/married-glade
Upgrading from <0.6.0 to 0.6.0

ngokevin:
updted / fixed

PR #2962, opened 2017-08-11
Information about bug for users using 0.6.0

rttomlinson:
Related to Issue #2961

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/guides/building-a-360-image-gallery.md
+++ b/docs/guides/building-a-360-image-gallery.md
@@ -11,6 +11,8 @@ order: 3
 > Gallery Boilerplate on
 > GitHub](https://github.com/aframevr/360-image-gallery-boilerplate).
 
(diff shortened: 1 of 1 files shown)


dmarcos:
I would not hardcode version numbers in the examples or reference specific bugs. The info will go quickly out of date and it will be hard to remember to update it

rttomlinson:
What would you recommend doing? It was frustrating not understanding why the example was not working properly.

dmarcos:
A bug in an example is the same as any other bug in the code base. The fix will ship in the next A-Frame version in a few weeks. You opened a good issue already to keep track of it. Any ideas on how we could fix it https://github.com/aframevr/aframe/issues/2961?

rttomlinson:
Not yet. I will look into whether the issue is with the `event_set` component or if it is with something else within the code base first.

PR #2963, opened 2017-08-11
Additional information for clarification

rttomlinson:
Information regarding plane and camera default values

**Description:**
More info for guide
**Changes proposed:**
- 
-
-

### DIFF
--- a/docs/guides/building-a-360-image-gallery.md
+++ b/docs/guides/building-a-360-image-gallery.md
@@ -116,6 +116,9 @@ of the images defined in the [Asset Management System][ams].
   geometry="primitive: plane; height: 1; width: 1"
   material="shader: flat; src: #cubes-thumb"></a-entity>
 ```
+>***Note that the planes will default to position="0 0 0" and since the camera is also in that position by default you will not see them. You can reposition the camera with "wasd". Move the camera in the positive z direction ("S" key if camera is facing the default direction) to see the planes.
+
+>***Note that the planes do not have "backs" by default. If you move to the opposite facing side of the plane, the plane will look like it disappears!
 
 [sound]: ../components/sound.md
 


rttomlinson:
I agree that this is not specific to 360 image example, but it does provide a little more information which would have been helpful when going through the guide. I found myself on the Slack channel in order to get over these hurdles.

ngokevin:
> since the camera is also in that position by default

The camera default position should be 1.6 meters up, no? If you use the default camera or `<a-camera>`.

rttomlinson:
Oh, you're right. The docs say it's "0 1.6 0" in desktop mode and "0 0 0" in VR mode. I'm in desktop mode.
https://github.com/aframevr/aframe/blob/master/docs/primitives/a-camera.md

docs/guides/building-a-360-image-gallery.md
```diff
@@ -116,6 +116,9 @@ of the images defined in the [Asset Management System][ams].
   geometry="primitive: plane; height: 1; width: 1"
   material="shader: flat; src: #cubes-thumb"></a-entity>
 ```
+>***Note that the planes will default to position="0 0 0" and since the camera is also in that position by default you will not see them. You can reposition the camera with "wasd". Move the camera in the positive z direction ("S" key if camera is facing the default direction) to see the planes.
```
rttomlinson: camera is by default in "0 1.6 0" in desktop mode and "0 0 0" in VR mode. https://github.com/aframevr/aframe/blob/master/docs/primitives/a-camera.md

ngokevin:
Yeah, and in VR mode (with positional tracking), it would be at the real-world head height, so greater than 0. Anything we should note given that?

rttomlinson:
What if the comment says something more generic like "You will need to reposition the camera in order to see the planes because of the default settings. You can reposition the camera with "WASD".

ngokevin:
It's not really anything specific with planes, but just placing stuff at the same position as the camera in general.

Perhaps a quick comment at https://aframe.io/docs/0.6.0/guides/building-a-basic-scene.html#placing-our-box-in-front-of-the-camera like "In general, note that if we add an entity at the same position as the camera, we won't see it from the default viewpoint". But we can leave that for another PR, thanks!

rttomlinson:
I'll look at the first guide again.

Are we considering the guides to be sequential or are each stand alone?

PR #2964, opened 2017-08-11
Add ability to override default mobile settings (hmd rotation, fixed pitch).

jaydansand:
**Description:**
It is currently impossible to provide a full non-HMD experience on mobile devices. The look-controls component bypasses `hmdEnabled` checks for mobile devices, and prevents touch-based pitch adjustment.

In some contexts, mobile users may want to choose between using their phone as an hmd or using pure touch-based interaction.

**Changes proposed:**
- Add new `mobileHmdAlways` configuration option that toggles the `isMobile` bypass.
- Add new `mobileFixedPitch` configuration option that toggles the `onTouchMove` pitch adjustment.
- Default both of the new options to preserve current behavior for backwards compatibility. Developers can use these options to opt in to the new behavior.


ngokevin:
Normally, we recommend just copying/customizing the look-controls if people need special behaviors. There are many permutations and edge cases of controls that people have come up with, but supporting all of the at once is pretty hard. Since now we have so many switches, it becomes harder to maintain.

Perhaps we might one day get something like Don's universal-controls in aframe-extras, but A-Frame is sort of VR-focused so doesn't cater as much to the many possible mobile use cases.

jaydansand:
Thanks for the quick response! I totally understand and that definitely makes sense for most customizations. I'd argue my pull request isn't a customization, but more a compromise of a bug fix.

I don't really want to change anything about how look-controls behaves, essentially, except I want mobile and desktop experiences to be consistent (in other words I don't want mobile devices to get special treatment). Right now look-controls carves out exceptions to its own behaviors (touchmove vs mousemove controlling pitch, `hmdEnabled=false` is ignored), *just for mobile devices*. Shouldn't that special mobile vs. desktop behavior be optional?

ngokevin:
The look-controls is meant to be a default implementation where A-Frame is VR-focused, meaning mobile uses HMD. I wouldn't try to add more and more things on top of look-controls to accommodate every use case. For cases such as this, it's not too difficult to copy/modify look-controls. We get a lot of PRs adding these types of switches to look-controls, I will try to add a FAQ/documentation on how to modify/customize look-controls in the future.

Thanks, let me know if you have other concerns.

ngokevin:
Well, will rethink this since it's a harmless switch. @dmarcos, opinions on having look-controls support full non-VR for mobile? I'd lean on not.

dmarcos:
I'm ok with this switch in isolation. The broader question is why are we taking this when we rejected similar PRs in the past? My worry is not this feature in particular, it's reasonable and the code is well written. My concern as we discussed is piling tons of small switches that grow the number of posible configurations and code paths we have to maintain. I feel that it's already hard enough to provide good support for the current broad range of devices for the VR use case. As a general approach I would wait to see more demand before taking up on the long term maintainance cost. I want to feel confident we can deliver good quality without going insane 🙂

ngokevin:
Mostly a balance between maintenance vs. value. This one doesn't cause for much maintenance, but possibly some value. I could see how people might want to disable the polyfill but keep the touch controls on mobile, or want to provide some UI to let user decide interaction. Hard to measure demand on these types of things, but the WebVR community does love its "responsiveness". Thin line.

machenmusik:
It may be a question of providing access to internals so it is simple to make an external component that does look-controls except for some small thing... asking devs to recreate entirely just to throw a flag (e.g. laser-controls) is another example with a different response.  There, the maintenance is basically zero, the utility is potentially significant, and the code is written in such a way as to be impractical to work around (e.g., as we support more controllers and laser-controls is updated, variants have to update in lockstep as there is no reuse / inheritance possible)

ngokevin:
The maintenance isn't zero. We have arrays of flags and need to work together, and we're continually adding more flags or getting PRs for more flags for non-VR purposes. It's less resistance to copy-and-paste (remix, not *recreate*) and tweak. Components should be small, simple, tweakable rather than large complexes.

`laser-controls` workflow and customization doesn't have any issues with it. It's a very small and simple component that sets other components. Need to add more controllers? Just attach them. That other PR about the `model` flag is an unrelated issue with why the line does not show up when the `model` is set to `false`, not to do with needing any flags.

jaydansand:
Thanks so much for taking the time to debate this issue! I don't know much about the other PRs that have been proposed, but here are my arguments in favor of this PR. I had mentioned (1) before, but the conversation seems to have moved on and I wanted to bring it up again because it's relevant to (2).

1. This could really be considered a bug report instead of a feature request: the switches added in the PR are a backwards-compatible way of making things consistent. Phones are not handled consistently compared with all other devices, which feels like a bug. For example, `hmdEnabled` is ignored, only for phones. Pitch control is locked, only for phones.
2. @dmarcos's concern about opening the floodgates is totally valid, but here's one critical difference: as described in (1) above, this isn't really adding new features or behaviors to `look-controls`; it's restoring current functionality across all devices. I don't think there are many other switches anyone could propose that fall within the same narrow scope; there aren't (to my admittedly limited understanding) a bunch of other places where `look-controls` carves out special exceptions just for some class of devices. That could be an easy test to limit the kinds of PRs that would be considered in the future.
3. As @ngokevin mentioned, web developers love their responsiveness, and there are definitely use cases out there (and cases like mine will probably grow, especially if this PR goes through). Making this small change will greatly expand aframe's applicability for web devs (allowing full-featured mixed-device operation), increasing interest and gaining contributors who are eager to help. I don't know if that's a good thing, but probably it is? Hooray for OSS!

On a related note, +1 for @machenmusik's suggestion about exposing more hooks into `look-controls`. It's currently fairly prohibitive to clone a complex component like `look-controls` and be forever locked into merging upstream for the 99% of unchanged code. :)

And lastly, thank you guys so much for your wonderful work that makes this all possible. It's truly appreciated, no matter how this PR goes.

ngokevin:
Would it make sense to just disable the mobile orientation-based movement when hmdEnabled is false?

jaydansand:
That would solve half of the PR, but what about fixed pitch?

ngokevin:
I wouldn't handle fixed pitch, at that point, you can customize manually the look-controls for you own needs.

jaydansand:
Is there a reason for defaulting to fixed pitch? It seems weird that mobile devices are gimped out of the gate, especially with the preponderance of Google Cardboard and related "VR headsets" that people use with phones - I added flags to be backwards-compatible, but it's arguably a bug and this is a bugfix. Would it help if I reclassify?

Copying look controls means buying into constantly merging and updating. There's a reasonable standard that could prevent this being a floodgate issue (see [August 28, 2017 comment](https://github.com/aframevr/aframe/pull/2964/#issuecomment-325404273), item #2) in this particular (bug-fixy) case.

donmccurdy:
> Copying look controls means buying into constantly merging and updating.

I don't think it's much trouble in practice. I've been maintaining an alternate version [here](https://github.com/donmccurdy/aframe-extras/tree/master/src/controls) and haven't had to touch the HMD/orientation code in 11 months.

jaydansand:
@dmarcos That probably depends on the usage - for example, all of our scenes have spherical image skyboxes with scenery, so you're not likely to forget which way is down (it's where the grass is),
 especially since it doesn't let you go 'round the horn and flip upside down. Anecdotally, we've hacked the fixed pitch out, and hundreds of users have tried it without being confused or disoriented (yet, anyway). Shouldn't the fixed pitch behavior be opt-in (or at least provide an opt-out)?

Thought from another angle, why isn't it considered as disorienting to pan the scene using mouse on a desktop (i.e. why doesn't desktop + mouse have fixed pitch)? The gestures are very similar: up is up, down is down, and you can't go past the zenith in either.

dmarcos:
@jaydansand the way to opt out is to implement your own component as you succesfuly did.

dmarcos:
@jaydandsand we want to keep the core components focused and with functionality that serves the majority of cases. By keeping components simple we prevent a combinatory explosion of flags permutations that is very hard to maintain over time. Simplicity also allows developers like you to adapt the code to particular scenarios we don’t cover. I’m going to close this for now and revisit in the future if there’s more demand

jaydansand:
As you wish; I'd think a simple opt-out for a workaround would make it more suitable to the majority of cases, but it's not my project.

I'm unclear: will `hmdEnabled:false` at least be respected on mobile devices going forward, as @ngokevin suggested in https://github.com/aframevr/aframe/pull/2964#issuecomment-333301342? If that requires an update to the PR, I'm happy to do it - half-acceptance is better than zero-acceptance.

donmccurdy:
Having `hmdEnabled=false` also apply to mobile devices seems reasonable to me; +1 on that.

PR #2966, opened 2017-08-12
fix for vr2vr in Oculus Browser

machenmusik:
With Oculus Browser, the vrdisplayactivate event fires before this.effect exists,
so try again next tick also... harmless if first time succeeded.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -123,7 +123,17 @@ module.exports.AScene = registerElement('a-scene', {
         this.exitVRTrueBound = function () { self.exitVR(true); };
 
         // Enter VR on `vrdisplayactivate` (e.g. putting on Rift headset).
-        window.addEventListener('vrdisplayactivate', this.enterVRBound);
+        window.addEventListener('vrdisplayactivate', function () {
+          // With Oculus Browser, vrdisplayactivate fires before this.effect exists.
+          if (!self.effect) {
+            // Try again to enter VR next tick.
+            // NOTE: Using self.enterVRBound apparently no longer counts as user gesture.
+            setTimeout(function () { self.enterVR(); });
+          } else {
+            // Enter VR now.
+            self.enterVR();
+          }
+        });
 
         // Exit VR on `vrdisplaydeactivate` (e.g. taking off Rift headset).
         window.addEventListener('vrdisplaydeactivate', this.exitVRBound);


src/core/scene/a-scene.js
```diff
@@ -118,7 +118,12 @@ module.exports.AScene = registerElement('a-scene', {
         window.addEventListener('vrdisplaypresentchange', this.onVRPresentChangeBound);
 
         // Enter VR on `vrdisplayactivate` (e.g. putting on Rift headset).
-        window.addEventListener('vrdisplayactivate', function () { self.enterVR(); });
+        window.addEventListener('vrdisplayactivate', function () {
```
ngokevin: What about having enterVR check if there is no `this.effect`, and registering an event listener to call itself when its ready.

machenmusik: Checking `this.effect` may work, but not sure whether subsequent event listener will work, as `vrdisplayactivate` initiation context would be lost; so far, firing next tick seems to be long enough for the renderer startup to happen, as that seems long enough for `render-target-loaded` handling to propagate.

machenmusik: This still seems to work.  Note that there is a quirk with VR-to-VR back navigation, but I think that is on the Oculus Browser side.
`works for me, although once you use back button to go back within VR, the next attempt to go forward again while staying in VR seems to fail, you get kicked back to 2D browser mode.  But if you then reload and re-enter VR, next in-VR traversal succeeds.`

dmarcos:
I suspect that the solution for this is going to be a bit more complicated. First, We have to determine when `vrdisplayactivate` is exactly supposed to be fired when navigating. The app logic has to have a consistent way to attach an event handler without missing the event. At the moment, we attach the handler on the `attachedCallback` method of `a-scene`. Is that early enough? I know that we also have a time limit to call `requestPresent` in the `vrdisplayactivate` handler after the event fires. VREffect is not created until the `canvas` component initializes and that can take a while because it's blocked by asset loading. We might have to initialize the canvas and VREffect as early as we can so we can call `requestPresent` even before the assets load.

machenmusik:
In case it wasn't clear, `5dc9f75` has been tested to work successfully on my GearVR phone, with the one quirk with back behavior I mentioned.

> At the moment, we attach the handler on the attachedCallback method of a-scene. Is that early enough?
Yes, it does appear to be early enough to attach.  The issue I saw was that Oculus Browser was firing the event itself before A-Frame was ready to actually execute the handler... given that a timeout of zero (next tick) succeeds, there may be some next-tick deferral that making `this.effect` available depends upon which is unique to A-Frame.

> We might have to initialize the canvas and VREffect as early as we can so we can call requestPresent before the assets load.
IMO we should do that so that one can present VR UX during asset load, even if you ignore VR-to-VR navigation entirely -- but I don't think we should wait for all that to be completed before making A-Frame VR-to-VR navigation work with Oculus Browser.

machenmusik:
something changed overnight, started getting conflicts and now the change doesn't work anymore.  investigating.

machenmusik:
OK, this latest version is tested to work using its integration into chenzlabs master - and this time I did not run into the back-behavior quirk.

machenmusik:
@ngokevin why am I getting this test failure? seems that a recent commit may have poisoned the well...
```
[error]: docs/introduction/index.md
    Link definition does not exist: [Mozilla][mozilla]
```

machenmusik:
@ngokevin @dmarcos if you have GearVR can you confirm it works for you?  any other issues preventing merge?

machenmusik:
I agree that this change does not address the fundamental issue completely, but it does buy us some time while we investigate further.  It is not entirely clear to me that splitting into an initial and then subsequent attempt to call requestPresent will not run into gesture permission issues the second time, as did some of the initial attempts at workaround.

dmarcos:
The fix won't ship until 0.7.0. I prefer to get a proper solution rather than something temporary we have to redo. Once you have aquired a vr session via requestPresent after `vrdisplayactivate` subsequent calls will also succeed. That's part of the spec

dmarcos:
quoting the spec: `If requestPresent() is called outside of an engagement gesture, the promise MUST be rejected unless the VRDisplay was already presenting.`

machenmusik:
Fair enough.  Is `vrdisplayactivate` the only event we need to treat in this fashion?

machenmusik:
@dmarcos at present, trying your recommended approach is not successful with Oculus Browser...

- Installing a `vrdisplayactivate` extremely early (at the bottom of src/index.js) causes problems in `requestPresent` since we don't have the canvas yet, and thus we don't have any layers, so `requestPresent` fails.  It is possible to remember the display, but if we can't make `requestPresent` succeed, we are still subject to gesture requirements later on.

- Cardboard polyfill seems to be triggering a doomed attempt to enterVR() when it does display connection (`vrdisplayconnect`).

- Inside the `render-target-loaded` handler after `this.resize()`, the subsequent attempt to `enterVR()` fails.
I wonder if we have to wait for next tick to succeed in any event due to deferred actions from setting up renderer/effect... which the prior workaround had been doing.

dmarcos:
You can use a dummy canvas different than the one that will be later created.

machenmusik:
(simply adding a dummy canvas to document body and passing that in doesn't work either.)

dmarcos:
What is not working? Firefox, Oculus Browser... It's probably a bug in the browser.

dmarcos:
We can create the canvas and then the canvas component will reuse it.

machenmusik:
Oculus Browser.  I am guessing there are some minimal requirements on the canvas that aren't satisfied.  But given that the current PR works, and no definite working implementation strategy for more optimal solution after a few attempts, I will be using the current PR in a fork for the time being.

machenmusik:
What is currently happening seems to be the following.

With the early handler attach and dummy canvas (assuming that a webgl context has been created),
the initial `vrdisplayactivate` is caught and the initial `requestPresent` with dummy canvas succeeds...

But then it looks like polyfill decides to connectDisplay to CardboardVRDisplay, which fires `vrdisplayconnect`, and the scene handler for that then triggers `enterVR()` before `this.effect` exists.

When that fails, there doesn't seem to be a subsequent attempt to access `requestPresent` (e.g. from the `render-target-loaded` handler upon noticing the artifact from the initial `vrdisplayactivate`) ... and so we end up never presenting the correct page content in VR (or at all).

dmarcos:
It seems a bug on the polyfill. it's should be a no op in an Oculus Browser context that has a native WebVR implementation

machenmusik:
Is this yet another missing DPDB problem?  I notice that the DPDB does not have my phone model, nor does it have sensible wildcards for the various Samsung S6/S7/S8 variants...

machenmusik:
Maybe #2985 will improve this...

machenmusik:
Unfortunately #2985 does not improve this, current PR is still needed

machenmusik:
Deprecated by #2991

PR #2968, opened 2017-08-12
Bump THREE to r86

dmarcos:
Bump THREE to `r86`
- This will unblock `glTF2` support (cc @donmccurdy)
- @mrdoob recommended to wait until `r87` to move from `VREffect` / `VRControls` to the new `renderer.vr` API

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "present": "0.0.6",
     "promise-polyfill": "^3.1.0",
     "style-attr": "^1.0.2",
-    "three": "^0.84.0",
+    "three": "^0.86.0",
     "three-bmfont-text": "^2.1.0",
     "webvr-polyfill": "^0.9.36"
   },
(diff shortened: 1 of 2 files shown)


donmccurdy:
This is working correctly for me, testing out the tracked-controls demo, but i do see a `TypeError: device is null` error in the console. Looks the same as whatever is making the tests fail.

dmarcos:
There is a bug in THREE where a `vrdisplaypresentchange` fires and there' no VRDevice assigned to the `WebVRManager`. I opened https://github.com/mrdoob/three.js/pull/11970 to consider the case.

dmarcos:
As a temporary workaround until `THREE` r87 ships I'm setting a dummy VRDevice on the THREE VRManager to prevent the `TypeError: device is null` error.

dmarcos:
Let `glTF 2.0` take over the world! :rocket:

PR #2971, opened 2017-08-14
add look-controls new attribute and default "tiltEnabled: true" so th…

c4augustus:
…at mobile tilt can be disabled

**Description:**
This adds to the component look-controls schema: "tiltEnabled: false/true (default)"

**Changes proposed:**
This allows disabling the current default behavior that tilting a mobile device changes the orientation of the camera--while not in VR mode--i.e. "look around with your phone" mode. The current behavior (tilting) remains the default.
-
-

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -17,7 +17,8 @@ module.exports.Component = registerComponent('look-controls', {
     enabled: {default: true},
     hmdEnabled: {default: true},
     reverseMouseDrag: {default: false},
-    standing: {default: true}
+    standing: {default: true},
+    tiltEnabled: {default: true}
   },
 
   init: function () {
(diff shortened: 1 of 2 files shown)


dmarcos:
There's already a flag called `hmdEnabled`. Can we reuse it?

c4augustus:
I considered that but then if you want to disable tilt only for non-VR mobile mode and then the user enters VR-mode, the code would have to react and reset the hmdEnable flag for the HMD to work properly. I'm assuming this, I didn't actually test to see if that would be the case.

dmarcos:
I really appreciate the PR. The code looks great and it even comes with tests 🥇 Our responsive approach considers magic window mode as the base line for mobile devices. A-Frame provides a set of sensible defaults out of the box that we think are useful for a majority of use cases. For custom functionality we recommend creating components using the core ones as a base as you did. Incorporating flags for a lot of particular cases will force us to test and maintain a lot of permutations making very difficult to maintain quality as we move forward. I'm going to close this PR but I'll keep it in mind if we see more demand for this feature.

ngokevin:
I'll try to write an FAQ and something in the docs about customizing look-controls for non-VR purposes. We've gotten many non-prompted PRs on `look-controls` in the past, and we usually repeat the same things. It's unfortunate to have to turn away work. But thanks though!

c4augustus:
Okay then. I'll just proceed with your recommendation of writing a customized/different look-controls component. I will probably want to do more with it anyway. My goal is to help build websites that can transition somewhat seamlessly from manually navigated 3D (the most likely usecase) through "tilted"/"look around" mobile 3D (less likely), and then finally VR (the coolest, but most niche at this time). I understand your desire to keep A-Frame focused on VR, and possibly AR as well.

PR #2972, opened 2017-08-14
In Python 3 SimpleHTTPServer is called http.server

mvasilkov:
**Description:**

In Windows (arguably a popular beginner environment), Python 3 installs the `python` binary, same as Python 2. The `python -m SimpleHTTPServer` command mentioned in the docs will not work in Python 3, which can be mildly confusing to people unfamiliar with the change.

**Changes proposed:**

- Add the alternative command for Py3.


dmarcos:
Thank you! 🥇

PR #2973, opened 2017-08-14
Enable touch input for cursor click

wmurphyrd:
**Description:**

I recently received a sweet Mozilla brand corrugated paperboard mobile VR viewer (no trademark issues here), and I was sad to learn that the nifty touch button/lever didnt't work with `cursor`.

This PR modifies `cursor` to listen for touch input events in addition to mouse events. With this small change, mobile users can have the same level of control as mouse users. This is especially useful in situations where gaze fusing is not optimal, like dragging an object around the scene to move it. (Here's an example of this with this modified version of `cursor` & `super-hands`: https://deserted-ladybug.glitch.me/)

**Changes proposed:**

- Register 'touchstart' and `touchend' listeners on the canvas alongside 'mousedown' and 'mouseup'

**Notes:**

f2bff70 is optional, it centralizes the listing of event types to facilitate future maintenance, but that might be unnecessary 

**Concerns:**

* It's not smart about multiple touches - the first 'touchend' will terminate the `cursor` interaction even if more touches remain active. I'd be happy to add handling for this, but didn't want to add complications unnecessarily. 
* It does not implement 'touchmove' alongside 'mousemove'. Since the event APIs for each are structured slightly differently, this would be more than a simple change. Again, happy to add this if you think it's worth the additional complication
* This might cause some unexpected behavior for projects using `cursor` for gaze fusing and capturing touch input manually for other purposes. If preferred, I could extend the schema make this configurable, either as a flag or as a list of event types.

Thanks to @andrewvee on slack for a discussion that prompted

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -26,11 +26,11 @@ traditional web development for now. Under the hood, the cursor component uses
 the `raycaster-intersected` and `raycaster-intersection-cleared` events,
 capturing the closest visible intersected entity.
 
-By default, the cursor is configured to be used in a gaze-based mode
-and will register user input via mouse or touch.
-Specifying the `downEvents` and `upEvents` properties allows the cursor to work
-with controllers.  The [laser-controls component][laser-controls] automatically
-configures those.
+By default, the cursor is configured to be used in a gaze-based mode and will
+register user input via mouse or touch. Specifying the `downEvents` and
+`upEvents` properties allows the cursor to work with controllers. For example,
+the [laser-controls component][laser-controls] automatically configures these
+properties to work with most controllers..
 
 [geometry]: ./geometry.md
 [line]: ./line.md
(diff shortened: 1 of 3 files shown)


machenmusik:
Careful testing this, as I think there may be namespace collision with tracked controls touch events

wmurphyrd:
I think we're safe because these are registered on the canvas, so no entity events will bubble to these listeners. Will double check with on Vive

ngokevin:
Cardboard buttons have been working with `mousedown`. `mousedown` should also register with touch. You weren't seeing it emitting?

wmurphyrd:
Here's the same example with the 0.6.1 A-Frame dist: https://grey-bench.glitch.me 
And it doesn't respond to touch on my surface in FF55 or Chrome nor on my Moto X in FF55

Putting some listeners on canvas to investigate, I see that a quick tap (our double tap on the surface) will generate 'mousedown', 'mouseup', and 'click' all at once, so that should trigger the cursor. 

However, pressing and holding or dragging does not generate any of the mouse events, so my example where the idea is to press and hold while moving the phone around doesn't work.

PR #2978, opened 2017-08-16
fix controller update issue #2975

### LINKED ISSUE #2975: controller position updates only every half second (experimental Chromium, Oculus Browser...)
As mentioned in Slack
https://aframevr.slack.com/archives/C0GG937RN/p1502830167000371

It seems likely that this is a Chromium-specific regression due to the throttled tick optimizations.

machenmusik:
Don't throttle tick; instead, only update controllers list when number of gamepads changes.

Fixes #2975.

### DIFF
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -11,7 +11,8 @@ module.exports.System = registerSystem('tracked-controls', {
     this.controllers = [];
 
     // For non-Firefox browsers, the gamepad poses only seem to update if getGamepads() is called.
-    // Therefore, don't throttle the tick frequency; instead, optimize the update check.
+    // Therefore, don't throttle the tick frequency.
+
     this.updateControllerList();
 
     if (!navigator.getVRDisplays) { return; }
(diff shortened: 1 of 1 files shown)


machenmusik:
Confirmed working with:
- Firefox 55, Rift & Touch
- experimental Chromium, Rift & Touch
- Oculus Browser, Gear VR & controller

Pretty sure it will also work for Daydream, but have not confirmed myself.

src/systems/tracked-controls.js
```diff
@@ -38,11 +36,23 @@ module.exports.System = registerSystem('tracked-controls', {
...
     var i;
     var prevCount;
+    var thisCount;
 
     gamepads = navigator.getGamepads && navigator.getGamepads();
     if (!gamepads) { return; }
 
+    // Count first, and only actually update controller list if length changes.
     prevCount = controllers.length;
+    thisCount = 0;
```
dmarcos: I think this check does not save much and is brittle. Controllers might change (one controller connects and another disconnects) and the number remains the same

machenmusik: OK.  I do think that if we can find some way to detect that there are no controller identity changes and not emit the controllersupdated event, that would be a significant savings.

machenmusik:
We merged #2984 instead (simpler).

PR #2983, opened 2017-08-17
Tiny change to the documentation

floostmodern:
Just a tiny change to the documentation that I believe makes it a bit more clearer.

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -508,9 +508,10 @@ AFRAME.registerComponent('c', {});
 
 [sound]: ../components/sound.md
 
-`multiple` allows for a component to have multiple instances per entities. By default, since
-`multiple` is set to `false`, entities could have only one instance of a component . For
-example, an entity could only have one geometry component.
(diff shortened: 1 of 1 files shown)


docs/core/component.md
```diff
@@ -508,8 +508,8 @@ AFRAME.registerComponent('c', {});
 
 [sound]: ../components/sound.md
 
-`multiple` allows for a component to have multiple instances. By default, since
-`multiple` is set to `false`, a component could have one instance. For
+`multiple` allows for a component to have multiple instances per entities. By default, since
```
dmarcos: Maybe simpler this way? `multiple allows multiple instances of the component on an entity`

docs/core/component.md
```diff
@@ -508,8 +508,8 @@ AFRAME.registerComponent('c', {});
...
+`multiple` is set to `false`, entities could have only one instance of a component . For
```
dmarcos: What about? `multiple is set to false, by default an entity can only have a single instance of the component`

PR #2984, opened 2017-08-17
Remove throttling of tracked-controls system tick. Chrome does not up…

dmarcos:
…date the gamepad poses unless we call getGamepads.

### DIFF
--- a/src/systems/tracked-controls.js
+++ b/src/systems/tracked-controls.js
@@ -1,5 +1,4 @@
 var registerSystem = require('../core/system').registerSystem;
-var utils = require('../utils');
 
 /**
  * Tracked controls system.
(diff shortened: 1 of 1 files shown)


machenmusik:
Why not #2978?

machenmusik:
I think that part of the prior optimization is to avoid rebuilding controller list and emitting event when not needed.

dmarcos:
I'm fine with https://github.com/aframevr/aframe/pull/2978 if it looks like this one 😄

machenmusik:
ok - modified #2978, although it may later be worth a little thought on what the right check to avoid emitting `controllersupdated` would be, since handling that is pretty expensive

dmarcos:
For the sake of simplicity I'm merging since it's what we need to fix the bug at hand.
We can do improvements on how we check the gamepads on https://github.com/aframevr/aframe/pull/2978

Issue #2967, opened 2017-08-12
vr2vr traversal fails with Oculus Browser


machenmusik:
A-Frame VR-to-VR link traversals fail at present with Oculus Browser.

See Slack discussion at https://aframevr.slack.com/archives/C0GG937RN/p1502489851298903

See #2966 for a workaround PR.

vincentfretin:
Testing with https://wmurphyrd.github.io/aframe-super-hands-component/examples/portals.html
so with aframe 0.7.1 and latest oculus browser, it now loads correctly the scene when traversing a portal. It was not the case in september, it did quit the vr mode and there was some sort of infinite loop.
But after traversing two portals, it starts dropping frames, and eventually freeze...

vincentfretin:
Well today it froze after 4 portals. Probably a memory issue? Is the previous scene freed when we enter a portal?

dmarcos:
Freeing resources of the previous page on navigation should be task of the browser. Not sure what A-Frame can do.

machenmusik:
depends whether a-frame is caching things that can be freed e.g. image pano textures

this magic incantation used to force that texture cache to be flushed 
`AFRAME.scenes[0].systems.material.clearTextureCache()`
but that may not be what you want

vincentfretin:
I don't know if I can analyze the memory of oculus browser with the remote chrome debugger on my machine. I didn't try it yet, I don't have much experience with this memory recording stuff on chrome.
I know how to use the allocations recording on firefox though, so I can already try with firefox on my machine to see if the memory is increasing when traversing portals.

vincentfretin:
@dmarcos you're right, the link component just do a `window.location = href` in the end.

I did some benchmarks with chrome profiling tool. I'm not seeing a real memory issue, only the usual increase and decrease of memory from the GC of camera position/rotation created objects and arrays (see #3183)
By looking at profiles and the code, I did find an optimization to do in the link component #3224 though.

I guess you can close the issue, the vr2vr traversal works great in itself. The freeze or crash of the Oculus browser is an issue of this browser.

dmarcos:
@vincentfretin thanks for investigating. it is really helpful 🙏

PR #2985, opened 2017-08-17
Moves canvas initialization logic from a component to the scene. This…

dmarcos:
… handle the case when calling enterVR before the renderer is initialized and also moves the canvas and renderer initialization before assets load (it's not a component anymore). This is a prequesite for a loading screen / spinner

**Description:**

**Changes proposed:**
-
-
-


tests/core/scene/a-scene.test.js
```diff
@@ -385,7 +385,7 @@ suite('a-scene (without renderer)', function () {
       var sceneEl = this.el;
       sceneEl.innerHTML = 'NEW';
       sceneEl.reload();
-      assert.equal(sceneEl.innerHTML, '');
+      assert.equal(sceneEl.innerHTML, '<canvas class="a-canvas" data-aframe-canvas="true"></canvas>');
```
ngokevin: We can do a more robust test with query selector rather than string check.

src/core/scene/a-scene.js
```diff
@@ -449,6 +450,46 @@ module.exports.AScene = registerElement('a-scene', {
       writable: window.debug
     },
 
+    setupCanvas: {
```
ngokevin: We could move this out of the prototype to pure function. `setupCanvas (sceneEl)`.

src/core/scene/a-scene.js
```diff
@@ -449,6 +450,46 @@ module.exports.AScene = registerElement('a-scene', {
...
+
+        // Prevent overscroll on mobile.
+        canvasEl.addEventListener('touchmove', function (event) {
+          event.preventDefault();
+        });
+
+        // Set canvas on scene.
+        sceneEl.canvas = canvasEl;
+        sceneEl.emit('render-target-loaded', {target: canvasEl});
+        // For unknown reasons a syncrhonous resize does not work on desktop when
```
ngokevin: can fix this typo on synchronous

ngokevin:
r+wc

machenmusik:
FYI - this does not solve #2967

dmarcos:
Yes, `vrdisplayactivate` event handler has to be attached earlier: when AFRAME loads instead of when a-scene initializes

machenmusik:
Ok tell you what, since I think we both agree solution to #2967 based on this would be better than current #2966, let me try to redo that on top of this PR.  May take a while for me to get to it

Issue #2693, opened 2017-05-23
Upgrade plans for glTF 2.0


donmccurdy:
I think GLTF2Loader will be ready for A-Frame v0.7.0, but I'm not sure COLLADA2GLTF and the Blender exporter will be updated to glTF2.0 quickly enough. We may not want to bump `gltf-model` until the workflows are stabilized. There is also a tool for upgrading glTF1.0 assets to glTF2.0, so if that's reliable then maybe we don't have to wait on other stuff. We could also have both `gltf-model` and `gltf2-model`, but since three.js is planning to drop GLTFLoader eventually, I think that makes less sense.

For now, I have a `gltf-model-next` component in aframe-extras.

takahirox:
glTF 2.0 has been released!
https://www.khronos.org/news/press/khronos-releases-gltf-2.0-specification

donmccurdy:
Yay!

I think my vote would be:

- Plan to upgrade to three.js r85 for A-Frame 0.7.0.
- Update `gltf-model` from GLTFLoader to GLTF2Loader then, without renaming the component.

Waiting for 0.7.0 also buys some time for the tooling, since COLLADA2GLTF updates haven't been merged yet.

sirkitree:
Just a note that I recently used the gltf 2.0 exporter for blender and found using the core `gltf-model` component doesn't work currently, subsequently found this thread.

Has there been any work on getting the 2.0 spec working with `gltf-model` yet?

mokargas:
@sirkitree You can use `gltf-model-next` from @donmccurdy A-Frame Extras package https://github.com/donmccurdy/aframe-extras/blob/master/src/loaders/gltf-model-next.js

takahirox:
And one more thing.
Make sure to use the latest Blender glTF exporter version.
Recently it's fixed a skinning bug.

donmccurdy:
Quick recap of my thoughts on timing here:

1. THREE.GLTF2Loader is stable already, available in three.js r85 and r86.
2. If we swap GLTF2Loader for GLTFLoader under the hood in `gltf-model`, that will be a breaking change. glTF1.0 models will no longer be compatible. Is that OK? Shall I just create a `gltf-model-legacy` component in aframe-extras, similar to the `gltf-model-next` component I'm maintaining now?
[...]

PR #2986, opened 2017-08-18
Update gltf-model to glTF 2.0.

donmccurdy:
Example assets bumped with https://github.com/aframevr/assets/pull/27.

the glTF2.0 model is not backward-compatible, so users upgrading to A-Frame 0.7.0 but wanting to continue using their glTF1.0 assets will need to use [`gltf-model-legacy` frome aframe-extras](https://github.com/donmccurdy/aframe-extras/tree/master/src/loaders).

Fixes https://github.com/aframevr/aframe/issues/2693.


donmccurdy:
Hm, in the interest of not breaking 0.6.0 examples, should the assets be added to a new folder, leaving the glTF 1.0 ones alone?

dmarcos:
I think we won't break anything because the site won't deploy until `0.7.0`. Am I right?

docs/components/gltf-model.md
```diff
@@ -91,17 +96,18 @@ rendering.
...
-- [glTF Workflow for A Saturday Night](https://blog.mozvr.com/a-saturday-night-gltf-workflow/)
-- [FBX &rarr; glTF][fbx-converter] - Coming soon.
-- [COLLADA &rarr; glTF][collada-converter]
+Over 100,000 glTF models are free for download on [Sketchfab][sketchfab], and
+various exporters and converters converters are available:
+
+- [Blender Exporter](blender-exporter)
 - [OBJ &rarr; glTF][obj-converter]
+- [COLLADA &rarr; glTF][collada-converter]
+- [glTF Workflow for A Saturday Night](https://blog.mozvr.com/a-saturday-night-gltf-workflow/)
```
dmarcos: Is the blog post still relevant? Is it worth to bring it up to date?

dmarcos:
It looks good to me. I think merging this won't affect the current deployed examples on the site. I was wondering if the `a-saturday-night` blog post should be updated with a glTF2 workflow (https://blog.mozvr.com/a-saturday-night-gltf-workflow/) but we should first update the assets to glTF2. Maybe not worth now. Should we keep the link then? Is it the info in the blog post still relevant?

dmarcos:
Thank you very much! This makes me ultra happy. 😄 🎉 🎈 🎆 🍾

donmccurdy:
@dmarcos Yay! Regarding the `a-saturday-night` blog post, a lot of what's written there is explaining how to work around limitations in the old Blender exporter, by exporting to COLLADA first. The new Blender exporter is much more reliable, and COLLDA2GLTF's glTF2.0 support is not finished, so I think we want to recommend using the Blender Exporter these days.

My vote would be to write a shorter new blog post, and then put an `UPDATE: ...` at the top of the old one linking to that. Unfortunately it's unlikely I would have time to write that in the next couple weeks. 😕

PR #2987, opened 2017-08-18
Correcting docs for various primitive attributes

omgitsraven:
**Description:**
A few of the a-frame primitives use 'height' and 'width' to refer to properties of the geometry, but the docs seem to assume they're referring to properties of the material. (I'm not sure if this suggests an underlying problem with whatever originally generated these tables in the docs, but it was suggested in the Slack that I just go ahead and edit the docs manually, so I have.)

**Changes proposed:**
- Fix the 'component mapping' and 'default value' for the 'height' and (if used in the geometry constructor) the 'width' attributes, in the docs for the 'a-box', 'a-cone', and 'a-cylinder' primitives
- It seems like someone already fixed the component mappings in 'a-plane', but didn't fix the defaults; I've done that now


dmarcos:
Thanks!

PR #2993, opened 2017-08-24
Updated html example in visual feedback

Absulit:
**Description:**
I was trying to use these examples and realized they were not working with the most recent versions of A-Frame. Made a copy of the codepen example and added some changes to make it work.

Also the HTML example in the page works different, so I made a few changes, the 'fill' property in both animations started with 'backwards', now the first one starts with 'forwards', and removed the 'fuseTimeout' since it seems that the 'dur' properties in animation Entity take care of that.

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -154,18 +154,18 @@ intersects the entity, it will emit an event, and the animation system will
 pick up event with the `begin` attribute:
 
 ```html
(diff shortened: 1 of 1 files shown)


ngokevin:
Cool, also feel free to create a Glitch instead so we can link it to the sidebar of that documentation page.

e.g.,

```
examples:
  - title: Adding Visual Feedback
     src: https://glitch.com/~absulit-cursor-feedback
```

PR #2994, opened 2017-08-24
Bump THREE to r87

dmarcos:


dmarcos:
@donmccurdy Bumping THREE to r87. It seems that `GLTF2Loader` has been renamed to `GLTFLoader`

donmccurdy:
Does r87 force us to change any other VREffect usage?

dmarcos:
I don't believe show. I will do a sanity check in VR to make sure it works.

dmarcos:
Tried on Gear VR and everything seems to work.

PR #2991, opened 2017-08-24
extend PR #2985 (moved canvas init) to also solve issue #2967 (vr2vr traversal in Oculus Browser)

machenmusik:
Deprecates #2966.
See further discussion in #2966 / #2967 / #2985.

PR #2996, opened 2017-08-24
redo #2991: extend PR #2985 (moved canvas init) to also solve issue #2967 (vr2vr traversal in Oculus Browser)

machenmusik:
per @dmarcos:
I merged by mistake and reverted. Sorry for that. Can you please reopen? #2985 has landed so we need to rebase

(Rebase forthcoming.)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -117,5 +117,6 @@ module.exports = window.AFRAME = {
 // Enter VR on `vrdisplayactivate`, to handle in-VR traversal much earlier in the lifecycle.
 // This event may also fire in other circumstances (e.g. putting on the Rift headset).
 window.addEventListener('vrdisplayactivate', function (evt) {
-  if (window.AFRAME.scenes.length > 0) { window.AFRAME.scenes[0].enterVR(); }
+  var s = window.AFRAME.scenes[0];
+  if (s.effect) { s.enterVR(); } else { setTimeout(s.enterVRBound); }
 });
--- a/src/index.js
+++ b/src/index.js
@@ -113,3 +113,10 @@ module.exports = window.AFRAME = {
   utils: utils,
   version: pkg.version
 };
+
+// Add vrdisplayactivate event listener early.
+window.addEventListener('vrdisplayactivate', function (evt) {
+  if (window.AFRAME.scenes.length > 0) {
+    window.AFRAME.scenes[0].enterVR();
+  }
+});
--- a/src/index.js
+++ b/src/index.js
@@ -114,7 +114,8 @@ module.exports = window.AFRAME = {
   version: pkg.version
 };
 
-// Add vrdisplayactivate event listener early.
+// Enter VR on `vrdisplayactivate`, to handle in-VR traversal much earlier in the lifecycle.
+// This event may also fire in other circumstances (e.g. putting on the Rift headset).
 window.addEventListener('vrdisplayactivate', function (evt) {
   if (window.AFRAME.scenes.length > 0) {
     window.AFRAME.scenes[0].enterVR();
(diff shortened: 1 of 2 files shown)


dmarcos:
rebase needed to remove the commit that has already landed

machenmusik:
rebase done - are you in a position to sanity check?

dmarcos:
I tried quickly but it does not seem to work for me. I did not have the controller though (I hacked the link traversal demo with a gaze based cursor). Tomorrow I'll try with the proper controller again

machenmusik:
Works for me with this test page as example

https://s.codepen.io/machenmusik/debug/XaVMWJ/mVAbGERyxabk

Once you enter VR, it reloads in 5 seconds.  So if you end up out of VR, you know it is not working.

dmarcos:
I'm trying the link traversal example with your branch and the controller and I get dropped of vr. Does examples/showcase/link-traversal work for you?

machenmusik:
for me, it is working for simpler scenes, but not for ones with big assets to download -- I suspect it runs into the timeout issue in those cases

machenmusik:
looks like the setTimeout from prior PR is still needed for the VREffect to be available in the heavier cases.

machenmusik:
@dmarcos can you confirm latest PR version works for you?

src/index.js
```diff
@@ -113,3 +113,10 @@ module.exports = window.AFRAME = {
   utils: utils,
   version: pkg.version
 };
+
+// Enter VR on `vrdisplayactivate`, to handle in-VR traversal much earlier in the lifecycle.
+// This event may also fire in other circumstances (e.g. putting on the Rift headset).
+window.addEventListener('vrdisplayactivate', function (evt) {
+  var s = window.AFRAME.scenes[0];
+  if (s.effect) { s.enterVR(); } else { setTimeout(s.enterVRBound); }
```
dmarcos: I don't think it is guaranteed that in the next tick the effect is going to be defined. I'm trying to find a more robust way to do this.

dmarcos: Also is not what I would not expect the event to behave following the spec. The user action policy should not apply to the next tick logic.

machenmusik: If you find a more robust way, great.  I tried several things and did not find another way for gesture permission to survive.  See https://github.com/WICG/interventions/issues/12

machenmusik: (More specifically, https://github.com/WICG/interventions/issues/12#issuecomment-209908148)

machenmusik:
@dmarcos any progress on a more robust solution?  (Where this.effect will definitely exist while preserving user gesture?)

cvan:
@machenmusik out of curiosity, which example(s) are you testing with this? just https://aframe.io/aframe/examples/showcase/link-traversal/, yeah?

dmarcos:
This will be handled in un upcoming PR implementing a loading screen

machenmusik:
@cvan perhaps irrelevant now, but link-traversal example, this codepen (https://s.codepen.io/machenmusik/debug/XaVMWJ/mVAbGERyxabk) and a non-public app

machenmusik:
@dmarcos ok, as we discussed at OC4

dmarcos:
@machenmusik thanks for the conversation at OC4

cvan:
@machenmusik that CodePen URL gave me a 403. just posting this in case anyone else's reading: https://codepen.io/machenmusik/pen/XaVMWJ/

machenmusik:
Thanks and sorry @cvan!

cvan:
no sweat! I was just testing WebVR flows in Oculus Browser. and I know you were working on this, so wanted to check.

thanks for working on this. before I do a deeper dive, are there any known issues you know of for A-Frame + Gear VR? I ask because the new Oculus headsets are coming soon, and I want to make sure A-Frame and three.js are prepared.

thanks again for all your PRs and keeping your finger on the pulse :+1:

PR #3000, opened 2017-08-25
Added support for VR headsets that do not provide stageParameters 

olga-microsoft:
**Description:**
Fixes issue of inconsistent head and controller height offset, for cases when stageParameters property on VR display is null.

**Changes Proposed:**
Provide default user height, if this is the case to be consistent with tracked controls.


src/components/look-controls.js
```diff
@@ -17,7 +18,9 @@ module.exports.Component = registerComponent('look-controls', {
     enabled: {default: true},
     hmdEnabled: {default: true},
     reverseMouseDrag: {default: false},
-    standing: {default: true}
+    standing: {default: true},
+    userHeight: {default: 0},
```
dmarcos: Are you using this property anywhere? it seems you are using userHeight from the camera component in `getUserHeight`

src/components/look-controls.js
```diff
@@ -54,11 +57,21 @@ module.exports.Component = registerComponent('look-controls', {
...
     this.controls.update();
     this.updateOrientation();
     this.updatePosition();
   },
 
+  /**
+   * Return user height to use for standing poses, where a device doesn't provide an offset.
+   */
+  getUserHeight: function () {
+    var headEl = this.data.headElement || this.el.sceneEl.camera.el;
```
dmarcos: What are the cases where the headElement is not the active camera?

src/components/look-controls.js
```diff
@@ -17,7 +18,9 @@ module.exports.Component = registerComponent('look-controls', {
     enabled: {default: true},
     hmdEnabled: {default: true},
     reverseMouseDrag: {default: false},
-    standing: {default: true}
+    standing: {default: true},
+    userHeight: {default: 0},
+    headElement: {type: 'selector'}
```
dmarcos: Can we add a description for the use cases of this property?

tests/components/look-controls.test.js
```diff
@@ -52,4 +53,27 @@ suite('look-controls', function () {
       window.dispatchEvent(new Event('mouseup'));
     });
   });
+
+  suite('head-height', function () {
```
dmarcos: Thanks for the tests!

olga-microsoft:
Thank you @dmarcos for review. I removed the properties under question. 
Merged master in order to get tests pass.
Please let me know, if there are any more concerns.

src/components/look-controls.js
machenmusik: (e.g. spectator camera?)

src/components/look-controls.js
```diff
@@ -56,11 +57,21 @@ module.exports.Component = registerComponent('look-controls', {
...
     this.controls.update();
     this.updateOrientation();
     this.updatePosition();
   },
 
+  /**
+   * Return user height to use for standing poses, where a device doesn't provide an offset.
+   */
+  getUserHeight: function () {
+    var headEl = this.el.sceneEl.camera.el;
```
dmarcos: A scene could have multiple cameras. In this case we probably only care about the camera on the entity where this component is setup. I would use the public API this way:
````javascript 
var el = this.el;
var userHeight = el.hasAttribute('camera') && el.getAttribute('camera').userHeight || DEFAULT_CAMERA_HEIGHT;
return userHeight;
````

dmarcos: Does this work in your case?

dmarcos: I'm worried about the coupling of look-controls and the camera right now. Moving userHeight to look-controls is something we have discussed in the past but we can do it in a separate PR

dmarcos:
Thanks! Pretty close! A small couple of tweaks and we're ready to go.

src/components/look-controls.js
olga-microsoft: Yes, this works and looks better. Thank  you!

dmarcos:
Nice! Thank you! Welcome as a contributor!

olga-microsoft:
Thank you!

Issue #3001, opened 2017-08-26
Appending a-abox in a-scene fails in Chrome


jgbarah:
**Description:**

The code below works well in Firefox 55.0 (shows a white box on a blue sky), but fails in Chrome (I've tested in "60.0.3112.113 (Official Build) (64-bit)" (for Debian) and some earlier versions). In Firefox, I don't see anything remarkable in the console, but in Chrome I see the errors below.

The problem seems to be in:

```
browser.js:111 core:a-node:warn You are attempting to attach <A-BOX> outside of an A-Frame scene. 
Append this element to `<a-scene>` instead. 
```

which is weird, since I'm only getting the a-scene, building a minimal a-box, and then attaching the a-box to the a-scene...

Errors in Chrome console:

```
A-Frame Version: 0.6.1 (Date 19-07-2017, Commit #a4b91bc)
index.js:87 three Version: ^0.84.0
index.js:88 WebVR Polyfill Version: ^0.9.36
browser.js:111 core:a-node:warn You are attempting to attach <A-BOX> outside of an A-Frame scene. Append this element to `<a-scene>` instead. 
browser.js:111 core:a-node:warn You are attempting to attach <A-BOX> outside of an A-Frame scene. Append this element to `<a-scene>` instead. 
material.js:151 Uncaught (in promise) TypeError: Cannot read property 'registerMaterial' of undefined
    at i.setMaterial (material.js:151)
    at i.updateShader (material.js:101)
    at i.update (material.js:47)
    at i.updateProperties (component.js:271)
    at i.module.exports.Component (component.js:52)
    at new i (component.js:458)
    at HTMLElement.value (a-entity.js:377)
    at HTMLElement.value (a-entity.js:544)
    at e (a-entity.js:513)
    at Array.forEach (<anonymous>)
...
value @ a-entity.js:495
(anonymous) @ a-entity.js:306
(anonymous) @ a-node.js:117
Promise resolved (async)
value @ a-node.js:115
value @ a-entity.js:302
value @ a-entity.js:84
value @ a-register-element.js:150
value @ a-register-element.js:150
(anonymous) @ (index):11
three.js:19802 THREE.WebGLRenderer 84
```

- A-Frame Version: 0.6.1
- Platform / Device: "60.0.3112.113 (Official Build) (64-bit)" (for Debian) 
- Reproducible Code Snippet or URL:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
[...]

PR #3002, opened 2017-08-26
Update javascript-events-dom-apis.md

jgbarah:
See #3001 for the problem I found with some code that didn't run in Chrome. Thanks to @dmarcos and @ngokevin, who showed the need to wait until `a-scene` is loaded.

This section of the tutorial is the first one entering into details of how to write JavaScript code to manipulate a scene, so I think it needs these clarifications. Otherwise, you can lose a lot of time and energy until you realize your code should be in a listener for `loaded` of the `a-frame` element.

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/introduction/javascript-events-dom-apis.md
+++ b/docs/introduction/javascript-events-dom-apis.md
@@ -71,6 +71,29 @@ Components encapsulate all of our code to be reusable, declarative, and
 shareable. Though if we're just poking around at runtime, we can use our
 browser's Developer Tools Console to run JavaScript on our scene.
 
+## Waiting until entities are loaded
+
+Entities, including `a-scene`, are not immediately available after they
+attached in the DOM. Some initialization code needs to be run before.
+When an A-Frame entity is ready, it will fire the `loaded` event.
+Therefore, if we need to something whith it, we can just add a listener
+for that event.
+
+For example, if we want to run almost any JavaScript code that
+manipulates the scene (for example, adding a new entity to it),
+we need to have `a-scene` loaded.
+Therefore, any such code should be included in the appropriate listener:
+
+```js
+document.querySelector('a-scene').addEventListener('loaded', function() { 
+  // your code goes here 
+});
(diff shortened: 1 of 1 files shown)


ngokevin:
We do not recommend this way. The very first section notes where to place code for JavaScript. Within a component, so that we don't have to wait to load at all:

https://aframe.io/docs/0.6.0/introduction/javascript-events-dom-apis.html#where-to-place-javascript-code-for-a-frame

> Before we go over the different ways to use JavaScript and DOM APIs, we recommend encapsulating your JavaScript code within A-Frame components. Components modularize code, make logic and behavior visible from HTML, and ensure that code is executed at the correct time (e.g., after the scene and entities have attached and initialized). As the most basic example, to register a console.log component.

The examples on the right hand sidebar also provide remixable examples to do so (as do all the other examples in the School):

https://glitch.com/edit/#!/aframe-school-js?path=solution3.html

Also covered in https://aframe.io/docs/0.6.0/core/scene.html#running-content-scripts-on-the-scene

> The recommended way is to write a component, and attach it to the scene element. The scene and its children will be initialized before this component.

jgbarah:
Thanks a lot. However, maybe a more prominent text about that would help. I went through [that page](https://aframe.io/docs/0.6.0/introduction/javascript-events-dom-apis.html#where-to-place-javascript-code-for-a-frame), but I didn't notice the problem. Maybe because I was not using a raw script after `<a-scene>`...

I've submitted a new pull request, #3005, about this, just in case you agree a clarification is convenient. Just close it if not.

In any case, thanks a lot for the support and explanations.

ngokevin:
The first section says explicit where to put JS code for A-Frame and numerous examples on the right hand side. The code in the original issue did not follow which was the issue, nowhere recommends using DOMContentLoaded in a script tag before the scene. Thanks though!

jgbarah:
OK. Sorry for the noise.

PR #3005, opened 2017-08-27
Add text: a-scene not ready right after being in DOM

jgbarah:
Add some text to stress that not only you're in trouble if you write some code in a `<script>` tag after `<a-scene>`, but in general if you attach to `<a-scene>` right after it is in the DOM (which IMHO is what is surprising, when you are used to other JS frameworks).

Context: #3002.

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/introduction/javascript-events-dom-apis.md
+++ b/docs/introduction/javascript-events-dom-apis.md
@@ -73,7 +73,10 @@ browser's Developer Tools Console to run JavaScript on our scene.
 [contentscripts]: ./scene.md#running-content-scripts-on-the-scene
 
 Do **not** try to put A-Frame-related JavaScript in a raw `<script>` tag after
-`<a-scene>` as we would with traditional 2D scripting. If we do, we'd have to
+`<a-scene>` as we would with traditional 2D scripting,
+and in general, don't asssume `<a-scene>` is ready to attach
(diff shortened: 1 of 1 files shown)


ngokevin:
What does attach to a-scene mean? Im not sure it provides any extra information. We've said where to put JS code and where not to put it.

ngokevin:
Thanks! Your original issue, I made stuff bolder and more clear hopefully if other people run into the same thing, and I've added additional examples. I'll remove the `after the scene` part to recommend not putting code in raw script tags outside components at all.

jgbarah:
> What does attach to a-scene mean? Im not sure it provides any extra information. We've said where to put JS code and where not to put it.

Maybe "attach an entity to `<a-scene>`"? The fact is that when I read the current text, I thought it only applied to inlined scripts right after `<a-scene>` and I thought a clarification would help to newbies as me... The fact that it is not enough to write the code in a listener for the DOM ready event should be highlighted. I naively expected that the `<a-scene>` element was ready for children as soon as it was in the DOM, and maybe other would think the same.

However, as I said in the issue, if you don't see it this way, don't mind. I already took too much time from you.

jgbarah:
OK, thanks a lot!

PR #3006, opened 2017-08-28
adding sky glitch exercise

afvc:

PR #3007, opened 2017-08-28
adding geometry glitches examples

afvc:

### DIFF
--- a/docs/components/geometry.md
+++ b/docs/components/geometry.md
@@ -6,7 +6,23 @@ parent_section: docs
 section_title: Components
 section_order: 4
 source_code: src/components/geometry.js
-examples: []
+examples: 
+
+ - title: Ex2 — Creating Shapes
+    src: https://glitch.com/edit/#!/ex-2-geometry?path=index.html:1:0
+
+ - title: Ex2.1 — Texture on Shapes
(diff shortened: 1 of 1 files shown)


docs/components/geometry.md
```diff
@@ -6,7 +6,23 @@ parent_section: docs
 section_title: Components
 section_order: 4
 source_code: src/components/geometry.js
-examples: []
+examples: 
+
+ - title:Creating Shapes
+    src: https://glitch.com/edit/#!/ex-2-geometry?path=index.html:1:0
```
cvan: you can probably remove `?path=index.html:1:0` (or add them to the other URLs)

cvan:
thanks for the contributions. these are nice 👍

docs/components/geometry.md
ngokevin: Nope, you need these so they don't route to the README.

ngokevin:
Like https://github.com/aframevr/aframe/pull/3006, these require some work to format and make consistent. If @afvc is free, or whenever I have a bit of free time, we can get around to it.

docs/components/geometry.md
cvan: my question was more about the consistency with the URLs defined below

PR #3008, opened 2017-08-28
light types glitch example

afvc:


ngokevin:
I think a light example should have a way to quick be able to see different light types because this is all people will see:

screen shot 2017-08-28 at 4 53 35 pm

I'd add in-VR buttons to swap the light types or even spawn lights. That would mean a pretty much reworked example so I'll close this for now. Thanks!

PR #3009, opened 2017-08-28
adding glitch example

afvc:


ngokevin:
In case of any future Glitches (more are welcome!), we can remix this base Glitch to maintain some consistency/formatting: https://glitch.com/~aframe

Working on merging them now

ngokevin:
Thanks! I think for UI, we should have componentized solutions versus JS in a script tag running animations (we prescribe JS code within components). Also some with more visual polish (e.g., demonstrating https://github.com/etiennepinchon/aframe-material with controllers).

PR #3012, opened 2017-08-29
Allow Decoupling of TouchEvents and Look-Contols

jsimonson2013:
**Description:**

While working on an AR.js + A-Frame app that allows users to draw/paint in AR, I wanted users to be able to use touch events to draw on a mobile device and use device orientation to change camera orientation. Touch events are included in `look-controls` by default, with no disable action allowed. Using `touchmove` events to draw resulted in drawing across the scene in a reverse pan because of the way that `touchmove` is coupled with `look-controls`. In magic window mode, there could be many applications of touch events that would not want to also change device orientation.

**Changes proposed:**
- Add property to look-controls schema
- Check property for true before binding, adding, and removing TouchEvent-related functions


src/components/look-controls.js
```diff
@@ -77,9 +78,11 @@ module.exports.Component = registerComponent('look-controls', {
...
     this.onMouseMove = bind(this.onMouseMove, this);
     this.onMouseUp = bind(this.onMouseUp, this);
-    this.onTouchStart = bind(this.onTouchStart, this);
-    this.onTouchMove = bind(this.onTouchMove, this);
-    this.onTouchEnd = bind(this.onTouchEnd, this);
+    if (this.data.touchEnabled) {
+      this.onTouchStart = bind(this.onTouchStart, this);
+      this.onTouchMove = bind(this.onTouchMove, this);
+      this.onTouchEnd = bind(this.onTouchEnd, this);
+    }
```
donmccurdy: This event binding should happen regardless of `data.touchEnabled`, because touch events might be enabled or disabled at a later time.

src/components/look-controls.js
```diff
@@ -144,9 +149,11 @@ module.exports.Component = registerComponent('look-controls', {
...
 
     // Touch events.
-    canvasEl.removeEventListener('touchstart', this.onTouchStart);
-    canvasEl.removeEventListener('touchmove', this.onTouchMove);
-    canvasEl.removeEventListener('touchend', this.onTouchEnd);
+    if (this.data.touchEnabled) {
+      canvasEl.removeEventListener('touchstart', this.onTouchStart);
+      canvasEl.removeEventListener('touchmove', this.onTouchMove);
+      canvasEl.removeEventListener('touchend', this.onTouchEnd);
+    }
```
donmccurdy: It may be better to return early from these event listeners when touch is disabled, rather than conditionally binding or not. Otherwise, (1) `setAttribute('look-controls', {touchEnabled: false})` will have no effect, and (2) calls to `pause()` after that will fail to remove the old touch listeners.

So instead:

```js
onTouchStart: function () {
  if (!this.data.touchEnabled) { return; }

  // ...
}
```

examples/index.html
```diff
@@ -172,6 +172,7 @@
 
     <ul class="links">
       <li><a href="boilerplate/hello-world/">Hello World</a></li>
+        <li><a href="boilerplate/hello-world-touch-disabled/">Hello World Touch Disabled</a></li>
```
donmccurdy: It's not necessary to add an example for this, but could you instead add an entry in the look-controls documentation for the new option?

examples/boilerplate/hello-world-touch-disabled/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
```
dmarcos: Not sure if we need an example specific for just one flag.

examples/index.html
dmarcos: Yes, It seems that just a flag does not justify a whole example.

src/components/look-controls.js
dmarcos: Yes, This condition is not necessary. We should bind the methods regardless of the flag

src/components/look-controls.js
```diff
@@ -123,9 +126,11 @@ module.exports.Component = registerComponent('look-controls', {
     window.addEventListener('mouseup', this.onMouseUp, false);
 
     // Touch events.
-    canvasEl.addEventListener('touchstart', this.onTouchStart);
-    window.addEventListener('touchmove', this.onTouchMove);
-    window.addEventListener('touchend', this.onTouchEnd);
+    if (this.data.touchEnabled) {
```
dmarcos: We also have to handle the dynamic case when the flag changes at runtime. We probably need a method `updateTouchListeners` that contains this logic and its called from here and the `update` method.

jsimonson2013:
Thanks for the feedback @donmcurdy and @dmarcos . I will implement those suggested changes soon.

src/components/look-controls.js
jsimonson2013: Addressed in commit 1d6930b61ac66d080bfe57eada841a032fd6e6e8

src/components/look-controls.js
jsimonson2013: Addressed in commit 1d6930b61ac66d080bfe57eada841a032fd6e6e8

examples/index.html
jsimonson2013: Addressed in commit 1d6930b61ac66d080bfe57eada841a032fd6e6e8

examples/boilerplate/hello-world-touch-disabled/index.html
jsimonson2013: Addressed in commit 1d6930b61ac66d080bfe57eada841a032fd6e6e8

src/components/look-controls.js
jsimonson2013: When the `enabled` flag is updated at runtime, a grab cursor is shown or hidden accordingly. I'm not sure what would need to be updated, other than the behavior that the flag controls (updated in commit 1d6930b61ac66d080bfe57eada841a032fd6e6e8). Would the behavior change accordingly when the flag is set at runtime?

donmccurdy: I think this is taken care of, given the last set of changes, yes?

dmarcos:
Thanks!

PR #3013, opened 2017-08-30
Added support for windows mixed reality controllers

leweaver:
**Description:**
Added support for windows motion controllers, with tests, which fixed one or two bugs in tracked-controls class.

Also consolidated the logic used in tracked-controls utils and component for isControllerPresent to a single function.

This PR does not yet have the final controller GLB asset files, - I do plan on uploading those to the a-frame CDN however I am still waiting for the OK internally here.

**Changes proposed:**
- Added error callback on glTF-model component
- Windows Motion Controller support 
- Consolidation of tracked-controls duplicated logic to a single utility function

### DIFF
--- a/docs/components/windows-motion-controls.md
+++ b/docs/components/windows-motion-controls.md
@@ -1,0 +1,60 @@
+---
+title: windows-motion-controls
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/windows-motion-controls.js
+examples: []
+---
+
+[trackedcontrols]: ./tracked-controls.md
+
+The windows-motion-controls component interfaces with any spatial controllers exposed through
+Windows Mixed Reality as Spatial Input Sources (such as Motion Controllers). 
+It wraps the [tracked-controls component][trackedcontrols] while adding button
+mappings, events, and a controller model that highlights applies position/rotation transforms 
+to the pressed buttons (trigger, grip, menu, thumbstick, trackpad) and moved axes (thumbstick and trackpad.)
+
+## Example
+
+```html
+<a-entity windows-motion-controls="hand: left"></a-entity>
+<a-entity windows-motion-controls="hand: right"></a-entity>
+```
+
+## Value
+
+| Property             | Description                                                                                       | Default Value  |
+|----------------------|---------------------------------------------------------------------------------------------------|----------------|
+| hand                 | The hand that will be tracked (i.e., right, left).                                                | right          |
+| pair                 | Which pair of controllers, if > 2 are connected.                                                  | 0              |
+| model                | Whether the controller model is loaded.                                                           | true           |
+| hideDisconnected     | Disable rendering of controller model when no matching gamepad (based on ID & hand) is connected. | true           |
+
+
+## Events
+
+| Event Name             | Description                                     |
+| ----------             | -----------                                     |
+| thumbstickdown         | Thumbstick button pressed.                      |
+| thumbstickup           | Thumbstick button released.                     |
+| thumbstickchanged      | Thumbstick button changed.                      |
+| thumbstickmoved        | Thumbstick axis moved.                          |
+| triggerdown            | Trigger pressed.                                |
+| triggerup              | Trigger released.                               |
+| triggerchanged         | Trigger changed.                                |
+| gripdown               | Grip button pressed.                            |
+| gripup                 | Grip button released.                           |
+| gripchanged            | Grip button changed.                            |
+| menudown               | Menu button pressed.                            |
+| menuup                 | Menu button released.                           |
+| menuchanged            | Menu button changed.                            |
+| trackpaddown           | Trackpad pressed.                               |
+| trackpadup             | Trackpad released.                              |
+| trackpadchanged        | Trackpad button changed.                        |
+| trackpadmoved          | Trackpad axis moved.                            |
+| controllerdisplayready | The model file is loaded and completed parsing. |
+
+## Assets
+
+TBC.
--- a/docs/components/windows-motion-controls.md
+++ b/docs/components/windows-motion-controls.md
@@ -53,7 +53,7 @@ to the pressed buttons (trigger, grip, menu, thumbstick, trackpad) and moved axe
 | trackpadup             | Trackpad released.                              |
 | trackpadchanged        | Trackpad button changed.                        |
 | trackpadmoved          | Trackpad axis moved.                            |
-| controllerdisplayready | The model file is loaded and completed parsing. |
+| controllermodelready   | The model file is loaded and completed parsing. |
 
 ## Assets
 
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -1,5 +1,6 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var warn = utils.debug('components:windows-motion-controls:warn');
 
 /**
  * glTF model loader.
@@ -27,6 +28,7 @@ module.exports.Component = registerComponent('gltf-model', {
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
     }, undefined /* onProgress */, function gltfFailed () {
+      warn('Failed to glTF-model: ' + src);
       el.emit('model-error', {format: 'gltf', src: src});
     });
   },
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -1,7 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
 var utils = require('../utils/');
-var warn = utils.debug('components:windows-motion-controls:warn');
+var warn = utils.debug('components:gltf-model:warn');
 
 /**
  * glTF model loader.
@@ -28,8 +28,9 @@ module.exports.Component = registerComponent('gltf-model', {
       self.model.animations = gltfModel.animations;
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
-    }, undefined /* onProgress */, function gltfFailed () {
-      warn('Failed to glTF-model: ' + src);
+    }, undefined /* onProgress */, function gltfFailed (error) {
+      var message = (error && error.message) ? error.message : 'Failed to load glTF model';
+      warn(message);
       el.emit('model-error', {format: 'gltf', src: src});
     });
   },
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -26,6 +26,8 @@ module.exports.Component = registerComponent('gltf-model', {
       self.model.animations = gltfModel.animations;
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
+    }, undefined /* onProgress */, function gltfFailed () {
+      el.emit('model-error', {format: 'gltf', src: src});
     });
   },
 
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -1,5 +1,6 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var utils = require('../utils/');
 var warn = utils.debug('components:windows-motion-controls:warn');
 
 /**
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -19,22 +19,25 @@ registerComponent('laser-controls', {
     el.setAttribute('windows-motion-controls', {hand: data.hand});
 
     // Wait for controller to connect, or have a valid pointing pose, before creating ray
-    el.addEventListener('controllerconnected', function (evt) { createRay(evt); });
-    el.addEventListener('controllerdisplayready', createRay);
+    el.addEventListener('controllerconnected', createRay);
+    el.addEventListener('controllermodelready', createRay);
 
     function createRay (evt) {
       var controllerConfig = config[evt.detail.name];
 
       if (!controllerConfig) { return; }
 
+      // Show the line unless a particular config opts to hide it, until a controllermodelready
+      // event comes through.
       var raycasterConfig = utils.extend({
         showLine: true
       }, controllerConfig.raycaster || {});
 
-      if (evt.detail.name === 'windows-motion-controls') {
-        var motionControls = el.components['windows-motion-controls'];
-        raycasterConfig = utils.extend(raycasterConfig, motionControls.rayOrigin);
-        raycasterConfig.showLine = motionControls.rayOriginInitialized;
+      // The controllermodelready event contains a rayOrigin that takes into account
+      // offsets specific to the loaded model.
+      if (evt.detail.rayOrigin) {
+        raycasterConfig = utils.extend(raycasterConfig, evt.detail.rayOrigin);
+        raycasterConfig.showLine = true;
       }
 
       el.setAttribute('raycaster', raycasterConfig);
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -36,7 +36,8 @@ registerComponent('laser-controls', {
       // The controllermodelready event contains a rayOrigin that takes into account
       // offsets specific to the loaded model.
       if (evt.detail.rayOrigin) {
-        raycasterConfig = utils.extend(raycasterConfig, evt.detail.rayOrigin);
+        raycasterConfig.origin = evt.detail.rayOrigin.origin;
+        raycasterConfig.direction = evt.detail.rayOrigin.direction;
         raycasterConfig.showLine = true;
       }
 
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -16,21 +16,33 @@ registerComponent('laser-controls', {
     el.setAttribute('gearvr-controls', {hand: data.hand});
     el.setAttribute('oculus-touch-controls', {hand: data.hand});
     el.setAttribute('vive-controls', {hand: data.hand});
+    el.setAttribute('windows-motion-controls', {hand: data.hand});
 
-    // Wait for controller to connect before
-    el.addEventListener('controllerconnected', function (evt) {
+    // Wait for controller to connect, or have a valid pointing pose, before creating ray
+    el.addEventListener('controllerconnected', function (evt) { createRay(evt); });
+    el.addEventListener('controllerdisplayready', createRay);
+
+    function createRay (evt) {
       var controllerConfig = config[evt.detail.name];
 
       if (!controllerConfig) { return; }
 
-      el.setAttribute('raycaster', utils.extend({
+      var raycasterConfig = utils.extend({
         showLine: true
-      }, controllerConfig.raycaster || {}));
+      }, controllerConfig.raycaster || {});
+
+      if (evt.detail.name === 'windows-motion-controls') {
+        var motionControls = el.components['windows-motion-controls'];
+        raycasterConfig = utils.extend(raycasterConfig, motionControls.rayOrigin);
+        raycasterConfig.showLine = motionControls.rayOriginInitialized;
+      }
+
+      el.setAttribute('raycaster', raycasterConfig);
 
       el.setAttribute('cursor', utils.extend({
         fuse: false
       }, controllerConfig.cursor));
-    });
+    }
   },
 
   config: {
@@ -50,6 +62,11 @@ registerComponent('laser-controls', {
 
     'vive-controls': {
       cursor: {downEvents: ['triggerdown'], upEvents: ['triggerup']}
+    },
+
+    'windows-motion-controls': {
+      cursor: {downEvents: ['triggerdown'], upEvents: ['triggerup']},
+      raycaster: {showLine: false}
     }
   }
 });
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -41,7 +41,11 @@ registerComponent('laser-controls', {
         raycasterConfig.showLine = true;
       }
 
-      el.setAttribute('raycaster', raycasterConfig);
+      // Only apply a default raycaster if it does not yet exist. This prevents it overwriting
+      // config applied from a controllermodelready event.
+      if (evt.detail.rayOrigin || !el.hasAttribute('raycaster')) {
+        el.setAttribute('raycaster', raycasterConfig);
+      }
 
       el.setAttribute('cursor', utils.extend({
         fuse: false
--- a/src/components/windows-motion-controls.js
+++ b/src/components/windows-motion-controls.js
@@ -1,3 +1,4 @@
+/* global THREE */
 var bind = require('../utils/bind');
 var registerComponent = require('../core/component').registerComponent;
 var controllerUtils = require('../utils/tracked-controls');
@@ -6,11 +7,11 @@ var utils = require('../utils/');
 var debug = utils.debug('components:windows-motion-controls:debug');
 var warn = utils.debug('components:windows-motion-controls:warn');
 
+var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
+
 // TODO: Point to final glb assets once they are hosted. For now, place them in this hard coded directory.
 var MODEL_BASE_URL = '/examples/assets/models/controllers/wmr/';
-var MODEL_LEFT_FILENAME = 'left.glb';
-var MODEL_RIGHT_FILENAME = 'right.glb';
-var MODEL_UNIVERSAL_FILENAME = 'universal.glb';
+var MODEL_FILENAMES = { left: 'left.glb', right: 'right.glb', default: 'universal.glb' };
 
 var GAMEPAD_ID_PREFIX = 'Spatial Controller (Spatial Interaction Source) ';
 var GAMEPAD_ID_PATTERN = /([0-9a-zA-Z]+-[0-9a-zA-Z]+)$/;
@@ -23,14 +24,17 @@ var GAMEPAD_ID_PATTERN = /([0-9a-zA-Z]+-[0-9a-zA-Z]+)$/;
  */
 module.exports.Component = registerComponent('windows-motion-controls', {
   schema: {
-    hand: {default: 'left'},
+    hand: {default: DEFAULT_HANDEDNESS},
+    // It is possible to have multiple pairs of controllers attached (a pair has both left and right).
+    // Set this to 1 to use a controller from the second pair, 2 from the third pair, etc.
+    pair: {default: 0},
     // If true, loads the controller glTF asset.
     model: {default: true},
     // If true, will hide the model from the scene if no matching gamepad (based on ID & hand) is connected.
     hideDisconnected: {default: true}
   },
 
-  _mapping: {
+  mapping: {
     // A-Frame specific semantic axis names
     axes: {'thumbstick': [0, 1], 'trackpad': [2, 3]},
     // A-Frame specific semantic button names
@@ -77,7 +81,7 @@ module.exports.Component = registerComponent('windows-motion-controls', {
     this.bindMethods();
 
     // Cache for submeshes that we have looked up by name.
-    this._loadedMeshInfo = {
+    this.loadedMeshInfo = {
       buttonMeshes: null,
       axisMeshes: null
     };
@@ -115,7 +119,8 @@ module.exports.Component = registerComponent('windows-motion-controls', {
 
   checkIfControllerPresent: function () {
     this.checkControllerPresentAndSetup(this, GAMEPAD_ID_PREFIX, {
-      hand: this.data.hand
+      hand: this.data.hand,
+      index: this.data.pair
     });
 
     if (this.data.hideDisconnected) {
@@ -156,6 +161,7 @@ module.exports.Component = registerComponent('windows-motion-controls', {
     var controller = trackedControlsComponent ? trackedControlsComponent.controller : null;
     var device = 'default';
     var hand = this.data.hand;
+    var filename;
 
     if (controller) {
       // Read hand directly from the controller, rather than this.data, as in the case that the controller
@@ -170,10 +176,7 @@ module.exports.Component = registerComponent('windows-motion-controls', {
     }
 
     // Hand
-    var filename;
-    if (hand === 'left') filename = MODEL_LEFT_FILENAME;
-    else if (hand === 'right') filename = MODEL_RIGHT_FILENAME;
-    else filename = MODEL_UNIVERSAL_FILENAME;
+    filename = MODEL_FILENAMES[hand] || MODEL_FILENAMES.default;
 
     // Final url
     return MODEL_BASE_URL + device + '/' + filename;
@@ -183,6 +186,7 @@ module.exports.Component = registerComponent('windows-motion-controls', {
     var data = this.data;
     this.el.setAttribute('tracked-controls', {
       idPrefix: GAMEPAD_ID_PREFIX,
+      controller: data.pair,
       hand: data.hand
     });
 
@@ -221,8 +225,11 @@ module.exports.Component = registerComponent('windows-motion-controls', {
 
   onModelLoaded: function (evt) {
     var controllerObject3D = evt.detail.model;
-    var loadedMeshInfo = this._loadedMeshInfo;
+    var loadedMeshInfo = this.loadedMeshInfo;
     var i;
+    var meshName;
+    var mesh;
+    var meshInfo;
 
     debug('Processing model');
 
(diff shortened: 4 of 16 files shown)


tests/components/hand-controls.test.js
```diff
@@ -41,6 +41,23 @@ suite('hand-controls', function () {
...
+      component.pressedButtons['BorY'] = false;
+      component.pressedButtons['surface'] = false;
+      assert.equal(component.determineGesture(), 'Point');
+    });
+
+    test('makes point gesture on vive', function () {
+      var trackedControls;
+      el.setAttribute('tracked-controls', '');
+      trackedControls = el.components['tracked-controls'];
+      trackedControls.controller = {id: 'OpenVR Vive', connected: true};
```
machenmusik: Is `OpenVR Vive` really how the controllers are appearing these days?

machenmusik:
Nice - I had actually done a version relying on `tracked-controls` changes from #2583 which are similar to those you made here, but I didn't have actual controller models.  Wonder if it makes sense to merge them

src/components/gltf-model.js
```diff
@@ -26,6 +26,8 @@ module.exports.Component = registerComponent('gltf-model', {
       self.model.animations = gltfModel.animations;
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
+    }, undefined /* onProgress */, function gltfFailed () {
+      el.emit('model-error', {format: 'gltf', src: src});
```
dmarcos: We probably should throw an error instead of an event? An similar scenario can be found in the `text` component: https://github.com/aframevr/aframe/blob/master/src/components/text.js#L391

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+module.exports.Component = registerComponent('windows-motion-controls', {
+  schema: {
+    hand: {default: 'left'},
+    // If true, loads the controller glTF asset.
+    model: {default: true},
+    // If true, will hide the model from the scene if no matching gamepad (based on ID & hand) is connected.
+    hideDisconnected: {default: true}
+  },
+
+  _mapping: {
```
dmarcos: We don't use underscore to indicate private variables. Can we keep just `mapping`?

tests/components/hand-controls.test.js
dmarcos: In Firefox is `OpenVR Gamapad`

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+    this.onButtonUp = function (evt) { self.onButtonEvent(evt, 'up'); };
+    this.onButtonTouchStart = function (evt) { self.onButtonEvent(evt, 'touchstart'); };
+    this.onButtonTouchEnd = function (evt) { self.onButtonEvent(evt, 'touchend'); };
+    this.controllerPresent = false;
+    this.lastControllerCheck = 0;
+    this.previousButtonValues = {};
+    this.bindMethods();
+
+    // Cache for submeshes that we have looked up by name.
+    this._loadedMeshInfo = {
```
dmarcos: no underscore on variable name

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
+var bind = require('../utils/bind');
+var registerComponent = require('../core/component').registerComponent;
+var controllerUtils = require('../utils/tracked-controls');
+var utils = require('../utils/');
+
+var debug = utils.debug('components:windows-motion-controls:debug');
+var warn = utils.debug('components:windows-motion-controls:warn');
+
+// TODO: Point to final glb assets once they are hosted. For now, place them in this hard coded directory.
+var MODEL_BASE_URL = '/examples/assets/models/controllers/wmr/';
```
dmarcos: As you pointed out we should host the models in the CDN https://github.com/aframevr/assets/tree/gh-pages/controllers

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+      hand = controller.hand;
+
+      if (!forceDefault) {
+        var match = controller.id.match(GAMEPAD_ID_PATTERN);
+        device = ((match && match[0]) || device);
+      }
+    }
+
+    // Hand
+    var filename;
```
dmarcos: define variables at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+    if (hand === 'left') filename = MODEL_LEFT_FILENAME;
```
dmarcos: curly braces around if clause `if (hand === 'left') {filename = MODEL_LEFT_FILENAME;}`

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+var registerComponent = require('../core/component').registerComponent;
+var controllerUtils = require('../utils/tracked-controls');
+var utils = require('../utils/');
+
+var debug = utils.debug('components:windows-motion-controls:debug');
+var warn = utils.debug('components:windows-motion-controls:warn');
+
+// TODO: Point to final glb assets once they are hosted. For now, place them in this hard coded directory.
+var MODEL_BASE_URL = '/examples/assets/models/controllers/wmr/';
+var MODEL_LEFT_FILENAME = 'left.glb';
```
dmarcos: it might be better to consolidate names in an object:
````javascript
var MODEL_FILENAMES = { left: 'left.glb`, right: 'right.glb', default: 'universal.glb' }
````

File: src/components/windows-motion-controls.js (hunk 7 above)
Comment (dmarcos) [inline src/components/windows-motion-controls.js]: if we consolidate the filenames in an object as described above we can consolidate this if statement into:
````javascript
filename = MODEL_FILES[hand] || MODEL_FILES.default
````

File: src/components/windows-motion-controls.js [hunk 9]
```diff
@@ -0,0 +1,382 @@
...
+    this.el.sceneEl.removeEventListener('controllersupdated', this.onControllersUpdate, false);
+  },
+
+  onControllersUpdate: function () {
+    this.checkIfControllerPresent();
+  },
+
+  onModelError: function (evt) {
+    var defaultUrl = this.createControllerModelUrl(true);
+    if (evt.detail.src !== defaultUrl) {
```
dmarcos: Do we need a default model when we already have hand and left ones? If another component like `hand-controls` decides to use a custom model (hence the model property === false) is responsibility of that higher level component to handle model errors. I might be misunderstanding something though

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+
+    // Reset the caches
+    loadedMeshInfo.buttonMeshes = {};
+    loadedMeshInfo.axisMeshes = {};
+
+    // Cache our meshes so we aren't traversing the hierarchy per frame
+    if (rootNode) {
+      // Button Meshes
+      for (i = 0; i < this._mapping.buttons.length; i++) {
+        var buttonMeshName = this._mapping.buttonMeshNames[this._mapping.buttons[i]];
```
dmarcos: variable defined at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+        if (!buttonMeshName) {
+          debug('Skipping unknown button at index: ' + i + ' with mapped name: ' + this._mapping.buttons[i]);
+          continue;
+        }
+
+        var buttonMesh = rootNode.getObjectByName(buttonMeshName);
```
dmarcos: variable defined at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+        if (!buttonMesh) {
+          warn('Missing button mesh with name: ' + buttonMeshName);
+          continue;
+        }
+
+        var buttonMeshInfo = {
```
dmarcos: variable defined at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+            '(VALUE: ' + !!buttonMeshInfo.value +
+            ', PRESSED: ' + !!buttonMeshInfo.pressed +
+            ', UNPRESSED:' + !!buttonMeshInfo.unpressed +
+            ')');
+        }
+      }
+
+      // Axis Meshes
+      for (i = 0; i < this._mapping.axisMeshNames.length; i++) {
+        var axisMeshName = this._mapping.axisMeshNames[i];
```
dmarcos: variable at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+        if (!axisMeshName) {
+          debug('Skipping unknown axis at index: ' + i);
+          continue;
+        }
+
+        var axisMesh = rootNode.getObjectByName(axisMeshName);
```
dmarcos: variable at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+        if (!axisMesh) {
+          warn('Missing axis mesh with name: ' + axisMeshName);
+          continue;
+        }
+
+        var axisMeshInfo = {
```
dmarcos: variable at the top

src/components/windows-motion-controls.js
```diff
@@ -0,0 +1,382 @@
...
+    var axisMeshInfo = this._loadedMeshInfo.axisMeshes[axis];
+    if (!axisMeshInfo) return;
+
+    var min = axisMeshInfo.min;
+    var max = axisMeshInfo.max;
+    var target = axisMeshInfo.value;
+
+    // Convert from gamepad value range (-1 to +1) to lerp range (0 to 1)
+    var lerpValue = axisValue * 0.5 + 0.5;
+    target.setRotationFromQuaternion(min.quaternion.clone().slerp(max.quaternion, lerpValue));
```
dmarcos: To appease the garbage collector we should avoid allocating new variables and reuse instead as much as we can. We usually use a closure pattern for temporary variables that don't hold state (https://github.com/aframevr/aframe/blob/master/src/components/cursor.js#L168). In this case I think we could do:
````javascript
lerpAxisTransform: (function () {
 var quaternion = new THREE.Quaternion();
 return function (axis, axisValue) {
   ....
      target.setRotationFromQuaternion(quaternion.copy(min.quaternion).slerp(max.quaternion, lerpValue));
   ....
 }
});

File: src/components/windows-motion-controls.js [hunk 17]
```diff
@@ -0,0 +1,382 @@
...
+
+  lerpButtonTransform: function (buttonName, buttonValue) {
+    var buttonMeshInfo = this._loadedMeshInfo.buttonMeshes[buttonName];
+    if (!buttonMeshInfo) return;
+
+    var min = buttonMeshInfo.unpressed;
+    var max = buttonMeshInfo.pressed;
+    var target = buttonMeshInfo.value;
+
+    target.setRotationFromQuaternion(min.quaternion.clone().slerp(max.quaternion, buttonValue));
```
dmarcos: Same as above. Try to avoid allocations and reuse

src/components/gltf-model.js
leweaver: Could do both? I listen to the error message to try loading a different (fallback) asset file, I think it could be useful in other contexts for the caller to know that the file load failed.

tests/components/hand-controls.test.js
leweaver: Makes sense to use a real value, I'll update to that.

src/components/windows-motion-controls.js
leweaver: This logic only applies in the case where the model property === true, and is a separate concept to the default hand... our device ID's have a suffix which represents a particular hardware form factor (as exposed by the underlying Windows Mixed Reality API). 
`Spatial Controller (Spatial Interaction Source) 045E-065A`

While there is currently a single device (the Motion Controller, 045E-065A), the underlying windows API supports addition of new hardware form factors. As such, it is possible that new hardware exists (with a new ID suffix) before we upload a corresponding asset - which is why we fallback to a 'default' GLB file.

Hopefully I understand your comment correctly?

dmarcos: I see. I have not seen the default model. How does it look like?

src/utils/tracked-controls.js
```diff
@@ -46,41 +46,81 @@ module.exports.checkControllerPresentAndSetup = function (component, idPrefix, q
...
+function findMatchingController (controllers, filterIdExact, filterIdPrefix, filterHand, targetControllerNumber) {
+  var controller;
+  var i;
+  var matchingControllerOccurence = 0;
+  targetControllerNumber = targetControllerNumber || 0;
+
+  for (i = 0; i < controllers.length; i++) {
+    controller = controllers[i];
+    // Determine if the controller ID matches our criteria
+    if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) continue;
```
dmarcos: curly brace around if statement:
````javascript
if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) { continue; }
if (!filterIdPrefix && controller.id !== filterIdExact) { continue; }
````

File: src/utils/tracked-controls.js [hunk 19]
```diff
@@ -46,41 +46,81 @@ module.exports.checkControllerPresentAndSetup = function (component, idPrefix, q
...
+      return controller;
     }
-    if (isPresent) { break; }
-    // Update count of gamepads with idPrefix.
-    if (isPrefixMatch) { index++; }
+    ++matchingControllerOccurence;
   }
-
-  return isPresent;
+  return null;
```
dmarcos: probably better `return undefined` for consistency with other functions across the code base

src/components/windows-motion-controls.js
machenmusik: in other cases we have used idPrefix e.g. if they will all be "Spatial Controller " but then all bets are off.  I think the version I did uses that strategy, let me find where I put that code and put it somewhere for reference

src/components/windows-motion-controls.js
machenmusik: For clarity, to get the models in the CDN, I believe one does a pull request against aframevr/assets with the appropriate changes.

dmarcos: Correct! Just a PR against the `gh-pages` branch on https://github.com/aframevr/assets/tree/gh-pages/controllers

src/components/windows-motion-controls.js
machenmusik: my personal preference is to have them as unpublicized properties on the component instance, so that other functions can reuse them, but yes I think the vast majority of code uses closure form

src/components/windows-motion-controls.js
dmarcos: `gh-pages` so they get deployed on the CDN

machenmusik: not unless something changed?  for example this is vive tracker https://github.com/aframevr/assets/pull/21

machenmusik: specific example for reference, this was vive-tracker: https://github.com/aframevr/assets/pull/21

src/components/windows-motion-controls.js
dmarcos: Two things:
1. If it's just a temporary value is better to use a closure. By having the variable on the closure you share one single variable across all the object instances.
2. In the `this`. it pollutes the object and it's tempting to use it to store state that will be destroyed if multiple methods use the variable for partial calculations.
3. If those variables are not longer useful they will be hanging from the `this`. it's easy to forget and leave them there orphan.

src/components/windows-motion-controls.js
leweaver: As an alternative, what if there is an additional http request prior to model download that retrieves a manifest file with mappings of suffix -> filename, with an entry for default?

My main motivator here is that we want to avoid making changes to the core library if we need to add new, radically different, models - since people don't update library versions after releasing their apps.

src/utils/tracked-controls.js
```diff
@@ -46,41 +46,81 @@ module.exports.checkControllerPresentAndSetup = function (component, idPrefix, q
...
+  var matchingControllerOccurence = 0;
+  targetControllerNumber = targetControllerNumber || 0;
+
+  for (i = 0; i < controllers.length; i++) {
+    controller = controllers[i];
+    // Determine if the controller ID matches our criteria
+    if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) continue;
+    if (!filterIdPrefix && controller.id !== filterIdExact) continue;
+
+    if (filterHand) {
```
dmarcos: Maybe simpler like this? And maybe more accurate if there are several controllers with the same handedness. I think it cannot happen today with the available hardware but who knows if one day 2 players can play in the same machine with two pairs of controllers
````javascript
// If the hand filter and controller handedness are defined we compare them.
if (filterHand && controller.hand && filterHand !== controller.hand) { continue; }
````

Comment (dmarcos) [reply src/utils/tracked-controls.js]: Maybe also more consistent with the negative statements above

Comment (leweaver) [reply src/utils/tracked-controls.js]: Good point re: multiple pairs, will make appropriate changes and add tests.

File: src/components/windows-motion-controls.js (hunk 16 above)
Comment (machenmusik) [reply src/components/windows-motion-controls.js]: Completely agree with 2 and 3.
On 1, I thought we ran into past issues with multiple instances / invocations there, but may be misremembering.

File: src/utils/tracked-controls.js (hunk 20 above)
Comment (machenmusik) [reply src/utils/tracked-controls.js]: from a little experience - undefined controller hand means none, but undefined filterHand means don't care, which is a completely different thing.  See #2583

Comment (machenmusik) [reply src/utils/tracked-controls.js]: you want to be able to specify a filter for no-handedness (vive-tracker), as well as for specific handedness (various handed controllers), as well as don't care (don't care whether the 3DOF controller is left/right/none, just give me the pose)

Comment (leweaver) [reply src/utils/tracked-controls.js]: It looks like your first and third scenarios can be handled in the same way, by simply passing in filterHand = undefined, and optionally using the targetControllerNumber to get a controller at specific index

Comment (machenmusik) [reply src/utils/tracked-controls.js]: no - filterHand undefined (default) means "I don't care whether controller.hand exists, much less what its value is".  That cannot be used for the first case (no-handedness), where the test must be performed, and must only succeed when controller.hand is undefined.

**Comment by machenmusik** [comment]
FYI, working implementation on top of #2583 is here:
https://github.com/chenzlabs/aframe/tree/support-spatial-controller

File: src/components/gltf-model.js (hunk 2 above)
Comment (dmarcos) [reply src/components/gltf-model.js]: yes, we could have both

Comment (donmccurdy) [reply src/components/gltf-model.js]: +1 for both. The errors should be helpful for developers debugging, who may not know to listen for the event.

But my inclination would be to log an error or warning, through the debug util, rather than throwing something that can't be caught. See [`obj-model`](https://github.com/aframevr/aframe/blob/master/src/components/obj-model.js#L5).

File: src/components/windows-motion-controls.js (hunk 9 above)
Comment (dmarcos) [reply src/components/windows-motion-controls.js]: I see. About the current code approach. If new hardware is released, Is a default model better than loading the 045E-065A one? Neither of them will be representative of the new hardware. If a new released controller is radically different I expect all the mappings and buttons to be differerent. In that case, we might better have a separate new component for it. The idea of the manifest is indeed intriging. How could we make sure that the developer has control over the experience? In the demos we have done so far that work across Oculus and Vive we had to make a lot of adjustments for each of the controllers: different button number, physical distribution and capabilities (capacitive, analog, axis...), haptics, button tooltip positioning, raycasting origin, location of menus and UI that are hand locked. It worked well for us to be explicit: this is the list of controllers my app is supporting and I know their capabilities, button mapping and physical appearance. If a new controller is released I will probably want to do some integration to make sure my interaction model does not break. It might be confusing if a single component can map to multiple physical controllers / models and a dev cannot make any assumptions.

File: src/components/gltf-model.js (hunk 2 above)
Comment (dmarcos) [reply src/components/gltf-model.js]: Cool! Let's add both then 👌

File: src/components/windows-motion-controls.js (hunk 9 above)
Comment (machenmusik) [reply src/components/windows-motion-controls.js]: if the functionality is same but appearance may be different, it would be better to have a default model than to throw up hands just because we're not sure exactly what this one looks like.  remember before we had real Touch controller model and we just used Vive so you could at least see what you were doing?  same for Daydream and GearVR, and Vive tracker...  That being said, we might just want to reuse existing model as the generic model.

Comment (machenmusik) [reply src/components/windows-motion-controls.js]: if functionality is actually different, then really gamepad id string should be different enough to distinguish (not just by cryptic suffix)

Comment (dmarcos) [reply src/components/windows-motion-controls.js]: I hope I'm not confusing things too much 🙂. Summing up. Based on experience, I prefer each controller component to map to one and just one physical device (with its corresponding model). As opposed to traditional games -where you don't usually see a gamepad in game- in VR the interaction model is built around the physical appearence of the controller and there are more things you want to control besides button mapping into actions.

Comment (dmarcos) [reply src/components/windows-motion-controls.js]: I'm open to evolve the current API in the future but it's probably beyond 0.7.0 scope

Comment (machenmusik) [reply src/components/windows-motion-controls.js]: yeah, I think we've been slowly evolving the controller components to use common pieces (buttons, axes, mapping, utils functions) but haven't gotten as far as model manipulations, explicit refactor may be pre-1.0 but probably not this coming release.

tactically speaking, I'd like to consider merging #2583 first, as some of the mods in this one are expanded to cover the interesting challenges the tracker poses, and rebasing this on top

Comment (machenmusik) [reply src/components/windows-motion-controls.js]: >  Based on experience, I prefer each controller component to map to one and just one physical device (with its corresponding model).

Tactically speaking, maybe it could be as simple as a generic component with the basic functionality (I had called mine `spatial-controls` along prior naming convention lines) matching on idPrefix, and then specific components that matched more specific gamepad id strings with model mappings / manipulations?

Comment (dmarcos) [reply src/components/windows-motion-controls.js]: Let's merge this first since it's higher priority than tracker support for 0.7.0

Comment (machenmusik) [reply src/components/windows-motion-controls.js]: I'd agree, this is more important, except we can't merge without the models, can we?
I think #2583 is ready to merge now.
If we have a little time, I'd rebase a fork off #2583 and PR back.  I expect the delta is not large

Comment (dmarcos) [reply src/components/windows-motion-controls.js]: I've been working on a separare PR for tracker support to implement a spectator camera. I prefer to keep https://github.com/aframevr/aframe/pull/2583 on hold and also not slow this one down.

Comment (machenmusik) [reply src/components/windows-motion-controls.js]: OK - would be happy to have this contribution from Microsoft - but if you will be doing tracker support, you may need to revisit https://github.com/aframevr/aframe/pull/3013#discussion_r136470194 -- the current changes in this PR make it impossible to match no-handedness

File: src/utils/tracked-controls.js (hunk 20 above)
Comment (dmarcos) [reply src/utils/tracked-controls.js]: @machenmusik not sure I fully understand. Can you suggest an alternative if statement that covers the case you're referring?

Comment (machenmusik) [reply src/utils/tracked-controls.js]: https://github.com/aframevr/aframe/pull/2583/files#r136465173 is a bit ugly but working.
- instead of relying on schema hand undefined (which is a little ambiguous about no-handedness vs. don't-care as discussed here), use explicit strings `any-or-none` and `none` to specify hand filter

Comment (machenmusik) [reply src/utils/tracked-controls.js]: I think the utils function change is https://github.com/aframevr/aframe/pull/2583/files#diff-24b039fab259237004f42c5bf3856cdfR69 -- use DEFAULT_HANDEDNESS to replace controller hand of undefined with an actual string to match

Comment (machenmusik) [reply src/utils/tracked-controls.js]: but actually I think `DEFAULT_HANDEDNESS` was supposed to be `none` there

Comment (dmarcos) [reply src/utils/tracked-controls.js]: We can revisit this later for tracker support.

Comment (machenmusik) [reply src/utils/tracked-controls.js]: generally though, something like:
```
// Only try to match filterHand if a filter value is specified.
if (filterHand) {
  if ((controller.hand ? controller.hand : DEFAULT_HANDEDNESS) !== filterHand) { continue; }
}
```

Comment (machenmusik) [reply src/utils/tracked-controls.js]: If the filter does not reject based on hand value, I believe we want to fall through against controller index check (if any) so you can match against the second set of left and right hand controllers, or the second or third trackers.

File: src/components/gltf-model.js (hunk 2 above)
Comment (leweaver) [reply src/components/gltf-model.js]: Will do both, thanks!

File: src/utils/tracked-controls.js [hunk 21]
```diff
@@ -78,46 +72,46 @@ module.exports.isControllerPresent = isControllerPresent;
...
-      }
-      // If we reach here, the controller is unhanded - check against targetControllerNumber
+    if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) { continue; }
+    if (!filterIdPrefix && controller.id !== filterIdExact) { continue; }
+
+    // If the hand filter and controller handedness are defined we compare them.
+    if (filterHand && controller.hand && filterHand !== controller.hand) { continue; }
+
+    // If we have detected an unhanded controller and the component was asking for a particular hand,
+    // we need to treat the controllers in the array as pairs of controllers. This effectively means that we
```
machenmusik: I am not sure this is needed or desirable; originally there were issues with browsers not uniformly exposing handedness, but that is gone now, so if the filter is requesting a hand, we should require one.

Still need coverage for the case where we want to match no-handedness, my suggestion is to use the string `none` such as the following:
```
if (filterHand && (filterHand !== (controller.hand ? controller.hand : 'none'))) { continue; }
```

dmarcos: Why do we need `none`? In the absence of a filterHand the condition is a pass through.

machenmusik: For example - trackers do not have handedness.  Also, if you want to use a third vive controller as spectator camera, my understanding is that the first two get hands, then the rest are no-handed (hand undefined) -- maybe that has changed.

machenmusik: (This bit is inside tracked-controls utils that gets used for everything with a pose, not just within the windows-motion-controls component)

dmarcos: Let's deal with the problem at hand and we will accommodate those no handedness / 3rd control edge cases when we integrate the trackers.

machenmusik: right, as I said, I'm fine with that, this comment is really more about the flagged code portion:
```
// If we have detected an unhanded controller and the component was asking for a particular hand,
// we need to treat the controllers in the array as pairs of controllers.
```
I don't think we need to do that, nor should we.
If filterHand is supplied, it should be used.  e.g. hand left, index 1 (to get the second left hand controller, instead of 0 for the first)

dmarcos: Yeah I agree this condition is a bit cryptic. The case is trying to cover is if there are more than 2 controllers connected the  filterControllerIndex will control which pair in the list is selected as left / right. I do not think this is a possible scenario on any platform today. I'm happy to let `left` === 0 and `right` === 1 if the `controller.hand` is not supplied.

dmarcos: These kind of details taken care for you is one of the reasons why people should be using A-Frame 😄 All those edge cases are not fun to deal with.

machenmusik: But that is not correct behavior.  Unlike Vive controllers, the Windows motion controllers are handed, more like Oculus Touch.  So handedness is not a function of cardinality

machenmusik: developer.microsoft.com/en-us/windows/mixed-reality/motion_controllers

Look closely at the picture.

machenmusik: (Microsoft Edge does in fact provide the hand property for appropriate gamepads/controllers.)

dmarcos: It's ok. If there's always a hand the filterHand is always supplied by `windows-motion-controls` so you won't reach this statement https://github.com/aframevr/aframe/pull/3013/files/638c6a2a2fad503ba67f291bc8a1d7b8df3e05d5#diff-bc91e5db13c01778a4dfb59ce1fd3f05R27

machenmusik: I seem to recall that the original mapping for Vive, before browsers exposed hand, was actually the opposite... 0 right, 1 left.  Anyway, now that browsers do the right thing, if the Windows motion controls behave that way, I'm not sure this index mapping should be applied more generally in tracked-controls used by all, versus in the windows-motion-controls component... since the convention may apparently vary by controller type.

machenmusik: If the filterHand is always supplied, why do we need this code snippet at all?  Now I am confused...

machenmusik: Meaning, the way the current code is designed, hand applies before index, so you ask for the second left handed controller with filter of hand=left and index=1 (since they start at zero).  No need to worry about whether pairs or not.

dmarcos: Vive handedness is determined at runtime based on the relative position to the head. I believe there's still a chance of `controller.hand` to come undefined.

machenmusik: Right, but if you specified that you wanted a handed controller, shouldn't you only get it when the system agrees that it is left hand?  Otherwise you will end up with hands changing abruptly for no reason

machenmusik: Thanks for indulging the long discussion.  My sense is that the windows-motion-controls PR can handle left hand, right hand, and multiple pairs without the changes to tracked-controls utils (which have been tested to work properly across Vive, Oculus touch, daydream and gearvr).  If we want to change behavior due to Vive indeterminate hand, I suggest we do that in separate PR

leweaver: Now I am starting to get confused too!

If, as you say, all browsers are now exposing the hand property correctly - it is very tempting to update the vive controls class to use that. We do still need to consider the 'unhanded case' though as 
the underlying windows API that drives the Motion Controllers support in Edge's WebVR does support unhanded controllers, even though there aren't any on the market at this time. 
(The current Motion Controllers are, as you observed, handed). If our platform exposes an unhanded controller, the gamepad 'hand' property will be an empty string: ''.

The reason I have the (yes, cryptic) snippet is that the underlying platform also supports more than 2 unhanded controls connected at once. So in theory the platform could tell Edge that the user has 4 or more unhanded controllers connected. The question is - do we want these to work with the hand-controls component? If so, we need them to work with a component that has requested a particular hand.

If so, we need to keep the logic here that puts unhanded controllers into pairs of left and right, based on their index.

Regarding the number of connected handed controllers, it is easy to connect 4 handed controllers today using two physical, and two simulated via the windows mixed reality simulator.

dmarcos: @machenmusik Do you think the code we have here is reasonable even tough there might be an edge case or two we're missing? Can we go with this and refine when integrating the Vive trackers and test on different browsers if our assumptions about `handeness` still apply?

machenmusik: @leweaver When doing #2583 I did in fact switch over vive-controls to use handedness, no more weird index stuff.

@dmarcos When doing #2583, I had tested against all supported controllers as well as vive-trackers, which is why I am personally more comfortable with the changes to tracked-controls utils there.  (Although it has been so long that other changes may have crept in that break it now, I suppose.)

On unhanded controllers - it sounds as though there will be WinMR equivalents of vive-trackers, in which case the main changes in #2583 may prove useful as well, namely handling no-handedness (although it sounds as though we now have empty string to contend with as well as undefined) and also adding full rotation offset to tracked-controls, not just Z (because the native orientation of vive-tracker is somewhat different from canonical model).

I am happy to welcome @leweaver to join the A-Frame project and thus take over WinMR controller support, just trying to share learnings from doing the last few.  Feel free to take whatever is worthwhile from #2583 and the spatial-controls branch I had done, if it is helpful.

As for whether unhandled controllers should be allowed to match handed requests, that's a bit of a philosophical question.  The dynamic Vive wand case is simpler, as the controllers will in all likelihood gain handedness eventually.  In the WinMR and Vive tracker case, it's less clear, although I would expect unhandled WinMR controllers to have different id strings and thus be handled by a related but separate component.

dmarcos: Ok, let's keep this part of the PR as is and improve it later in follow up PRs

machenmusik: > If so, we need to keep the logic here that puts unhanded controllers into pairs of left and right, based on their index.

The problem I see with this is that all unhandled controllers may not be created equal.  Let's assume that unhandled controllers include totems that can be used in either hand (with no dynamic hand assignment), and also other things like gun controllers as are popular with PSVR.  The user has limited or no control over enumerated order.  What if the enumerated order is totem, gun, totem?  Then you have two left-handed totems and a gun for a right hand, which may not be the intended outcome...

machenmusik: > If so, we need to keep the logic here that puts unhanded controllers into pairs of left and right, based on their index.

The problem I see with this is that all unhandled controllers may not be created equal.  Let's assume that unhandled controllers include totems that can be used in either hand (with no dynamic hand assignment), and also other things like gun controllers as are popular with PSVR.  The user has limited or no control over enumerated order.  What if the enumerated order is totem, gun, totem?  Then you have two left-handed totems and a gun for a right hand, which may not be the intended outcome...

machenmusik: > If so, we need to keep the logic here that puts unhanded controllers into pairs of left and right, based on their index.

The problem I see with this is that all unhandled controllers may not be created equal.  Let's assume that unhandled controllers include totems that can be used in either hand (with no dynamic hand assignment), and also other things like gun controllers as are popular with PSVR.  The user has limited or no control over enumerated order.  What if the enumerated order is totem, gun, totem?  Then you have two left-handed totems and a gun for a right hand, which may not be the intended outcome...

dmarcos:
I missed one important thing. We should incorporate the `windows mixed reality controlers` into `laser-controls` too which are the A-Frame cross device universal controls (3DOF and 6DOF). It should be pretty straight forward. Just add the controllers to the list: 

https://github.com/aframevr/aframe/blob/master/src/components/laser-controls.js#L18

And provide the appropriate event mapping and offset for the raycaster (so it's casted from the proper position of the controller):

https://github.com/aframevr/aframe/blob/master/src/components/laser-controls.js#L46

dmarcos:
Thanks for the patience. This is a big patch with a lot to discuss.

dmarcos:
We're pretty close though

leweaver:
Oh yes - I will add to the laser controls class and perform testing. Let me know if you are happy for me to update the PR with that additional commit (and the other minor resolved issues we have discussed) now, or wait.

dmarcos:
Yes please, add the commit to this PR

leweaver:
I've updated the PR to add laser-controls support. I would have preferred to do it without emitting the new 'controllerdisplayready' event; however I couldn't get the line-component to update the start/end positions once they have been set once. Have you encountered that?

Note - that the offsets and rotations default to zero/forward; but can be overridden by the data from the glTF model file (we store the offsets on specifically named nodes in the tree, similar to how we do the button transforms).

src/components/gltf-model.js
```diff
@@ -1,5 +1,7 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var utils = require('../utils/');
+var warn = utils.debug('components:windows-motion-controls:warn');
```
donmccurdy: This should be:

```js
var warn = utils.debug('components:gltf-model:warn');
```

...since the warning will appear for an glTF model.

src/components/gltf-model.js
```diff
@@ -26,6 +28,9 @@ module.exports.Component = registerComponent('gltf-model', {
       self.model.animations = gltfModel.animations;
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
+    }, undefined /* onProgress */, function gltfFailed () {
+      warn('Failed to glTF-model: ' + src);
```
donmccurdy: The gltfFailed callback will receive an argument, `error` — IMO the warning and event should use that message, as it's probably more informative than the URL. Suggested:

```js
var message = error ? error.message : 'Failed to load glTF model';
warn(message);
el.emit('model-error', {format: 'gltf', message: message});

```

machenmusik:
Hi folks, where are we on this one?  Going back to the original description of changes, it seems as though the priority order is this...

> Changes proposed:

> Windows Motion Controller support

This we definitely want ASAP, for the currently publicized handed controllers at least. I think that only a small subset of this PR is needed for that.

> Consolidation of tracked-controls duplicated logic to a single utility function

I think much of this PR is for this consolidation actually.  Do we need it?  We have been talking about refactoring, this could be first step.

> Added error callback on glTF-model component

Nice to have.

docs/components/windows-motion-controls.md
```diff
@@ -0,0 +1,60 @@
...
+| gripup                 | Grip button released.                           |
+| gripchanged            | Grip button changed.                            |
+| menudown               | Menu button pressed.                            |
+| menuup                 | Menu button released.                           |
+| menuchanged            | Menu button changed.                            |
+| trackpaddown           | Trackpad pressed.                               |
+| trackpadup             | Trackpad released.                              |
+| trackpadchanged        | Trackpad button changed.                        |
+| trackpadmoved          | Trackpad axis moved.                            |
+| controllerdisplayready | The model file is loaded and completed parsing. |
```
dmarcos: Maybe `controllermodelready` is a better name?

src/components/laser-controls.js
```diff
@@ -16,21 +16,33 @@ registerComponent('laser-controls', {
...
 
-      el.setAttribute('raycaster', utils.extend({
+      var raycasterConfig = utils.extend({
         showLine: true
-      }, controllerConfig.raycaster || {}));
+      }, controllerConfig.raycaster || {});
+
+      if (evt.detail.name === 'windows-motion-controls') {
+        var motionControls = el.components['windows-motion-controls'];
+        raycasterConfig = utils.extend(raycasterConfig, motionControls.rayOrigin);
```
dmarcos: Can we pass the `rayOrigin` in the `evt.detail` object? This way we don't have to check for `windows-motion-controls` We should maybe move all the controllers to use a `controllermodelready` event for consistency. Maybe better in a separate PR.

dmarcos:
This is pretty close! I think some rethinking refactoring is due to the controllers API after 0.7.0 release. The remaining changes If I have not missed anything are:
https://github.com/aframevr/aframe/pull/3013/files#r137131772
https://github.com/aframevr/aframe/pull/3013/files#r136705108
https://github.com/aframevr/aframe/pull/3013/files#r136705073
https://github.com/aframevr/aframe/pull/3013/files#r137129564

src/components/laser-controls.js
leweaver: I like the pattern of passing the ray details in the event, will modify to follow that pattern. Happy to also modify the others, just let me know if you would prefer that in a separate PR once this is completed.

dmarcos: Thanks! We can do it in a separate PR so we move things along.

leweaver:
Thanks all for the reviewing! I've pushed another revision that should address those 4 changes.

src/components/laser-controls.js
```diff
@@ -16,21 +16,36 @@ registerComponent('laser-controls', {
...
+      // event comes through.
+      var raycasterConfig = utils.extend({
         showLine: true
-      }, controllerConfig.raycaster || {}));
+      }, controllerConfig.raycaster || {});
+
+      // The controllermodelready event contains a rayOrigin that takes into account
+      // offsets specific to the loaded model.
+      if (evt.detail.rayOrigin) {
+        raycasterConfig = utils.extend(raycasterConfig, evt.detail.rayOrigin);
```
dmarcos: is it clear to do? 
````javascript
raycasterConfig.reayOrigin = evt.detail.rayOrigin;
````

Comment (leweaver) [reply src/components/laser-controls.js]: you'd need to do something like..
```
raycasterConfig.origin = evt.detail.rayOrigin.origin
raycasterConfig.direction = evt.detail.rayOrigin.direction
```

but we expand that out if preferred?

dmarcos: yeah, probably better to make it explicit to avoid surprises. nothing hidden 🙂

leweaver:
@dmarcos  - Something I have noticed is that the controllers are quite dark in some scenes. It would be nice to have a high-intensity directional light (position: 0, 1, 0) that only lights the controller, not the rest of the scene. I can't see an obvious way to do this, have any tricks?

dmarcos:
@leweaver Is there anything that can be done at the model / materials level? Adding a light  to the controller is a possibility but maybe too prescriptive and confusing when authors want to have control of the visual appearance of the scene. If nothing can be done at the material level we maybe consider add a light property to the `windows-motion-controls` component that it's `false` by default. This way devs that want to turn the lights on can opt into it.

dmarcos:
I believe this is the last pending change? https://github.com/aframevr/aframe/pull/3013/files#r137319556

leweaver:
That (which I'll get done now) - and also getting controller assets uploaded to the aframevr/assets repo so that we can modify the URL at the top of the file: [windows-motion-controls.js#L12
](https://github.com/leweaver/aframe/blob/c5df51e70dff4f830ee2bd9a6d2ebc0df39e32cf/src/components/windows-motion-controls.js#L12
) - I put a note this morning on aframevr/assets#1 related to this.

leweaver:
I believe I am finished with my changes to this PR. Do I also need to also include the built files under dist/?

dmarcos:
Thanks! 🥇 No need to include the built files.

dmarcos:
@leweaver Will you open a separate PR to point the models URL whenever we merge them on the `assets` repo?

dmarcos:
Thank you!

leweaver:
Sorry forgot to hit enter on this comment box! As you probably noticed, right now this PR version is pointing to: https://cdn.aframe.io/controllers/microsoft/ - I have just submitted https://github.com/aframevr/assets/pull/29 so the URL should be valid soon.

Thanks very much for merging in!

PR #3014, opened 2017-09-01
Support and tests for vrdisplaypointerrestricted events to the a-scene…

olga-microsoft:
…element with the possibility to disable the behavior (on by default)

**Description:**
Support mouse input in WebVR by handling vrdisplaypointerrestricted and vrdisplaypointerunrestricted events in MS Edge. It is by design that mozPointerLock* and webkitPointerLock* implementations are not included.

**Changes proposed:**
- In response to vrdisplaypointerrestricted perform requestPointerLock on the canvas element that is used for WebVR rendering
- In response to vrdisplaypointerunrestricted perform document.exitPointerLock if the locked element is the canvas that is used for WebVR rendering
- Give the developer an option to turn off this behavior by calling disableCanvasPointerLock on the scene object.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -31,7 +31,7 @@ var warn = utils.debug('core:a-scene:warn');
  * @member {object} object3D - Root three.js Scene object.
  * @member {object} renderer
  * @member {bool} renderStarted
- * @member (object) effect - three.js VREffect
+ * @member {object} effect - three.js VREffect
  * @member {object} systems - Registered instantiated systems.
  * @member {number} time
  */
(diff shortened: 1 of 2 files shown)


src/core/scene/a-scene.js
```diff
@@ -317,6 +342,50 @@ module.exports.AScene = registerElement('a-scene', {
       writable: window.debug
     },
 
+    enableCanvasPointerLock: {
```
dmarcos: What are the scenarios where someone would want to enable / disable this? I don't fully understand Edge WebVR behavior. I'm poking on this with partial information.

olga-microsoft: Just wanted to be flexible. I can't think of a scenario that this would be useful. I will remove this if it's okay with you.

dmarcos: As a maintainer I'm becoming more and more flag averse over time 😄 since new flags combine with previous ones that can combine in multiple weird ways that are difficult to debug and maintain over time. As a conservative measure I would get rid of the flag now and reintroduce it later if we found a legitimate use case for it. I cannot think of any at the moment.

olga-microsoft: Sure, this makes sense, removing!

dmarcos:
Thanks!

PR #3016, opened 2017-09-01
[gltf-model] Update examples to glTF2.0 assets.

donmccurdy:


src/components/gltf-model.js
```diff
@@ -26,6 +29,8 @@ module.exports.Component = registerComponent('gltf-model', {
       self.model.animations = gltfModel.animations;
       el.setObject3D('mesh', self.model);
       el.emit('model-loaded', {format: 'gltf', model: self.model});
+    }, undefined, function gltfFailed (error) {
```
dmarcos: The open PR for the microsoft controllers is also adding an error handler. https://github.com/aframevr/aframe/pull/3013/files#diff-7d85b9d27d0b70ce1f5dcb099f712247R29

Should we include this change there?

donmccurdy: Sounds good!

PR #3021, opened 2017-09-02
fix component build data when previousData is object with null prop

ngokevin:
**Description:**

Fixes `null has no properties` type error.

**Changes proposed:**
- Check for truthiness before running `.constructor`.


ngokevin:
@donmccurdy r? if you're around

dmarcos:
Thanks!

PR #3025, opened 2017-09-06
hand-controls bugfixes

leweaver:
**Description:**
The cubes in the tracked-controls example do not move when grabbed by Windows Motion Controllers. Additionally, the hand-controls meshes are incorrectly offset from the element center when initially loaded, as they inherit the parent transform (which has been set by the tracked-controls component). Lastly, there is a race condition somewhere, such that sometimes for a single frame `mesh.mixer` does not exist causing a null reference exception.

**Changes proposed:**
- Set the hand-controls mesh position and rotation to zero to ensure it is not offset from the element center
- In the grab component, make a copy of the position rather than pointing to an object reference for use in delta calculations.
- Added a null check to mesh.mixer prior to use.

### DIFF
--- a/examples/showcase/tracked-controls/components/grab.js
+++ b/examples/showcase/tracked-controls/components/grab.js
@@ -1,4 +1,4 @@
-/* global AFRAME */
+/* global AFRAME, THREE */
 
 /**
 * Handles events coming from the hand-controls.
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks!

PR #3026, opened 2017-09-06
Handle vrdisplayconnect/vrdisplaydisconnect events in VREffect and VRControls (fix #3019)

### LINKED ISSUE #3019: Plugging in VR headset after page load not detected
**Description:**
When plugging in an HMD after page load, clicking the Enter VR button still takes you to the 2D mouse driven mode, rather than entering VR.

This appears to be related to ([three.js #12109](https://github.com/mrdoob/three.js/issues/12109))
A workaround could be to re-implement the VRControls class in a-frame; it is only an example in three.js after all and is quite a small class to begin with.

- A-Frame Version: master
- Platform / Device: Windows Mixed Reality
- Reproducible Code Snippet or URL: https://aframe.io/examples/showcase/helloworld/

dmarcos:
I think this is the proper solution for https://github.com/aframevr/aframe/issues/3019 but I have not tested yet on headset. I can do a pass tomorrow on Oculus / Vive and Microsoft Mixed Reality headsets. @leweaver Can you also verify on your side that this fixes the problem?

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -128,8 +128,8 @@ module.exports.AScene = registerElement('a-scene', {
         // Exit VR on `vrdisplaydeactivate` (e.g. taking off Rift headset).
         window.addEventListener('vrdisplaydeactivate', this.exitVRBound);
 
-        // Enter VR on `vrdisplayconnect` (e.g. plugging on Rift headset).
-        window.addEventListener('vrdisplayconnect', this.enterVRBound);
+        // Exit VR on `vrdisplaydisconnect` (e.g. unplugging Rift headset).
+        window.addEventListener('vrdisplaydisconnect', this.exitVRTrueBound);
 
         // Exit VR on `vrdisplaydisconnect` (e.g. unplugging Rift headset).
         window.addEventListener('vrdisplaydisconnect', this.exitVRTrueBound);
--- a/tests/core/scene/a-scene.test.js
+++ b/tests/core/scene/a-scene.test.js
@@ -61,21 +61,6 @@ suite('a-scene (without renderer)', function () {
     });
   });
 
-  suite('vrdisplayconnect', function () {
-    test('tells A-Frame about entering VR when the headset is connected', function (done) {
-      var event;
-      var sceneEl = this.el;
-      var enterVRStub = this.sinon.stub(sceneEl, 'enterVR');
-      sceneEl.effect = {requestPresent: function () { return Promise.resolve(); }};
-      event = new CustomEvent('vrdisplayconnect');
-      window.dispatchEvent(event);
-      process.nextTick(function () {
-        assert.ok(enterVRStub.called);
-        done();
-      });
-    });
-  });
-
   suite('vrdisplaydisconnect', function () {
     test('tells A-Frame about entering VR when the headset is disconnected', function (done) {
       var event;
--- a/vendor/VRControls.js
+++ b/vendor/VRControls.js
@@ -19,6 +19,9 @@ THREE.VRControls = function ( object, onError ) {
 
 	}
 
+	window.addEventListener('vrdisplayconnect', function (evt) { vrDisplay = evt.display; });
+	window.addEventListener('vrdisplaydisconnect', function () { vrDisplay = undefined });
+
 	function gotVRDisplays( displays ) {
 
 		vrDisplays = displays;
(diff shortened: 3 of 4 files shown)


dmarcos:
Tried this on Oculus / Firefox 55 and it fixes the issue.

leweaver:
Thanks! We are taking a look on Edge and will let you know

olga-microsoft:
Hi Diego,
The issue described by Lewis is fixed in Edge too, but this leads to another issue:
1. Load the page 2. Connect a headset 3. Disconnect the headset -> page animation loop is broken.
I couldn't root cause, but VREffect is calling vrDisplay.requestAnimationFrame( f ); even without entering VR.
Does this repro in Firefox? I could open a new issue, but I think it should be fixed as a part of this change if it does.

olga-microsoft:
One more thing. After disconnecting the headset while in VR the behavior should be the same as after pressing exit VR button. Currently it stops rendering the scene.

dmarcos:
@olga-microsoft Thanks for testing this. I made a couple of changes to consider the cases you mentioned. Now when disconnecting the headseat VR mode disengages automatically. Let me know if this covers every case now.

olga-microsoft:
@dmarcos The disconnect is still not handled properly. After I disconnected the headset, Enter VR button appeared and clicking it enters fullscreen, but the animation loop is stopped, the scene is not being rendered anymore (fullscreen is black). 
If I plug in headset again, the Enter VR button starts VR as expected (but renders black too)

dmarcos:
@olga-microsoft I might need some help here. What browsers / headsets are you testing on? It is working as expected for me on Firefox / Oculus:

1. I start Firefox with the headset connected.
2. Click on the Googles to enter VR (VR mode engages and head tracking works as expected).
3. I unplug the cables (I tried the USB and HDMI independently and both at the same time) VR mode exits, the googles icon renders again and I can move the camera with mouse and WASD keys.
4. I click on the googles icon and the browser goes on fullscreen mode.

Do you see any errors in console?

olga-microsoft:
The behavior I described above is with Edge + Win Mixed Reality. 
I just tried with Firefox + Vive. Unplugging freezes the split screen picture and plugging resumes VR. There is no option to return to Desktop on the webpage.

dmarcos:
`vrdisplaydisconnect` does not fire reliably with Firefox + Vive. It's a known bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1397397). Do you have an Oculus to try on? 

On Edge + Win Mixed Reality. Can you try the following example (https://webvr.info/samples/00-hello-webvr.html) and see if you get the proper messages for `vrdisplayconnect` / `vrdisoplaydisconnect`?

image

image

olga-microsoft:
Yes, on https://webvr.info/samples/00-hello-webvr.html I am getting all events for either USB or HDMI plug/unplug. Your change is handling the events, which I can see with the VR button working as expected (fullscreen/HMD is toggled correctly). 
vrDisplay/window.requestAnimationFrame loop terminates after headset is disconnected.
Don't have Oculus

dmarcos:
Interesting. It seems you're seeing the render loop crashing and does not recover. Could it be that when the headset is unplugged in the middle of the render loop there's a last call to `submitFrame` that does not finish properly? The `vrdisplaydisconnect` would be dispatched after that call.

dmarcos:
It looks there's a time discrepancy between Firefox and Edge somewhere. When unplugging the headset on Edge the vrDisplay.requestAnimationFrame is cancelled implicitly and the callback never invoked stopping A-Frame rendering loop. In Firefox somehow the callback of the ongoing requestAnimationFrame is invoked one last time after the headset is unplugged so the render loop has a chance to seamlessly switch to window.requestAnimationFrame. I'm not sure what is the expected behavior. It's not something we define in the spec (https://w3c.github.io/webvr/spec/1.1/#dom-vrdisplay-requestanimationframe) I pushed a change to restart the render loop when disconnecting the headset that you can try. @olga-microsoft let me know if that solves the issue on Edge. I still have to improve the approach since we don't want to end up with multiple render loops on Firefox

olga-microsoft:
Thank you! With this change all looks good, rendering restarts. cancelAnimationFrame should be called in response to vrdisplaypresentchange to prevent spawning multiple render loops.

dmarcos:
Thanks! I'm now calling `cancelRequestAnimationFrame` on the handler for `vrdisplaydisconnect`. I also works on Firefox 55 / Vive

PR #3031, opened 2017-09-07
Do not resize the canvas while presenting (leads to resolution drop)

olga-microsoft:
Do not resize canvas in VR mode (not only for mobile, but for desktop too)
Tested on Edge + WMR, Firefox + Vive and Gear VR

Resizing the browser window while presenting leads to the canvas being resized accordingly, which updates resolution in the headset, causing a bad experience. Since Aframe does not rely on Three.js built-in mechanism for resizing the canvas this needs to be handled explicitly in a-scene.js.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -459,12 +459,11 @@ module.exports.AScene = registerElement('a-scene', {
         var canvas = this.canvas;
         var embedded = this.getAttribute('embedded') && !this.is('vr-mode');
         var size;
-        // Possible camera or canvas not injected yet.
-        // ON MOBILE the webvr-polyfill relies on the fullscreen API to enter
-        // VR mode. The canvas is resized by VREffect following the values returned
-        // by getEyeParameters. We don't want to overwrite the size with the
-        // windows width and height.
-        if (!camera || !canvas || this.is('vr-mode') && isMobile) { return; }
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -464,7 +464,7 @@ module.exports.AScene = registerElement('a-scene', {
         // VR mode. The canvas is resized by VREffect following the values returned
```
dmarcos: Can we update the comment to reflect the new logic? Something like this:
```Do not resize canvas If the a camera or canvas have not been injected. On VR mode, VREffect handles canvas resize based on the dimensions returned by the getEyeParameters function of the WebVR API. We want to honor those values instead of overwriting the size with the windows and width and height```

olga-microsoft: Updated, is this clear enough?

zcfan:
This pr cause me a problem. After apply it, I can't get a "full" fullscreen on pc. The canvas keep the same size as normal state, with black background.After I add `this.is('vr-mode')` back, everything is fine. Is this behavior made on purpose?

This [jsfiddle](https://jsfiddle.net/zcfan/fh7qhzdw/) and very example using 0.7.0 can reproduce it.

olga-microsoft:
I will look into it @zcfan

PR #3032, opened 2017-09-08
Update sound.md

clavis-magna:
distance model defaults to 'inverse' not linear

**Description:**
distance model defaults to 'inverse' not linear
**Changes proposed:**
change default from linear to inverse
-
-


dmarcos:
Thanks!

PR #3035, opened 2017-09-09
[docs] Update index.md

p4cm4nn:
**Description:**
Edited and updated the introduction in index.md
**Changes proposed:**
-Reorganised content to follow a logical order as such: "What is is A-Frame", "Features of A-Frame", "Getting Started with A-Frame", "Off You Go!"

### DIFF
--- a/docs/introduction/index.md
+++ b/docs/introduction/index.md
@@ -16,39 +16,6 @@ examples:
 [three.js]: https://threejs.org
 [webvr]: https://iswebvrready.com
 
-## Getting Started
-
-[glitch]: http://glitch.com/~aframe
-
-A-Frame can be developed from a plain HTML file without having to install
-anything. A great way to try out A-Frame is to **[remix the starter example on
-Glitch][glitch]**, an online code editor that instantly hosts and deploys for
-free. Alternatively, create an `.html` file and include A-Frame in the `<head>`:
-
-```html
-<html>
-  <head>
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks! Without supporting data and based on personal experience when you end up reading a Readme file in a github repo you are likely to have an idea of what the project is about and want to kick the tyres as fast as you can. I prefer to see clear and concise instructions to get started first. The organization you're proposing is more appropriate for a project website instead of the README file of the source code. @ngokevin what do you think?

dmarcos:
Sorry I just realized that it's not a PR against the README file. I still favor having onboarding instructions first. @ngokevin has probably more to say

p4cm4nn:
Hello @dmarcos. That is the distinction I made (i.e., readme file versus introduction). The rationale is that an intro will have a logical flow whereas the readme is to get up and running as fast as possible. It therefore makes sense for intro to contain what A-frame is, it's main features, how to get started using it, etc.

cvan:
thanks for opening this PR. I do agree with @dmarcos' recommendation about keeping the section placement as is. although it make hierarchically flow better with this change, presenting this information above is likely better for onboarding developers quicker with useful sample code.

ngokevin:
This was done on purpose to keep the quick started information above the fold versus paragraphs on exposition. Thanks though.

PR #3040, opened 2017-09-12
Add clarification about MutationObserver.

### LINKED ISSUE #3039: Clarify in docs that `MutationObserver` won't work due to performance optimizations
Following conversations on #3028 , it would be convenient to clarify this API won't work.

delapuente:
Fix #3039 

The clarification use MutationObserver as an example but the warning generally
states that A-Frame is not updating the DOM.

### DIFF
--- a/docs/introduction/javascript-events-dom-apis.md
+++ b/docs/introduction/javascript-events-dom-apis.md
@@ -250,6 +250,26 @@ entityEl.parentNode.removeChild(entityEl);
 A blank entity doesn't do anything. We can modify the entity by adding
 components, configuring component properties, and removing components.
 
+[debug]: ../components/debug.md
+[mutation-observer]: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+[flushtodom]: https://aframe.io/docs/0.6.0/core/component.html#flushtodom
+It is important to note that, for [performance reasons][flushtodom], updating
+an attribute which represents a component will bypass updating the DOM. For this
+reason, some DOM APIs would not work as expected. For instance, the
+[`MutationObserver`][mutation-observer] won't registaer any attribute update
+unless you use the [`debug`][debug] component in your scene:
+
+```js
+// <a-entity id="test" position="0 0 0"></a-entity>
+var test = document.querySelector('#test');
+var obs = new MutationObserver(function (mutations) {
+  console.log(mutations);
+  // will never print, add the debug component to the scene to make it work
+});
+obs.observe(test, { attributes: true });
+test.setAttribute('position', '0 1 0');
+```
+
 ### Adding a Component with `.setAttribute()`
 
 To add a component, we can use `.setAttribute(componentName, data)`. Let's add


docs/introduction/javascript-events-dom-apis.md
```diff
@@ -250,6 +250,26 @@ entityEl.parentNode.removeChild(entityEl);
...
+```js
+// <a-entity id="test" position="0 0 0"></a-entity>
+var test = document.querySelector('#test');
+var obs = new MutationObserver(function (mutations) {
+  console.log(mutations);
+  // will never print, add the debug component to the scene to make it work
+});
+obs.observe(test, { attributes: true });
+test.setAttribute('position', '0 1 0');
+```
```
donmccurdy: I don't think it's necessary to give a code sample showing what doesn't work. If the user knows what MutationObservers are, the note would be enough. If they don't, they're probably not concerned with this issue.

docs/introduction/javascript-events-dom-apis.md
```diff
@@ -250,6 +250,26 @@ entityEl.parentNode.removeChild(entityEl);
...
 components, configuring component properties, and removing components.
 
+[debug]: ../components/debug.md
+[mutation-observer]: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+[flushtodom]: https://aframe.io/docs/0.6.0/core/component.html#flushtodom
+It is important to note that, for [performance reasons][flushtodom], updating
+an attribute which represents a component will bypass updating the DOM. For this
+reason, some DOM APIs would not work as expected. For instance, the
+[`MutationObserver`][mutation-observer] won't registaer any attribute update
+unless you use the [`debug`][debug] component in your scene:
```
donmccurdy: Would suggest making this a blockquote-style note, like:

> **NOTE:** For performance reasons, updating a component will not update the DOM...

docs/introduction/javascript-events-dom-apis.md
```diff
@@ -250,6 +250,26 @@ entityEl.parentNode.removeChild(entityEl);
 A blank entity doesn't do anything. We can modify the entity by adding
 components, configuring component properties, and removing components.
 
+[debug]: ../components/debug.md
+[mutation-observer]: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
+[flushtodom]: https://aframe.io/docs/0.6.0/core/component.html#flushtodom
+It is important to note that, for [performance reasons][flushtodom], updating
+an attribute which represents a component will bypass updating the DOM. For this
+reason, some DOM APIs would not work as expected. For instance, the
+[`MutationObserver`][mutation-observer] won't registaer any attribute update
```
donmccurdy: *register

ngokevin:
I've noted it more simpler and more appropriate places: 
 https://github.com/aframevr/aframe/commit/1d92dbb3d0680cec3c1218c718bb44c639ab9f5c

Note A-Frame mentions to be mostly tool agnostic, but does not mean compatible with all Web. Mutation observers are also not a common API web developers will use or be aware of.

delapuente:
Thank you @ngokevin but I think the place @donmccurdy suggested  (`docs/introduction/javascript-events-dom-apis.md `) was more convenient since it is more likely developers will read it. Also, the warning was more general since we don't know all the DOM APIs that the optimisation could break. E.g.: another API also breaking is `.querySelector()` whith attribute selectors using the value of the attributes. I.e., this won't work:

```js
// <a-entity id="test"></a-entity>
document.querySelector('#test').setAttribute('position', '0 2 0');
console.log('element:', document.querySelector('[position="0 2 0"]'));
```

ngokevin:
Good point, didn't see that. I'd add that to the FAQ and to the debug component...and then perhaps a Caveats section in the JS section linking to the FAQ. Would you mind adding that? Don't think we need a whole paragraph and code section though; mutation observers and attribute selectors on positions are not very common operations.

donmccurdy:
Ok, yeah I thought MutationObservers were too narrow a topic for the FAQ, but something like `el.querySelector('[position=0 2 0]')` does maybe broaden the scope enough to include. That's fine with me. I do think we want to keep the code part light (ideally inline, if any) since these are not examples that people should imitate.

PR #3042, opened 2017-09-12
Allow MediaStream to be sound sources.

### LINKED ISSUE #3041: Add support for MediaStream in `sound` component
Currently, only buffered audio files are supported by A-Frame but supporting [`MediaStream`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream) sources would enable interesting scenarios such as positional WebRTC audio.

delapuente:
Fix #3041 

To be consistent with the A-Frame API, the patch allows the use of an `<audio>`
tag with a `MediaStream` object in the `srcObject` attribute. E.g.:

```js
// <a-assets></a-assets>
// <a-entity id="speaker"></a-entity>
const speaker = document.getElementById('speaker');
const assets = document.querySelector('a-assets');
navigator.mediaDevices.getUserMedia({ audio: true })
.then(stream => {
  const voice = new Audio();
  voice.id = 'speaker-voice';
  voice.srcObject = stream;
  assets.appendChild(voice);
  speaker.setAttribute('sound', `src:${voice.id}`);
});
```

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -66,19 +66,42 @@ module.exports.Component = registerComponent('sound', {
     if (srcChanged) {
       var self = this;
 
+      var isStream = typeof window.MediaStream === 'function' &&
+                     data.src.srcObject instanceof window.MediaStream;
+      if (isStream) {
+        var stream = data.src.srcObject;
+        this.pool.children.forEach(function (sound) {
+          sound.setNodeSource(sound.context.createMediaStreamSource(stream));
+        });
+        if (data.src.readyState >= 1) {
+          self.completeLoading();
+        } else {
+          data.src.addEventListener('loadedmetadata', function onMetadata () {
+            data.src.removeEventListener('loadedmetadata', onMetadata);
+            self.completeLoading();
+          });
+        }
+        return;
+      }
+
+      var path = typeof data.src === 'string' ? data.src : data.src.src;
       this.loaded = false;
-      this.audioLoader.load(data.src, function (buffer) {
+      this.audioLoader.load(path, function (buffer) {
         self.pool.children.forEach(function (sound) {
           sound.setBuffer(buffer);
         });
-        self.loaded = true;
 
         // Remove this key from cache, otherwise we can't play it again
-        THREE.Cache.remove(data.src);
+        THREE.Cache.remove(path);
         if (self.data.autoplay || self.mustPlay) { self.playSound(); }
-        self.el.emit('sound-loaded');
+        self.completeLoading();
       });
     }
+  },
+
+  completeLoading: function () {
+    this.loaded = true;
+    this.el.emit('sound-loaded');
   },
 
   pause: function () {
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -72,6 +72,7 @@ function arrayStringify (value) {
 function assetParse (value) {
   var el;
   var parsedUrl;
+  var mediaTags = ['CANVAS', 'VIDEO', 'IMG', 'AUDIO'];
 
   // If an element was provided (e.g. canvas or video), just return it.
   if (typeof value !== 'string') { return value; }
@@ -86,7 +87,7 @@ function assetParse (value) {
     if (el) {
       // Pass through media elements. If we have the elements, we don't have to call
       // three.js loaders which would re-request the assets.
-      if (el.tagName === 'CANVAS' || el.tagName === 'VIDEO' || el.tagName === 'IMG') {
+      if (mediaTags.indexOf(el.tagName) >= 0) {
         return el;
       }
       return el.getAttribute('src');


src/components/sound.js
```diff
@@ -64,17 +64,29 @@ module.exports.Component = registerComponent('sound', {
     }
     // All sound values set. Load in `src`.
     if (srcChanged) {
-      var self = this;
+      var isStream = (data.src.srcObject instanceof MediaStream);
```
cvan: if `MediaStream` isn't available or enabled (such as in release-channel Safari desktop or mobile < iOS 11), this will throw a `ReferenceError`:

```
ReferenceError: Can't find variable: MediaStream
```

you can change this to

```js
var isStream = (data.src.srcObject instanceof window.MediaStream);
```

or this:

```js
var isStream = ('MediaStream' in window && data.src.srcObject instanceof MediaStream);
```

if you decide to use `MediaStream` (instead of `window.MediaStream`), you'll probably want to denote the globals in comments:

```js
/* global MediaStream */
```

src/components/sound.js
```diff
@@ -64,17 +64,29 @@ module.exports.Component = registerComponent('sound', {
...
+
+      if (isStream) {
+        var stream = data.src.srcObject;
+        this.pool.children.forEach(function (sound) {
+          sound.setNodeSource(sound.context.createMediaStreamSource(stream));
+        });
+        this.loaded = true;
+        this.el.emit('sound-loaded');
```
cvan: do we want to emit which src was emitted? or should developers do the look ups in the event listener and assume nothing's changed since the event was emitted?

src/components/sound.js
```diff
@@ -64,17 +64,29 @@ module.exports.Component = registerComponent('sound', {
...
     if (srcChanged) {
-      var self = this;
+      var isStream = (data.src.srcObject instanceof MediaStream);
+
+      if (isStream) {
+        var stream = data.src.srcObject;
+        this.pool.children.forEach(function (sound) {
+          sound.setNodeSource(sound.context.createMediaStreamSource(stream));
+        });
+        this.loaded = true;
```
cvan: should this wait for the `loadedmetadata` event to fire?

src/core/propertyTypes.js
```diff
@@ -86,7 +86,7 @@ function assetParse (value) {
     if (el) {
       // Pass through media elements. If we have the elements, we don't have to call
       // three.js loaders which would re-request the assets.
-      if (el.tagName === 'CANVAS' || el.tagName === 'VIDEO' || el.tagName === 'IMG') {
+      if (['CANVAS', 'VIDEO', 'IMG', 'AUDIO'].indexOf(el.tagName) >= 0) {
```
cvan: do you want to put this as a constant up top?

cvan: probably not really used, but what about `<object>`?

src/components/sound.js
cvan: I assume by this point [autoplay](https://github.com/aframevr/aframe/pull/3042/files#diff-6e2259d7fcbd1d0b371631b04575bc07R90) is already assumed? what about with Notifications API prompts, etc.?

haven't tested that part - just figured I'd ask.

ngokevin:
Perhaps makes more sense for the use case of media streamed audio over multiuser to handle it within a separate component. Don't feel comfortable have the sound component handle both buffered audio as well as streams. Multiuser setups will have their own custom components for several things.

netpro2k:
Though I can see the argument for not bundling streaming sources into `sound` I think the change to allow audio elements to get passed through untouched is still probably a good idea as it allows one to write a `stream-sound` component without having to do anything hacky.

src/components/sound.js
cvan: @delapuente friendly ping on this

delapuente:
Thank you all for the feedback. Working on it.

delapuente:
Hello @cvan, next version is ready for review.

@ngokevin, @netpro2k Despite I understand the argument against `sound` component dealing with streams, I think the patch gives you support for a very simple but common use case in a consistent and straightforward way. Furthermore, the source code might serve as a model for more complex implementations.

ngokevin:
I think this should be a component. That's what A-Frame is set up to do. Anyone can create components that do one thing and do it well. Supporting more things in the core sets a dangerous precedent to support even more things in the core. I think the sound component already does too much since it supports pooling (should have been a separate `sound-pool` component).

We might find as well that the Positional Audio default with three.js does not work well for multiuser, and we'd later be forced to continue to add more features and such to maintain it proper. I see there's already custom stuff happening in https://github.com/networked-aframe/networked-aframe/issues/11

So I recommend a component + Glitch, and we'd be happy to feature it in the documentation for people to remix. Next time I'll try to make the decision to close earlier so efforts aren't wasted. Thanks for understanding.

netpro2k:
As I mentioned above, I can totally see the argument for not adding this feature to the `sound` component, but I think the change in this PR to propertyTypes.js that allow audio elements to get passed through untouched is still probably a good idea, otherwise its difficult for people to write a component that handles MediaStreams themselves.

ngokevin:
Not sure whether that property type should pass through audio. The current three.js audio helpers do an XHR directly to the URL, they do not touch the audio element. It's better to pass a string. You also don't need to use `<a-assets>` for media streams because `<a-assets>` is for preloading only, and streams can't be preloaded.

delapuente:
Even if "preloading only" was the original intention, `<a-assets>` is also the place for mixins and, as far as I know, mixins can be generated dynamically. From the semantic perspective, it makes sense to contain all the experience assets inside the `<a-assets>` tag. From an API perspective, it is more homogeneous if media elements are treated equally (if the core sound component only needs the URL, that's perfectly fine).

Finally, considering the argument of providing only basic functionality in the core, I'm wondering why #3176 got merged.

netpro2k:
Yes the `sound` component only needs the url, which it can get from the audio element, other componets that may want an `type: 'audio'` property might need more, so it makes sense for the element to just be passed through unchanged, like canvas, video, and image properties.

ngokevin:
You can always use a string type or selector type and handle how you need. Right now, the sound component does not use the raw audio element so it doesn't make sense to change the behavior.

PR #3043, opened 2017-09-12
Windows motion controller glb format update

leweaver:
**Description:**
Updates to be compliant with final assets (https://github.com/aframevr/assets/pull/30)

**Changes proposed:**
- Updated pointing pose node name
- Fixed pointing pose being incorrect when the controller is not at origin on initialization (calculating transforms from glTF root, not parent world position)
- Fixed laser-controls not showing pointing ray on controller reconnection
- Removed erroneous debug code.

### DIFF
--- a/src/components/windows-motion-controls.js
+++ b/src/components/windows-motion-controls.js
@@ -254,8 +254,6 @@ module.exports.Component = registerComponent('windows-motion-controls', {
 
     // Cache our meshes so we aren't traversing the hierarchy per frame
     if (rootNode) {
-      rootNode.updateMatrixWorld();
-
       // Button Meshes
       for (i = 0; i < this.mapping.buttons.length; i++) {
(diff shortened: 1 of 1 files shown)


src/components/windows-motion-controls.js
```diff
@@ -319,36 +318,36 @@ module.exports.Component = registerComponent('windows-motion-controls', {
         }
       }
 
-      // Calculate the pointer pose (used for rays), by applying the inverse holding pose, then the pointing pose.
+      // Calculate the pointer pose (used for rays), by applying the world transform of th POINTER_POSE node
+      // in the glTF (assumes that root node is at world origin)
```
dmarcos: Can we move the logic to calculate the ray origin to a separate function? The onModelLoaded function is getting pretty big

dmarcos:
Thank you!

PR #3046, opened 2017-09-13
Hide the controller model instead of the whole entity

dmarcos:
`windows-motion-controllers` hides the whole entity if the expected controllers are not present. This will also make any other controllers (Vive, Oculus...) invisible. I replaced the change in visibility of the controllers entity with the visiblity of just the windows motion controllers model if there's one already loaded. I tested more thoroughly the windows motion controllers in Edge and I find inconsistencies: I ofthen see just one controller when starting Edge and switching it on/off does not help. Also switching off and back on when the controllers are properly tracked does not seem to work. Not sure yet if the problem is Edge or A-Frame but Vive + Firefox seem to work more reliably when it comes to detect and track the controllers. Are you experiencing the same problems on Edge? @leweaver @olga-microsoft

### DIFF
--- a/src/components/windows-motion-controls.js
+++ b/src/components/windows-motion-controls.js
@@ -130,8 +130,8 @@ module.exports.Component = registerComponent('windows-motion-controls', {
       index: this.data.pair
     });
 
-    if (this.data.hideDisconnected) {
-      this.el.setAttribute('visible', this.controllerPresent);
+    if (this.data.hideDisconnected && this.controllerModel) {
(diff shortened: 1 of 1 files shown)


leweaver:
Thanks for the visibility change!

Regarding the other issues, it takes sometimes 5-10 seconds for a controller to be removed from the gamepads array after it is powered off. To determine if the not appearing issue is A-frame or edge, perhaps try the WebVR.info controller sample? https://webvr.info/samples/XX-vr-controllers.html

PR #3047, opened 2017-09-14
Do not use deltas to update headset and controllers. Use absolute values instead

dmarcos:
This solves a lot of issues with controllers connecting / disconnecting and the cases where the camera starts on the floor or unexpected position. We should have done this a loooooong time ago 😄 @olga-microsoft this improves a lot the behavior of the `windows motion controllers` in Edge.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -28,7 +28,7 @@ module.exports.Component = registerComponent('look-controls', {
     this.previousHMDPosition = new THREE.Vector3();
     this.hmdQuaternion = new THREE.Quaternion();
     this.hmdEuler = new THREE.Euler();
-    this.position = {};
+    this.position = new THREE.Vector3();
     this.rotation = {};
 
     this.setupMouseControls();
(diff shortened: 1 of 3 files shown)


src/components/look-controls.js
```diff
@@ -227,34 +227,27 @@ module.exports.Component = registerComponent('look-controls', {
...
-      var previousHMDPosition = this.previousHMDPosition;
-      var sceneEl = this.el.sceneEl;
-
-      if (!sceneEl.is('vr-mode')) { return; }
+  updatePosition: function () {
+    var el = this.el;
+    var currentHMDPosition;
+    var currentPosition;
+    var position = this.position;
+    var previousHMDPosition = this.previousHMDPosition;
```
olga-microsoft: Is previousHMDPosition defined if not vr-mode? Move below return?

olga-microsoft:
Tried out laser controls with WMR. Works much smoother and connect/disconnect is handled correctly

src/components/look-controls.js
```diff
@@ -227,34 +227,27 @@ module.exports.Component = registerComponent('look-controls', {
...
+    currentPosition = el.getAttribute('position');
 
-      currentPosition = el.getAttribute('position');
-      position.x = currentPosition.x + deltaHMDPosition.x;
-      position.y = currentPosition.y + deltaHMDPosition.y;
-      position.z = currentPosition.z + deltaHMDPosition.z;
-      el.setAttribute('position', position);
-    };
-  })(),
+    position.x = currentPosition.x - previousHMDPosition.x + currentHMDPosition.x;
```
ngokevin: Wonder if we can make `this.position` a `Vector3` and use the `sub`/`add` methods instead of doing individual x/y/z operations. Goes for the tracked-controls as well.

src/components/look-controls.js
ngokevin: Probably harmless, we have an informal code style to define vars up top and alias `this.` variables to remove `this.` clutter.

ngokevin:
r+wc

ngokevin:
suite.only

PR #3053, opened 2017-09-15
Forgo gamepadconnected / gamepaddisconnected event handlers. It adds …

dmarcos:
…multiple entry points to the checkIfControllerPresent logic, the event is not reliable across browsers and and it's redundant because we poll already the gamepads per tick


machenmusik:
+1, since we re-confirmed that Chromium/Chrome derivatives require getGamepads() call per tick for updates to work, those events don't help us.

dmarcos:
@machenmusik thanks for veryfing

Issue #2701, opened 2017-05-25
hide hand-controls model by default, only show model once controller connected


ngokevin:
**Description:**

Wait for `controllerconnected` event and check the `evt.detail.name` matches `vive-controls` or `oculus-touch-controls`.

I think we can wait until `controllerconnected` to start the XHR for the hand-model OBJ/MTL.

machenmusik:
I would strongly suggest that in the data schema there should be an array of strings that are component names to both attach initially and respond to upon controller connected, that defaults to whatever you like (here, vive-controls and oculus-touch-controls)

rttomlinson:
Is this issue still being worked on?

gurumukhi:
Hi @rttomlinson, thanks for stepping up for this bug I dont think anyone is working on this. Feel free to work on this bug and create a PR, if you have any questions you can ask on #issues slack channel of a-frame (https://aframe.io/community/#slack).

jselmani:
I see that @rttomlinson has considered working on this and probably still is.  I am a new contributor to open source projects and would like to get involved.  Considering this is still an open issue, can I work on this issue as well?

Thanks!

rttomlinson:
Hi Jiel,

Feel free to take it over. I haven't had much time lately. Let me know if
you have any questions!

jselmani:
Thanks @rttomlinson.  There has been a commit that I referenced that looks as if the issue has been solved, so before I start I would like to hear from someone from the A-Frame team so I can perhaps get working on another issue.  If I have any questions though, I will definitely ask!

dmarcos:
@jselmani good catch! yes, this should be fixed in 0.7.0

jselmani:
Thanks for clarifying @dmarcos, I appreciate it!

PR #3059, opened 2017-09-15
Hide hands model when controller connected/disconnected

dmarcos:


jselmani:
Does this commit solve issue [#2701](https://github.com/aframevr/aframe/issues/2701)?  Just wanted to know because I wanted to take a jab at it if it is still available.

PR #3066, opened 2017-09-18
Preserving the render callback when cancelling the VR animation frame

leweaver:
**Description:**
Since [#0dba182](https://github.com/aframevr/aframe/commit/0dba1828118eed1f69edb9c2bc403b587e3399be#diff-2bf1de3dd522db1514c4c8ff0e3256a8) the 2D render loop does not restart once the headset is disconnected, as `scope.f` is set to undefined in the `cancelAnimationFrame` function.

**Changes proposed:**
- Return the old value for `f` from the `cancelAnimationFrame` function
- Pass that returned value into the new `requestAnimationFrame` function


dmarcos:
Thanks! 👌

PR #3068, opened 2017-09-19
Adding example to interaction page

PAPERPANKS:
**Changes proposed:**
- adding an example related to events and interaction in A-frame
- editing `readme.md file` , adding `target="_blank" ` to `<a>` tags

This commit is ready for review.


ngokevin:
This is cool, but swapping the environment causes hangs in the scene, right?

PAPERPANKS:
@ngokevin may be there are some back-end issue but this may be a best example to add and showcase.

ngokevin:
I mean the performance is not really good, we can't really recommend slow scenes.

ngokevin:
Perhaps do something less intensive on the clicks. For example, changing some text, or making something a different color, popping up elements

PR #3069, opened 2017-09-20
Correct property name for whitespace

brianpeiris:
**Description:**
three-bmfont-text uses the "mode" property to specify the whitespace mode. The text component just passes its component data directly to the bmfont library. 

Perhaps the component should be changed to rename the property before sending it to the library instead?

**Changes proposed:**
- Change the docs to use the actual bmfont property name.


dmarcos:
`mode` is a vague name. I would keep the `whiteSpace` property name and make sure that value is mapped to the `mode` attribute of `three-bmfont-text`

brianpeiris:
Gotcha, I'll open a new pull request with that change.

Issue #3067, opened 2017-09-19
json-model entity can't be clicked


yursile:
when i use v0.6.0, the json-model entity can't be selected nor clicked
but v0.5.0 goes well.
so what is the best practice on v0.6.0?
any suggestion would be grateful

here is my code:
`
<a-entity id="bot" cursor-listener class="clickable" animation-control="clip: run; : .3" scale="3.5 3.5 3.5" position="0 -4 -2.5" rotation="0 0 0" animation-mixer="clip: idle"
            json-model="src: url(models/bot4.json);"></a-entity>
`

donmccurdy:
I believe this is the same issue as #2980, the raycaster's cache must be cleared after models load. We're considering ways to do this automatically, but you can add a component in the meantime:

```js
AFRAME.registerComponent('raycaster-refresh', {
  init: function () {
    var sceneEl = this.el.sceneEl;
    sceneEl.addEventListener('object3dset', function () {
      var raycasterEl = sceneEl.querySelector('[raycaster]');
      raycasterEl.components.raycaster.refreshObjects();
    });
  }
});
```

```html
<a-scene raycaster-refresh>
  ...
</a-scene>
```

Issue #2980, opened 2017-08-16
Raycaster line not truncating correctly on OBJs for 'raycaster' component


evhan55:
**Description:**

- A-Frame Version: 0.6.1
- Platform / Device: Chrome / Google Pixel
- Reproducible Code Snippet or URL: https://slime-cereal.glitch.me

**Steps To Reproduce**

- Load URL on a Google Pixel device in Chrome
- Put on Daydream View Headset
- Calibrate Daydream controller...
- Point laser at orange/black striped model
- Expected: raycaster truncates at model
- Observed: raycaster shoots right through model, and also through two other OBJs I also tested

**Note**
- Raycaster line seemingly truncates correctly when using the `laser-controls` component but not the `raycaster` component

![screenshot_20170816-144602](https://user-images.githubusercontent.com/699840/29380029-5cbe618a-8292-11e7-971d-509f6ae38aae.png)

evhan55:
Possibly related to this older issue: https://github.com/ngokevin/aframe-core/issues/797

donmccurdy:
Thanks for the test example!

Found the problem: We typically think of the `raycaster.refreshObjects()` as needing to be called after the DOM changes, but actually `raycaster.objects` is caching the list of `Object3D`s associated with each entity, not the entity itself. So its cache needs to be cleared not just when the DOM changes, but when a model is added to an entity.

In this case, the [stub mesh created by the `material` component](https://github.com/aframevr/aframe/blob/master/src/components/material.js#L155) is what got cached. If I manually call `el.components.raycaster.refreshObjects()` after the model loads, everything works. That stub mesh has bitten me before (https://github.com/aframevr/aframe/issues/2464) — I wonder if we could do without it?

I'd like to find a more intuitive way to keep the raycaster up to date, but I'm not sure how to handle this case. Here's the [relevant snippet of code](https://github.com/aframevr/aframe/blob/master/src/components/raycaster.js#L122-L141). We could listen for `object3dset` on the scene, with a debounce?

machenmusik:
Another approach, can raycaster attach listener for model loaded event to entities of interest?

ngokevin:
We could have raycaster listen for `object3dset` on a opt-in flag?

Or provide harder documentation and Glitch examples how to manually refresh the raycaster as you need in docs. That'd be the most performant approach. Perhaps a component `raycaster-refresh="raycasterEl: [raycaster]"`.

machenmusik:
[...]

PR #3070, opened 2017-09-20
[raycaster] Use MutationObserver and object3dset/remove to refresh.

donmccurdy:
This is a proposed fix for #2980 and #3067. It's not possible to perfectly detect when the whitelist may have changed — direct calls to `el.object3D.add(mesh)` will be missed, for example — but this covers all of the A-Frame API surface I think.

Changes:
- Use a MutationObserver to detect new elements and new attributes. While attribute *values* are not flushed to the DOM, the attribute names are, so this at least covers cases like adding a new component or changing a className.
- Listen for `object3dset` and `object3dremove`. Unlike child-added/removed, these are only triggered after the entity has loaded, so no need for a delay.
- When any of the above listeners fire, just set a dirty flag.
- Before each raycast, if dirty flag is set, refresh the object list.
- Warn users if the `raycaster.objects` selector is exotic, like`[position=0 2 0]`, which will be missed by our MutationObserver.

If this direction looks OK, I'll add tests. It's a bigger change than we should include for 0.7.0, and can wait for 0.7.1 or 0.8.0.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -12,7 +12,7 @@ var dummyVec = new THREE.Vector3();
 // refresh the whitelist. Matches classnames, IDs, and presence of attributes.
 // Selectors for the value of an attribute, like [position=0 2 0], cannot be
 // reliably detected and are therefore disallowed.
-var OBSERVER_SELECTOR_RE = /^[\w\s-.[\]#]*$/;
+var OBSERVER_SELECTOR_RE = /^[\w\s-.,[\]#]*$/;
 
 // Configuration for the MutationObserver used to refresh the whitelist.
 // Listens for addition/removal of elements and attributes within the scene.
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -9,6 +9,9 @@ var warn = utils.debug('components:raycaster:warn');
 
 var dummyVec = new THREE.Vector3();
 
+// Defines selectors that should be 'safe' for the MutationObserver used to
+// refresh the whitelist. Less common selectors, like [position=0 2 0], cannot
+// be detected here.
 var SAFE_SELECTOR_RE = /^[\w\s-.[\]#]+$/;
 
 /**
@@ -144,6 +147,7 @@ module.exports.Component = registerComponent('raycaster', {
       // Update check time.
       this.prevCheckTime = time;
 
+      // Refresh the object whitelist if needed.
       if (this.dirty) { this.refreshObjects(); }
 
       // Store old previously intersected entities.
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -12,7 +12,7 @@ var dummyVec = new THREE.Vector3();
 // refresh the whitelist. Matches classnames, IDs, and presence of attributes.
 // Selectors for the value of an attribute, like [position=0 2 0], cannot be
 // reliably detected and are therefore disallowed.
-var OBSERVER_SELECTOR_RE = /^[\w\s-.[\]#]+$/;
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks I'll look into it post 0.7.0 release

donmccurdy:
Added `raycaster.watch` property, if users would like to manage refreshes themselves.

/cc @machenmusik

src/components/raycaster.js
```diff
@@ -70,19 +80,32 @@ module.exports.Component = registerComponent('raycaster', {
...
+
+    this.setDirty();
   },
 
   play: function () {
-    this.el.sceneEl.addEventListener('loaded', this.refreshObjects);
-    this.el.sceneEl.addEventListener('child-attached', this.refreshOnceChildLoaded);
-    this.el.sceneEl.addEventListener('child-detached', this.refreshObjects);
+    if (!this.data.watch) { return; }
+    this.observer.observe(this.el.sceneEl, {
```
ngokevin: Could store/reuse this object as a constant

src/components/raycaster.js
```diff
@@ -38,13 +47,14 @@ module.exports.Component = registerComponent('raycaster', {
...
     this.intersectedEls = [];
-    this.objects = null;
+    this.objects = [];
     this.prevCheckTime = undefined;
     this.prevIntersectedEls = [];
     this.raycaster = new THREE.Raycaster();
     this.updateOriginDirection();
-    this.refreshObjects = bind(this.refreshObjects, this);
-    this.refreshOnceChildLoaded = bind(this.refreshOnceChildLoaded, this);
+    this.setDirty = bind(this.setDirty, this);
```
ngokevin: I re-ran some jsperf stuff, we can just start using vanilla bind. I don't think there's perf difference

src/components/raycaster.js
```diff
@@ -28,7 +36,8 @@ module.exports.Component = registerComponent('raycaster', {
     origin: {type: 'vec3'},
     recursive: {default: true},
     showLine: {default: false},
-    useWorldCoordinates: {default: false}
+    useWorldCoordinates: {default: false},
+    watch: {default: true}
```
ngokevin: perhaps `autoRefresh`

src/components/raycaster.js
```diff
@@ -1,11 +1,19 @@
...
 
 var bind = utils.bind;
+var warn = utils.debug('components:raycaster:warn');
 
 var dummyVec = new THREE.Vector3();
 
+// Defines selectors that should be 'safe' for the MutationObserver used to
+// refresh the whitelist. Less common selectors, like [position=0 2 0], cannot
+// be detected here.
+var SAFE_SELECTOR_RE = /^[\w\s-.[\]#]+$/;
```
ngokevin: Can explain more what this selector matches, and name maybe like `OBSERVER_SELECTOR_RE`

src/components/raycaster.js
```diff
@@ -70,19 +80,32 @@ module.exports.Component = registerComponent('raycaster', {
       el.removeAttribute('line');
     }
 
-    this.refreshObjects();
+    if (data.objects !== oldData.objects && !SAFE_SELECTOR_RE.test(data.objects)) {
+      warn('Selector "' + data.objects + '" may not update automatically with DOM changes.');
+    }
+
+    if (data.watch !== oldData.watch && el.isPlaying) {
+      data.watch ? this.play() : this.pause();
```
ngokevin: Should not call handlers from other handlers, can extract to separate methods

ngokevin:
The approach looks good to me to continue with tests, will want to mention in perf best practices section to watch out for this potentially

donmccurdy:
Updated based on feedback, and added a couple tests.

PR #3075, opened 2017-09-21
Fix fullscreen problem on pc browser

zcfan:
**Description:**

I can't get a "full" fullscreen on pc. The canvas keep the same size as normal state, with black background. After I add `this.is('vr-mode')` back, everything is fine.

Related: https://github.com/aframevr/aframe/pull/3031#issuecomment-331105971

**Changes proposed:**
- Fix fullscreen problem on pc browser


olga-microsoft:
Thank you for reporting the problem! Please see my suggested solution in #3080
Adding isMobile back will fix fullscreen mode and reintroduce the problem of resizing browser window, while in immersive mode on desktop.

zcfan:
Thank you, I thought immersive mode and fullscreen is the same thing 😂

PR #3080, opened 2017-09-21
Do not resize the canvas while presenting in immersive mode or on mobile

olga-microsoft:
Renderer size should only be updated on desktop in fullscreen mode.
Resizing the browser window while presenting leads to the canvas being resized accordingly, which updates resolution in the headset, causing a bad experience. Since Aframe does not rely on Three.js built-in mechanism for resizing the canvas this needs to be handled explicitly in a-scene.js.
Alternative to #3075, follow-up of #3031

### DIFF
--- a/tests/core/scene/a-scene.test.js
+++ b/tests/core/scene/a-scene.test.js
@@ -388,6 +388,58 @@ suite('a-scene (without renderer)', function () {
     });
   });
 
+  suite.only('resize', function () {
+    var sceneEl;
+    var setSizeSpy;
+
+    setup(function () {
+      sceneEl = this.el;
+      AScene.prototype.resize.restore();
+      sceneEl.camera = { updateProjectionMatrix: function () {} };
+      sceneEl.canvas = document.createElement('canvas');
+      sceneEl.renderer = { setSize: function () {} };
+
+      setSizeSpy = this.sinon.spy(sceneEl.renderer, 'setSize');
+    });
+
+    test('resize renderer when not in vr mode', function () {
+      sceneEl.resize();
+
+      assert.ok(setSizeSpy.called);
+    });
+
+    test('resize renderer when in vr mode in fullscreen presentation (desktop, no headset)', function () {
+      sceneEl.effect = {
+        isPresenting: false
+      };
+      sceneEl.addState('vr-mode');
+
+      sceneEl.resize();
+
+      assert.ok(setSizeSpy.called);
+    });
+
+    test('does not resize renderer when in vr mode on mobile', function () {
+      sceneEl.isMobile = true;
+      sceneEl.addState('vr-mode');
+
+      sceneEl.resize();
+
+      assert.notOk(setSizeSpy.called);
+    });
+
+    test('does not resize renderer when in vr mode and presenting in a headset', function () {
+      sceneEl.effect = {
+        isPresenting: true
+      };
+      sceneEl.addState('vr-mode');
+
+      sceneEl.resize();
+
+      assert.notOk(setSizeSpy.called);
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


tests/core/scene/a-scene.test.js
```diff
@@ -388,6 +388,70 @@ suite('a-scene (without renderer)', function () {
     });
   });
 
+  suite('resize', function () {
+    test('resize not in vr updates renderer size', function () {
+      var sceneEl = this.el;
+      AScene.prototype.resize.restore();
+      sceneEl.camera = { updateProjectionMatrix: function () {} };
+      sceneEl.canvas = document.createElement('canvas');
+      sceneEl.renderer = { setSize: function () {} };
```
dmarcos: This code bellow is common to all the tests in the suite. Can we factor it out to a setup function? e.g: https://github.com/aframevr/aframe/blob/master/tests/core/scene/a-scene.test.js#L214

````javascript
 var sceneEl = this.el;
 AScene.prototype.resize.restore();
 sceneEl.camera = { updateProjectionMatrix: function () {} };
 sceneEl.canvas = document.createElement('canvas');
 sceneEl.renderer = { setSize: function () {} };
````

File: tests/core/scene/a-scene.test.js [hunk 2]
```diff
@@ -388,6 +388,70 @@ suite('a-scene (without renderer)', function () {
     });
   });
 
+  suite('resize', function () {
+    test('resize not in vr updates renderer size', function () {
```
dmarcos: difficult to read. May be better like this? `resize renderer when not in vr mode`

tests/core/scene/a-scene.test.js
```diff
@@ -388,6 +388,70 @@ suite('a-scene (without renderer)', function () {
...
+      sceneEl.canvas = document.createElement('canvas');
+      sceneEl.renderer = { setSize: function () {} };
+      var setSizeSpy = this.sinon.spy(sceneEl.renderer, 'setSize');
+
+      sceneEl.resize();
+
+      assert.ok(setSizeSpy.called);
+    });
+
+    test('resize in non-immersive vr on desktop updates renderer size', function () {
```
dmarcos: `resize renderer when in vr mode in fullscreen presentation (no headset)`

tests/core/scene/a-scene.test.js
```diff
@@ -388,6 +388,70 @@ suite('a-scene (without renderer)', function () {
...
+      };
+      sceneEl.addState('vr-mode');
+      var setSizeSpy = this.sinon.spy(sceneEl.renderer, 'setSize');
+
+      sceneEl.resize();
+
+      assert.ok(setSizeSpy.called);
+    });
+
+    test('resize in vr on mobile does not update renderer size', function () {
```
dmarcos: `does not resize renderer in vr mode on mobile devices`

tests/core/scene/a-scene.test.js
```diff
@@ -388,6 +388,70 @@ suite('a-scene (without renderer)', function () {
...
+      sceneEl.isMobile = true;
+      sceneEl.addState('vr-mode');
+      var setSizeSpy = this.sinon.spy(sceneEl.renderer, 'setSize');
+
+      sceneEl.resize();
+
+      assert.notOk(setSizeSpy.called);
+    });
+
+    test('resize in immersive vr (while presenting) does not update renderer size', function () {
```
dmarcos: `does not resize renderer when in vr mode and presenting in a headset`

tests/core/scene/a-scene.test.js
olga-microsoft: Good idea, thank you

dmarcos:
Thank you!

PR #3091, opened 2017-09-28
[docs] Adding `hasLoaded` property

delapuente:
`hasLoaded` property is missing from `entity` documentation.

### DIFF
--- a/docs/core/entity.md
+++ b/docs/core/entity.md
@@ -93,19 +93,9 @@ document.querySelector('a-entity[sound]').components.sound.pause();
 
 ### `hasLoaded`
 
-Whether the entity has fully initialized. Initialization ends after entity's chidren and components have finished loading. A common pattern to wait for an entity to load before doing something is:
-
- ```js
- function doSomething() {
-   if (!entity.hasLoaded) {
-     entity.addEventListener('loaded', doSomething);
-     return;
-   }
-   // do something with entity
- }
- var entity = document.querySelector('a-entity');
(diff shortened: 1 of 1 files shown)


docs/core/entity.md
```diff
@@ -91,6 +91,22 @@ Or if a component exposes an API, we can call its methods:
...
+ ```js
+ function doSomething() {
+   if (!entity.hasLoaded) {
+     entity.addEventListener('loaded', doSomething);
+     return;
+   }
+   // do something with entity
+ }
+ var entity = document.querySelector('a-entity');
+ doSomething();
```
donmccurdy: To encourage people to write their JS using components, could I suggest reorganizing this as:

```js
AFRAME.registerComponent('my-behavior', {
  init: function () {
    var el = this.el;
    if (el.hasLoaded) {
      this.doSomething();
    } else {
      el.addEventListener('loaded', this.doSomething.bind(this));
    }
  },
  doSomething: function () {
    // ...
  }
});
```

ngokevin: was literally just changing to something like that

ngokevin: though even then i'm not sure if that pattern is appropriate. better to explicitly wait on something or add a dependency, i don't know how common it is to run something on `loaded`

donmccurdy: ^For a scene component I use this. For normal components more often using a dependency, `model-loaded`, or `object3dset`, yeah.

PR #3095, opened 2017-09-29
clone default array property type, use default array value if not defined (fixes #3092)

### LINKED ISSUE #3092: Schema `default` property is not working for `array` type
The following schema is not working as expected:
```js
schema: { type: 'array', default: ['a'] }
```

Demo (open the console):
https://codepen.io/lodr/pen/oGWyoM

Expected log:
```
["a"]
```

Actual log:
```
[]
```

ngokevin:
**Description:**

1. Arrays were not getting cloned.
2. Default arrays value were not being used if no data was explicitly defined...since array parser will parse `''` to `[]`. `buildData` would then treat `[]` as a defined value.

**Changes proposed:**
- Apply several checks used by Object type also to the Array type for cloning.
- Add a case in `buildData` to treat `[]` not as a defined component value.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -325,7 +325,7 @@ Component.prototype = {
    * @return {object} The component data
    */
   buildData: function (newData, clobber, silent, skipTypeChecking) {
-    var componentDefined = newData !== undefined && newData !== null;
+    var componentDefined;
     var data;
     var defaultValue;
     var keys;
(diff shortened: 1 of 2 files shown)


src/core/component.js
```diff
@@ -547,3 +549,7 @@ function wrapPlay (playMethod) {
     sceneEl.addBehavior(this);
   };
 }
+
+function objectOrArray (value) {
```
dmarcos: Maybe `isObjectOrArray` is a better name

ngokevin:
r?

PR #3101, opened 2017-09-30
add a missing verb

vincentfretin:
I'm not sure what you meant in this sentence, but there is definitely a verb missing.


ngokevin:
should be something like `enhances the behavior of`

vincentfretin:
Ok, I changed it.

Issue #3051, opened 2017-09-14
Increasing offset when entering VR for non positional devices


DavidHGillen:
**Description:** Every time the user enters the app or VR they are moved upwards by userHeight if one is anything other than 0. and if it is 0 then it is the default value of 1.6.

This bug is present on non positional devices like GearVR, chrome and firefox.

- A-Frame Version: 0.6.1+, not present in 0.6.0
- Platform / Device: Android Chrome, Android Firefox55, Gear VR. Not present w/ VIVE. Others not tested.
- Reproducible Code Snippet or URL:
```
<html>
<head>
	<script src="https://rawgit.com/aframevr/aframe/master/dist/aframe-master.js"></script>
</head>
<body>
	<a-scene>
		<a-plane position="0 -1.6 0" rotation="-90 0 0" width="1.8288" height="1.8288" color="#A48080"></a-plane>
		<a-plane position="0 0 0" rotation="-90 0 0" width="1.8288" height="1.8288" color="#A4A4A4"></a-plane>
		<a-plane position="0 1.6 0" rotation="-90 0 0" width="1.8288" height="1.8288" color="#80A480"></a-plane>
		<a-plane position="0 3.2 0" rotation="-90 0 0" width="1.8288" height="1.8288" color="#80C680"></a-plane>
	</a-scene>
</body>
</html>
```
https://youtu.be/awjY0GMPox8

ngokevin:
Not quite understanding the issue. userHeight can't be disabled, not applyingl, incorrectly applying>

DavidHGillen:
Here's a video if that makes it clearer:
https://youtu.be/awjY0GMPox8

delapuente:
Confirmed. It's happening to me too:

![increasinf-offset](https://user-images.githubusercontent.com/757942/30964996-ed1239e0-a453-11e7-9b95-59c9c77bf4e4.gif)

vincentfretin:
I can confirm there is an issue with 0.7.0 with GearVR and Oculus brower. No issue with 0.6.0 and 0.6.1.
In 0.7.0, no issue with `<a-camera user-height="0" />`, there is an issue with `<a-camera user-height="1" />` (or any value > 0), it increases each time by user-height when you enter, exit, enter vr.

vincentfretin:
The regression should be in a commit between 19 july (post 0.6.1 release) and 14 september (date of this issue) I guess.

vincentfretin:
[...]

PR #3105, opened 2017-10-01
Fix increasing offset when entering VR for non positional devices

vincentfretin:
This fixes issue #3051

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -69,7 +69,7 @@ module.exports.Component = registerComponent('look-controls', {
    */
   getUserHeight: function () {
     var el = this.el;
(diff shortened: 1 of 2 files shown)


vincentfretin:
If `el.getAttribute('camera').userHeight` returned 0, `getUserHeight()` always returned DEFAULT_CAMERA_HEIGHT instead of 0.

` <a-entity camera look-controls />` now renders again userHeight to 0 like it's supposed to.

vincentfretin:
But maybe this fix introduces a regression for #3000, @olga-microsoft can you please verify?

ngokevin:
thanks!

vincentfretin:
Ah, I just looked at it again. It doesn't fix the issue with `<a-camera />` which set camera.userHeight to 1.6 by default". I still have an increasing offset. This fixes only the case where you really want a camera on the floor.

PR #3108, opened 2017-10-02
deprecate states, remove state mixin feature (fixes #2958)

### LINKED ISSUE #2958: consider removing *state* mixins (e.g., `cube-hovered`)
**Description:**

What state mixins do is when a state is added `.addState` (e.g., `hovered`), this code will take every mixin (e.g., `foo`), and add a new mixin `foo-hovered` which the developer can use to automatically bind some components. Vice-versa, on `.removeState`, `foo-hovered` would be removed.

```html
<script>
AFRAME.registerComponent('someScript', {
  init: function () {
    this.el.addEventListener('mouseenter', () => { this.addState('hovered'); }
    this.el.addEventListener('mouseleave', () => { this.removeState('hovered'); }
  }
});
</script>

<a-scene>
  <a-assets>
    <a-mixin id="cube" material="color: red"></a-mixin>
    <a-mixin id="cube-hovered" material="color: blue"></a-mixin>
  </a-assets>

  <a-entity mixin="cube" someScript></a-entity>
</a-scene>
```

Reasons to consider removing:

- Roundabout: half in JS, half in DOM. We add/remove state via JavaScript in one place, have a mixin in another place with the same state name in another place, attach the mixin to the entity in another place. When the state updates, a mixin will be added to the entity at runtime.
- Keeps track of data within a string (space-delimited list of mixins): every time it fetches the new mixins, it has to run `.split` operations creating new arrays. Then needs to run diff operations between two arrays. Also relies on Mutation Observers.
- Not used: I don't think I or anyone else has used this feature.

I think an alternative would be to keep this type of logic at the application level, how the application determines state and how entities respond to that state. For example, an application might do instead make it declarative, explicit, and all on the entity:

```html
<a-entity
  state-bind-add__hovered="material.color: red"
  state-bind-remove__hovered="material.color: blue">
```

Or just do everything entirely in JS, removing having to proxy through this States API:

```
<script>
AFRAME.registerComponent('someScript', {
  init: function () {
    this.el.addEventListener('mouseenter', () => { this.setAttribute('material', 'color', 'red' }
    this.el.addEventListener('mouseleave', () => { this.removeState('material', 'color', 'blue'); }
  }
});
</script>
```

Here's the code proposed to remove, but will currently keep the `addState`/`removeState` API:

```js
  mapStateMixins: {
    value: function (state, op) {
      var mixins;
      var mixinIds;
      var i;

      mixins = this.getAttribute('mixin');

      if (!mixins) { return; }
      mixinIds = mixins.split(' ');
      for (i = 0; i < mixinIds.length; i++) {
        op(mixinIds[i] + '-' + state);
      }
      this.updateComponents();
    }
  },

  /**
   * Handle update of mixin states (e.g., `box-hovered` where `box` is the mixin ID and
   * `hovered` is the entity state.
   */
  updateStateMixins: {
    value: function (newMixins, oldMixins) {
      var diff;
      var newMixinsIds;
      var oldMixinsIds;
      var i;
      var self = this;

      newMixinsIds = newMixins.split(' ');
      oldMixinsIds = (oldMixins || '') ? oldMixins.split(' ') : [];

      // List of mixins that might have been removed on update.
      diff = oldMixinsIds.filter(function (i) { return newMixinsIds.indexOf(i) < 0; });

      // Remove removed mixins.
      diff.forEach(function (mixinId) {
        // State Mixins
        var stateMixinsEls = document.querySelectorAll('[id^=' + mixinId + '-]');
        Array.prototype.forEach.call(stateMixinsEls, function (el) {
          self.unregisterMixin(el.id);
        });
      });

      // Add new mixins.
      this.states.forEach(function (state) {
        newMixinsIds.forEach(function (id) {
          var mixinId = id + '-' + state;
          self.registerMixin(mixinId);
        });
      });
    }
  },
```

ngokevin:
**Description:**

#2958 

The original intention was to add something similar to pseudo CSS selectors. But:

1. State mixins are a scattered / roundabout way of applying DOM data based on an event. The same can be accomplished more simply with events. I use [event-set](https://github.com/ngokevin/kframe/tree/master/components/event-set) and it's easy.
2. State mixins can be slow. Any time a state is added, the entity force refreshes all of its components despite data not having changed. This caused some unexpectedness for me when components were trying to get "updated" with the same data.
3. `.addState`/`.removeState`/`.is` can be done with classes or data attributes while being visible from the DOM.
4. Usage is low. The appearances outside A-Frame are mostly when people copied code from the original `tracked-controls` example. a-blast/a-painter/a-saturday-night does not use states.
5. Less maintenance. The cursor component had lots of states and had to be maintained alongside the events which accomplished the same thing.

**Changes proposed:**
- Remove state mixins auto-appliers.
- Deprecate states.

### DIFF
--- a/docs/core/entity.md
+++ b/docs/core/entity.md
@@ -151,23 +151,6 @@ console.log(entity.sceneEl === sceneEl);  // >> true.
 
 ## Methods
 
-### `addState (stateName)`
-
-`addState` will push a state onto the entity. This will emit the `stateadded`
-event, and we can check the state can for existence using `.is`:
-
-
-```js
-entity.addEventListener('stateadded', function (evt) {
-  if (evt.detail.state === 'selected') {
-    console.log('Entity now selected!');
-  }
-});
-
-entity.addState('selected');
-entity.is('selected');  // >> true
-```
-
 ### `emit (name, detail, bubbles)`
 
 [animation-begin]: ./animations.md#begin
@@ -438,25 +421,6 @@ AFRAME.registerComponent('example-light', {
     // object3DMap.light is now null.
   }
 });
-```
-
-### `removeState (stateName)`
-
-`removeState` will pop a state from the entity. This will emit the
-`stateremoved` event, and we can check the state its removal using `.is`:
-
-```js
-entity.addEventListener('stateremoved', function (evt) {
-  if (evt.detail.state === 'selected') {
-    console.log('Entity no longer selected.');
-  }
-});
-
-entity.addState('selected');
-entity.is('selected');  // >> true
-
-entity.removeState('selected');
-entity.is('selected');  // >> false
 ```
 
 ## Events
--- a/examples/showcase/tracked-controls/components/aabb-collider.js
+++ b/examples/showcase/tracked-controls/components/aabb-collider.js
@@ -58,12 +58,13 @@ AFRAME.registerComponent('aabb-collider', {
       collisions.forEach(handleHit);
       // No collisions.
       if (collisions.length === 0) { self.el.emit('hit', {el: null}); }
-      // Updated the state of the elements that are not intersected anymore.
+
       this.collisions.filter(function (el) {
         return collisions.indexOf(el) === -1;
-      }).forEach(function removeState (el) {
-        el.removeState(self.data.state);
+      }).forEach(function (el) {
+        el.emit('unhit');
       });
+
       // Store new collisions
       this.collisions = collisions;
 
@@ -90,7 +91,6 @@ AFRAME.registerComponent('aabb-collider', {
 
       function handleHit (hitEl) {
         hitEl.emit('hit');
-        hitEl.addState(self.data.state);
         self.el.emit('hit', {el: hitEl});
       }
 
--- a/examples/showcase/tracked-controls/components/grab.js
+++ b/examples/showcase/tracked-controls/components/grab.js
@@ -7,7 +7,6 @@
 */
 AFRAME.registerComponent('grab', {
   init: function () {
-    this.GRABBED_STATE = 'grabbed';
     // Bind event handlers
     this.onHit = this.onHit.bind(this);
     this.onGripOpen = this.onGripOpen.bind(this);
@@ -45,7 +44,7 @@ AFRAME.registerComponent('grab', {
     var hitEl = this.hitEl;
     this.grabbing = false;
     if (!hitEl) { return; }
-    hitEl.removeState(this.GRABBED_STATE);
+    hitEl.emit('ungrab');
     this.hitEl = undefined;
   },
 
@@ -55,7 +54,7 @@ AFRAME.registerComponent('grab', {
     // If the hand is not grabbing the element does not stick.
     // If we're already grabbing something you can't grab again.
     if (!hitEl || hitEl.is(this.GRABBED_STATE) || !this.grabbing || this.hitEl) { return; }
-    hitEl.addState(this.GRABBED_STATE);
+    hitEl.emit('grab');
     this.hitEl = hitEl;
   },
 
--- a/examples/showcase/tracked-controls/index.html
+++ b/examples/showcase/tracked-controls/index.html
@@ -6,20 +6,20 @@
     <meta name="description" content="Tracked Controls – A-Frame">
     <script src="../../../dist/aframe-master.js"></script>
     <script src="components/aabb-collider.js"></script>
+    <script src="components/cube.js"></script>
     <script src="components/grab.js"></script>
     <script src="components/ground.js"></script>
+    <script src="https://unpkg.com/aframe-event-set-component@4.0.1/dist/aframe-event-set-component.min.js"></script>
     <script src="shaders/skyGradient.js"></script>
   </head>
   <body>
     <a-scene fog="color: #bc483e; near: 0; far: 65;">
       <a-assets>
         <a-mixin id="cube"
+                 event-set__grab="color: #FFE646"
+                 event-set__hit="color: #F2E646"
                  geometry="primitive: box; height: 0.30; width: 0.30; depth: 0.30"
                  material="color: #EF2D5E;"></a-mixin>
-        <a-mixin id="cube-collided"
-                 material="color: #F2E646;"></a-mixin>
-        <a-mixin id="cube-grabbed"
-                 material="color: #F2E646;"></a-mixin>
       </a-assets>
       <!-- Hands -->
       <a-entity hand-controls="left" aabb-collider="objects: .cube;" grab></a-entity>
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -13,12 +13,6 @@ var EVENTS = {
   MOUSEUP: 'mouseup'
 };
 
-var STATES = {
-  FUSING: 'cursor-fusing',
-  HOVERING: 'cursor-hovering',
-  HOVERED: 'cursor-hovered'
-};
-
 var CANVAS_EVENTS = {
   DOWN: ['mousedown', 'touchstart'],
   UP: ['mouseup', 'touchend']
@@ -83,11 +77,7 @@ module.exports.Component = registerComponent('cursor', {
   },
 
   remove: function () {
-    var el = this.el;
-    el.removeState(STATES.HOVERING);
-    el.removeState(STATES.FUSING);
     clearTimeout(this.fuseTimeout);
-    if (this.intersectedEl) { this.intersectedEl.removeState(STATES.HOVERED); }
     this.removeEventListeners();
   },
 
@@ -272,16 +262,12 @@ module.exports.Component = registerComponent('cursor', {
     this.intersectedEl = intersectedEl;
 
     // Hovering.
-    cursorEl.addState(STATES.HOVERING);
-    intersectedEl.addState(STATES.HOVERED);
     self.twoWayEmit(EVENTS.MOUSEENTER);
 
     // Begin fuse if necessary.
     if (data.fuseTimeout === 0 || !data.fuse) { return; }
-    cursorEl.addState(STATES.FUSING);
     this.twoWayEmit(EVENTS.FUSING);
     this.fuseTimeout = setTimeout(function fuse () {
-      cursorEl.removeState(STATES.FUSING);
       self.twoWayEmit(EVENTS.CLICK);
     }, data.fuseTimeout);
   },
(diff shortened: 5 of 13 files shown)


vincentfretin:
FYI, super-hands currently uses this feature.

dmarcos:
The `tracked-controls` examples also uses this features

ngokevin:
It's easy to port. I can port the tracked-controls example while removing lines of code. Don't think because a couple places uses this prevents us from cleaning up API.

```js
      <a-assets>
        <a-mixin id="cube"
                 geometry="primitive: box; height: 0.30; width: 0.30; depth: 0.30"
                 material="color: #EF2D5E;"
                 event-set__hit="material.color: #F2E646"
                 event-set__grab="material.color: #F2E646"></a-mixin>
      </a-assets>
```

And then delete all the state-related JS code.

ngokevin:
Would even propose to remove the state API overall in general since data attributes or classes can do the same thing.

dmarcos:
I still want to give the API a chance. I think it is useful, similar to CSS pseudo selectors. The state API is used in the A-Frame code itself. It can be implemented too with data attributes but it means more coupling with the DOM and not as terse and elegant as doing `el.is('alive')`

ngokevin:
Every time a state is added (e.g., the cursor component), even if I'm using mixins at all, the mixin state updates will do a mass component update (`.updateComponents`), which is a large waste of perf. I just ran into a bug just now because it was running a mass `.updateComponents` even when no relevant data has changed any time state was added/removed, I had no mixins, and no event listeners.

People haven't been "relying" on it as much as it was used in tracked-controls examples and we shipped it, and then the code was copied and pasted around. I wouldn't say because it was useful. A-Frame is inherently DOM so coupling is not a concern, and even so I'd rather see this done outside the core repo if it's an experiment. The cursor and vr-mode can be simply ported to just get rid of states and use a data attribute/class/special method.

I'm not sold on the CSS psuedo selector idea. It's a scattered/roundabout way of updating component data. We have JS code that adds state, we have to add a mixin in `<a-assets>`, and then another mixin based on that mixin name with a suffix, and then apply the mixins on the entity. It's much less magical to have a JS component or e.g., my `event-set` component.

ngokevin:
super-hands does not rely on states that heavily. The codebase has states because a lot of the code was based/copied on the tracked-controls collider and cursor. Actual usage of the states is low in the codebase.

donmccurdy:
> Every time a state is added (e.g., the cursor component), even if I'm using mixins at all, the mixin state updates will do a mass component update (.updateComponents), which is a large waste of perf. I just ran into a bug just now because it was running a mass .updateComponents even when no relevant data has changed any time state was added/removed, I had no mixins, and no event listeners.

I'm not 100% following here... why were states triggering component updates? Just to explore options, couldn't we keep the State API but not use it this way in Mixins? I.e., still have methods for `addState`, `removeState`, and `is`, which fire events, but not tie mixins into that?

dmarcos:
Yeah. I found the state API useful specially to develop UIs. We're providing pretty high level UI components like `hand-controls` and `laser-controls` that emit events and also modify the state of the entities. Event driven architectures are sometimes confusing and is useful to have a way to explicitly query the state of an entity. The state API is handy and very low footprint. My vote to keep it.

ngokevin:
Here's the footprint, consider you use states instead of events, which run frequently:

1. [`addState`](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L822) calls [`mapStateMixins`](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L836)
2. [`mapStateMixins`](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L137) registers a new mixin for every mixin the entity has, and then it runs [`this.updateComponents`](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L150) which does a mass component update.
3. [`registerMixin`](https://github.com/aframevr/aframe/blob/master/src/core/a-node.js#L150) as a side-effect also creates a Mutation Observer.
4. An event is emitted.

The similar happens on `removeState`. So if everything is running on the state API rather than events as we often use now, these run on any thing that happens in the application. By providing both states and events, there's redundancy on how to handle things, and I think events are fine for handling these.

The way states are used today, are pretty much for events. In `tracked-controls`, all it is used for is "color this when the cube is hit, color this when the cube is unhit", it doesn't use the `.is()`. The `laser-controls` examples don't use it either, nor the painter/blast/saturday. If we want to query whether something has happened, it should be up to the application on how to keep the state.

ngokevin:
Just testing on one very simple constrained event in `tracked-controls` example:

## With States (Before)

Changing the color of a cube on mouseenter takes **8ms**. Bottleneck on `updateProperties` calls from which is what state mixins. `buildData` shows up high on the call tree and memory allocator.

screen shot 2017-10-12 at 9 53 39 pm
screen shot 2017-10-12 at 9 51 48 pm

## Removed States (After)

Changing the color of a cube on mouseenter takes **6ms**. No large `updateProperties` call. `buildData` does not show up on the call tree and memory allocator. And simpler user-facing API IMO with `event-set`.

screen shot 2017-10-12 at 9 58 28 pm

screen shot 2017-10-12 at 10 05 38 pm

donmccurdy:
I don't have any issue with removing the state mixins feature. Ok to go ahead, IMO.

But the case for deprecating states entirely is less clear. It is kind of nice being able to just write `el.is('cursor-hovered')` without needing to have been listening to that element from the beginning. What if my component was created after the hovering began? It's nice that it is a very small API.

donmccurdy:
More specifically, I'm still hopeful states will be helpful for managing transitions between animations, NPCs, and little things like that. I've been planning to do more to test that, but have been blocked by (1) poor support for FBX/COLLADA animation, and (2) lack of a way to export multiple animations from glTF.

ngokevin:
I'm not as hardly opposed to keeping states, but I'm not quite comfortable either. I like not having special ways to do things that can already be done easily (e.g., data attributes, data sets, classes, a variable in a component, a variable in system). It also signals to developers "this is the way to do states". It's not quite how I'm handling state, and I don't think we've built enough applications to decide it's the best way. At least painter/blast/saturdaynight are not using them.

donmccurdy:
> I don't think we've built enough applications to decide it's the best way. At least painter/blast/saturdaynight are not using them.

I agree we haven't researched enough to decide what the best way is now. But, we might as well keep the status quo (i.e. not delete an API that has been around a while) until we have something to recommend. Is there anything in painter/blast/saturdaynight that we'd suggest users imitate?

donmccurdy:
> It also signals to developers "this is the way to do states".

Pretty weak signal, given that it's used so rarely. 😉 I don't think it hurts anything to keep this API around until we can make a more definitive suggestion on how to do states.

machenmusik:
Until we don't need to do `.is('vr-mode')` that way, we still need states, right?

ngokevin:
Yeah, it's definitely easier to keep the state API in. But I do see though like if you leave it in, there are components like `super-hands` that start to use it, and people copy/paste, and then it spreads. Then if you ever want to change it, it becomes harder down the road. But I guess we can do something when we get there. If we keep states, are we OK with removing it from like `cursor.js`? Or wonder if we can agree about the state mixins?

In the PR, I replaced the `.is('vr-mode')`with data attributes, but that was just showing what it would look like if we removed completely.

ngokevin:
They still feel a bit redundant. The addState/removeState does the same thing as events (e.g., `addState('hovered'); removeState('hovered');` vs `emit('hover'); emit('unhover')`, except that it keeps track of a boolean. I don't think keeping track of a boolean is necessary to introduce an API largely similar to events.

Several ways to do the same in a separate component without prescribing components to both emit state changes and events. The cursor component already does this, for every event, it also has to maintain state stuff. These examples, the states are only maintained on demand:

```html
<a-entity state__hovered="on: mouseenter; off: mouseenter">
<!-- el.getAttribute('state__hovered'); -->
```

```html
<a-entity state="on: triggerdown; off: triggerup; name: click">
<!-- el.dataset.click or el.getAttribute('data-click') or el.states.click or el.is('click'), whatever. -->
```

Looking to at least remove the state mixins though. For performance (measured above), and also I'm experiencing weirdness where because the cursor component uses states, when I hover, the entity force-refreshes all of its components, and I get odd behaviors of entities jittering around.

donmccurdy:
> Looking to at least remove the state mixins though.

Again, certainly, +1 for removing state mixins. 

>  I don't think keeping track of a boolean is necessary to introduce an API largely similar to events.

The question is whether now is the right time to delete an API we've already introduced. I do not think we should, until we have docs and examples to point to a replacement way of doing things and determine that it's better than what we're currently doing. The `state` component you show looks OK, but would need some thought about how to use it programmatically. `el.components.state.is('foo')`? Is that `state` component going to be in Core?  I don't think this all needs to be one PR.

wmurphyrd:
Was the state mixin feature ever documented? I remember staring at the tracked controls example source dumbfounded about how the color changed until I found state mixins code in core. 

I've already been working towards deprecating the only real dependency on state in super hands (i.e. the interface with sphere-collider), so I won't miss it. 

I did push state mixins as the easiest way to customize reactions in super hands. Would it be possible to pull `event-set` into core when dropping state mixins? It's nice to have a non-JS option for newcomers to add basic interactivity without adding another dependency

dmarcos:
`event-set` is opening a can of worms by introducing logic into the DOM a la angular. I think it should remain as a 3rd party component.

wmurphyrd:
I can always bundle it in super-hands

ngokevin:
I can split it. It's less risky to keep it but clarity is nice, and if you plan on deprecating something, better to do it early before it spreads via copy/paste. Just wanted to discuss while it's up.

The programmatic is described in the examples: a `.getAttribute` call (`el.getAttribute('state__foo')`), no extra API, and follows the same as the rest of A-Frame. No need to depend on the emitting component to implement states, components just emit as usual.

Would not put it in core as I haven't seen much usage. Most instances of set/removeState are from copy/pasting tracked-controls example code. I don't think it's a problem transitioning that small audience of the API since they maintain states but do not use them in applications. Deprecating an API is fine since upgrading is a deliberate step, and we just add migration notes/tips in the changelog.

I'll open a PR for the state mixins though.

donmccurdy:
> ```
> <a-entity state__hovered="on: mouseenter; off: mouseenter">
> 
> ```

This syntax doesn't work — the component still needs to be present while it's waiting for the event to fire, and presumably after it turns off.

Not to say we can't make a `state` component workable.. just needs some thought. Having this API spread only to eventually move it to a third-party component would be no worse, IMO, than having the community components all handle state in slightly different ways, so that there's no consistent API.

ngokevin:
I wouldn't introduce event-set in core, but FWIW, I've had no issues with event-set being hard to follow for toggling simple things. Comparing to Angular is extreme where it was messy due to the two-way data bindings that propagated changes throughout the application all in DOM. The way event-set is used, I wouldn't call logic as much as CSS pseudo-selectors are.

> This syntax doesn't work — the component still needs to be present while it's waiting for the event to fire, and presumably after it turns off.

I think that's a lot easier to control than having to depend on the component implement states. Some will, some won't. At least events are dependable on being available. States will always be handled in slightly separate ways, there are too many different ways to keep track of a boolean.

ngokevin:
Closing for now after state mixins removal and will re-evaluate later, thanks!

PR #3111, opened 2017-10-02
only set this.controls.userHeight if device has positional tracking

vincentfretin:
This definitely fix issue #3051 
I don't know how to write a test for this, this seems bound to the VRControls and how it renders in the browser...

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -59,12 +59,19 @@ module.exports.Component = registerComponent('look-controls', {
 
   tick: function (t) {
     var data = this.data;
-    var hasPositionalTracking = this.hasPositionalTracking !== undefined ? this.hasPositionalTracking : checkHasPositionalTracking();
+    var hasPositionalTracking;
+
     if (!data.enabled) { return; }
+
     this.controls.standing = data.standing;
+
+    hasPositionalTracking = this.hasPositionalTracking !== undefined
+      ? this.hasPositionalTracking
+      : checkHasPositionalTracking();
     if (hasPositionalTracking) {
       this.controls.userHeight = this.getUserHeight();
     }
+
     this.controls.update();
     this.updateOrientation();
     this.updatePosition();
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -1,7 +1,10 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var utils = require('../utils/');
 var DEFAULT_CAMERA_HEIGHT = require('../constants').DEFAULT_CAMERA_HEIGHT;
 var bind = require('../utils/bind');
+
+var checkHasPositionalTracking = utils.device.checkHasPositionalTracking;
 
 // To avoid recalculation at every mouse movement tick
 var GRABBING_CLASS = 'a-grabbing';
@@ -56,9 +59,12 @@ module.exports.Component = registerComponent('look-controls', {
 
   tick: function (t) {
     var data = this.data;
+    var hasPositionalTracking = this.hasPositionalTracking !== undefined ? this.hasPositionalTracking : checkHasPositionalTracking();
     if (!data.enabled) { return; }
     this.controls.standing = data.standing;
-    this.controls.userHeight = this.getUserHeight();
+    if (hasPositionalTracking) {
+      this.controls.userHeight = this.getUserHeight();
+    }
     this.controls.update();
     this.updateOrientation();
     this.updatePosition();
--- a/tests/__init.test.js
+++ b/tests/__init.test.js
@@ -8,12 +8,13 @@ window.debug = true;
 navigator.getVRDisplays = function () {
   var resolvePromise = Promise.resolve();
   var mockVRDisplay = {
-    requestPresent: resolvePromise,
+    cancelAnimationFrame: function (h) { return window.cancelAnimationFrame(1); },
+    capabilities: {},
     exitPresent: resolvePromise,
-    submitFrame: function () { return; },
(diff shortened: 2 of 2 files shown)


dmarcos:
The proper solution for this is probably moving the logic to the restore pose of the camera component: https://github.com/aframevr/aframe/blob/master/src/components/camera.js#L99

vincentfretin:
@dmarcos There is already a hasPositionalTracking check in camera to save the pose (position and rotation of the camera) yes. So you mean saving userHeight? But this is set on the VRControls object, not the camera. I'm really lost with this, and I don't have any positional tracking device to test this, only GearVR. 

What I know it that before #3000, there was no `this.controls.userHeight = this.getUserHeight();` line in look-controls, so the `this.controls.userHeight` stayed at 0.0. Adding a condition to it solved the issue for GearVR.

machenmusik:
It feels to me as though we are trying to bake too many smarts into the `camera` -- when it comes to uses for render textures and spectator views, the `camera` is just that - not meant to be a stand-in for the user and presumed HMD positioning.  I thought that (in current incarnation) `look-controls` was intended to contain the latter, and that from prior discussions we thought this area might need some refactoring.  My recollection is that recent changes have handed to use absolute positioning not relative positioning for room scale etc., and this repeated application of relative offset may be a vestige of doing things the old way that should not be re-introduced regardless of where?

dmarcos:
All right. I think we should then move the logic to save / restore the camera pose / position from the `camera` to look controls as well as the `userHeight` property.

vincentfretin:
One other thing that seems to differ between positional and non positional tracking devices is that with positional tracking you are placed to z -2 or something like that, and with GearVR you are probably at z: 0. So there are some examples, notably https://github.com/wmurphyrd/aframe-super-hands-component/issues/89 where you are too close to the object, here with the big cube.

DavidWyand:
I just wanted to add that this fix works for me.  It allows my project to work on GearVR with the Oculus browser, and the Vive using Firefox 56.0, without having to make any camera changes.  Thanks!

vincentfretin:
@dmarcos Maybe we can merge this and you folks work eventually on a refactoring in another branch?

ngokevin:
Do this to fix tests: https://github.com/aframevr/aframe/pull/3149/files#diff-f9aac19c37c38417114cd801ed8016c4R12

ngokevin:
Fixed the tests.

@dmarcos this fix make sense? We can refactor later.

olga-microsoft:
@vincentfretin sorry for breaking GearVR and thank you for issuing the fix. I verified that it works as expected with Window Mixed Reality.

src/components/look-controls.js
```diff
@@ -56,9 +59,19 @@ module.exports.Component = registerComponent('look-controls', {
...
+
     if (!data.enabled) { return; }
+
     this.controls.standing = data.standing;
-    this.controls.userHeight = this.getUserHeight();
+
+    hasPositionalTracking = this.hasPositionalTracking !== undefined
+      ? this.hasPositionalTracking
+      : checkHasPositionalTracking();
+    if (hasPositionalTracking) {
```
ngokevin: Trying to understand. How come we still want to  apply user height if there is positional tracking? Seems to me you don't want to apply an offset if your height is tracked, or we should check more specifically that there are no stage parameters? And how come it was adding up more and more each time enter VR?

src/components/look-controls.js
```diff
@@ -1,8 +1,11 @@
 var registerComponent = require('../core/component').registerComponent;
 var THREE = require('../lib/three');
+var utils = require('../utils/');
 var DEFAULT_CAMERA_HEIGHT = require('../constants').DEFAULT_CAMERA_HEIGHT;
 var bind = require('../utils/bind');
```
cvan: feel free to change this to `utils.bind`

src/components/look-controls.js
```diff
@@ -1,8 +1,11 @@
...
 
+var checkHasPositionalTracking = utils.device.checkHasPositionalTracking;
```
cvan: do you want to import this function above?

src/components/look-controls.js
```diff
@@ -56,9 +59,19 @@ module.exports.Component = registerComponent('look-controls', {
...
   tick: function (t) {
     var data = this.data;
+    var hasPositionalTracking;
+
     if (!data.enabled) { return; }
+
     this.controls.standing = data.standing;
-    this.controls.userHeight = this.getUserHeight();
+
+    hasPositionalTracking = this.hasPositionalTracking !== undefined
```
cvan: if this is already expect to be a Boolean, you can simplify this expression

tests/__init.test.js
```diff
@@ -8,12 +8,13 @@ window.debug = true;
...
+    cancelAnimationFrame: function (h) { return window.cancelAnimationFrame(1); },
+    capabilities: {},
     exitPresent: resolvePromise,
-    submitFrame: function () { return; },
-    getPose: function () { return { orientation: null, position: null }; },
+    getPose: function () { return {orientation: null, position: null}; },
     requestAnimationFrame: function () { return 1; },
-    cancelAnimationFrame: function (h) { return window.cancelAnimationFrame(1); }
+    requestPresent: resolvePromise,
+    submitFrame: function () { return; }
```
cvan: can you add a test for checking the tracking method and user height?

dmarcos:
VRControls / VREffect will be soon gone. See https://github.com/aframevr/aframe/pull/3152 I leave it open as a reference and a reminder to port the logic. Nothing to do here.

dmarcos:
After the removal of VREffect / VRControls this PR is not applicable anymore.

PR #3115, opened 2017-10-03
remove two array allocations per deepEqual call (critical path)

ngokevin:
**Description:**

From the browser performance tab, deepEqual was a large memory allocator. After this patch, deepEqual no longer even registers on the memory tab.

utils.deepEqual is called on every single component update. Saves two arrays per component update. For things like look-controls which updates pos/rot every frame, saves a lot.

**Changes proposed:**
- Don't use Object.keys in utils.deepEqual, use the same two arrays.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -4,6 +4,7 @@ var debug = require('./debug');
 var deepAssign = require('deep-assign');
 var device = require('./device');
 var objectAssign = require('object-assign');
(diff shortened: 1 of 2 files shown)


dmarcos:
Failing tests

ngokevin:
Fixed, I moved the object pooling utility from the other PR since the deepEqual call can get recursive, there's an uncertain amount of arrays we'd need to create.

PR #3119, opened 2017-10-04
Update cursor.md

weddingdj:
**Description:**

In this case it is raycaster-interected-cleared and not raycaster-intersection-cleared

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -23,7 +23,7 @@ The cursor component listens to events and keeps state on what's being hovered
 and pressed in order to provide `mousedown`, `mouseup`, `mouseenter`,
 `mouseleave`, and `click` events. We use the name `mouse` to mimic
 traditional web development for now. Under the hood, the cursor component uses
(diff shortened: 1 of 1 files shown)


vincentfretin:
@weddingdj This fix is wrong.
In cursor.js, there is 
```
el.addEventListener('raycaster-intersection', this.onIntersection);
el.addEventListener('raycaster-intersection-cleared', this.onIntersectionCleared);
```
So I think the error was in the first event in the sentence, ``raycaster-intersected` should be `raycaster-intersection`. For me, the sentence should be:
```
Under the hood, the cursor component uses the `raycaster-intersection` and `raycaster-intersection-cleared` events, capturing the closest visible intersected entity.
```

vincentfretin:
The raycaster component is emitting raycaster-intersected and raycaster-intersected-cleared on the intersected entity, yes, but the cursor component is not using these events.

vincentfretin:
Please merge #3122 if you agree.

PR #3125, opened 2017-10-06
batch raycaster-intersection events, reuse event detail

ngokevin:
**Description:**

**Changes proposed:**
- Cache event detail object.
- Emit one raycaster-intersection event per tick for all intersected entities.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -249,11 +249,13 @@ module.exports.Component = registerComponent('cursor', {
     var index;
     var intersectedEl;
     var intersection;
+    var raycaster;
 
     // Select closest object, excluding the cursor.
-    index = evt.detail.els[0] === cursorEl ? 1 : 0;
-    intersection = evt.detail.intersections[index];
-    intersectedEl = evt.detail.els[index];
+    raycaster = cursorEl.components.raycaster;
+    index = raycaster.intersectedEls[0] === cursorEl ? 1 : 0;
+    intersection = raycaster.intersections[index];
+    intersectedEl = raycaster.intersectedEls[index];
 
     // If cursor is the only intersected object, ignore the event.
     if (!intersectedEl) { return; }
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -54,6 +54,7 @@ module.exports.Component = registerComponent('raycaster', {
     // adjust line length.
     this.clearedIntersectedEls = [];
     this.unitLineEndVec3 = new THREE.Vector3();
+    this.intersections = [];
     this.intersectedEls = [];
     this.objects = [];
     this.prevCheckTime = undefined;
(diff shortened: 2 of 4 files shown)


machenmusik:
(This may break existing code, right?)

vincentfretin:
Yes definitely, this will break super-hands for GearVR. super-hands is already broken with the merge of #3126

dmarcos:
@vincentfretin Is it too hard for `super-hands` to accommodate to the new API?

ngokevin:
Yeah, didn't see super-hands coming. I can try making a small PR.

machenmusik:
This is one of those features where it is rarely used, but for any that do, the event detail is always used at the moment. Would it be feasible to determine if there was anything listening and if not just don’t emit at all?

ngokevin:
Updated to use @dmarcos' cached event pattern.

machenmusik:
Nice!

tests/components/raycaster.test.js
```diff
@@ -207,16 +207,6 @@ suite('raycaster', function () {
...
-      var raycasterEl = el;
-      raycasterEl.addEventListener('raycaster-intersection', function (evt) {
-        assert.equal(evt.detail.els[0], targetEl);
-        assert.equal(evt.detail.intersections[0].object.el, targetEl);
-        done();
-      });
-      component.tick();
-    });
-
     test('emits event on intersected entity with details', function (done) {
```
vincentfretin: You should put back this test, no?

vincentfretin:
@dmarcos @ngokevin I made a small PR to super-hands https://github.com/wmurphyrd/aframe-super-hands-component/pull/80

src/components/raycaster.js
```diff
@@ -148,98 +151,84 @@ module.exports.Component = registerComponent('raycaster', {
...
-      // Emit intersected on intersected entity per intersected entity.
-      for (i = 0; i < intersections.length; i++) {
-        intersections[i].object.el.emit('raycaster-intersected', {
-          el: el,
-          intersection: intersections[i]
-        });
-      }
+    // Emit intersected on intersected entity per intersected entity.
+    for (i = 0; i < intersections.length; i++) {
+      intersections[i].object.el.emit('raycaster-intersected', {
```
dmarcos: we can use `intersectionDetail` for this one too

src/components/cursor.js
```diff
@@ -233,9 +233,9 @@ module.exports.Component = registerComponent('cursor', {
     var intersection;
 
     // Select closest object, excluding the cursor.
-    index = evt.detail.els[0] === cursorEl ? 1 : 0;
-    intersection = evt.detail.intersections[index];
-    intersectedEl = evt.detail.els[index];
+    index = cursorEl.components.raycaster.intersectedEls[0] === cursorEl ? 1 : 0;
```
dmarcos: temp variable `var raycasterComponent = cursorEl.components.raycaster`?

src/components/raycaster.js
dmarcos: We probably need an `intersectedDetail` object

ngokevin: This one is different because it's different data per intersected entity.

ngokevin:
Updated.

ngokevin:
This has been done as part of another one of my PRs https://github.com/aframevr/aframe/pull/3250

Issue #3124, opened 2017-10-05
laser-controls line does not respect intersection line length


DavidWyand:
**Description:**
The line component that is part of laser-controls does not change its length to the intersection distance as determined by the raycaster component.  What should happen is the line component shortens itself to not extend past the nearest intersection point.  What ends up happening is the line component remains at the raycaster's far value.

- A-Frame Version: 0.7.0
- Platform / Device: Tested with Vive on Firefox 56.0, and GearVR with GearVR controller using Oculus' browser
- Reproducible Code Snippet or URL: May be seen in link-traversal showcase example

Spent a long time trying to figure out why the laser-controls component wasn't correctly changing its laser line length in my own project.  But then I found that the link-traversal example also exhibits the same issue: its laser lines go right through the link portals.

I've found a couple of reasons for this.  One I solved, and the other I worked around.

The first issue I found was that the raycaster stopped calling its drawLine() method as soon as there was a valid intersection.  The reason for this is due to a _return_ call in its tick() method.  Change the _return_ in this code block starting at line 218 in raycaster.js from:

```
// Emit intersection cleared on both entities per formerly intersected entity.
for (i = 0; i < prevIntersectedEls.length; i++) {
  if (intersectedEls.indexOf(prevIntersectedEls[i]) !== -1) { return; }
  el.emit('raycaster-intersection-cleared', {el: prevIntersectedEls[i]});
  prevIntersectedEls[i].emit('raycaster-intersected-cleared', {el: el});
}
```

to a _break_:

```
// Emit intersection cleared on both entities per formerly intersected entity.
for (i = 0; i < prevIntersectedEls.length; i++) {
  if (intersectedEls.indexOf(prevIntersectedEls[i]) !== -1) { break; }
  el.emit('raycaster-intersection-cleared', {el: prevIntersectedEls[i]});
  prevIntersectedEls[i].emit('raycaster-intersected-cleared', {el: el});
}
```

This allows for drawLine() to be properly called with the correct length, which should have been it.  Unfortunately, there appears to be a deeper issue with A-Frame for the next reason.

Within raycaster's drawLine() method there is the following at line 307:

    el.setAttribute('line', lineData);

[...]

PR #3127, opened 2017-10-06
fix raycaster line update that wasn't triggered anymore (refs #3124)

vincentfretin:
I implemented the solution proposed in https://github.com/aframevr/aframe/issues/3124#issuecomment-334589365


vincentfretin:
I just rebased because #3127 was merged.

ngokevin:
Thanks!

Issue #3131, opened 2017-10-06
emit the intersection cleared events before the intersection events


ngokevin:
**Description:**

If you are changing intersection between two overlapping entities, from one entity to another, then right now the cleared events will fire after the intersection events.

Imagine you are setting a hover color to white on intersection, and back to black on clear. Right now:

1. intersection start with entity 1: white
2. intersection start with entity 2: white
3. intersection end with entity 1: black

What will end up happening is while we expect our entity to now be black since it's still intersecting an entity, it is now white due to the order What we want is:

1. intersection start with entity 1: white
2. intersection end with entity 1: black
3. intersection end with entity 2: white

PR #3134, opened 2017-10-06
fix cleared events not emitted in some cases

vincentfretin:
From the comment https://github.com/aframevr/aframe/commit/0b60c12bd48de67c5e82eb1c19159800a07f1ca1#commitcomment-24807425
we changed the `return` to a `break`, but it should really be a `continue`.

I tested with two overlapping cubes, and with the break, sometime I had a cube that stayed in a hover state and the ray wasn't on it. I couldn't reproduce the issue by replacing `break` with `continue`.

#3131 is maybe related to this?


dmarcos:
Thanks!

Issue #2981, opened 2017-08-16
Propagate 'model' property from daydream-controls to laser-controls component?


evhan55:
Would it be possible to propagate the `model` property from the `daydream-controls` component up to the `laser-controls` component so that we can hide the controller model but still use laser controls?

https://aframe.io/docs/0.6.0/components/laser-controls.html
https://aframe.io/docs/0.6.0/components/daydream-controls.html

donmccurdy:
Do you still want tracking from daydream controls? If so this should work:

```html
<a-entity laser-controls="hand: left" daydream-controls="model: false"></a-entity>
```

Or just a floating laser that you can use however you want?

```html
<a-entity cursor="fuse: false" raycaster="showLine: true"></a-entity>
```

... but overriding `model:false` for all the different controller types is sort of a pain, so I'd be OK with surfacing a `model` property on `laser-controls` too.

evhan55:
Yes! I would still like Daydream controls, I had already tried this before posting the issue and it didn't seem to work:

```html
<a-entity laser-controls="hand: left" daydream-controls="model: false"></a-entity>
```

I just tried again and it didn't work again, no controller model, but also no tracked laser pointer either.

However, the whole reason I was trying to have Daydream-tracked lasers without a model was because I was trying to reduce lag on loading of the 3D controller model, in trying to resolve a bug where the controller wasn't tracking at all for a colleague of mine.  But, overnight I realized that we had just forgotten to enable the `gamepad-extensions` flag on their Pixel device.

Some follow ups:

- For people who still may want to have controller tracking, laser pointing, but no controller model, the workaround that you posted would be fine (for me, anyway) if it worked as expected, but it doesn't seem to at the moment.  Should I open a new issue for that?
- It might be useful for the A-Frame controller docs to put in reminders that the `webvr` and `gamepad-extensions` flags might need to be set on mobile (and possibly other headsets?)..

Thank you!

machenmusik:
> ... but overriding model:false for all the different controller types manually is sort of a pain

Exactly... If we want to have easy to use laser-controls, we should add values that should cascade downward

ngokevin:
For edge and custom cases, I'd copy and paste the laser-controls component and do what you need rather than adding more API. laser-controls is an out of the box implementation and the code is very simple, it sets other components. If you need to create your own controls, you can modify it or add anothe component that overrides daydream-controls

evhan55:
Thanks, @ngokevin ! Makes sense.

[...]

PR #3138, opened 2017-10-09
Laser controls cleanup

wmurphyrd:
**Description:** Improve the cleanup of `laser-controls` on removal and allow an option to not load controller models. I'm suggesting this to improve my `progressive-controls` component, which uses `laser-controls` for pointing mode. I have an issue where the raycaster from laser-controls cannot be disabled for touch mode because it is added back on event triggers even if laser-controls is removed. I have also had a feature request for the model property from a user.

**Changes proposed:**

- When `laser-controls` is removed, also remove registered event listeners
- To accomplish the above, move helper functions from within `init` into component methods
- Add a `model` schema property that propagates to controller components

### DIFF
--- a/docs/components/laser-controls.md
+++ b/docs/components/laser-controls.md
@@ -56,9 +56,10 @@ to the distance to the intersection point.
 
 ## Properties
 
-| Properties | Description        |
-|------------|--------------------|
-| hand       | `left` or `right`. |
+| Properties | Description                                         | Default |
+|------------|-----------------------------------------------------|---------|
+| hand       | `left` or `right` when two controllers are present. | `right` |
+| model      | Whether the default controller models are loaded.   | `true`  |
 
 ## Customizing the Raycaster
 
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -4,19 +4,21 @@ var bind = utils.bind;
 
 registerComponent('laser-controls', {
   schema: {
-    hand: {default: 'right'}
+    hand: {default: 'right'},
+    model: {default: true}
   },
 
   init: function () {
     var data = this.data;
     var el = this.el;
+    var controllerConfig = {hand: data.hand, model: data.model};
 
     // Set all controller models.
-    el.setAttribute('daydream-controls', {hand: data.hand});
-    el.setAttribute('gearvr-controls', {hand: data.hand});
-    el.setAttribute('oculus-touch-controls', {hand: data.hand});
-    el.setAttribute('vive-controls', {hand: data.hand});
-    el.setAttribute('windows-motion-controls', {hand: data.hand});
+    el.setAttribute('daydream-controls', controllerConfig);
+    el.setAttribute('gearvr-controls', controllerConfig);
+    el.setAttribute('oculus-touch-controls', controllerConfig);
+    el.setAttribute('vive-controls', controllerConfig);
+    el.setAttribute('windows-motion-controls', controllerConfig);
 
     // Bind methods
     this.createRay = bind(this.createRay, this);
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -1,5 +1,6 @@
 var registerComponent = require('../core/component').registerComponent;
 var utils = require('../utils/');
+var bind = utils.bind;
 
 registerComponent('laser-controls', {
   schema: {
@@ -7,10 +8,8 @@ registerComponent('laser-controls', {
   },
 
   init: function () {
-    var config = this.config;
     var data = this.data;
     var el = this.el;
-    var self = this;
 
     // Set all controller models.
     el.setAttribute('daydream-controls', {hand: data.hand});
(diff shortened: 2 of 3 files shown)


src/components/laser-controls.js
```diff
@@ -1,69 +1,79 @@
...
-        el.setAttribute('raycaster', 'showLine', true);
-      }
-
-      el.setAttribute('cursor', utils.extend({
-        fuse: false
-      }, controllerConfig.cursor));
+    // Wait for controller to connect, or have a valid pointing pose, before creating ray
+    el.addEventListener('controllerconnected', this.createRay);
+    el.addEventListener('controllerdisconnected', this.hideRay);
+    el.addEventListener('controllermodelready', this.createRay);
```
dmarcos: Why two separate event listeners attached to the same event?

wmurphyrd: Fixed in 59d2430

wmurphyrd:
I just found #2981 in which @ngokevin advises against adding the model property. Sorry I didn't search issues before sending this. Shall I strip out the commits related to the model property and just keep the `remove` cleanup?

docs/components/laser-controls.md
```diff
@@ -56,9 +56,10 @@ to the distance to the intersection point.
 
 ## Properties
 
-| Properties | Description        |
-|------------|--------------------|
-| hand       | `left` or `right`. |
+| Properties | Description                                         | Default |
+|------------|-----------------------------------------------------|---------|
+| hand       | `left` or `right` when two controllers are present. | `right` |
+| model      | Whether the default controller models are loaded.   | `true`  |
```
dmarcos: The position of the laser is model specific and letting users override the model won't play gracefully

dmarcos:
What is the use case for hiding the controller model?

ngokevin:
Looking back at the conversation, people should be able to do:

`<a-entity laser-controls daydream-controls="model: false">`

to disable. That should work, but there were some reports of it not.

wmurphyrd:
@dmarcos: I think it is because they want to load their own model instead of the controllers, like the guns in A-Blast. @omgitsraven @evhan55 Can you comment on use cases for which you wanted `model: false` on `laser-controls` / `progressive-controls`?

@ngokevin: yeah, but that gets a lot more verbose when we're talking about setting it for all the controllers loaded by `laser-controls`

omgitsraven:
The idea was to provide my own model instead, yeah. Something relevant to the scene, that would be used instead of either Vive or Touch controllers.

docs/components/laser-controls.md
wmurphyrd: It looks like there was some effort put into allowing the model to specify its own offset
[original](https://github.com/aframevr/aframe/blob/master/src/components/laser-controls.js#L43)/[proposed](https://github.com/aframevr/aframe/pull/3138/files#diff-3f66fed14508fae90dc58b66fac99025R51)

ngokevin:
re: a-blast For something complex as guns, you'd want to use your own controller component. There's no need for laser/raycaster there. All it is is setting your own model.

re:verbose It's not too bad really, to just set the components. It'd take like 15 seconds. Alternatively, copy and paste the `laser-controls` component, all it does is set other components. It's very easy. I'm in favor of not adding more API or emphasizing components that can be done easily manually, to have multiple ways to do the same thing.

dmarcos:
We could have write an article on on how to make custom components using `a-blast` as a cases study. I opened https://github.com/aframevr/aframe-site/issues/469

wmurphyrd:
What I'm really invested in here is the improved removal cleanup, so I'm
going to strip out the `model` related commits and force push an update
with just the cleanup

evhan55:
In my case, I had a user who was on a satellite link, and the laser was not showing up at all.  I thought for a while that it was the controller model geometry needed to be downloaded in order for laser controls to work, so I was looking for ways to disable the model-loading, since the laser itself was more important.

I think I was wrong about the reason it wasn't showing up, though, but I did feel limited by needing to load extra geometry for low-bandwidth cases.

evhan55:
Having articles explaining how to make custom components sounds like a great idea!

machenmusik:
For better or worse, I think the examples like `laser-controls` work well enough that folks don't try to clone it and make their own versions like a-blast... so if that a-blast example will show how easy it is (and expose some of the annoyances like having to keep your own list of the higher level controls you need to prospectively add) there will probably be less questions / requests here going forward

ngokevin:
It's already in the docs https://aframe.io/docs/0.7.0/introduction/interactions-and-controllers.html#creating-custom-controllers

evhan55:
@ngokevin Ooh yeah, I guess I meant specific examples for a few use cases to see it in action. Even though they are discoverable in the a-frame source, it's not usually my first impulse to check there.

ngokevin:
I think these changes make laser-controls try to do too much, which was an almost reference out-of-the-box implementation that all it does is set other components. It's pretty much a config with no logic. If people need anything deeper than that, I'd start with a copy of laser-controls as a base and modify. If people need to do things such as disable the laser or remove models, that can be done with `setAttribute` calls. So I'd vote to close.

wmurphyrd:
I reverted the dependency, so I no longer have any stake in this.

PR #3142, opened 2017-10-09
In the text it says 1.6 while in the table 0

nikgraf:
**Description:**

Fixed the docs.

### DIFF
--- a/docs/components/camera.md
+++ b/docs/components/camera.md
@@ -27,7 +27,7 @@ A camera situated at the average height of human eye level (1.6 meters).
 | far        | Camera frustum far clipping plane.                                                                                              | 10000         |
 | fov        | Field of view (in degrees).                                                                                                     | 80            |
 | near       | Camera frustum near clipping plane.                                                                                             | 0.005         |
-| userHeight | How much height to add to the camera when *not* in VR mode. The default camera has this set to 1.6 (meters, to represent average eye level.). | 0             |
+| userHeight | How much height to add to the camera when *not* in VR mode. The default camera has this set to 1.6 (meters, to represent average eye level.). | 1.6             |
 | zoom       | Zoom factor of the camera.                                                                                                      | 1             |
 
 ## VR Behavior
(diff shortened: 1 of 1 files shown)


ngokevin:
Actually the default camera generated by A-Frame has a user height of 1.6, but using the camera component itself has a default user height of 0. Perhaps we can reword to make it clear.

vincentfretin:
The documentation is correct. Nonetheless if you're using non positional tracking device with aframe 0.7.0, and you add a <a-camera/> yourself in the scene, your eyes will be at 1.6 because of the following issue https://github.com/aframevr/aframe/pull/3111

nikgraf:
Any suggestions re-word this? It indeed is confusing 😄

ngokevin:
`Height offset to add to the camera when *not* in VR mode so the camera is not in the ground. The default camera that A-Frame injects or the `<a-camera>` primitive has this set to 1.6 (meters, to represent average eye level.). Note the default camera component alone defaults this to 0.`

nikgraf:
@ngokevin thx I updated the PR, but changed the wording `is not in the ground` to `is not on ground level`

I still don't understand what is said here due the fact: I don't understand the difference between the camer when no in VR and default camera. Can you elaborate the difference?

docs/components/camera.md
```diff
@@ -27,7 +27,7 @@ A camera situated at the average height of human eye level (1.6 meters).
 | far        | Camera frustum far clipping plane.                                                                                              | 10000         |
 | fov        | Field of view (in degrees).                                                                                                     | 80            |
 | near       | Camera frustum near clipping plane.                                                                                             | 0.005         |
-| userHeight | How much height to add to the camera when *not* in VR mode. The default camera has this set to 1.6 (meters, to represent average eye level.). | 0             |
+| userHeight | Height offset to add to the camera when *not* in VR mode so the camera is not on ground level. The default camera that A-Frame injects or the primitive has this set to 1.6 (meters, to represent average eye level.). Note the default camera component alone defaults this to 0. | 1.6             |
```
ngokevin: the default is still 0

nikgraf:
@ngokevin ohhh, now after reading this new section everything makes sense! thanks for the explaination

PR #3143, opened 2017-10-09
Cursor update: new raycaster API and touch handling for rayOrigin: mouse

### LINKED ISSUE #3114: cursor="rayOrigin: mouse" does not work on touchscreens
**Description:**

- A-Frame Version: 
A-frame 7.0
- Platform / Device: 
Windows 10 - Firefox 55 and Chrome 61
Android - Chrome Browser 61
https://jsfiddle.net/8jmcuaLw/4/

When using a touchscreen the 'click' event does not register properly. 
In above url try clicking the blue box on a touch screen.
Following would be expected:
 - When clicked the blue box should change to a random hue.
What actually happens:
- the 'touch'/'click' does not register on first touch.
- It registers second time but this time I can trigger the click event on the box by clicking anywhere on the screen. Even besides the box.

### LINKED ISSUE #3140: cursor intersections never clear
**Description:** `cursor` needs to be updated for the new `raycaster` api

- A-Frame Version: master
- Platform / Device: any
- Reproducible Code Snippet or URL: https://github.com/aframevr/aframe/blob/master/src/components/cursor.js#L276

I noticed this while working on #3114. I'll include this fix in the PR

wmurphyrd:
**Not yet complete - need advice**

**Description:** Improve handling of touch input for `rayOrigin: mouse` mode. Make necessary updates for compatibility with `raycaster` post #3126.

Fixes #3140 
Fixes #3114 

**Changes proposed:**

- Bind touch events to onMouseMove handler
- Alter handler to find coordinates in either MouseEvent or TouchEvent touch 0
- Update onIntersectionCleared to handle array of entities and `clearedEls` property name
- Cancel handled touch events so that mobile browsers do not emulate duplicate MouseEvents

**Problem:**

Adapting `cursor` to handle touch events was straightforward, but it didn't completely solve the problem. For a mouse click, we know the mouse position in advance, and the raycaster has had time to refresh. For a touch, however, we don't know the position until the `touchstart`. As the raycaster only refreshes intersections on an interval, it is outdated when `onCursorDown` checks the current intersection. The current effect is that, with each new touch, the `cursor` events/states will effect entities under the previous touch location, always lagging one touch behind. 

I see two potential solutions, and I'd like to know which the maintainers favor:

1. Set the touch event handling on a timeout equal to the current raycaster interval so that they are called after the next raycaster refresh
2. Modify `raycaster` component to allow on-demand refresh (i.e. move everything from `tick` after the interval check into a separate method)

Other concerns:

The current touch tracking is simplistic: starts on the first touch, ignores additional new touches, and ends when any touch ends (even if that first touch is still going). Two other options that could be implemented without major changes would be to keep track of that first touch in a state variable in order to not end until it ends or to update the ray location with every new touch but still end on the first `touchend`.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -151,15 +151,21 @@ module.exports.Component = registerComponent('cursor', {
     el.removeEventListener('raycaster-intersection', this.onIntersection);
     el.removeEventListener('raycaster-intersection-cleared', this.onIntersectionCleared);
     window.removeEventListener('mousemove', this.onMouseMove);
+    window.removeEventListener('touchstart', this.onMouseMove);
+    window.removeEventListener('touchmove', this.onMouseMove);
     window.removeEventListener('resize', this.updateCanvasBounds);
   },
 
   updateMouseEventListeners: function () {
     var el = this.el;
     window.removeEventListener('mousemove', this.onMouseMove);
+    window.removeEventListener('touchstart', this.onMouseMove);
+    window.removeEventListener('touchmove', this.onMouseMove);
     el.setAttribute('raycaster', 'useWorldCoordinates', false);
     if (this.data.rayOrigin !== 'mouse') { return; }
     window.addEventListener('mousemove', this.onMouseMove, false);
+    window.addEventListener('touchstart', this.onMouseMove, false);
+    window.addEventListener('touchmove', this.onMouseMove, false);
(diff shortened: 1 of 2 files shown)


ngokevin:
We can split the tick out to a separate method that can be manually called. I'll work on merging what we have here now though first. Thanks!

ngokevin:
Thanks! We can do the manual raycaster intersection check as a follow up.

PR #3147, opened 2017-10-10
Support the latest Web Browser for the ODG R-7.

machenmusik:
The changes here use an updated version 0.9.40 of the polyfill that make the accelerometer / gyroscope based sensor fusion work properly on the latest Web Browser for the ODG R-7, which is a landscape-primary device, as well as making A-Frame realize that it should be using it for R-7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "style-attr": "^1.0.2",
     "three": "^0.87.0",
     "three-bmfont-text": "^2.1.0",
-    "webvr-polyfill": "chenzlabs/webvr-polyfill#r7"
+    "webvr-polyfill": "^0.9.40"
   },
   "devDependencies": {
     "browserify": "^13.1.0",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "style-attr": "^1.0.2",
     "three": "^0.87.0",
     "three-bmfont-text": "^2.1.0",
-    "webvr-polyfill": "^0.9.36"
+    "webvr-polyfill": "chenzlabs/webvr-polyfill#r7"
   },
   "devDependencies": {
     "browserify": "^13.1.0",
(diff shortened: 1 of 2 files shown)


dmarcos:
Can we open a PR on the polyfill repo?

machenmusik:
ahead of you, https://github.com/googlevr/webvr-polyfill/pull/286

machenmusik:
once @jsantell merges into polyfill, if we are using latest master of polyfill in general, then the package.json change can be reverted

dmarcos:
Good! I prefer to wait for googlevr/webvr-polyfill#286 to land and just bump the `webvr-polyfill`. Cleaner and less work.

dmarcos:
Pointing dependencies to private of forks should be just a last resource

machenmusik:
understood.  looks like it just merged, let me retry without package.json, hang tight

jsantell:
just published webvr-polyfill v0.9.40 with the changes to support the R7 :metal:

machenmusik:
nice!  OK so now the latest ODG R-7 Web Browser can use polyfill with correct accelerometer/gyro behavior, so you can load up https://r7-aframe.glitch.me/ and look at the boxes to select them.  (Although the display is still monoscopic at the moment.)

machenmusik:
@dmarcos OK to merge, thx @jsantell !

package.json
```diff
@@ -32,9 +32,9 @@
   "dependencies": {
     "@tweenjs/tween.js": "^16.8.0",
     "browserify-css": "^0.8.2",
-    "debug": "ngokevin/debug#noTimestamp",
+    "debug": "github:ngokevin/debug#noTimestamp",
```
dmarcos: Why is this necessary?

package.json
```diff
@@ -32,9 +32,9 @@
...
     "deep-assign": "^2.0.0",
-    "document-register-element": "dmarcos/document-register-element#8ccc532b7",
+    "document-register-element": "github:dmarcos/document-register-element#8ccc532b7",
```
dmarcos: Why is this change necessary?

machenmusik:
those github: changes shouldn't have been necessary, I undid by checking out an older version, let me fix

machenmusik:
ok, package.json fixed

machenmusik:
looks correct now:
- bump polyfill to 0.9.40, which has R-7 changes
- make A-Frame use accelerometer and gyro for R-7 (with polyfill fixes), like mobile phones
- tweak landscape utils call to work for R-7

dmarcos:
Thanks! 🏅

PR #3153, opened 2017-10-11
fix data from dom not parsed if skipTypeChecking (fixes #3150)

### LINKED ISSUE #3150: component values from DOM never parsed due to skipTypeChecking
**Description:**

`<a-entity id="raycaster" raycaster="origin: 0 0 0">`

This data gets stored in `attrValue` cache as `{origin: '0 0 0'}`.

```js
var updateObj = {interval: 100};
raycasterEl.setAttribute(updateObj);
updateObj.interval = 200;
raycasterEl.setAttribute(updateObj);
```

Now every time the component goes to update with the reused objects, `attrValue.origin` is used as is, and is never parsed. As a result, we get: `{origin: '0 0 0', interval: 200}` when they merge.

ngokevin:
**Description:**

**Changes proposed:**
- Only skip type checking if there is not extra data coming from DOM that needs parsing.
- Simplify and fix attrValue update mutating the object used for setAttribute.

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -156,15 +156,16 @@ Component.prototype = {
     if (value === undefined) { return; }
 
     // Merge new data with previous `attrValue` if updating and not clobbering.
-    if (!isSinglePropSchema && !clobber && this.attrValue) {
-      for (property in this.attrValue) {
-        if (!(property in attrValue)) {
-          attrValue[property] = this.attrValue[property];
-        }
+    if (!isSinglePropSchema && !clobber) {
+      this.attrValue = this.attrValue ? cloneData(this.attrValue) : {};
+      for (property in attrValue) {
+        this.attrValue[property] = attrValue[property];
       }
+      return;
     }
 
-    this.attrValue = extendProperties({}, attrValue, isSinglePropSchema);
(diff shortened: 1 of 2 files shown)


src/core/component.js
```diff
@@ -228,9 +230,25 @@ Component.prototype = {
     }
 
     isSinglePropSchema = isSingleProp(this.schema);
+
     // Disable type checking if the passed attribute is an object and has not changed.
     skipTypeChecking = attrValue !== null && typeof this.previousAttrValue === 'object' &&
                        attrValue === this.previousAttrValue;
+    if (skipTypeChecking) {
+      for (key in this.attrValue) {
```
dmarcos: Probably more efficient this way:
````javascript 
var oldKeys = Object.keys(this.attrValue);
var newKeys = Object.keys(attrValue);
var key;
if (oldKeys.length !== newKeys.length) { skipTypeChecking = false; }
for (var i =0; i < oldKeys.length; ++i) { 
  key = oldKeys[i];
  if (!(key in attrValue) { 
     skipTypeChecking = false;
     break;
  }
 }
````

ngokevin: That creates two arrays

Issue #2464, opened 2017-03-07
obj-model dependency injects material component


donmccurdy:
**Description:**

- A-Frame Version: 0.5.0
- Platform / Device: Chrome 56.
- Reproducible Code Snippet or URL:

Scenario: I have a simple component that does a thing to its host mesh. If the mesh is available at load, it applies then. If not, it waits for `object3dset`.

```html
<a-entity obj-model="obj: foo.obj; mtl: foo.mtl;" doer-of-things></a-entity>
```

```js
module.exports = AFRAME.registerComponent('doer-of-things', {
  init: function () {
    const mesh = this.el.getObject3D('mesh');

    if (!mesh) {
      this.el.addEventListener('object3dset', this.init.bind(this));
    }

    mesh.traverse(function(node) {
      // does things
    });
  }
});
```

Somewhat surprisingly, `mesh` is set from the beginning, and the warning `Material component properties are ignored when a .MTL is provided` *always* appears. This seems to be because `obj-model` depends on `material`, the `material` component gets added, and creates a Mesh for itself.

If I redesign `doer-of-things` to allow the underlying mesh to change, the problem goes away, but maybe there's enough of a gotcha here to be worth fixing. Ideas:

1. Does the `material` component need to create an empty mesh? Could it just hang onto a material reference until a mesh appears?
2. Should `obj-model` be allowed some sort of "optional" dependency on `material`, such that if the component isn't provided, it won't be added?

ngokevin:
I think both 1 and 2 are valid.

1. We can change material to wait for `object3Dset` if it doesn't exist rather than `getOrCreateObject3D('mesh').

2. We can move the `material` initialization into the handlers. Detect if there's a material trying to be set, and wait for it.

donmccurdy:
I think fixing (1) would be helpful, in any case. What do you mean by 'the handlers'? 

I'm also occasionally writing custom components to override properties on models (e.g. `envMap`, which PBR models expect but don't provide). Might be benefits to allowing the `material` component to override properties of existing materials rather than creating a new THREE.Material from scratch, but maybe out-of-scope or better to prototype as a separate component.

ngokevin:
Lifecycle handlers*

donmccurdy:
Another example of the problem:

http://stackoverflow.com/questions/43371917/how-do-i-access-the-attributes-of-buffergeometry-from-an-an-a-frame-component/43373818?noredirect=1#comment73820235_43373818

[...]

PR #3154, opened 2017-10-11
Better shader docs

### LINKED ISSUE #3151: Better documentation on `AFRAME.registerShader()` needed!
Per discussion at https://aframevr.slack.com/archives/C0GG937RN/p1507678947000225
The functionality of `AFRAME.registerShader()` has improved to the point where it is pretty useful, but the documentation for shaders is buried inside `material`, and the documentation for `registerShader()` does not have an example of recent changes which allow very straightforward usage without dipping down into `THREE` materials:
```
/* global AFRAME */
// Grid along X and Z (horizontal planes).

AFRAME.registerShader('grid', {
  schema: {
    color: {type: 'color', is: 'uniform', default: 'white'},
    opacity: {type: 'number', is: 'uniform', default: 1.0},
    thickness: {type: 'number', is: 'uniform', default: 1.0},
    interval: {type: 'number', is: 'uniform', default: 0.01}
  },

  raw: true,

  vertexShader: [
    'precision mediump float;',
    'attribute vec2 uv;',
    'attribute vec3 position;',
    'uniform mat4 projectionMatrix;',
    'uniform mat4 modelViewMatrix;',
    'uniform float interval;',
    'varying vec2 coord;',
    'void main(void) {',
    '  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);',
    '  coord = vec2(position.x / interval, position.z / interval);',
    '}'
  ].join('\n'),

  fragmentShader: [
    '#extension GL_OES_standard_derivatives: enable',
    'precision mediump float;',
    'uniform vec3 color;',
    'uniform float opacity;',
    'uniform float thickness;',
    'varying vec2 coord;',
    'void main() {',
    '  vec2 grid = abs(fract(coord - 0.5) - 0.5) / fwidth(coord);',
    '  float line = min(grid.x, grid.y);',
    '  if (line > thickness) { discard; } else',
    '  gl_FragColor = vec4(color, opacity);',
    '}'
  ].join('\n')
});
```

machenmusik:
Fix #3151.


docs/components/material.md
```diff
@@ -378,6 +451,137 @@ AFRAME.registerShader('custom', {
...
+component:
+
+```js
+AFRAME.registerComponent('custom-material', {
+  schema: {
+    // Add properties.
+  },
+
+  init: function () {
+    this.material = this.el.getOrCreateObject3D('mesh').material = new THREE.ShaderMaterial({
```
donmccurdy: This predates your PR, but i think `getOrCreateObject3D('mesh').material` is a bad pattern. It screws with components waiting for `object3dset`, for example. IMO this initializer should see if a mesh is available already, and if not wait for the `object3dset` method.

(this has the added benefit of making it compatible with custom models, which `registerShader` is not afaik)

ngokevin: How does it mess with components waiting for object3dset?

donmccurdy: See: https://github.com/aframevr/aframe/issues/2464

machenmusik: Point is, the framework should abstract all this away so you can say `shader:displacement` and let A-frame worry about when.  That is what `registerShader` allows, and this entire approach completely fails to do.

machenmusik: > if you use a traverse(), which registerShader cannot do
Shouldn't the `material` component be the one doing that?

donmccurdy: whynotboth

machenmusik: My thought was a simple `recursive:true` (or we can still call it traverse).  I am guessing we can detect the aframe entity object3D structure and refuse to cascade, but I haven't looked.

machenmusik: My thought was a simple `recursive:true` (or we can still call it traverse).  I am guessing we can detect the aframe entity object3D structure and refuse to cascade, but I haven't looked.

machenmusik: You can always make your own replacement for the material component, which obviously you do by default, but for the simple example cases that seems like overkill.

What I am not sure about is whether the model object trees have consistent DOM-ish naming, for better declarative mapping etc.  I think I saw discussion on that in another issue/PR

machenmusik:
OK so -

We seem to agree these docs are an improvement.

Both the examples (grid and displacement) are done both as registered shader and custom component, so we actually have @donmccurdy's comment `whynotboth` already.

Ok to merge while working #3155?

donmccurdy:
I haven't had the chance to do a thorough read-through yet, but assuming someone has reviewed for any errors then +1 from me for merging and not waiting on the other discussions.

docs/components/material.md
```diff
@@ -498,44 +504,231 @@ void main() {
 
 varying vec2 vUv;
 uniform vec3 color;
-uniform float time;
```
dmarcos: `time1000` is a weird name. Let’s keep time. It is already consistent with other time related variables that use miliseconds in the code base

machenmusik: So the reason I didn't use `time` is because that is commonly used to represent time, but in seconds not milliseconds.  Unfortunately the time type is processed as milliseconds, so I wanted to pick a different name and convert to the time in seconds that typical sharers expect.  Maybe `timeMsec` or something more descriptive?

machenmusik: To clarify, there is inherent support for type `time`, without using a `tick()` handler to update anything, but it is in milliseconds rather than seconds (which is what you generally see in common sharers from Shadertoy, Shaderfrog, etc.)

dmarcos: In the DOM and A-Frame we use miliseconds for time. I would keep the name `time`

docs/components/material.md
```diff
@@ -327,42 +327,115 @@ visual defects.
...
+
+[example]: https://codepen.io/machenmusik/pen/WZyQNj
+Let's walk through an [example CodePen][example] with step-by-step commentary.
+
+As always, we need to include the A-Frame script.
+Here, we use RawGit to include the latest master version.
 
 ```js
-AFRAME.registerComponent('custom-material', {
+<script src="https://rawgit.com/aframevr/aframe/master/dist/aframe-master.min.js"></script>
```
donmccurdy: Any reason we need to be suggesting the master version in the docs?

@ngokevin do versions get automatically find/replaced when we do a release?

docs/components/material.md
```diff
@@ -327,42 +327,115 @@ visual defects.
...
+```js
+    color: {type:'color', is:'uniform', default:'red'},
+    opacity: {type:'number', is:'uniform', default:1.0}
   },
+```
 
-  update: function () {
-    // Update `this.material`.
+Setting raw to true uses THREE.RawShaderMaterial instead of ShaderMaterial,
+so your shader strings are used as-is, for advanced shader usage.
```
donmccurdy: I'd link to the docs for [THREE.RawShaderMaterial](https://threejs.org/docs/#api/materials/RawShaderMaterial) and ShaderMaterial here.

> so your shader strings are used as-is, for advanced shader usage.

maybe better to be explicit: "so built-in uniforms and attributes are not automatically added to your shader code."

docs/components/material.md
```diff
@@ -498,44 +504,231 @@ void main() {
...
+  fragmentShader: fragmentShader
+});
+```
+
+that the vertex shader uses:
+
+```glsl
+// vertex.glsl
+
+#pragma glslify: pnoise3 = require(glsl-noise/periodic/3d)
```
donmccurdy: I probably wouldn't go to the trouble of explaining how to set it up here, but worth noting somewhere that:

> This example uses [Browserify](http://browserify.org/) and [glslify](https://github.com/glslify/glslify) to include  [glsl-noise](https://github.com/hughsk/glsl-noise) as a dependency of our shader.

(similar options are available for rollup etc.)

dmarcos:
We have an example of custom shaders already. Do we want to point to it in the docs? https://aframe.io/aframe/examples/test/shaders/

docs/components/material.md
machenmusik: No reason other than not having to change it every release.

docs/components/material.md
machenmusik: I may not be doing a good job of explaining.

The shader examples that are time-varying already use a uniform named `time` that is in seconds.  Rather than trying to force already working shaders to rename, what I am illustrating is how to do a simple two line addition that allows the shader code to be reused without modification.

docs/components/material.md
machenmusik: done

docs/components/material.md
machenmusik: done

machenmusik:
@dmarcos added that example

machenmusik:
OK, I think this looks good now, unless someone has a real example to contribute of a custom shader that uses a type that `registerShader` doesn't support

docs/components/material.md
machenmusik: Changed to `timeMsec`, and added comments in the shader to explain the conversion to seconds.

docs/introduction/shaders.md
```diff
@@ -0,0 +1,26 @@
+---
```
ngokevin: I think this part is too thin for a separate section. I think it's OK to be contained in the material section for now. Perhaps as the API and the ecosystem around shaders possibly matures, we move it out a separate section?

docs/components/material.md
```diff
@@ -516,26 +523,248 @@ void main() {
...
+  <a-sphere material="shader:grid-glitch; color: blue;"
             radius="0.5"
             position="0 1.5 -2">
   </a-sphere>
 </a-scene>
 ```
 
-* [Live demo](https://aframe-simple-shader.glitch.me/)
-* [Remix this on Glitch](https://glitch.com/edit/#!/aframe-simple-shader)
+* [Live demo](https://aframe-simpler-shader.glitch.me/)
```
ngokevin: All the Glitch demos can be listed in the `examples` array in the header:

```
examples:
  - title: Grid Glitch Shader
    url: https://glitch.com/edit/#!/aframe-simpler-shader?path=client/shader-grid-glitch.js:1:0
```

For ease of copy/pasting, I'd remove the GLSLify build step and use [multiline strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Multi-line_strings).

docs/components/material.md
ngokevin: I've found that when I need a custom material, I just use a raw component. I think @donmccurdy does the same? It's simpler than hooking into `material`. Should we recommend that?

machenmusik: The idea is that if you only need to use your own (vertex and fragment) shader, that is most easily done with `registerShader` and then `material="shader:my-shader"`.  The examples that were given previously are actually easier to do that way (which is hopefully illustrated).

Other examples where you only need shader and not custom material, besides the ones listed, include the typical top/bottom and left/right stereo shaders we used to recommend separate geometry and/or components for.

If `registerShader` doesn't work for various reasons, then it does recommend creating a custom component (with this example). 

I did ask for an example that actually requires custom component, but so far no one has suggested one.

docs/introduction/shaders.md
machenmusik: The non-duplicated part here that seems worth keeping is the reference to shader repositories like ShaderToy and/or ShaderFrog (there are probably others too) and pointers toward components that help with reusing those.  Where would you suggest that appear in `material` documentation?

docs/components/material.md
machenmusik: It was that way originally (using GLSL-ify) so I left it that way; you can see that the new one I added as walkthrough does, in fact, use multi-line strings.  Are you asking to make changes to the existing example instances that will retroactively alter documentation for prior versions?

docs/components/material.md
ngokevin: Nevermind, I saw this is already done as a raw component. I thought it was trying registerShader.

docs/introduction/shaders.md
ngokevin: Perhaps when shaders are first introduced.

docs/components/material.md
ngokevin: You can remix the Glitch and not alter the old one.

docs/introduction/shaders.md
machenmusik: Looks like your shaderfrog gallery went away, was that intentional?  Changed reference to just the shaderfrog site.

docs/components/material.md
machenmusik: FYI - current examples header looks pretty unattractive, and no other good examples to follow?

docs/introduction/shaders.md
ngokevin: Yeah, it was an empty repo 🙄

docs/components/material.md
ngokevin: It's a data structure, it gets rendered underneath the table of contents as a list. I think there are two Glitches you listed, can add more as we go.

ngokevin:
Thanks, will work on merging this soon.

docs/components/material.md
ngokevin: Just checked, there are more than two. Good list!

machenmusik: updated the Glitches I made, just the page and script, no more extraneous server and files

HawkenKing:
the shader docs are good, but it goes from 0 to 9000mph in one doc, the displacement shader example is far to advanced for a casual user just getting to grips with a-frame.

A useful example might be how to register THREE.MeshLambertMaterial as a shader.

PR #3164, opened 2017-10-14
input-mapping system

fernandojsg:
Added input-mapping system: https://github.com/fernandojsg/aframe-input-mapping-component

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -204,6 +204,7 @@
       <li><a href="test/geometry-gallery/">Geometry Gallery</a></li>
       <li><a href="test/geometry-merging/">Geometry Merging</a></li>
       <li><a href="test/gltf-model/">glTF Model</a></li>
+      <li><a href="test/input-mapping/">Input mapping</a></li>
       <li><a href="test/laser-controls/">Laser Controls</a></li>
       <li><a href="test/mixin/">Mixin</a></li>
       <li><a href="test/model/">Model <em>(glTF, OBJ, COLLADA)</em></a></li>
--- a/examples/test/input-mapping/index.html
+++ b/examples/test/input-mapping/index.html
@@ -1,0 +1,109 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>A-Frame Input Mapping Component - Basic</title>
+    <meta name="description" content="Basic example for Input Mapping component."></meta>
+    <script src="../../../dist/aframe-master.js"></script>
+  </head>
+  <body>
+    <a-scene>
+      <a-assets>
+        <img id="groundTexture" src="https://cdn.aframe.io/a-painter/images/floor.jpg">
+        <img id="skyTexture" src="https://cdn.aframe.io/a-painter/images/sky.jpg">
+      </a-assets>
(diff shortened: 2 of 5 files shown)


fernandojsg:
Not sure if it's the best idea to start the PR until we'll solve the issues we're discussing in my repo and then push all the changes, or merge it and evolve it here, moving the discussion to this repo too.
/cc @ngokevin @dmarcos thoughts?

dmarcos:
@fernandojsg I think we can deal with the issues in further iterations of the mapping. The current design is useful and easy to understand / use. I commented on https://github.com/fernandojsg/aframe-input-mapping-component/issues/3 that it is the only one that could change the current design, the others are refinements. We had discussed that problem before and settled on organizing the mappings per controller instead of per action for the reasons described in the issue: Organizing mappings per controller is closer to how users and developers reason about mappings. I grab a controller and I imagine how the interaction will feel and how the different actions map to the buttons.

cvan:
@trevorfsmith: since you’ve been assessing and preparing for tackling universal XR/MR input fragmentation, I wanted to get your insight on this - if you haven’t already take a look. thanks for taking a look :smile:

TrevorFSmith:
This looks like a good start and addresses a lot of the issues with mapping lots of input types to higher level semantic events.

A few notes:

When I wrote a similar input manager for Spaciblō (classes here: https://github.com/Spaciblo/spaciblo-core/blob/master/docroot/js/spaciblo-input.js and default schema here: https://github.com/Spaciblo/spaciblo-core/blob/master/docroot/js/spaciblo-default-input-schema.js) it became quickly apparent when making apps that I needed  to know how the higher level event was triggered. So, the app might treat differently events from left- or right-handed controller or from gaze control.

One of my working theories is that a MR app framework needs to keep track of global app state (type of pose data, type of display, type of controllers, etc) so that the input mapping can be entirely swapped out. For example, if a handset is in flat page mode then it needs a different input map handling touch inputs than it would in magic window pass-through AR mode which is also different from when in a HMD doing VR.  So, the same lower level inputs lead to different higher level events, depending on MR mode.

Diego, I suspect that this is related to the conversation that you're having with @blairmacintyre about the use of "presenting" to indicate global app state. It seems like we need state with more info, like a data structure that represents the current display, the UI mode (flat, overlay, in-scene), the reality (pass-through AR, see-through AR, VR), and the input schema.

fernandojsg:
I believe we could still handle these global states as presenting or not with the states implemented currently and the app could swap between these mapping states as the app change itself to another state.

ngokevin:
Maybe we should let the component/system API bake before we place in the core. We've been quick to put components into core before (object `pool`) that haven't gotten a lot of run. I'll use it in one of my projects and can provide feedback/PRs.

fernandojsg:
@ngokevin yep I agree with you. I'll keep improving it on my repo and testing it with the rest of the demos and once it get mature we could decide how to include it here

dmarcos:
The `input-mapping` component is still under heavy development. It is a bit premature to pull into core. I'm closing this for now and we will reconsider for `0.9.0`. Better let it mature as a 3rd party component for now.

PR #3169, opened 2017-10-15
Factor out onButtonEvent method from controlers

dmarcos:
It factors out the `onButtonEvent` method common to all controls. Other methods can be also factored out like `onButtonChanged` or `onAxisMoved`. I will do follow up PRs with those. I started with `onButtonEvent` for easier review

### DIFF
--- a/src/components/vive-controls.js
+++ b/src/components/vive-controls.js
@@ -2,8 +2,10 @@ var registerComponent = require('../core/component').registerComponent;
 var utils = require('../utils/');
 
 var bind = utils.bind;
-var checkControllerPresentAndSetup = utils.trackedControls.checkControllerPresentAndSetup;
-var emitIfAxesChanged = utils.trackedControls.emitIfAxesChanged;
+var trackedControlsUtils = require('../utils/tracked-controls');
+var checkControllerPresentAndSetup = trackedControlsUtils.checkControllerPresentAndSetup;
+var emitIfAxesChanged = trackedControlsUtils.emitIfAxesChanged;
+var onButtonEvent = trackedControlsUtils.onButtonEvent;
 
 var VIVE_CONTROLLER_MODEL_OBJ_URL = 'https://cdn.aframe.io/controllers/vive/vr_controller_vive.obj';
 var VIVE_CONTROLLER_MODEL_OBJ_MTL = 'https://cdn.aframe.io/controllers/vive/vr_controller_vive.mtl';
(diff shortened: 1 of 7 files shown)


src/utils/controls.js
```diff
@@ -0,0 +1,16 @@
+/**
```
ngokevin: There's a tracked-controls utils where all the controllers share code already

src/components/vive-controls.js
```diff
@@ -195,35 +197,18 @@ module.exports.Component = registerComponent('vive-controls', {
...
+    var isTouch;
     if (!this.data.model) { return; }
 
+    isTouch = evtName.indexOf('touch') !== -1;
     // Don't change color for trackpad touch.
     if (isTouch) { return; }
 
     // Update colors.
     color = evtName === 'up' ? this.data.buttonColor : this.data.buttonHighlightColor;
-    if (Array.isArray(buttonName)) {
```
ngokevin: Don't need the array handler?

src/utils/controls.js
dmarcos: It makes sense for me to separate `tracked-controls` from the higher level controls utils

src/components/vive-controls.js
dmarcos: I don’t think we need it. I don’t see it used anywhere. I believe we went ahead of ourselves and thought we would neef multiple events when a button is pressed. I can reintroduce but I couldn’t find where this code path is used.

dmarcos:
I think is a bit premature to do a registerController API. There’s plenty of ad-hoc logic to deal with the specifics of each controller and I don’t feel ready for abstraction. Factoring out a few functions will clean up and make code testable. We can reevaluate an`registerController` API a bit later.

src/utils/controls.js
```diff
@@ -0,0 +1,16 @@
...
+ * Handle a button event and reemits the events.
+ *
+ * @param {string} id - id of the button.
+ * @param {string} evtName - name of the reemitted event
+ * @param {object} component - reference to the component
+ * @param {string} hand - handedness of the controller: left or right.
+ */
+module.exports.onButtonEvent = function (id, evtName, component, hand) {
```
ngokevin: Concerned about passing in `component` and having a utility call a bunch of the component's methods.

For example, we did this with `checkControllerPresentAndSetup`, I found it hard to follow. The utils depends on `component.el`, `component.controllerPresent`, `component.addEventListeners`, `component.controllerEventsActive`, `component.injectTrackedControls`. There's a huge amount of interface expected to be available.

Since we want to make available these shared methods and have the controllers implement this common interface, we might want to think early about a higher level API/prototype mixin early. Splitting out to a util makes it seem refactored, but the util and component are heavily coupled and not any more testable. It might be the wrong road to go down.

machenmusik: Agree that although the attempt to refactor into utils allowed code reuse, it made it nearly impossible to understand what things were doing without a pretty thorough look

dmarcos: You're correct. It's essentially a mixin but passing the `this` as an argument rather than `this.foo()`. I want to stay focused on the task of improving controls. My goal is to easily be able to add `thumbpadup`, `thumbstickup` methods without having to copy code across controllers. There will be just three methods factored out from controllers. In a separate API we can think of a mixin mechanism that consolidates `checkControllerPresentAndSetup` and these factored out methods in a more sustainable pattern.

ngokevin: I'm not sure the utils improves the code quality of the controls since it makes things hard to understand, I'd favor copy and pasting code. Ideally, if we're refactoring, we might as well refactor in the right direction. We can merge something temporary if we agree on the mechanism and someone promises to do it soon.

ngokevin: Perhaps not agree on the specific code, but that we can have a mixin-style refactor instead soon after you are able to finish the thumbpad/stick work. Then mostly OK with something temporary, these methods are at least less coupled than the tracked-controls utils.

src/components/vive-controls.js
machenmusik: I believe that was originally there for buttons that were aliased to trigger more than one named event (e.g. 'menu' and 'app') but then we resolved the naming such that there were no duplicates.  So it should not currently be used anywhere

dmarcos:
I consolidated the new methods in `tracked-controls` utils

PR #3171, opened 2017-10-16
remove state mixins

ngokevin:
**Description:**

Split from #3108. For performance and reducing bug potential from force refreshing all components when a state or mixin is added/removed. And to reduce API surface and favor more predictable APIs.

**Changes proposed:**
- Remove state mixins.
- Update tracked-controls to not use state mixins for toggling colors. I use event-set here, open to other ways (e.g., vanilla JS) to perhaps not have a dependency in the examples.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -5,7 +5,6 @@ var THREE = require('../lib/three');
 var utils = require('../utils/');
 
 var AEntity;
-var bind = utils.bind;
 var debug = utils.debug('core:a-entity:debug');
 var warn = utils.debug('core:a-entity:warn');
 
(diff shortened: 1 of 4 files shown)


src/core/a-entity.js
```diff
@@ -833,8 +774,7 @@ var proto = Object.create(ANode.prototype, {
       var stateIndex = this.states.indexOf(state);
       if (stateIndex === -1) { return; }
       this.states.splice(stateIndex, 1);
-      this.mapStateMixins(state, bind(this.unregisterMixin, this));
-      this.emit('stateremoved', {state: state});
+      this.emit('stateremoved', state);
```
donmccurdy: Intentional? I assumed we want to discuss recommendations for an alternative state API outside of this PR, so breaking changes to the current API don't seem immediately useful.

donmccurdy:
One comment, otherwise LGTM

src/core/a-entity.js
ngokevin: We can save an object in case the API does stick around for a version or further even if an application is not using it. It's not looking like there'll be consensus resulting in removing this API at the moment.

PR #3179, opened 2017-10-19
add cursor event listeners to canvas, not window

ngokevin:
**Description:**

Cursor only cares about actual touch events.

This fixes an issue where vive-controls touchstart bubbled up to window and cursor was receiving them.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -150,22 +150,25 @@ module.exports.Component = registerComponent('cursor', {
     });
     el.removeEventListener('raycaster-intersection', this.onIntersection);
     el.removeEventListener('raycaster-intersection-cleared', this.onIntersectionCleared);
-    window.removeEventListener('mousemove', this.onMouseMove);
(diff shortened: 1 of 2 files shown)


wmurphyrd:
Another option would be to register those `onMouseMove` listeners on the `canvas` like the `onCursorDown`/`Up` listeners, as this would take them out of the bubble path of controllers and other scene elements.

I thought it was odd they were on the window the last time I was in there. Perhaps just an artifact of the merge from the previous mouse cursor component.

ngokevin:
Thanks, better idea.

ngokevin:
Needs another review due to changing the solution.

dmarcos:
The problem with this approach is that if you have an embedded scene that it's small on screen it won't be a good experience.

dmarcos:
Oh sorry. looking at wrong file

dmarcos:
Failing tests

ngokevin:
Fixed. The embedded scene example works fine, same as before for me.

PR #3152, opened 2017-10-11
Remove VREffect / VRControls in favor of the new WebGLRenderer API

dmarcos:
Tested on:

- [x] Desktop no headset
- [x] Vive
- [x] Oculus
- [x] Windows Mixed Reality
- [x] Daydream
- [x] GearVR
- [x] Cardboard (Android)
- [x] Cardboard (iOS)

Moving THREE dependency to the dev branch temporarly until r88 release.

PR #3181, opened 2017-10-19
Webxr refactor

joshmarinacci:
**WebXR bindings**

**Changes proposed:**
- add XREffect
- add XRControls
- add xr-anchor component


dmarcos:
Thanks! `VREffect` / `VRControls` will be gone in the next A-Frame version (see https://github.com/aframevr/aframe/pull/3152). I recommend starting making a WebXRManager similar to the WebVRManager that replaces VREffect / VRControls (https://github.com/mrdoob/three.js/blob/dev/src/renderers/webvr/WebVRManager.js). WebXR is still highly experimental so I would keep it as a separate `aframe-xr.js` library that bundles both the WebXRManager and the webxr-polyfill.js. At load, the library would swap WebVRManager with the WebXRManager in the renderer: `sceneEl.renderer.vr = new WebXRManager()` and register the relevant components for ar scenarios.

ngokevin:
Agreed with not tunneling down into that deep of an experimental level, and three.js has moved away from Effect/Controls, so this direction won't be viable regardless. We can discuss in an issue next time so we don't waste your efforts (unless you were planning on doing this anyways). Thanks!

PR #3196, opened 2017-10-26
Make camera system aware of mixins.

### LINKED ISSUE #3090: The camera component can not be used inside a mixin
The following example fails:
https://codepen.io/lodr/pen/yzbvba

It is caused by:
```html
<a-mixin id="c" camera></a-mixin>
```

And the console says:
```
TypeError: cameraEl.pause is not a function
[Learn More]
```

captura de pantalla 2017-09-28 a las 13 57 36

delapuente:
Fix #3090

Prevents the camera system from taking into account a-mixin elements when
handling the active camera.

### DIFF
--- a/src/systems/camera.js
+++ b/src/systems/camera.js
@@ -100,11 +100,15 @@ module.exports.System = registerSystem('camera', {
     cameraEls = sceneEl.querySelectorAll('[camera]');
     for (i = 0; i < cameraEls.length; i++) {
       cameraEl = cameraEls[i];
-      if (newCameraEl === cameraEl) { continue; }
+      if (isMixin(cameraEl) || newCameraEl === cameraEl) { continue; }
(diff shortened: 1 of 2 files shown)


src/systems/camera.js
```diff
@@ -100,11 +100,15 @@ module.exports.System = registerSystem('camera', {
     cameraEls = sceneEl.querySelectorAll('[camera]');
     for (i = 0; i < cameraEls.length; i++) {
       cameraEl = cameraEls[i];
-      if (newCameraEl === cameraEl) { continue; }
+      if (isMixin(cameraEl) || newCameraEl === cameraEl) { continue; }
```
ngokevin: You can do `!cameraEl.isEntity`

delapuente: Will it return `true` for primitives also?

ngokevin: Yeah.

PR #3199, opened 2017-10-27
Fix IIFE typo in Best Practices code sample

brianpeiris:
**Description:**
The IIFE in the example code is supposed to be the outer function, not the inner function.
**Changes proposed:**
- Fixed the typo


dmarcos:
Thank you

machenmusik:
Aren't we finding ourselves replacing IIFE and closure variables due to issues like #3200?

ngokevin:
It's only an issue when it's used at the wrong time.

PR #3203, opened 2017-10-28
Change date format to ISO8601

### LINKED ISSUE #3202: Documentation - date version ambiguity
**Description:**

The version date is ambiguous. 

For example, on https://github.com/aframevr/aframe/blob/43a113600b47031ff60b63c9007e08cd7e8a597f/src/index.js#L86

Is that 5th October, or 10th May?

May I suggest that `getBuildTimestamp ()` in [`updateVersionLog.js`](https://github.com/aframevr/aframe/blob/master/scripts/updateVersionLog.js#L19) be changed to use ISO 8601 format.

YYYY-MM-DD is the international standard for representing dates - it's unambiguous and is well understood by machines and people.

I'll submit a small PR.  Please let me know if it is inappropriate or if it breaks anything.

edent:
Less ambiguous date format.

**Description:**
Fixes #3202 

**Changes proposed:**
- Use ISO8601 Format
- Dates will be YYYY-MM-DD

### DIFF
--- a/scripts/updateVersionLog.js
+++ b/scripts/updateVersionLog.js
@@ -16,17 +16,9 @@ gitRev.short(hash => {
 });
 
 /**
- * @returns {string} of format `1-1-2017`.
+ * @returns {string} of format `2017-12-25`.
  */
 function getBuildTimestamp () {
-  function pad2 (value) {
-    return ('0' + value).slice(-2);
-  }
-  const date = new Date();
-  const timestamp = [
-    pad2(date.getUTCDate()),
-    pad2(date.getUTCMonth() + 1),
-    date.getUTCFullYear()
-  ];
-  return timestamp.join('-');
+  const timestamp = new Date().toISOString().slice(0, 10);
+  return timestamp;
 }


dmarcos:
I would keep the current format so the log message remains consistent unless there's clear value on changing the date format.

edent:
As I mentioned in #3202, the current format is ambiguous - both for humans and machines.

I'm European, and often find it confusing working out whether 05-10-2017 is in October (as we would write it) or May as Americans would write it.

Similarly, computers parsing log messages will find it easier to ingest the standard ISO8601 format - which is used all over the web.

dmarcos:
Thanks. I'm also European and I have never had trouble with the current format. It might not be the case for the general population. I have no data on either direction just personal evidence. Just curious, Is not the problem the same for humans if you see the date and don't know whether it is ISO8601? `05-10-2017`  and  `2017-05-10` are equally ambiguous to me if you do not know about data formats and most humans don't. Putting the year after or before the `month-day` does not change much in practice. In computer world I personally always assume that month goes first. 

The log is also a message for humans (developers more precisely) and not for machines to parse. In this case keeping consistency might be more valuable that potential benefit of the change.

edent:
ISO8601 goes largest element to smallest element. Year-Month-Day Hour Minutes Seconds.

I would expect most JS programmers reading the logs to be familiar with it.

* It's the [HTML standard way of representing time](https://www.w3.org/TR/html51/infrastructure.html#dates-and-times)
* [Schema.org](https://schema.org/Date) uses it
* [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) is functionally identical and commonly used.
* As noted in the PR, JavaScript uses it as [`toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)
* [Keep A Changelog](http://keepachangelog.com/en/1.0.0/) also recommend that format.


If you want reduce confusion, you could use [RFC 822](http://www.rssboard.org/rss-profile#data-types-datetime) which looks like `Mon, 15 Oct 2007 14:10:00 GMT` - but that will be confusing to people who don't speak English as the month names will be different.

As I said, I found the logs confusing. I think this simple change would improve things.

machenmusik:
So I do think there is confusion between mm-dd-yyyy and dd-mm-yyyy, the most common occurrence for me (as I am not European) being travel documentation... but when year comes first it is understood to be yyyy-mm-dd since no one appears to use yyyy-dd-mm (thanks ISO8601 and friends)

> In computer world I personally always assume that month goes first.

If the vast majority of programmers agree with the personal assumption, then agreed not much impetus to change; however if there is significant confusion, this seems like a simple enough change to make

dmarcos:
Thanks!

edent:
Thank you 😀

Issue #916, opened 2016-01-31
Multiple a-scenes with a page


zitterbewegung:
Hi, I would like to have multiple a-scenes within a page so that I can have multiple 3d models mixed in with other html.

RangerMauve:
What about embedding iframes for the time being?

cvan:
Fullscreening (entering VR) with iframes work with Firefox but not with Chrome

brianblakely:
This is probably the reason we'll need to abandon A-Frame for our project. It stinks, but we can't use the iframe solution.

cvan:
> This is probably the reason we'll need to abandon A-Frame for our project. It stinks, but we can't use the iframe solution.

Actually, this is supported now. Sorry, this issue didn't get closed with a mention of the proper pull request that enabled this to work.

@dmarcos worked on this for [A-Frame v0.3.0](https://aframe.io/blog/aframe-v0.3.0/). [Here's an example of a mock Wikipedia entry](https://aframe.io/aframe/examples/showcase/wikipedia/), demonstrating the ability to embed one scene (but could be multiple) in an existing 2D page alongside content.

brianblakely:
Maybe it is just a-videosphere, in that case?






On Tue, Sep 20, 2016 at 4:59 AM -0400, "chris van wiemeersch" <notifications@github.com> wrote:













This is probably the reason we'll need to abandon A-Frame for our project. It stinks, but we can't use the iframe solution.




Actually, this is supported now. Sorry, this issue didn't get closed with a mention of the proper pull request that enabled this to work.



@dmarcos worked on this for A-Frame v0.3.0. Here's an example of a mock Wikipedia entry, demonstrating the ability to embed one scene (but could be multiple) in an existing 2D page alongside content.

ngokevin:
@cvan multiple scenes on a page do not work yet. It is complex to do. And the issue is still

ngokevin:
I frames don't quite work entering fullscreen yet

cvan:
> I frames don't quite work entering fullscreen yet

even with `<iframe allowfullscreen>`? what's the issue? is there an issue on file in this repo?

ngokevin:
I'm actually not sure. I neglected the `allowfullscreen`. I remember it didn't work in CodePen. Need to check.

hawaiikaos:
[...]

PR #3206, opened 2017-10-31
fixed: components+systems would misbehave with multiple scenes

ProPuke:
**Description:**

Currently, one of the barriers to supporting multiple scenes (see #916) on the same page is that components and systems do not interact correctly when multiple scenes exist.

Each scene contains its own systems, but the last scene to initialise always overwrites the global `Component.__proto__.system` property of its components. This means that systems and components for earlier scenes no longer function correctly (for instance lights don't attach themselves to their correct lights system, meaning default lighting isn't disabled).

This fixes that by initialising components _after_ they have been attached to the scene, and setting their `Component.system` instance property to the correct scene system. Any attempts to update component properties before this happens are queued and then processed after attachment and initialisation; However DOM updates still occur as expected, to ensure DOM manipulation and debug serialisation still all work as expected.

**Changes proposed:**
- `Component.__proto__.system` is now longer set by systems upon system initialisation
- Component initialisation now occurs as a result of the `load()` callback and `updateComponents()` once the entity has been attached to a scene
- Component initialisation sets its local `system` by accessing its local `el.sceneEl.systems`. This occurs immediately within the `Component` base ctor, before local construction and `init` occurs.
- Attempts to update the properties of components before initialisation occurs succeed successfully and update the DOM where expected. The actual component update, however, is queued and only executed once `updateComponents()` has been called on the attached parent entity.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -36,6 +36,7 @@ var proto = Object.create(ANode.prototype, {
       this.components = {};
       // To avoid double initializations and infinite loops.
       this.initializingComponents = {};
+      this.deferredComponentUpdates = [];
       this.isEntity = true;
       this.isPlaying = false;
       this.object3D = new THREE.Group();
@@ -74,6 +75,8 @@ var proto = Object.create(ANode.prototype, {
       var self = this;  // Component.
 
       this.addToParent();
+
+      this.updateComponents();
 
       // Don't .load() scene on attachedCallback.
       if (this.isScene) { return; }
@@ -430,7 +433,7 @@ var proto = Object.create(ANode.prototype, {
         var i;
         var name;
 
-        if (!this.hasLoaded) { return; }
+        if (!this.hasLoaded || !this.parentEl) { return; }
 
         // Gather mixin-defined components.
         for (i = 0; i < this.mixinEls.length; i++) {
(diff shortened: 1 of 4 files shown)


ProPuke:
I also started clearing sceneEl on entity detach, but this was breaking unit tests, so I undid that.

This feels like a dependency on broken behaviour?

The docs simply say "An entity has a reference to its scene element.". Which would indicate to me it would be null if an entity was no longer attached a scene.

But if it's cleared on detach things break, so I let it linger to whatever it used to be :S

ProPuke:
Okay, there's also a few tests that intentionally try and access the raw component objects of entities before they're attached to a scene, or attach entities to a `<body>` instead of an `<a-scene>`. These fail :S

I'd be in favour of adding a caveat rule to the first (that the entity must be in a scene before the component objects can be accessed), in order to support multiple scenes, and changing the latter as it feels like incorrect/undefined behaviour.

dmarcos:
We recommend replacing the content of a single scene instead of having multiple in the same page if you want to swap content. Do you want to render multiple scenes simultaneously? It’s already problematic to get good performance on a single scene. These are also API breaking changes for components that access systems. In general tests should not be skipped to accomodate new functionality. Either the tests have to change if they don’t check the right things anymore or the PR has to be modified to not break pre existing functionality.

ngokevin:
While this fix may improve things for multiple scenes, I think we shouldn't go down that route for now as it will make things more complex. For support of multiple scenes, more things have to happen (e.g., three.js does not even support multiple scenes with WebGL renderer). See that #916 is closed.

PR #3207, opened 2017-10-31
added: depthTest and depthWrite properties to the Text component

### LINKED ISSUE #2914: request: Ability to control depth test/write with Text
Currently there are no `depthTest`/`depthWrite` properties for `Text`, and `Material`s do not work with `Text`.

Often you do not want text intersecting with geometry. Disabling `depthTest` for hud content and floating in-world labels would be useful.

ProPuke:
**Description:**

This closes #2914. There is currently no way to control depth testing or depth writing with the Text component. This adds `depthTest` and `depthWrite` properties to the Text component, as per Materials.

Ideally it would be nice in the future if the Material component could be used with objects using custom shaders, such as Text. But in the (hopeful) meanwhile this adds `depthTest` and `depthWrite` properties to `Text` so you can also control that stuff, should you need to (because sometimes you want 3d positional text that's guaranteed not to clip geometry)

**Changes proposed:**
- `depthTest` and `depthWrite` properties are added to the Text component. These are forwarded to the custom material in `updateBaseMaterial()` (like `side`)

### DIFF
--- a/src/components/text.js
+++ b/src/components/text.js
@@ -49,6 +49,8 @@ module.exports.Component = registerComponent('text', {
   schema: {
     align: {type: 'string', default: 'left', oneOf: ['left', 'right', 'center']},
     alphaTest: {default: 0.5},
+    depthTest: {default: true},
+    depthWrite: {default: true},
     // `anchor` defaults to center to match geometries.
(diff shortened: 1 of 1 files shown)


jbolten:
Doesn't seem to work any more.
Is there any reason why this commit was removed on 2017-12-13?
Is there another way to use depthTest with text?

PR #3210, opened 2017-10-31
make the example 360 video page redirect to example Glitch

machenmusik:
Per discussion on #2830, this PR replaces the previous example of 360 video in the repo to redirect to an example Glitch.  The example takes advantage of recently-merged #3176 to show how a video can be specified as HLS and MP4.

### DIFF
--- a/examples/boilerplate/360-video/index.html
+++ b/examples/boilerplate/360-video/index.html
@@ -1,20 +1,9 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <title>360 Video</title>
-    <meta name="description" content="360 Video — A-Frame">
-    <script src="../../../dist/aframe-master.js"></script>
+    <meta http-equiv="refresh" content="0; https://aframe-360-video-native-hls-example.glitch.me">
   </head>
   <body>
-    <a-scene>
-      <a-assets>
-        <video id="video" autoplay loop crossorigin="anonymous">
-          <source src="https://ucarecdn.com/fadab25d-0b3a-45f7-8ef5-85318e92a261/"></source>
-        </video>
-      </a-assets>
-
-      <a-videosphere src="#video" rotation="0 180 0"></a-videosphere>
-    </a-scene>
+    <a href="https://aframe-360-video-native-hls-example.glitch.me">Redirecting... click to go directly</a>
   </body>
 </html>
--- a/examples/index.html
+++ b/examples/index.html
@@ -173,7 +173,7 @@
     <ul class="links">
       <li><a href="boilerplate/hello-world/">Hello World</a></li>
       <li><a href="boilerplate/panorama/">360&deg; Image</a></li>
-      <li><a href="boilerplate/360-video/">360&deg; Video</a></li>
+      <li><a href="https://aframe-360-video-native-hls-example.glitch.me/">360&deg; Video</a></li>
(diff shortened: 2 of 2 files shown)


ngokevin:
Thanks, I'd focus on placing this within the documentation (FAQ, material page, videosphere page, future video guide) and move away from expecting the `examples/` directory to be the official place. Perhaps an HTML comment in the existing videosphere page, but we keep the `examples/` directory mainly for testing core features.

https://github.com/aframevr/aframe/blob/master/CONTRIBUTING.md#add-glitch-examples-to-documentation

machenmusik:
The primary usage I saw was the sidebar link for 360 video, and rather than changing all of those links to point at Glitch directly (which is not versioned, so potential issues if we change things later), I changed the target to be a stub.  That also allows previous links to the destination from external sites or personal bookmarks to be redirected to the new Glitch.

Is there a list of all the other places to change?  Happy to change those to direct Glitch links where it makes sense

machenmusik:
As an example of the kind of concerns I have:

- The docs/primitive/videosphere.md page links to the edit page for a Glitch example.  It is unclear to me whether this example link should be changed or not.

- The Glitch example, when you go and look at it, is referencing an older version of A-Frame (0.6.1).  If there were breaking changes between 0.6.1 and now, the same Glitch would be stuck failing for one or the other, as the Glitch itself is not versioned.

- The Glitch example, when you go and look at it, has been edited (probably unintentionally) so that it is rotated 80 degrees instead of 180 degrees.  We need a way to make the Glitch viewable, but require remix to edit.

ngokevin:
The site sidebar link can be changed to your Glitch

machenmusik:
ok - anywhere else?

ngokevin:
There is some confusion here. What we should do is:

Add the example to the **documentation** as noted here: https://github.com/aframevr/aframe/blob/master/CONTRIBUTING.md#add-glitch-examples-to-documentation

The examples don't need to be touched as we don't want to use them for developer reference as much as documentation and Glitches. I was talking about the A-Frame homepage sidebar, we can update the link to the source code.

machenmusik:
ok, so yes there is some confusion.

isn't the A-Frame homepage sidebar this?  (when you click it, currently goes to the boilerplate 360-video example not its source; this changes to go to glitch not its source, but could easily go to its source if that is what you wanted)
https://github.com/aframevr/aframe/pull/3210/commits/7309eca64660e6c7e494afdf7c031fd4e7b23b1a

and you'd simply like to revert this, leaving the previous example unchanged but no longer linked?
https://github.com/aframevr/aframe/pull/3210/commits/aeb4f8166176f37e58eb2d52ae77af87be46f583

machenmusik:
I did specifically ask about the existing documentation but no answer, only about site sidebar:
```
The docs/primitive/videosphere.md page links to the edit page for a Glitch example. It is unclear to me whether this example link should be changed or not.
```

ngokevin:
Okay to change the link in `a-videosphere` and https://github.com/aframevr/aframe/commit/7309eca64660e6c7e494afdf7c031fd4e7b23b1a is not the homepage. If we can keep threads concise, it'd be easier to not miss things.

machenmusik:
got it, thanks!

The branch for this PR now has just the docs example link change.  Did you want to reopen this one, or have me do another PR?

The homepage link is in a different repo I presume, will find it and address there.

PR #3212, opened 2017-11-01
Prevent component leakage in updateComponents

### LINKED ISSUE #3200: entity.updateComponents leaks components into other entities
**Description:** `entity.updateComponents` can end up being called with a dirty [`componentsToUpdate` closure object](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L425) that contains component keys from another entity, causing those components to be erroneously added to the current entity.

One such scenario is a component using `setAttribute` to set the mixins of a child entity from within its `update` handler will end up also adding itself to the child entity because it is still listed in the shared closure object as a component needing update on its own entity. If said component also creates that child entity on initialization, then you have the recursion problem in the reprex below. 

- A-Frame Version: 0.7.1 & master
- Platform / Device: Windows Chrome & Firefox
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/glorious-gargoyle?path=index.html:21:0 (there should only be 1 sphere)

I think this could be fixed similarly to #3170 , by making `componentsToUpdate` an instance object.

wmurphyrd:
**Description:** Prevent an edge case wherein an entity could erroneously add its components to its child entities. Fix #3200 

**Changes proposed:**
- Move `componentsToUpdate` object from IIFE closure to instance variable

The diff is messy due to a change in indentation level for `updateComponents`. The actual changes to the method were 1) remove the IIFE and 2) declare a local componentsToUpdate variable that points to the instance variable

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -36,6 +36,7 @@ var proto = Object.create(ANode.prototype, {
       this.components = {};
       // To avoid double initializations and infinite loops.
       this.initializingComponents = {};
+      this.componentsToUpdate = {};
       this.isEntity = true;
       this.isPlaying = false;
       this.object3D = new THREE.Group();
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

PR #3213, opened 2017-11-01
Use component event handlers insted of bound a-scene methods

dmarcos:
It is more hygienic to use dedicated event handlers than directly wrapped methods of `a-scene`. This became a problem with some experiments that were attempting at wrapping `a-scene` methods.

### DIFF
--- a/src/components/scene/vr-mode-ui.js
+++ b/src/components/scene/vr-mode-ui.js
@@ -24,15 +24,14 @@ module.exports.Component = registerComponent('vr-mode-ui', {
 
     if (utils.getUrlParameter('ui') === 'false') { return; }
 
-    this.enterVR = bind(sceneEl.enterVR, sceneEl);
-    this.exitVR = bind(sceneEl.exitVR, sceneEl);
     this.insideLoader = false;
     this.enterVREl = null;
     this.orientationModalEl = null;
+    this.bindMethods();
 
     // Hide/show VR UI when entering/exiting VR mode.
-    sceneEl.addEventListener('enter-vr', bind(this.updateEnterVRInterface, this));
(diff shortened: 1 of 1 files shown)


src/components/scene/vr-mode-ui.js
```diff
@@ -55,10 +61,10 @@ module.exports.Component = registerComponent('vr-mode-ui', {
     if (this.enterVREl || this.orientationModalEl) { return; }
 
     // Add UI if enabled and not already present.
-    this.enterVREl = createEnterVRButton(this.enterVR);
+    this.enterVREl = createEnterVRButton(this.onEnterVRButtonClick.bind(this));
```
ngokevin: I'd bind in `init`

src/components/scene/vr-mode-ui.js
```diff
@@ -124,16 +130,18 @@ function createEnterVRButton (enterVRHandler) {
...
+    onClick();
   });
   return wrapper;
 }
 
 /**
- * Create a modal that tells mobile users to orient the phone to landscape.
- * Add a close button that if clicked, exits VR and closes the modal.
+ * Creates a modal dialog to request the user to switch to landscape orientation.
+ * *
```
ngokevin: double star

src/components/scene/vr-mode-ui.js
```diff
@@ -46,6 +44,14 @@ module.exports.Component = registerComponent('vr-mode-ui', {
                             bind(this.toggleOrientationModalIfNeeded, this));
   },
 
+  onModalClick: function () {
```
ngokevin: docstrings (`Exit VR when modal clicked.` - `Enter VR when VR button clicked.`)

PR #3214, opened 2017-11-02
allow disable shadow map auto update, fix shadow system init

ngokevin:
**Description:**

**Changes proposed:**
- Shadow system setup was never getting called.
- Allow disable auto update shadow map to allow one-time shadow baking and on-demand updates.

### DIFF
--- a/docs/components/shadow.md
+++ b/docs/components/shadow.md
@@ -24,7 +24,6 @@ scene but not receive shadows itself.
 ```
 
 [light]: ./light.md#configuring-shadows
-
 **IMPORTANT:** Adding the `shadow` component alone is not enough to display
 shadows in your scene. We must have at least one light with `castShadow:
 true` enabled.  Additionally, the light's shadow camera (used for depth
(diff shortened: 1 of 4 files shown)


docs/components/shadow.md
```diff
@@ -37,3 +36,16 @@ component for more information.
...
+
+## Scene Properties
+
+The shadow system exposes scene-level properties for configuring the renderer
+for shadows. These are set on `<a-scene>` (e.g., `<a-scene shadow="autoUpdate:
+  false">`).
+
+| Property           | Description                                                                                                                                                                                           | Default Value                       |
+|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| autoUpdate         | Whether to dynamically update the shadow map every frame. Disable and manually update by setting `renderer.shadowMap.needsUpdate = true` for best performance. Calculating shadow maps are expensive. | true                                |
```
donmccurdy: nit: s/are expensive/is expensive

dmarcos:
Tests failing

ngokevin:
Updated.

`needs test fixed` label? GitHub shows red X everywhere already.

dmarcos:
It helps me visually and sometimes you get an `X` on the pull request list when in reality is not broken tests but some sort of code coverage compliance thing.

ngokevin:
OK I don't like more process, but it's fine. FYI those code coverage compliance things were from when people accidentally included `.only` in their tests (46c7cfcd12f6ef4bf3d30f6c744035ffd173510f). I've written a hook to prevent that.

PR #3215, opened 2017-11-02
cursor only re-check for intersections on cursor down for touch

ngokevin:
**Description:**

@wmurphyrd 

**Changes proposed:**
- Seems like that new re-cast is not needed for mouse


wmurphyrd:
Probably true. I thought about this, but I landed on leaving the check in for mouse use for the edge case in which a really fast user moves the mouse over the target and clicks within the 100ms default raycast refresh period. No strong opinion either way though.

ngokevin:
Alright makes sense. Just gonna push whitespace changes then

wmurphyrd:
Feel free to request whitespace and style changes from me on my PRs before merging

PR #3218, opened 2017-11-03
Better hls example

machenmusik:
Per discussion on #3210.

### DIFF
--- a/docs/primitives/a-videosphere.md
+++ b/docs/primitives/a-videosphere.md
@@ -6,7 +6,7 @@ parent_section: primitives
 source_code: src/extras/primitives/primitives/a-videosphere.js
 examples:
   - title: Adding a Videosphere
-    src: https://glitch.com/edit/#!/aframe-adding-a-videosphere?path=index.html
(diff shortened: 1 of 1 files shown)


ngokevin:
Sweet! Found a few issues just now:

- Clicking the window causes video to pause (in case of mouse drag).
- First time loading, I got a DOM timing issue that `play()` call was interrupted by a `pause()` call.
- Sometimes WebGL errors when the video not yet loaded and gets passed to WebGL and we have a texture with no image.

machenmusik:
re: clicking - I can remove the toggling and just make click trigger play, to fix that.
re: first time DOM timing issue and WebGL errors - can you say more about your configuration / browser at the time?  always, or intermittent?  main page, or detail?

machenmusik:
OK I think those issues are addressed now
- click to play only, no toggling to avoid pause on mouse drag, etc.
- for main page, which uses MP4 and may not be streaming, don't shorten the a-assets timeout

https://github.com/chenzlabs/aframe-360-video-example

PR #3219, opened 2017-11-03
don't use getOrCreate mesh for geometry/material

ngokevin:
**Description:**

@donmccurdy raised this in an issue before. I just ran into problems with it.

The `getOrCreateObject3D` makes listening on the `object3dset` event unreliable, since the `getOrCreateObject3D` will emit the `object3dset` event while the object has an **empty mesh**, before the geometry/material is properly set.

After this, I think we can deprecate `getOrCreateObject3D`.

**Changes proposed:**
- Only set the object3d *after* the material/geometry are set.
- Have material not set the Mesh, it depends on other components to provide mesh (e.g, geometry/model). Waits for object3dset event.

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -152,11 +152,24 @@ module.exports.Component = registerComponent('material', {
    * @returns {object} Material.
    */
   setMaterial: function (material) {
-    var mesh = this.el.getOrCreateObject3D('mesh', THREE.Mesh);
+    var el = this.el;
+    var mesh;
     var system = this.system;
+
     if (this.material) { disposeMaterial(this.material, system); }
-    this.material = mesh.material = material;
+
+    this.material = material;
     system.registerMaterial(material);
+
+    // Set on mesh. If mesh does not exist, create it.
+    mesh = el.getObject3D('mesh');
+    if (mesh) {
+      mesh.material = this.material;
+    } else {
+      mesh = new THREE.Mesh();
+      mesh.material = this.material;
+      el.setObject3D('mesh', mesh);
+    }
   }
 });
 
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -161,14 +161,16 @@ module.exports.Component = registerComponent('material', {
     this.material = material;
     system.registerMaterial(material);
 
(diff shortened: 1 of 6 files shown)


donmccurdy:
+1 for deprecating `getOrCreateObject3D`, but I think this implementation has the same effect. Is it necessary to create a stub `THREE.Mesh` to hold the material?  [`obj-model` already gets the material from `el.components.material.material`](https://github.com/aframevr/aframe/blob/v0.7.0/src/components/obj-model.js#L58-L67), so I don't see what the stub mesh accomplishes.

ngokevin:
So material should wait for object3dset, right?

ngokevin:
Updated for that.

donmccurdy:
I'm worried about #3220 and would prefer that we check for `event.target === this.el` instead, but otherwise this looks good.

src/components/material.js
```diff
@@ -152,11 +152,26 @@ module.exports.Component = registerComponent('material', {
...
+    // Set on mesh. If mesh does not exist, wait for it.
+    mesh = el.getObject3D('mesh');
+    if (mesh) {
+      mesh.material = material;
+    } else {
+      el.addEventListener('object3dset', function waitForMesh (evt) {
+        if (evt.detail.type !== 'mesh') { return; }
+        el.getObject3D('mesh').material = material;
+        el.removeEventListener('object3dset', waitForMesh);
+      });
```
donmccurdy: According to http://caniuse.com/#feat=once-event-listener, support for the `addEventListener(type, fn, {once: true})` option is good enough to start using.  MDN claims Edge doesn't have it yet, so we should check on that first.

ngokevin:
Reverted the bubble change. Added the target check. Fixed tests.

src/components/material.js
ngokevin: Too bad it wouldn't work since we have to do the type and target checks.

donmccurdy: Ah, yeah. 🙁

donmccurdy:
Still something up with the tests that isn't obvious to me, otherwise this looks good. Thanks!

PR #3220, opened 2017-11-03
don't bubble object3dset/remove events

ngokevin:
**Description:**

I think there's potential for surprise if we're listening to an object3dset event, and accidentally catch an object3dset event from a child. One would have to know to check the event target, which is unlikely to remember.

**Changes proposed:**
- Don't bubble object3dset/remove events.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -170,7 +170,7 @@ var proto = Object.create(ANode.prototype, {
       // Add.
       this.object3D.add(obj);
       this.object3DMap[type] = obj;
-      this.emit('object3dset', {object: obj, type: type});
+      this.emit('object3dset', {object: obj, type: type}, false);
     }
   },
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
I'm not sure about this change. There are side effects:

1. No reliable way to listen for a new mesh entering the scene. Can try `model-loaded`, but you'll miss primitive geometries. Can try `child-attached`, but you'll get noise from entities that haven't finished loading, or that aren't intended to contain meshes.
2. The new `raycaster` refresh logic relies on this event reaching the scene, and will break with this change.

The risk of catching `object3dset` from children is there too, of course, but at least it's a risk that we share with most DOM APIs. Better the devil you know, etc.

ngokevin:
That's true. I think we can revert then, especially given the raycaster refresh logic. Too bad the tests missed that, good catch.

PR #3221, opened 2017-11-03
emit event for geometry set

ngokevin:
**Description:**

Got a use case when I need to know when a new geometry is set so I can do stuff like reapply UVs or vertex colors.

**Changes proposed:**
- `geometryset` event, does not bubble, no event detail needed.

### DIFF
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -36,6 +36,8 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+
+    this.el.emit('geometryset', null, false);
   },
 
   /**


src/components/geometry.js
```diff
@@ -36,6 +36,8 @@ module.exports.Component = registerComponent('geometry', {
 
     // Create new geometry.
     this.geometry = mesh.geometry = system.getOrCreateGeometry(data);
+
+    this.el.emit('geometryset', null, false);
```
donmccurdy: Per comment on #3219, I think this should work a bit differently — the `material` component should cache its `THREE.Material` instance if no mesh is available, without creating a mesh, and wait for the `object3dset` event to fire. The geometry components should always check `getObject3D('mesh')` first, and call `setObject3D('mesh', mesh)` if nothing exists yet.

ngokevin:
Actually, I can just use `componentchanged`.

PR #3222, opened 2017-11-04
[deprecate] Entity.getOrCreateObject3D

ngokevin:
**Description:**

Move away from this as it creates empty stub meshes which are confusing and can mess things up. Geometry and material have been changed to not use these already ,which were the only cases in the core.

Use `setObject3D()` and `object3dset` event instead.

**Changes proposed:**
- deprecate getOrCreateObject3D


ngokevin:
The failing test is related to code coverage, but I removed the test. We can ignore that.

PR #3224, opened 2017-11-07
avoid constantly cloning and copying quaternion in link component

vincentfretin:
**Description:**

avoid constantly cloning and copying quaternion when we are close to portal (<= 20)

**Changes proposed:**
- avoid quaternion.clone() on each tick by reusing a quaternionClone variable and doing instead a copy.
- If I well understood the code, object3D.quaternion is modified only by object3D.lookAt, so to avoid doing quaternionClone.copy and object3D.quaternion.copy on each tick when we are close to portal, simply do quaternionClone.copy before the object3D.lookAt

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -25,6 +25,7 @@ module.exports.Component = registerComponent('link', {
   init: function () {
     this.navigate = this.navigate.bind(this);
     this.previousQuaternion = undefined;
+    this.quaternionClone = new THREE.Quaternion();
     // Store hidden elements during peek mode so we can show them again later.
     this.hiddenEls = [];
     this.initVisualAspect();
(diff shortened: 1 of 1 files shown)


src/components/link.js
```diff
@@ -171,6 +171,7 @@ module.exports.Component = registerComponent('link', {
     var cameraWorldPosition = new THREE.Vector3();
     var elWorldPosition = new THREE.Vector3();
     var quaternion = new THREE.Quaternion();
+    var quaternionClone = new THREE.Quaternion();
```
ngokevin: This should be on the instance of the component, else all links will share this. It seems `this.previousQuaternion` is meant to be stateful.

dmarcos: Yes, the variable is stateful.

vincentfretin:
I moved quaternionClone to init and rebased from master. Is this ok now?

PR #3227, opened 2017-11-09
Changed GearVR device to not be mobile

Marchhill:
**Description:**
As mentioned [here](https://github.com/aframevr/aframe/issues/1346#issuecomment-340851877) the Samsung Internet VR Browser does not correctly display aframe, instead showing the game through two eyeholes. This pull request was intended to fix that but instead it simply shows the aframe game on a screen (like a watching it on a cinema screen). So this is an improvement but not a fix.

**Changes proposed:**
- Make _isMobile false if on gearvr.
- See [this](https://github.com/aframevr/aframe/issues/1346#issuecomment-340861743) comment made by @dbradleyfl

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -45,6 +45,9 @@ var isMobile = (function () {
     if (isIOS() || isTablet() || isR7()) {
       _isMobile = true;
     }
+    if (isGearVR()) {
+      _isMobile = false;
+    }
   })(window.navigator.userAgent || window.navigator.vendor || window.opera);
 
   return function () { return _isMobile; };


vincentfretin:
I don't understand what is the issue here.
I'm using GearVR with Oculus browser and Samsung Internet VR with no issue with the S8 phone.
Unless you have this issue https://twitter.com/fernandojsg/status/924985106372100096 ?
@fernandojsg was this with a S6 phone and Samsung Internet? Do you have still this issue?
@Marchhill what phone do you use?

fernandojsg:
@vincentfretin yes it was with my S6 + samsung internet. I don't have my device here but I'll let you know once I'll test it again.

Marchhill:
@vincentfretin sorry forgot to mention I am using S7

Marchhill:
Here is a comparison of before and after the pr:
before
after

dmarcos:
Is non A-Frame webvr content working for you? Does the Oculus Browser work? I believe there was a recent regression on the Internet Browser these days

mkungla:
This PR is not a fix. A-Frame and other VR Content works fine with Oculus Browser.
Issue described here is latest problem with Samsung Internet it self that no VR content works correctly with that so therefore nothing to do with A-Frame.

mkungla:
GearVR users have to make sure that their Oculus Software is up to date and use shipped Oculus Browser instead of Samsung Internet.

machenmusik:
OK then why don't we suggest that an issue be filed against https://github.com/SamsungInternet/support
similar to 
https://github.com/SamsungInternet/support/issues/22
and close this one here?

dmarcos:
There’s an open issue that it’s probably related:

https://github.com/SamsungInternet/support/issues/19

@marchill Thanks for investigating this. Can you check if the issue above is the root cause?

Marchhill:
@dmarcos aha yes that is it

PR #3229, opened 2017-11-11
allow radians for rotation, make rotation multi-prop

ngokevin:
**Description:**

Most 3D math stuff is done in radians. Allowing radians allows applications that use radians to not have to convert to degrees only for A-Frame to convert back to radians. Makes it easy to do things like `.setAttribute('rotation', object3D.rotation)`.

Made rotation multi-prop to support `unit` property. Will later also move scale and position to be multi-prop to allow easier updating of single axis of these components.

**Changes proposed:**
- Radians prop type.
- Multi-prop rotation component.

### DIFF
--- a/src/components/rotation.js
+++ b/src/components/rotation.js
@@ -1,16 +1,39 @@
 var degToRad = require('../lib/three').Math.degToRad;
 var registerComponent = require('../core/component').registerComponent;
+var utils = require('../utils/');
+
+var defaultRotation = {x: 0, y: 0, z: 0};
+
+var UNIT_DEGREES = 'degrees';
+var UNIT_RADIANS = 'radians';
 
 module.exports.Component = registerComponent('rotation', {
-  schema: {type: 'vec3'},
+  schema: {
+    unit: {default: UNIT_DEGREES, oneOf: [UNIT_DEGREES, UNIT_RADIANS]},
+    x: {type: 'number'},
+    y: {type: 'number'},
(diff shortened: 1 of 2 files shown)


dmarcos:
This could make components incompatible with one another if they use different radians. Unit conversation is one common source of catastrophic errors and headaches. See Mars Climate Orbiter. I’m open to change to radians but I would not support both. It is opening a can of worms

dmarcos:
As an alternative I propose using `el.getObject3D()` so the Object3D can be manipulated directly and the info gets synced backed up to the rotation, scale and position components of the entity. Object3D comes with callbacks already when Euler and Quaternion change. For scale, position and rotation I would wrap the method set of the Object3Ds of each entity so we can get notified.

ngokevin:
Simple wrapping won't work well for vector3s. .set is not the only method used to manipulate a vector, there's all the other vector methods as well as simply doing direct set like position.x = 10

ngokevin:
Maybe if we link the data object directly to the vector

dmarcos:
You can wrap the `updateMatrix` method of the Object3D to keep position, rotation, scale data in sync

ngokevin:
The matrix won't be updated immediately though

ngokevin:
I'd wrap getAttribute to return data directly from the Object3D

ngokevin:
Will work on this later and as a fuller solution with three.js -> A-Frame syncing

Issue #3230, opened 2017-11-13
Only half of the screen is present after changing mobile's orientation


Lefortov:
The bug is present on iOS using Chrome browser, on other browsers it works well. When you rotate your phone so it's changing orientation the canvas is present only on the half of the screen, other part of the screen is empty.

- A-Frame Version: 0.7.1
- Platform / Device: IOS, Chrome browser

jaydansand:
This also happens on iOS Safari (iOS 9, 10, and 11) when launched as a full-window app from a home screen bookmark.

Here was my half-baked patch to fix the home screen issue; probably a similar approach is necessary for Chrome. At least for the Safari issue, the problem was at the point of the `orientationchange` event, `window.innerHeight` and `window.innerWidth` (as returned by `getCanvasSize()` when the view is *not* embedded) were reporting incorrect (too small) values, but after the transition, they were correct - so I added a 100ms `setTimeout` workaround.

This is based on 0.6.1, but I could roll a pull request if desired against the current version if desired:
```
--- aframe-master.js
+++ aframe-master.js
@@ -48678,13 +48678,23 @@
                 var canvas = this.canvas;
                 var embedded = this.getAttribute('embedded') && ! this.is('vr-mode');
                 var size;
+                var self = this;
                 if (! camera || ! canvas || this.is('vr-mode') && isMobile) {
                   return;
                 }
-                size = getCanvasSize(canvas, embedded);
-                camera.aspect = size.width / size.height;
-                camera.updateProjectionMatrix();
-                this.renderer.setSize(size.width, size.height);
+                var setSize = function() {
+                  size = getCanvasSize(canvas, embedded);
+                  console.log(size, canvas, embedded);
+                  camera.aspect = size.width / size.height;
+                  camera.updateProjectionMatrix();
+                  self.renderer.setSize(size.width, size.height);
+                };
+                if (self.isIOS && navigator.standalone) {
+                  setTimeout(setSize, 100);
+                }
+                else {
+                  setSize();
+                }
               }, writable: window.debug
             },
             setupRenderer: {

```

jaydansand:
Yep, same issue, same fix. I've updated the patch and filed a pull request (#3232).

PR #3232, opened 2017-11-13
Fix erroneous canvas size on iOS devices

jaydansand:
This PR is related to issue #3230.

**Description:**
In iOS, orientation changes firing `AScene.resize` may erroneously set the canvas size, reducing the visible area by half:
portrait
landscape

This occurs only in non-embedded scenes because the values in `window.innerHeight` and `window.innerWidth` (used by `getCanvasSize()`) are incorrect mid-transition in some browsers. Affected (so far): iOS Chrome, and Safari when the page is launched full-screen via a home screen bookmark.

**Changes proposed:**
- Modify `AScene.resize` to institute a 100ms delay before resizing via `getCanvasSize()` for non-embedded scenes on iOS devices.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -457,6 +457,7 @@ module.exports.AScene = registerElement('a-scene', {
         var embedded = this.getAttribute('embedded') && !this.is('vr-mode');
         var size;
         var isEffectPresenting = this.effect && this.effect.isPresenting;
+        var self = this;
         // Do not update renderer, if a camera or a canvas have not been injected.
         // In VR mode, VREffect handles canvas resize based on the dimensions returned by
         // the getEyeParameters function of the WebVR API. These dimensions are independent of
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -457,18 +457,32 @@ module.exports.AScene = registerElement('a-scene', {
...
+          // Notify renderer of size change.
+          self.renderer.setSize(size.width, size.height, false);
+        };
+        if (this.isIOS && !embedded) {
+          // On iOS devices (at least as of iOS 11.1.1), particularly in Chrome
+          // or home-screen-icon Safari, when the orientation is changed the
+          // canvas size returned by getCanvasSize() is incorrect during the
+          // transition. Delay resizing for 100ms to allow the transition to
+          // finish.
+          setTimeout(setSize, 100);
```
dmarcos: This seems brittle. Should we listen to [orientationchange](https://developer.mozilla.org/en-US/docs/Web/Events/orientationchange) to also trigger a resize cycle?

jaydansand: I agree `setTimeout` is a kludge and the value of 100ms was imprecise (it was experimentally-derived and I added some padding to be sure). Yes, that's brittle. But the alternative is perhaps more brittle:

The order of `resize` and `orientationchange` is [undefined and implementation-specific](https://stackoverflow.com/questions/14157942/orientationchange-event-fires-scroll-resize-event). Various Stack Overflow answers suggest using a `setTimeout()` and debounce function if you need to bind to both events, due to the lack of standardization of event order (or even because some events don't always fire when both are applicable).

Anecdotally, I just experimented with my web app, and I can confirm that `resize` fires *after* `orientationchange` (iPhone 6S, iOS 11) so that wouldn't solve the problem of needing to re-run `resize`.

dmarcos:
Thanks! Closed in favor of the simpler https://github.com/aframevr/aframe/pull/3306

PR #3269, opened 2017-12-06
add xOffset option to allow text padding

ngokevin:
**Description:**

xOffset to allow for padding relative to the geometry plane behind it without having to separate them into two entities.,

**Changes proposed:**
- text.xOffset


machenmusik:
Does the padding handle the various align settings e.g. if you right align, does the padding work correctly, or does it act as left padding and the text goes outside the plane?  If center align, does it pad both sides?

ngokevin:
Whatever you want, you can set negative or positive value. It's naive.

machenmusik:
Oh ok, so literally just offset, not actually padding.  Nice!

PR #3271, opened 2017-12-06
optimize aframe.utils.diff

ngokevin:
**Description:**

**Changes proposed:**
- Remove creation objects, arrays, functions, closures.
- Allow specify targetObject to be the return object versus creating an objevt.


ngokevin:
r?

PR #3274, opened 2017-12-07
[docs] update path of link in html-and-primitives

laurasimmons30:
**Description:**
- link to building a basic scene resulted in 404

**Changes proposed:**
-
-
-


dmarcos:
Thanks 🏆

PR #3280, opened 2017-12-12
Add spectator camera

dmarcos:
- [x] - Test on headset
- [x] - Add tests

### DIFF
--- a/docs/components/camera.md
+++ b/docs/components/camera.md
@@ -27,6 +27,8 @@ A camera situated at the average height of human eye level (1.6 meters).
 | far        | Camera frustum far clipping plane.                                                                                                                                                                                                                                                  | 10000         |
 | fov        | Field of view (in degrees).                                                                                                                                                                                                                                                         | 80            |
 | near       | Camera frustum near clipping plane.                                                                                                                                                                                                                                                 | 0.005         |
+| spectator   | Whether the camera is used to render a third-person view of the scene on the 2D display while in VR mode.
+                                                                                                                                                                                                                | false            |
 | zoom       | Zoom factor of the camera.                                                                                                                                                                                                                                                          | 1             |
 
 ## Default Camera
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -11,6 +11,7 @@ module.exports.Component = registerComponent('camera', {
     far: {default: 10000},
     fov: {default: 80, min: 0},
     near: {default: 0.005, min: 0},
+    spectator: {default: false},
     zoom: {default: 1, min: 0}
   },
 
@@ -31,10 +32,8 @@ module.exports.Component = registerComponent('camera', {
    * Update three.js camera.
    */
   update: function (oldData) {
-    var el = this.el;
     var data = this.data;
     var camera = this.camera;
-    var system = this.system;
 
     // Update properties.
     camera.aspect = data.aspect || (window.innerWidth / window.innerHeight);
@@ -44,8 +43,16 @@ module.exports.Component = registerComponent('camera', {
     camera.zoom = data.zoom;
     camera.updateProjectionMatrix();
 
+    this.updateActiveCamera(oldData);
+    this.updateSpectatorCamera(oldData);
+  },
+
+  updateActiveCamera: function (oldData) {
+    var data = this.data;
+    var el = this.el;
+    var system = this.system;
     // Active property did not change.
-    if (oldData && oldData.active === data.active) { return; }
+    if (oldData && oldData.active === data.active || data.spectator) { return; }
 
     // If `active` property changes, or first update, handle active camera with system.
     if (data.active && system.activeCameraEl !== el) {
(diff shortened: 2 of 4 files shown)


docs/components/camera.md
```diff
@@ -27,6 +27,7 @@ A camera situated at the average height of human eye level (1.6 meters).
...
 | fov        | Field of view (in degrees).                                                                                                                                                                                                                                                         | 80            |
 | near       | Camera frustum near clipping plane.                                                                                                                                                                                                                                                 | 0.005         |
+| spectator   | The camera is used to render the scene on a 2D display in VR mode.                                                                                                                                                                                                                | false            |
```
ngokevin: `Whether the camera is used to render a third-person view of the scene on the 2D display while in VR mode.`

src/components/scene/spectator.js
```diff
@@ -0,0 +1,40 @@
...
+  setupDefaultSpectatorCamera: function () {
+    var sceneEl = this.el;
+    var spectatorCameraEl = sceneEl.systems.camera.spectatorCameraEl;
+    if (spectatorCameraEl) { return; }
+    spectatorCameraEl = this.defaultSpectatorCameraEl = document.createElement('a-entity');
+    spectatorCameraEl.setAttribute(DEFAULT_SPECTATOR_CAMERA_ATTR, '');
+    spectatorCameraEl.setAttribute(constants.AFRAME_INJECTED, '');
+    spectatorCameraEl.setAttribute('camera', {spectator: true});
+    spectatorCameraEl.setAttribute('wasd-controls', '');
+    spectatorCameraEl.setAttribute('look-controls', {hmdEnabled: false});
```
ngokevin: i think an ideal default would be orbit controls versus the dreaded WASD + mousedrag

src/components/scene/spectator.js
```diff
@@ -0,0 +1,40 @@
...
+
+var bind = require('../../utils/').bind;
+var constants = require('../../constants/');
+var DEFAULT_SPECTATOR_CAMERA_ATTR = 'data-aframe-default-spectator-camera';
+
+/**
+ * spectator component.
+ * It injects a spectator camera if there's not one already defined.
+ */
+module.exports.Component = register('spectator', {
```
ngokevin: Is a wrapper component necessary? I'd rather define my own camera via HTML, providing like id/position/rotation/scale with possibility of custom controls (maybe defaulting to orbit).

src/components/scene/spectator.js
dmarcos: Orbit controls are more appropriate when you want to focus the attention on a specific subject in the scene. I don't have any major issues with WASD + mousedrag.

src/components/scene/spectator.js
dmarcos: You can provide your own camera too but it is nice to be able to do just:
`<a-scene spectator>`

ngokevin: I think it's too much cruft, 40 LOC / documentation page / need to understand difference between `<a-scene spectator>` vs `<a-entity camera="spectator: true">`. Spectator camera will often be a deliberate feature addition, and in most cases we'd want to tailor the camera to the app which an injected camera won't help

src/components/scene/spectator.js
ngokevin: If you develop using WASD + mousedrag for two years, it's easy to get used to it. The specific subject would be the play space.

ngokevin: Though wouldn't matter if we can remove the wrapper and leave it to the dev to specify controls. You can do `<a-spectator-camera>` to default the properties

dmarcos: It does not matter if we remove the default. However, when implementing the viewer of A-Painter orbit controls I realized that are not very useful unless the subject is exactly on the center of the orbit controls. More often than not you want to let the user explore freely and orbit controls feel then limited. WASD + mousedrag are much more appropriate as a general solution and it is something that anybody that has played a video game is familiar with.

ngokevin: We can default to wasd:drag (but use a primitive not a wrapper component), but in practice I'd use orbit.

Games don't use mouse drag though, they use pointer lock.

And we overestimate the number of people that have played a first person game with WASD. I've put aframe in front of people and they have trouble with controls.

And I think orbit controls are less limited. The head height is not fixed, you can go under or over and fly. You can use with one hand. You can have a view of the whole scene or focus on one area. We did spectator mode for multiuser a-painter with Hayden, it worked okay.

dmarcos:
@ngokevin ?

ngokevin:
Looks good, we could have `<a-spectator-camera>` too since it's just a few lines for primitives.

dmarcos:
I'm not a fan of `<a-spectator-camera>` because it conveys the idea that you can add multiple of them. You can but it gets messy to manage. I would prefer people adding just one and position it as needed. That's why I had added the spectator component to the scene to act more like a singleton. I want also to de emphasize primitives and not reinforce the concept.

ngokevin:
r+

ngokevin:
Nice!

PR #3290, opened 2017-12-13
Ignore package-lock.json for now

mkungla:
Some point it may be committed into source repository!
https://docs.npmjs.com/files/package-lock.json

**Description:**

**Changes proposed:**
-
-
-


brianpeiris:
Out of curiosity, what was the reason behind ignoring package-lock.json?

PR #3293, opened 2017-12-13
default the raycaster interval/throttle to 0

ngokevin:
**Description:**

I recommend raycaster should be checked every frame so that interactions are responsive. Otherwise they can feel delayed or laggy, impacting the perceived performance.

**Changes proposed:**
- Remove default raycaster 100ms throttle.


vincentfretin:
please update the default value in the documentation as well

PR #3303, opened 2017-12-16
End video tag missing

entrebot:
I'm a total rookie so I'm probably making a mistake but it seems this line may be missing a '</video>' tag?

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
🙏 Thanks

PR #3305, opened 2017-12-17
mixin composition + string split cache util (fixes #3270)

### LINKED ISSUE #3270: Mixin composition
**Description:**

It'd be nice to be able to compose mixins together. Then we'd be able to define clear classes made of mixins. Otherwise, each entity has to specify every single mixin. This would match better with ECS.

```html
<a-mixin id="text" text="font: myfont.fnt"></a-mixin>
<a-mixin id="blackText mixin="text" text="color: #000"></a-mixin>
<a-mixin id="bigText" text="width: 3"></a-mixin>
<a-mixin id="header" mixin="bigText blackText"></a-mixin>
```

Yielding:

```
<a-entity mixin="header">

versus

<a-entity mixin="text blackText bigText header">
```

It'd be easy to absorb the other mixin into the mixin cache. But there's maintainence cost in needing to perhaps wait for other mixins to attach and cascading/handling dynamically added/updated mixins (perhaps we can just not handle that). 

I guess alternatively, a third-party system could handle it:

```
<a-mixin-composer id="header" mixins="blackText bigText"></a-mixin>
```

ngokevin:
**Description:**

Allow for mixin composition to build "classes" out of multiple mixins.

```
<a-mixin id="font" text="font: myfont.json"></a-mixin>
<a-mixin id="red" text="color: red"></a-mixin>

<a-mixin id="redText" mixin="red font"></a-mixin>

<a-entity mixin="redText someMixin">
// This will turn into `<a-entity mixin="red font redFont someMixin">`

```

**Changes proposed:**
- Allow for mixin composition.
- String split util (used by mixin) which caches string split operations to save memory.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -600,8 +600,9 @@ var proto = Object.create(ANode.prototype, {
         return;
       }
       if (attr === 'mixin') {
+        // Ignore if `<a-node>` code is just updating computed mixin in the DOM.
+        if (newVal === this.computedMixinStr) { return; }
         this.mixinUpdate(newVal, oldVal);
-        return;
       }
     }
   },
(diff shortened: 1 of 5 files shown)


ngokevin:
https://github.com/aframevr/aframe/pull/3302 is relevant for making sure defined attributes don't get overridden by mixins.

dmarcos:
Can we write some tests for this?

ngokevin:
Yeah, will do

ngokevin:
Updated.

dmarcos:
Tests failing on Firefox

ngokevin:
Updated

PR #3306, opened 2017-12-17
Postpone the resize operation in iOS to ensure that the window size matches the viewport on orientation changes (fix #3282)

### LINKED ISSUE #3282: a-canvas not always fullscreen in iOS Chrome
**Description:**
Scene doesn't take up whole screen. Even if it does initially, it breaks after changing device's orientation. In Safari it works alright - i expierienced it only on Chrome iOS. Chrome version: 62.0.3202.70
25344290_1651791048206153_1585761493_o

- A-Frame Version: 0.5.0, i also expierenced it on my own demos with 0.7.0 version
- Platform / Device: iPad Air 2
- Reproducible Code Snippet or URL: https://aframe-gallery.glitch.me/

dmarcos:
It is a simpler version of https://github.com/aframevr/aframe/pull/3232 Once the bug is fixed in Webkit (https://bugs.webkit.org/show_bug.cgi?id=170595) we can just remove the `if` statement.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -104,7 +104,18 @@ module.exports.AScene = registerElement('a-scene', {
 
         resize = bind(this.resize, this);
         window.addEventListener('load', resize);
-        window.addEventListener('resize', resize);
+        window.addEventListener('resize', function () {
+          // Workaround for a Webkit bug (https://bugs.webkit.org/show_bug.cgi?id=170595)
+          // where the window does not contain the correct viewport size
+          // after an orientation change. The window size is correct if the operation
+          // is postponed a few milliseconds.
+          // self.resize can be called directly once the bug above is fixed.
+          if (this.isIOS) {
+            setTimeout(resize, 100);
+          } else {
+            resize();
+          }
+        });
         this.play();
 
         // Add to scene index.


src/core/scene/a-scene.js
```diff
@@ -450,10 +450,20 @@ module.exports.AScene = registerElement('a-scene', {
...
+        var embedded;
         var size;
         var vrDevice;
         var isVRPresenting;
+        // Workaround for a Webkit bug (https://bugs.webkit.org/show_bug.cgi?id=170595)
+        // where the window does not contain the correct viewport size
+        // after an orientation change. The window size is correct if the operation
+        // is postpoend a few miliseconds.
+        // The if satement below can be removed when the bug is fixed.
+        if (this.iOS) {
```
ngokevin: Wouldn't this be a never-ending loop?

dmarcos: you looked to an outdated push. Now it is ready for review 😄

src/core/scene/a-scene.js
```diff
@@ -450,10 +450,23 @@ module.exports.AScene = registerElement('a-scene', {
...
+        var embedded;
         var size;
         var vrDevice;
         var isVRPresenting;
+        // Workaround for a Webkit bug (https://bugs.webkit.org/show_bug.cgi?id=170595)
+        // where the window does not contain the correct viewport size
+        // after an orientation change. The window size is correct if the operation
+        // is postponed a few miliseconds.
+        // The if satement below can be removed when the bug is fixed.
+        if (this.iOS && !this.delayedResize) {
```
ngokevin: Wouldn't it be simpler to go to the first `this.resize()` call and do the timeout check there versus binding + keeping state?

```
if (isIOS) {
  setTimeout(this.resize.bind(this), 200);
} else {
  this.resize();
}
```

dmarcos: That will work too

dmarcos: changed

src/core/scene/a-scene.js
```diff
@@ -104,7 +104,18 @@ module.exports.AScene = registerElement('a-scene', {
 
         resize = bind(this.resize, this);
         window.addEventListener('load', resize);
-        window.addEventListener('resize', resize);
+        window.addEventListener('resize', function () {
+          // Workaround for a Webkit bug (https://bugs.webkit.org/show_bug.cgi?id=170595)
+          // where the window does not contain the correct viewport size
+          // after an orientation change. The window size is correct if the operation
+          // is postpoend a few miliseconds.
+          // The if satement below can be removed when the bug is fixed.
```
ngokevin: Couple typos `postponed` `milliseconds` `statement`

antoniohof:
In Cordova/PhoneGap using WKWebview, it does not enter the conditional "if (this.isIOS) ", so the issue remains.

dmarcos:
@antoniohof thanks for investigating. Is there a way to crate an `isCordova()` function to detect the scenario and the condition to the if statement. Feel free to open a PR.

PR #3310, opened 2017-12-19
Improve function names in debugger call stacks

brianpeiris:
**Description:**
Currently most functions in an A-Frame call stack are named "value" due to the way elements are defined and registered.
This change makes use of the [displayName](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName) property on functions to improve call stack readability.

Before:
Debugger call stack with multiple functions named "value"

After:
Debugger call stack with properly named functions

**Changes proposed:**
- Name functions on registered elements using the displayName property.

### DIFF
--- a/src/core/a-register-element.js
+++ b/src/core/a-register-element.js
@@ -59,8 +59,8 @@ module.exports.registerElement = function (tagName, obj) {
   }
 
   // Give all functions their proper name.
-  for (const propName of Object.getOwnPropertyNames(newObj.prototype)) {
-    const propVal = newObj.prototype[propName];
+  for (var propName of Object.getOwnPropertyNames(newObj.prototype)) {
+    var propVal = newObj.prototype[propName];
     if (propVal && typeof propVal === 'function') {
       propVal.displayName = propName;
     }
--- a/src/core/a-register-element.js
+++ b/src/core/a-register-element.js
@@ -61,7 +61,7 @@ module.exports.registerElement = function (tagName, obj) {
   // Give all functions their proper name.
   for (var propName of Object.getOwnPropertyNames(newObj.prototype)) {
     var propVal = newObj.prototype[propName];
-    if (propVal && typeof propVal === 'function') {
+    if (typeof propVal === 'function') {
       propVal.displayName = propName;
     }
   }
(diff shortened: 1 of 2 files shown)


src/core/a-register-element.js
```diff
@@ -58,6 +58,14 @@ module.exports.registerElement = function (tagName, obj) {
     newObj = {prototype: Object.create(proto, newObj)};
   }
 
+  // Give all functions their proper name.
+  for (const propName of Object.getOwnPropertyNames(newObj.prototype)) {
```
ngokevin: not an es6 codebase

tests/core/a-register-element.test.js
```diff
@@ -65,6 +65,20 @@ suite('a-register-element', function () {
       });
       document.body.appendChild(document.createElement('a-test-entity-2'));
     });
+
+    test.only('Names functions correctly', function (done) {
```
ngokevin: remove .only, uncapitalize test name

src/core/a-register-element.js
```diff
@@ -58,6 +58,14 @@ module.exports.registerElement = function (tagName, obj) {
     newObj = {prototype: Object.create(proto, newObj)};
   }
 
+  // Give all functions their proper name.
+  for (var propName of Object.getOwnPropertyNames(newObj.prototype)) {
+    var propVal = newObj.prototype[propName];
+    if (propVal && typeof propVal === 'function') {
```
dmarcos: Is checking for `if ( typeof propVal === 'function')` sufficient?

brianpeiris: Ah, yes. Will fix.

dmarcos:
There's a `test.only` left to remove and this is mergeable:

https://github.com/aframevr/aframe/pull/3310#discussion-diff-157641487R69

brianpeiris:
The `test.only` was removed in `829dcc0`.

dmarcos:
You're right! The Github PR conversation is not the best to keep track of updates.

dmarcos:
Thanks! 👍

brianpeiris:
Sweet, thanks.

PR #3312, opened 2017-12-19
set transforms using object3d vs setAttribute in look-controls

ngokevin:
**Description:**

Porting some critical code to faster/cleaner object3D API for transform updates.

**Changes proposed:**
- Use 3js object3D API for transforms in look-controls for bit of speed.
- Don't need to re-allocate object `this.savedPose` every enter VR.


dmarcos:
I have a PR that touched this logic to fix the user height issues. Some parts of this might end up obsolete. I put it on hold and I’ll submit mine today or tomorrow

dmarcos:
https://github.com/aframevr/aframe/pull/3327 delegates pose update to the VRManager from THREE that changes the object3D directly. It makes this PR redundant.

PR #3316, opened 2017-12-19
get rid of some string split operations

ngokevin:
**Description:**

Copied a util over from #3305 

**Changes proposed:**
- Get rid of few string split operations, namely for `multiple` components

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -234,7 +234,7 @@ function extractDomain (url) {
   var domain = url.indexOf('://') > -1 ? url.split('/')[2] : url.split('/')[0];
 
   // Find and remove port number.
-  return domain.split(':')[0];
+  return domain.substring(0, domain.indexOf(':'));
 }
 
 /**
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -322,3 +322,19 @@ module.exports.findAllScenes = function (el) {
 
 // Must be at bottom to avoid circular dependency.
 module.exports.srcLoader = require('./src-loader');
+
+/**
+ * String split with cached result.
+ */
+module.exports.split = (function () {
+  var splitCache = {};
+
+  return function (str, delimiter) {
+    if (!(delimiter in splitCache)) { splitCache[delimiter] = {}; }
+
+    if (str in splitCache[delimiter]) { return splitCache[delimiter][str]; }
+
+    splitCache[delimiter][str] = str.split(delimiter);
+    return splitCache[delimiter][str];
+  };
+})();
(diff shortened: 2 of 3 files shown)


src/core/a-assets.js
```diff
@@ -234,7 +234,7 @@ function extractDomain (url) {
   var domain = url.indexOf('://') > -1 ? url.split('/')[2] : url.split('/')[0];
 
   // Find and remove port number.
-  return domain.split(':')[0];
+  return domain.substring(0, domain.indexOf(':'));
```
vincentfretin: What about the url.split just above?

ngokevin: Seems a bit unclear about doing it with string splits and didn't want to risk breaking it :)

src/utils/index.js
```diff
@@ -322,3 +322,19 @@ module.exports.findAllScenes = function (el) {
...
+/**
+ * String split with cached result.
+ */
+module.exports.split = (function () {
+  var splitCache = {};
+
+  return function (str, delimiter) {
+    if (!(delimiter in splitCache)) { splitCache[delimiter] = {}; }
+
+    if (str in splitCache[delimiter]) { return splitCache[delimiter][str]; }
```
dmarcos: Not sure how this would actually save since you trade the `split` operation with a key lookup on the object of a potentially long string.

ngokevin: The JS engine hashes strings so string length doesn't affect object lookup. The savings are more around memory versus time. I think in time, a lookup is faster than a split since a split operation needs to walk the string, but also create an array.

ngokevin: The split is faster, but at 350 million operations a second for lookup, I don't think it'll make much difference. But saving on memory should help with GC, where GC can cause hangs on the order of milliseconds. So that's the tradeoff

https://jsperf.com/string-split-vs-cached-lookup/1

dmarcos: I'm not claiming that this patch won't improve things but we have to be always aware of hidden costs. `it is a hash the size of the string does not affect lookup` is not a correct statement. It depends on the implementation. If you have a hash function that additions the character encodings the cost is dependent on the length. If you have a multilevel hash table where you use substrings to organize the keys you might also have different performance depending on the string length. Those costs might trump the benefit of a split operation in the average case.

ngokevin: I agree, I measured it, and the conclusion is I'd prefer to trade off garbage creation which has larger potential impact than the relatively miniscule CPU time saved on split vs. lookup.

ngokevin: Due diligence analysis:

**Saving on memory if caching**

About 32 bytes to create a JavaScript array

**Cost on CPU if caching vs split**

My laptop on FF does:

- 920 million `str.split` per second = 0.00000108695 ms
- 310 million cached lookups per second = 0.0000032258 ms

Costing 0.00000213885 ms versus a string split

So in this trial, it costs 0.00000213885 ms to save 32 bytes which I think is worth the tradeoff.

PR #3317, opened 2017-12-20
save array creations on each raycaster update

ngokevin:
**Description:**

`flattenChildrenShallow` was creating two arrays on each raycaster refresh. And raycasters auto refresh every time there is an attribute change or child append on the scene. And a scene has often multiple raycasters. So this saves quite a ton of arrays.

**Changes proposed:**
- Reuse arrays for raycaster flattenChildrenShallow.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -148,11 +148,13 @@ module.exports.Component = registerComponent('raycaster', {
    */
   refreshObjects: function () {
     var data = this.data;
+    var els;
+
     // If objects not defined, intersect with everything.
-    var els = data.objects
+    els = data.objects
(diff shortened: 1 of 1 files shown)


src/components/raycaster.js
```diff
@@ -38,7 +38,7 @@ module.exports.Component = registerComponent('raycaster', {
     direction: {type: 'vec3', default: {x: 0, y: 0, z: -1}},
     enabled: {default: true},
     far: {default: 1000},
-    interval: {default: 0},
+    interval: {default: 100},
```
vincentfretin: I don't think you meant to revert this change.

ngokevin: I'll do in another commit, i think 100 was appropriate for perf

PR #3320, opened 2017-12-21
Add transparent flag to be able to define transparent backgrounds

dmarcos:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
What about making `background` not a default component, which effectively defaults to transparent. If developer wants a color, they can then set `background.color`. If want transparent, then don't set. At the moment, it defaults to a dark/black, which I'm not sure about.

dmarcos:
@ngokevin it defaults to black. You usually don't want a transparent background to start with.

machenmusik:
The usual case for transparent background is AR.  I don't think we need to force background scene component, if you want to change the background you can just add it yourself...

PR #3323, opened 2017-12-25
Select another intersected element if theres any after the current intersection has cleared

dmarcos:

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -255,9 +255,7 @@ module.exports.Component = registerComponent('cursor', {
    * Handle intersection.
    */
   onIntersection: function (evt) {
-    var self = this;
     var cursorEl = this.el;
-    var data = this.data;
     var index;
     var intersectedEl;
     var intersection;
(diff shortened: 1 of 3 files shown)


src/components/cursor.js
```diff
@@ -325,6 +333,15 @@ module.exports.Component = registerComponent('cursor', {
...
     // Clear fuseTimeout.
     clearTimeout(this.fuseTimeout);
+
+    // Set intersection to another raycasted element if any.
+    intersections = this.el.components.raycaster.intersections;
+    if (intersections.length === 0) { return; }
+    // exclude the cursor.
+    index = intersections[0].object.el === cursorEl ? 1 : 0;
+    intersection = intersections[index];
+    if (!intersections[index]) { return; }
```
ngokevin: can do `if (!intersection)`

src/components/cursor.js
```diff
@@ -279,14 +277,33 @@ module.exports.Component = registerComponent('cursor', {
...
     if (this.intersectedEl) { this.clearCurrentIntersection(); }
 
+    this.setIntersection(intersectedEl, intersection);
+  },
+
+  /**
+   * Handle intersection cleared.
+   */
+  onIntersectionCleared: function (evt) {
+    var clearedEls = evt.detail.clearedEls;
```
ngokevin: could do 

```js
if (evt.detail.clearedEls.indexOf(this.intersectedEl) === -1) { return; }
this.clearCurrentIntersection();
```

dmarcos:
Done

PR #3324, opened 2017-12-26
Document that component data is readonly

brianpeiris:
**Description:**
It's not obvious that component data is meant to be readonly. This can lead to confusion if a developer modifies data directly and it is later clobbered by a-frame.

**Changes proposed:**
- Add a note to the docs to clarify that component data is readonly and that `setAttribute` should be used instead.

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -12,7 +12,7 @@ examples: []
 [entity]: ./entity.md
 [multiple]: #multiple
 [three]: http://threejs.org/
-[setAttribute]: ../introduction/javascript-events-dom-apis.md#updating-a-component-with-setattribute-
+[setAttribute]: ../introduction/javascript-events-dom-apis.md#updating-a-component-with-setattribute
 
 In the [entity-component-system pattern][ecs], a component is a reusable and
 modular chunk of data that we plug into an entity to add appearance, behavior,
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -12,6 +12,7 @@ examples: []
 [entity]: ./entity.md
 [multiple]: #multiple
 [three]: http://threejs.org/
+[setAttribute]: ../introduction/javascript-events-dom-apis.md#updating-a-component-with-setattribute-
 
 In the [entity-component-system pattern][ecs], a component is a reusable and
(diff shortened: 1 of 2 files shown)


docs/core/component.md
```diff
@@ -244,12 +244,12 @@ the data to modify the entity. The handlers will usually interact with the
...
-| this.el.sceneEl | Reference to the [scene][scene] as an HTML element.                                                                                           |
-| this.id         | If the component can have [multiple instances][multiple], the ID of the individual instance of the component (e.g., `foo` from `sound__foo`). |
+| Property        | Description                                                                                                                                                                                                                                         |
+|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| this.data       | Parsed component properties computed from the schema default values, mixins, and the entity's attributes. <br/>__Note:__ The data property should be considered readonly. If a component wants to update its own data, it should use *setAttribute*.|
```
dmarcos: Should we remove it all together from the docs to avoid temptations? 😄

brianpeiris: well, component authors need to know about the existence of `this.data`. Ideally we could solve this in code by using getters and setters and write-disabled properties but that's probably more work than it's worth.

dmarcos: I was thinking that maybe the mention in the different methods documentation could be enough: https://aframe.io/docs/0.7.0/core/component.html#update-olddata
But yeah you also need to be aware of it for any other component method.

dmarcos: What about:  
**Important:** Do not modify the data attribute directly. It is updated internally by A-Frame. To modify a component use *[setAttribute](https://aframe.io/docs/0.7.0/introduction/javascript-events-dom-apis.html#updating-a-component-with-setattribute)*

brianpeiris: 👍 sounds good to me. I've pushed the change.

dmarcos:
Sweet! Thanks 👍

dmarcos:
Doc tests failing. It complains about the link to `setAttribute`

dmarcos:
Thanks 🎉

Issue #3118, opened 2017-10-04
Camera height changes every time user enters in VR mode


VitoMatarazzo:
**Description:**

- A-Frame Version: 0.7.0
- Platform / Device: Chrome on Android phone

I have a scene with "a-camera" placed in the default position. When I enter VR mode on a mobile browser on different Android smartphones the camera automatically goes up of 1.6, while upon exiting VR mode the height doesn't return to the previous value. So the height increases everytime the user enters VR. I have tried to manually set the position of the camera and change the userHeight parameter but the problem remains.

I managed to solve this problem by resetting the camera to the default height -1.6 upon entering VR, but I don't think this is how the camera should work.

dmarcos:
Thanks! This is a dup of https://github.com/aframevr/aframe/issues/3051

alessandrocaprarelli:
I have the same issue in Chrome version 62 on Android phone with A-frame Version 0.7.1.

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/3327

TheDirigible:
It's still happening for me, what's the solution?

Issue #2487, opened 2017-03-12
Camera userHeight offset is requiring some hacks.


donmccurdy:
- A-Frame Version: 0.5.0

I'm having recurring problems stemming from the `userHeight` offset in `camera`.

1. It defaults to `0`, except if you don't include a camera in your scene, in which case it defaults to `1.6`. This is likely confusing for new users.
2. One of the less-awful locomotion mechanisms for mobile users is something like `checkpoint-controls`. I would like to do one of the following:

```html
<a-entity universal-controls="movementControls: checkpoint"
          checkpoint-controls="mode: teleport">
  <a-entity camera="userHeight: 1.6"></a-entity>
</a-entity>
```

^This breaks, because the camera's offset is inside the entity being rotated. Looking down moves the camera closer to the ground.

***

```html
<a-entity camera 
          position="0 1.6 0"
          universal-controls="movementControls: checkpoint"
          checkpoint-controls="mode: animate">
</a-entity>
```

^This mostly works with mobile, but breaks with roomscale devices, and all of the checkpoints need to include this 1.6 offset as their `y` values.

***

```html
<a-entity camera="userHeight: 1.6"
          universal-controls="movementControls: checkpoint"
          checkpoint-controls="mode: animate">
</a-entity>
```

This looks better, but fails in all the same ways as the last example – the camera is going to directly modify the position of the entity (which ideally a camera would not do...) and now the checkpoints need to somehow be device-aware and include or exclude the offset appropriately.

***

Getting into the longer discussion of #2265, but I wanted to give a concrete example of the difficulty. I'd like to be able to give people a default that more-or-less works with both mobile and roomscale, but it's not obvious how to do that without hardcoding userHeight in a lot of places.

machenmusik:
so iy seems that you want a separate component that will apply default user height like camera, and use that instead of default camera behavior?

donmccurdy:
I'm not sure. I think that these methods...

```
camera.removeHeightOffset()
camera.saveCameraPose()
camera.restoreCameraPose()
```

... would ideally be in `look-controls`, not `camera`. In that case I could implement my own component that handles the height offset without worrying about a camera that mutates itself.

dmarcos:
yeah. probably look-controls is a better place for that logic

machenmusik:
look-controls is used for more than just cameras, which is why i think carving off the mutating portion into a separate component may be safer...

machenmusik:
[...]

PR #3327, opened 2017-12-27
look-controls no longer manage the camera pose in VR mode (fix #3315 #3118 #3051 #2487)

### LINKED ISSUE #3315: userHeight not applied when there are not stageParameters
**Description:**

- A-Frame Version: 0.7
- Platform / Device: 6DOF Mobile headset

After https://github.com/aframevr/aframe/commit/3d188db12742d162b2c9e60ffd0222f75acaffa1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 it seems that userHeight is not applied in devices using null stageParameters and the camera is on ground level. 

Should this be fixed on A-frame or do we need to fix this in Three.js? https://github.com/mrdoob/three.js/issues/11619

I have a wip in the Three.js WebVRManager class to to handle standingTransform and could make a PR if you want the code to be there.

What do you think?

dmarcos:
Instead VRManager from THREE does update the pose applying always `standing` coordinates so a user height is always present in the pose: defaulting to 1.6 in the case of 3DOF headsets.

### DIFF
--- a/docs/components/camera.md
+++ b/docs/components/camera.md
@@ -16,7 +16,7 @@ to move and rotate the camera.
 A camera situated at the average height of human eye level (1.6 meters).
 
 ```html
-<a-entity camera="userHeight: 1.6" look-controls></a-entity>
+  <a-entity position="0 1.6 0" camera look-controls></a-entity>
 ```
 
 ## Properties
@@ -36,21 +36,13 @@ A camera situated at the average height of human eye level (1.6 meters).
 If a camera is not specified, A-Frame will inject a default camera:
 
 ```html
-<a-entity camera="active: true; userHeight: 1.6" look-controls wasd-controls position="0 0 0" data-aframe-default-camera></a-entity>
+<a-entity camera="active: true" look-controls wasd-controls position="0 0 0" data-aframe-default-camera></a-entity>
 ```
 
 If a camera is specified (e.g., our own `<a-camera>` or `<a-entity camera>`),
 then the default camera will not be added.
 
 ## VR Behavior
-
-When not in VR mode, `userHeight` translates the camera up to approximate
-average height of human eye level. The injected camera has this set to 1.6
-(meters). When entering VR, this height offset is *removed* such that we used
-absolute position returned from the VR headset. The offset is convenient for
-experiences that work both in and out of VR, as well as making experiences look
-decent from a desktop screen as opposed to clipping the ground if the headset
-was resting on the ground.
 
 When exiting VR, the camera will restore its rotation to its rotation *before*
 it entered VR. This is so when we exit VR, the rotation of the camera is back
--- a/docs/components/look-controls.md
+++ b/docs/components/look-controls.md
@@ -26,14 +26,13 @@ component](camera.md).
 
 [pointer-lock-api]: https://developer.mozilla.org/docs/Web/API/Pointer_Lock_API
 
-| Property           | Description                                                      | Default Value |
-|--------------------|------------------------------------------------------------------|---------------|
-| enabled            | Whether look controls are enabled.                               | true          |
-| hmdEnabled         | Whether to use VR headset pose in VR mode.                       | true          |
-| reverseMouseDrag   | Whether to reverse mouse drag.                                   | false         |
-| touchEnabled       | Whether to use touch controls in magic window mode.              | true          |
+| Property         | Description                                                      | Default Value |
+|------------------|------------------------------------------------------------------|---------------|
+| enabled          | Whether look controls are enabled.                               | true          |
+| hmdEnabled       | Whether to use VR headset pose in VR mode.                       | true          |
+| reverseMouseDrag | Whether to reverse mouse drag.                                   | false         |
+| touchEnabled     | Whether to use touch controls in magic window mode.              | true          |
 | pointerLockEnabled | Whether to hide the cursor using the [Pointer Lock API][pointer-lock-api]. | true |
-| userHeight         | Height offset to add to the camera when *not* in VR mode so the camera is not on ground level. The default camera that A-Frame injects or the `<a-camera>` primitive sets this to 1.6 meters. But note the default camera component alone (`<a-entity camera>`) defaults this to 0. | 0 |
 
 ## Customizing look-controls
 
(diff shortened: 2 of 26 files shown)


ngokevin:
r- as is

This stacks the Vive's absolute pose in the room on top of a defined 1.6m on the camera rig. Resulting in being way too high up.

dmarcos:
This PR adapts to the changes proposed to the THREE VRManaged in https://github.com/mrdoob/three.js/pull/13158 Tested on:

- [x] Oculus Rift
- [x] Windows MR
- [x] Vive

ngokevin:
Shouldn't the non-VR mode user offset logic remain? We want a 1.6m (or configurable) user height on 2D mode, but when entering VR, use the headset pose.

See `laser-controls` is on the ground:

screen shot 2018-01-23 at 3 18 00 pm

If we try to add `position="0 1.6 0"` to the camera rig, then in VR, the height will be too high.

dmarcos:
@ngokevin It is up to you to put the camera where you want in 2D mode. You can just place the camera where you want. You don't need a camera rig. The default camera A-Frame camera is placed at 1.6m already for people they don't want to set their own camera.

dmarcos:
@ngokevin Look at `tracked-controls` as an example

ngokevin:
r+

Can you dump some knowledge here for future reference? Everything about the sitting to standing transform, how the camera is handled internally by three.js, etc.

dmarcos:
The code is now much simpler than before. The camera pose is managed completely by THREE and you will always have a user scale height: Either the one coming from the VRDevice sensors or a 1.6m default otherwise. The use of sittingToStandingTransform can be safely ignored by A-Frame. All poses coming from VRDevices are normalized to `standing mode`. In other words, the camera height will always be relative to the ground: measured by the tracking system, inferred by the headset user configuration or default to 1.6m in the case of 3DOF headsets.  I updated the commit comment to reflect what this PR ended up doing that it is different that the original camera rig approach. By normalizing to `standing position` it is still possible to provide a camera rig to move the user around like for instance the teleport component.

donmccurdy:
Lots of questions about how to use this now. 😅 If camera pose is handled by THREE for 6DOF, I don't see any code here that would set the camera's position back onto the entity now... How do I look up the user's position in the scene? And how would components like `teleport-controls` move the player from some current position to, say, a target position at a different elevation?

In `tracked-controls/index.html` example I see:

```html
<a-entity position="0 1.6 2" camera look-controls wasd-controls></a-entity>
```

Please correct me on any of this that is wrong:

1. For 2D devices, camera height is `1.6` and that's configurable through the `position` component.
2. For 3DOF devices, camera height is `1.6` and that's __not__ configurable because it's hard coded in WebVRManager.
3. For 6DOF devices, camera position is handled by THREE, and is entirely independent of the starting position placed on the camera entity? Or it reflects initial `position.x` and `.z` but not `.y`?

dmarcos:
Magic, in 0.8.0 you don't need to do `el.setAttribute('position')` you can modify directly `el.object3D.position.set()` and you will get back the correct value when doing `el.getAttribute('position')`. The same goes for the rotation. THREE updates the pose on the entity.

To move the user around you just create a camera rig that you move around. `teleport-controls` is moving to a camera rig model already (https://github.com/fernandojsg/aframe-teleport-controls/blob/master/examples/mesh/index.html#L44)

```
<a-entity id="cameraRig">
  <a-entity camera look-controls></a-entity>
  <a-entity hand-controls="left"></a-entity>
  <a-entity hand-controls="right"></a-entity>
</entity>
```
To position the camera in 2D mode you just position the camera:
```
<a-entity position="0 1.6 0"></a-entity>
````
The position is saved when entering VR mode and restored afterwards. The A-Frame default camera already sets the height to 1.6.

For 3DOF devices the height is 1.6 by default. Not configurable but you can move up and down the user with a camera rig if necessary.

**Comment by dmarcos** [comment]
Thanks to @arturitu for testing on:
- [x] Chrome on Daydream
- [x] Oculus Browser on GearVR
- [x] Internet Browser on GearVR

**Comment by donmccurdy** [comment]
Ok, I think this all sounds good but just to clarify....

```html
<a-entity id="cameraRig" position="25 100 25">
  <a-entity id="head" camera position="0 1.6 0" look-controls></a-entity>
  <a-entity teleport-controls="cameraRig: #cameraRig; teleportOrigin: #head;"></a-entity>
  <a-entity teleport-controls="cameraRig: #cameraRig; teleportOrigin: #head;"></a-entity>
</entity>
```

Means that...

1. The "floor" is essentially at y=100, so if my height is 1.51 meters in a 6DOF device the `#cameraRig` position will be left alone, but the `#head` position will change to `0 1.51 0`, for a cumulative world position of `25 101.51 25`.
2. If I need the position of the player relative to the scene floor (e.g. for nav mesh stuff) I'll use `cameraRigEl.object3D.position`
3. If I need the position of the player's head, I'd use `cameraEl.object3D.getWorldPosition()`
4. Teleport controls will move the rig, not the camera, and most community components should do the same.
5. The cameraRig can safely be rotated around the Y axis but probably users shouldn't change pitch/roll.

dmarcos:
@donmccurdy correct to all of the above 😄

donmccurdy:
Ok, LGTM. 👍

Other thoughts: 

1. `wasd-controls` should probably either offer a `cameraRig` or `cameraEl` property, so that it can move the rig but set direction according to the camera heading.

2. I might delete a lot of the rotation-related stuff from `universal-controls`, and focus only on components that affect position instead... so for example components like `touch-` `checkpoint-` and `gamepad-controls` will move the rig around but will leave rotation to the `look-controls` component. That way things are less complicated for everyone and I can focus on things like navmesh traversal rather than these userHeight issues. 😅

dmarcos:
@donmccurdy Can you file an issue for the `wasd-controls` property to indicate a camera rig? 

Rotation composition is tricky 😄

dmarcos:
Thanks everybody for your help on this PR. I got stuck for a while unable to make a decision I would feel comfortable with.

machenmusik:
Nice!

machenmusik:
I do think that as best practice for 6DOF we could use example that use a player rig and teleport (even if we only do the shoot-the-floor version with laser-controls)

mattrei:
The gaze cursor  is now disappearing when in VR mode on mobile devices. See related issue  #3375

PR #3334, opened 2018-01-11
[docs] Update cursor page with mouse click example.

madebynoam:
Added a Glitch example to the docs, to demonstrate how to allow mouse clicks on entities with A-Frame.

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -4,7 +4,9 @@ type: components
 layout: docs
 parent_section: components
 source_code: src/components/cursor.js
-examples: []
+examples:
+  - title: Mouse Click Example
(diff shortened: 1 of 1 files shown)


dmarcos:
Cool thanks! Can you add a hover effect on the elements to emphasize you can interact with them using the mouse?

madebynoam:
Sounds good. Just updated the Glitch project with basic hover effects for interaction feedback.
Decided to use the event-set component for those feedback interactions, since it's simpler to demonstrate than writing custom components in this context. But let me know if you think it should use a custom component with addEventListerer.

dmarcos:
Thanks! Congrats for your first PR

madebynoam:
Awesome, thanks!

PR #3335, opened 2018-01-11
Updated links to use https when possible

mbrickn:
**Description:**

I updated links to use HTTPS instead of HTTP when possible.
This should help improve user experience by avoiding redirects.

Thank you for your time!

**Changes proposed:**

- Use https when possible in documentation, while leaving links to websites without certs and localhost as http.


mbrickn:
Thank you! 😄

PR #3341, opened 2018-01-17
[look-controls] Support Pointer Lock API.

### LINKED ISSUE #2625: Proposal: add Pointer Lock to look-controls only for desktop fullscreen mode without headset
**Description:**

Pointer Lock is an easier way to look around on desktop than mouse-drag. Mouse-drag also has problems that it triggers clicks on time users move. We could add Pointer Lock to the default control scheme if:

1. User is on desktop
2. User does not have a headset detected
3. User enters fullscreen mode

It seems like Pointer Lock doesn't prompt permissions on Firefox and Chrome anymore too when I try Pointer Lock example on MDN with user gesture. So it's not intrusive, use the VR Button click as the user gesture.

We maintain the mouse-drag for now when not fullscreen.

donmccurdy:
Fixes #2625.

From the last comment:

> Yeah, I'm in favor of a Pointer Lock on by default. I shared some A-Frame stuff to people on desktop and the click-drag is not usable. For default desktop mode, I'd favor Pointer Lock/a-cursor versus Click-Drag/mouse-cursor (they conflict anyways).

... it sounds like we're leaning toward "on by default" rather than "on by default after entering fullscreen". If that's still correct and we want to merge this, I'll add the tests and docs.

/cc @ngokevin

### DIFF
--- a/docs/components/look-controls.md
+++ b/docs/components/look-controls.md
@@ -10,7 +10,7 @@ examples: []
 The look-controls component:
 
 - Rotates the entity when we rotate a VR head-mounted display (HMD).
-- Rotates the entity when we click-drag mouse.
+- Rotates the entity when we move the mouse.
 - Rotates the entity when we touch-drag the touchscreen.
 
 ## Example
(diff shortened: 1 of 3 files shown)


donmccurdy:
The A-Frame website with a scene embedded in other 2D page content seems like the exception rather than the rule, and it can always be disabled... would you suggest other defaults? Default the option to off? Remove the option but automatically enable it pointerlock in fullscreen mode?

It's a pretty significant usability improvement for more cases than not, in my opinion.

ngokevin:
I still prefer defaulting it. Embedded cases seem like an exception to me too, and even then, the canvas is often still the focus of the page. With embedded sites, mouse-drag becomes more difficult to use because the canvas is a smaller click target to constantly aim back at after you drag away. When you click on the canvas, you intend to engage with it, so it's fine to me to provide Pointer Lock since it's easier to use.

dmarcos:
It needs tests fixed

donmccurdy:
Docs and tests added. ✅

dmarcos:
Thanks!

PR #3347, opened 2018-01-21
Adds backToFront param to pool.requestEntity(), to help w/ transparency

DougReeder:
**Description:**
See https://aframe.io/docs/0.7.0/components/material.html#transparency-issues for context.

See https://glitch.com/edit/#!/trail-cause for an example of transparency issues.

One cannot use HTML to re-order pool objects, so this adds a backToFront param to pool.requestEntity(). If true, the available object created *last* is returned.  This is useful when each entity pulled from the pool is placed *farther* from the camera than previous entity, as when the player is flying through clouds.

For an example of use, see https://github.com/DougReeder/elfland-glider/tree/master/ginnungagap/index.html#L41

If it would work better as a parameter when creating the pool, I can rework this.

**Changes proposed:**
-
-
-

I agree to license my contributions under the MIT License.

### DIFF
--- a/src/components/scene/pool.js
+++ b/src/components/scene/pool.js
@@ -67,8 +67,11 @@ module.exports.Component = registerComponent('pool', {
 
   /**
    * Used to request one of the available entities of the pool
+   * @param backToFront pass falsy if objects retrieved from the pool will generally be placed *nearer* the camera than
+   * older objects, such as bullets from the player's gun.  Pass truthy if they'll be *farther*, such as clouds the
+   * player is flying through.  @See https://aframe.io/docs/0.7.0/components/material.html#transparency-issues
    */
-  requestEntity: function () {
+  requestEntity: function (backToFront) {
     var el;
     if (this.availableEls.length === 0) {
       if (this.data.dynamic === false) {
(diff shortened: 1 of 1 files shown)


dmarcos:
I’m not 100% sure what it is the problem this solves. When you pull entities from the pool you can position them wherever you want. In the case of bullets you pull an entity from the pool, place it at the tip of the gun and animate it. The application has full control where the entities are placed. There is also the renderOrder parameter from Object3D for finer control: https://threejs.org/docs/#api/core/Object3D.renderOrder

DougReeder:
Setting the position of entities works fine.

The end goal is transparency working properly.  The renderOrder property does exist on the meshes, but setting it does not appear to have any effect on transparency.  Transparency *does* depend on the order in which objects are created.   See [this glitch](https://glitch.com/edit/#!/trail-cause).  The second cloud is created after the first cloud and is nearer the camera. Transparency works properly for these two entities.  The third cloud is created after the first two, but is further away. The first two entities completely occlude the third, instead of being transparent.

dmarcos:
I would let content manage the position of the entities. The main problem of this PR is that it couples the internal data structure of the pool with the public API. The pool contract should be just `give me an available entity if any` / `return this entity to the pool`. External code should not care if there's a list, stack or any data structure internally. If that data structure changes the public API breaks.

DougReeder:
I'd prefer that transparency be independent of object allocation. However, *renderOrder does not* appear to have an effect, but the *order of allocation does*.  Can you show me an example like [this glitch](https://glitch.com/edit/#!/trail-cause) where transparency works properly, without using the order of allocation?

dmarcos:
I think the proper solution is just position the entities so they stack in the proper order.

dmarcos:
@ngokevin What do you think here?

PR #3369, opened 2018-02-02
Bug in 360 image gallery tutorial

anselanza:
**Description:**
The id for the "plane" script (template) is incorrectly set to "link" in the example, which is confusing and also breaks the example.

**Changes proposed:**
- Use correct id in the example


anselanza:
Closing because there are multiple confusing things about this tutorial!

Issue #3288, opened 2017-12-13
360 video is broken


machenmusik:
https://aframe.io/aframe/examples/boilerplate/360-video/ works.
https://a-frobot.github.io/aframe/examples/boilerplate/360-video/ only shows black.

machenmusik:
https://a-frobot.github.io/aframe/examples/test/videosphere/ is also broken and only shows black.

mkungla:
duplicate of #3287

ngokevin:
Reports that this can be fixed with `material.side` = `back`. Perhaps the negative Z scale thing isn't working out anymore?

machenmusik:
@mkungla it's not a duplicate in the sense that text being broken is separate from 360 video being broken, right?  although the root cause may be the same.  @ngokevin hmm, so that isn't great... hadn't seen any reason for that to break with newer three.js?

machenmusik:
So what did we decide to do here, change videosphere to use `side:back`, or make negative scale work again?  (If the former, we probably want to note negative scale change as potentially breaking)

dmarcos:
We bumped a-frame to r89. Is this still a problem?

dmarcos:
Yep, it seems to reproduce on master.

machenmusik:
(Sadly, yes)

ngokevin:
https://github.com/mrdoob/three.js/pull/12787

You need to try to not use negative scale anymore

vincentfretin:
I found this issue when trying to use aframe master on my project.
I confirm the issue is related to mrdoob/three.js#12787
I have lots of the following warnings too:
`
THREE.Matrix4: .getInverse() can't invert matrix, determinant is 0
THREE.Matrix3: .getInverse() can't invert matrix, determinant is 0
`
because of `frontFaceCW` calculation in mrdoob/three.js#12787 (I tested to set it to false and do my own aframe build, the warnings and the black texture issue are gone)

You need `material="side:back; ..."` and a negative scale like scale="-1 1 1" otherwise the texture is flipped. But I still have warnings `.getInverse() can't invert matrix, determinant is 0`.

dmarcos:
@vincentfretin Thanks for investigating. It would be great if you have some bandwidth to open a PR with the fix.

machenmusik:
My recollection is that changes are needed to `a-sky`, `a-videosphere`, and `a-curvedimage` although I haven't had time yet for PR.  But I haven't seen that matrix error yet

[...]

PR #3370, opened 2018-02-02
fix black texture issue with a-sky and a-videosphere components with three r89

vincentfretin:
Fix black texture issue with a-sky and a-videosphere components with three r89 because of the changes https://github.com/mrdoob/three.js/pull/12787

scale needs to stay -1 1 1 otherwise texture is flipped.

no-change needed for a-curvedimage

This fixes https://github.com/aframevr/aframe/issues/3288


machenmusik:
Nice!

PR #3371, opened 2018-02-02
Detail how to enable Daydream controls support

gmarty:
**Description:**
It took me a long time to realise that the Gamepad extensions had to be enabled in order to use the controller.

**Changes proposed:**
This is a documentation change. I hope it is clear enough.

### DIFF
--- a/docs/components/daydream-controls.md
+++ b/docs/components/daydream-controls.md
@@ -14,6 +14,13 @@ It wraps the [tracked-controls component][trackedcontrols] while adding button
 mappings, events, and a Daydream controller model that highlights the touched
 and/or pressed buttons (trackpad).
 
+As a first time set up, to use the Daydream controller on Chrome for Android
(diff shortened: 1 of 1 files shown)


gmarty:
Yes, I thought about that and the info is up to date on webvr.rocks.
We can then add a link to https://webvr.rocks/chrome_for_android#android, that contains the steps required to make it work.

Another way around would be to inform the developers. For example if a Daydream controller is detected but it doesn't have a `pose` property, then we can emit an event or a warning in the console asking to enable the Gamepad extensions.

ngokevin:
I think this is a sufficient place to put it. If it becomes outdated, won't be too bad, and someone will probably fix it.

gmarty:
Is this PR ready to be merged? It was approved 10 days ago.

dmarcos:
Thanks! 👍

PR #3372, opened 2018-02-03
Fix parameters passed to the onButtonEvent utils function from the wi…

dmarcos:
…ndows-motion-controls component

**Description:**

**Changes proposed:**
-
-
-


ngokevin:
r+

PR #3376, opened 2018-02-05
Fix clipAction loop method in hand controls.

johnshaughnessy:
`THREE.PingPong` is not a defined constant. `THREE.LoopPingPong` is, but does not exhibit the desired behaviour. `THREE.LoopRepeat` seems to be correct.


dmarcos:
@johnshaughnessy Thanks! Have you tried the track-controls example to check that the hand animations still work as expected?

ngokevin:
wfm. Doesn't change the behavior much, but I guess this is more correct.

PR #3378, opened 2018-02-08
fix for transparency issue with png images

js-d-coder:
https://github.com/aframevr/aframe/issues/666#issuecomment-302831463

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -335,7 +335,7 @@ the elements in markup. We can often run into scenarios where foreground
 entities occlude background entities. This creates confusion and unwanted
 visual defects.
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
+1 for mentioning this, although I might clarify it a bit:

> When using PNG images as cutouts or masks (where part of the image should be fully transparent, and the rest fully opaque), use `transparent=false` and set `alpha-test` to something like 0.5.

dmarcos:
@js-d-coder This almost ready. Are you incorporating the changes that @donmccurdy suggested?

ngokevin:
Would also use `material="alphaTest: 0.5"` versus the primitive form.

ngokevin:
Will address comments in follow up commit.

PR #3383, opened 2018-02-10
[text] Add .negate property for msdfgen >1.4

### LINKED ISSUE #3382: Add 'negate' option to text component
three-bmfont-text is based on an older version of msdfgen, and fonts generated with the newer version don't work. I've proposed a fix (https://github.com/Jam3/three-bmfont-text/pull/24) but we'd need to add the same option to our own shader.

donmccurdy:
Fixes #3382.

Testing with fonts created here: http://msdf-bmfont.donmccurdy.com/

```html
text="align: center; width: 6;
    font: /examples/assets/fonts/yahei.json;
    value: 你好，世界;
    negate: false;
    shader: msdf;"
```

screen shot 2018-02-09 at 11 12 25 pm


machenmusik:
Nice!  Maybe we should regenerate the older default font with the new tool, and default negate to false (at least for msdf)  so all new msdf fonts we create will work without the flag?

donmccurdy:
The fonts in aframe-fonts are generated from https://github.com/Jam3/msdf-bmfont, which is still using the old msdfgen version and would need to be updated, too. Users reporting issues because of that: https://github.com/etiennepinchon/aframe-fonts/issues/1#issuecomment-352407302

I'm hoping my little web app can help out with that, and let you generate just the character set you need (especially necessary for chinese users), but i need to [iron out some issues](https://github.com/soimy/msdf-bmfont-xml/issues/21) first. So, probably won't want to change the default for 0.8.0 I don't think.

PR #3388, opened 2018-02-13
AFRAME.registerComponent("name",...

polygontwist:
// <a-entity "name"></a-entity> 
that should be the same

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/introduction/developing-with-threejs.md
+++ b/docs/introduction/developing-with-threejs.md
@@ -140,7 +140,7 @@ denotes the `Object3D`s purpose.
 For example, to set a point light from within a component:
 
 ```js
-AFRAME.registerComponent('pointlight', {
+AFRAME.registerComponent('light', {
   init: function () {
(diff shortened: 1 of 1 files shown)


donmccurdy:
I think this example should remain "pointlight", registering "light" will throw an error as it's already defined. The intention here is to show how a user could create a new component for their particular needs. Maybe `area-light`, using `new THREE.RectAreaLight( 0xffffff, undefined,  2, 10 );`, would be a better example since that's not already provided by A-Frame?

dmarcos:
@donmccurdy is correct. Thanks for submitting

polygontwist:
in the following "Removing an Object3D From an Entity" is the same error

dmarcos:
@polygontwist You might have got confused with object3D named `light` which is different than the component name. The example illustrate how to add / remove object3Ds on a custom component and it is named `pointLight` so it does not collide with the built in `light` component in A-Frame.

Issue #3389, opened 2018-02-15
aframe master  0.7.0 (Date 2018-02-10, Commit #5f4179d) stats vr-mode-ui and inspector not shown ( behind sky) 


arpu:
**Description:**
aframe master  0.7.0 (Date 2018-02-10, Commit #5f4179d) stats vr-mode-ui and inspector not shown ( behind sky) 
- A-Frame Version: aframe master  0.7.0 (Date 2018-02-10, Commit #5f4179d)
- Platform / Device: fedora chrome
- Reproducible Code Snippet or URL:https://a-frobot.github.io/aframe/examples/boilerplate/hello-world/

if i remove the sky and the set the background none  i can see the stats but the 
vr-mode-ui is not clickable ->
https://subsequent-morocco.glitch.me/
looks like a z-index problem and all the gui elements are behind the sky

PR #3391, opened 2018-02-19
remove a-canvas z-index

arpu:
solves the problem in https://github.com/aframevr/aframe/issues/3389
in chrome and firefox on fedora


dmarcos:
Thanks!

PR #3407, opened 2018-02-28
Fix control transform components and rotationOffset

brianpeiris:
**Description:**
#3327 changed the way we calculate tracked controller pose:
- As a side effect, the code no longer updated the individual `position` and `quaternion` properties of the controller object. This prevents users and libraries (like `networked-aframe`) from reading those properties.
- It also removed use of `tracked-control`'s `rotationOffset` property, this lead to `hand-controls` with incorrectly rotated hand models when using the Oculus Rift and Touch, as evident in the tracked-controls example.
- Lastly, it regressed standing matrix application for 3DoF controls -- i.e. standing matrix should only be applied for 6DoF controls, since the controller position is controlled by the arm model, which bases it off the head position.

**Changes proposed:**
- Decompose the tracked object's matrix in to its individual components
- Reintroduce use of the `rotationOffset` property.
- Only apply the standing matrix if the controller is 6DoF
- Add tests to cover these changes.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -169,8 +169,13 @@ module.exports.Component = registerComponent('tracked-controls', {
       object3D.matrixAutoUpdate = false;
       object3D.matrix.compose(object3D.position, object3D.quaternion, object3D.scale);
       object3D.matrix.multiplyMatrices(standingMatrix, object3D.matrix);
-      object3D.matrixWorldNeedsUpdate = true;
+      object3D.matrix.decompose(object3D.position, object3D.quaternion, object3D.scale);
     }
+
+    object3D.rotateZ(this.data.rotationOffset * THREE.Math.DEG2RAD);
+
+    object3D.updateMatrix();
+    object3D.matrixWorldNeedsUpdate = true;
   },
 
   /**
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -152,19 +152,19 @@ module.exports.Component = registerComponent('tracked-controls', {
     // Compose pose from Gamepad.
     pose = controller.pose;
 
-    if (pose.position !== null) {
+    if (pose.position) {
       object3D.position.fromArray(pose.position);
     } else {
       // Controller not 6DOF, apply arm model.
       if (this.data.armModel) { this.applyArmModel(object3D.position); }
     }
 
-    if (pose.orientation !== null) {
+    if (pose.orientation) {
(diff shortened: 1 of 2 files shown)


src/components/tracked-controls.js
```diff
@@ -169,8 +169,13 @@ module.exports.Component = registerComponent('tracked-controls', {
       object3D.matrixAutoUpdate = false;
       object3D.matrix.compose(object3D.position, object3D.quaternion, object3D.scale);
       object3D.matrix.multiplyMatrices(standingMatrix, object3D.matrix);
-      object3D.matrixWorldNeedsUpdate = true;
+      object3D.matrix.decompose(object3D.position, object3D.quaternion, object3D.scale);
     }
+
+    object3D.rotateZ(this.data.rotationOffset * THREE.Math.DEG2RAD);
+
+    object3D.updateMatrix();
```
brianpeiris: I haven't tested this PR in a Daydream/GearVR with a 3DoF controller yet. We disable `matrixAutoUpdate` in `init`, so I'm not sure how this code could have worked for those controllers without an explicit `updateMatrix`. Perhaps it _was_ actually broken. I'll test with a 3DoF controller when I get a chance.

src/components/tracked-controls.js
```diff
@@ -164,13 +164,18 @@ module.exports.Component = registerComponent('tracked-controls', {
     }
 
     // Apply transforms, if 6DOF and in VR.
-    if (vrDisplay) {
+    if (vrDisplay && pose.position !== null) {
```
brianpeiris: @dmarcos After testing in Daydream, I found that this change was necessary as well. Just wanted to ensure it makes sense to you as well.

dmarcos: It makes sense, we can just make position` (vrDisplay && pose.position)`

tests/components/tracked-controls.test.js
```diff
@@ -130,6 +132,23 @@ suite('tracked-controls', function () {
       component.tick();
       assertVec3(el.getAttribute('position'), [2, 4, 6]);
     });
+
+    test('applies standing matrix transform', function () {
```
dmarcos: Thanks for the tests

dmarcos:
Thank you @brianpeiris 👍

src/components/tracked-controls.js
brianpeiris: Confirmed. Daydream controller does not update without this.

dmarcos: Thanks for testing on it.

PR #3409, opened 2018-03-01
Clone the properties that are object type into oldData so the update …

### LINKED ISSUE #3403: Line not being updated correctly
**Description:**

On the following example: https://glitch.com/edit/#!/inconclusive-thrill I'm updating the line between to spheres:
```
line.setAttribute('line', {start: posL, end: posR});
``` 
It doesn't work as expected, but If I add `debug: true` to the scene, it works.

I guess it's something related to the latest changes to optimize the set/getAttribute?

dmarcos:
…method is called when they change. We were holding a reference and the deepEqual test was passing and the update method was never called (fix #3403)

-

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -522,7 +522,16 @@ function cloneData (data) {
 */
 function extendProperties (dest, source, isSinglePropSchema) {
   if (isSinglePropSchema && (source === null || typeof source !== 'object')) { return source; }
-  return utils.extend(dest, source);
+  var copy = utils.extend(dest, source);
+  var keys = Object.keys(copy);
+  var key;
+  var i;
+  for (i = 0; i < keys.length; i++) {
+    key = keys[i];
+    if (!copy[key] || copy[key].constructor !== Object) { continue; }
+    copy[key] = utils.clone(copy[key]);
+  }
+  return copy;
 }
 
 /**
--- a/tests/components/line.test.js
+++ b/tests/components/line.test.js
@@ -69,6 +69,17 @@ suite('line', function () {
       assert.equal(positionArray[5], 9);
     });
 
+    test('set end but no start position', function () {
+      var end = {x: 4, y: 5, z: 6};
+      el.setAttribute('line', {end: end});
+      end.z = 9;
+      var updateSpy = this.sinon.spy(el.components.line, 'update');
+      el.setAttribute('line', {end: end});
+      var endPoint = el.getAttribute('line').end;
+      assert.equal(endPoint.z, 9);
+      assert.ok(updateSpy.called);
+    });
+
     test('modify color', function () {
       el.setAttribute('line', {color: '#f9a'});
 


dmarcos:
This fixes the problem but introduces extra object copies when instantiating `oldData`. I'm open to alternative solutions.

ngokevin:
Yeah, need to think of another way. The perf hit isn't worth it (also using a Object.keys/forEach/fn creates an array and function over for...in).

I'd say create a method of the line component to accept updates. Assumingly this would be for `tick`ed changes, we wouldn't want to go thru `.setAttribute` in that case.

dmarcos:
Bare in mind we are already instantiating and copying properties to oldData. Strings, numbers, etc. It is allocated memory that is also eventually garbage collected. This patch makes also copies of plain object attributes. In practice the performance won’t change much.

dmarcos:
The problem is not exclusive to line either but to all the component that have vector properties so a specific method to line will not fix the bug.

src/core/component.js
```diff
@@ -522,7 +522,12 @@ function cloneData (data) {
 */
 function extendProperties (dest, source, isSinglePropSchema) {
   if (isSinglePropSchema && (source === null || typeof source !== 'object')) { return source; }
-  return utils.extend(dest, source);
+  var copy = utils.extend(dest, source);
+  Object.keys(copy).forEach(function copyObject (key) {
```
ngokevin: Use for loop

src/core/component.js
```diff
@@ -522,7 +522,12 @@ function cloneData (data) {
...
+    if (!copy[key] || copy[key].constructor !== Object) { return; }
+    copy[key] = utils.extend({}, copy[key]);
```
ngokevin: utils.clone

fernandojsg:
I'm afraid that with all the latest optimizations to improve performance we could end up losing consistency if things don't work as expected depending on the type of component (eg: As for pos/rot/sca https://github.com/aframevr/aframe-inspector/issues/534).
I honestly would prefer consistency over performance, as you will always have the option to go deep and use the object3D anyway.
Not sure if it could be even needed something similar to `debug=true` to ensure that the `setAttribute/getAttribute` works as you could expect even if you will lose some performance.

ngokevin:
The getAttribute flow often doesn't matter because for most entities, we never need to retrieve the data, only set. For example, it would be rare to want to retrieve the opacity of an animating button. If we set start/end points of a line and want to keep track, then can store those a in vector in component controlling the line.

The pos/rot/sca was a good perf improvement, with a minor fix on Inspector needed because Inspector was depending on componentchanged event, which also isn't a necessary API. I just simply told Inspector to update the panel when the handles are used. Performance is #1 priority IMO, because sites are unusable otherwise.

this commit should be fine, but worth nothing setAttribute has some overhead that should be avoided on frequent operations.

dmarcos:
@fernandojsg If this PR lands, what other inconsistencies are you worried about? If you modify obj/rot/scale via `setAttribute` componentchanged should fire normally as before. It is when you modify the object3D directly, which is an advanced feature is when the event is not fired because you bypass the `setAttribute` code path.

dmarcos:
This should be ready to go.

fernandojsg:
@dmarcos @ngokevin I believe with this change everything should be fine now. Although I'm not sure about how the getAttribute works, I would expect that If I'm modifying the value accessing directly to the `object3D` that won't get me the expected value, but if I'm setting attributes using `setAttribute` it will always return the same value when I'll call `getAttribute` is that correct?
I totally understand that performance is a top priority here but I was afraid to have edge cases because of the performance optimizations that didn't work as you could expect or as they used to work before these optimizations. I believe we must ensure that the whole api regarding `set/get Attributes` is consistent but expose optimizations to the user knowing that they could lead to inconsistent results if you don't know what you're doing.

dmarcos:
@fernandojsg There are no inconsistencies that I'm aware of. If you modify the object3D directly, `getAttribute` / `setAttribute` will work exactly as before. See example below on: https://rawgit.com/aframevr/aframe/master/examples/showcase/anime-UI/
getset

PR #3413, opened 2018-03-02
keyboard-shortcuts adds and removes event listeners in play/pause

Adam4lexander:
Changes keyboard-shortcuts so event listener for key presses is attached in 'play' and removed in 'pause'. 
Linked to https://github.com/aframevr/aframe-inspector/pull/530 - we dont want the 'f' key to enter vr mode while the inspector is open and this component is paused.

### DIFF
--- a/src/components/scene/keyboard-shortcuts.js
+++ b/src/components/scene/keyboard-shortcuts.js
@@ -11,7 +11,7 @@ module.exports.Component = registerComponent('keyboard-shortcuts', {
     var self = this;
     var scene = this.el;
 
-    this.listener = window.addEventListener('keyup', function (event) {
+    this.listener = function (event) {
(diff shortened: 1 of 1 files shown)


src/components/scene/keyboard-shortcuts.js
```diff
@@ -11,23 +11,27 @@ module.exports.Component = registerComponent('keyboard-shortcuts', {
     var self = this;
     var scene = this.el;
 
-    this.listener = window.addEventListener('keyup', function (event) {
+    this.listener = function (event) {
```
dmarcos: Can we make a method `onKeyup` following the same pattern found in the code base? e.g: https://github.com/aframevr/aframe/blob/master/src/components/scene/inspector.js#L50

Adam4lexander:
Sure thing, its done!

PR #3418, opened 2018-03-06
adding look-controls to the example

Utopiah:
**Description:** lot of people get confused on why the camera "doesn't move" once they add it to be able to add the reticle

**Changes proposed:**
- Added look-controls to avoid that.


dmarcos:
Thank you!

PR #3423, opened 2018-03-08
Support Gear VR and Oculus Go controllers with gearvr-controls

machenmusik:
allow multiple ID prefixes with pipe delimiter; 
support GearVR, Gear VR and Oculus Go with gearvr-controls.

Tested with laser-controls example.

### DIFF
--- a/src/components/gearvr-controls.js
+++ b/src/components/gearvr-controls.js
@@ -9,7 +9,7 @@ var GEARVR_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/samsun
 var GEARVR_CONTROLLER_MODEL_OBJ_URL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.obj';
 var GEARVR_CONTROLLER_MODEL_OBJ_MTL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.mtl';
 
-var GAMEPAD_ID_PREFIX = 'Gear VR';
+var GAMEPAD_ID_PREFIX = 'Gear VR|GearVR|Oculus Go';
 
 /**
  * Gear VR controls.
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -88,10 +88,20 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
   var i;
   var matchingControllerOccurence = 0;
   var targetControllerMatch = filterControllerIndex || 0;
-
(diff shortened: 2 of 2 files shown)


machenmusik:
As mentioned, tested with laser-controls example, using Gear VR controller and current Oculus Browser (that has the controller naming quirk and is otherwise broken)

src/components/gearvr-controls.js
```diff
@@ -9,7 +9,7 @@ var GEARVR_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/samsun
 var GEARVR_CONTROLLER_MODEL_OBJ_URL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.obj';
 var GEARVR_CONTROLLER_MODEL_OBJ_MTL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.mtl';
 
-var GAMEPAD_ID_PREFIX = 'Gear VR';
+var GAMEPAD_ID_PREFIX = 'Gear VR|GearVR|Oculus Go';
```
DigiTec: Please remove GearVR as that is a bug in our code that will be fixed next Tuesday in 4.3.1

src/utils/tracked-controls.js
```diff
@@ -88,10 +88,20 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
   var i;
   var matchingControllerOccurence = 0;
   var targetControllerMatch = filterControllerIndex || 0;
-
+  var filterIdPrefixes;
```
DigiTec: This would all be cleaner if you just made a RegExp that matches on the controller.id with a starts with operator "^". This will be jit compiled (faster) and remove a bunch of code.

src/components/gearvr-controls.js
dmarcos: What about making this variable an array instead of parsing each time?

src/utils/tracked-controls.js
dmarcos: That's a good idea. I always think twice about having RegExps if I can avoid them.

machenmusik:
We are actually discussing withdrawing this one, and making `oculus-go-controls` with separate prefix and model, which `gearvr-controls` could include as a dependency.

src/utils/tracked-controls.js
machenmusik: If we go with this approach, probably better to just make all the prefixes into regex and use it as such...

machenmusik:
... since if `GearVR` is going away, we will no longer have a case with two different prefixes for same controller

PR #3424, opened 2018-03-09
[renderer] Add scene component for renderer config.

donmccurdy:
Resolves https://github.com/aframevr/aframe/issues/666.

```html
<a-scene renderer="antialias: true;
                   sortObjects: true;
                   gammaOutput: true;">
```

### DIFF
--- a/src/components/scene/renderer.js
+++ b/src/components/scene/renderer.js
@@ -1,0 +1,81 @@
+/* global HTMLElement */
+var register = require('../../core/component').registerComponent;
+var debug = require('../../utils/debug');
+
+var warn = debug('components:renderer:warn');
+
+/**
+ * Determines state of various renderer properties.
+ * This could be done by a component, but it's important
+ * to know the values when the renderer is created, so
+ * that materials don't need to be recompiled.
+ */
+module.exports.Component = register('renderer', {
+  schema: {
+    antialias: {default: 'auto', oneOf: ['true', 'false', 'auto']},
+    gammaOutput: {default: false},
+    physicallyCorrectLights: {default: false},
+    sortObjects: {default: false}
+  },
+  init: function () {
+    var el = this.el;
+    var data = this.data;
+    var renderer = el.renderer;
+
+    if (!el.isScene) {
+      warn('Fog component can only be applied to <a-scene>');
+    }
+
+    // Default not AA for mobile.
+    // NOTE: This default is also applied in a-scene.js.
+    var attrString = HTMLElement.prototype.getAttribute.call(el, this.name);
+    if (attrString && !attrString.match(/($|\W)antialias\W/)) {
+      el.setAttribute(this.name, {antialias: !el.isMobile});
+    }
+
+    renderer.gammaOutput = data.gammaOutput;
+    renderer.sortObjects = data.sortObjects;
+    renderer.physicallyCorrectLights = data.physicallyCorrectLights;
+  },
+  update: function (prevData) {
+    if (!Object.keys(prevData).length) { return; }
+
+    var data = this.data;
+    var sceneEl = this.el;
+    var renderer = sceneEl.renderer;
+    var needsShaderUpdate = false;
+
+    if (data.antialias !== prevData.antialias) {
+      warn('Property "antialias" cannot be changed after scene initialization');
+    }
+
+    if (data.sortObjects !== prevData.sortObjects) {
+      renderer.sortObjects = data.sortObjects;
+    }
+
+    if (data.gammaOutput !== prevData.gammaOutput) {
+      renderer.gammaOutput = data.gammaOutput;
+      needsShaderUpdate = true;
+    }
+
+    if (data.physicallyCorrectLights !== prevData.physicallyCorrectLights) {
+      renderer.physicallyCorrectLights = data.physicallyCorrectLights;
+      needsShaderUpdate = true;
+    }
+
+    if (needsShaderUpdate) {
+      warn('Modifying renderer properties at runtime requires shader update and may drop frames.');
+      sceneEl.object3D.traverse(function (node) {
+        if (node.isMesh) {
+          if (Array.isArray(node.material)) {
+            node.material.forEach(function (material) {
+              material.needsUpdate = true;
+            });
+          } else {
+            node.material.needsUpdate = true;
+          }
+        }
+      });
+    }
+  }
+});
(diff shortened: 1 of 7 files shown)


src/core/scene/a-scene.js
```diff
@@ -670,22 +672,38 @@ function exitFullscreen () {
...
+  physicallyCorrectLights: 'physically-correct-lights'
+};
+
 /**
- * Determines if renderer anti-aliasing should be enabled.
- * Enabled by default if has native WebVR or is desktop.
+ * Determines state of various renderer properties.
+ * This could be done by a component, but it's important
+ * to know the values when the renderer is created, so
+ * that materials don't need to be recompiled.
```
donmccurdy: Correction: we only need to know `antialias` up front.

ngokevin:
I would try using `renderer="foo: bar"` syntax, but not thru the component flow? Perhaps just parse it with `AFRAME.utils.styleParser.parse`.

donmccurdy:
Would you want to roll antialias into that, or leave it as-is? Other than that a “real” component would be fine.

ngokevin:
Yeah, we can roll anti-alias into that, and deprecate the current `antialias="true"`.

donmccurdy:
Ok, implemented `renderer="antialias: true; gammaOutput: true"` here. Will add docs + test if the gist looks good.

ngokevin:
Looks good. There's no problem with the component setting up too late?

src/components/scene/renderer.js
```diff
@@ -0,0 +1,82 @@
...
+      warn('Fog component can only be applied to <a-scene>');
+    }
+
+    if (el.hasAttribute('antialias')) {
+      warn('Component `antialias` is deprecated. Use `renderer="antialias: true"` instead.');
+    }
+
+    // Default not AA for mobile.
+    // NOTE: This default is also applied in a-scene.js.
+    var attrString = HTMLElement.prototype.getAttribute.call(el, this.name);
```
ngokevin: We've been doing vars-on-top (declarations only). Though tempted to just adopt prettier :P

src/components/scene/renderer.js
```diff
@@ -0,0 +1,82 @@
...
+
+    if (data.physicallyCorrectLights !== prevData.physicallyCorrectLights) {
+      renderer.physicallyCorrectLights = data.physicallyCorrectLights;
+      needsShaderUpdate = true;
+    }
+
+    if (needsShaderUpdate) {
+      warn('Modifying renderer properties at runtime requires shader update and may drop frames.');
+      sceneEl.object3D.traverse(function (node) {
+        if (node.isMesh) {
```
ngokevin: can do an early return and keep an indent `if (!node.isMesh) { return; }`

src/components/scene/renderer.js
```diff
@@ -0,0 +1,82 @@
...
+    if (attrString && !attrString.match(/($|\W)antialias\W/)) {
+      el.setAttribute(this.name, {antialias: !el.isMobile});
+    }
+
+    renderer.gammaOutput = data.gammaOutput;
+    renderer.sortObjects = data.sortObjects;
+    renderer.physicallyCorrectLights = data.physicallyCorrectLights;
+  },
+  update: function (prevData) {
+    if (!Object.keys(prevData).length) { return; }
```
ngokevin: I guess this check isn't even really needed?

src/components/scene/renderer.js
```diff
@@ -0,0 +1,82 @@
...
+
+    if (el.hasAttribute('antialias')) {
+      warn('Component `antialias` is deprecated. Use `renderer="antialias: true"` instead.');
+    }
+
+    // Default not AA for mobile.
+    // NOTE: This default is also applied in a-scene.js.
+    var attrString = HTMLElement.prototype.getAttribute.call(el, this.name);
+    if (attrString && !attrString.match(/($|\W)antialias\W/)) {
+      el.setAttribute(this.name, {antialias: !el.isMobile});
```
ngokevin: Can do `setAttribute('renderer', 'antialias', `!el.isMobile`);`

Don't need `this.name` above either, can just be `renderer`

src/components/scene/renderer.js
donmccurdy: Done

src/components/scene/renderer.js
donmccurdy: Done

src/components/scene/renderer.js
donmccurdy: It's needed, `update()` is called during initialization and we don't want to set `.needsUpdate=true` unnecessarily.

src/components/scene/renderer.js
donmccurdy: Done

dmarcos:
Default components are gone on https://github.com/aframevr/aframe/pull/3490. We need to adapt this PR

donmccurdy:
> There's no problem with the component setting up too late?

Well, the `renderer.antialias` handling is a bit of a hack because `init()` runs too late. Otherwise it's fine, we can set the other properties before any shaders have compiled.

Having trouble doing a `git push` from the cafe, so the last commit is still missing and my "Done" comments were premature.

@dmarcos — without default components, the only issue I see is that I can't detect whether the `renderer` component has been added before or after scene initialization. `sceneEl.hasLoaded` seems to be true when `init()` runs either way. I don't want to set `.needsUpdate` on all of the materials if the renderer hasn't compiled them yet... is there another way to see if the first render has already happened?

donmccurdy:
Resolved the initialization issue by checking `sceneEl.time > 0`. Added tests and docs. I think this is ready to go except one thing — are we OK deprecating `antialias="true"` on an `0.8.X` release? I've kept it functional here but the warning is still logged.

dmarcos:
Besides the warning messages. Should we reflect in the docs that antialias cannot be changed after initialization?

donmccurdy:
Done (and rebased).

dmarcos:
Better not to introduce API changes on a minor version. We can deprecate `<a-scene antialias="true">` later

dmarcos:
Thank you!

donmccurdy:
> Better not to introduce API changes on a minor version. We can deprecate <a-scene antialias="true"> later

Ok, note this PR does log a warning about `antialias` (but keeps it working). We'll want to remove that before 0.8.2 then?

PR #3425, opened 2018-03-09
fix cursor mouseenter/mouseleave issues on overlapping entities, do not grab new intersections after cleared in cursor component

ngokevin:
**Description:**

There was some old code in cursor component lying around that when its intersected entity is cleared, it will look for the next one. 

This causes issues. It's better to just rely on the incoming raycaster-intersection/raycaster-intersection-cleared events rather than trying to manage intersections in cursor component.

**Changes proposed:**
- Don't set new intersection after intersection clear in cursor component.
- Reuse event detail in raycaster.
- Remove old vars.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -59,7 +59,7 @@ module.exports.Component = registerComponent('cursor', {
     // Debounce.
     this.updateCanvasBounds = utils.debounce(function updateCanvasBounds () {
       self.canvasBounds = self.el.sceneEl.canvas.getBoundingClientRect();
-    }, 200);
+    }, 1000);
 
     this.eventDetail = {};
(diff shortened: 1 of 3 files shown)


src/components/cursor.js
```diff
@@ -59,7 +59,7 @@ module.exports.Component = registerComponent('cursor', {
     // Debounce.
     this.updateCanvasBounds = utils.debounce(function updateCanvasBounds () {
       self.canvasBounds = self.el.sceneEl.canvas.getBoundingClientRect();
-    }, 200);
+    }, 1000);
```
dmarcos: Is this necessary?

ngokevin: Not really, but really don't need to check window resize that often.

PR #3429, opened 2018-03-09
tidy up loose ends from #3423 (fixes #3447 and #3483)

### LINKED ISSUE #3447: 0.8.0 laser-controls with Oculus Touch
When using `laser-controls` with Oculus Touch controllers the following effects are present:

- Only the left controller is ever tracked (sometimes cases where it isn't tracked at all).
- When both controllers are added, the right controller is attached to the left controller.
- Supplying the controller with any custom `raycaster` properties stops the controller from being tracked, it just sits in its default position in the scene.
- Occasionally the GearVR controller will be loaded in instead.

Here's a demo: https://codepen.io/PButcher/pen/GxoOBQ

As far as I can tell the `oculus-touch-controls` component is working just fine in 0.8.0.

I experienced this earlier when testing for #3445 and also when poking around for solutions to this `super-hands` issue: https://github.com/wmurphyrd/aframe-super-hands-component/issues/128

- A-Frame Version: 0.8.0
- Platform / Device: Win10, Rift, Firefox
- Reproducible Code Snippet or URL: https://codepen.io/PButcher/pen/GxoOBQ

machenmusik:
as discussed in #3423, 
use RegExp and update multi-prefix syntax; remove GearVR
(might be best to merge *after* the Oculus Browser fix is deployed next Tuesday)

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -88,19 +88,15 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
   var i;
   var matchingControllerOccurence = 0;
   var targetControllerMatch = filterControllerIndex || 0;
-  var filterIdPrefixes;
+  var filterIdRegex;
   if (filterIdPrefix && filterIdPrefix.indexOf('|') >= 0) {
-    filterIdPrefixes = filterIdPrefix.split('|');
+    filterIdRegex = new RegExp('^' + filterIdPrefix);
   }
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     // Determine if the controller ID matches our criteria
-    if (filterIdPrefixes) {
-      var matches = false;
-      for (var prefix in filterIdPrefixes) {
-        if (prefix && controller.id.indexOf(prefix) === -1) { matches = true; }
-      }
-      if (!matches) { continue; }
+    if (filterIdRegex) {
+      if (!filterIdRegex.test(controller.id)) { continue; }
     } else
     if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) { continue; }
     if (!filterIdPrefix && controller.id !== filterIdExact) { continue; }
--- a/src/components/gearvr-controls.js
+++ b/src/components/gearvr-controls.js
@@ -9,7 +9,7 @@ var GEARVR_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/samsun
 var GEARVR_CONTROLLER_MODEL_OBJ_URL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.obj';
 var GEARVR_CONTROLLER_MODEL_OBJ_MTL = GEARVR_CONTROLLER_MODEL_BASE_URL + 'gear_vr_controller.mtl';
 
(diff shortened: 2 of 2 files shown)


machenmusik:
@digitec can you confirm whether the Oculus Browser fix has been deployed yet?

vincentfretin:
From #gearvr slack channel today by 
```
DigiTec [3 h 51]
Update, Oculus Browser 4.3.1 is now rolling out and has fixed the GearVR -> Gear VR naming on the controllers. It’s at 10% and on the PTC (public test channel) and will ramp up to 100% between now and Thursday assuming no new issues are uncovered.

If you aren’t familiar with the PTC it is a channel you can get on to get Oculus software a bit earlier (usually a few days in advance) and test it and provide feedback before we launch to public.
```

machenmusik:
built version is at https://rawgit.com/chenzlabs/aframe/fix-oculus-gearvr-controller-dist/dist/aframe-master.min.js for testing convenience

ngokevin:
Let's just keep all the strings and merge it.

src/utils/tracked-controls.js
```diff
@@ -88,19 +88,15 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
   var i;
   var matchingControllerOccurence = 0;
   var targetControllerMatch = filterControllerIndex || 0;
-  var filterIdPrefixes;
+  var filterIdRegex;
   if (filterIdPrefix && filterIdPrefix.indexOf('|') >= 0) {
-    filterIdPrefixes = filterIdPrefix.split('|');
+    filterIdRegex = new RegExp('^' + filterIdPrefix);
```
dmarcos: You don't need to create a Regexp for each time we ant to find and match controllers. Can we instantiate and reuse?

machenmusik: I considered that, but for example https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/compile says it's deprecated so didn't seem great.  Thoughts?

machenmusik: My understanding is that because it is passed to tracked-controls, we can't simply make GAMEPAD_ID_PREFIX a RegExp (which is what both @DigiTec and I would ideally prefer) as that will not carry over properly.  As a slightly clunky workaround to prevent subsequent construction, we can probably create a global map that holds the RegExp created for a given prefix string... Is that the sort of thing you had in mind here?

dmarcos: Can we make filterIdPrefix parameter to take a String or an Array an avoid using RegExps that are always hard to understand and maintain?

machenmusik: The prefix needs to be passed into tracked-controls, not sure that would work as string-or-array.  The original version used split to make the array if it saw pipe delimiter, which also makes garbage, but is easier to read and maintain; would you prefer a different PR that doesn't use RegExp at all, but instead provides an alternative solution to the related issues? (i.e. simple bugfix to previous pipe delimited split)

dmarcos: I would prefer to avoid RegExp if we can. When I saw `var GAMEPAD_ID_PREFIX = '((Gear VR)|(Oculus Go))'; ` it made me feel confused right off the bat :)

ngokevin: regex is fine for me. it's the fastest way. probably just rename the var is good.

machenmusik:
too late @ngokevin -- see #3489

PR #3434, opened 2018-03-11
[look-controls] Fixes pointerlock jank, primitive shorthand.

donmccurdy:


dmarcos:
Thanks!

donmccurdy:
Ah, ran look-controls tests but not the integration test... will fix that, i think the test just needs to be changed.

PR #3440, opened 2018-03-11
use scene.onAfterRender for tock

wizgrav:
Aframe 0.8.0 switched to using the webvr support provided by three's webglrenderer. The problem is that the renderer submits the frame as part of its render() operations so by the time the tock methods are called the frame is already submitted. This breaks post processing.

To facilitate this case, this PR was submitted in three https://github.com/mrdoob/three.js/pull/11582 . This PR utilizes this functionality to fix post proc


src/index.js
```diff
@@ -83,7 +83,7 @@ require('./core/a-mixin');
 require('./extras/components/');
 require('./extras/primitives/');
 
-console.log('A-Frame Version: 0.8.1 (Date 2018-03-11, Commit #ea836cf)');
+console.log('A-Frame Version: 0.8.0 (Date 2018-03-11, Commit #f6761393)');
```
dmarcos: We don’t need this change

src/core/scene/a-scene.js
```diff
@@ -50,10 +50,16 @@ module.exports.AScene = registerElement('a-scene', {
         this.isMobile = isMobile;
         this.isScene = true;
         this.object3D = new THREE.Scene();
+        this.vrCamera = null;
+        var self = this;
+        this.object3D.onAfterRender = function (renderer, scene, camera) {
+          self.vrCamera = camera;
```
dmarcos: What is this variable for!

dmarcos:
Thanks for staying on top of postprocessing and make sure it still works.

src/core/scene/a-scene.js
wizgrav: the issue with three's internal vr handling was that it was swapping the provided camera internally, with an ArrayCamera for the twin views in the case of VR. This camera was not exposed but needed for post processing and tiled forward rendering. onBeforeRender and onAfterRender where added to address that, the correct camera will be provided in every case.

So .vrCamera will be the same as .camera in the case of non-vr rendering and the ArrayCamera otherwise so it can be used in view dependent post proc filters like god rays.

The naming was completely arbirtrary, if you prefer something else just say so

dmarcos:
Rebase needed

src/core/scene/a-scene.js
dmarcos: I see. I comment about what the variable is for will be hepful. We can also think about a better name: `cameraUsedToRender` , `rendererInternalCamera` ... not sure. Something more descriptive

wizgrav: renderCamera?

wizgrav:
Variable name changed, console.log fixed

src/core/scene/a-scene.js
```diff
@@ -50,10 +50,17 @@ module.exports.AScene = registerElement('a-scene', {
         this.isMobile = isMobile;
         this.isScene = true;
         this.object3D = new THREE.Scene();
+        this.renderUsedCamera = null;
+        var self = this;
+        this.object3D.onAfterRender = function (renderer, scene, camera) {
+          // THREE may swap the camera used for the rendering if in VR, so expose it
+          self.renderUsedCamera = camera;
```
dmarcos: Should we pass the camera as a third argument of the tock call? It might be cleaner

dmarcos:
@wizgrav This is almost ready. Since the main use of the tock method is for post processing it makes sense to pass the camera as a paremeter. What do you think?

wizgrav:
camera is now passed as a third argument to tock which does look much more cleaner

wizgrav:
Fscked up rebase closing this to submit a clean PR

PR #3451, opened 2018-03-14
Update camera.md, mention camera rig

donmccurdy:


docs/components/camera.md
```diff
@@ -11,14 +11,27 @@ The camera component defines from which perspective the user views the scene.
 The camera is commonly paired with controls components that allow input devices
 to move and rotate the camera.
 
-## Example
+## Examples
 
 A camera situated at the average height of human eye level (1.6 meters).
+When used with controls that receive rotation or position (e.g. from a
+VR device) this position will be overridden.
```
dmarcos: But also restored when exiting VR. Should we mention that?

ngokevin:
Looks good as is, just point is that controls overwrite position so don't rely much on setting them on camera directly.

PR #3456, opened 2018-03-15
default camera position changed

payamspot:
0 0 0 changed to 0 1.6 0


dmarcos:
Thank you. Good catch.

payamspot:
Thanks :)

PR #3464, opened 2018-03-16
2 changes

payamspot:
2 minor changes
**Changes proposed:**
- default camera height changed to 1.6
- table cells format corrected


dmarcos:
Thanks

PR #3468, opened 2018-03-17
use scene.onAfterRender() for tock

wizgrav:
Correct handling of post processing when using three's internal webvr support


wizgrav:
@dmarcos I redid the PR cause the rebase didn't went well

dmarcos:
I merge and I'll add docs

PR #3472, opened 2018-03-18
Fix for consistent firing of buttonup/down/touchstart/end/buttonchanged

DigiTec:
This change delinks these event streams so that they can fire independenty of one another. I've also added a sufficiently complete number of tests for the new behaior to make sure all events that should be firing are, looking for negative events (ensuring some events don't fire when they should not) and that the values all come through as expected.

As part of the fix I've supplied some helpers that are now consistently used in tracked-controls.test.js for verifying individual events and streams of events in a more concise way to avoid duplicating a lot of code.

I've also made consistent the use of sinon's spy asserts, which were previously using the ".called" property which reaches into the abstraction further than it should. This also allows us to upgrade certain spy verification's to look for a specific number of calls (calledOnce) which is an overall improvement.

This change cascaded into the oculus-touch-controls.test.js and gearvr-controls.test.js as well. For the oculus-touch-controls, this is mostly assert fixups, but also does some additional deep object verification.

For the gearvr-controls.test.js this included some bug fixes (using wrong event names) when I was trying to make the test fail and could not, only to realize it wasn't sniffing for the right event names. This is in preparation for creation of the oculus-go-controls component and test suite.

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -51,7 +51,9 @@ module.exports.Component = registerComponent('tracked-controls', {
   tick: function (time, delta) {
     var mesh = this.el.getObject3D('mesh');
     // Update mesh animations.
-    if (mesh && mesh.update) { mesh.update(delta / 1000); }
+    if (mesh && mesh.update) {
+      mesh.update(delta / 1000);
+    }
     this.updateGamepad();
     this.updatePose();
     this.updateButtons();
@@ -87,6 +89,12 @@ module.exports.Component = registerComponent('tracked-controls', {
     this.controller = controller;
   },
 
+  /**
+   * Applies an artificial arm model to simulate elbow to wrist positioning
+   * based on the orientation of the controller.
+   *
+   * @param {object} controllerPosition - Existing vector to update with controller position.
+   */
   applyArmModel: function (controllerPosition) {
     // Use controllerPosition and deltaControllerPosition to avoid creating variables.
     var controller = this.controller;
@@ -156,7 +164,9 @@ module.exports.Component = registerComponent('tracked-controls', {
       object3D.position.fromArray(pose.position);
     } else {
       // Controller not 6DOF, apply arm model.
-      if (this.data.armModel) { this.applyArmModel(object3D.position); }
+      if (this.data.armModel) {
+        this.applyArmModel(object3D.position);
+      }
     }
 
     if (pose.orientation) {
@@ -210,8 +220,8 @@ module.exports.Component = registerComponent('tracked-controls', {
    * @returns {boolean} Whether button has changed in any way.
    */
   handleButton: function (id, buttonState) {
-    var changed = this.handlePress(id, buttonState) ||
-                  this.handleTouch(id, buttonState) ||
+    var changed = this.handlePress(id, buttonState) |
+                  this.handleTouch(id, buttonState) |
                   this.handleValue(id, buttonState);
     if (!changed) { return false; }
     this.el.emit('buttonchanged', {id: id, state: buttonState});
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -164,9 +164,7 @@ module.exports.Component = registerComponent('tracked-controls', {
       object3D.position.fromArray(pose.position);
     } else {
       // Controller not 6DOF, apply arm model.
-      if (this.data.armModel) {
-        this.applyArmModel(object3D.position);
-      }
+      if (this.data.armModel) { this.applyArmModel(object3D.position); }
     }
 
     if (pose.orientation) {
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -51,9 +51,7 @@ module.exports.Component = registerComponent('tracked-controls', {
   tick: function (time, delta) {
     var mesh = this.el.getObject3D('mesh');
     // Update mesh animations.
-    if (mesh && mesh.update) {
-      mesh.update(delta / 1000);
-    }
+    if (mesh && mesh.update) { mesh.update(delta / 1000); }
     this.updateGamepad();
     this.updatePose();
     this.updateButtons();
--- a/tests/components/gearvr-controls.test.js
+++ b/tests/components/gearvr-controls.test.js
@@ -1,4 +1,4 @@
-/* global assert, process, setup, suite, test */
+/* global assert, process, setup, sinon, suite, test */
 var entityFactory = require('../helpers').entityFactory;
 
 suite('gearvr-controls', function () {
@@ -27,8 +27,8 @@ suite('gearvr-controls', function () {
     test('returns not present if no controllers on first on first call', function () {
       var el = this.el;
       var component = el.components['gearvr-controls'];
-      var addEventListenersSpy = this.sinon.spy(component, 'addEventListeners');
-      var injectTrackedControlsSpy = this.sinon.spy(component, 'injectTrackedControls');
+      var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
+      var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
 
       el.sceneEl.systems['tracked-controls'].controllers = [];
 
@@ -36,17 +36,17 @@ suite('gearvr-controls', function () {
 
       component.checkIfControllerPresent();
 
-      this.sinon.assert.notCalled(injectTrackedControlsSpy);
-      this.sinon.assert.notCalled(addEventListenersSpy);
+      sinon.assert.notCalled(injectTrackedControlsSpy);
+      sinon.assert.notCalled(addEventListenersSpy);
       assert.strictEqual(component.controllerPresent, false);
     });
 
     test('does not remove event listeners if no controllers', function () {
       var el = this.el;
       var component = el.components['gearvr-controls'];
-      var addEventListenersSpy = this.sinon.spy(component, 'addEventListeners');
-      var injectTrackedControlsSpy = this.sinon.spy(component, 'injectTrackedControls');
-      var removeEventListenersSpy = this.sinon.spy(component, 'removeEventListeners');
+      var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
+      var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
+      var removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
 
       el.sceneEl.systems['tracked-controls'].controllers = [];
 
@@ -55,18 +55,18 @@ suite('gearvr-controls', function () {
 
       component.checkIfControllerPresent();
 
-      this.sinon.assert.notCalled(injectTrackedControlsSpy);
-      this.sinon.assert.notCalled(addEventListenersSpy);
-      this.sinon.assert.notCalled(removeEventListenersSpy);
+      sinon.assert.notCalled(injectTrackedControlsSpy);
+      sinon.assert.notCalled(addEventListenersSpy);
+      sinon.assert.notCalled(removeEventListenersSpy);
       assert.strictEqual(component.controllerPresent, false);
     });
 
     test('attaches events if controller is newly present', function () {
       var el = this.el;
       var component = el.components['gearvr-controls'];
-      var addEventListenersSpy = this.sinon.spy(component, 'addEventListeners');
-      var injectTrackedControlsSpy = this.sinon.spy(component, 'injectTrackedControls');
-      var removeEventListenersSpy = this.sinon.spy(component, 'removeEventListeners');
+      var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
+      var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
+      var removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
 
       el.sceneEl.systems['tracked-controls'].controllers = component.controllersWhenPresent;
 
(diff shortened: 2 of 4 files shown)


src/components/tracked-controls.js
```diff
@@ -51,7 +51,9 @@ module.exports.Component = registerComponent('tracked-controls', {
   tick: function (time, delta) {
     var mesh = this.el.getObject3D('mesh');
     // Update mesh animations.
-    if (mesh && mesh.update) { mesh.update(delta / 1000); }
+    if (mesh && mesh.update) {
```
DigiTec: Updated for consistency. It seems that the pattern for a complicated line is to break and not do it all on one line.

src/components/tracked-controls.js
```diff
@@ -210,8 +220,8 @@ module.exports.Component = registerComponent('tracked-controls', {
    * @returns {boolean} Whether button has changed in any way.
    */
   handleButton: function (id, buttonState) {
-    var changed = this.handlePress(id, buttonState) ||
-                  this.handleTouch(id, buttonState) ||
+    var changed = this.handlePress(id, buttonState) |
```
DigiTec: This is the core bug fix. We no longer do the short circuit operator. Instead we process all three handle* calls so that all events get generated and prior button states get saved correctly.

tests/components/gearvr-controls.test.js
```diff
@@ -149,8 +151,8 @@ suite('gearvr-controls', function () {
       component.checkIfControllerPresent();
 
       // Fail purposely.
-      el.addEventListener('thumbstickmoved', function (evt) {
-        assert.ok(false);
+      el.addEventListener('trackpadmoved', function (evt) {
```
DigiTec: This was previously incorrect. Likely since it was copied from the oculus-touch-controls.test.js I think. Once I changed it I was able to verify that I could indeed hit the fail when I updated the changed value accordingly.

tests/components/gearvr-controls.test.js
```diff
@@ -206,7 +214,14 @@ suite('gearvr-controls', function () {
       var trackedControls = el.components['tracked-controls'];
       var applyArmModelSpy = this.sinon.spy(trackedControls, 'applyArmModel');
       trackedControls.tick();
-      assert.notOk(applyArmModelSpy.called);
+
+      // Verify that the function which applies arm model is not called when disabled.
+      this.sinon.assert.notCalled(applyArmModelSpy);
+
+      // Additionally verify that no other offets have been applied.
+      assert.strictEqual(el.object3D.position.x, 0);
```
DigiTec: Added additional verification for the arm model applying.

tests/components/gearvr-controls.test.js
```diff
@@ -216,7 +231,29 @@ suite('gearvr-controls', function () {
       var trackedControls = el.components['tracked-controls'];
       var applyArmModelSpy = this.sinon.spy(trackedControls, 'applyArmModel');
       trackedControls.tick();
-      assert.ok(applyArmModelSpy.called);
+
+      // Verify that the function which applies arm model is called.
+      this.sinon.assert.calledOnce(applyArmModelSpy);
+    });
+
+    test('verifies armModel position is applied for the right hand', function () {
```
DigiTec: Verify right handed arm model works correctly (positive X offset)

tests/components/gearvr-controls.test.js
```diff
@@ -216,7 +231,29 @@ suite('gearvr-controls', function () {
...
+      makePresent(el);
+      var trackedControls = el.components['tracked-controls'];
+      trackedControls.tick();
+      assert.ok(el.object3D.position.x > 0);
+    });
+
+    test('verifies armModel position is applied for the left hand', function () {
+      var el = this.el;
+      el.setAttribute('gearvr-controls', 'armModel', true);
+      el.setAttribute('gearvr-controls', 'hand', 'left');
```
DigiTec: Verify left handed arm model works correctly (negative X offset)

tests/components/oculus-touch-controls.test.js
```diff
@@ -28,9 +28,9 @@ suite('oculus-touch-controls', function () {
 
       component.checkIfControllerPresent();
 
-      assert.notOk(injectTrackedControlsSpy.called);
-      assert.notOk(addEventListenersSpy.called);
-      assert.ok(removeEventListenersSpy.called);
+      this.sinon.assert.notCalled(injectTrackedControlsSpy);
```
DigiTec: There is a series of these updates. We shouldn't be breaking the sinon abstraction on the spy and accessing called directly if we can avoid it. Instead use the helpers for notCalled, called, calledOnce, etc...

tests/components/oculus-touch-controls.test.js
```diff
@@ -83,10 +83,10 @@ suite('oculus-touch-controls', function () {
...
       component.checkIfControllerPresent();
 
-      assert.notOk(injectTrackedControlsSpy.called);
-      assert.notOk(addEventListenersSpy.called);
-      assert.notOk(removeEventListenersSpy.called);
-      assert.ok(component.controllerPresent);
+      this.sinon.assert.notCalled(injectTrackedControlsSpy);
+      this.sinon.assert.notCalled(addEventListenersSpy);
+      this.sinon.assert.notCalled(removeEventListenersSpy);
+      assert.strictEqual(component.controllerPresent, true);
```
DigiTec: More consistently use strictEqual

tests/components/tracked-controls.test.js
```diff
@@ -258,10 +257,7 @@ suite('tracked-controls', function () {
         call => call.args[0] === 'buttonchanged');
       assert.equal(emitChangedCalls.length, 1);
 
-      const emitCall = emitChangedCalls[0];
-      assert.equal(emitCall.args[0], 'buttonchanged');
-      assert.deepEqual(emitCall.args[1].id, 0);
-      assert.deepEqual(emitCall.args[1].state, controller.buttons[0]);
+      assertButtonEvent(emitChangedCalls[0], 'buttonchanged', 0, controller.buttons[0]);
```
DigiTec: Helper is at the bottom. Basically avoids us writing a 4 lines of assert code and consolidates into a useful single call.

tests/components/tracked-controls.test.js
```diff
@@ -273,10 +269,129 @@ suite('tracked-controls', function () {
...
+    });
+
+    test('emits buttonchanged if value changed', function () {
```
DigiTec: New test for 'value' property.

tests/components/tracked-controls.test.js
```diff
@@ -273,10 +269,129 @@ suite('tracked-controls', function () {
...
+      component.tick();
+
+      const emitChangedCalls = emitSpy.getCalls().filter(
+        call => call.args[0] === 'buttonchanged');
+      assert.equal(emitChangedCalls.length, 1);
+
+      assertButtonEvent(emitChangedCalls[0], 'buttonchanged', 0, controller.buttons[0]);
+    });
+
+    test('emits independent streams for buttondown, touchstart and buttonchanged', function () {
```
DigiTec: New test to verify we get separate event streams

tests/components/tracked-controls.test.js
```diff
@@ -273,10 +269,129 @@ suite('tracked-controls', function () {
...
+      const emitChangedCalls = emitSpy.getCalls().filter(
+        call => call.args[0] === 'buttonchanged');
+      assert.equal(emitChangedCalls.length, 1);
+
+      assertButtonEvent(emitButtonCalls[0], 'buttondown', 0, controller.buttons[0]);
+      assertButtonEvent(emitTouchCalls[0], 'touchstart', 0, controller.buttons[0]);
+      assertButtonEvent(emitChangedCalls[0], 'buttonchanged', 0, controller.buttons[0]);
+    });
+
+    test('emits independent streams for buttonup, touchend and buttonchanged', function () {
```
DigiTec: Same for if we release the button, we should get all 3 event series.

tests/components/tracked-controls.test.js
```diff
@@ -273,10 +269,129 @@ suite('tracked-controls', function () {
...
+      const emitChangedCalls = emitSpy.getCalls().filter(call => call.args[0] === 'buttonchanged');
+      assert.equal(emitChangedCalls.length, 1);
+
+      // Verify each of the 3 events has the correct event name and state.
+      assertButtonEvent(emitButtonCalls[0], 'buttonup', 0, controller.buttons[0]);
+      assertButtonEvent(emitTouchCalls[0], 'touchend', 0, controller.buttons[0]);
+      assertButtonEvent(emitChangedCalls[0], 'buttonchanged', 0, controller.buttons[0]);
+    });
+
+    test('emits correct event stream for spaced out interaction.', function () {
```
DigiTec: Verifies that multi-tick interactions produce the correct series of events for a simple press and release.

tests/components/tracked-controls.test.js
```diff
@@ -273,10 +269,129 @@ suite('tracked-controls', function () {
...
+      // Fourth round, verify we only see a touchend and button changed when we lift our touch.
+      emitSpy = this.sinon.spy(el, 'emit');
+      controller.buttons[0].touched = false;
+      component.tick();
+
+      assertEventStream(emitSpy.getCalls(), ['buttonchanged', 'touchend'], ['buttondown', 'buttonup', 'touchstart']);
+      emitSpy.restore();
+    });
+
+    test('emits correct event states on a fast click', function () {
```
DigiTec: Verifies that if the states all change in single ticks that we still get proper event streams. Special case of the prior test that also tests that we have independent event streams.

tests/components/tracked-controls.test.js
```diff
@@ -465,3 +568,43 @@ function toQuaternion (x, y, z) {
     return quaternion.toArray();
   })();
 }
+
+/**
```
DigiTec: New helpers so we can consolidate some existing code.

DigiTec:
Test failure is FireFox only. Unrelated to the changes I've made unless I'm missing something. Maybe this is a known issue?

SUMMARY:
✔ 11 tests completed
✖ 1 test failed
FAILED TESTS:
  ✖ "after each" hook for "emits mousedown event on intersectedEl"
    Firefox 59.0.0 (Linux 0.0.0)
  Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  require<["/home/travis/build/aframevr/aframe/tests/__init.test.js"]</</<@/tmp/tests/__init.test.js:57:0 <- /tmp/f633f031b631506f1dd8c8e8c60c723e.browserify:64579:5

DigiTec:
Verified that the offending test failure is also on master (updated my local repository master to verify).

src/components/tracked-controls.js
dmarcos: Can we keep this as a one liner? In modern displays aspect ratio vertical space is at a premium.

src/components/tracked-controls.js
```diff
@@ -156,7 +164,9 @@ module.exports.Component = registerComponent('tracked-controls', {
       object3D.position.fromArray(pose.position);
     } else {
       // Controller not 6DOF, apply arm model.
-      if (this.data.armModel) { this.applyArmModel(object3D.position); }
+      if (this.data.armModel) {
+        this.applyArmModel(object3D.position);
```
dmarcos: Can we keep this as a one liner?

tests/components/gearvr-controls.test.js
dmarcos: good catch 👍

tests/components/tracked-controls.test.js
```diff
@@ -465,3 +568,43 @@ function toQuaternion (x, y, z) {
...
+ * data matches.
+ *
+ * @param {object} eventCall - The spied call to emit.
+ * @param {string} eventName - The expected name of the event.
+ * @param {number} eventId - The button id firing the event.
+ * @param {object} eventState - The full event state.
+ */
+function assertButtonEvent (eventCall, eventName, eventId, eventState) {
+  assert.equal(eventCall.args[0], eventName);
+  assert.deepEqual(eventCall.args[1].id, eventId);
```
dmarcos: Should this just be `assert.equal(eventCall.args[1].id, eventId);` ?

DigiTec: yes.

tests/components/tracked-controls.test.js
dmarcos: cool

src/components/tracked-controls.js
DigiTec: Sure, what is the rationale for 1 line versus line breaking? there seems to be no formatter or consistency across the code base with regards to this. The only consistency I did find was that most complex statements (like this one) are broken across lines and return clauses within the statement block tend to be on the same line.

src/components/tracked-controls.js
dmarcos: I think this is the last comment to address and this is ready to go.

dmarcos:
@DigiTec Thanks for the fix and the test coverage. Much appreciated 👍

src/components/tracked-controls.js
dmarcos: You're right that the rule is a bit loose. If the line is short we keep it in one line. If you have a very long statement we break it in multiple lines. We don't have a specific number of characters. That's the best answer I can give 😄

dmarcos:
Thank you!

tests/components/gearvr-controls.test.js
```diff
@@ -204,19 +212,48 @@ suite('gearvr-controls', function () {
...
       var trackedControls = el.components['tracked-controls'];
-      var applyArmModelSpy = this.sinon.spy(trackedControls, 'applyArmModel');
+      var applyArmModelSpy = sinon.spy(trackedControls, 'applyArmModel');
       trackedControls.tick();
-      assert.notOk(applyArmModelSpy.called);
+
+      // Verify that the function which applies arm model is not called when disabled.
+      sinon.assert.notCalled(applyArmModelSpy);
+
+      // Additionally verify that no other offets have been applied.
```
vincentfretin: offets => offsets

PR #3473, opened 2018-03-19
Add additional button tests for DayDream controls.

DigiTec:
The DayDream controls were missing menu and system button tests to make sure that they get remapped and reemitted properly given their remapped names.

Additionally was able to fix up the trackpadchanged to use the deep object check for event state and then copy this across the two new tests.

Found one test that had a leading space in the description, fixed and then changed one assert.ok -> assert.fail which is more appropriate given the context.

### DIFF
--- a/tests/components/daydream-controls.test.js
+++ b/tests/components/daydream-controls.test.js
@@ -126,14 +126,14 @@ suite('daydream-controls', function () {
       component.checkIfControllerPresent();
 
       // Install event handler listening for thumbstickmoved.
-      this.el.addEventListener('trackpadmoved', function (evt) {
+      el.addEventListener('trackpadmoved', function (evt) {
         assert.equal(evt.detail.x, 0.1);
         assert.equal(evt.detail.y, 0.2);
         done();
       });
 
       // Emit axismove.
-      this.el.emit('axismove', {axis: [0.1, 0.2], changed: [true, false]});
+      el.emit('axismove', {axis: [0.1, 0.2], changed: [true, false]});
     });
 
     test('does not emit trackpadmove on axismove with no changes', function (done) {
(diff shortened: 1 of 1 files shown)


tests/components/daydream-controls.test.js
```diff
@@ -158,23 +157,64 @@ suite('daydream-controls', function () {
...
+      var component = el.components['daydream-controls'];
+
+      el.sceneEl.systems['tracked-controls'].controllers = component.controllersWhenPresent;
+
+      component.checkIfControllerPresent();
+
+      const eventState = {value: 0.5, pressed: true, touched: true};
+
+      // Install event handler listening for triggerchanged.
+      this.el.addEventListener('menuchanged', function (evt) {
```
dmarcos: You can just do `el.addEventListener('menuchanged', function (evt) {`

tests/components/daydream-controls.test.js
```diff
@@ -158,23 +157,64 @@ suite('daydream-controls', function () {
...
+        assert.deepEqual(evt.detail, eventState);
+        done();
+      });
+
+      // Emit buttonchanged for the menu button.
+      this.el.emit('buttonchanged', {id: 1, state: eventState});
```
dmarcos: `el.emit('buttonchanged', {id: 1, state: eventState});`

tests/components/daydream-controls.test.js
```diff
@@ -158,23 +157,64 @@ suite('daydream-controls', function () {
...
+      var component = el.components['daydream-controls'];
+
+      el.sceneEl.systems['tracked-controls'].controllers = component.controllersWhenPresent;
+
+      component.checkIfControllerPresent();
+
+      const eventState = {value: 0.5, pressed: true, touched: true};
+
+      // Install event handler listening for triggerchanged.
+      this.el.addEventListener('systemchanged', function (evt) {
```
dmarcos: this not needed `el.addEventListener('systemchanged', function (evt) {`

tests/components/daydream-controls.test.js
```diff
@@ -158,23 +157,64 @@ suite('daydream-controls', function () {
...
+        assert.deepEqual(evt.detail, eventState);
         done();
       });
 
-      // Emit buttonchanged.
-      this.el.emit('buttonchanged',
-                   {id: 0, state: {value: 0.5, pressed: true, touched: true}});
+      // Emit buttonchanged for the system button.
+      this.el.emit('buttonchanged', {id: 2, state: eventState});
```
dmarcos: You can do `el.emit('buttonchanged', {id: 2, state: eventState});`

tests/components/daydream-controls.test.js
DigiTec: Nice, I copied that from another place, so I fixed up all instances.

dmarcos:
Thank you! 🎉

Issue #3445, opened 2018-03-13
0.8.0 Raycaster Issue: Multiple intersected interactive elements along ray path


PButcher:
I've been sitting on this problem for a couple of hours now and can't put my finger on what exactly is going on under the hood to cause the issue but the symptoms are easy to reproduce:

https://codepen.io/PButcher/pen/PRZGRr

Move the mouse cursor down over the blue plane (which sits behind the red plane), on to the red plane, then move the cursor left across the face of the red plane until the raycaster is not intersecting with the blue plane, then move the cursor back to the right. The raycaster emits a `mouseleave` and `mouseenter` event on the red plane as well as a `mouseenter` and `raycaster-intersected` event on the blue plane even though it is not the closest intersection. Furthermore, when you move the cursor upwards over the blue again, the raycaster emits a `raycaster-intersected-cleared` event but not a `mouseleave` event on the red plane.

Here's a terrible GIF of one of my apps that I was upgrading to 0.8.0 from 0.7.x that alerted me to the bug: https://streamable.com/ntaum

I know a lot of work went into optimising the raycaster for 0.8.0, but it looks like some logic was removed that stops events being emitted on new intersections that are further along the ray while there are currently other foreground intersections.

Cheers!

- A-Frame Version: 0.8.0
- Platform / Device: First noticed on Win10, Rift, Firefox. Debugged on Mac with Chrome 64.
- Reproducible Code Snippet or URL: https://codepen.io/PButcher/pen/PRZGRr

donmccurdy:
Thanks! I think this is the same issue as https://github.com/aframevr/aframe/issues/3467, which should be fixed after https://github.com/aframevr/aframe/pull/3475.

PR #3475, opened 2018-03-19
[cursor] Use recent intersection data from raycaster.

### LINKED ISSUE #3467: rayOrigin: mouse click event has wrong intersection location
**Description:**

- A-Frame Version: 0.8.0
- Platform / Device: Firefox or Chrome on desktop
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/lace-coast

The example code should be placing objects so that they appear to be at the location clicked.

The intersection location given seems to have something to do with the location of the last object to have a `mouseleave` event, as if you set `raycaster: "objects: a-sky"` then the location never changes.

donmccurdy:
* Pulls most recent intersection info from raycaster, removes `cursor.intersection` property.
* Remove redundant `updateMatrixWorld()` call.
* Ensure `clearCurrentIntersection()` does not select a new intersection entity if used in a context where another intersection is already known.

Fixes #3467, #3485, #3445.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -52,7 +52,6 @@ module.exports.Component = registerComponent('cursor', {
 
     this.fuseTimeout = undefined;
     this.cursorDownEl = null;
-    this.intersection = null;
     this.intersectedEl = null;
     this.canvasBounds = document.body.getBoundingClientRect();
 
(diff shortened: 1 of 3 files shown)


ngokevin:
Looking, just gonna give it a test.

donmccurdy:
thanks!

PR #3481, opened 2018-03-21
Audio file in guide should link to media on CDN

anselanza:
**Description:**
In the *Building a 360° Image Gallery* Guide, the example for the "basic scene" has a path to an audio file that is relative. Trying to run this locally will not work unless the user finds the file, downloads it and copies it to the correct path. Not very friendly for first-time users.

**Changes proposed:**
- Use src URL for media on CDN, as per the Glitch example.


dmarcos:
Thanks! Congrats for your first patch!

Issue #3483, opened 2018-03-22
Gear VR Controller does not work.


sabbir420:
**Description:**
I am facing problem. I am using gear VR and I use 'gearvr-controls' but it does not work. It does not click. If I put cursor then by cursor is clicking but gear vr controller does not.  Here is an example. You can check from [here](https://odd-fortnight.glitch.me).

- A-Frame Version: 0.8.0
- Platform / Device: Gear VR
- Reproducible Code Snippet or URL:

machenmusik:
this should be fixed by https://github.com/aframevr/aframe/pull/3429
see https://codepen.io/anon/pen/JLNRvx

sabbir420:
@machenmusik  See my example. Clicking does not work. I checked your example too.

machenmusik:
I don't think clicking does anything in the codepen I mentioned above (there are no handlers attached to the box entity that would do anything with clicks).

With respect to your previous glitch 
https://glitch.com/edit/#!/odd-fortnight?path=index.html:1:0
a few things:
- gearvr-controls is now built into A-Frame, don't include aframe-gearvr-controls-component anymore
- due to #3483, you need to use an A-Frame version that has #3429 built into it
- seems as though trackpad is acting as click with GearVR, rather than trigger

see this remix, which works with GearVR controller in Oculus Browser (and also Touch controllers in Firefox)
https://glitch.com/edit/#!/watery-teeth?path=index.html:56:0

machenmusik:
note that in watery-teeth: 
- old component was removed
- patched A-Frame version is referenced instead of broken 0.8.0
- used laser-controls (which automatically includes cursors for controllers) instead of gearvr-controls
- removed explicit head-locked cursor (on the assumption that you'd want the controller cursor that laser-controls provides)

machenmusik:
This should be fixed now in latest A-Frame master via #3489

PR #3489, opened 2018-03-23
simple fix for Gear VR and Oculus Touch controller issues (fixes #3447 and #3483)

### LINKED ISSUE #3447: 0.8.0 laser-controls with Oculus Touch

machenmusik:
Per discussion on #3429, this is a simpler fix avoiding RegExp usage.

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -97,12 +97,16 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
     // Determine if the controller ID matches our criteria
     if (filterIdPrefixes) {
       var matches = false;
-      for (var prefix in filterIdPrefixes) {
-        if (prefix && controller.id.indexOf(prefix) === -1) { matches = true; }
+      for (var j = 0; j < filterIdPrefixes.length; j++) {
+        var prefix = filterIdPrefixes[j];
+        if (prefix && !controller.id.indexOf(prefix)) {
+          matches = true;
+          break;
+        }
       }
       if (!matches) { continue; }
     } else
-    if (filterIdPrefix && controller.id.indexOf(filterIdPrefix) === -1) { continue; }
+    if (filterIdPrefix && controller.id.indexOf(filterIdPrefix)) { continue; }
     if (!filterIdPrefix && controller.id !== filterIdExact) { continue; }
 
     // If the hand filter and controller handedness are defined we compare them.
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -1,6 +1,8 @@
+var split = require('./split').split;
+
 var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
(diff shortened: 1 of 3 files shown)


src/utils/tracked-controls.js
```diff
@@ -90,19 +101,24 @@ function findMatchingController (controllers, filterIdExact, filterIdPrefix, fil
...
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     // Determine if the controller ID matches our criteria
     if (filterIdPrefixes) {
       var matches = false;
-      for (var prefix in filterIdPrefixes) {
-        if (prefix && controller.id.indexOf(prefix) === -1) { matches = true; }
+      for (var j = 0; j < filterIdPrefixes.length; j++) {
+        var prefix = filterIdPrefixes[j];
+        console.log(prefix, ' ?= ', controller.id);
```
dmarcos: don't need log

src/utils/tracked-controls.js
```diff
@@ -2,6 +2,17 @@ var DEFAULT_HANDEDNESS = require('../constants').DEFAULT_HANDEDNESS;
 var AXIS_LABELS = ['x', 'y', 'z', 'w'];
 var NUM_HANDS = 2; // Number of hands in a pair. Should always be 2.
 
+var splitOnPipeMap = {};
+
+function splitOnPipeOnce (stringToSplit) {
+  var rtn = splitOnPipeMap[stringToSplit];
+  if (!rtn) {
```
dmarcos: Early return `if (rtn) { return; }`

machenmusik: (I'm not sure it actually looks better now, but I understand your stylistic preference.)

machenmusik:
The prefix also needs to be passed into tracked-controls, not sure that will survive as array...?

machenmusik:
If we don't mind doing the split each time, can remove the split once map stuff and it will look much more like what is in there now, just not broken.  Is that preferable?

machenmusik:
Have "put the candle back" to provide a minimal fix to what is already in master.
Hopefully we can agree to merge the simple fix, and subsequently discuss optimizations.

machenmusik:
Travis CI failure, but seems entitely unrelated

PR #3490, opened 2018-03-23
remove default components

ngokevin:
**Description:**

Save 4 component initializations per entity.

Allows for setting with object3D API on creation without the component overriding with 0/0/0:

```js
var entity = document.createElement('a-entity');
entity.object3D.position.set(1, 2, 3);
scene.appendChild(entity);
```

**Changes proposed:**
- Save 4 component initializations per entity.
- Don't need pos/rot/scale/vis components initializations by default if not defined.
- Removes a bunch of code in the component lifecycle that checks for default component

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -22,15 +22,6 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
  * @member {boolean} isPlaying - false if dynamic behavior of the entity is paused.
  */
 var proto = Object.create(ANode.prototype, {
-  defaultComponents: {
-    value: {
-      position: '',
-      rotation: '',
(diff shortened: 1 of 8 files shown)


src/core/a-entity.js
```diff
@@ -767,6 +727,7 @@ var proto = Object.create(ANode.prototype, {
       // If cached value exists, return partial component data.
       var component = this.components[attr];
       if (component) { return component.attrValue; }
+      // if (OBJECT3D_PROPERTIES.indexOf(attr) !== -1) { return this.getAttribute(attr); }
```
dmarcos: remove

ngokevin: k

PR #3492, opened 2018-03-23
Update webvr-polyfill to 0.10.5

jsantell:
This fixes an issue from late December where objects close to the camera appear duplicated/out of focus (https://github.com/immersive-web/webvr-polyfill/issues/312)


madebynoam:
Woo hoo! Thank you!

PR #3497, opened 2018-03-24
Update material.md - default metalness was incorrect

omgitsraven:
As seen in https://github.com/aframevr/aframe/blob/master/src/shaders/standard.js , the default value for metalness is 0, not 0.5


donmccurdy:
Thanks!

PR #3511, opened 2018-03-31
Fix Touch pose (fixes #2965)

### LINKED ISSUE #2965: oculus-touch-controls seems to be off real controller by 45 degrees
I have a basic a-frame scene with oculus-touch-controls (same thing with laser-controls too). Its off the real angle of my controller by almost exactly 45 degrees. I'm not applying rotation to the controller or any of the parent entities, I've stripped my a-frame to just controller entitties and still have this issue.

oculus-error

The blue 'cone' is showing the correct level "straight ahead" vector. Both the oculus controls and the laser ray are tipped 45 degrees relative to reality.

machenmusik:
Modify oculus-touch-controls to accommodate current model and semi-recent browser / SDK behavioral changes in Touch controller pose.  See discussion in #2965.

Note that the model pose changes are made conditionally to the default Touch model only, and that the positional offset are not exactly perfect still, but IMO slightly better aligned with real-life than previous suggestions.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -2,6 +2,7 @@ var bind = require('../utils/bind');
 var registerComponent = require('../core/component').registerComponent;
 var trackedControlsUtils = require('../utils/tracked-controls');
 var onButtonEvent = trackedControlsUtils.onButtonEvent;
+var THREE = require('../lib/three');
 
 var TOUCH_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/oculus/oculus-touch-controller-';
 var TOUCH_CONTROLLER_MODEL_OBJ_URL_L = TOUCH_CONTROLLER_MODEL_BASE_URL + 'left.obj';
@@ -11,7 +12,7 @@ var TOUCH_CONTROLLER_MODEL_OBJ_MTL_R = TOUCH_CONTROLLER_MODEL_BASE_URL + 'right.
 
 var GAMEPAD_ID_PREFIX = 'Oculus Touch';
 
-var PIVOT_OFFSET = {x: 0, y: -0.015, z: 0.04};
+var PIVOT_OFFSET = {x: 0, y: 0, z: 0};
 
 /**
  * Oculus Touch controls.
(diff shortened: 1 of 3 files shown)


machenmusik:
(Could use some additional opinions on the pose offsets...)

src/components/oculus-touch-controls.js
```diff
@@ -11,7 +12,7 @@ var TOUCH_CONTROLLER_MODEL_OBJ_MTL_R = TOUCH_CONTROLLER_MODEL_BASE_URL + 'right.
 
 var GAMEPAD_ID_PREFIX = 'Oculus Touch';
 
-var PIVOT_OFFSET = {x: 0, y: -0.015, z: 0.04};
+var PIVOT_OFFSET = {x: 0, y: 0, z: 0};
```
dmarcos: We don't this anymore. Can we remove?

src/components/oculus-touch-controls.js
```diff
@@ -191,6 +192,18 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
     // Offset pivot point
     controllerObject3D.position = PIVOT_OFFSET;
```
dmarcos: We can probably remove this line since it does not have any effect

src/components/oculus-touch-controls.js
```diff
@@ -191,6 +192,18 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+
+    // Fix issue #2965 for default Touch model, rotate it 45 degrees.
```
dmarcos: No need to reference issues on source comments. The reference in the commit is good enough

src/components/oculus-touch-controls.js
```diff
@@ -191,6 +192,18 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    // But if it isn't the default Touch model, don't.
+    controllerObject3D.traverse(function (o3d) {
+      if (o3d instanceof THREE.Mesh) {
+        if (o3d.name.startsWith('body_oculus-touch-controller-')) {
```
dmarcos: It this check necessary. We're returning early above if `this.data.model === false`  (https://github.com/aframevr/aframe/pull/3511/files#diff-3056bccc97756be52ace666f5e213434R180)

src/components/oculus-touch-controls.js
```diff
@@ -191,6 +192,18 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
     // Offset pivot point
     controllerObject3D.position = PIVOT_OFFSET;
+
+    // Fix issue #2965 for default Touch model, rotate it 45 degrees.
+    // But if it isn't the default Touch model, don't.
+    controllerObject3D.traverse(function (o3d) {
```
dmarcos: Avoid acronyms. `o3d` -> `object3D`

machenmusik:
I previously left the pivot offset in there in case others tried it and determined that pivot offset of zero was wrong - I am interpreting the comments as if you've tried it and it looks fine, thus removed that stuff now.

Not sure whether we need the traversal, rather than just manipulating controllerObject3D; if we could just manipulate controllerObject3D, then I might argue you'd want to put the pivot offset code back but use the values in the onModelLoaded adjustment.

I'm not in a position to check at the moment (may not happen this weekend)

ngokevin:
Thanks!

Issue #3505, opened 2018-03-28
Sound component `name` and `id` not present in event details for `sound-loaded` and `sound-ended`


dsinni:
**Description:** When applying multiple sound components to an entity (e.g., `sound__click`), there doesn't seem to be an obvious way (or possibly any way) to identify the sound component's `name` and/or `id` (e.g., `sound__click` and `click`, respectively).

This makes listening for the `sound-loaded` and `sound-ended` events somewhat ineffective as the source component and audio are unknown.

- A-Frame Version: `0.7.*` - `0.8.0`
- Platform / Device: Firefox, Chrome, Edge on Windows 10 (all I've tested)
- Reproducible Code Snippet or URL: https://codepen.io/dansinni/pen/ZxvyLx

koganei:
@dsinni What do you think of this PR? Instead of looking for ```e.target.id```, you would look for ```e.detail.id```, and it will give you the id of the sound being loaded/ended, even if there are multiple sounds on the a-box.

dsinni:
Sounds good, @koganei. I only mentioned `e.target` because I couldn't figure for the life of me how to access that data. `e.detail` definitely seems more appropriate. Thanks.

PR #3514, opened 2018-04-01
fix for issue #3505: add name and id to sound events

koganei:
**Description:**
#3505 proposes to add the details of the sound being loaded 

**Changes proposed:**
- Added id, name and component to sound-loaded event.
- Added id, name, component and sound to sound-ended event.

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -76,7 +76,7 @@ module.exports.Component = registerComponent('sound', {
         // Remove this key from cache, otherwise we can't play it again
         THREE.Cache.remove(data.src);
         if (self.data.autoplay || self.mustPlay) { self.playSound(); }
-        self.el.emit('sound-loaded', {id: self.id, name: self.name});
+        self.el.emit('sound-loaded', self.evtDetail);
       });
     }
   },
@@ -156,7 +156,7 @@ module.exports.Component = registerComponent('sound', {
     this.pool.children.forEach(function (sound) {
       sound.onEnded = function () {
         sound.isPlaying = false;
-        el.emit('sound-ended', {index: i, id: self.id, name: self.name});
+        el.emit('sound-ended', self.evtDetail);
       };
     });
   },
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -76,7 +76,7 @@ module.exports.Component = registerComponent('sound', {
         // Remove this key from cache, otherwise we can't play it again
         THREE.Cache.remove(data.src);
         if (self.data.autoplay || self.mustPlay) { self.playSound(); }
-        self.el.emit('sound-loaded', {component: self});
+        self.el.emit('sound-loaded', {component: self, id: self.evtDetail.id, name: self.evtDetail.name});
       });
     }
   },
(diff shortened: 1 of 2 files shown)


koganei:
Hi all, this is my first PR to this repository. Please let me know if anything needs to be changed! The issue only mentions wanting 'id' and 'name', but I added 'component' and 'sound' to be as complete about the event details as possible.

koganei:
@dmarcos good point. I've removed them and now only 'id' and 'name' are emitted.

koganei:
Not sure what's happening, with the same code, the first Travis build failed for Firefox, the second build failed for Chrome. Different errors. :/

src/components/sound.js
```diff
@@ -155,7 +156,7 @@ module.exports.Component = registerComponent('sound', {
     this.pool.children.forEach(function (sound) {
       sound.onEnded = function () {
         sound.isPlaying = false;
-        el.emit('sound-ended', {index: i});
+        el.emit('sound-ended', {index: i, id: self.evtDetail && self.evtDetail.id, name: self.evtDetail && self.evtDetail.name});
```
dmarcos: Where is `evtDetail` defined?

koganei: It's an attribute that all components have. https://github.com/aframevr/aframe/blob/master/src/core/component.js#L38

koganei:
@dmarcos Looking into how evtDetail is built, I found that I could directly use self.id and self.name, so I re-wrote it in a cleaner way.

src/components/sound.js
```diff
@@ -155,7 +156,7 @@ module.exports.Component = registerComponent('sound', {
     this.pool.children.forEach(function (sound) {
       sound.onEnded = function () {
         sound.isPlaying = false;
-        el.emit('sound-ended', {index: i});
+        el.emit('sound-ended', {index: i, id: self.id, name: self.name});
```
koganei: I believe that the index sent here is actually the wrong value. It's always going to be the length of the for loop above it (Line 150), rather than by the forEach that this line is nested in (Line 156). I think that the i referenced here is meant to be the second argument of the forEach, but I don't know enough about the code to know if I'm right or not.

ngokevin: Is the index really necessary? I think it's sufficient to send id/name (create the evtDetail object in `init` once). For sound pools, this event will get fired a lot, and this will create a bunch of objects.

Actually don't think this event in general is really needed?

koganei: Ah, that makes sense. I'll replace them with `evtDetail` then. The event seems to be wanted by the reporter. Sounds like the sort of events that would be useful to chain sounds one after the other, no?

ngokevin: I think that's a rare case, and if dev wants, they can wire the events up themselves. But sound pooling is used when you have lots of sounds that fire rapidly and overlapping each other.

Dev can either reach and grab the pool, or pass in an audio element, or keep track of the sound triggers elsewhere, etc.

koganei: Do you want a new PR that removes the event then?

ngokevin: It's probably okay right now. At the moment, I've been making my own sound pools not using the sound component. It's fine for now

PR #3527, opened 2018-04-07
vr-mode-ui: click event stopped from propagation

DougReeder:
The trigger on a Cardboard V2 causes a screen touch, so it's common to listen for an click on the scene element or window.  This changes prevents a tap on the a-enter-vr button from propagating to the scene and window, so the other listener doesn't have to disentangle an "enter vr" tap from the Cardboard button.


ngokevin:
thanks!

PR #3529, opened 2018-04-07
apply rotation component, if present, in look-controls

machenmusik:
Per discussion on Slack:

In version 0.8.x, `look-controls` was changed to ignore the `rotation` component and directly override `object3d.rotation` -- which appears to have broken some use cases.

This PR modifies `updateOrientation()` in `look-controls` to also take into account the value of the `rotation` component, if present.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -187,13 +187,27 @@ module.exports.Component = registerComponent('look-controls', {
     var pitchObject = this.pitchObject;
     var yawObject = this.yawObject;
     var sceneEl = this.el.sceneEl;
+    var q = this.hmdQuaternion;
 
     // In VR mode, THREE is in charge of updating the camera rotation.
     if (sceneEl.is('vr-mode') && sceneEl.checkHeadsetConnected()) { return; }
 
     // Calculate polyfilled HMD quaternion.
     this.polyfillControls.update();
-    hmdEuler.setFromQuaternion(this.polyfillObject.quaternion, 'YXZ');
+
+    // If rotation component exists, get its values and apply them.
+    var rotationComponent = el.components.rotation;
+    if (rotationComponent) {
+      hmdEuler.set(
+        THREE.Math.DEG2RAD * rotationComponent.data.x,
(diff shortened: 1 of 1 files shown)


src/components/look-controls.js
```diff
@@ -187,13 +187,27 @@ module.exports.Component = registerComponent('look-controls', {
...
 
     // In VR mode, THREE is in charge of updating the camera rotation.
     if (sceneEl.is('vr-mode') && sceneEl.checkHeadsetConnected()) { return; }
 
     // Calculate polyfilled HMD quaternion.
     this.polyfillControls.update();
-    hmdEuler.setFromQuaternion(this.polyfillObject.quaternion, 'YXZ');
+
+    // If rotation component exists, get its values and apply them.
+    var rotationComponent = el.components.rotation;
```
dmarcos: There's always a rotation component

dmarcos:
This breaks consistency because the approach won't work in VR mode when the pose of the camera is modified internally by the VRManager in THREE.

src/components/look-controls.js
machenmusik: I thought that thanks to @ngokevin, that is no longer true?

machenmusik:
@ngokevin is that a theoretical concern, or a finding from using supermedium?  Agreed obviously that if it doesn't work in VR then an alternative implementation is needed

dmarcos:
@machenmusik Any position and rotation in the entity will be overriden by THREE in VR mode.

src/components/look-controls.js
ngokevin: Yup, this check is necessary now. Really don't need to go thru rotation component at all, just modify object3D directly.

ngokevin:
@machenmusik It's just a fact, I would ask three.js to provide an offset on the camera. I think given that, it seems reasonable to close. Thanks for digging into this! @dmarcos spent *a lot* of time on this, we know the pain :)

machenmusik:
Ok makes sense to close until we can get accommodation in Three

PR #3537, opened 2018-04-12
Oculus touch model position

dmarcos:
The position offset was not applied correctly for the oculus touch model. The orientation did not match the real world position of the controller. Introduced a `orientationOffset` attribute to be able to match any controller model with real world orientation. rotationOffset should be deprecated but I left it for backwards compatiblity. We can remove in 0.9.0

### DIFF
--- a/docs/components/daydream-controls.md
+++ b/docs/components/daydream-controls.md
@@ -36,15 +36,17 @@ It can occassionally take a few seconds before the controller can be used.
 
 ## Value
 
-| Property             | Description                                        | Default |
-|----------------------|----------------------------------------------------|---------|
-| armModel             | Whether the arm model is used for positional data. | true    |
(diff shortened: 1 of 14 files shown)


machenmusik:
Originally orientation offset was for vive-tracker, did we want to put that back in now?

dmarcos:
`rotationOffset` was introduced for the vive tracker and not used anywhere else. It is safe to remove in favor orientationOffset

machenmusik:
Wait, I thought rotation offset was for hand-controls - if you changed hand-controls to use orientationOffset should be fine though

dmarcos:
@machenmusik `hand-controls` uses `orientationOffset` now https://github.com/aframevr/aframe/pull/3537/files#diff-02099d98da3e9b5683f5a6ac5092aec4L173

Issue #3502, opened 2018-03-26
Material blending property


feiss:
Threejs' Material class has the `blending` property which is really useful for faking materials that emit light or other effects. (For example, see the disco lights in [saturday night](https://aframe.io/a-saturday-night))

https://threejs.org/docs/#api/materials/Material.blending

It would be nice to have it available in A-Frame material directly, without having to make a custom component just for changing its value.

samrae7:
I would like to have a crack at this. As far as I can see it just means creating a `blending` component that passes the property to three.js in the sme way that e.g. the position component works. Does that make sense? Thanks

donmccurdy:
Hi @samrae7, please feel free! My feeling is that it should probably be a sub-property of the `material` component, like:

```html
<a-entity material="src: foo.png; blending: additive" />
```

where values are `none`, `normal`, `additive`, `subtractive`, `multiply`. I think it is OK to leave out three.js's "custom" blending modes for now.

@feiss does that sound reasonable for most use cases?

feiss:
@donmccurdy yup, that is exactly what I was asking for :)

samrae7:
@donmccurdy @feiss thanks. I'm working on this. Seems straightforward so will be done soon.

feiss:
thanks!

ngokevin:
https://github.com/aframevr/aframe/pull/3543

PR #3543, opened 2018-04-16
Material blending

samrae7:
**Description:**
Add `blending` property to Aframe Material component
https://github.com/aframevr/aframe/issues/3502

**Changes proposed:**
- Add blending prop to Material component
- Add test

====

I think I've done this right. There's one thing I didn't understand though. If anyone could clarify it would help my understanding of the code.

As i understand it these lines are checking if certain props change, and if they do, we set `needsUpdate` to true
https://github.com/aframevr/aframe/blob/master/src/components/material.js#L127

As explained in the threejs docs this is only needed for certain properties: (https://github.com/mrdoob/three.js/blob/25f4534a5609f55a9d13cff47b1e535b4b109659/docs/manual/introduction/How-to-update-things.html#L175)
 and blending is not one of these.

My question is that I noticed (when writing the tests) that `material.needsUpdate` was true after setting the material.blending attribute. Why is this? I thought it would only be true if we manually set it when we want to trigger a re-render in webGL because e.g. vertex-colors changed.

Thanks

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -30,7 +30,8 @@ module.exports.Component = registerComponent('material', {
     side: {default: 'front', oneOf: ['front', 'back', 'double']},
     transparent: {default: false},
     vertexColors: {type: 'string', default: 'none', oneOf: ['face', 'vertex']},
-    visible: {default: true}
+    visible: {default: true},
+    blending: {default: 'normal', oneOf: ['none', 'normal', 'additive', 'subtractive', 'multiply']}
   },
 
   init: function () {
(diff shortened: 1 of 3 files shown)


ngokevin:
Cool! Would you mind adding to docs as well? `docs/components/material.md` in the properties table.

samrae7:
Ah yes! Sure will do

src/components/material.js
```diff
@@ -121,6 +122,7 @@ module.exports.Component = registerComponent('material', {
     material.transparent = data.transparent !== false || data.opacity < 1.0;
     material.vertexColors = parseVertexColors(data.vertexColors);
     material.visible = data.visible;
+    material.blending = parseBlending(data.blending);
 
     // Check if material needs update.
     if (Object.keys(oldData).length &&
```
ngokevin: If you want another easy task, we should remove this Object.keys call to save an array creation. Instead use a for...in loop to check for key existence or something.

ngokevin:
Thanks!

donmccurdy:
This will allow some nice effects, thanks! 🎉

src/components/material.js
ngokevin: That works. You don't even need the `if (key)` check I believe.

Issue #3509, opened 2018-03-30
Best practices for color space management


donmccurdy:
A-Frame 0.8.0 brings in some upstream fixes to THREE.GLTFLoader that are technically correct, but potentially inconsistent with existing scenes and examples. Pre-0.8.0, `gltf-model` assumed all textures were in linear color space — `material.color` and `material.src` still make this same assumption today. That's incorrect — images are usually sRGB, and the glTF spec explicitly requires color and emissive textures to use sRGB.

Following best practice:

1. Textures are converted `sRGB` -> `linear` at the beginning of the fragment shader if the image encoding is set to sRGB.
2. Renderer calculations are done in linear space.
3. Renderer output is converted `linear` -> `sRGB` for the screen if `renderer.gammaOutput=true`.

By default in A-Frame and three.js, `renderer.gammaOutput` is _false_ and step (3) never happens. As a result, glTF models look darker than they should. Textures using `material.src` look mostly right, because we don't set the encoding to sRGB, even though the images usually are sRGB. So the colors are being treated as linear in lighting calculations, which practically isn't very noticeable.

User **danilo** in Slack created this PDF, which illustrates the problem nicely:

https://drive.google.com/file/d/1BRdCU9nDzUHOfDYYXdbAV1cMW60Md6uv/view

I don't have any quick answers on what we should do here... in short, I think the three.js defaults assume some familiarity with color spaces, and ideally we would abstract this away from our users, or at least provide some best practices as a reference for users who are going to care about matching a color palette precisely.

If we _were_ to change our defaults, here's a straw proposal:

* [ ] By default, set `renderer="gammaOutput: true"`.
* [ ] By default, set `src` and `emissiveMap` both to sRGB encoding. Perhaps provide `srcColorSpace` and `emissiveColorSpace` to override this, but I think those properties would be an escape hatch to restore legacy behavior and not otherwise useful.

The one thing this doesn't fix is `material.color` values, which are assumed to be already linear in three.js, and don't have an encoding setting. So either (1) we could provide guidance about converting an sRGB hex code to linear, or (2) we would automatically do this in the `material` component.

I'm also not sure what this all means for vertex colors...

My vote for first step would be to add `srcColorSpace` and `emissiveColorSpace` in a PR, verify that it improves outcomes, and then enable them without changing any defaults. It would be nice to have feedback from someone more familiar with color spaces and/or rendering pipelines. 🙂

webdan76:
[...]

PR #3547, opened 2018-04-20
add support for setting the encoding of the main map

ncoder:
**Description:**
Allow specifying the encoding of the main map on the material. This is important for correct linear/gamma settings.

Looks like:
` <a-entity id="sky" 
            geometry="primitive:sphere; radius:30; phiLength:360; phiStart:0; thetaLength:180" 
            material="src:#skymap; encoding:sRGB"
            rotation="30 0 0">
`
 
**Changes proposed:**
I don't think this change is merge able as-is. But I wanted to get this out there to see if there was interest in doing something similar, or if there are plans to solve this some other way.

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -30,7 +30,8 @@ module.exports.Component = registerComponent('material', {
     side: {default: 'front', oneOf: ['front', 'back', 'double']},
     transparent: {default: false},
     vertexColors: {type: 'string', default: 'none', oneOf: ['face', 'vertex']},
-    visible: {default: true}
+    visible: {default: true},
+    encoding: {type: 'string', oneOf: ['sRGB', 'Linear']}
   },
 
   init: function () {
--- a/src/systems/material.js
+++ b/src/systems/material.js
@@ -346,6 +346,16 @@ function setTextureProperties (texture, data) {
   if (offset.x !== 0 || offset.y !== 0) {
     texture.offset.set(offset.x, offset.y);
   }
+
+  if( data.encoding) {
+    let newEncoding = THREE[ data.encoding] || THREE[ data.encoding + "Encoding"];
+    if(newEncoding === undefined) {
+      error("Unrecognized encoding: " +  data.encoding);
+    }
+    else {
+      texture.encoding = newEncoding;
+    }
+  }
 }
 
 /**
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -39,7 +39,7 @@ module.exports.updateMapMaterialFromData = function (materialName, dataName, sha
     // Load texture for the new material src.
     // (And check if we should still use it once available in callback.)
     el.sceneEl.systems.material.loadTexture(src,
-      {src: src, repeat: data.repeat, offset: data.offset, npot: data.npot},
+      {src: src, repeat: data.repeat, offset: data.offset, npot: data.npot, encoding: data.encoding},
       checkSetMap);
   }
 


src/systems/material.js
```diff
@@ -346,6 +346,16 @@ function setTextureProperties (texture, data) {
...
+
+  if( data.encoding) {
+    let newEncoding = THREE[ data.encoding] || THREE[ data.encoding + "Encoding"];
+    if(newEncoding === undefined) {
+      error("Unrecognized encoding: " +  data.encoding);
+    }
+    else {
+      texture.encoding = newEncoding;
+    }
+  }
```
donmccurdy: Will need to use ES5 `var` and match the if/else style of the code above here.

donmccurdy:
+1 from me for providing a way of setting sRGB here. It should apply to `map` and `emissiveMap` but not other textures. See also: https://github.com/aframevr/aframe/issues/3509. The alternative to this approach would be setting sRGB on `map` and `emissiveMap` _by default_, as well as `gammaOutput=true`, but that would make a lot of breaking color changes for 0.9.0. That sounds like too much pain, IMO.

Despite the three.js terminology, I would suggest naming this option `colorSpace` rather than `encoding`, and using `sRGB` and `linear` as the values.

src/components/material.js
```diff
@@ -30,7 +30,8 @@ module.exports.Component = registerComponent('material', {
     side: {default: 'front', oneOf: ['front', 'back', 'double']},
     transparent: {default: false},
     vertexColors: {type: 'string', default: 'none', oneOf: ['face', 'vertex']},
-    visible: {default: true}
+    visible: {default: true},
+    encoding: {type: 'string', oneOf: ['sRGB', 'Linear']}
```
dmarcos: It needs a default value

src/systems/material.js
```diff
@@ -346,6 +346,16 @@ function setTextureProperties (texture, data) {
   if (offset.x !== 0 || offset.y !== 0) {
     texture.offset.set(offset.x, offset.y);
   }
+
+  if( data.encoding) {
+    let newEncoding = THREE[ data.encoding] || THREE[ data.encoding + "Encoding"];
+    if(newEncoding === undefined) {
+      error("Unrecognized encoding: " +  data.encoding);
```
dmarcos: Probably better a warning than an error.

dmarcos: I would do. warning + fallback to existing encoding if it's not valid

src/systems/material.js
```diff
@@ -346,6 +346,16 @@ function setTextureProperties (texture, data) {
   if (offset.x !== 0 || offset.y !== 0) {
     texture.offset.set(offset.x, offset.y);
   }
+
+  if( data.encoding) {
+    let newEncoding = THREE[ data.encoding] || THREE[ data.encoding + "Encoding"];
```
dmarcos: `var encoding` define at top of function

dmarcos:
Thank you. @ncode congrats for your first PR. A few changes and this is ready to go.

ngokevin:
Hey @ncode, want to make a few changes? Thanks!

ngokevin:
@donmccurdy does this approach still look good? i don't have much experience with colorspace and encodings.

donmccurdy:
This is still correct, but I think i'd prefer to do something at the scene level (like https://github.com/aframevr/aframe/pull/3757) rather than putting a setting on every component in the scene.

ncoder:
Sweet! Thanks, guys.

Issue #3390, opened 2018-02-17
Oculus Go controllers do not work in A-Frame


frankolivier:
A-Frame controller experiences do not work on Oculus Browser 4.3.0+ on an Oculus Go device.

Previously, the Oculus Go controllers identified themselves as GearVR controllers. However, they are now properly identified as Go controllers, and so now all of the A-Frame controller experiences no longer work; the A-Frame controller code only works on a known set of named controllers. 

In Oculus Browser 4.3.0 and newer - on Oculus Go devices only - the Gamepad.id property now returns "Oculus Go Controller". 

Before Oculus Browser 4.3.0, GamePad.id always returned "Gear VR Controller" on Oculus Go.

On Gear VR devices, Gamepad.id will still be "Gear VR Controller".

Oculus Go controllers have the same functionality as Gear VR controllers:
https://www.oculus.com/blog/pioneering-the-frontier-of-vr-introducing-oculus-go-plus-santa-cruz-updates/

- A-Frame Version: 0.7.1
- Platform / Device: Oculus Browser on Oculus Go
- Reproducible Code Snippet or URL: See https://aframe.io/examples/showcase/a-blast/ - the various A-Frame 'universal' controls components (such as super-hands: https://github.com/wmurphyrd/aframe-super-hands-component) also do not work.

mkeblx:
If possible it would be nice to remove looking at these strings. Also in WebXR API they won't be available.

machenmusik:
I would counter with: when the strings are available, it is much more intuitive for the user to have a representation that matches what they would see from their system outside Web context (e.g. to avoid "hey why does my browser think my Touch controller looks like that" as well as "which button means what in this Web thing") - perhaps you are arguing for a generic controller representation, but we basically started with that in A-Frame and quickly evolved to understanding specific controllers instead.  Still being debated in WebXR API discussions from what I see

dmarcos:
I do not have Oculus Go to test. How is the Oculus Go controller. @frankolivier Does it have the same buttons than the regular GearVR remote? @mkeblx What do you propose to use instead of the string to discriminate between controllers and properly map buttons to actions?

machenmusik:
Should be same (or else they would not have been able to function properly as Gear VR Controller)

We should probably keep this issue limited to current Gamepad API, and discuss WebXR future in https://github.com/immersive-web/webxr/issues/319

Send me one and I will fix however we decide ;-)

Artyom17:
[...]

PR #3548, opened 2018-04-20
add Oculus Go controller support

dmliao:
**Description:** Adds support for using / displaying the Oculus Go controller. Addresses #3390.

**Changes proposed:**
- Adds a new component and tests for `oculus-go-controls`, which is pretty close in functionality to the GearVR controller. A lot of this is duplicated code right now, which can likely be consolidated.
- Sets up the Oculus Go controller model from the CDN to support button highlighting.
- Removes workaround from `tracked-controls` to have multiple gamepad id prefixes in one string (since we now have a separate component for Oculus Go controllers)


src/components/oculus-go-controls.js
```diff
@@ -0,0 +1,179 @@
...
+
+  onModelLoaded: function (evt) {
+    var controllerObject3D = evt.detail.model;
+    var buttonMeshes;
+    if (!this.data.model) { return; }
+    buttonMeshes = this.buttonMeshes = {};
+    buttonMeshes.trigger = controllerObject3D.children[0].children[1].children[4];
+    buttonMeshes.trackpad = controllerObject3D.children[0].children[1].children[1];
+
+    // we need to set trigger and trackpad material separately so that we can edit them independently
```
dmliao: I'm not too happy with how I fiddle with the materials here, but without these changes, changing the color doesn't seem to actually create any visual difference. Is this a concern, or is it something I shouldn't be worried about?

Tests/components/oculus-go-controls.test.js
```diff
@@ -0,0 +1,263 @@
+/* global assert, process, setup, sinon, suite, test */
+var entityFactory = require('../helpers').entityFactory;
```
dmarcos: the path for this file should be `tests/components/oculus-go-controls.test.js` with lowercase `tests` instead of `Tests`

src/components/oculus-go-controls.js
```diff
@@ -0,0 +1,179 @@
...
+  // No need for onButtonChanged, since Oculus Go controller has no analog buttons.
+
+  onModelLoaded: function (evt) {
+    var controllerObject3D = evt.detail.model;
+    var buttonMeshes;
+    if (!this.data.model) { return; }
+    buttonMeshes = this.buttonMeshes = {};
+    buttonMeshes.trigger = controllerObject3D.children[0].children[1].children[4];
+    buttonMeshes.trackpad = controllerObject3D.children[0].children[1].children[1];
+
```
dmarcos: Are not the objects in the glTF named? It will be more robust if we use names like in the Oculus Touch Controls: https://github.com/aframevr/aframe/blob/master/src/components/oculus-touch-controls.js#L184

src/components/oculus-go-controls.js
```diff
@@ -0,0 +1,179 @@
...
+    // we need to set trigger and trackpad material separately so that we can edit them independently
+    for (const buttonName in buttonMeshes) {
```
dmarcos: Declare `var buttonName` at the top

dmarcos: We don't use const

src/components/oculus-go-controls.js
dmarcos: Is this because the material is shared across different parts and you need to clone them so each button has its own?

dmarcos: Can this be handled at the glTF level?

dmarcos:
Congrats for your first PR. Almost ready to go! I made a few comments / questions.

src/components/oculus-go-controls.js
dmliao: There's two problems: one is that it's a shared material, the other is that the material itself doesn't seem to obviously change colors (though that might be because of how shiny / dark my scenes made the material look - I'm not too familiar with how the material properties should behave). 

It probably can be handled at the GLTF level, but I'm not sure how to go about doing that.

dmarcos:
Looking good. Thanks!

kevinsimper:
@dmliao Is there somewhere I can test this easily? Would love to test it out!

dmarcos:
@kevinsimper https://rawgit.com/aframevr/aframe/master/dist/aframe-master.min.js

kevinsimper:
@dmliao @dmarcos I tried it and it works! Really great!

However it works different from the normal in-game Oculus Go controller in that it follows the heads rotation where the Oculus Go controller normally follows a certain circle on the radius on that, eg. does not follow your heads rotation.

I think I like the default behavior as when I move my head, I don't my move hands at the same time, but with Aframe it looks like I move my hand which I don't.

Can the Oculus Go controller be made fixed as an option?

kevinsimper:
@machenmusik 

> Not using any arm model at all could be somewhat problematic as the controller would be rooted at the wrong spot it you turn 180 degrees and start interacting

There are several things that makes it works without arm:
- You will still have the ray
- The controller runs on its own circle, so it can go all around but on either your right and left side from a start
- The Oculus button pressing it resets the direction, but looking and pointing the same direction.

But doing the arm model, it will be more confusing as with WebVR you just came from the default and then into a weird Aframe arm model. It feels really weird.

machenmusik:
I am not sure what you are suggesting would be providing the positional update for the Go controller with no arm model, as it is not a 6DOF controller and the Go doesn't know where it is.  If you are saying that the browser is providing a full 6DOF pose using the native arm model, that would be great, but I would like to get some level of confirmation from the committers, as I expect they would have disabled the arm model if it was not required.

machenmusik:
That being said, it looks as though there is a simple flag to disable arm model if you like, see
https://github.com/aframevr/aframe/blob/961ba488d68e04eb1880e2988e53846279ed9b70/src/components/oculus-go-controls.js#L29

kevinsimper:
@machenmusik The oculus controller turns in the black radius where the red circle is the head 

image

machenmusik:
@kevinsimper right, but if you turn around (turn your head 180 degrees), without any arm model, your right hand would then become your left hand... the arm model is what keeps your right hand on the right.  Also you can turn it off if you don't like it, as described above.

kevinsimper:
@machenmusik We can't change how it works inside the Oculus Go, so shouldn't it work the same inside Aframe so that people don't get confused?

dmarcos:
@kevinsimper Did you get a chance to disable armModel? How does it work?

machenmusik:
@kevinsimper actually there is a native arm model for Oculus Go, and here is how to prove it... find a swivel chair and sit in it; put on your Oculus Go, go to the Home screen, center the controller and hold it out in front of you with your right hand; spin in the swivel chair slowly clockwise.  You'll notice at first the controller doesn't move in space, as you mention in your diagram above.  But as you keep spinning, you will notice that the controller starts to move in space to follow your rotation as well; it's the native arm model that is doing that.  Without any arm model, the controller would stay in the same position and would no longer be visible once you had trotted far enough.

The Oculus Go native arm model is almost certainly more sophisticated than the current basic arm model in A-Frame; pull requests with better arm model are certainly welcome, as are suggestions on how it could be improved even without implementation, as you have been doing (thank you BTW!)... and if the behavior with `armModel:false` is more to your liking, that is already a built-in A-Frame option.

dmarcos:
Is not the native arm model already applied and passed to the gamepad pose?

machenmusik:
https://judicious-random.glitch.me seems fine on Oculus Go
(https://glitch.com/edit/#!/judicious-random?path=index.html:3:2)

If all the other 3DOF combinations apply native arm model, maybe we can default to `armModel:false`
- [x] Oculus Browser on Go
- [ ] Oculus Browser on GearVR
- [ ] Samsung Internet on GearVR
- [ ] Chrome on Daydream
- [ ] anything else I forgot

machenmusik:
oh, I am wrong, armModel doesn't pass through `laser-controls`.  ewww.  will keep trying with that Glitch

machenmusik:
So the current state of https://judicious-random.glitch.me is coded for the Go only, and there is apparently no position provided when the arm model is disabled; the controller stays put in the floor. :-/

PR #3563, opened 2018-04-29
do not set `parentNode` -- it's a read-only property

bitonic:
we noticed this after getting errors when running babel on
a-frame (which we had to do since it now uses some features
that uglify does not support)

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
We should have documented why that was necessary. We should avoid touching it if we can. Can we tell Babel somehow to ignore it?

PR #3579, opened 2018-05-11
Substantially optimize coordinates.parse

mquander:
`coordinates.parse` winds up accounting for a moderate amount of time if you do per-frame updates of attributes containing new string or object coordinates. (In our project, an example was a raycaster component which we updated with a new direction and origin each frame.) This change makes it between 2x and 10x faster in both the string-input and object-input cases in Firefox and Chrome (depending on input characteristics) while being more or less equally straightforward.

[Here's a JSPerf comparing the old and new versions](https://jsperf.com/a-frame-coordinates-parsing/8).

There's one edge case I'm worried about. In the old version, if you gave the input `{x: 1, y: "2", splatoon: "42", "whiz": "bang"}`, it would return `{x: 1, y: 2, splatoon: 42, whiz: NaN}`. That is, it would parse any string keys at all in the input, even ones that it didn't know anything about, and even ones that aren't floats. That behavior seems strange to me, it didn't seem documented anywhere, and no test was exercising it, but it's true. If that behavior is important, then let me know and I'll hack it back in.

### DIFF
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -43,7 +43,7 @@ function parse (value, defaultVec) {
 
   vec = {};
   COORDINATE_KEYS.forEach(function (key, i) {
-    if (coordinate[i] !== undefined) {
+    if (coordinate[i]) {
       vec[key] = parseFloat(coordinate[i], 10);
     } else {
       var defaultVal = defaultVec && defaultVec[key];
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -5,7 +5,7 @@ var extend = require('object-assign');
 var warn = debug('utils:coordinates:warn');
 
 // Order of coordinates parsed by coordinates.parse.
-var coordinateKeys = ['x', 'y', 'z', 'w'];
+var COORDINATE_KEYS = ['x', 'y', 'z', 'w'];
 
 // Coordinate string regex. Handles negative, positive, and decimals.
 var regex = /^\s*((-?\d*\.{0,1}\d+(e-?\d+)?)\s+){2,3}(-?\d*\.{0,1}\d+(e-?\d+)?)\s*$/;
@@ -42,17 +42,15 @@ function parse (value, defaultVec) {
   coordinate = value.trim().split(/\s+/g);
 
   vec = {};
-  for (var i = 0; i < coordinateKeys.length; i++) {
-    var key = coordinateKeys[i];
-    if (coordinate.length > i) {
+  COORDINATE_KEYS.forEach(function (key, i) {
+    if (coordinate[i] !== undefined) {
       vec[key] = parseFloat(coordinate[i], 10);
     } else {
       var defaultVal = defaultVec && defaultVec[key];
-      if (defaultVal !== undefined) {
-        vec[key] = parseIfString(defaultVal);
-      }
+      if (defaultVal === undefined) { return; }
+      vec[key] = parseIfString(defaultVal);
     }
-  }
+  });
   return vec;
 }
 module.exports.parse = parse;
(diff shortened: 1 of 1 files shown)


src/utils/coordinates.js
```diff
@@ -21,26 +24,36 @@ function parse (value, defaultVec) {
...
+
   vec = {};
-  vec.x = coordinate[0] || defaultVec && defaultVec.x;
-  vec.y = coordinate[1] || defaultVec && defaultVec.y;
-  vec.z = coordinate[2] || defaultVec && defaultVec.z;
-  vec.w = coordinate[3] || defaultVec && defaultVec.w;
-  return vecParseFloat(vec);
+  for (var i = 0; i < coordinateKeys.length; i++) {
+    var key = coordinateKeys[i];
+    if (coordinate.length > i) {
```
dmarcos: Probably more clear if you do `if (coordinate[i] !== undefined)`

src/utils/coordinates.js
```diff
@@ -21,26 +24,36 @@ function parse (value, defaultVec) {
...
+      vec[key] = parseFloat(coordinate[i], 10);
+    } else {
+      var defaultVal = defaultVec && defaultVec[key];
+      if (defaultVal !== undefined) {
```
dmarcos: Early return preferable: 
```
  if (defaultVal === undefined) { continue; }
  vec[key] = parseIfString(defaultVal); 
```

src/utils/coordinates.js
```diff
@@ -21,26 +24,36 @@ function parse (value, defaultVec) {
...
-  coordinate = value.trim().replace(/\s+/g, ' ').split(' ');
+  coordinate = value.trim().split(/\s+/g);
+
   vec = {};
-  vec.x = coordinate[0] || defaultVec && defaultVec.x;
-  vec.y = coordinate[1] || defaultVec && defaultVec.y;
-  vec.z = coordinate[2] || defaultVec && defaultVec.z;
-  vec.w = coordinate[3] || defaultVec && defaultVec.w;
-  return vecParseFloat(vec);
+  for (var i = 0; i < coordinateKeys.length; i++) {
```
dmarcos: A bit more compact using forEach
```javascript
coordinateKeys.forEach(function (key, i) {
   if (coordinate[i] !== undefined) {
      vec[key] = parseFloat(coordinate[i], 10); 
   } else {
      if (defaultVal === undefined) { continue; }
      vec[key] = parseIfString(defaultVal); 
  }
}
```

src/utils/coordinates.js
```diff
@@ -4,6 +4,9 @@ var extend = require('object-assign');
 
 var warn = debug('utils:coordinates:warn');
 
+// Order of coordinates parsed by coordinates.parse.
+var coordinateKeys = ['x', 'y', 'z', 'w'];
```
dmarcos: Capitalize since it's a const `COORDINATE_KEYS`

dmarcos:
Some of the unit tests are failing with the change.

dmarcos:
Thanks for the PR. It looks good. One thing that you might to consider. If you are calling `setAttribute` on a tick I recommend reusing the same object, the values will be considered trusted. Type checking / parsing will be skipped altogether (https://aframe.io/docs/0.8.0/introduction/best-practices.html#tick-handlers) and you will reduce object allocation. For instance:
```javascript
init: function () {
 this.yourComponentData = {coords: {x: 0, y: 0, z:0}, enabled: true};
},
tick: function () {
  this.yourComponentData.coords.x += 1;
  this.el.setAttribute('your-component', this.yourComponentData);
}
```

ngokevin:
If you are in a tick, better to just avoid setAttribute altogether. We should have the raycaster component update the origin/direction directly. We already do the object reuse, that only saves on parsing.

mquander:
I made the style changes you suggested. The test failure was a real bug; it was the new version doing a bad job of handling the empty string or a string with all spaces as input. It should line up now. Performance still seems equally good.

Thanks for the advice about reusing the same object, that advice could really be worth its weight in gold for us. I didn't realize it before.

dmarcos:
Thank you. First PR achievement unlocked 🎉 Congrats!

PR #3588, opened 2018-05-18
use trigger as well as trackpad press for cursor (mouse) events

JoeStrout:
Unlike gearvr-controls, oculus-go-controls was only mapping trackpad press to mouse up/down.  This is contrary to standard usage of the Go controller, which almost always uses the trigger for point-and-click tasks.  So I added triggerdown/triggerup, just like on GearVR.

**Description:**

**Changes proposed:**
-
-
-


machenmusik:
Nice job, well done - and welcome to the community of A-Frame committers!

PR #3589, opened 2018-05-19
remove garbage and bubbling from tracked-controls

ngokevin:
**Description:**

Objects and strings were being created on every button touch, press, and up.

**Changes proposed:**
- Cache event details.
- Use constants versus concatenating strings for event names.

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -9,15 +9,29 @@ var EYES_TO_ELBOW = {x: 0.175, y: -0.3, z: -0.03};
 // Vector from eyes to elbow (divided by user height).
 var FOREARM = {x: 0, y: 0, z: -0.175};
 
+// Due to unfortunate name collision, add empty touches array to avoid Daydream error.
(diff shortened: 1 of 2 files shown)


src/components/tracked-controls.js
```diff
@@ -249,7 +269,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     for (i = 0; i < controllerAxes.length; i++) {
       this.axis.push(controllerAxes[i]);
     }
-    this.el.emit('axismove', this.axisMoveEventDetail);
+    this.el.emit(EVENTS.AXISMOVE, this.axisMoveEventDetail, false);
```
dmarcos: I’m ok with the PR but I believe these strings should be allocated on stack and not on the heap so no subjects for GC

ngokevin: I read [strings get collected](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management). But this PR also removes the string concatenations which do allocate and are messy anyways, constants are cleaner.

PR #3590, opened 2018-05-19
look-controls: in flat mode, allows camera z-rotation (by other control)

DougReeder:
**Description:**
Removes the line where camera z-rotation is always forced to 0

**Changes proposed:**
In v0.7.x, look-controls allowed experiences with the feel of flying; specifically, banking into turns. (for example, https://elfland-glider.surge.sh/)  The current implementation sets camera z-rotation to 0 on every tick, making look-controls incompatible with such experiences.  As z-rotation is initially zeroed, that line is not necessary to keep z-rotation zero on non-banking experiences.

If an experience needs to alter the z-rotation, it must still implement that separately.


DougReeder:
On my machine (MacBook Pro, mid-2014, running OX X 10.11.6) the tests pass when running
npm run test:chrome -- --single-run
SUMMARY:
✔ 1067 tests completed
ℹ 15 tests skipped

PR #3594, opened 2018-05-22
clear raycaster intersections when disabled (fixes #3593)

### LINKED ISSUE #3593: Raycaster should clear all current intersections when disabled.
**Description:**

- A-Frame Version:
- Platform / Device:

ngokevin:
**Description:**

Fix an issue where entities would not get `mouseleave`d when raycaster was disabled.

**Changes proposed:**
- When toggling raycaster to disabled, clear its intersections.
- Use event constants to define in one place now that they are used in multiple methods.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -20,6 +20,13 @@ var OBSERVER_CONFIG = {
   subtree: true
 };
 
+var EVENTS = {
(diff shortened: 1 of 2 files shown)


ngokevin:
Addressed / ready.

Issue #3129, opened 2017-10-06
consider having raycaster only emit the start of intersections


ngokevin:
**Description:**

Currently, the raycaster seems to emit an event + object per tick per intersected entity, which is quite a bit. Even if the intersected entity was already intersecting on the last tick. So imagine you are just hovering over anything without moving, they'll get continuously pounded with events. I think we need only emit these events on the intersection start.

```js
    // Emit intersected on intersected entity per intersected entity.
    for (i = 0; i < intersections.length; i++) {
      // TODO: Check previousIntersectedEls to see if this is a new intersection.
      intersections[i].object.el.emit('raycaster-intersected', {
        el: el,
        intersection: intersections[i]
      });
    }
```

machenmusik:
Is there a better way to understand when and how the intersection point changes?  I think that is the primary value of the subsequent events.  Use case - drawing a dot / placing an object on a plane or box where the intersection point is.

machenmusik:
Maybe emit the first intersection, and then the recipient should register on the raycaster for further updates for that element?  At the moment, I don't think the intersections are objects that survive across raycasts and update live.

ngokevin:
Hmm, I forgot about that. Perhaps a property to always keep emitting in case you need that...? Even then you might not need it for every object. Theoretically, it'd be most efficient to look up the intersection data yourself which is not too hard.

ngokevin:
Maybe a subscribe method to put your entity on a list to get a callback for intersection updates?

machenmusik:
I'm thinking use cases like - raycaster detects ground plane and also other objects in environment; if ground plane is first hit, you want live updates to highlight where you want the object placed, but if other objects, you don't highlight (and don't care where until).  For that use case, you want a property that will throttle the intersection count for which events are emitted (e.g. to 1, or 2 if you want to be able to point through objects if required)

ngokevin:
Could be done like so, perhaps not worth maintaining an API for what can be done with a bit of JS?

```js
groundEl.addEventListener('raycaster-intersected', el => {
  this.intersectingRaycaster = evt.detail.el;
});

groundEl.addEventListener('raycaster-intersected-cleared', () => {
  this.intersectingRaycaster = null;
});

...

tick: () => {
  if (!this.intersectingRaycaster) {  return; }
  var intersectionIndex = this.intersectingRaycaster.components.raycaster.intersectedEls.indexOf(this.el);
[...]

PR #3595, opened 2018-05-22
[breaking] remove raycaster-intersected garbage in favor of fetching getIntersection manually

ngokevin:
**Description:**

The intersection in the event detail requires object creation. Pass a function instead. The intersection data is only used in certain and more involved cases and calling a function to get the intersection data is easy.

Can update docs if looks good.

**Changes proposed:**
- Instead of emit an object per intersection on tick, reuse the event detail, and offer `getIntersection` method to fetch intersection data.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -75,6 +75,8 @@ module.exports.Component = registerComponent('raycaster', {
     this.otherLineEndVec3 = new THREE.Vector3();
     this.lineData = {end: this.lineEndVec3};
 
+    this.getIntersection = this.getIntersection.bind(this);
+    this.intersectedDetail = {el: this.el, getIntersection: this.getIntersection};
     this.intersectedClearedDetail = {el: this.el};
     this.intersectionClearedDetail = {clearedEls: this.clearedIntersectedEls};
     this.intersectionDetail = {};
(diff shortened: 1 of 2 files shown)


machenmusik:
Yeah, that is what I thought you and Don meant while discussing #3129

ngokevin:
I think you still need to filter compute newly intersected entities / cleared for the `raycaster-intersection` and `raycaster-intersection-cleared` events, even if you ignore the `raycaster-intersected*` events. The raycaster-side events only get emitted on newly intersected or cleared events.

machenmusik:
Right, but if you are already doing that yourself programmatically, the raycaster component doesn't really help because (unlike intersections) I don't think new and cleared are saved on the component.

machenmusik:
I think the use cases that were discussed were things like "I'm manipulating something, and don't care about others until no longer intersecting" which can be programmatic calls to getIntersection for the entity (or entities) of interest.

wmurphyrd:
Are the intersect**ed** events commonly used in practice? I've always just used on the inersect**ion** events as `super-hands` does its own version of communicating to the affected entities. Could save even more performance if these were removed.

ngokevin:
Used sometimes when you want to know something is intersected. There is the cursor, but that only emits on the nearest intersected.

The only thing we'd save is some event emits. I don't think that has much effect. The loops and processing will be done regardless. If you aren't using the events, then I'd make your own raycaster.

ngokevin:
Rebased.

mquander:
Would it be simpler to just include a reference to the raycaster component in the event detail, instead of a reference to its `getIntersection` function?

ngokevin:
Possibly, I think `getIntersection` is a bit more prescriptive / offers more guidance on what to do with the event. We could do both.

PR #3599, opened 2018-05-24
fix canvas textures being created as video textures

ngokevin:
**Description:**

https://stackoverflow.com/questions/50493717/how-to-get-aframe-to-only-upload-canvas-contents-to-texture-when-needed/50505535#50505535

Canvas textures was re-uploading to GPU every frame.

**Changes proposed:**
- Don't use VideoTexture for canvas...

### DIFF
--- a/examples/test/canvas-texture/components/canvas-updater.js
+++ b/examples/test/canvas-texture/components/canvas-updater.js
@@ -1,0 +1,14 @@
+/* global AFRAME */
+
+AFRAME.registerComponent('canvas-updater', {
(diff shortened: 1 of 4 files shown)


greggman:
this will probably work but it seems like the [canvas-texture](https://github.com/aframevr/aframe/blob/master/examples/test/canvas-texture/components/draw-canvas-rectangles.js) example will need to do something to set `needsUpdate` for the texture inside its `tick` handler

ngokevin:
Good catch, updated.

ngokevin:
Updated.

PR #3608, opened 2018-05-31
Fix bug in raycaster example

mquander:
It seems to me like this was always broken, and that it meant to get the top-level `target` on the event. After the fix, the example seems to correctly change the color of the big box that is the raycaster target when it is intersected, as I presume it was meant to. Tested on Firefox and Chrome.


mquander:
(Aha, it broke here: https://github.com/aframevr/aframe/pull/3308. That makes sense.)

PR #3609, opened 2018-05-31
Bump three.js to r93

mquander:
https://github.com/mrdoob/three.js/wiki/Migration-Guide#r92--r93
https://github.com/mrdoob/three.js/milestone/6?closed=1

`npm test` passes, and I ran all tests and examples in localhost:9000/examples on Ubuntu Firefox, without VR, and fixed a few simple warnings that flew out related to recent three.js API deprecations. Everything looks good at a glance.

It would probably be nice to test VR.

(Actually, the "Generic Logo" example doesn't work for me. But it doesn't seem related -- it also doesn't work for me on the [live site](https://aframe.io/aframe/examples/animation/generic-logo/). Tried Firefox and Chrome both.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "present": "0.0.6",
     "promise-polyfill": "^3.1.0",
     "style-attr": "^1.0.2",
-    "three": "0.92.0",
+    "three": "0.93.0",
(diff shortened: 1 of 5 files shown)


dmarcos:
Thanks! 👍

Issue #3573, opened 2018-05-08
WebGLInfo: Renamed primitives since three v91 (info)


arpu:
- A-Frame Version: A-Frame Version: 0.8.2 (Date 2018-05-07, Commit #050cbaf) three Version: 0.92.0 
- Platform / Device: Version 66.0.3359.139 (Offizieller Build) (64-Bit)
- Reproducible Code Snippet or URL: https://a-frobot.github.io/aframe/examples/performance/cubes/

https://github.com/mrdoob/three.js/commit/bd8f751f73a8c849540ffeb7346d697145c2b9f5#diff-489f2404e69a7f3736ada32f3af6c3c3

dmarcos:
What is exactly the problem?

arpu:
the stats Vertices and Faces are NaN
looks like the https://github.com/aframevr/aframe/blob/master/vendor/rStats.extras.js needs an update

ngokevin:
A PR for that would be welcome!

PR #3616, opened 2018-06-04
Fix NaN for faces & vertices on stats panel

benjaminleonard:
**Description:**
https://github.com/aframevr/aframe/issues/3573
Matching the name change in three.js to fix a NaN error in the stats panel

Not sure why the vertices stopped working also, as I couldn't find the change that broke this – are vertices shared? If so the calculation for vertices will be incorrect.

EDIT: Ignore my stupid mistake with the vertex calculation!

### DIFF
--- a/vendor/rStats.extras.js
+++ b/vendor/rStats.extras.js
@@ -122,9 +122,6 @@ window.threeStats = function ( renderer ) {
         },
         'renderer.info.render.points': {
             caption: 'Points'
-        },
-        'renderer.info.render.vertices': {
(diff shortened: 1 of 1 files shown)


benjaminleonard:
Just found this: https://github.com/mrdoob/three.js/pull/13404
Should vertices be removed from the panel all together?

ngokevin:
Thanks!

Could you also PR this to https://github.com/spite/rstats?

benjaminleonard:
Done, thanks!

PR #3620, opened 2018-06-05
Remove fromExternal flag from exitVR / enterVR. VRManager vr mode cannot be enabled / disabled with the flag present

dmarcos:
@machenmusik Probably the `fromExternal` flag is no longer necessary. I cannot find a reason why we would still need it. Happy to reintroduce if still relevant.


dmarcos:
@machenmusik Am I missing any scenario where the `fromExternal` flag might be necessary?

Issue #3623, opened 2018-06-05
Sound component can't use the same audio resource more than once.


yuseitahara:
If I use an audio resource(sound.mp3 for example) in an entity only, then it works, but if I use it in more than two entities, then only the first entity can load that resource and the other entities fail to load it. That is because AudioContext.decodeAudioData() is a destructive function, once an ArrayBuffer is given to that function, that ArrayBuffer becomes empty. An empty ArrayBuffer remains in THREE.Cache. Then if the next sound component uses the same audio resource(the same asset or the same url), since that ArrayBuffer is already empty, decodeAudioData fails to load, so other sound components can't load the audio.

AFAIK this is a problem of THREE.AudioLoader. But I would like to fix it in aframe's sound component because it is a basic feature of aframe and beginners like me must be confused by this bug. Could you advise me something? My idea is to implement a custom audio loader based on THREE.AudioLoader. I will make a pull request.

- A-Frame Version:0.8.2
- Platform / Device:All
- Reproducible Code Snippet or URL:https://jsfiddle.net/yuseit/egsdcm47/23/

dmarcos:
I would circle with the THREE folks to see if it makes sense to enhance THREE.AudioLoader to fulfill your needs. Can you open an issue in THREE with a proposal and reference this issue to keep track?

yuseitahara:
Yes, I will do. Thank you.

yuseitahara:
This issue has already been reported on THREE.

mrdoob/three.js#13710

dsinni:
This may be a dupe of #3454

yuseitahara:
This problem has been fixed in three.js r94.

PR #3626, opened 2018-06-07
Add a test for sound component (#3623)

yuseitahara:
This test makes sure that sound component can load the same audio resource more than once.


yuseitahara:
A test for #3623. This test fails now.

dmarcos:
Cool, thanks. We don’t want to merge failing tests. We can incorporate this once the problem is solved on the THREE side

yuseitahara:
I will recreate this PR.

PR #3641, opened 2018-06-13
Ability to cap the canvas size

benjaminleonard:
**Description:**
Adding an attribute to the scene component which allows you to set a maximum height/width.
Currently, you get some big frame drops on high-density screens in a full width window – especially on iMacs & Macbooks with retina screens and slightly less powerful GPUs.

Was just a first stab.
Would it be more useful with a separate width and height attribute?
Tried this rather than a pixel ratio attribute, as you can make the most of the high-density screen at smaller sizes.


ngokevin:
Nice. I wanted this because, yeah, fullscreen on Macbook is a drain.

You don't have to build the dist for PRs, you can remove that.

Perhaps `renderer="maxCanvasWidth: 1280; maxCanvasHeight: 720"`? And I would make it so that if you enter VR, the height and width are ignored?

benjaminleonard:
Great!

I've implemented the two separate attributes, and added them to the renderer schema to stop the warning. Though the default doesn't do anything. 

Size caps are ignored in VR mode.

src/core/scene/a-scene.js
```diff
@@ -652,19 +659,50 @@ module.exports.AScene = registerElement('a-scene', {
...
   return {
-    height: window.innerHeight,
-    width: window.innerWidth
+    height: size.height,
+    width: size.width
   };
 }
 
+function getMaxSize (max, isVR) {
+  var size = {
```
ngokevin: define vars-on-top and initialize where needed.

```
var aspectRatio;
var size;
```

src/core/scene/a-scene.js
```diff
@@ -652,19 +659,50 @@ module.exports.AScene = registerElement('a-scene', {
...
+
   return {
-    height: window.innerHeight,
-    width: window.innerWidth
+    height: size.height,
+    width: size.width
   };
 }
 
+function getMaxSize (max, isVR) {
```
ngokevin: I would call all these variables like `definedMaxSize` or `maxSize`.

ngokevin: Can you document this function and what `isVR` will do?

src/core/scene/a-scene.js
```diff
@@ -652,19 +659,50 @@ module.exports.AScene = registerElement('a-scene', {
  * @param {object} canvasEl - the canvas element
  * @param {boolean} embedded - Is the scene embedded?
  */
-function getCanvasSize (canvasEl, embedded) {
+function getCanvasSize (canvasEl, embedded, max, isVR) {
```
ngokevin: I think better order is `canvasEl, maxSize, isVR, isEmbedded` of importance

src/core/scene/a-scene.js
```diff
@@ -496,10 +496,17 @@ module.exports.AScene = registerElement('a-scene', {
         if (this.hasAttribute('renderer')) {
           rendererAttrString = this.getAttribute('renderer');
           rendererAttr = utils.styleParser.parse(rendererAttrString);
+
           if (rendererAttr.antialias && rendererAttr.antialias !== 'auto') {
             rendererConfig.antialias = rendererAttr.antialias === 'true';
           }
+
+          this.maxSize = {
```
ngokevin: `this.maxCanvasSize`

docs/components/renderer.md
```diff
@@ -16,7 +16,9 @@ The `renderer` component configures a scene's
 <a-scene renderer="antialias: true;
                    gammaOutput: true;
                    sortObjects: true;
-                   physicallyCorrectLights: true;"></a-scene>
+                   physicallyCorrectLights: true;
+                   maxCanvasWidth: 1920;
+                   maxCanvasHeight: 1920;"></a-scene>
```
ngokevin: 1080 would be better default

src/components/scene/renderer.js
```diff
@@ -13,6 +13,8 @@ var warn = debug('components:renderer:warn');
 module.exports.Component = register('renderer', {
   schema: {
     antialias: {default: 'auto', oneOf: ['true', 'false', 'auto']},
+    maxCanvasWidth: {default: -1},
```
ngokevin: 1920 x 1080 might be a reasonable default?

src/core/scene/a-scene.js
```diff
@@ -652,19 +659,50 @@ module.exports.AScene = registerElement('a-scene', {
...
   };
 }
 
+function getMaxSize (max, isVR) {
+  var size = {
+    width: window.innerWidth,
+    height: window.innerHeight
+  };
+
+  if (!max || isVR || (max.width === -1 && max.height === -1)) return size;
```
ngokevin: Braces around if statement

docs/components/renderer.md
```diff
@@ -61,3 +66,8 @@ be adjusted when making this change. Performance is not significantly affected i
 
 > **NOTE:** When glTF models contain lights, use the physically-correct lighting mode to match
 > the results in the original modeling tool.
+
+### maxCanvasWidth / maxCanvasHeight
```
ngokevin: I think can be explained in the attributes table

src/components/scene/renderer.js
benjaminleonard: Shouldn't the default be uncapped?

docs/components/renderer.md
benjaminleonard: Can do – but I suppose in most instances you'd be best having them the same value thereby capping for both landscape and portrait aspect ratios

src/components/scene/renderer.js
ngokevin: I think it might be good to limit as default to 1920x1080 because higher might be overkill and degrade performance across the board. I think all A-Frame apps should be capped because so many people have Retina screens. For example, ShaderToy caps to a very small resolution.

VR mode is the more important part, I think it's okay to cap what is essentially the preview. What do you think?

ngokevin: Alright

benjaminleonard: So I added the defaults here, since it avoids the warning for an undefined attribute – but the default value doesn't have an effect, hence this line:

`width: rendererAttr.maxCanvasWidth ? parseInt(rendererAttr.maxCanvasWidth) : -1`

Is that a problem?

ngokevin:
I played around. I am not sure setSize helps with Retina. The size is already reported as 1440x900 and pixel ratio is 2. Setting pixel ratio to 1 seems like what you'd want to do for Retina?

benjaminleonard:
Ahh, of course.

Let's say your screen was 1920x1080 at pixel ratio of 3 – and the cap is set to 2560 x 2560. If the pixel ratio was then set to 1 but a size larger than the screen, would it cause any problems?

benjaminleonard:
Agreed that would be much simpler – but maybe there's value in the max size, since it means you can make the most of the pixel ratio at smaller window sizes where it wouldn't be a frame rate issue?
Perhaps we could add an option for pixel ratio also?

I've just pushed a change that should hopefully fix the sizing on retina screens.

ngokevin:
Thanks! Let's see how it goes.

PR #3643, opened 2018-06-14
[docs] Add a section about non-ascii characters to text component.

yuseitahara:
It is essential for non-English users like me to know how to use non-ASCII characters in A-Frame. I added a section about how to generate a custom msdf font to use non-ASCII characters. In my case, it is Japanese characters.

### DIFF
--- a/docs/components/text.md
+++ b/docs/components/text.md
@@ -166,6 +166,29 @@ the SDF font, DejaVu:
 [fontexampleimage]: http://i.imgur.com/iWtXHm5.png
 ![Arial Black vs. Deja Vu][fontexampleimage]
 
+### Non-ASCII Characters
+
+To use non-ascii characters, you need to create your own custom font. The easiest way is to use [a web-based MSDF tool][msdf-bmfont-web]. Select your required character set and generate your own custom msdf font zip file.
+
+Once you download your custom msdf font zip file, extract it and rename `*.png` to `*-msdf.png`, then put both png and json files to your A-Frame directory.
+
+Lastly, you should specify the character set used in your HTML by using `<meta>` tag to avoid text to be garbled. If your text is garbled, it is not rendered.
+
+```html
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <script src="https://aframe.io/releases/0.8.2/aframe.min.js"></script>
+  </head>
+  <body>
+    <a-scene>
+      <a-sky color="lightblue"></a-sky>
+      <a-text value="ABCあいうえお日本語" font="custom-msdf.json" negate="false" scale="2 2 1" position="-2 2 -4"></a-text>
+    </a-scene>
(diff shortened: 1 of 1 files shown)


docs/components/text.md
```diff
@@ -166,6 +166,29 @@ the SDF font, DejaVu:
...
+
+Once you download your custom msdf font zip file, extract it and rename `*.png` to `*-msdf.png`, then put both png and json files to your A-Frame directory.
+
+Lastly, you should specify the character set used in your HTML by using `<meta>` tag to avoid text to be garbled. If your text is garbled, it is not rendered.
+
+```html
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <script src="https://aframe.io/releases/0.8.2/aframe.min.js"></script>
```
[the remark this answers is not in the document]
yuseitahara: It seems that https://aframe.io/releases/latest/aframe.min.js is deprecated and is permanently pinned to v0.1.3...

docs/components/text.md
```diff
@@ -166,6 +166,29 @@ the SDF font, DejaVu:
 [fontexampleimage]: http://i.imgur.com/iWtXHm5.png
 ![Arial Black vs. Deja Vu][fontexampleimage]
 
+### Non-ASCII Characters
+
+To use non-ascii characters, you need to create your own custom font. The easiest way is to use [a web-based MSDF tool][msdf-bmfont-web]. Select your required character set and generate your own custom msdf font zip file.
+
+Once you download your custom msdf font zip file, extract it and rename `*.png` to `*-msdf.png`, then put both png and json files to your A-Frame directory.
```
donmccurdy: Hm do we know why the PNG needs to be renamed? The filename is listed inside the JSON file; I'm surprised by this. 🤔

yuseitahara: That is because text component loads an image named fontName-msdf.png by default. It does not use image filename in json.

See https://github.com/aframevr/aframe/blob/dcdc1235600202c8c0b1f7ac0df21a8cc4018ef6/src/components/text.js#L253

yuseitahara: Should we fix text component?

yuseitahara:
Is it okay to merge this PR? Is there anything missing?

donmccurdy:
Let's try to fix the `text` component's reliance on same-name textures: https://github.com/donmccurdy/aframe-extras/pull/247. Otherwise this looks good to me. If that PR will take some time, then I'd prefer to merge this one and recommend that users use the `fontImage` path rather than relying on our inference of the image URL from the JSON filename.

docs/components/text.md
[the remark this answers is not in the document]
ngokevin: Correct, we do not use `latest`. People would use `latest`, then A-Frame upgrades, then hundreds of sites would break instantly.

PR #3648, opened 2018-06-16
Eliminate redundant work in raycaster component

mquander:
`THREE.Object3D.getWorldPosition` [already internally calls](https://github.com/mrdoob/three.js/blob/12e7e072a595dfba6374a36c76da308ec0a04cf4/src/core/Object3D.js#L446) `updateMatrixWorld` in order to always produce an up-to-date position, so we don't want to call it ourselves first -- that's just doing the work twice.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -320,8 +320,7 @@ module.exports.Component = registerComponent('raycaster', {
         return;
       }
 
-      // Grab the position and rotation.
-      el.object3D.updateMatrixWorld();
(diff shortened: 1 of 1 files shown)


mquander:
By the way, I would be interested to understand why `raycaster.checkIntersections` calls `setTimeout(drawLine)` at the end. Why can't it just call `drawLine` directly? Alternatively, can `drawLine` happen in the next `tick`, instead of in a separate timeout? I ask because the raycaster code is fairly performance-sensitive for our team and `setTimeout` is relatively slow.

ngokevin:
Not sure, should have left a comment back then. You can try removing it or moving to a tock.

PR #3651, opened 2018-06-18
[text] Read URL of first texture page from JSON.

donmccurdy:
The JSON file in an MSDF font should contain the paths of any textures it relies on. Currently those are ignored (https://github.com/aframevr/aframe/pull/3643#discussion_r195370194) and we try to infer the path from the JSON file name. For fonts created with http://msdf-bmfont.donmccurdy.com/, the paths in the JSON are correct but inferred paths are not.

For some of the default fonts, it looks like the JSON contains absolute paths on disk, which clearly won't work — not sure what tool was used to create those, and whether it should be fixed? For now this PR looks for absolute non-HTTP(S) paths and ignores them.

### DIFF
--- a/src/components/text.js
+++ b/src/components/text.js
@@ -36,6 +36,9 @@ var cache = new PromiseCache();
 var fontWidthFactors = {};
 var textures = {};
 
+// Regular expression for detecting a URLs with a protocol prefix.
+var protocolRe = /^\w+:/;
+
 /**
  * SDF-based text component.
  * Based on https://github.com/Jam3/three-bmfont-text.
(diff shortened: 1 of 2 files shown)


src/components/text.js
```diff
@@ -249,8 +251,15 @@ module.exports.Component = registerComponent('text', {
   },
 
   getFontImageSrc: function () {
+    if (this.data.fontImage) { return this.data.fontImage; }
     var fontSrc = this.lookupFont(this.data.font || DEFAULT_FONT) || this.data.font;
```
ngokevin: Don't know if we're following the `vars-on-top` style...it was insisted a while back and sometimes it's not followed very strictly, and sometimes just results in a mess of unordered var declarations / initializations at the top. And then when you make small code changes, need to shuffle the function to match the rule. Just ranting 😛

donmccurdy: No preference, I don't use it strictly in my own code.  ¯\_(ツ)_/¯

donmccurdy:
Added tests.

donmccurdy:
😅 https://github.com/etiennepinchon/aframe-fonts/pull/5

Issue #3276, opened 2017-12-10
Raycaster reports intersections with entities excluded by objects property when nested


wmurphyrd:
**Description:**

Entities that are excluded by `raycaster.objects` can be reported as an intersected entity if they are a child of a selected entity. It makes sense that descendant meshes are included in the raycaster's search, but I would expect that the the parent which is specified in `objects` would be reported as the intersected entity in the `raycaster-intersection` event so that the reported intersections are exclusively with entities specified in `objects`

I can work on this if you agree

- A-Frame Version: 0.7.0 & master
- Platform / Device: desktop Firefox
- Reproducible Code Snippet or URL: https://cdn.rawgit.com/aframevr/aframe/c2f08cf0/dist/aframe-master.min.js

Only '#box' is included in `raycaster.objects`, but '#sphere` will appear as the event target when clicked.

ngokevin:
Agreed, just ran into this bug.

wmurphyrd:
Fixed in #3652

PR #3652, opened 2018-06-19
fix raycaster flatten to only include object3DMap objects

ngokevin:
**Description:**

I was doing a non-recursive raycast, and defined some objects that had children. These children were getting added to the `objects` set due to the `flattenChildrenShallow`. So I had an entity with 700+ non-visible children that were toggled off, and they were getting added to the raycast set.

The original intent of the logic was to only add like `getObject3D('mesh')` and such. So rather than loop over children, loop over `object3DMap`. Makes the logic simpler as well.

**Changes proposed:**
- Don't add all of groups' children to `.objects`, only add `object3DMap` keys.

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -169,7 +169,7 @@ module.exports.Component = registerComponent('raycaster', {
     els = data.objects
       ? this.el.sceneEl.querySelectorAll(data.objects)
       : this.el.sceneEl.children;
-    this.objects = this.flattenChildrenShallow(els);
+    this.objects = this.flattenObject3DMaps(els);
     this.dirty = false;
   },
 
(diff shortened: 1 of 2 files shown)


ngokevin:
r?

ngokevin:
@donmccurdy want to check it out since you did the initial flatten

3DataWill:
This resolves #3276.  Perhaps worth noting/documenting that it also changes the behavior when recursive is true such that it no longer traverses the entity hierarchy.

joshmarinacci:
I think I just ran into this issue. Using the latest from master recursive:true no longer works. I can target the nested elements by using a class but if they don't have a class there is no way to target them.  This pretty much breaks the default behavior.

PR #3662, opened 2018-06-22
Adding Workshop demos from the A-Frame NYC meetup

rdub80:
**Description:**
adding workshop resources


ngokevin:
This perhaps belongs on the community page, next to the existing NYC entry (https://github.com/aframevr/aframe-site/blob/master/src/community/index.md), rather than the root introduction page rather than as the core steps to getting started with A-Frame.

PR #3668, opened 2018-06-26
Log errors during node initialization

mquander:
Previously, this would just fail silently. This branch is running a lot of user code (i.e. component `init`) so it's frustrating not to get any console output if it fails.

An easy way to test is to open one of the examples and throw an error during `init`.


ngokevin:
Thanks!

PR #3680, opened 2018-07-03
[doc] document property interface

### LINKED ISSUE #3531: Document property interface
As far as I can tell there is no documentation of the full interface for property schema. in *Component -> Schema* we mention `type`, `default`, and `parse`, but then `if` and `stringify` are not mentioned.

https://aframe.io/docs/0.8.0/core/component.html#schema

krzysztof-grzybek:
Fixes #3531

I added docs for `stringify`, but I don't what `if` is about (mentioned in issue).  @donmccurdy any hint about that?


ngokevin:
Thanks! `if` is not terribly important. Just for Inspector for certain types of components.

PR #3682, opened 2018-07-04
a-image transparent is by default true

T00mm:
**Description:**
According the source code this value is by default true. Is the documentation at fault or the code?

**Changes proposed:**
- Edit the documentation to correctly show the default value
-
-


ngokevin:
Thanks!

PR #3683, opened 2018-07-04
initialize/prioritize camera initialization in system step before loading the scene (fixes #2332)

### LINKED ISSUE #2332: Prioritize camera initialization
**Description:**

While the scene is loading, the camera component and system will set `sceneEl.camera` and make it active. But the camera entity will be loaded in the same priority as all the other entities. Any components wishing to make use of the camera have to wait for `camera-set-active` event like we do in the motion capture project

If we have the scene initialize the camera before all the other components (whether the camera is default, or defined), then it'll be easier for other components to use the camera.

ngokevin:
**Description:**

Make A-Frame a bit more robust with components not needing to wait on camera initialization of active set (unless the component wants to handle active camera switching). Before some components had access to the camera, and some components did not, depending on luck on the load order. 

Similar to how we prioritized canvas initialization.

**Changes proposed:**
- Initialize a camera before loading the scene/entities/components.


ngokevin:
Fixing tests for Chrome...

Issue #3454, opened 2018-03-15
Regression: Sound component not playing audio asset across multiple entities in 0.8.*


dsinni:
**Description:**

When multiple entities utilize the `sound` component, and use the same audio asset source, only one instance of audio is created in `0.8.*`.

This does work as expected in `0.7.*` as long as an `<audio>` element is used rather than an `<a-asset-item>`.

In `0.8.*`, only the first entity of its kind in the code emits audio and `TypeError: Argument of AudioContext.decodeAudioData can't be a detached buffer - three.js:37939:18` is thrown.

As I expected, `poolSize` has no effect, I'm assuming due to the fact that this is across multiple instances of the `sound` component and not contained within one component.

- A-Frame Version: `0.8.0` and `0.8.1`
- Platform / Device: Chrome, Firefox, Edge on Windows 10 (all I've tested thus far)
- Reproducible Code Snippet or URL: https://codepen.io/dansinni/pen/yKOjem

arpu:
looks like the three positional audio works normal
https://threejs.org/examples/#webaudio_sandbox

dsinni:
@arpu Thanks for the reply. Yeah, it's not so much the positional audio that's not working; it's more specific to when you have multiple entities using the same audio asset, which is pretty common in my work, so this has been a blocking issue for me. It works well in `0.7.*`, just not in `0.8.0`. Hoping this is addressed soon so I can use `0.8.*` in production.

meatwallace:
we've resolved the issue on a local branch and will be getting a PR up in the next few hours.

dsinni:
Rad. Thanks, @meatwallace.

meatwallace:
@dsinni - sorry, jumped the gun a little - we'll be opening a PR, but the issue is in Three.js, so not certain what that means for A-Frame exactly as it's using a fork. regardless, i'll post back as soon as I open the PR.

dsinni:
Ah, bummer. Okay. Thanks, @meatwallace. Appreciate it. Hoping it can be resolved without too much hassle. It may not be a breaking issue, but it really is an important bug, IMO; I'm surprised more people haven't noticed/reported it (or maybe they have on `three.js`).

meatwallace:
@dsinni - the PR can be found [here](https://github.com/mrdoob/three.js/pull/13710). Not certain it'll be accepted, as it might be more of a workaround rather than a fix.

[...]

PR #3687, opened 2018-07-07
Add a test for sound component.

yuseitahara:
This test makes sure that sound component can load the same audio resource more than once. This is a test for #3623.


yuseitahara:
This is a test for #3454 too.

dmarcos:
Cool. Glad that it got fixed in THREE r94. Thanks for remembering writing the test and congrats for your first PR. Keep them coming 🥇

Issue #3525, opened 2018-04-06
Scene becomes "squished" when changing orientation on iOS Firefox and Safari


SteveX:
**Description:**
Viewing the most basic example https://aframe.glitch.me/ on iPhone SE or iPad Pro 12.9" using Firefox and Safari, when you change device orientation the sphere, cylinder and cube can become "squished", so the sphere is ovoid etc. 

On the iPhone this happens when the browser is showing the controls, and also when started from a Safari Add to Home icon. 

On the iPad 12.9", this doesn't happen when the browser controls are showing, but does when the scene is started from a Safari Add to Home icon.

- A-Frame Version:0.8.0
- Platform / Device: Firefox and Safari on iOS / iPhone SE & iPad Pro 12'9"
- Reproducible Code Snippet or URL: https://aframe.glitch.me/

RSpace:
I've investigated this issue. There's some funky bugs making it out into the Mobile Safari releases, like this one:
https://hackernoon.com/onresize-event-broken-in-mobile-safari-d8469027bf4d

However, there's already been a fix added for that (https://github.com/aframevr/aframe/issues/3282), and so this appears to be something else. Specifically, it has do to with how Mobile Safari returns the width and height of the window. Inspired by the article above and how it's done on apple.com, I found out that this bug goes away if we read the size of the viewport from `document.body.offsetWidth` and -`Height` instead of from `window.innerWidth` and -`Height`.

I've proposed this fix in this PR:
https://github.com/aframevr/aframe/pull/3711

PR #3711, opened 2018-07-30
Fix iOS orientation change squeeze

RSpace:
**Description:**

See https://github.com/aframevr/aframe/issues/3525

**Changes proposed:**
- Fix bug where `isiOS` is not checked on the right scope
- Use document body offset width and height instead of window inner width and height

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -120,7 +120,7 @@ module.exports.AScene = registerElement('a-scene', {
           // after an orientation change. The window size is correct if the operation
           // is postponed a few milliseconds.
           // self.resize can be called directly once the bug above is fixed.
-          if (this.isIOS) {
+          if (self.isIOS) {
             setTimeout(resize, 100);
           } else {
             resize();
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -709,7 +709,7 @@ function getMaxSize (maxSize, isVR) {
   var size;
   var pixelRatio = window.devicePixelRatio;
 
-  size = {height: window.innerHeight, width: window.innerWidth};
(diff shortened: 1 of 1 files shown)


src/core/scene/a-scene.js
```diff
@@ -709,7 +709,7 @@ function getMaxSize (maxSize, isVR) {
   var size;
   var pixelRatio = window.devicePixelRatio;
 
-  size = {height: window.innerHeight, width: window.innerWidth};
+  size = {height: document.body.offsetHeight, width: document.body.offsetWidth};
```
dmarcos: Can you describe what's the difference with respect to `window.innerHeight`?

RSpace: That's a fair question, but I can't give a crystal clear answer. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements), the `offsetWidth` and `offsetHeight` is the size of an element _including_ padding and border, but _excluing_ any margin.

The `<body>` element is always 100% wide AFAIK, but not necessarily 100% height. The height of a `<body>` is determined by what's inside it and how it's styled, so while `document.body.offsetWidth` should generally always be equal to `window.innerWidth`, `document.body.offsetHeight` is certainly not guaranteed to always be equal to `window.innerHeight`.

However, in this case we only use the `offsetWidth` and `offsetHeight` when the A-Frame scene is not embedded, meaning it should be styled to take up 100% width and height. This in turn should mean that the body has also been styled to a 100% width and height.

Finally, while is debatable whether `window.innerHeight` or `document.body.offsetHeight` is the best way to determine the height, it _does_ fix this very annoying bug in Mobile Safari, so I guess the question is if anyone has a non-standard setup where making this change will break their experience.

PR #3714, opened 2018-08-01
Update a-video.md

KevinEverywhere:
The caveats at the bottom of the page are no longer accurate. Specifically, the video can work in the browser when not on the home screen (line removed); and since iOS 11,  user interaction is required to play video. 

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thanks! Congrats on your first PR! 🥇

PR #3717, opened 2018-08-03
add logarithmicDepthBuffer option

Tino-F:
**Description:**
I have ran into an issue while experimenting with AR.js (https://github.com/jeromeetienne/AR.js/issues/146) that required me to enable [logarithmicDepthBuffer](https://threejs.org/examples/?q=log#webgl_camera_logarithmicdepthbuffer). AFrame does not include this option, so I thought that I would include it myself. 

**Changes proposed:**
- added logarithmicDepthBuffer option to the scene as an attribute.
- added logarithmicDepthBuffer option to renderer component (not able to modify after scene initialization)

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -515,7 +515,7 @@ module.exports.AScene = registerElement('a-scene', {
         var rendererAttrString;
         var rendererConfig;
 
-        rendererConfig = {alpha: true, antialias: !isMobile, canvas: this.canvas};
+        rendererConfig = {alpha: true, antialias: !isMobile, canvas: this.canvas, logarithmicDepthBuffer: false};
         if (this.hasAttribute('antialias')) {
           rendererConfig.antialias = this.getAttribute('antialias') === 'true';
         }
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -520,6 +520,10 @@ module.exports.AScene = registerElement('a-scene', {
           rendererConfig.antialias = this.getAttribute('antialias') === 'true';
         }
 
+        if (this.hasAttribute('logarithmicDepthBuffer')) {
+          rendererConfig.logarithmicDepthBuffer = this.getAttribute('logarithmicDepthBuffer') === 'true';
+        }
+
         this.maxCanvasSize = {height: 1920, width: 1920};
 
         if (this.hasAttribute('renderer')) {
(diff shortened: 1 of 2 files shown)


Tino-F:
*(not able to modify AFTER scene initialization)

src/core/scene/a-scene.js
```diff
@@ -520,6 +520,10 @@ module.exports.AScene = registerElement('a-scene', {
           rendererConfig.antialias = this.getAttribute('antialias') === 'true';
         }
 
+        if (this.hasAttribute('logarithmicDepthBuffer')) {
```
dmarcos: Let's initialize `logarithmicDepthBuffer` to false here: https://github.com/aframevr/aframe/pull/3717/files#diff-eb9bb1084863a2f65678ea67d6e7cf5eR518

dmarcos:
Thanks! A small modification needed and ready to go.

src/core/scene/a-scene.js
Tino-F: The logarithmicDepthBuffer option has been set to false by default!

andreyrd:
I don't think the component version of this actually works.

`<a-scene renderer="logarithmicDepthBuffer: true">`
When setting logarithmicDepthBuffer on [the renderer component](https://github.com/aframevr/aframe/blob/master/src/components/scene/renderer.js#L47), the property is checked to make sure it's not being changed after initialization, but it's never actually set on the renderer, nor is it [used by a-scene](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L529). I think it needs to be treated just like antialias [here](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L533).

`<a-scene logarithmicDepthBuffer="true">`
The attribute on the primitive works fine.

Tino-F:
Yes @andreyrd you're exactly right. `logarithmicDepthBuffer` is a property that can only be set on the renderer's creation just like `antialias`.  The renderer component only allows you to modify the renderer's settings after initialization, limiting you to only a few different properties to modify.

I noticed a note made in the code [here](https://github.com/Tino-F/aframe/blob/d210a212d5666c7e5e5b36e557073e674df85f86/src/components/scene/renderer.js#L31) to enable modification of `antialias` in the renderer component for v0.0.9. Of course when this gets implemented, doing the same for logarithmicDepthBuffer will more than likely be a matter of copying and pasting

pancakewaffles:
Not sure if this is a bug but using logarithmicDepthBuffer, rendering text is problematic. In fact, no text is rendered. 

JSFiddle here: https://jsfiddle.net/7aLh69gq/17/

You can toggle the logarithmicDepthBuffer setting from false to true and see the effects of it on the text. Moving the text forwards or backwards has no effect.

PR #3725, opened 2018-08-09
Emit correct detail for stateadded event (fix #3723)

### LINKED ISSUE #3723: stateadded/stateremoved documentation is wrong
Documenation for `a-entity` `stateadded` and `stateremoved` event says `evt.detail.state` contains the state that was added/removed, and the example shows the same: 
```
entity.addEventListener('stateadded', function (evt) {
  if (evt.detail.state === 'selected') {
    console.log('Entity now selected!');
  }
});
```
but  `state` is not actually defined on detail. Instead `evt.detail` is set directly to the state. https://github.com/aframevr/aframe/blob/030023f5aab128ce0cbe7aa24b83de5ec248cc04/src/core/a-entity.js#L750

Either the documentation is wrong or `a-entity` should be updated to emit `{state: state}`. A third option would be to imply that `addState` should be passed an object, but given nothing in a-frame does that, it is probably not correct.

- A-Frame Version: head
- Platform / Device: any

dmarcos:
**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -747,7 +747,7 @@ var proto = Object.create(ANode.prototype, {
     value: function (state) {
       if (this.is(state)) { return; }
       this.states.push(state);
-      this.emit('stateadded', state);
+      this.emit('stateadded', {state: state});
     }
   },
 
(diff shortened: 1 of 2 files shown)


ngokevin:
This was removed to save an object allocation, use `evt.detail` and if the docs are incorrect, correct the docs instead.

dmarcos:
What about reusing an object to prevent the allocation? It would be nice to keep detail consistently as an object

dmarcos:
Just ammended docs

PR #3730, opened 2018-08-13
fix systems.md

gftruj:
By accident in my [commit](https://github.com/aframevr/aframe/commit/6c593aafea5f0eee59b5f673cfa679ed0da8cba1), I've deleted `---` which seem to be necessary for the docs creator.


ngokevin:
thanks!

PR #3733, opened 2018-08-15
Include source files in the installed A-Frame package

RSpace:
Include source files in the installed A-Frame package to allow developers to reference individual files and debug easier.

See discussion on https://github.com/aframevr/aframe/pull/3712

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
   "license": "MIT",
   "files": [
     "dist/*",
-    "docs/**/*"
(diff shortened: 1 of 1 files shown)


ngokevin:
I'm okay with requiring from `src` if it's just for debugging A-Frame installed through npm. Otherwise, the alternative is to go thru `git clone`, `npm install`, `npm link`, and then build the `dist` on every change.

PR #3741, opened 2018-08-17
check geometry is set first in text updateLayout

ngokevin:
**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Tests failing

PR #3742, opened 2018-08-17
make a-frame unit test setups more robust in the asynchrony 

ngokevin:
**Description:**

Don't rely on a timeout for the node to attach.

**Changes proposed:**
-
-
-


ngokevin:
updated. just small improvements  / code quality to test setups. gonna leave for now, and perhaps go thru more later.

Issue #3675, opened 2018-06-30
Add aframe-event-decorators to utils.


ngokevin:
**Description:**

It's useful!

@Adam4lexander

Adam4lexander:
Sure! I'll be happy to get on this.

ngokevin:
Hey, got time to copy it in? Add as a separate module to `utils.js` in like `events.js` or `decorators.js`, then attach it to `AFRAME.utils.bindEvent`.

Adam4lexander:
Sure thing, will look into this within the next few days! Thanks for the clarification of where to put it.

dmarcos:
Any more context about this? It’s hard to follow what this is for. The associated PR is adding a fair amount of code of complexity and the value is not obvious

ngokevin:
Closing in favor of https://github.com/aframevr/aframe/pull/4114 , thanks for the work though!

Benefits are made more obvious in this form.

PR #3744, opened 2018-08-21
Add event binder

Adam4lexander:
https://github.com/aframevr/aframe/issues/3675

Adding in utils.bindEvent like you suggested. I had to make a small change to component.js to execute each decorator when a new component is instantiated.. In my own version I was just monkey patching this in by wrapping some core A-frame functions.

Should this also go in the A-frame documentation? I'm happy to add it in just let me know where you think it should go.


src/utils/event-decorators.js
```diff
@@ -0,0 +1,163 @@
...
+ *     console.log(evt.detail);
+ *   })
+ * })
+ *
+ * By default the function will be bound to events corresponding to its property name, in this case: 'componentchanged'.
+ * It will listen for events on its parent element, and will begin listening or end listening when init or remove is
+ * called. However this can also be configured by passing a configuration object:
+ *
+ * AFRAME.registerComponent("foo", {
+ *   whenSceneLoads: bindEvent( {
```
ngokevin: I don't think we need this customizable of an API. Should just take a function, listen to `this.el`, default to the property name, and set up on init/remove.

src/core/component.js
```diff
@@ -268,6 +269,8 @@ Component.prototype = {
       if (el.initializingComponents[this.name]) { return; }
       // Prevent infinite loop in case of init method setting same component on the entity.
       el.initializingComponents[this.name] = true;
+      // Execute decorator functions on component before init
+      executeDecorators(this);
```
ngokevin: Is this avoidable, can we do pass in `this` explicitly?

src/utils/event-decorators.js
```diff
@@ -0,0 +1,163 @@
...
+ */
+function decorate (func, decoratorFunc) {
+  return decorations.add(func, decoratorFunc);
+}
+
+/*
+ * Executes all decorated functions on the given component. You shouldn't need to call
+ * this yourself.
+ */
+function executeDecorators (component) {
```
ngokevin: If this were called on every component, we'd want to optimize it. No closures (`forEach`), no array allocations.

tests/utils/event-decorators.test.js
```diff
@@ -0,0 +1,53 @@
...
+        myEventHandler: bindEvent({
+          event: 'someevent',
+          target: 'a-scene',
+          listenIn: 'init',
+          removeIn: 'remove'
+        }, function () {
+          spy();
+        })
+      });
+      var el = this.el;
```
ngokevin: vars on top, or can use `let` / `const` in tests

src/utils/event-decorators.js
dmarcos: I agree, this API seems too high of an abstraction for the core. I would probably keep it as a 3rd party component / library

src/core/component.js
Adam4lexander: Unfortunately that's the trick to get this all working. I couldn't find any way of avoiding it.. although I thought a lot about it. In my own node package I had to monkey patch this call in by wrapping some core functions

Adam4lexander:
Understood, it is quite a high level abstraction. Now that I know that this complexity is definitely not what you want I can simplify the code a great deal if that can change your mind? Atm its written a lot more generally then is needed to handle the specific case of binding event listeners in init/remove and defaulting to the property name. Unfortunately it would still need that call to executeDecorators (or something thereto) in component.js.

I came up with this because I wanted event handling in A-Frame to look and feel more like in Unity 3D. In Unity common GameObject events like 'OnCollisionEnter', 'OnEnable' etc are mapped functions on components with the same name. Compared to addEventListener/removeEventListener/function.bind() I think its a more declarative, easier on the eyes way of managing events. Although there needs to be a bit of magic in the background to make it work.

src/utils/event-decorators.js
ngokevin: Not what I meant. I mean the extra options are not needed. The utility still belongs in core, optional to use. I use it extremely often not to have to bind methods, add event listeners up top, and have to manually remove event listeners on `pause` or `remove`, which no one ever does because it's too much work.

ngokevin:
This is a utility that can be called optionally. It's extremely useful. Rather than:

```js
{ 
  init: function () {
   this.bindMethods();
  },

  bindMethods: function () {
    this.click = this.click.bind(this);
  },

  addEventListeners: function () {
    this.el.addEventListener('click', this.click);
  },

  removeEventListeners: function () {
    this.el.addEventListener('click', this.click);
  },

  pause: function () {
   this.removeEventListeners(); 
  },

  play: function () {
    this.removeEventListeners();
  },

  remove: function () {
    this.removeEventListeners();
  },

  click: function () {

  }
}
```

We do:

```js
{
  click: utils.bindEvent(function () {

  })
}
```

ngokevin:
It could be an auto-magic option by default, though there's less magic in calling it manually.

Adam4lexander:
I'm rewriting it now, it can be way waaayy more simple. With all jargon fluff removed.

Adam4lexander:
New version is much simpler. No more jargon fluff about decorator functors. It should be more clear how it actually works now. I'm still not sure how to avoid that change in component.js, any suggestions for what you had in mind?

ngokevin:
It's explicit because you explicitly call it. It only does those behavior if you use it.

ngokevin:
Hey, sorry for the work and thanks for the time you spent. After creating the hot reloader and continuing the ideas from your API, I came up with: https://github.com/aframevr/aframe/pull/4114

Adam4lexander:
All good! Happy to help out. Your new event handler system looks great

PR #3749, opened 2018-08-23
Expanded commit hash for document-register-element dependency

5310:
**Description:**

Expanding the commit-ish to the full hash doesn't harm us (and is good practice anyway,) and by doing so it lets us far more easily resolve the dependency and actually be able to install the package under, for example, PNPM. 

See pnpm/pnpm#1127

**Changes proposed:**
- Expand the commit-ish for document-register-element to the full hash


dmarcos:
Thanks 👍

5310:
Thank you!

PR #3752, opened 2018-08-28
Fix controller reconnecting on Oculus Go / GearVR

netpro2k:
**Description:**
This fixes controllers not being able to reconnect on Oculus Go and GearVR. The bug seems to just be due to a copy/paste error. This brings these 2 in line with the otehr controller components, keeping the `controllersupdated` event bound solong as the entity is playing. This is particularly important on Oculus Go as exiting VR causes the controller to disconnect (as far as the gamepad API is concenred) and the controller will reconnect when entering VR.

**Changes proposed:**
- Support disconnecting and reconnecting Oculus Go and GearVR controllers


dmarcos:
Thanks! 👍

PR #3760, opened 2018-09-05
Add configurable loading screen.

dmarcos:
**Description:**

Loading screen to avoid blank screen while loading assets and scene render loop has not yet started. The loader animation doesn't require 60fps to mask main thread pauses.

The logic is a `pseudo-component` since it has to initialize and render while the assets load. 

The title of the page is displayed above the loading indicator.

The color of the dots, background and style of the title bar are configurable.

loader

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -1,6 +1,7 @@
 /* global Promise, screen */
 var initMetaTags = require('./metaTags').inject;
 var initWakelock = require('./wakelock');
+var loadingScreen = require('../loadingScreen');
 var re = require('../a-register-element');
 var scenes = require('./scenes');
 var systems = require('../system').systems;
@@ -85,7 +86,6 @@ module.exports.AScene = registerElement('a-scene', {
     attachedCallback: {
       value: function () {
         var self = this;
-
         // Renderer initialization
         setupCanvas(this);
         this.setupRenderer();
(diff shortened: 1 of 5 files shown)


src/core/scene/a-scene.js
```diff
@@ -507,6 +510,80 @@ module.exports.AScene = registerElement('a-scene', {
       writable: true
     },
 
+    setupLoader: {
```
ngokevin: Can place all the loader code in a module that exports functions `setupLoader(sceneEl)` to not pollute scope, not meant to be callable by user

src/core/scene/a-scene.js
```diff
@@ -507,6 +510,80 @@ module.exports.AScene = registerElement('a-scene', {
...
+      value: function () {
+        var self = this;
```
ngokevin: a little daunting to read a web of instantiations that aren't used until later in various places. need to jump up/down the function to follow.

src/core/scene/a-scene.js
```diff
@@ -507,6 +510,80 @@ module.exports.AScene = registerElement('a-scene', {
...
+        var size = getCanvasSize(this.canvas, false, this.maxCanvasSize, this.is('vr-mode'));
+        camera.aspect = size.width / size.height;
+        camera.updateProjectionMatrix();
+         // Notify renderer of size change.
+        this.renderer.setSize(size.width, size.height, false);
+      },
+      writable: true
+    },
+
+    removeLoader: {
```
ngokevin: Can also free the created geometry/materials

dmarcos:
@ngokevin This is ready to go:

- Extracted `loadingScreen` logic to module.
- Added tests.
- Now the loading screen also works on vr if `vrdisplayactivate` fires. A-Frame will always render something on link traversal while the scene is loading (unless the user disables the loader).

ngokevin:
I'd document the `document.title` behavior

PR #3761, opened 2018-09-10
added a reverseTouchDrag parameter

Utopiah:
Equivalent to reverseMouseDrag for mobile and tablet.

**Description:** added a reverseTouchDrag parameter

**Changes proposed:**
- using direction as it was used for reverseMouseDrag


dmarcos:
What about a single `reverseDrag` flag that works for both mobile and mouse?

Utopiah:
Probably better especially since I can't imagine situations where one would want to have the opposite behavior on mouse vs touch. Still being curious about "creative" or artistic usages and wanted to preserve compatibility I didn't want to "take the risk". Up to you.

ngokevin:
I think this is fine

PR #3764, opened 2018-09-13
simplify text shaders and prevent ugly alpha issues when text is small/far away in front of element with alpha (fixes #3557)

### LINKED ISSUE #3557: entity text is blurry in a-box element
**Description:**

- A-Frame Version: 0.8.0
- Platform / Device: Desktops

Not sure if this topic belongs here but I have an a-entity text inside an a-box element and the text looks blurred out 

words blurry why

These are the lines the box is in
```
<a-box id="box1" material="src: img/info.svg; transparent:true;" position="0 2 -1.8" rotation="0 0 0" depth="0" width="1" height="1" scale="0.001 0.001 0">
    <a-entity id="info" width="1" position="0 0 0.6" text="value: This is the Mac desktop computer IXD has purchased to supply their students with adequate equipment so that they are capable of accomplishing projects even if they do not have their own Mac OS based system; color:#000;">
    </a-entity>
    <a-circle id="box1-close" radius="0.06" class="clickable" material="src: img/close.svg; transparent:true;" remove-yellow position="0.4 0.4 0.6" rotation="0 0 0" scale="1 1 0"></a-circle>
</a-box>
```
I suspect the problem may be with the anti aliasing but I have already manually turned it on in my a-scene

ngokevin:
**Description:**

All of the fiddling with the alpha didn't really help and made things look worse when blending with other entities with alpha, and at distance.

Text example looks unchanged, but in certain conditions the text degrades badly. Can see it below, but it often manifests worse:

## Before

screen shot 2018-09-13 at 3 25 59 pm

## After

screen shot 2018-09-13 at 3 25 33 pm

**Changes proposed:**

- Remove alpha modifications that were done based on experiment / guess & checks.
- Remove the logic `when texel is too small, blend raw alpha value rather than supersampling`
- Remove the logic `when texel is big enough, do standard alpha test`
- Less shader branching!

### DIFF
--- a/src/shaders/msdf.js
+++ b/src/shaders/msdf.js
@@ -33,44 +33,32 @@ module.exports.Shader = registerShader('msdf', {
     '#endif',
 
     'precision highp float;',
-    // FIXME: Experimentally determined constants.
-    '#define BIG_ENOUGH 0.001',
-    '#define MODIFIED_ALPHATEST (0.02 * isBigEnough / BIG_ENOUGH)',
-    '#define ALL_SMOOTH 0.4',
-    '#define ALL_ROUGH 0.02',
-    '#define DISCARD_ALPHA (alphaTest / (2.2 - 1.2 * ratio))',
+    'uniform bool negate;',
+    'uniform float alphaTest;',
+    'uniform float opacity;',
     'uniform sampler2D map;',
     'uniform vec3 color;',
-    'uniform float opacity;',
-    'uniform float alphaTest;',
-    'uniform bool negate;',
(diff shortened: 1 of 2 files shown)


machenmusik:
I think the original experiments with simpler shader had other artifacts that were font specific.
If those are gone now (either no one uses the broken fonts anymore, or something else fixed it) then I am all for simplifying, +1 from me

ngokevin:
looks fine to me in the text example

ngokevin:
screen shot 2018-09-25 at 11 18 21 pm

PR #3767, opened 2018-09-21
don't bubble state events

ngokevin:
**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -747,7 +747,7 @@ var proto = Object.create(ANode.prototype, {
     value: function (state) {
       if (this.is(state)) { return; }
       this.states.push(state);
-      this.emit('stateadded', state);
+      this.emit('stateadded', state, false);
     }
   },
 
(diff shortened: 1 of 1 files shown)


wmurphyrd:
I use state event bubbling quite a lot for advanced interactions using super hands. For example, a selectable component which listens for state end events coming from the grabbed or clicked states of any if its children to keep track of the last selected child.

dmarcos:
What's the rational to not bubble state events?

machenmusik:
I assume it is performance, but if we're not going to bubble them (which is kinda the point of using events for that IMO), then maybe you want to consider removing and reimplementing...

ngokevin:
It was for performance and semantics. For my laptop, a stateadded or stateremoved call each take ~.1ms (1% of a 90FPS budget), didn't quite like the cursor used it, so figured I could ameliorate it a bit. I thought it didn't make sense to bubble these events when they are very single-entity-specific.

I could go either way, my full desire would be to remove the API, and just use the cursor events provided. Second would be to remove event bubbling so it doesn't fire on like 10+ ancestors. If there's disagreement, don't mind closing either.

dmarcos:
When I write that API bubbling was what made it valuable. I used it to have a bunch of entities (created dynamically) changing to a specific state (`dead`) that I wanted to get notified for. Maybe we don't bubble but emit also on the scene so one can attach a single listener? @wmurphyrd Would that work for you?

ngokevin:
Closing for now, but recommend to switch components to use events.

PR #3772, opened 2018-09-25
reduce memory allocation in component update core codepath

ngokevin:
**Description:**

Rather than create a new object each time to create a new `oldData`, just have two objects representing `oldData` and `data`. Also when updating, rather than create a new `data` object, have a `newData` object that is used for holding the new data which then gets copied into `data`. So in total, only use three objects: `oldData`, `data`, `newData`.

**Changes proposed:**
- Adapt `style-attr` implementation to be able to pass in an object for reuse instead of creating a new one each time. Also remove lots of callbacks, `Object.keys()` array creations from their code.  Also tweaked to do prettier stringifies. We manage this forked dependency now. 
- Use object pooling in core component code to reduce object allocations. Recycle them when component is removed.
  - 1. Remove object allocations on each update for deep cloning data to `oldData`.
  - 2. Remove object allocation on initialization for initial `oldData`.
  - 3. Remove object allocation on each update during `buildData` for default value copying.  
  - 4. Remove object alloaction during `buildData`, was using `Object.keys`.
- Add some helper properties to component (`isSingleProp`, `isSinglePropObject`, `isObjectBased`)
- Update our `styleParser` utils a bit to not allocate object + array + use callbacks to convert keys to camelCase, and to cache regexes.

Updated to not delete the keys when recycling an object, but just set every key to `undefined`. This means object pooling best used when objects share a common schema, in case we iterate over keys. So I create different object pools for each type of component, since their object keys will be the same. I also update schema parser to ignore undefined keys.

## Before

screen shot 2017-10-03 at 7 16 43 pm

## After

**Improvements:** See the `updateProperties` and `updateCachedAttrValue` no longer on the list of frequent memory allocators. 

screen shot 2017-10-03 at 7 17 57 pm

- [x] Still need to polish/comment a bit more (and try to remove the clones for default values). But up and working again.
- [x] Tests

### DIFF
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -79,6 +79,11 @@ module.exports.Component = registerComponent('camera', {
       // Camera disabled. Set camera to another camera.
       system.disableSpectatorCamera();
     }
+
+    this.savedPose = {
+      position: el.getAttribute('position'),
+      rotation: el.getAttribute('rotation')
+    };
   },
 
   /**
--- a/src/components/camera.js
+++ b/src/components/camera.js
@@ -79,11 +79,6 @@ module.exports.Component = registerComponent('camera', {
       // Camera disabled. Set camera to another camera.
       system.disableSpectatorCamera();
     }
-
-    this.savedPose = {
-      position: el.getAttribute('position'),
-      rotation: el.getAttribute('rotation')
-    };
   },
 
   /**
--- a/src/components/geometry.js
+++ b/src/components/geometry.js
@@ -61,12 +61,10 @@ module.exports.Component = registerComponent('geometry', {
 
   /**
    * Update geometry component schema based on geometry type.
-   *
-   * @param {object} data - New data passed by Component.
    */
   updateSchema: function (data) {
+    var currentGeometryType = this.oldData && this.oldData.primitive;
     var newGeometryType = data.primitive;
-    var currentGeometryType = this.data && this.data.primitive;
     var schema = geometries[newGeometryType] && geometries[newGeometryType].schema;
 
     // Geometry has no schema.
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -53,10 +53,16 @@ module.exports.Component = registerComponent('material', {
   },
 
   updateSchema: function (data) {
-    var newShader = data.shader;
-    var currentShader = this.data && this.data.shader;
-    var shader = newShader || currentShader;
-    var schema = shaders[shader] && shaders[shader].schema;
+    var currentShader;
+    var newShader;
+    var schema;
+    var shader;
+
+    newShader = data && data.shader;
+    currentShader = this.oldData && this.oldData.shader;
+    shader = newShader || currentShader;
+    schema = shaders[shader] && shaders[shader].schema;
+
     if (!schema) { error('Unknown shader schema ' + shader); }
     if (currentShader && newShader === currentShader) { return; }
     this.extendSchema(schema);
--- a/src/components/obj-model.js
+++ b/src/components/obj-model.js
@@ -21,12 +21,16 @@ module.exports.Component = registerComponent('obj-model', {
   update: function () {
     var data = this.data;
     if (!data.obj) { return; }
-    this.remove();
+    this.resetMesh();
     this.loadObj(data.obj, data.mtl);
   },
 
   remove: function () {
     if (!this.model) { return; }
+    this.resetMesh();
+  },
+
+  resetMesh: function () {
     this.el.removeObject3D('mesh');
   },
 
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -16,7 +16,9 @@ var warn = utils.debug('core:component:warn');
 
 var aframeScript = document.currentScript;
 var upperCaseRegExp = new RegExp('[A-Z]+');
-var objectPool = utils.objectPool.createPool();
+
+// Object pools by component, created upon registration.
+var objectPools = {};
 
 /**
  * Component class definition.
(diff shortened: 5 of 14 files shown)


ngokevin:
Ready to go. Did another test with the /performance/animation-set-attribute stress test to measure memory. These recordings were taken over **4 seconds**. We save the following allocation calls:

- buildData (component) (~400KB)
- getKeyValueChunks (style-parser) (~160KB)
- cloneData (~80KB)
- updateProperties (~28KB)

With the particular stress test, it does lose out on ~5FPS (20FPS vs 15FPS), but it's also an unrealistic example since we wouldn't use setAttributes that hard and often for animations. Dodging GC is more important because GC will lose you entire frames.

# Before 

before

# After

after

arpu:
@ngokevin i would like test this change, how do you generate the memory statistics?

src/components/camera.js
```diff
@@ -79,6 +79,11 @@ module.exports.Component = registerComponent('camera', {
       // Camera disabled. Set camera to another camera.
       system.disableSpectatorCamera();
     }
+
+    this.savedPose = {
+      position: el.getAttribute('position'),
```
dmarcos: I think we have to allocate new objects here. You will be holding a reference to the entity shared objects that will change.

ngokevin: I just removed it, it wasn't used.

ngokevin:
browser memory tools

dmarcos:
It would be nicer if this PR could
be broken in smaller PRs. It touches a lot
of files and it’s hard to follow and see implications

src/components/obj-model.js
```diff
@@ -21,12 +21,16 @@ module.exports.Component = registerComponent('obj-model', {
   update: function () {
     var data = this.data;
     if (!data.obj) { return; }
-    this.remove();
+    this.resetMesh();
```
dmarcos: Unrelated to the purpose of the PR?

src/components/material.js
```diff
@@ -53,10 +53,16 @@ module.exports.Component = registerComponent('material', {
   },
 
   updateSchema: function (data) {
-    var newShader = data.shader;
-    var currentShader = this.data && this.data.shader;
-    var shader = newShader || currentShader;
-    var schema = shaders[shader] && shaders[shader].schema;
+    var currentShader;
+    var newShader;
```
dmarcos: unrelated to PR?

src/components/geometry.js
```diff
@@ -61,12 +61,10 @@ module.exports.Component = registerComponent('geometry', {
 
   /**
    * Update geometry component schema based on geometry type.
-   *
-   * @param {object} data - New data passed by Component.
    */
   updateSchema: function (data) {
+    var currentGeometryType = this.oldData && this.oldData.primitive;
```
dmarcos: unrelated to PR?

dmarcos:
styleParser.js and component.js changes could be separate PRs

src/components/obj-model.js
ngokevin: fixes a test

src/components/material.js
ngokevin: fixes test, checks for data / oldData existence

src/components/geometry.js
ngokevin: yes, it checks for oldData rather than data

ngokevin:
It's hard and effort to break up. They both relate to the object pooling and both tie together. There's only two major changes:

- remove allocations from style parser by passing an object in (managed with object pool)
- remove allocations from component.js by removing the cloning (managed with object pool)

I detail listed all the changes in the original description

arpu:
this breaks my little app with 

body.js:122 Uncaught TypeError: Cannot read property 'toUpperCase' of undefined at initBody (body.js:122) at HTMLElement.<anonymous> (a-node.js:300) at a-node.js:131 initBody @ body.js:122 emit @ a-node.js:300 (anonymous) @ a-node.js:131 Promise.then (async) load @ a-node.js:128 load @ a-entity.js:255 (anonymous) @ a-scene.js:591 setTimeout (async) play @ a-scene.js:590 attachedCallbackPostCamera @ a-scene.js:128 (anonymous) @ a-scene.js:102 emit @ a-node.js:300 checkUserCamera @ camera.js:98 (anonymous) @ camera.js:58 emit @ a-node.js:300 setObject3D @ a-entity.js:166 init @ camera.js:28 updateProperties @ component.js:315 updateComponent @ a-entity.js:489 updateComponents @ a-entity.js:463 (anonymous) @ a-entity.js:259 (anonymous) @ a-node.js:130 Promise.then (async) load @ a-node.js:128 load @ a-entity.js:255 (anonymous) @ a-entity.js:84 emit @ a-node.js:300 (anonymous) @ a-node.js:131 Promise.then (async) load @ a-node.js:128 load @ a-assets.js:83 (anonymous) @ a-assets.js:70 setTimeout (async) value @ a-assets.js:66


https://github.com/donmccurdy/aframe-physics-system/blob/master/src/components/body/body.js#L64

ngokevin:
can you log this.data, and more information if you're updating it, initializing it...a live example would be great

arpu:
here is a live example https://codesandbox.io/s/5k1wxylxkx

ngokevin:
Thanks, possible to have one with the physics as an unminified project file so I can poke around in it?

arpu:
https://codesandbox.io/s/vv6jm1l06l

with unminified physics

arpu:
@ngokevin  should i open a new bugreport?

ngokevin:
thanks sure

PR #3776, opened 2018-09-27
Update a-cylinder.md

nbackous:
Unless my understanding of basic geometry is incorrect (I always reserve the right to be incorrect), a hexagon has ~7~  **6** segments.

**Description:**

**Changes proposed:**
-
-
-


dsinni:
Hexagon actually has 6; heptagon has 7.

nbackous:
@dsinni lol, uh, yeah, whoops

ngokevin:
thx

PR #3792, opened 2018-10-08
[gltf-model] Add system to configure and enable Draco compression.

### LINKED ISSUE #3640: New GLTF Loader for GLTF2 - unable to pass DracoLoader instance
Hi,
I have draco compressed GLTF images which needs to be loaded using aframe. When trying to load it using the new gltf-model component I am getting "components:gltf-model:warn THREE.GLTFLoader: No DRACOLoader instance provided." error. I hope this is because there is no option to specify the dracoloader which can be passed as argument to the threejs GLTFLoader -> setDRACOLoader function. Please help...

donmccurdy:
Fixes #3640.

Adds a `gltf-model` system, through which compression (via Draco) can be enabled. `THREE.DRACOLoader` loads a WASM or JS decoder, choosing automatically based on browser capabilities. The decoders must be placed in a directory accessible to the application. Example:

```html
<a-scene stats gltf-model="dracoDecoderPath: /lib/decoder/;">
  <a-entity gltf-model="foo.glb" ></a-entity>
</a-scene>
```

If the change and the API look ok I'll update docs.

### DIFF
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -10,8 +10,12 @@ module.exports.Component = registerComponent('gltf-model', {
   schema: {type: 'model'},
 
   init: function () {
+    var dracoLoader = this.system.getDRACOLoader();
     this.model = null;
     this.loader = new THREE.GLTFLoader();
+    if (dracoLoader) {
+      this.loader.setDRACOLoader(dracoLoader);
+    }
   },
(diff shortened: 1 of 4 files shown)


dmarcos:
Yeah 👍 Thanks for stay on top of all things glTF

PR #3801, opened 2018-10-09
Remove trackpaddown event from the downEvents to not monopolize all buttons on GearVR and Oculus Go

dmarcos:


machenmusik:
The intent here is: if controller has trigger, use that, else use trackpad (e.g. on daydream) correct?

dmarcos:
correct. We map click to trigger button on all controllers except daydream that does not have one. This PR makes it more consistent and frees the trackpad for other input on Oculus Go and Gear VR.

PR #3805, opened 2018-10-12
support preprocessing sound, remove forEaches

ngokevin:
**Description:**

Allow preprocessing the sound before playing it.

Lets me position the sound, or depitch / detune it dynamically when I call playSound dynamically. This is so I can have a sound pool container entity that is not explicitly attached to a bunch of individual children entities.

**Changes proposed:**
- Allow for doing something to the audio object / sound buffer before playing it in sound component.
- Remove forEaches.
- No bubble the sound events. Sound pools emit fairly often (e.g., bullet sounds).

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -36,7 +36,10 @@ module.exports.Component = registerComponent('sound', {
 
   update: function (oldData) {
     var data = this.data;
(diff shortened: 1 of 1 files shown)


ngokevin:
👍

PR #3813, opened 2018-10-16
improve grabbing cursor perf (fixes #3790)

### LINKED ISSUE #3790: Don't set grabbing class on the body. Causing 2ms style recalculations.
**Description:**

For look-controls, setting class on the body on each click, causing 2ms style recalculations. We should just set the style directly.

krzysztof-grzybek:
Fixes #3790

Hi @ngokevin , I added changes, but I didn't spot any performance improvement (I checked it in Chrome performance devtools). Maybe I didn't understand something?

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -5,7 +5,6 @@ var bind = utils.bind;
 var PolyfillControls = require('../utils').device.PolyfillControls;
 
 // To avoid recalculation at every mouse movement tick
-var GRABBING_CLASS = 'a-grabbing';
 var PI_2 = Math.PI / 2;
 
 var checkHasPositionalTracking = utils.device.checkHasPositionalTracking;
@@ -248,7 +247,7 @@ module.exports.Component = registerComponent('look-controls', {
 
     this.mouseDown = true;
     this.previousMouseEvent = evt;
-    document.body.classList.add(GRABBING_CLASS);
+    this.showGrabbingCursor();
 
     if (this.data.pointerLockEnabled && !this.pointerLocked) {
       if (canvasEl.requestPointerLock) {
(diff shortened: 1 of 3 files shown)


src/components/look-controls.js
```diff
@@ -259,12 +258,34 @@ module.exports.Component = registerComponent('look-controls', {
     }
   },
 
+  /**
+   * Shows grabbing cursor on scene
+   */
+
```
ngokevin:
```suggestion

```

src/components/look-controls.js
```diff
@@ -259,12 +258,34 @@ module.exports.Component = registerComponent('look-controls', {
...
+      'cursor: -moz-grabbing;' +
+      'cursor: -webkit-grabbing;';
+    var newStyle = currentStyle.replace(/;$/, '') + ';' + cursorStyle;
+    this.el.sceneEl.canvas.setAttribute('style', newStyle);
+  },
+
+  /**
+   * Hides grabbing cursor on scene
+   */
+
```
ngokevin:
```suggestion

```

src/components/look-controls.js
```diff
@@ -259,12 +258,34 @@ module.exports.Component = registerComponent('look-controls', {
     }
   },
 
+  /**
+   * Shows grabbing cursor on scene
+   */
+
+  showGrabbingCursor: function () {
+    var currentStyle = this.el.sceneEl.canvas.getAttribute('style') || '';
```
ngokevin: Just this:

```js
this.el.sceneEl.canvas.style.cursor = 'grabbing';
this.el.sceneEl.canvas.style.mozCursor = 'grabbing';
this.el.sceneEl.canvas.style.webkitCursor = 'grabbing';
```

ngokevin:
I saw in devtools taking time for "Apply Style Changes".

src/components/look-controls.js
krzysztof-grzybek: I'm afraid it won't work.. Take a look, what has to be prefixed is css property value, not property name. On the other hand, are You sure that those prefixes are needed? I tested it on latest chrome and firefox and it works without prefix, also take a look [here](https://caniuse.com/#search=grabbing).

ngokevin: Yeah can just get rid of prefixes. It's not that big a feature anyways.

Issue #3793, opened 2018-10-08
Consider removing collada-model.


donmccurdy:
I'm not sure we need collada-model in the core library any longer. The COLLADA2GLTF converter should be at least as robust as THREE.ColladaLoader at this point, and among three.js users COLLADA use appears to be less common than glTF, OBJ, or FBX: https://discourse.threejs.org/t/survey-which-is-your-primary-3d-format-for-three-js-apps/3015.

donmccurdy:
^If we decide to remove it, I'll likely add it to the aframe-extras loaders for a those who still want it.

dmarcos:
Happy to remove it for 0.9.0

1d10t:
> ^If we decide to remove it, I'll likely add it to the aframe-extras loaders for a those who still want it.

Hello, Don
will you add collada loader to aframe-extras repo ? at least i still need it

ngokevin:
You can just include the three.js COLLADA Loader in your project as a script tag, then call it (`THREE.ColladaLoader.load`) or copy the old component into your project. Until it's added.

donmccurdy:
Added in A-Frame Extras v6.0.0.

ngokevin:
noice

olivierFalloff:
Hello
Before thank you for your amazing work!!!
Hello
I have been using a-frame since 2018 to show my scenographies to the directors I work for.I used the colladas files for my objects and everything worked very well and my textures were as I wanted them, a little dull, my materials came from bake of textures. But since the webxr appeared it is no longer possible to use. I have tried the Gltf files but the GLTF format is less beautiful, it shines, it is heavier and there is too much aliaising ...
I wrote an addon for Blender which automatically wrote all the html lines in a-frame and now it's trash, I no longer have access to my collada files !!! how many hours wasted ...
Why do we have to reinvent the hammer in IT every 5 years ??? Please keep the things working thanks.

donmccurdy:
@olivierFalloff We have made a legacy COLLADA loader available for those who need it, in https://github.com/n5ro/aframe-extras. If you are having trouble getting that legacy COLLADA loader to work, or getting materials to appear the way you want in more modern and efficient formats like glTF, I would recommend starting a new question on Stack Overflow, or asking in the A-Frame Slack, where you can share enough information for someone to reproduce the issue and help you debug it.

olivierFalloff:
Hello
[...]

PR #3817, opened 2018-10-18
[collada-model] Deprecate 'collada-model'.

donmccurdy:
Per https://github.com/aframevr/aframe/issues/3793. 

What do you think of just marking it deprecated for a release before we remove anything? Or better to just go for it?

### DIFF
--- a/docs/components/collada-model.md
+++ b/docs/components/collada-model.md
@@ -9,6 +9,12 @@ examples: []
 
 The collada-model component loads a 3D model using a [COLLADA][wiki-collada] (.DAE) file.
 
+> **DEPRECATED**: The `collada-model` component is deprecated. Consider
+converting models with [COLLADA2GLTF][collada2gltf] and using `gltf-model`
+instead.
+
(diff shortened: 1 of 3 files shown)


ngokevin:
I'm okay just going for it. I haven't seen COLLADA use, and many of the popular asset places don't export them now. Also need to remove from examples.

ngokevin:
I think realistically 2-3 weeks.

dmarcos:
Happy to remove it. 0.9.0 is a good heads up before 1.0.0 when we should consolidate to glTF.

ngokevin:
@donmccurdy Want to execute the removal? Also want to have a go at reviewing some of my recent PRs?

donmccurdy:
I'm working on the `collada-model` removal now. I think it will require merging something like #3757 to keep consistent colors, as we replace the `.dae` examples with `.glb`. #3757 is still waiting for me to make changes at this point too I think. 😅

ngokevin:
Going with https://github.com/aframevr/aframe/pull/3866

Issue #3838, opened 2018-11-02
`animation` `startEvents` not triggering animations in `master`


dsinni:
**Description:**
Using `animation="startEvents: ...;"` doesn't work in `master` (`843a67f`) as it did previously with `0.8.2` and the `aframe-animation-component`. The animation isn't triggered when the event fires.

- A-Frame Version: `master` (`843a67f`)
- Platform / Device: Windows 10 
- Reproducible Code Snippet or URL:
```html
<a-box color="#fff"
       position="0 1 -3"
       animation="property: object3D.position.x; from: 0; to: 10; startEvents: click;">
</a-box>
```

CodePen: https://codepen.io/dansinni/pen/NEKrja

Switching to `0.8.2` with the separate component works as expected.

ngokevin:
Set `.objects` on your raycaster. The box is not set up as a raycast target and perhaps some asynchrony  from using a-box over a-entity. (e.g., `<a-cursor raycaster="objects: a-box">` (mentioned in https://aframe.io/docs/0.8.0/introduction/best-practices.html#performance).

dsinni:
Ah, okay. I knew it was a best practice to set raycaster objects, but I didn't know that it was now mandatory. 

I left it out since the scene was ridiculously basic.

So, that is the case then? It's now mandatory to set raycaster objects?

ngokevin:
It should be. But I'm looking why the raycaster didn't pick it up in the first place, but not problem with animation.

ngokevin:
I have a fix coming to fix raycasters when defaulting to intersecting everything in scene. The wrapper entity threw it off.

Issue #3818, opened 2018-10-20
cursor: "rayOrigin: mouse" detects wrong element


naraesk:
**Description:**

I have a visualization with many occluded elements and onclick event handlers for every element. ```cursor="rayOrigin: mouse"``` is used to detect mouse clicks. But sometimes the event handler of an occluded element is called: box A occludes box B. When clicking on box A the event handler of box B is called, although it is hidden completely behind box A. Most of time it works corect (event handler of A is called). Just click around a bit in the attched visualization, sooner or later you will notice the wrong behaviour.

- A-Frame Version: 0.8
- Platform / Device: Desktop/Chromium
- Reproducible Code Snippet or URL:
[mouseray.zip](https://github.com/aframevr/aframe/files/2498070/mouseray.zip)

arpu:
but there is problem with entities within groups (nested in an entity)  since this commit see commit user message 

https://github.com/aframevr/aframe/commit/8809e7fd5eccd4c4ce76892f4a5673b911dfd60f

ComputerTime:
On this latest version of master, i can't get the objects array in a
raycaster to not be equal to an empty string or null, which means the whole
scenes children array is being used, so it seems to be fine now.

dmarcos:
@ComputerTime Thanks for verifying. Closing and will reopen if the problem comes back.

PR #3840, opened 2018-11-02
fix raycaster not grabbing all entities when objects is not set

ngokevin:
**Description:**

found by @dsinni. r? @donmccurdy

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -388,7 +388,7 @@ module.exports.Component = registerComponent('raycaster', {
    * so that non-recursive raycasting remains useful.
    *
    * Only push children defined as component attachemnts (e.g., setObject3D),
-   * not actual children in the scene graph hierarchy.
+   * NOT actual children in the scene graph hierarchy.
    *
    * @param  {Array<Element>} els
    * @return {Array<THREE.Object3D>}
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -177,7 +177,7 @@ module.exports.Component = registerComponent('raycaster', {
     // If objects not defined, intersect with everything.
     els = data.objects
       ? this.el.sceneEl.querySelectorAll(data.objects)
-      : this.el.sceneEl.children;
+      : this.el.sceneEl.querySelectorAll('*');
     this.objects = this.flattenObject3DMaps(els);
     this.dirty = false;
   },
@@ -401,7 +401,7 @@ module.exports.Component = registerComponent('raycaster', {
     // Push meshes and other attachments onto list of objects to intersect.
     objects.length = 0;
     for (i = 0; i < els.length; i++) {
-      if (els[i].object3D) {
+      if (els[i].isEntity && els[i].object3D) {
         for (key in els[i].object3DMap) {
           objects.push(els[i].getObject3D(key));
         }
--- a/docs/components/raycaster.md
+++ b/docs/components/raycaster.md
@@ -50,19 +50,19 @@ AFRAME.registerComponent('collider-check', {
 
 ## Properties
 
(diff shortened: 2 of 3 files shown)


src/components/raycaster.js
```diff
@@ -168,7 +168,7 @@ module.exports.Component = registerComponent('raycaster', {
     // If objects not defined, intersect with everything.
     els = data.objects
       ? this.el.sceneEl.querySelectorAll(data.objects)
-      : this.el.sceneEl.children;
+      : this.el.sceneEl.querySelectorAll('*');
```
donmccurdy: By default raycasting is recursive, so `querySelectorAll('*')` will check nested entities multiple times. What issue does this fix? The current behavior does not look obviously wrong to me.

dsinni: @donmccurdy: See #3838.

`animation` `startEvents` are not working when `raycaster` `objects` are not defined. I haven't yet tested further to see if it was limited to `animation`, though,

donmccurdy: Hm ok. Maybe I've misread and each mesh is only raycast against once, then? If so `flattenObject3DMaps` could probably use a bit clearer documentation and tests.

dsinni: Okay, just tested this further and it appears that it's not limited to `animation` and likely _is_ a general raycasting issue.

I usually wrap the scene in a "stage" entity, save for the camera, which stays outside of it. This allows me to translate the scene if need be.

I'm just noticing that simply having a parent wrapper is causing the issue, which was not the case previously.

You can see it demonstrated here: https://glitch.com/~imported-wallaby

If you try clicking any of the entities, nothing happens. There should be an animation for each, excluding the plane, and the console should log the event.

If you comment out the #stage entity, it works as expected.

arpu: maybe looks like more this problem https://github.com/aframevr/aframe/issues/3818 @donmccurdy  reopen the bugreport?

donmccurdy: Given this scene:

```html
<a-scene>
  <a-entity id="room1" gltf-model="room1.glb">
    <a-box/>
    <a-box/>
    <a-box/>
  </a-entity>
  <a-entity id="room2" gltf-model="room2.glb">
    <a-box/>
    <a-box/>
    <a-box/>
  </a-entity>
</a-scene>
```

`scene.children` will give a list of the outer "room" wrappers, whereas `querySelectorAll('*')` will list those wrappers _and_ each of their children:

[ room1, room2, box1, box2, ... ]

If we passed `room1.object3D` directly to the raycaster in recursive mode, it would be hit-testing each box multiple times. I think the `flattenObject3DMaps` implementation avoids that. If this is the tree structure:

```
- room1.object3D
  - mesh
  - box1.object3D
  - box2.object3D
  - ...
```

... it should only add the mesh, which has no children, to the objects list. But that is a bit hard to confirm from the current documentation and tests, and should probably be more explicit.

arpu: dont know if this helps here but the sample from @dsinni  works with aframe A-Frame Version: 0.8.2 (Date 2018-06-28, Commit #94c6b33)
https://guttural-sagittarius.glitch.me/  ( i can see the console events)

donmccurdy: Yeah to be specific — I'm pretty sure this PR is correct and fixes the issue, but we should add a unit test ensuring we don't raycast the same mesh twice when it's nested under another entity.

dmarcos: It sounds reasonable to have a test to ensure there are no redundant checks.

ngokevin:
added test. the behavior was described internally at: https://github.com/aframevr/aframe/blob/master/src/components/raycaster.js#L390 and i updated doc.

Issue #3846, opened 2018-11-06
Some scenes don't load on latest version of Chrome on iOS


thfranco:
I have an issue on the latest version of Chrome on iOS where my scenes are all blank. This works fine on Safari.
It seems that the aframe scripts are not even loaded.
This is also reproducible with the aframe samples, but some are still working.

- A-Frame Version: 0.8.2
- Platform / Device: iOS 12.1, iPhone SE, Chrome 70.0.3538.75
- Reproducible URLs: 

https://aframe.glitch.me/ and https://aframe.io/examples/showcase/sky/
Stays blank on chrome, loads on safari

https://aframe.io/examples/showcase/360-image-gallery/
Works on both chrome and safari

Chrome:
![image](https://user-images.githubusercontent.com/1555766/48056104-5f5a1400-e1b1-11e8-92f2-fa3abbeac910.png)

Safari: 
![image](https://user-images.githubusercontent.com/1555766/48056119-66812200-e1b1-11e8-9724-ea59c91965aa.png)

dmarcos:
Is there any way to get console logs? Something has changed on Chrome side but hard to tell with more info.

thfranco:
I debugged on windows with weinre.

I'm not able to see any console logs, like the all the script loadings:
A-Frame Version: 0.8.2
WebVR Polyfill Version: ^0.10.5
..

I successfully saw some logs on some other pages, but not on the pages that were blank. It might be linked to the weinre hook, so I'm not sure if aframe scripts are even loaded or not.
But I definitely didn't succeed to catch any error logs.

i3games:
I can reproduce on an iPad (same versions as above). A workaround is to "Request Desktop Site" - displays both scenes correctly.

thfranco:
Thanks for the tip, requesting desktop site is indeed displaying correctly, but it loses all the device motion events.
The scene won't move with phone orientation, and if I try to go to VR mode it just freezes the page.

But this is a good start to investigate, I tried to play with the user agent but didn't succeed to get motion events working.

dmarcos:
Just fixed on master.

Issue #3788, opened 2018-10-05
Uncaught TypeError: Cannot set property 'BUFFER_SCALE' of undefined


BoryaMogila:
**Description:**

- A-Frame Version: 0.8.2
- Platform / Device: Windows 10 / Google Chrome 69
- Reproducible Code Snippet or URL:

Sometimes this code fail at aframe/src/index.js
`if (utils.device.isIOSOlderThan10(window.navigator.userAgent)) {
  window.WebVRConfig.BUFFER_SCALE = 1 / window.devicePixelRatio;
}`
Not

benjaminleonard:
I get the same error on Chrome on a iPhone 8 running iOS 12.0.1.
Works fine in mobile Safari

benjaminleonard:
This is fixed for me now

dmarcos:
@benjaminleonard thanks for letting us know

PR #3852, opened 2018-11-08
Change the isIOSOlderThan10 so it doesn't catch Chrome Webview

shaun1x:
**Description:**

The `*` in the regex was catching the UA of the Chrome Webview on iOS. 

**Changes proposed:**
- Check for an underscore after the version number

** Notes **

There might be a better regex to accomplish this.


shaun1x:
This should fix #3846 and #3788

PR #3859, opened 2018-11-12
have mixinUpdate wait for entity to load first

ngokevin:
**Description:**

Fix culmination of weird cases when scene was still initializing, components were still initializing, node wasn't fully attached (so setAttribute wasn't fully overridden), and mixin update was called on a still initializing component, causing issues if init/update depended on el.sceneEl.


dmarcos:
tests fail

ngokevin:
fixed

PR #3860, opened 2018-11-12
When creating custom controls you need to explicitly define the camera

codeinabox:
**Description:**

I found when I was creating custom controls I had some issues. I initially tried the following

```html
<a-entity touch-controls></a-entity>
```

But when I explicity added the camera my controls worked fine

```html
<a-entity camera touch-controls></a-entity>
```

**Changes proposed:**

Add to caveats section of the documentation

### DIFF
--- a/docs/components/look-controls.md
+++ b/docs/components/look-controls.md
@@ -59,3 +59,9 @@ replace the `require` statements with A-Frame globals (e.g.,
 If you want to create your own component for look controls, you will have to
 copy and paste the HMD-tracking bits into your component. In the future, we may
 have a system for people to more easily create their controls.
+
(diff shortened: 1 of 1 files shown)


dsinni:
This isn't really a caveat so much as just how A-Frame works.

If you want a component to affect the camera, it needs to be attached to the `camera` entity or a parent entity. The same goes for `wasd-controls`, `look-controls`, etc.

In most cases, you're going to want to override the default camera configuration anyway.

Attaching `touch-controls` to an empty entity will only give you control over an empty entity, if it does anything at all.

ngokevin:
Yeah, controls just move position / rotation. If you want to have that effect on the camera, then you need to attach to a camera. Just like in games, controls can move different objects, not always the first-person camera, they are decoupled

PR #3867, opened 2018-11-15
tracked-controls allow matrixautoupdate for camera rig since matrix is decomposed anyways

ngokevin:
**Description:**

Seems like the matrix auto update disabled would prevent like controllers following the camera if the camera rig was moved. I don't see need to disable autoupdate since the matrix calculations are flushed with decompose.

To test:

1. Have a camera rig with controllers.
2. Go into VR with Vive/Rift.
3. Move camera rig position via command line.

The controllers don't follow you.

**Changes proposed:**
- Enabled tracked-controls matrixAutoUpdate
- Was hard to write a test since would have to run the three.js renderer to test matrix auto update logic.


ngokevin:
Actually, needs to only obey the position, not rotation.

ngokevin:
Okay, this is good to review. Tested a couple times and found these weren't necessary, and blocked camera rig transformation rom applying to controllers.

PR #3870, opened 2018-11-16
Fix vrDevice not initialized on polyfilled devices due to utils/devis running before the WebVRPolyill is initialized

dmarcos:
Polyfill is broken in `master` due to change in initialization order. `utils/devices` queries for available headsets and it misses the polyfilled vrDevice because WebVRPolyfilled is initialized later. This PR changes initialization order, bumps the polyfill and moves function out of `devices` needed to initialize the polyfill. Tested on:

- Pixel 2, Chrome 66
- iPhone X, iOS 12.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "style-attr": "^1.0.2",
     "three": "0.95.0",
     "three-bmfont-text": "^2.1.0",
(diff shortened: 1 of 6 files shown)


ngokevin:
Fix test then looks good

PR #3875, opened 2018-11-21
WebXR Support

dmarcos:
Add WebXR support along WebVR and Cardboard. `tracked-controls` is now a trivial component that sets either `tracked-controls-webvr` or `tracked-controls-webxr` depending on the available API.  Easier to maintain and easier to deprecate the webvr component when the time comes. A reference to the controller, axis and buttonStates is set on `tracked-controls` to avoid breaking some 3rd party components that access them directly. 

Moved THREE dependency to temporary branch until https://github.com/mrdoob/three.js/pull/15290 and https://github.com/mrdoob/three.js/pull/15242 merge.

Tested on:

- Supermedium + Oculus
- Firefox 63 desktop (WebVR), Oculus Rift, Samsung Odyssey+ 
- Chrome 70 and 72 desktop (WebXR enabled via chrome://flags), Oculus Rift, Samsung Odyssey+
- Pixel 2, Chrome 70 and 72 (WebXR enabled via chrome://flags)
- Lenovo Mirage, Chrome 70 and 72 (WebXR enabled via chrome://flags)

Pending:

- [x]  `tracked-controls` and `tracked-controls-webxr` tests.

- [x] Controller events. Current WebXR spec only emits one event / action as if the controller has a single button. The XRInputSource API is going to be revisited in the January F2F meeting and tracked-controls-webxr will evolve with it. As an interim solution Chrome has a WebXR Gamepad Support that provides a reference to a Gamepad in the InputSource object.

- [x] Testing on Daydream devices


jsantell:
Rad! Can the [webxr-polyfill](https://github.com/immersive-web/webxr-polyfill) help at all here?

dmarcos:
@jsantell Thanks for chipping in here. Does the `webxr-polyfill` have any advantage? I thought that the `webvr-polyfill` might be enough. If there's no native WebVR or WebXR APIs we polyfill WebVR. We will have to support both APIs  for at least a year.

dmarcos:
@jsantell thanks for the info. Super appreciated.

jsantell:
@dmarcos no problem! I haven't evaluated all the post-TPAC WebXR changes yet, and for the next few months until thing stabilize, not sure what to do with the polyfill in the near term, but maybe something to bring up at the next IWWG call, webxr-polyfill support plan/goals etc., because the current state of WebXR is only available in Chrome, and in 3 months, it sounds like supporting Chrome 71ish and Chrome ~77 (or whenever WebXR 1.0 is slated, or looking rather stable) will be a lot of work, and low value considering the current iteration of WebXR will be mostly out of circulation at that point

src/components/look-controls.js
```diff
@@ -182,22 +185,36 @@ module.exports.Component = registerComponent('look-controls', {
...
+    var posePosition = this.posePosition;
+    var poseRotation = this.poseRotation;
+    var poseScale = this.poseScale;
+
+    // WebXR API updates applies headset pose to the object3D matrixWorld internally.
+    // Reflect values back on position, rotation, scale so setAttribute returns expected values.
+    if (sceneEl.is('vr-mode') && utils.device.isWebXRAvailable) {
+      poseMatrix.elements = sceneEl.renderer.vr.getCameraPose().poseModelMatrix;
+      poseMatrix.decompose(posePosition, poseRotation, poseScale);
+      object3D.position.copy(posePosition);
```
ngokevin: posePosition, poseRotation variables not needed. Just decompose right into the object3D position and rotation. poseScale can be defined in the IIFE scope.

src/components/look-controls.js
```diff
@@ -326,6 +343,8 @@ module.exports.Component = registerComponent('look-controls', {
    */
   onEnterVR: function () {
     this.saveCameraPose();
+    this.el.object3D.position.set(0, 0, 0);
+    this.el.object3D.updateMatrix();
```
ngokevin: updateMatrix needed?

src/utils/tracked-controls.js
```diff
@@ -1,3 +1,4 @@
+var isWebXRAvailable = require('./device').isWebXRAvailable;
```
ngokevin: `hasWebXR` might be a nice shorter name since it's used everywhere

tests/components/vive-focus-controls.test.js
```diff
@@ -30,7 +30,7 @@ suite('vive-focus-controls', function () {
       var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
       var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
 
-      el.sceneEl.systems['tracked-controls'].controllers = [];
```
ngokevin: Would be nice if tracked-controls adopted all the properties/methods from `tracked-controls-whatever`. Would maintain backwards compatibility. I have a lot of components that read `axis` and would be nice if they didn't have to check the existence of either component. Both webvr and webxr have the `axis` property

dmarcos: there's no axis and won't likely be one in the WebXR API. Components will have to target the minimum common denominator (triggerdown event) or do component / API detection. Would a flag `isWebVR` / `isWebXR` in `tracked-controls` help or better just one global boolean like `isWebXRAvailable`?

dmarcos: For your components you will have to do feature detection already even with the `controller` reference in `tracked-controls`. WebVR gives you a Gamepad and WebXR an InputSource

src/utils/tracked-controls.js
ngokevin: hasWebXR not hasWebVR?

tests/components/vive-focus-controls.test.js
dmarcos: For your components probably the easiest is to do A-Frame version detection and access `tracked-controls` for A-Frame < 0.9.0 and `tracked-controls-webvr` >= 0.9.0. The rest of your code should remain the same.

dmarcos: An alternative is to start making a more abstract API to retrieve necessary info without accessing internal variables. A bit worry to over do it prematurely.

ngokevin:
good work

ngokevin:
One thing I would test is that in VR mode, children of the camera still get tracked. I saw matrixAutoUpdate was disabled, so the camera children trying to apply parent matrix might not update correctly.

remkoboschker:
I think I have the issue mentioned by @ngokevin that the children of the camera component are not tracking. I have a scene with a camera with look-controls rotation a sphere with bands surrounding it to make rotations more visible. There is a mirror in the scene and I cannot see myself move. But only in vr mode and only on the Oculus go. It works fine on Chrome 72.

Issue #3762, opened 2018-09-11
is yOffset property disabled in text component?


adrixp:
Hi,

When I try to use yOffset in text component nothing happens.

- A-Frame Version: 0.8.2
- Platform / Device: web (Chrome / Firefox) / Windows, Linux

Thanks and Greetings,
Adrian

PR #3886, opened 2018-11-29
Fixed yOffset not being used to change y position of text

Pluto8195:
addresses issue mentioned in  #3762

simply included (+ data.yOffset) to mesh position of text
`mesh.position.y = y * textScale + data.yOffset;`

### DIFF
--- a/src/components/text.js
+++ b/src/components/text.js
@@ -328,7 +328,7 @@ module.exports.Component = registerComponent('text', {
 
     // Position and scale mesh to apply layout.
     mesh.position.x = x * textScale + data.xOffset;
-    mesh.position.y = y * textScale;
+    mesh.position.y = y * textScale + data.yOffset;
     // Place text slightly in front to avoid Z-fighting.
     mesh.position.z = data.zOffset;
     mesh.scale.set(textScale, -1 * textScale, textScale);


ngokevin:
Hm, seems it's used at https://github.com/aframevr/aframe/blob/ea2142b4b487bf41e435b28ea2d089739a6dd31c/src/components/text.js#L391 in a different way.

Pluto8195:
@ngokevin can you explain what the yOffset value is suppose to effect?

ngokevin:
Supposed to effect per-char offset but this solution might be able to replace it. I have to dig down to why the original offset logic was meant for.

Pluto8195:
comment above schema
`// `yOffset` to adjust generated fonts from tools that may have incorrect metrics.`

but I find that being able to just manipulate the y makes formatting text generally much easier

ngokevin:
Thanks for waiting.

Can you figure out what effect it has on these lines? https://github.com/aframevr/aframe/blob/ea2142b4b487bf41e435b28ea2d089739a6dd31c/src/components/text.js#L401 ... Does the yOffset stack? Is the Roboto font hack necessary any more?

dmarcos:
Closing due to inactivity. can reopen if necessary. Thanks

PR #3894, opened 2018-12-06
?inspector={element} and Entity.inspect() helpers to reduce inspector friction

ngokevin:
**Description:**

When working with Inspector, there's friction in refreshing the scene + opening the Inspector + searching the entity / selecting the entity + repeat.

**Changes proposed:**
- Add optional `?inspector={elementId}` URL parameter that automatically opens the Inspector and focuses on the passed-in entity (if any). Then if you are working in an Inspector workflow (e.g., tweaking an entity), just keep the URL parameter there and refresh.
- Add `Entity.inspect()` method to easily open/focus any entity in Inspector from DevTools console.

If merged, matching Inspector changes will follow.

### DIFF
--- a/src/components/scene/inspector.js
+++ b/src/components/scene/inspector.js
@@ -1,8 +1,8 @@
 /* global AFRAME */
 var AFRAME_INJECTED = require('../../constants').AFRAME_INJECTED;
-var bind = require('../../utils/bind');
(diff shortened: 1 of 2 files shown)


dmarcos:
Should have waited for review.

PR #3895, opened 2018-12-06
My Changes (fixes #3865)

### LINKED ISSUE #3865: <a-camera> behavior is broken or has changed.
`<a-camera>` primitive not working as expected. Scene does not render correctly. I didn''t investigate further but replacing `<a-camera>` with `<a-entity camera look-controls>` solves the problem

- A-Frame Version: master
- Platform / Device: all
- Reproducible Code Snippet or URL: examples/tests/cursor

beyerz:
#3865 
**Description:**
When using position coupled with look-controls and the scene does not have webxr.
Then there is a matrix update performed.
In order for this to work, the position should be fully defined, other x and z values are undefined and result in no sky image being shown
**Changes proposed:**
- update a-camera default Component position to be fully defined

### DIFF
--- a/src/extras/primitives/primitives/a-camera.js
+++ b/src/extras/primitives/primitives/a-camera.js
@@ -5,7 +5,7 @@ registerPrimitive('a-camera', {
     'camera': {},
     'look-controls': {},
     'wasd-controls': {},
-    'position': {y: 1.6}
+    'position': {x: 0, y: 1.6, z: 0}
   },
 
   mappings: {


beyerz:
I don't understand what is causing this failure, I am confident that it is not as a result of my changes.
I have run `npm run test:chrome -- --single-run` locally and all tests are passing.

`SUMMARY:
✔ 1081 tests completed
ℹ 15 tests skipped
`

dmarcos:
Thanks for submitting. Could this indicate a deeper issue where position component does not default to 0 for the omitted coordinates. `el.attribute(‘position’, {x:1})`

beyerz:
> Thanks for submitting. Could this indicate a deeper issue where position component does not default to 0 for the omitted coordinates. `el.attribute(‘position’, {x:1})`

Yes I think that what you are saying is correct.
Position component does not define default values, only the type as vec3
```schema: {type: 'vec3'}```

The questions is, should position define a default and even if it does, should camera also define an overriding default in the event that position default may change?

I like to assume that every component should be responisble for itself so that the code can be more loosely-coupled.

ngokevin:
There is a deeper issue, but will merge this first.

PR #3897, opened 2018-12-08
set ssl true for dev server for working webXR

arpu:
**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/scripts/budo.js
+++ b/scripts/budo.js
@@ -25,6 +25,7 @@ var opts = {
   debug: process.env.NODE_ENVIRONMENT !== 'production',
   verbose: true,
   live: true,
+  ssl:true,
   stream: process.stdout,
   host: process.env.HOST,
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. Not sure if SSL should be the default.

arpu:
the only disadvantage is, developer has do allow the unknown ssl certification in the browser ( chrome and firefox will point to this after pageload)

ngokevin:
webxr requires https even for localhost?

arpu:
cannot test this
webxr compatible device is android phone or windows wmr 
(aframe is running on my fedora dev pc)

dmarcos:
WebXR does *not* require HTTPS for localhost

dmarcos:
I would make HTTP as a default and HTTPS as an option. `npm start https`

PR #3898, opened 2018-12-09
fix fullscreen logic and test, remove duplicate success handlers

ngokevin:
**Description:**

A requestFullscreen test was failing.

The enterVRSuccess function was for fully entering success and entering fullscreen if necessary.

It was copy and pasted to enterXRSuccess with all the fullscreen logic, but we really want a separate success handler just for the XRSession. And then after all that XR stuff is done, call a complete aframe-specific enterVRSuccess handler.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -256,53 +256,50 @@ module.exports.AScene = registerElement('a-scene', {
         var self = this;
         var vrDisplay;
         var vrManager = self.renderer.vr;
+
         // Don't enter VR if already in VR.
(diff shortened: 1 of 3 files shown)


dmarcos:
Has been this tested in a browser with WebXR available?

ngokevin:
Tests pass now. I tried testing WebXR on Chrome and couldn't be bothered trying to get it working, since my dev is not on Windows and I don't use `localhost`, and I probably need Chrome origin token.

ngokevin:
needs review to fix tests

ngokevin:
@dmarcos Works for me in Chrome.

PR #3918, opened 2018-12-16
Emit event when a headset is connected

dmarcos:
`navigator.getVRDisplays` (WebVR) and `navigator.xr.requestDevice` (WebXR) are both asynchronous. An event makes easier to detect when a headset is connected and implement enter VR UIs / workflows.

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -4,13 +4,17 @@ var vrDisplay;
 if (navigator.xr) {
   navigator.xr.requestDevice().then(function (device) {
     device.supportsSession({immersive: true, exclusive: true}).then(function () {
+      var sceneEl = document.querySelector('a-scene');
       vrDisplay = device;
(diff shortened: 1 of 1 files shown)


src/utils/device.js
```diff
@@ -4,13 +4,17 @@ var vrDisplay;
 if (navigator.xr) {
   navigator.xr.requestDevice().then(function (device) {
     device.supportsSession({immersive: true, exclusive: true}).then(function () {
+      var sceneEl = document.querySelector('a-scene');
       vrDisplay = device;
+      if (sceneEl) { sceneEl.emit('headsetconnected', {vrDisplay: vrDisplay}); }
```
ngokevin: maybe just `display`

ngokevin:
r+wc

src/utils/device.js
dmarcos: It's used similarly to `checkHeadsetConnected` from utils

PR #3919, opened 2018-12-16
Check for null pose

dmarcos:
**Description:**

**Changes proposed:**
-
-
-


ngokevin:
Oh, i guess a test would be good too.

PR #3932, opened 2018-12-28
Migrate hand model to glTF. Old JSONLoader has been deprecated from latest THREE

dmarcos:
I added the glTF version of the hand models to the assets repo. Everything works as expected but two remaining quirks:

- [Missing a few triangles](https://cl.ly/5f7e7e93681b) around the thumb in the left hand.
- Close hand and grip animations don't play in reverse. 

After figuring out those details this can merge.

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -3,8 +3,8 @@ var registerComponent = require('../core/component').registerComponent;
 
 // Found at https://github.com/aframevr/assets.
 var MODEL_URLS = {
-  left: 'https://cdn.aframe.io/controllers/oculus-hands/v2/leftHand.json',
-  right: 'https://cdn.aframe.io/controllers/oculus-hands/v2/rightHand.json'
+  left: 'https://cdn.aframe.io/controllers/hands/leftHand.glb',
+  right: 'https://cdn.aframe.io/controllers/hands/rightHand.glb'
 };
 
 // Poses.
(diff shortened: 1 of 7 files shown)


dmarcos:
Missing triangles problem is now solved with a new model.

donmccurdy:
Why include LegacyJSONLoader? I don't see it used here.

I tested the grip animation (just in my viewer) and `action.loop = THREE.LoopPingPong` seemed to work as intended for playing in reverse. Are we doing something different than that here? And what version of threejs are you synced against? A fix from r96 may be necessary: https://github.com/mrdoob/three.js/pull/14550.

dmarcos:
@donmccurdy the LegacyJSONLoader is needed to load the examples terrain (updated PR). We should convert to glTF too. I prioritized the hands model that is used by A-Frame core and consolidate all its assets to glTF.

Closing hand animation for example does not play in reverse. The hand animate as expected when closing but pops up open without transition. I'm testing on Firefox / Oculus Rift. What's your setup?

hand

donmccurdy:
I just played the animation forward and backward by loading it in and using dev tools to play it, didn't check on a device yet.

The branch of threejs should have this patch, right? https://github.com/mrdoob/three.js/pull/14550

Also the new call to `stopAllAction()` looks suspect. Note that it calls `reset()` on each action. Is that addition needed?

src/components/hand-controls.js
```diff
@@ -309,14 +322,19 @@ module.exports.Component = registerComponent('hand-controls', {
...
   playAnimation: function (gesture, lastGesture, reverse) {
+    var clip;
     var fromAction;
     var mesh = this.el.getObject3D('mesh');
     var toAction;
 
     if (!mesh) { return; }
 
+    // Stop all current animations.
+    mesh.mixer.stopAllAction();
```
donmccurdy: this looks like it would break the reverse animation?

dmarcos:
Fixed the remaining animation issue. This is ready to go.

PR #3940, opened 2019-01-05
Feature/auto property mapping

LukeTOBrien:
**Description:**  
Created an `autoPropertyMapping` function that creates the mappings between the component and the primitive.

**Issue the change resolves**  
 The **a-link** primitive did not provide mappings for all the properties. 
``` JavaScript
  mappings: {
    href: 'link.href',
    image: 'link.image',
    title: 'link.title'
  }
```
Now the mappings for `a-link` are auto generated from the component's schema:
``` JavaScript
mappings: autoPropertyMapping('link')
```


dmarcos:
Thanks. I think I prefer to keep the mapping explicit. If a new property is added to the schema it will be implicitly added to the primitive. That's not what you always want and introduces inconsistencies inadvertently like possible collisions with other mappings or missing documentation. Can you open a PR with just the properties you are missing?

dmarcos:
I also recomend using the component form instead of primitives:

`<a-entity link="..."></a-entity>`

ngokevin:
Thanks for the PR, I think we prefer explicit. You could send a PR adding the mappings if you want!

Issue #3944, opened 2019-01-09
Thumbstick not rotating on oculus-touch-control.


vt5491:
**thumbstick not animating on oculus-touch-control:**

- A-Frame Version: 0.8.2
- Platform / Device: Oculus Rift
- Reproducible Code Snippet or URL:

This originally came up in 3928, but wasn't the main thrust of that problem.  The other analog buttons now animate properly with the introduction of 3931, but this did not address the thumbstick movement, which has never been animated.  

I have made a fix that allows for thumbstick rotation and am submitting a pull request.

vt5491:
Problem unresolved.  Closed due to lack of interest.

PR #3946, opened 2019-01-09
Add thumbstick movement to oculus touch controller.

vt5491:
**Description:**
fixes ticket #3944 

**Changes proposed:**
- update 'onModelLoaded' to create a pivot group to which thumbstick is attached to properly set pivot point to be the approx. center of the thumbstick.
- added a 'thumbstickmoved' event handler
- rotate the thumbstick in two orthogonal directions as necessary, including one angled at approx. 30 deg.

My main thrust was basically just getting the thumbstick to rotate, which this fix does.  I'm not quite as sure about some of the "architectural" issues that may arise, but hopefully we can work something out.  I do have the following questions:

1.  Should the 'onThumbstickMoved' event handler instead be done by 'onAxisMove' instead?
2.  Should I emit a 'thumbstickchanged' event at the end of the event handler, like 'onButtonChanged' does?  Something like:

[this.el.emit('thumbstick' + 'changed', evt.detail);](url)

3.  I'm willing to write unit tests, but frankly was having a little trouble figuring out how to exactly test this.  I decided to wait on writing any until I get more feedback on my implementation.

While it's not an obviously hard thing to do, the two tricky things were getting the pivot point properly set and rotating about an angled axis.  The default pivot point for the thumbstick is definitely not where it should be for rotations.  See [stackoverflow question](https://stackoverflow.com/questions/54017039/how-to-programmatically-undo-positional-translation-to-pivot-point/54084288#54084288) for more detail.  Also, you have to do a rotation around one axis at a 30 deg angle to accommodate the beveled face of the controller, which was fixed by using 'rotateOnAxis'.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -12,7 +12,6 @@ var TOUCH_CONTROLLER_MODEL_URL = {
 
 var GAMEPAD_ID_PREFIX = 'Oculus Touch';
 
-var DEFAULT_MODEL_PIVOT_OFFSET = new THREE.Vector3(0, 0, -0.053);
 var RAY_ORIGIN = {
   left: {origin: {x: 0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
   right: {origin: {x: -0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}}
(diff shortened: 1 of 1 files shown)


vt5491:
Well, this pull request has been sitting idle for nearly two weeks.  I haven't received any feedback, or even any indication that anyone has even looked at it or even acknowledged it's existence.  The same goes with the associated problem ticket.   Is this standard procedure for the a-frame community?  

At this point, the only thing I can conclude is that there is zero interest among the a-frame maintainers in this fix.  Therefore, I withdraw my request.

micahstubbs:
yikes, it's too bad you didn't hear a reply back then @vt5491.

I'm working on a new project, and would like to implement a `teleport on thumbstick forward` interaction.  It looks like your change would enable that.

Issue #3602, opened 2018-05-25
Using inspector leaves pointerlock in a broken state


donmccurdy:
Opening and closing the inspector while pointerlock is enabled leaves look-controls in a broken state, listening for mousemove events while not actually able to re-enter pointerlock.

PR #3947, opened 2019-01-09
[look-controls] Fix issue accessing inspector from pointerlocked mode.

donmccurdy:
@ngokevin I think this fixes https://github.com/aframevr/aframe/issues/3602, but the default version of the inspector on current aframe doesn't pause/play the scene. It looks like you fixed that recently, so I updated to the master branch, but there I'm seeing the camera freeze after exiting the inspector regardless of whether pointerlock is involved. 🤔


ngokevin:
Thanks, I'll check out Inspector later.

PR #3963, opened 2019-01-23
Use system tick to retrieve controllers instead of unreliable inputsourceschange event

dmarcos:


ngokevin:
I would keep the way it is and file a bug with the browser, and workaround for now with a custom component. In the future, the spec should work as advertised.

PR #3967, opened 2019-01-24
highRefreshRate property to enable 72hz mode on Oculus Browser (fix #3943)

### LINKED ISSUE #3943: Ability to set VRAttributes.highRefreshRate
It would be great to able to set VRAttributes.highRefreshRate.

This allows Oculus Go to run the aframe scene with 72 fps instead of 60 fps.

dmarcos:
the `highRefreshRate` flag on `requestPresent` does not seem to have any effect. Enabling high Refresh Rate on `chrome://flags` does work as expected.

@DigiTec Am I doing things correctly? I added a console message that shows the frame time.

This is WIP simplified for evaluation. The final API will use the renderer component.

### DIFF
--- a/docs/components/renderer.md
+++ b/docs/components/renderer.md
@@ -29,7 +29,7 @@ The `renderer` system configures a scene's
 |-------------------------|---------------------------------------------------------------------------------|---------------|
 | antialias               | Whether to perform antialiasing. If `auto`, antialiasing is disabled on mobile. | auto          |
 | colorManagement         | Whether to use a color-managed linear workflow.                                 | false         |
-| highRefreshRate         | Oculus Browser has a 72hz optional mode vs. the default 60hz                                 | false         |
+| highRefreshRate         | Toggles 72hz mode on Oculus Browser. Defaults to 60hz.                          | false         |
 | sortObjects             | Whether to sort objects before rendering.                                       | false         |
 | physicallyCorrectLights | Whether to use physically-correct light attenuation.                            | false         |
 | maxCanvasWidth          | Maximum canvas width. Uses the size multiplied by device pixel ratio. Does not limit canvas width if set to -1.                                | 1920            |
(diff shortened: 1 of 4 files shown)


Artyom17:
Sorry @dmarcos, didn't see this question. Yes, what you are doing - seems correct to me (https://developer.oculus.com/documentation/oculus-browser/latest/concepts/browser-optimize-rendering/).

dmarcos:
@Artyom17 No problem. I know it's specific to Oculus Browser. Do you have an example with the `highRefreshRate` flag set to true that works? Is it a known browser bug or something in our end?

Artyom17:
> @Artyom17 No problem. I know it's specific to Oculus Browser. Do you have an example with the `highRefreshRate` flag set to true that works? Is it a known browser bug or something in our end?

This example works for me on Go: https://artyom17.github.io/webvr.info/samples/03-vr-presentation-72Hz.html (source https://github.com/Artyom17/webvr.info/blob/gh-pages/samples/03-vr-presentation-72Hz.html).
`        var attributes = {
          highRefreshRate: true,
        };
        vrDisplay.requestPresent([{ source: webglCanvas, attributes: attributes}]).then(function () {
`

dmarcos:
@Artyom17 Thanks. I see. It seems I have to pass the flag in an object on `attributes` property. That seemed to be my mystake.

Do you recommend us to default to 72Hz on Oculus Go or it should be opt-in for developers?

Artyom17:
@dmarcos I'd make it optional because of one issue - not all experiences would be able to make 72 fps on Go.

dmarcos:
@Artyom17 Thanks for the feedback

docs/components/renderer.md
```diff
@@ -29,6 +29,7 @@ The `renderer` system configures a scene's
 |-------------------------|---------------------------------------------------------------------------------|---------------|
 | antialias               | Whether to perform antialiasing. If `auto`, antialiasing is disabled on mobile. | auto          |
 | colorManagement         | Whether to use a color-managed linear workflow.                                 | false         |
+| highRefreshRate         | Oculus Browser has a 72hz optional mode vs. the default 60hz                                 | false         |
```
ngokevin: Description can say what the property does a bit more directly.  

`Toggles 72fps/72hz on Oculus Browser. Will default to 60fps/60hz by default`

PR #3974, opened 2019-01-29
Re-export collada models with new COLLADA2GLTF exporter

dmarcos:


donmccurdy:
@dmarcos the tree models were not a direct conversion with COLLADA2GLTF. They use a separate alpha map for the shadows which didn't convert cleanly, so I fixed that in Blender, along with the colors. on current master I'm seeing:

screen shot 2019-01-28 at 7 44 47 pm

PR #3989, opened 2019-01-31
Add precission property to renderer component

dmarcos:
Setting `mediump` in mobile by defult resulted in very noticeable visual artifacts. It's better if the developers can set the value deliberately knowing the trade offs.

### DIFF
--- a/docs/components/renderer.md
+++ b/docs/components/renderer.md
@@ -23,6 +23,8 @@ The `renderer` system configures a scene's
 
 ## Properties
 
+[precision]: #precision
+
 | Property                | Description                                                                     | Default Value |
 |-------------------------|---------------------------------------------------------------------------------|---------------|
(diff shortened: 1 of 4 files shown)


docs/components/renderer.md
```diff
@@ -32,6 +32,7 @@ The `renderer` system configures a scene's
 | maxCanvasWidth          | Maximum canvas width. Uses the size multiplied by device pixel ratio. Does not limit canvas width if set to -1.                                | 1920            |
 | maxCanvasHeight         | Maximum canvas height. Behaves the same as maxCanvasWidth.                      | 1920          |
 | logarithmicDepthBuffer  | Whether to use a logarithmic depth buffer.                                      | auto          |
+| precision  |       Shader Precision                                | high          |
```
ngokevin: should describe it more here, what it does, including the options. I'd almost just shove the whole paragraph from below into here, or at least link it.

ngokevin:
r+wc

Issue #1473, opened 2016-05-18
Document how to enter/exit VR


cvan:
**Description:**

``` js
document.querySelector('a-scene').enterVR();
document.querySelector('a-scene').exitVR();
```

And also document how to when `<iframe>`-ing content.
- A-Frame Version: any
- Platform/Device: all

ngokevin:
https://github.com/aframevr/aframe/blob/master/docs/core/scene.md#methods

cvan:
[Nice section here](https://github.com/aframevr/aframe/blob/master/docs/core/scene.md#running-content-scripts-on-the-scene), but for some reason waiting for `loaded` doesn't work. you have to add a `setTimeout`. gotta figure out why:

```js
<script>
   (function () {
       // switch to stereoscopic mode directly on page load, this needs to be after the a-scene load.
       var scene = document.querySelector('a-scene');
       if (scene.hasLoaded) {
           scene.enterVR();
       } else {
           scene.addEventListener('loaded', function () {
               setTimeout(function () {
                   scene.enterVR();
               }, 1000);
           });
       };
   })();
</script>
```

Also, not exactly specific to this issue, but still think we should support #705.

ngokevin:
Well, no one is against #705. Though I might also just recommend creating a scene component. They have the advantage that you don't have to worry about query selecting, waiting for events. Just write JS and attach:

```
AFRAME.registerComponent('do-whatever', {
  init: function () {
    // Do whatever.
  }
});
```

cvan:
like it

ngokevin:
Though I guess you might still have to add the event listener...but at least you don't have to query selector or worry whether it's already loaded. And you get reusable code for free.

manojkota:
This is not working with version >=0.3.0

cvan:
> This is not working with version >=0.3.0

Are you testing in Chromium or Firefox Nightly? Chromium currently has a temporary user-gesture requirement; Firefox Nightly does not.

cvan:
(Accidentally closed issue.)

manojkota:
I have tested in firefox & chrome in Android. It is not working. When I change the afrmae version to 0.2.0, it is directly going to vr mode.

HAZARDU5:
The following code seems to work with v0.4.0 in Firefox Nightly when `auto-init-vr` component is attached to the scene. Unfortunately Chromium still requires user interaction before it can auto-trigger VR mode:

```
[...]

PR #3994, opened 2019-02-04
Clarify seamless VR link traversal

0atman:
**Description:**

We should make it clear that the VR experience is only seamless on Vive et al. We are not managing our user's expectations correctly, see https://github.com/aframevr/aframe/issues/1473

**Changes proposed:**
- Vive works
- Cardboard doesn't work
- regular fullscreen doesn't work

### DIFF
--- a/docs/components/link.md
+++ b/docs/components/link.md
@@ -19,6 +19,8 @@ across pages, the following conditions must apply:
 - The browser is up-to-date with the WebVR specification and implements the `vrdisplayactivate` event.
 - The destination web page listens to the window [`vrdisplayactivate`] event and enters VR.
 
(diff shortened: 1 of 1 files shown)


0atman:
I see. I still think it's confusing: if you use Firefox mobile for cardboard VR, or Firefox desktop for a fullscreen 2D experience, you still get bounced out of VR/fullscreen.

dmarcos:
Added info in docs:

https://github.com/aframevr/aframe/commit/8f94626b81bc7a2c6faaf0742d4af35f2bd63c06

Let me know if you are missing any other details

0atman:
Perfection! Thank you so much.

PR #4003, opened 2019-02-06
Exception in coordinates.js fix

Artyom17:
In the case when defaultVec is null the parseIfString throws an exception, because val becomes null. 

val becomes null because of function parse: for example we have a vec3 with x, y, z and w is undefined. The defaultVec can be null. In this case, line 40 would make local var 'w' to equal to null. Then, on the line 44 it will do strict comparison with 'undefined' which fails and would call parseIfString passing null as a parameter, which will throw an exception.

### DIFF
--- a/src/utils/coordinates.js
+++ b/src/utils/coordinates.js
@@ -38,10 +38,10 @@ function parse (value, defaultVec) {
     y = value.y === undefined ? defaultVec && defaultVec.y : value.y;
     z = value.z === undefined ? defaultVec && defaultVec.z : value.z;
     w = value.w === undefined ? defaultVec && defaultVec.w : value.w;
-    if (x !== undefined) { value.x = parseIfString(x); }
-    if (y !== undefined) { value.y = parseIfString(y); }
-    if (z !== undefined) { value.z = parseIfString(z); }
-    if (w !== undefined) { value.w = parseIfString(w); }
+    if (x !== undefined && x !== null) { value.x = parseIfString(x); }
+    if (y !== undefined && y !== null) { value.y = parseIfString(y); }
+    if (z !== undefined && z !== null) { value.z = parseIfString(z); }
+    if (w !== undefined && w !== null) { value.w = parseIfString(w); }
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


ngokevin:
Hi thanks, is it possible to write a test case, or post some code that would trigger this exception to get a better idea?

src/utils/coordinates.js
```diff
@@ -96,7 +96,7 @@ module.exports.isCoordinate = function (value) {
 };
 
 function parseIfString (val) {
-  if (val.constructor === String) {
+  if (val != null && val.constructor === String) {
```
ngokevin: If we're worried about nulls. This change should be all that's needed?

Also use `!==` not `!=`.

Artyom17: I intentionally used '!=' because in this case both 'undefined' and 'null' will satisfy the condition. Do you think it is better to do 'if (val !== undefined && val !== null)' instead?

Artyom17: Answering to the question "This change should be all that's needed?": you can get away with only this change, however, 'value.w' in parse function will get 'null' assigned (but value.w shouldn't be even defined, so this may cause potential side effects).

Artyom17:
> Hi thanks, is it possible to write a test case, or post some code that would trigger this exception to get a better idea?

Uhm... I got the described situation while was working on A-Blast: I got parse call with vec3 and defaultVec = null. That is basically all the test case.

ngokevin:
I'd like to see code so I can verify it explicitly versus parsing a paragraph and potentially misunderstanding. If this change makes sense, then a unit test is needed anyways.

Artyom17:
> I'd like to see a line of example code pasted so I can verify it explicitly versus parsing a paragraph and potentially misunderstanding. If this change makes sense, then a unit test is needed anyways.

Fair point. Modified PR to include the test case.

src/utils/coordinates.js
Artyom17: Technically speaking, the change in this line is not mandatory, since the main checks are in 'parse' function and I've added it here  'to be safe'. We can remove this modification.

dmarcos: better to use `!==` for consistency with the rest of the code base.

Artyom17:
Have updated PR to comply to coding conventions around != vs !==

PR #4011, opened 2019-02-12
Document uppateSchema changes, add override

3DataWill:
**Description:** Congratulations on the new release! This updates the docs for the optimized approach to `updateSchema` and the new `schemaChange` flag in component property definitions. Also, if you'll accept it, adds the `schemaAlwaysUpdates` component definition property to override the optimized behavior.

I've been using `updateSchema` to give some components entirely dynamic schemas that take new properties as they come in and give them a property type definition. I couldn't figure a way to do this with the `schemaChange` flag as the updates often contain only previously unknown properties.

**Changes proposed:**
- `schemaAlwaysUpdates` component definition property
- `updateSchema` is called on every update if `schemaAlwaysUpdates` is `true`

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -473,7 +473,8 @@ AFRAME.registerComponent('sound', {
 `.updateSchema ()`, if defined, is called when certain flagged properties are
 updated to check if the schema also needs to be modified. To flag a property
 as one that triggers a schema udpate, add `schemaChange: true` to the property
-definition.
+definition. For a schema that always updates, for example to accept new
+properties, set the `schemaAlwaysUpdates` definition property to `true`.
 
 The `updateSchema` handler is often used to:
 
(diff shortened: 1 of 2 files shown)


ngokevin:
Would it be possible to pass the flag into every property?

3DataWill:
I read it as only checking for the flag on existing properties (new ones
don't have a schema definition yet until after updateSchema is called)

ngokevin:
What do entirely dynamic schemas look like?

3DataWill:
I've included an example in the docs in this PR for how to setup an entirely dynamic schema. It allows me to have a single data management component that works with all different types of plot components that each have a different set capabilities and whose capabilities can be expanded by other components on the entity.

3DataWill:
I also looked at a less exposed option of a non-enumerable property property on the schema object to flag the need for continuous schema updates, but that would have required additional code changes in the component initialization path to pass the hidden property from the schema definition on to the instance schema object.

ngokevin:
That's kinda funky. Can you just dynamically register components?

ngokevin:
Closing for now. I think this case is too edge-y, feel free to post a code sample of your API though. I recommend breaking your component up if possible. Or finding a workaround (e.g., can you call something manually?).

PR #4016, opened 2019-02-14
Remove extra properties from objectPool objects

### LINKED ISSUE #4012: objectPool property leakage
**Description:** For components with schemas that change (e.g. geometry), the object pooling can lead to the presence of inappropriate keys in a components data object when a used object is picked up from the pool during component initialization

- A-Frame Version: 0.9.0
- Platform / Device: Desktop FF & Chrome
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/ultra-woolen?path=index.html:17:13
Expected behavior: console log of geometry.data keys contains only sphere-related values
Observed behavior: console log includes height,width,depth from recycled box data object

3DataWill:
@ngokevin what do you think of this approach?

**Description:** Prevents keys added to component data pooled objectss by dynamic schemas from appearing in other instances of the component. Only needs to be done on initialization because the object copy methods used in the update cycle already skip over keys with undefined values. Should have no effect for components without dynamic schemas.

Close #4012 

**Changes proposed:**
- `utils.objectPool.removeUnusedKeys` deletes properties not found in a reference object (i.e. the component template schema)
- Apply above method when fetching initial objects during component initialization

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -50,10 +50,22 @@ var Component = module.exports.Component = function (el, attrValue, id) {
 
   // Store component data from previous update call.
   this.attrValue = undefined;
-  this.nextData = this.isObjectBased ? this.objectPool.use() : undefined;
-  this.oldData = this.isObjectBased ? this.objectPool.use() : undefined;
-  this.previousOldData = this.isObjectBased ? this.objectPool.use() : undefined;
-  this.parsingAttrValue = this.isObjectBased ? this.objectPool.use() : undefined;
+  if (this.isObjectBased) {
+    this.nextData = this.objectPool.use();
+    // Drop any properties added by dynamic schemas in previous use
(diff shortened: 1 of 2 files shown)


ngokevin:
Thanks, I'll take a look later.

src/core/component.js
```diff
@@ -50,10 +50,22 @@ var Component = module.exports.Component = function (el, attrValue, id) {
 
   // Store component data from previous update call.
   this.attrValue = undefined;
-  this.nextData = this.isObjectBased ? this.objectPool.use() : undefined;
```
ngokevin: Perhaps I'd just add a `keysFilter` argument to `objectPool.use` and pass that in versus a separate function.

ngokevin: @3DataWill What do you think about that?

PR #4022, opened 2019-02-15
Consider case when both WebVR and WebXR are available

dmarcos:


modulesio:
LGTM, for at least not crashing if both APIs exist.

PR #4033, opened 2019-02-20
Replace isOculusGo with isMobileVR function (fix #4032)

### LINKED ISSUE #4032: Have `AFRAME.utils.device.isMobile()` return `false` for Firefox Reality
## Description
A-Frame identifies Firefox Reality as a mobile browser since its `User-Agent` string contains 'Android' and 'Mobile VR', but it should be recognised as a standalone VR browser (like Oculus Browser is).

## Current behaviour
`AFRAME.utils.device.isMobile()` returns `true` in Firefox Reality but not in Oculus Browser.

## Expected behaviour
`AFRAME.utils.device.isMobile()` returns `false` in Firefox Reality as it does in Oculus Browser.

## Context
I've seen projects using A-Frame's [`AFRAME.utils.device.isMobile()`](https://github.com/aframevr/aframe/blob/aa792c9/src/utils/device.js#L52) (and the deprecated `AFRAME.utils.isMobile()`), [`AFRAME.utils.device.checkHasPositionalTracking()`](https://github.com/aframevr/aframe/blob/aa792c9/src/utils/device.js#L39), `window.webvrpolyfill.hasNative`, and `window.hasNativeWebVRImplementation` to infer whether the user is in a mobile Cardboard-esque environment vs. a standalone VR browser (e.g., on the Oculus Go), in a WebVR-capable desktop browser, or in a non-WebVR-capable browser (mobile or desktop).

Specifically, [`AFRAME.utils.device.isOculusGo()`](https://github.com/aframevr/aframe/blob/aa792c9/src/utils/device.js#L91) could be modified to match these Oculus Browsers (ordered by Oculus Browser, Samsung Browser, and Firefox Reality).

```js
  return /Pacific Build.+OculusBrowser.+SamsungBrowser.+Mobile VR/i.test(window.navigator.userAgent);
```

to

```js
  return /(Pacific Build.+OculusBrowser.+SamsungBrowser.+)|(SamsungBrowser)|(Mobile VR)/i.test(window.navigator.userAgent);
```

This check will need to change for the Oculus Go, especially with the upcoming Oculus Quest.

I've been working with content providers to improve their WebCompat for the VR browsers on the Go. For now, this has involved mostly checking for `/(OculusBrowser)|(SamsungBrowser)|(Mobile VR)/i.test(window.navigator.userAgent)`. We have no tentative plans to add the `Pacific Build` bit to Firefox's `User-Agent`, but we'll likely be using [`WEBGL_debug_renderer_info`](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_renderer_info), [`navigator.hardwareCurrency`](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorConcurrentHardware/hardwareConcurrency), and other signals to infer browser + platform feature capabilities + performance.

<hr>

- A-Frame Version: 0.8.2, 0.9.0
- Platform / Device: Firefox Reality (v1.1, v1.2)
    - https://www.eff.org/spot-vr
    - Context:
      - https://github.com/EFForg/spot_the_surveillance/issues/85#issuecomment-465398460
      - https://github.com/EFForg/spot_the_surveillance/issues/85#issuecomment-465652378

<hr>

dmarcos:
This combines `isGearVR` / `isOculusGo` into an `isMobileVR` function to differentiate from `isMobile` (aka Cardboard / Magic Window). Considers Samsung Internet, Firefox Reality and Oculus Browser

Also removes unused `checkHasPositionalTracking ` function

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -32,17 +32,7 @@ function checkHeadsetConnected () { return !!getVRDisplay(); }
 module.exports.checkHeadsetConnected = checkHeadsetConnected;
 
 /**
- * Check for positional tracking.
- */
-function checkHasPositionalTracking () {
-  var vrDisplay = getVRDisplay();
-  if (isMobile() || isGearVR() || isOculusGo()) { return false; }
-  return vrDisplay && vrDisplay.capabilities.hasPosition;
-}
-module.exports.checkHasPositionalTracking = checkHasPositionalTracking;
-
-/**
- * Checks if browser is mobile.
+ * Checks if browser is mobile and not stand-alone dedicated vr device.
  * @return {Boolean} True if mobile browser detected.
  */
 var isMobile = (function () {
@@ -55,7 +45,7 @@ var isMobile = (function () {
     if (isIOS() || isTablet() || isR7()) {
       _isMobile = true;
     }
-    if (isOculusGo()) {
+    if (isMobileVR()) {
       _isMobile = false;
     }
   })(window.navigator.userAgent || window.navigator.vendor || window.opera);
@@ -79,18 +69,13 @@ function isIOS () {
 }
 module.exports.isIOS = isIOS;
 
-function isGearVR () {
-  return /SamsungBrowser.+Mobile VR/i.test(window.navigator.userAgent);
-}
-module.exports.isGearVR = isGearVR;
-
 /**
- *  Detect Oculus Go device
+ *  Detect browsers in Stand-Alone headsets
  */
-function isOculusGo () {
-  return /Pacific Build.+OculusBrowser.+SamsungBrowser.+Mobile VR/i.test(window.navigator.userAgent);
+function isMobileVR () {
+  return /(OculusBrowser)|(SamsungBrowser)|(Mobile VR)/i.test(window.navigator.userAgent);
 }
-module.exports.isOculusGo = isOculusGo;
+module.exports.isMobileVR = isMobileVR;
 
 function isR7 () {
   return /R7 Build/.test(window.navigator.userAgent);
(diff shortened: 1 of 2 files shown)


src/utils/device.js
```diff
@@ -79,18 +69,13 @@ function isIOS () {
 }
 module.exports.isIOS = isIOS;
 
-function isGearVR () {
-  return /SamsungBrowser.+Mobile VR/i.test(window.navigator.userAgent);
-}
-module.exports.isGearVR = isGearVR;
```
cvan: similar to the deprecation warnings in `src/utils/index.js`, would it be useful to first deprecate this method with a `console.warn` before removing?

I hope folks are not calling these methods, but I have seen them in various A-Frame projects recently.

src/utils/device.js
```diff
@@ -79,18 +69,13 @@ function isIOS () {
...
-
 /**
- *  Detect Oculus Go device
+ *  Detect browsers in Stand-Alone headsets
  */
-function isOculusGo () {
```
cvan: same as my comment above

src/utils/device.js
dmarcos: There's a warning that offers alternative method https://github.com/aframevr/aframe/pull/4033/files#diff-8968c62593320351ced6a0ef776ef599R30
Not only a a deprecation. isGearVR was no longer correct. It was checking for Samsung Internet browser. Written when Oculus Go was not a thing. There's no longer a way to discriminate between OculusGo and GearVR and probably not needed. `isMobileVR` replaces prior functionality.

src/utils/device.js
dmarcos: https://github.com/aframevr/aframe/commit/6e7d33bc5ed97eb7e979d0e8d02c0155f14eec82

src/utils/device.js
cvan: ah, understood. thanks for explaining.

PR #4041, opened 2019-02-26
Add alpha property to renderer to disable transparent backbuffer

robertlong:
**Description:**
In Hubs we saw a few bugs with glTF models where models would render completely white or black depending on the platform. We traced it back to glTF models with transparent baseColor textures and no blend mode. Three.js still writes these alpha values to the frame buffer even if `transparent` is set to false. If the WebGLRenderingContext is passed `alpha: true` then the backbuffer is transparent and whatever is behind the canvas is used as the backbuffer.

**Changes proposed:**
- Add the `alpha` property to the `renderer` attributes to allow for disabling the transparent backbuffer.


dmarcos:
Thanks!

PR #4043, opened 2019-02-26
WebGL2Renderer (WIP)

robertlong:
**Description:**
In Hubs we'd like to use ETC2 texture compression for reducing texture upload times. The [WEBGL_compressed_texture_etc](https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc/) WebGL extension is only available in WebGL2. Three.js has shipped a WebGL2Renderer, but it's not finished yet. This PR adds a `webgl2` property to `renderer` that creates a WebGL2RenderingContext and passes it to the WebGLRenderer. This causes shaders to be upgraded to OpenGL ES 3.0 and we can also use all of the WebGL2 extensions.

We're currently merging this into Hubs, but I could understand if you want to hold off on adding this upstream. Regardless, feedback is welcome!

**Changes proposed:**
- Add `webgl2` property to `renderer`
- Upgrade shaders at runtime to OpenGL ES 3.0
- Remove reserved keyword `sampler` from msdf shader.

**To Do:**
- [ ] Determine how best to upgrade RawShaderMaterials on `AFRAME.registerShader()`
- [ ] Add documentation

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -578,6 +578,18 @@ module.exports.AScene = registerElement('a-scene', {
             rendererConfig.logarithmicDepthBuffer = rendererAttr.logarithmicDepthBuffer === 'true';
           }
 
+          if (rendererAttr.webgl2 && rendererAttr.webgl2 === 'true') {
+            rendererConfig.context = this.canvas.getContext('webgl2', {
+              alpha: rendererConfig.alpha,
+              depth: true,
+              stencil: true,
+              antialias: rendererConfig.antialias,
+              premultipliedAlpha: true,
+              preserveDrawingBuffer: false,
+              powerPreference: 'default'
+            });
+          }
+
           this.maxCanvasSize = {
             width: rendererAttr.maxCanvasWidth
               ? parseInt(rendererAttr.maxCanvasWidth)
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -11,6 +11,7 @@ var utils = require('../../utils/');
 var AEntity = require('../a-entity');
 var ANode = require('../a-node');
 var initPostMessageAPI = require('./postMessage');
+var shaders = require('../shader').shaders;
 
 var bind = utils.bind;
 var isIOS = utils.device.isIOS();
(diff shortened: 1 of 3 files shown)


ngokevin:
Thanks so much! We'll probably wait until full WebGL2 support in three.js, but will definitely refer this to people whenever WebGL2 comes out. Closing for now, but let's reopen when it's more production-ready.

arpu:
@ngokevin why do you think this is not  production-ready?

dmarcos:
definitively not ready

arpu:
do you have some more information on this? why its not ready?

ngokevin:
It says in the pr description the renderer is work in progress

dmarcos:
@arpu You can check progress on THREE: https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGL2Renderer.js

Not sure if the strategy is still to have a separate renderer for wegl2 as it hasn’t been much progress

robertlong:
Maybe @takahirox can give a short summary on the WebGL2Renderer progress?

As for our use of WebGL2 in Hubs, we really aren't using it for much outside of experimenting with the multiview extension. We'll get more use out of it in the future as we adopt UBOs and VAOs for reducing gl call overhead.

takahirox:
`WebGLRenderer` started to support WebGL2. You can use WebGL2 as this PR does.

```javascript
var canvas = document.createElement( 'canvas' );
var context = canvas.getContext( 'webgl2' );
var renderer = new THREE.WebGLRenderer( { canvas: canvas, context: context } );
```

 https://threejs.org/docs/#manual/en/introduction/How-to-use-WebGL2

I don't really think we update `WebGL2Renderer` unless we realize we should optimize renderer very differently between WebGL1 and WebGL2. (But it's kinda very dependent on mrdoob's decision.)

/cc @fernandojsg (Let me know if I'm wrong)

kmilo9999:
Hi,
Is there any update on supporting webgl2 on Aframe? 
I have developed a vr volume data visualizer for Aframe, and I like to publish it as a component. However, it depends on my modified version of aframe-master.

Thanks

arpu:
same here with Multiview

Issue #4042, opened 2019-02-26
Only one steamvr controller visible


InfiniteLee:
**Description:**
I'm not sure exactly what changed but it seems that now when using steamvr, one controller is always hidden. I've traced this to https://github.com/aframevr/aframe/blob/65952d6133ff23f542e43beb3ae587d66a971c21/src/components/tracked-controls-webvr.js#L107
which relies on a controller being found in https://github.com/aframevr/aframe/blob/bc577c968cf94e78f0706cae99148d254d370eba/src/utils/tracked-controls.js#L115

Because `1` is passed in for `filterControllerIndex` when looking for the `right` controller, this causes the condition here: https://github.com/aframevr/aframe/blob/bc577c968cf94e78f0706cae99148d254d370eba/src/utils/tracked-controls.js#L145 to never be true. 

The simplest fix I can think of is removing https://github.com/aframevr/aframe/blob/65952d6133ff23f542e43beb3ae587d66a971c21/src/components/tracked-controls-webvr.js#L100 as it seems extraneous (and wrong), though I'm not entirely sure that is the correct fix or not. My suspicion is that the bug where the controller is not found has been around for a while but was made only "visible" because of this commit: https://github.com/aframevr/aframe/commit/7754b2c478369aa4dbe1d8730788d8335ecd2ab8#diff-efa22db2dae70644518618ea4dbfde3e

- A-Frame Version: 0.9.0
- Platform / Device: firefox, steamvr
- Reproducible Code Snippet or URL: reproducible in hubs (hubs.mozilla.com)

dmarcos:
This was fixed by https://github.com/aframevr/aframe/pull/4024

Does master work?

InfiniteLee:
I didn't see that, yea that should fix it for some cases. However in the case that someone actually does specify `controller` to be non-default (non-zero) in `tracked-controls`, I believe this bug will still occur. Given the stated purpose of that property ("Index of controller in array returned by Gamepad API"), it doesn't seem unreasonable that would happen. I also still think that it is still invalid to use that value for `filterControllerIndex` when calling `findMatchingControllerWebVR`. With #4024, the issue is just being avoided by virtue of that value being kept to the default value of 0. As such, I still think that it should be removed.
I would also point out this: https://github.com/aframevr/aframe/blob/65952d6133ff23f542e43beb3ae587d66a971c21/src/components/tracked-controls-webvr.js#L55
[...]

PR #4047, opened 2019-02-28
invert the index values assumed for left and right vive controllers

InfiniteLee:
see: #4042


ngokevin:
@kfarr has tested this in all browsers. There are oddities when a Vive Tracker is connected, but this PR is an improvement. We will want to revisit this later. Thanks!

PR #4051, opened 2019-03-03
Fix the animation component for custom vec3 properties

Galadirith:
**Description:**

Fix the `animation` component to correctly animate `vec3` properties of custom components. The current implementation handles the cases for `position`, `scale` and `rotation` but does not handle custom properties that do not represent properties of `Object3D` (`PROP_POSITION`, `PROP_ROTATION`, `PROP_SCALE`).

**Changes proposed:**
- Change the `animations` property reference used for custom `vec3` properties
 https://github.com/aframevr/aframe/blob/ff2c7a4b0d6eba24ab538fb2e66a22efcfafbd5e/src/components/animation.js#L404
  to `animatables` as is used for `position`, `scale` and `rotation`
 https://github.com/aframevr/aframe/blob/ff2c7a4b0d6eba24ab538fb2e66a22efcfafbd5e/src/components/animation.js#L377

**My thanks:**

You all really inspire me ❤️ I am so grateful for all the hard work you've all put into developing A-Frame, and your commitment to an open VR web (sorry for being unprofessional and including this message in an PR).

### DIFF
--- a/src/components/animation.js
+++ b/src/components/animation.js
@@ -401,7 +401,7 @@ module.exports.Component = registerComponent('animation', {
     config.update = (function () {
       var lastValue = {};
       return function (anim) {
-        var value = anim.animations[0].target;
+        var value = anim.animatables[0].target;
 
         // Animate rotation through radians.
         // For animation timeline.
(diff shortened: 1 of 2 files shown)


ngokevin:
Thanks much for the kind words, and the solid PR and tests!

PR #4054, opened 2019-03-05
 Reenable hand-controls test, add tests, simplify determineGesture logic

DigiTec:
**Description:**
This re-enables hand-controls test which look errantly skipped by a prior submit. I've also added tests for the missing cases for the generic hand controls. Finally I simplified the logic for Vive controllers out of the logic for all of the other controllers. This is all in preparation for adding more potential hand controls type devices which may also branch logic, but in less fundamental ways than the Vive wands do.

**Changes proposed:**
- Re-enable hand-controls.test.js
- Add tests for hold, thumbs up and point + thumb
- Simplify determineGesture by lifting out the isVive special cases.


src/components/hand-controls.js
```diff
@@ -245,23 +245,23 @@ module.exports.Component = registerComponent('hand-controls', {
     var isTrackpadActive = this.pressedButtons['trackpad'] || this.touchedButtons['trackpad'];
     var isTriggerActive = this.pressedButtons['trigger'] || this.touchedButtons['trigger'];
     var isABXYActive = this.touchedButtons['AorX'] || this.touchedButtons['BorY'];
-    var isVive = isViveController(this.el.components['tracked-controls']);
 
     // Works well with Oculus Touch and Windows Motion Controls, but Vive needs tweaks.
-    if (isGripActive) {
-      if (isVive) {
+    if (isViveController(this.el.components['tracked-controls'])) {
```
dmarcos: Is possible to keep `isVive` variable? It makes the code a bit more readable

DigiTec: Yep, since it was only used once it felt redundant, but easy enough to put back in.

DigiTec:
@dmarcos - Not getting the failures in my local runs, only seems to be in Travis-CI and seems entirely unrelated to the changes. How do we normally deal with these if this is indeed a spurious error?

dmarcos:
Thanks. We have occasional test timeouts in Travis. I restarted the test and now all is green. Thank you.

PR #4059, opened 2019-03-06
Oculus go controls tests

DigiTec:
**Description:**

Fixes up tests to be more consistent for the Oculus GO controls and rely more heavily on the setup methods and state rather than trying to mock state (no reason to mock state if we can effectively use the system to set everything up). Since the tests bootstrap through all cases, we can be assured that tests later in the series that rely on capabilities earlier will have a canary test fail first.

This points out there is a lot of potential to have a suite like this shared by all with the right factoring. I will convert two more of the controls suites to use similar methodology and then see if commonality can be lifted out. For now, this makes the Go controls and tests easier for me to maintain and improve.

**Changes proposed:**
- Factor our better setup functionality for each test.
- Remove redundancy and rely on helpers for validation of common state.
- Remove a redundant test that wasn't testing a unique conditional state.

### DIFF
--- a/tests/components/oculus-go-controls.test.js
+++ b/tests/components/oculus-go-controls.test.js
@@ -7,6 +7,9 @@ suite('oculus-go-controls', function () {
     el.setAttribute('oculus-go-controls', 'hand: right');  // To ensure index is 0.
     el.addEventListener('loaded', function () {
       var component = el.components['oculus-go-controls'];
+      // Initially no controllers are present
+      component.controllers = [];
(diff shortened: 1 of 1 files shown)


ngokevin:
I rebased the branch. Looks alright, thanks! I don't think we need at the moment to make a full shared suite system for controllers. I think copy/paste in some areas is okay for now and maybe even easier to read/modify a test without referencing other stuff. I think it's a good idea if we ever have a common system in the core for controllers.

PR #4061, opened 2019-03-06
use self instead of this in sound.onEnded callback

hars-shhh:
**Description:**
In sound component on sound ended isPlaying is not set to false

```javascript
sound.onEnded = function () {
    this.isPlaying = false;
    el.emit('sound-ended', self.evtDetail, false);
};
```

**Changes proposed:**
- use self instead of this or change the onEnded callback function to arrow function.
```javascript
sound.onEnded = function () {
    self.isPlaying = false;
    el.emit('sound-ended', self.evtDetail, false);
};
```
or
```javascript
sound.onEnded = () => {
    this.isPlaying = false;
    el.emit('sound-ended', self.evtDetail, false);
};
```

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -33,6 +33,8 @@ module.exports.Component = registerComponent('sound', {
     this.loaded = false;
     this.mustPlay = false;
 
+    this.onEnded = this.onEnded.bind(this);
+
     // Don't pass evt because playSound takes a function as parameter.
     this.playSoundBound = function () { self.playSound(); };
   },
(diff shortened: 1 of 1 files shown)


src/components/sound.js
```diff
@@ -173,7 +173,7 @@ module.exports.Component = registerComponent('sound', {
     for (i = 0; i < this.pool.children.length; i++) {
       sound = this.pool.children[i];
       sound.onEnded = function () {
-        this.isPlaying = false;
+        self.isPlaying = false;
```
DigiTec: Question for whomever does review this from the AFrame team.

Are we explicitly avoiding arrow functions to capture this?

If we are, should we instead be using the bind() helper instead of capturing a "self"? This would make things more consistent. We tend to use bind() for simple functions and any complex function wrappers we instead use self.

dmarcos: Yes, no arrow functions: Not supported in IE 11. As @DigiTec said probably bind a bit cleaner than self.

PR #4065, opened 2019-03-07
Unused mock methods (at least in local tests). Try to remove.

DigiTec:
**Description:**

Having extra methods on the instance leaks additional abstraction that is unnecessary. Standardize how all controllers pull in trackedControlsUtils and related methods, remove the local storage on the instance since nobody mocks them out and verify tests all pass.

**Changes proposed:**
- Standardize how trackedControlsUtils is pulled in and methods are retrieved.
- Remove the mockable instance variables and use the direct methods instead.


ngokevin:
thanks!

PR #4066, opened 2019-03-07
There is a double bind on onAxisMoved, remove for perf/consistency

DigiTec:
**Description:**

We double bind the method in init and bindMethods. Remove the redundant bind for perf/consistency/profiling

**Changes proposed:**
- Remove the extra bind and assignment.


dmarcos:
Good catch. Thanks

PR #4067, opened 2019-03-07
Oculus touch controls tests standardization.

DigiTec:
**Description:**

Mirrors the changes in oculus-go-controls tests but updates for oculus-touch-controls. This is the last precursor change to lifting out these tests into a helper and then updating all 7 of our controls types. The "varying" information is the component name and the controllers when present.

**Changes proposed:**
- Follow the same pattern as oculus-go-controls when replicating the tests.
- Update the component name and present controller information to match Oculus Touch controllers.

**Future proposed changes:**
- Convert both oculus-go-controls and oculus-touch-controls for this functionality to a shared suite.
- Develop consensus and a pattern for shared suites/tests/generation since one doesn't appear to exist.

### DIFF
--- a/tests/components/oculus-touch-controls.test.js
+++ b/tests/components/oculus-touch-controls.test.js
@@ -10,97 +10,103 @@ suite('oculus-touch-controls', function () {
     el.setAttribute('oculus-touch-controls', '');
     el.addEventListener('loaded', function () {
       component = el.components['oculus-touch-controls'];
-      component.controllersWhenPresent = [
-        {id: 'Oculus Touch', index: 0, hand: 'left', pose: {}}
-      ];
+      // Initially no controllers are present
+      component.controllers = [];
+      // Our Mock data for enabling the controllers.
+      component.controllersWhenPresent = [{
+        id: 'Oculus Touch',
+        index: 0,
+        hand: 'left',
+        pose: {}
+      }];
       done();
     });
   });
 
   suite('checkIfControllerPresent', function () {
-    test('removes event listeners if controllers not present', function () {
-      var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
-      var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
-      var removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
+    var component;
+    var controllerSystem;
+    var addEventListenersSpy;
+    var injectTrackedControlsSpy;
+    var removeEventListenersSpy;
 
-      // Mock has not been checked previously.
-      delete component.controllerPresent;
+    setup(function (done) {
+      component = this.el.components['oculus-touch-controls'];
+      controllerSystem = this.el.sceneEl.systems['tracked-controls-webvr'];
+      addEventListenersSpy = sinon.spy(component, 'addEventListeners');
+      injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
+      removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
+      done();
+    });
+
+    /**
+     * Verifies that the method spy's are in the right state for a controller
+     * that has been injected.
+     *
+     * @param {object} component - The oculus-go-controls component to verify fields.
+     */
+    function verifyControllerSetup (component) {
+      sinon.assert.calledOnce(injectTrackedControlsSpy);
+      sinon.assert.calledOnce(addEventListenersSpy);
+      sinon.assert.notCalled(removeEventListenersSpy);
+      assert.strictEqual(component.controllerPresent, true);
+    }
+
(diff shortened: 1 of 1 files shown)


ngokevin:
Needs a rebase. I think I prefer not having a shared suite for now, it's less flexible and harder to follow what's being asserted. A suite for the core controller code might be useful, but needs thinking on the API. Perhaps unit tests on the underlying utils would be better alternative for now.

DigiTec:
Wow, I don't think that rebase worked. I might have to cherry pick this into a new branch and try again.

ngokevin:
I use `git pull --rebase upstream master`

DigiTec:
Thanks for the tip, that was definitely more efficient. I still had an issue that src\index.js was stuck on a 3/5 hash and not updated to 3/8, I wound up doing a git reset hard + cherry pick of the single change and that seems to have worked out well.

tests/components/oculus-touch-controls.test.js
```diff
@@ -10,38 +10,51 @@ suite('oculus-touch-controls', function () {
...
+        index: 0,
+        hand: 'left',
+        pose: {}
+      }];
       done();
     });
   });
 
   suite('checkIfControllerPresent', function () {
-    test('removes event listeners if controllers not present', function () {
```
dmarcos: This test seems to be gone after the changes. Intentional?

tests/components/oculus-touch-controls.test.js
```diff
@@ -10,38 +10,51 @@ suite('oculus-touch-controls', function () {
...
+    setup(function (done) {
+      component = this.el.components['oculus-touch-controls'];
+      controllerSystem = this.el.sceneEl.systems['tracked-controls-webvr'];
+      addEventListenersSpy = sinon.spy(component, 'addEventListeners');
+      injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
+      removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
+      done();
     });
 
-    test('does not call removeEventListeners multiple times', function () {
```
dmarcos: Same for this. Is this test intentionally removed?

tests/components/oculus-touch-controls.test.js
DigiTec: This test was invalid. The state it was testing for couldn't happen in production code. The correct state is that you add then remove a controller and then remove event listeners is called, which is covered by test 'removes event listeners if controller disappears' on line 90.

tests/components/oculus-touch-controls.test.js
```diff
@@ -10,38 +10,51 @@ suite('oculus-touch-controls', function () {
...
   });
 
   suite('checkIfControllerPresent', function () {
-    test('removes event listeners if controllers not present', function () {
-      var addEventListenersSpy = sinon.spy(component, 'addEventListeners');
-      var injectTrackedControlsSpy = sinon.spy(component, 'injectTrackedControls');
-      var removeEventListenersSpy = sinon.spy(component, 'removeEventListeners');
-
-      // Mock has not been checked previously.
-      delete component.controllerPresent;
```
DigiTec: Specifically, this line sets the value of controllerPresent to undefined, but this is never the case in the real code. The "init" call sets it to false and attachment sets it to true, so setting it to undefined in order to fail a strict equality check was just a buggy test.

tests/components/oculus-touch-controls.test.js
DigiTec: This is redundant with the base state that removeEventListeners is not called if the controller is not present. This is covered by test 'returns not present if no controllers on first call'. This is also the same set of tests for oculus-go-controls. This is meant to standardize down to the correct set of compliant tests for these components so they can be updated more easily, specifically, these improvements are meant to make adding Quest controller support easier (I'm going to upgrade the oculus-touch-controls for this since Quest controllers are really just generation 2 of Oculus Touch and share nearly all of the same behavior except for the models).

DigiTec:
@dmarcos I left comments and explained where the tests simplified and removed. Let me know if you have any other issues or disagree. Thanks for the sharp eye on the review. I appreciate it!

tests/components/oculus-touch-controls.test.js
dmarcos: I like the idea of consolidating Quest / Rift to `oculus-controls` but two reasons against:

- Buttons animation: There's code in the component that is model dependent and would have to be duplicated making the component more complicated. Debugging, bug diagnosis, maintenance will be harder.
- Tool tips and instructions: Separate components make easier and consistent with other controllers for an application to discriminate and display instructions and tooltips. 

At the moment, I favor keeping separate `oculus-touch-controls` and `oculus-quest-controls` even though there's some not ideal code and test duplication.

dmarcos:
Thanks!

PR #4069, opened 2019-03-08
don't ignore boolean values in flushToDOM

hars-shhh:
**Description:**
flushToDOM is not working for boolean values
eg: if you flushToDOM visible component it won't work for false value
This is piece of code that has the problem
```javascript
flushToDOM: function (isDefault) {
    var attrValue = isDefault ? this.data : this.attrValue;
    if (!attrValue) { return; }
    window.HTMLElement.prototype.setAttribute.call(this.el, this.attrName, this.stringify(attrValue));
},
```

**Changes proposed:**
My suggestion to add a boolean check for the attrValue

```javascript
flushToDOM: function (isDefault) {
    var attrValue = isDefault ? this.data : this.attrValue;
    //  don't ignore boolean values
    if (!attrValue && typeof attrValue !== 'boolean') { return; }
    window.HTMLElement.prototype.setAttribute.call(this.el, this.attrName, this.stringify(attrValue));
},
```

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -240,8 +240,7 @@ Component.prototype = {
    */
   flushToDOM: function (isDefault) {
     var attrValue = isDefault ? this.data : this.attrValue;
-    //  don't ignore boolean values
-    if (!attrValue && typeof attrValue !== 'boolean') { return; }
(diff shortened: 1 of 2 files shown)


ngokevin:
Thanks!

PR #4072, opened 2019-03-10
Fixes #4070: Replace .setTexturePath() call with .setResourcePath()

### LINKED ISSUE #4070: "Warning: Tried to remove `Object3D` of type:  mesh which was not defined." How to avoid it?
**Description:**
**Every** time we load entities into a scene at runtime we get the following warning, which is a mess with >100 entities: 

> THREE.MTLLoader: .setTexturePath() has been renamed to .setResourcePath(). aframe.min.js:442:484

> core:a-entity:warn Tried to remove `Object3D` of type:  mesh which was not defined.

Can you tell me how to avoid this warning? 

- A-Frame Version: 0.9 
- Platform / Device: MacOS Mojave

tomegz:
Also add meaningful variable name for basePath for improved readability.

**Description:**

- replace `.setTexturePath()` call with `.setResourcePath()` (method was renamed),
- move basePath to a meaningful variable for improved readability

### DIFF
--- a/src/components/obj-model.js
+++ b/src/components/obj-model.js
@@ -48,13 +48,14 @@ module.exports.Component = registerComponent('obj-model', {
     var mtlLoader = this.mtlLoader;
     var objLoader = this.objLoader;
     var rendererSystem = this.el.sceneEl.systems.renderer;
+    var BASE_PATH = mtlUrl.substr(0, mtlUrl.lastIndexOf('/') + 1);
 
     if (mtlUrl) {
       // .OBJ with an .MTL.
       if (el.hasAttribute('material')) {
         warn('Material component properties are ignored when a .MTL is provided');
       }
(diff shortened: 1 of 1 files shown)


src/components/obj-model.js
```diff
@@ -48,13 +48,14 @@ module.exports.Component = registerComponent('obj-model', {
     var mtlLoader = this.mtlLoader;
     var objLoader = this.objLoader;
     var rendererSystem = this.el.sceneEl.systems.renderer;
+    var basePath = mtlUrl.substr(0, mtlUrl.lastIndexOf('/') + 1);
```
dmarcos: Rename to BASE_PATH. Convention used for consts

dmarcos:
Thanks!

src/components/obj-model.js
tomegz: Done 👍

dmarcos:
Thanks and congrats for your first PR 🎆

PR #4073, opened 2019-03-11
Quest controller prep

DigiTec:
**Description:**

The generation 2 touch controllers used by the Oculus Quest will have all of the same buttons and will only differ in the appearance of the model. For this reason, we are factoring the code to allow for only key attributes such as the model url, ray origin and pivot points to be changed. The rest of the model should have similar button mesh structures and so all of the same code should be used.

**Changes proposed:**
- Factor existing globals for the touch controller model into a set configured by controllerType
- Mirror the gen1 config to the gen2 config until we can upload final models.
- Detect the device type and change the displayModel used for Oculus Quest.
- Update code to use the displayModel properties for ray cast origin and pivot point.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -14,30 +14,34 @@ var TOUCH_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/oculus/
 // For now the generation 2 model is the same as the original until a new one is prepared for upload.
 var TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL = TOUCH_CONTROLLER_MODEL_BASE_URL;
 
-var CONTROLLER_DEFAULT = 'oculus_touch_gen1';
+var CONTROLLER_DEFAULT = 'oculusTouchGen1';
 var CONTROLLER_PROPERTIES = {
-  oculus_touch_gen1: {
+  oculusTouchGen1: {
     left: {
       modelUrl: TOUCH_CONTROLLER_MODEL_BASE_URL + 'left.gltf',
-      rayOrigin: {origin: {x: 0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, -0.053)
+      rayOrigin: {origin: {x: 0.008, y: -0.01, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0, 0, -0.053),
+      modelPivotRotation: new THREE.Euler(0, 0, 0)
     },
     right: {
       modelUrl: TOUCH_CONTROLLER_MODEL_BASE_URL + 'right.gltf',
-      rayOrigin: {origin: {x: -0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, -0.053)
+      rayOrigin: {origin: {x: -0.008, y: -0.01, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0, 0, -0.053),
+      modelPivotRotation: new THREE.Euler(0, 0, 0)
     }
   },
-  oculus_touch_gen2: {
+  oculusTouchGen2: {
     left: {
-      modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'left.gltf',
-      rayOrigin: {origin: {x: 0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, -0.053)
+      modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'gen2-left.gltf',
+      rayOrigin: {origin: {x: -0.01, y: 0, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0, 0.012, 0),
+      modelPivotRotation: new THREE.Euler(-Math.PI / 4, 0, 0)
     },
     right: {
-      modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'right.gltf',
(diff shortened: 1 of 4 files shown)


src/components/oculus-touch-controls.js
```diff
@@ -7,18 +7,39 @@ var checkControllerPresentAndSetup = trackedControlsUtils.checkControllerPresent
...
+var TOUCH_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/oculus/oculus-touch-controller-';
+// For now the generation 2 model is the same as the original until a new one is prepared for upload.
+var TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL = TOUCH_CONTROLLER_MODEL_BASE_URL;
 
-var DEFAULT_MODEL_PIVOT_OFFSET = new THREE.Vector3(0, 0, -0.053);
-var RAY_ORIGIN = {
-  left: {origin: {x: 0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}},
-  right: {origin: {x: -0.008, y: -0.004, z: 0}, direction: {x: 0, y: -0.8, z: -1}}
+var CONTROLLER_DEFAULT = 'oculus_touch_gen1';
+var CONTROLLER_PROPERTIES = {
```
dmarcos: Is not gen1 / gen2 more of an Oculus nomenclature? Wouldn’t be more clear for developers to call it Rift and Quest?

src/components/oculus-touch-controls.js
```diff
@@ -34,6 +55,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     buttonTouchColor: {type: 'color', default: '#8AB'},
     buttonHighlightColor: {type: 'color', default: '#2DF'},  // Light blue.
     model: {default: true},
+    controllerType: {default: 'auto', oneOf: ['auto', 'oculus_touch_gen1', 'oculus_touch_gen2']},
```
dmarcos: Maybe not needed. Any scenario
where someone wants to support just one of them?

src/components/oculus-touch-controls.js
```diff
@@ -189,12 +229,12 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
 
     // Offset pivot point
-    controllerObject3D.position.copy(DEFAULT_MODEL_PIVOT_OFFSET);
+    controllerObject3D.position.copy(this.displayModel[this.data.hand].modelPivotOffset);
 
     this.el.emit('controllermodelready', {
       name: 'oculus-touch-controls',
       model: this.data.model,
-      rayOrigin: RAY_ORIGIN[this.data.hand]
+      rayOrigin: this.displayModel[this.data.hand].rayOrigin
```
dmarcos: Maybe pass the controller type (Rift, Quest) as well?

src/components/oculus-touch-controls.js
tcs-ulli: Rift S and Quest have same Controllers

ngokevin:
@DigiTec Any ideas on differentiating touch and quest controllers?

dmarcos:
@DigiTech Congrats on the Quest Launch! A-Frame works fantastically on the Oculus browser. If you can share a gltf model of the controllers we can ship this PR. Thanks for your help.

DigiTec:
My proposal for the controller model selection is that we deliver the
details in the payload as suggested in the thread. This will solve two
cases, the one we are in with Quest/Rift S/Rift where these are all Oculus
Touch controllers but with some visual differences. The second case it will
solve is specifying a model override. While testing improvements to the old
Rift model it was not possible for me to iterate quickly and locally
because of the way we fetch directly from the CDN. Being able to specify a
model override will be useful for the debugging case, but also if someone
has authored a better model for their scenario (performance, capabilities,
or some other vector of improvement I haven't thought about).

The current glTF I have was not good. We have to iterate more on the model.
I've been out of office for a bit and have been unable to drive all of the
model fixes needed. I hope to jump on that next week. I suspect it will be
a couple more weeks before we have a useful model (one where we can
properly articulate the joints/buttons, and highlight etc...)

dmarcos:
@DigiTec Thanks for the update and welcome back. 

Using an alternative model should be possible by doing `el.setAttribute('oculus-touch-controls', 'model', false);` and then adding your own model `el.setAttribute('gltf-model', 'url(...)');`. Overriding the built-in one might be tricky. For buttons highlight and animations to work, the user's model would have to have same parts and ids that would have to document and maintain moving forward.

nrgapple:
Are there any updates on the issue? It seems we are still missing models for the quest.

dmarcos:
No models yet. You should be able to use [oculus-touch-controls](https://aframe.io/docs/0.9.0/components/oculus-touch-controls.html#sidebar) in the meantime (with old touch models that you can replace). Buttons mapping remains the same.

ganeshv:
FWIW Firefox Reality has Quest controller models here: https://github.com/MozillaReality/FirefoxReality/tree/master/app/src/oculusvr/assets

dmarcos:
@DigiTec Are there any official Quest controllers models we can use to finish up this PR? Thanks for all your help.

DigiTec:
Having this conversation internally now. We need to build some. The set we have isn’t compatible with standard three.js rendering.

dmarcos:
> Having this conversation internally now. We need to build some. The set we have isn’t compatible with standard three.js rendering.

@DigiTec Thanks for quick update. We use the [glTF viewer](https://gltf-viewer.donmccurdy.com/) to make sure that glTF models work in THREE / A-Frame. If possible, separate parts to animate / highlight would be great. The [Touch v1 models](https://github.com/aframevr/assets/blob/master/controllers/oculus/oculus-touch-controller-left.gltf) we use are a good example. Let me know if I can do anything to help.

dmarcos:
@knochi has kindly contributed gen2 models on https://github.com/aframevr/assets/pull/34 If they look good for everybody, we can use them with this PR in the interim until official models are available.

Knochi:
I think these are as official as the gen1 ones. They are from the MozillaReality Repo

https://github.com/aframevr/assets/tree/master/controllers/oculus

dmarcos:
I integrated the models that @Knochi kindly made based on the [Mozilla Reality](https://github.com/MozillaReality/FirefoxReality/tree/master/app/src/oculusvr/assets) ones. I also adjusted the model pivot and `laser-controls` positions accordingly. Thanks @DigiTec and @Knochi for the beautiful work. It's much appreciated.

PR #4082, opened 2019-03-14
if el is undefined on remove child entity, call removeFromParent

edsilv:
**Description:**

We're trying to get aframe to work in a stencil.js render method.

Before adding this fix we were seeing the following error when entities were removed from the VDOM:

`Cannot read property 'object3D' of undefined`

This was caused by `el` being undefined here:

```
remove: {
  value: function (el) {
    this.object3D.remove(el.object3D);
  }
},
```

**Changes proposed:**

If `el` is undefined, use `this.removeFromParent()`:

```
remove: {
  value: function (el) {
    if (el) {
      this.object3D.remove(el.object3D);
    } else {
      this.removeFromParent();
    }
  }
},
```

We're sure this is probably naive, but would love to hear why!

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -261,7 +261,9 @@ var proto = Object.create(ANode.prototype, {
       if (el) {
         this.object3D.remove(el.object3D);
       } else {
-        this.object3D.remove(...this.object3D.children);
+        while (this.object3D.children.length) {
+          this.object3D.remove(this.object3D.children[0]);
+        }
         this.parentNode.removeChild(this);
       }
     }
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -258,7 +258,11 @@ var proto = Object.create(ANode.prototype, {
    */
   remove: {
     value: function (el) {
-      this.object3D.remove(el.object3D);
+      if (el) {
+        this.object3D.remove(el.object3D);
+      } else {
+        this.object3D.remove(...this.object3D.children);
+      }
     }
   },
 
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -261,9 +261,6 @@ var proto = Object.create(ANode.prototype, {
       if (el) {
         this.object3D.remove(el.object3D);
       } else {
-        while (this.object3D.children.length) {
-          this.object3D.remove(this.object3D.children[0]);
-        }
         this.parentNode.removeChild(this);
       }
     }
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -262,6 +262,7 @@ var proto = Object.create(ANode.prototype, {
         this.object3D.remove(el.object3D);
       } else {
         this.object3D.remove(...this.object3D.children);
+        this.parentNode.removeChild(this);
       }
     }
   },
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -1522,6 +1522,25 @@ suite('a-entity component lifecycle management', function () {
     el.play();
     assert.equal(el.sceneEl.behaviors.tock.indexOf(testComponentInstance), -1);
   });
+
+  suite('remove', function () {
+    test('detaches if called with no arguments', function (done) {
+      el.remove();
+      setTimeout(() => {
+        assert.notOk(el.parentNode);
+        done();
+      });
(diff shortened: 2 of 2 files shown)


edsilv:
Here's an example of the render method:

```
  render(): JSX.Element {
    return (
      <a-scene background="color: #ECECEC">
        {
          this.boxEnabled ? <a-entity id='box' geometry="primitive: box;" position="-1 0.5 -3" rotation="0 45 0" material="color: #4CC3D9" shadow></a-entity> : null
        }
        <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4" shadow></a-plane>
      </a-scene>
    )
  }
```

If we use an empty `a-entity` instead of `null` when `boxEnabled` is false we don't see the error. However, we think this would lead to a build-up of empty entities and isn't ideal for our use case where we'll be adding an arbitrary number of entities to our scene.

src/core/a-entity.js
```diff
@@ -258,7 +258,11 @@ var proto = Object.create(ANode.prototype, {
    */
   remove: {
     value: function (el) {
-      this.object3D.remove(el.object3D);
+      if (el) {
+        this.object3D.remove(el.object3D);
+      } else {
+        this.object3D.remove(...this.object3D.children);
```
dmarcos: Syntax not correct:

`this.object3d.remove(...this.object3D.children);`

edsilv: We needed to add `this.parentNode.removeChild(this)` too, otherwise the nodes weren't being deleted from the DOM.

dmarcos: Notice the three dots. That's not correct JS syntax: ` this.object3D.remove(...this.object3D.children);`

edsilv: Have removed the spread operator. Using a while loop instead to remove each child Object3D.

ngokevin:
Try just toggling the visibility? That's better in terms of perf anyways, and I don't know if we want to have A-Frame hack around VDOM when we don't recommend that path. Perhaps also trying an empty string instead of null?

edsilv:
Our project is to allow the annotation of 3D objects using a set of measuring and labeling tools. The box example above is really just a simplified scenario. Initially we used plain three.js with a 3d widgets library that seemed suitable, but soon this became very difficult to debug. 

We refactored to use aframe instead of plain three.js which gave us extra declarative expressivity and meant we could use a familiar reactive redux + jsx pattern. However, we then ran into the above issue. As the collection of tools dynamically varies in content I'm not sure setting visibility per entity will work.

We read your comments on aframe-react, but the problem is that our app includes regular HTML forms that also use the same redux store. Ideally we want to be able to have a single source of truth for everything to aid in debugging.

Right now with our hack it seems to be working pretty nicely, except for another gotcha that `registerSystem` isn't working, whereas `registerComponent` does. This seems to have been solved in aframe-react by wrapping `a-entity` in a react component. We didn't really want to go down that route though, preferring to keep our aframe tags generic. We don't think it really matters as our "system" is redux, and we're mainly just using aframe for declarative rendering and the nice component model.

I sounds like we'll need to maintain our own fork with the hack? We're huge fans of aframe, but it sounds like our use case is slightly at odds with the project's direction?

ngokevin:
Couple options are to toggle the visibility instead of attach / detach, try using an empty string, or you can still have a React app communicate to an A-Frame app without placing A-Frame within the React tree.

edsilv:
I think toggling visibility/using an empty string would work if we were rendering a list of fixed length. However, the list can grow/shrink dynamically to any size, so having say 100 elements that we selectively enable/disable isn't a good fit.

So far our approach is proving robust and a good developer debugging experience. The ideal outcome for us is to be able to treat A-Frame elements like other custom elements rendered with jsx in stencil.

Just checking https://github.com/aframevr/aframe/issues/567 It seems like custom elements are on the roadmap? If we're just extending the DOM with A-Frame, isn't it arguable that these tags should function like any other when used with jsx/vdom?

ngokevin:
BTW, you can also use object pooling: https://aframe.io/docs/0.9.0/components/pool.html which allows for dynamically creating and recycling entities from a fixed pool (that can expand if needed). This is better for perf than continually creating and destroying entities.

ngokevin:
Thanks!

PR #4085, opened 2019-03-15
Make vrdisplayactivate driven enter-vr path more robust

dmarcos:
When `vrdisplayactivate` is handled the scene and its renderer might no be initialized. The new handler stores the vrDisplay passed to the event for `a-scene` to use during initialization.

### DIFF
--- a/src/core/scene/loadingScreen.js
+++ b/src/core/scene/loadingScreen.js
@@ -9,22 +9,6 @@ var getSceneCanvasSize;
 var ATTR_NAME = 'loading-screen';
 var LOADER_TITLE_CLASS = 'a-loader-title';
 
-// It catches vrdisplayactivate early to ensure we can enter VR mode after the scene loads.
-window.addEventListener('vrdisplayactivate', function () {
-  var vrManager = sceneEl.renderer.vr;
-  var vrDisplay;
-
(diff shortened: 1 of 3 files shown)


src/core/scene/loadingScreen.js
```diff
@@ -10,19 +10,13 @@ var ATTR_NAME = 'loading-screen';
...
   if (navigator.xr) { return; }
-
-  vrDisplay = utils.device.getVRDisplay();
-  vrManager.setDevice(vrDisplay);
-  vrManager.enabled = true;
-  if (!vrDisplay.isPresenting) {
-    return vrDisplay.requestPresent([{source: sceneEl.canvas}]).then(function () {}, function () {});
-  }
+  window.AFRAME = window.AFRAME || {};
+  var vrDisplay = window.AFRAME.vrDisplay = evt.display;
```
ngokevin: I'd store this in a module versus checking `window` everywhere or attempting to initialize window.AFRAME in a util. 

I don't think it needs to be a global on AFRAME.

`core/vrDisplay.js`

`module.exports = {vrDisplay: null};`

`require('core/vrDisplay').vrDisplay = vrDisplay`

PR #4086, opened 2019-03-15
Check for navigator.xr.requestDevice to determine if WebXR API is available

dmarcos:
All versions of Chrome now define `navigator.xr` even with webxr flags disabled and both in Windows and Mac (as reported by some users)

image

Checking for `navigator.xr.requestDevice` seems to be a more reliable way to determine if WebXR API is present and available.


dmarcos:
Problem seems to be dipper. WebXR API has changed in chrome

PR #4094, opened 2019-03-20
Get rid of warning message due to deprecation of renderTarget argument on renderer.render method

dmarcos:


netpro2k:
While making this same change in our fork I noticed the tests still expect `renderTarget` to be null (https://github.com/aframevr/aframe/blob/master/tests/core/scene/a-scene.test.js#L69). The test check should probably just be removed, since I agree removing `renderTarget` is the right call (if you want to mess with that you should just get the renderer directly and do it yourself).

PR #4103, opened 2019-03-27
Add update method for gltf-model system

idrumgood:
**Description:** gltf-model System would fail to load Draco decompressor if the system initialized before the scene was present. This was most apparent using aframe-react but will handle all cases of dynamically adding the gltf-model System to any scene through `.setAttribute`

**Changes proposed:**
- Add an update method to the gltf-model System

### DIFF
--- a/src/systems/gltf-model.js
+++ b/src/systems/gltf-model.js
@@ -20,6 +20,13 @@ module.exports.System = registerSystem('gltf-model', {
     this.dracoLoader = path ? new THREE.DRACOLoader() : null;
   },
 
+  update: function () {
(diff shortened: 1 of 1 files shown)


donmccurdy:
Is this working for you? The last time I tried, systems didn't support updates.

idrumgood:
@Dongeg Yea, making this change in my local node_modules/aframe-master.js fixes the issue I was having. Draco compression loads, my model loads. The basic `System` has an `update` method stubbed out.

ngokevin:
systems have updates

Issue #4081, opened 2019-03-14
A-link fails in 0.9.0 in Oculus GO Browser when next scene is too heavy


PlumCantaloupe:
**Description:**
In A-Frame 0.8.2 one could recently traverse using the link component between scenes on the oculus GO using the Oculus browser, even if they were "heavier" scenes. Now in 0.9.0 the user will be stuck within a spinning ball screen waiting indefinitely for the next scene to load. If we then exit out of VR mode, refresh, and go back in it does load.

Below are the console logs from the GO when (1) first entering the "Reproducible Code Snippet or URL" given below, then (2) clicking on the link to move to load scene2, then (3) clicking in scene2's link to come back to the original scene where it gets stuck in the loading screen.

I have yet to dive into the code too far, but it seems to stem from THREE.js trying to request a different "size while VR device is presenting."

EDIT: with new changes, Vive now work with Firefox, and Oculus now just kicks out to 2D mode when entering a hyperlink.

**1.** <img width="1016" alt="Screen Shot 2019-03-14 at 12 28 32 AM" src="https://user-images.githubusercontent.com/264985/54331824-0b918880-45f2-11e9-977c-26f7d7a7d44d.png">
**2.** <img width="1022" alt="Screen Shot 2019-03-14 at 12 28 54 AM" src="https://user-images.githubusercontent.com/264985/54331865-324fbf00-45f2-11e9-8aea-91a295b96107.png">
**3.** <img width="1022" alt="Screen Shot 2019-03-14 at 12 29 16 AM" src="https://user-images.githubusercontent.com/264985/54331878-3976cd00-45f2-11e9-8a0d-987b8376e9bc.png">

Any ideas? :)

- A-Frame Version: 0.9.0
- Platform / Device: Oculus GO / Oculus Browser
- Reproducible Code Snippet or URL: https://glitch.com/~portal-test-af-1

dmarcos:
`vrdisplayactivate` handler likely has a time limit to enter VR mode. We're maybe taking too long, stuck at loading and times out.

PlumCantaloupe:
Thanks, I will take a look there as that seems a fruitful direction. The only thing I can note is that I tried replacing the version of three.js all way back to r93 and it was the same issues (r92 broke AF.0.9.0 in other ways).

PlumCantaloupe:
[...]

PR #4107, opened 2019-04-02
A-Link Fix for Oculus Browser

PlumCantaloupe:
**Description:**
The Oculus GO browser appears to refuse permission on vrDisplay.requestPresent if the correct canvas is not passed in. This causes issues when using hyperlinks ( see issue #4081 ) as A-Frame currently handles the "vrdisplayactivate" event by passing in a dummy canvas in Device.js.

**Changes proposed:**
To make the requestPresent method the same in both src/core/scene/a-scene.js and src/utils/Device.js so that some browsers (specifically the Oculus Browser on the Oculus GO) do not reject the presentation. To maintain consistency the presentation attributes are also passed in again.

_This would probably be better as a public method in device.js that both a-scene and device can call to reduce code redundancy; but I was unsure of how the A-Frame maintainers prefer to handle global functions._

**Needs Review:**
My PC died last night so this needs to be tested with other devices also (i.e. Firefox and the Vive / Oculus Rift) as they appear to be okay with the dummy canvas pass.

NOTE: I also fixed a formatting issue in an unrelated file so that I could commit.

### DIFF
--- a/src/systems/gltf-model.js
+++ b/src/systems/gltf-model.js
@@ -21,7 +21,7 @@ module.exports.System = registerSystem('gltf-model', {
   },
 
   update: function () {
-    var path;	  
+    var path;
     if (this.dracoLoader) { return; }
     path = this.data.dracoDecoderPath;
     THREE.DRACOLoader.setDecoderPath(path);
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -4,13 +4,33 @@ var vrDisplay;
 
 // Catch vrdisplayactivate early to ensure we can enter VR mode after the scene loads.
 window.addEventListener('vrdisplayactivate', function (evt) {
-  var canvasEl;
   // WebXR takes priority if available.
   if (navigator.xr) { return; }
-  canvasEl = document.createElement('canvas');
+
   vrDisplay = evt.display;
+
+  const scene = document.querySelector('a-scene');
+  const rendererSystem = scene.getAttribute('renderer');
+  const canvas = scene.canvas; // Oculus Browser will fail if we don't provide correct canvas VRLayer
+
+  // Not sure if required but seems good practise to stay consistent on request present
+  const presentationAttributes = {
+    highRefreshRate: rendererSystem.highRefreshRate,
+    foveationLevel: rendererSystem.foveationLevel
+  };
+
   // Request present immediately. a-scene will be allowed to enter VR without user gesture.
-  vrDisplay.requestPresent([{source: canvasEl}]).then(function () {}, function () {});
+  return vrDisplay.requestPresent([{
+    source: canvas,
+    attributes: presentationAttributes
+  }]).then(function () { console.warn('Succeeded to enter VR mode (`requestPresent`)'); },
+    function (err) {
+      if (err && err.message) {
+        throw new Error('Failed to enter VR mode (`requestPresent`): ' + err.message);
+      } else {
+        throw new Error('Failed to enter VR mode (`requestPresent`).');
+      }
+    });
 });
 
 // Support both WebVR and WebXR APIs.


PlumCantaloupe:
cc @Artyom17 @DigiTec

PlumCantaloupe:
Ah, I see a good point about the scene. Perhaps it is worth checking for scene loaded, and if not doing a 'loaded' callback? 

``` javascript
// !!not tested yet
// Catch vrdisplayactivate early to ensure we can enter VR mode after the scene loads.
window.addEventListener('vrdisplayactivate', function (evt) {
  // WebXR takes priority if available.
  if (navigator.xr) { return; }

  const scene = document.querySelector('a-scene');
  vrDisplay = evt.display;

  // make it a function we can use as callback ...
  const requestPresentFunc = () => {
    const rendererSystem = scene.getAttribute('renderer');
    const canvas = scene.canvas; // Oculus Browser will fail if we don't provide correct canvas VRLayer

    // Not sure if required but seems good practise to stay consistent on request present
    const presentationAttributes = {
      highRefreshRate: rendererSystem.highRefreshRate,
      foveationLevel: rendererSystem.foveationLevel
    };

    // Request present immediately. a-scene will be allowed to enter VR without user gesture.
    return vrDisplay.requestPresent([{
      source: canvas,
      attributes: presentationAttributes
    }]).then(function () { console.warn('Succeeded to enter VR mode (`requestPresent`)'); },
      function (err) {
        if (err && err.message) {
          throw new Error('Failed to enter VR mode (`requestPresent`): ' + err.message);
        } else {
          throw new Error('Failed to enter VR mode (`requestPresent`).');
        }
      });
    };

  // is scene exist requestPresent .. else wait until it is
  if ( scene.hasLoaded ) {
    requestPresentFunc();
  }
  else {
    scene.addEventListener('loaded', requestPresentFunc);
  }  
});
```

As discussed in the issue referenced above ( #4081 )  it is true that the link traversal examples do not trigger this behaviour because they are too simple. With a heavier scene this behaviour becomes unavoidable. Try the [glitch from the issue](https://glitch.com/~portal-test-af-1).

Also, though I understand this would not be under your control, is it not strange that we call a function like [requestPresent](https://developer.mozilla.org/en-US/docs/Web/API/VRDisplay/requestPresent) without explicit parameters? (i.e. accepts dummy arguments). Seems ripe for fragmented implementation and misunderstanding.

dmarcos:
Current code in master is compliant with spec. I would prefer to understand why Oculus Browser is not working in your case. Maybe because two requestPresent calls are 5+ seconds apart and the first one is not honored and then the 2nd one fails?

PlumCantaloupe:
What do you mean by the second requestPresent? 

In master, I see it only called once when a UA triggers 'EnterVR" (in a-scene.js); and only once when "vrdisplayactivate" is called in Device.js (through console logging pathways when testing). I don't see any other calls to requestPresent in the library, or when it runs. Should requestPresent be called twice on link traversal?

Thanks for your patience @dmarcos. Doing my best to understand how all these moving pieces come together.

dmarcos:
In a vrdisplayactivate workflow requestPresent is [called twice](https://github.com/aframevr/aframe/search?utf8=✓&q=requestPresent&type=). First early so the even is not missed and avoid timeouts (Oculus seems to have a 5 sec limit to call requestPresent from handler) and then a second time when the scene has loaded and 
canvas and rest of components are ready:

https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L435

PlumCantaloupe:
Okay, hope we can hear from the Oculus team then. I just did a few more tests adding in some logs and timers into both master and my PR to compare output below with the following steps to give us an idea how both "light" and "heavy" scenes react.

1. I start on this ["heavy" scene on Glitch](https://portal-test-af-1.glitch.me)
2. I click on link (in VR mode with Oculus GO) to go to this ["light" scene on Glitch](https://portal-test-af-2.glitch.me)
3. Then click on that link to go back to the ["heavy" scene on Glitch](https://portal-test-af-1.glitch.me)

The below output for all three transitions suggests that it might be that tiomeout issue as the second enterVR is not ever called on master "heavy" scene.

You can play with code for both tests here (check commented out scripts for master script with logs)
- https://glitch.com/~portal-test-af-1
- https://glitch.com/~portal-test-af-2

### With **current master** we get this:

Master

### With **this alink-pr** we get:
_note that the type error for "isPresenting" happens sporadically and does not seem to affect anything one way or another in this PR or master._

PR

DigiTec:
@Artyom17 and I are going to vet this and get back to you. There is a timeout and a heavy page could hit the timeout and fall out of VR presentation. This has been an issue, but is critical to avoid bad end user behavior of a black screen.

Why it would fail to present to a new canvas after presenting to a temporary canvas is beyond me. Most likely this is some sort of exclusivity check or something.

ngokevin:
Might be a reason to consider an alternate API for navigation that doesn't rely on such tight timing? https://github.com/immersive-web/navigation/pull/2#issuecomment-478738120

PlumCantaloupe:
So I am coming back to this hoping that the Oculus browser has been updated to handle this case but now this workaround doesn't work either on the Oculus Quest Browser (amazing device and browser for WebVR btw!).

I would eternally grateful for any workaround/hack on this issue as I am hoping to run some studies this summer that will allow users to move between worlds, but _I am completely blocked_ as they keep popping out of VR on link traversal with these "heavy" scenes :(

Any ideas @DigiTec or @Artyom17?

The errors I get on the quest for this PR (updated to current A-Frame master) now is telling me I need a uder gesture:
Screen Shot 2019-05-28 at 5 52 17 PM

Using base master I first get a rendering context error (like before), then the above erro:
Screen Shot 2019-05-28 at 5 52 12 PM
Screen Shot 2019-05-28 at 5 52 17 PM

PlumCantaloupe:
Huge thanks to @Artyom17 and @dmarcos  this has been fixed by PR #4199.

PR #4118, opened 2019-04-14
Update globals.md

weddingdj:
**Description:**
Removed tween.js since it is not used anymore.

**Changes proposed:**
- Removed tween.js since it is not used anymore.
-
-


ngokevin:
Thanks!

PR #4121, opened 2019-04-16
keep components attached to entity if whole entity is detached (fixes #4027)

### LINKED ISSUE #4027: geometry primitive undefined so fallback to box
**Description:**

- A-Frame Version: 0.9.0
- Platform / Device: Chrome on desktop

I have simple code that do
```
const plane = document.createElement("a-plane");
plane.setAttribute("height", 0.196);
plane.setAttribute("width", 0.196);
AFRAME.scenes[0].appendChild(plane)
```
initially it shows a plane, now if I remove the plane element `AFRAME.scenes[0].removeChild(plane)` and recreate it, it will take an object from the pool and it will sometime show a box instead of a plane. Somehow I get "primitive: undefined" in the object.

Putting a `console.log(this.attrValue, this.data)` in `src/core/component.js` `initComponent`, I get initially 
```
attrValue: {width: "0.196", height: "0.196", primitive: "plane"}
data: {
  buffer: true
  height: 0.196
  primitive: "plane"
  segmentsHeight: 1
  segmentsWidth: 1
  skipCache: false
  width: 0.196
}
```

and if it take an existing object from the pool (when I remove plane and recreate it), I can get this:
```
attrValue: {width: "0.196", height: "0.196", primitive: undefined}
or attrValue: {height: "0.196", width: "0.196", primitive: undefined, buffer: undefined, skipCache: undefined, …}
data: {
  buffer: true
  depth: 1
  height: 0.196
  primitive: "box"
  segmentsDepth: 1
  segmentsHeight: 1
  segmentsWidth: 1
  skipCache: false
  width: 0.196
}
````
and other time:
`{radius: undefined, width: 0.196, height: 0.196, primitive: "plane", buffer: true, …}`
triggering a warning:
core:schema:warn Unknown property `radius` for component/system `geometry`.  see #4012 for this one.

Now If I change the implementation of `clearObject` to delete the keys:
`for (key in obj) { delete obj[key]; }`
all the issues goes away.

The issue comes from the commit
https://github.com/aframevr/aframe/commit/6deaea0564d18c91b94d801db0e39535b094f044
from the component.js file. I tested master with the component.js file before this commit and I don't have the issue. The PR #4016 doesn't fix the issue.

ngokevin:
**Description:**

Preserves component data if entity is detached so entity could be reattached okay. If entity is detached, it should run component pause/remove callbacks, but can still keep the component reference + data.

If a component is explicitly removed, then we destroy the component and its data.

- [x] Need to figure out how to possibly avoid memory leak if developer detaches an entity, but then does nothing with it, need to return the component objects to pool. Do we need an `Entity.destroy()`? If this use case is supported (detach entity and attach somewhere else in scenegraph), then do we need components to have a separate destroy handler? Or do we just not call `Component.remove` unless it is being destroyed and count the `remove` as the destroy handler.

**Changes Proposed:**

- Allow entities to be detached and reattached while maintaining component data.
- `Entity.destroy()` method for when you detach an entity and never want to use it anymore, you can call the `.destroy()` method.

### DIFF
--- a/docs/core/entity.md
+++ b/docs/core/entity.md
@@ -320,6 +320,10 @@ object:
 entity.setAttribute('position', { x: 1, y: 2, z: 3 });
 ```
 
+### `destroy ()`
+
+Clean up memory related to the entity such as clearing all components and their data.
+
 #### Updating Multi-Property Component Data
 
 To update component data for a multi-property component, we can pass the name
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -100,7 +100,9 @@ var proto = Object.create(ANode.prototype, {
       if (!this.parentEl) { return; }
 
       // Remove components.
-      for (componentName in this.components) { this.removeComponent(componentName); }
+      for (componentName in this.components) {
+        this.removeComponent(componentName, false);
+      }
 
       if (this.isScene) { return; }
 
@@ -373,7 +375,7 @@ var proto = Object.create(ANode.prototype, {
   },
 
   removeComponent: {
-    value: function (name) {
+    value: function (name, destroy) {
       var component;
 
       component = this.components[name];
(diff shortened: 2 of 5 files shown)


Moon-ChangHyun:
There seems to be a mistake in Entity.destroy().
In that method it seems that this.el.parentNode should be modified to this.parentNode.

PR #4123, opened 2019-04-16
Component.load handler (fixes #3556)

### LINKED ISSUE #3556: Consider stricter lifecycle guarantees
Got bit recently with an unexpected issue writing a component, thought I'd write it up for ideas...

The `nav-mesh` component needs the mesh from the entity it's attached to, and needs access to any transformations on its parents. I'd imagined this would be simple enough:

```js
AFRAME.registerComponent('nav-mesh', {
  init: function () {
    this.loadNavMesh();
    this.el.addEventListener('object3dset', () => this.loadNavMesh());
  },

  loadNavMesh: function () {
    const navMesh = this.el.getObject3D('mesh');

    if (!navMesh) return;

    const navMeshGeometry = navMesh.geometry.isBufferGeometry
      ? new THREE.Geometry().fromBufferGeometry(navMesh.geometry)
      : navMesh.geometry.clone();

    scene.updateMatrixWorld();
    navMeshGeometry.applyMatrix(navMesh.matrixWorld);
    this.system.setNavMeshGeometry(navMeshGeometry);
  }
});
```

Unfortunately this fails — when `init` runs, the entity's parents do not have their position/rotation/scale applied yet, so `scene.updateMatrixWorld()` does nothing, and the transformation isn't applied to the Geometry.

Next I assumed `update()` would be a safe place to put the `loadNavMesh()` call, but it too runs before parents are initialized. I don't want this code to run on `play` (it shouldn't reload when the scene pauses) so instead I ended up with:

```js
tick: function (t) {
  if (t === 0) this.loadNavMesh();
}
```

Anyway, all of this was a little more complicated than I expected, and harder than component lifecycles in e.g. Angular. Saw this issue (or similar) called out as a [pain point compared to Unity recently, too](http://portfolio.anthony-scavarelli.com/2018/04/24/thoughts-on-multi-user-vr-development-using-unity-and-aframe-webvr/):

> Have to wrestle with asynchronicity issues that you generally don’t run into with Unity (e.g. why is this model loading BEFORE something else it is dependent on …)

Anyway, I don't have a concrete proposal at the moment but wanted to think about tightening up component lifecycles in a future version and documenting clearly what component authors should expect.

ngokevin:
**Description:**

#3556 

The `.play()` handler is currently overloaded to handle both:

- Scene / component first load.
- Resume from pause or activated from pool.

More often than not, we care more about the first case, a handler to run the component after the scene has loaded or the component is attached to an already-running scene. Common cases are waiting for things to attach to run query selectors, waiting until other components in the scene have run, waiting until ancestors have initialized in order to get accurate world transforms, or just wanting to run something while scene is actually playing. 

But with the `.play()` handler as is, we sometimes need to check it's only called on first run so that it **doesn't** run after like Inspector open/close, or being request from pool. To solve the overload, I add a `.load()` handler similar to `.play()` that is ensured to only be called once.

@donmccurdy Thoughts?

**Changes proposed:**
- `Component.load()` handler similar to `Component.play()` but is only ensured to be called called once.
- [x] Tests / docs.

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -86,10 +86,12 @@ AFRAME.registerComponent('foo', {
   schema: {},
   init: function () {},
   update: function () {},
+  load: function () {},
+  play: function () {},
+  pause: function () {},
   tick: function () {},
   remove: function () {},
-  pause: function () {},
-  play: function () {}
+  events: {}
 });
 ```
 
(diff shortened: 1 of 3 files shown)


dmarcos:
> Common cases are waiting for things to attach to run query selectors, waiting until all initial components in scene has attached, or just wanting to run something while scene is actually playing.

Do we need both init and load? Can we just postpone `init` call until the scene has loaded? It might cover the use cases above

ngokevin:
It would almost work except the definition of the scene `loaded` is that all entities and their components have initialized. Would break most things if `loaded` would be emitted and no components have been set up yet.

I use `init` to set up variable and state, and would use `load` for things that need to wait until the scene is set up (e.g., you want to call `getWorldPosition` but parents need to init first).

donmccurdy:
Please don't postpone `init` – that would break a lot of code, and delay the overall setup process.

I don't personally think I'd need `load` all that often, using `play` has worked well enough. Often I don't care that much whether the scene itself is set up – what I need to know is that other components (even those higher in the tree than the current one) have already initialized.

What do you think of `mount` rather than `load`? Not a strong preference but it makes a nice parallel to React's `didMount` and `willUnmount`.

Another idea was mentioned in https://github.com/donmccurdy/aframe-physics-system/pull/125 – i'll point it out here but it's probably out of scope – to have `init` (or something like `load` I guess) return a promise that delays further traversal. Haven't used that pattern but I kind of like the idea.

ngokevin:
> what I need to know is that other components (even those higher in the tree than the current one) have already initialized.

Yeah, `.load` would ensure that and addresses your filed issue #3556: `The fact that I can't know the initial position of my entity until the first tick has already begun seems like a hazard.`. Sometimes you want to wait until things set up including ancestor components, their side effects, and transformations. `load` is just a catch-all way to do that.

Another big problem solved is that `.play` is called on initialization + scene resume + when entity is activated from an object pool. Often had code that had to check whether this `play` was the first instance of the call.

I prefer `load` since it points to when the component is loaded and running, which on first render means scene load. `mount` sounds like on very first component attach which wouldn't be accurate, since `load` would run later than that.

donmccurdy:
> Yeah, .load would ensure that and addresses your filed issue #3556

Ah good point 👍

ngokevin:
One example is here: https://github.com/supermedium/moonrider/commit/633b7f6932c1f631919255478ad920093a6c76e8#diff-bb259426f7c04432298f7755d7853bacR52

I prefer to create a reference to a scene component when the scene loads (i.e., `.load`). It is preferable not do it in `.play` because `.play` will get re-called every time the entity gets pulled from the pool, and I only need to create that reference once. Instead, I have to just reference at run-time. This is a simple example, but can take place with more complex operations you only want to run once on scene load.

dmarcos:
Points made really worth and might consider something like this in the future. Closing this PR for since it's stale.

PR #4126, opened 2019-04-17
Add optional Logo capability to loading screen

Srile:
**Description:**
This adds the option to add a logo to the loading screen component.


dmarcos:
The load screen should not depend on any assets or you would need a loading screen for the loading screen 😄 It's meant to fill the gap while the scenes assets load.

For 2D loading screen one can [modify the `a-loader-title` style](https://aframe.io/docs/0.9.0/components/loading-screen.html#sidebar)

Srile:
Alright, that's a good point. A seperate component is definitely a better solution for this.

PR #4146, opened 2019-04-30
Metatag to hide URL bar in iOS when entering fullscreen mode

dmarcos:
Holy grail metatag? Tested on iPhone X and iOS 12.2


eyexpo-dan:
I tried out `"aframe": "github:aframevr/aframe#master"` on iOS 12.2 with iPhone 6 Plus.

enterVR() still shows the URL bar.

dmarcos:
It works on iPhone X and Safari. Behavior variates across devices and screen sizes,!It seems only reliable way to get rid of the bar is still adding site to homescreen. It’s a browser behavior. There’s not much we can do about.

PR #4153, opened 2019-05-06
Update typo on Building a Scene Basics

jannaee:
**Description:**
Used the contraction and `lets us` where it should `let us`


ngokevin:
Thanks for the PR! I believe the current is correct as in:

"Oxygen lets us breathe".

lets is a present tense of let.

Issue #4164, opened 2019-05-12
Component ticks behave differently in Chrome vs Firefox when entities are removed


brianpeiris:
**Description:**
Chrome seems to run `detachedCallback` synchronously when an entity is removed. This causes component ticks to be removed from the `behaviors` array during the tick, which in turn causes ticks to be missed, since the array is being mutated while its being iterated.
Firefox on the other hand seems to call `detachedCallback` asynchronously, after a tick.

I've opened PR #4165 with a failing test case and you can see a visual demonstration of the bug by opening the following glitch link in Firefox and Chrome. The demonstration contains a scene with multiple entities. The entities all have a component on them that removes its own entity from the scene, and there is a component on the scene that pauses the scene after the first frame to show the skipped ticks.

- A-Frame Version: 0.92 & master
- Platform / Device: Windows & Android, Firefox and Chrome. Haven't tested Safari.
- Reproducible Code Snippet or URL: https://aframe-tick-bug.glitch.me/

dmarcos:
Thanks. Probably the difference is that Chrome has a registerElement implementation while Firefox runs on polyfill. Chrome is deprecating soon registerElement in favor of custom elements v1. Both browsers will run on polyfill soon

dmarcos:
Is this issue still relevant with https://github.com/aframevr/aframe/pull/4167 ?

brianpeiris:
Hmm, #4167 does make the behavior consistent across Chrome and Firefox. However, it does so by making the bug consistent -- i.e. it mutates the `behaviors` array while it's iterating it, so ticks end up being skipped when entities are removed.

I've setup a branch with my tests on top of #4167:
https://github.com/brianpeiris/aframe/tree/customElementsV1-tick-test

You can run the visual test yourself in Chrome and Firefox to see what I mean:
https://rawcdn.githack.com/brianpeiris/aframe/6e94225ebd01a90e7512a8294b0fbc21110e58c5/examples/test/ticks/index.html

brianpeiris:
To be clear. I expect this demo to remove all entities in the first tick, since they all have the `remove-self` component on them. If you open this link in Firefox, which uses a-frame 0.9.2, you'll see that it does the right thing: https://aframe-tick-bug.glitch.me/

PR #4167, opened 2019-05-13
Migrate to custom elements v1 API

dmarcos:
Migrate all A-Frame elements to [custom elements V1 API](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements). Both Firefox and Chrome ship implementations today. Chrome is soon deprecating the old `document.registerElement`. This PR will make A-Frame less reliant on the polyfill, behavior more consistent across browsers (https://github.com/aframevr/aframe/issues/4164) and likely will come with perf improvements as well. 

These changes touch most of the lines in a bunch of files. It would be good to merge at the beginning of the release cycle to have time to work out the kinks before 1.0. Examples work fine in both desktop Chrome (m74) and Firefox (66) as is.

The main drawback is that Custom Elements v1 enforces JS classes that are not available in older browsers.

Pending work:

- [x] Better test polyfill path.
- [x] Test in mobile browsers (Chrome Pixel 3 m74  and Safari iOS 12.2)
- [x] Port all unit tests.
- [x] Test with production complex apps (https://github.com/supermedium/beatsaver-viewer)


arpu:
nice! testing on my vrland.io project and it works without any problem

dmarcos:
I think this is ready to go. Tested on Edge (42), Firefox Desktop (66), Chrome Desktop (m74), Safari and Chrome for iOS 12.2,  Chrome for Android (m74), Oculus Browser, Supermedium and Exokit.

Only compatibility issue is with old IE due to Custom Elements V1 enforcing classes.

dmarcos:
There's a travis failing test that I don't understand. Not sure what `mocha --ui tdd tests/node` tests and why it's failing on the polyfill

src/core/a-entity.js
```diff
@@ -23,155 +22,156 @@ var ONCE = {once: true};
...
-      this.removeFromParent();
-      ANode.prototype.detachedCallback.call(this);
+    if (this.isScene) { return; }
 
-      // Remove cyclic reference.
-      this.object3D.el = null;
-    }
-  },
+    this.removeFromParent();
+    super.disconnectedCallback.call(this);
```
justinfagnani:
```suggestion
    super.disconnectedCallback();
```

src/core/a-node.js
```diff
@@ -1,269 +1,266 @@
...
+      self.hasLoaded = true;
+      if (cb) { cb(); }
+      self.emit('loaded', undefined, false);
+    }).catch(function (err) {
+      error('Failure loading node: ', err);
+    });
+  }
+
+  getChildren () {
+    return Array.prototype.slice.call(this.children, 0);
```
justinfagnani: Could be `[...this.children]`

src/core/a-entity.js
```diff
@@ -677,90 +633,81 @@ var proto = Object.create(ANode.prototype, {
...
+    var delimiterIndex;
+    var isDebugMode;
+    var key;
+    delimiterIndex = attrName.indexOf(MULTIPLE_COMPONENT_DELIMITER);
+    componentName = delimiterIndex > 0 ? attrName.substring(0, delimiterIndex) : attrName;
+
+    // Not a component. Normal set attribute.
+    if (!COMPONENTS[componentName]) {
+      if (attrName === 'mixin') { this.mixinUpdate(arg1); }
+      super.setAttribute.call(this, attrName, arg1);
```
justinfagnani:
```suggestion
      super.setAttribute(attrName, arg1);
```

src/extras/primitives/primitives.js
```diff
@@ -19,147 +20,147 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
...
+    }
+
+    /**
+     * Sync to attribute to component property whenever mapped attribute changes.
+     * If attribute is mapped to a component property, set the component property using
+     * the attribute value.
+     */
+    attributeChangedCallback (attr, oldVal, value) {
+      var componentName = this.mappings[attr];
+      super.attributeChangedCallback.call(this, attr, oldVal, value);
```
justinfagnani:
```suggestion
      super.attributeChangedCallback(attr, oldVal, value);
```

src/core/a-assets.js
```diff
@@ -10,126 +10,127 @@ var warn = debug('core:a-assets:warn');
...
+
+  disconnectedCallback () {
+    // ANode method.
+    super.disconnectedCallback();
+
+    if (this.timeout) { clearTimeout(this.timeout); }
+  }
+
+  load () {
+    super.load.call(this, null, function waitOnFilter (el) {
```
justinfagnani:
```suggestion
    super.load(null, function waitOnFilter (el) {
```

src/core/a-entity.js
```diff
@@ -180,112 +180,96 @@ var proto = Object.create(ANode.prototype, {
...
+  load () {
+    var self = this;
 
-      if (this.hasLoaded || !this.parentEl) { return; }
+    if (this.hasLoaded || !this.parentEl) { return; }
 
-      ANode.prototype.load.call(this, function entityLoadCallback () {
-        // Check if entity was detached while it was waiting to load.
-        if (!self.parentEl) { return; }
+    super.load.call(this, function entityLoadCallback () {
```
justinfagnani:
```suggestion
    super.load(function entityLoadCallback () {
```

src/core/scene/a-scene.js
```diff
@@ -34,709 +32,656 @@ var warn = utils.debug('core:a-scene:warn');
...
-        window.removeEventListener('vrdisplaydisconnect', this.exitVRTrueBound);
-        window.removeEventListener('vrdisplaypointerrestricted', this.pointerRestrictedBound);
-        window.removeEventListener('vrdisplaypointerunrestricted', this.pointerUnrestrictedBound);
+    // Default components.
+    this.setAttribute('inspector', '');
+    this.setAttribute('keyboard-shortcuts', '');
+    this.setAttribute('screenshot', '');
+    this.setAttribute('vr-mode-ui', '');
+
+    super.connectedCallback.call(this);
```
justinfagnani:
```suggestion
    super.connectedCallback();
```

src/core/a-mixin.js
```diff
@@ -9,106 +9,97 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
-     */
-    cacheAttribute: {
-      value: function (attr, value) {
-        var component;
-        var componentName;
+  /**
+   * setAttribute that parses and caches component values.
+   */
+  setAttribute (attr, value) {
+    window.HTMLElement.prototype.setAttribute.call(this, attr, value);
```
justinfagnani: You can call `super.setAttribute(attr, value)` without reaching to the prototype directly, but you might also consider doing this in `attributeChangedCallback` if you can.

src/core/a-mixin.js
```diff
@@ -9,106 +9,97 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
-        return this.componentCache[attr] ||
-               window.HTMLElement.prototype.getAttribute.call(this, attr);
-      }
-    },
+    // Get component data.
+    componentName = utils.split(attr, MULTIPLE_COMPONENT_DELIMITER)[0];
+    component = components[componentName];
+    if (!component) { return; }
+    if (value === undefined) {
+      value = window.HTMLElement.prototype.getAttribute.call(this, attr);
```
justinfagnani: `super.setAttribute(attr, value)` should work here too.

src/core/a-mixin.js
```diff
@@ -9,106 +9,97 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
+    this.componentCache[attr] = component.parseAttrValueForCache(value);
+  }
+
+  /**
+   * If `attr` is a component, then grab pre-parsed value from the cache.
+   * Else do a normal getAttribute.
+   */
+  getAttribute (attr) {
+    return this.componentCache[attr] ||
+      window.HTMLElement.prototype.getAttribute.call(this, attr);
```
justinfagnani: `super...` here as well

src/core/a-assets.js
```diff
@@ -10,126 +10,127 @@ var warn = debug('core:a-assets:warn');
...
-          loaded.push(new Promise(function (resolve, reject) {
-            // Set in cache because we won't be needing to call three.js loader if we have.
-            // a loaded media element.
-            THREE.Cache.files[imgEls[i].getAttribute('src')] = imgEl;
-            imgEl.onload = resolve;
-            imgEl.onerror = reject;
-          }));
-        }
+class AAssets extends ANode {
+  constructor (self) {
```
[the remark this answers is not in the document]
WebReflection: The correct form here is:
```js
constructor(self) {
  self = super(self);
}
```

Both native and polyfilled browsers will receive the right upgraded context.

As it is now, you lose that upgrade in polyfilled browsers calling just `super(self)` without addressing its return.

This is explained in here https://github.com/WebReflection/document-register-element#v1-caveat and it's been working just fine since about ever in every other library based on DRE polyfill (StencilJS, Google AMP, etc)

src/core/a-assets.js
```diff
@@ -10,126 +10,127 @@ var warn = debug('core:a-assets:warn');
...
+    super(self);
```
WebReflection: as explained in here https://github.com/aframevr/aframe/pull/4167/files/a3041384c73e829bfe91b110afa7e8a6bf30af5f#r285005172 you need to upgrade self to either itself or the returned node.

I also suggest to cut the mustard with a base class that does just this:
https://github.com/WebReflection/document-register-element#skipping-the-caveat-through-extends

so that you'll have a proper `init() {}` method that won't have ever issues and there's less to explain to newcomers.

src/core/a-assets.js
```diff
@@ -10,126 +10,127 @@ var warn = debug('core:a-assets:warn');
...
-        }, function handleOnError (xhr) {
-          self.emit('error', {xhr: xhr});
-        });
-      }
-    }
-  })
-});
+class AAssetItem extends ANode {
+  constructor (self) {
+    super(self);
```
WebReflection: same as above, please consider adding an `init`, or `created` method to `ANode` class

```js
class Anode extends HTMLElement {
  constructor(...$) { const _ = super(...$); _.init(); return _; }
  init() { /* override as you like */ }}
}

class AAssetItem extends ANode {
  init() {
    this.data = null;
    this.isAssetItem = true;
  }
}
```

src/core/a-cubemap.js
```diff
@@ -1,48 +1,43 @@
...
-      value: function () {
-        this.srcs = this.validate();
-      },
-      writable: window.debug
-    },
+class ACubeMap extends HTMLElement {
+  /**
+   * Calculates this.srcs.
+   */
+  constructor (self) {
```
WebReflection: same as in previous comments

src/core/a-mixin.js
```diff
@@ -9,106 +9,97 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
+  }
 
-    attributeChangedCallback: {
-      value: function (attr, oldVal, newVal) {
-        this.cacheAttribute(attr, newVal);
-        this.updateEntities();
-      }
-    },
+  constructor (self) {
+    super(self);
```
WebReflection: same as in previous comments

src/core/a-mixin.js
```diff
@@ -9,106 +9,97 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
+   */
+  cacheAttributes () {
+    var attributes = this.attributes;
+    var attrName;
+    var i;
+    for (i = 0; i < attributes.length; i++) {
+      attrName = attributes[i].name;
+      this.cacheAttribute(attrName);
+    }
+  }
```
WebReflection: following Justin hints, while it's not necessary to refactor everything at this point, an example on how above method could be improved is:

```js
const {attributes} = this;
for (let i = 0; i < attributes.length; i++) {
  const {name} = this.attributes[i]; // or {name, value}
  this.cacheAttribute(name);
}
```

src/extras/primitives/primitives.js
```diff
@@ -19,147 +20,147 @@ module.exports.registerPrimitive = function registerPrimitive (name, definition)
...
+  var mappings = definition.mappings || {};
+  var primitiveClass = class extends AEntity {
+    static get observedAttributes () {
+      return Object.keys(components)
+             .concat(Object.keys(mappings))
+             .concat(['mixin']);
+    }
+
+    constructor (self) {
+      super(self);
```
WebReflection: same gotcha with the `self = super(self)`

src/index.js
```diff
@@ -78,7 +80,7 @@ require('./components/index'); // Register standard components.
 require('./geometries/index'); // Register standard geometries.
 require('./shaders/index'); // Register standard shaders.
 require('./systems/index'); // Register standard systems.
-var ANode = require('./core/a-node');
+var ANode = require('./core/a-node').ANode;
```
WebReflection: const {ANode} = require('./core/a-node');

src/core/a-assets.js
```diff
@@ -1,7 +1,7 @@
-var ANode = require('./a-node');
+/* global customElements */
+var ANode = require('./a-node').ANode;
```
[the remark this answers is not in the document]
dmarcos: not for now. maybe in the future.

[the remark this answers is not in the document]
dmarcos: notice that this is a surgical PR to just migrate to customElements V1. A-Frame code base is ES5 for IE 11 compatibility. We can revisit and refactor in the future.

src/core/a-assets.js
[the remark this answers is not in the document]
dmarcos: @justinfagnani That's a requirement of the polyfill as @WebReflection mentioned

src/core/a-assets.js
dmarcos: thanks

src/core/a-assets.js
dmarcos: I tried that but `init` would have to also remember to call `super.init` in derived classes.

src/core/a-mixin.js
dmarcos: We try to avoid `attributeChangedCallback` as much as as possible to reduce asynchrony.

src/core/a-mixin.js
dmarcos: thanks. that's code predating this PR.

src/core/a-mixin.js
dmarcos: thanks

src/core/a-mixin.js
dmarcos: thanks, as mentioned above, A-Frame code base is ES5 prior to this PR. We can revisit and refactor later.

src/core/a-assets.js
WebReflection: I guess my hint was rather for developers extending A-Frae components too, where writing `super.init();` instead of remembering the `self = super(self)` dance seems to be a win/more maintainable approach, but of course it's your choice.

The idea is that the day you won't need the polyfill anymore, you can simply rename all `init` as `constructor` and call it a day.

src/core/a-entity.js
```diff
@@ -23,155 +22,156 @@ var ONCE = {once: true};
...
-      this.object3D.el = this;
-      this.object3DMap = {};
-      this.parentEl = null;
-      this.rotationObj = {};
-      this.states = [];
-    }
-  },
+class AEntity extends ANode {
+  static get observedAttributes () {
+    return Object.keys(COMPONENTS).concat(['mixin']);
```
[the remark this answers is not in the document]
dmarcos: thank you for the info and the heads up

src/core/a-mixin.js
dmarcos: In this case we don't want the `a-node` setAttribute logic to run but we still want the new values to be reflected in the DOM

src/core/a-mixin.js
dmarcos: similar to above. we don't want the a-node setAttribute logic to run but we still want the new values to be reflected in the DOM

dmarcos:
@WebReflection @justinfagnani I incorporated your feedback. Thanks a lot. That's very generous of your time, it's a big PR 👍

src/core/a-mixin.js
```diff
@@ -9,106 +9,91 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
-    createdCallback: {
-      value: function () {
-        this.componentCache = {};
-        this.id = this.getAttribute('id');
-        this.isMixin = true;
-      }
-    },
+class AMixin extends ANode {
+  constructor () {
+    super();
```
WebReflection: is ANode is extending HTMLElement, same `self = super(self)` dance might apply

dmarcos: Yeah, I think you're commenting in an outdated snippet:

https://github.com/aframevr/aframe/pull/4167/files#diff-631b4510fca3a68661ae2833c57e1e3dR18

Thanks

WebReflection: I did, apologies for the confusion

edsilv:
This is awesome! I'd started an experimental project to try porting A-Frame to Typescript + webpack, but soon realised that upgrading to custom elements v1 was necessary first. 

I was thinking to use esmodule imports instead of require too.

> The main drawback is that Custom Elements v1 enforces JS classes that are not available in older browsers.

We could make multiple UMD dist targets for es3, es5 etc with webpack.

Not sure how keen the community would be on Typescript, but happy to have a go if the consensus is that it would increase robustness. My setup generates the TS definitions too, so you don't need to maintain them separately.

edsilv:
It was a bit more complicated than I thought, but I've got a cross-browser ts + webpack + custom elements setup working here: https://github.com/edsilv/aframe-ts-webpack

arpu:
@dmarcos  i have this branch running since some time with any problems but now some trouble wiith exokit 
same problem accourse if in chrome if aframe load script is set to the end of the html 

the exoKit dev means this is a problem with this branch 

```NODE_TLS_REJECT_UNAUTHORIZED='0' node --experimental-worker . -x webvr https://vrland.io/exo_customElementsV1.html                                                                                                                    INT(-2) ↵  10084  10:04:54 
(node:19816) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(node:19816) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
THREE.WebGLRenderer 103dev
ErrorEvent {
  type: 'error',
  target: <script src="aframe_customElementsV1.js"></script>,
  bubbles: false,
  cancelable: false,
  defaultPrevented: false,
  propagationStopped: false,
  currentTarget: <script src="aframe_customElementsV1.js"></script>,
  message: 'systems[name] is not a constructor',
  stack: 'https://vrland.io/aframe_customElementsV1.js:76834\n    ' +
    'this.systems[name] = new systems[name](this);\n                        ' +
    ' ^\n\nTypeError: systems[name] is not a constructor\n    at ' +
    'AScene.initSystem ' +
    '(https://vrland.io/aframe_customElementsV1.js:76834:26)\n    at ' +
    'AScene.initSystems ' +
    '(https://vrland.io/aframe_customElementsV1.js:76821:10)\n    at ' +
    'AScene.connectedCallback ' +
    '(https://vrland.io/aframe_customElementsV1.js:76787:10)\n    at ' +
    'CustomElementRegistry.upgrade ' +
    '(/home/arpu/Work/githubsources/exokit/src/Window.js:285:38)\n    at ' +
    '/home/arpu/Work/githubsources/exokit/src/Window.js:210:14\n    at ' +
    '_recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1305:22)\n   ' +
    ' at _recurse ' +
    '(/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)\n    at ' +
    '_recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)\n   ' +
    ' at _recurse ' +
    '(/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)\n    at ' +
    'Document.traverse ' +
    '(/home/arpu/Work/githubsources/exokit/src/DOM.js:1322:12)'
}
https://vrland.io/aframe_customElementsV1.js:76834
    this.systems[name] = new systems[name](this);
                         ^

TypeError: systems[name] is not a constructor
    at AScene.initSystem (https://vrland.io/aframe_customElementsV1.js:76834:26)
    at AScene.initSystems (https://vrland.io/aframe_customElementsV1.js:76821:10)
    at AScene.connectedCallback (https://vrland.io/aframe_customElementsV1.js:76787:10)
    at CustomElementRegistry.upgrade (/home/arpu/Work/githubsources/exokit/src/Window.js:285:38)
    at /home/arpu/Work/githubsources/exokit/src/Window.js:210:14
    at _recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1305:22)
    at _recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)
    at _recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)
    at _recurse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1311:28)
    at Document.traverse (/home/arpu/Work/githubsources/exokit/src/DOM.js:1322:12)

```

modulesio:
☝️ I found the above crash occurs anytime `<a-scene>` is seen by the `aframe.js` initial tick -- e.g. if the script is at the end of the `</body>`. That behavior is not specific to Exokit.

arpu:
can this be updated to latest master?

WebReflection:
> can this be updated to latest master?

FYI the latest, better, improved, polyfill, targeting IE11 and every other browser, is this one:
https://github.com/WebReflection/custom-elements#readme

dmarcos:
@arpu Not a top priority at the moment. Any particular needs?

arpu:
@dmarcos nothing special want to look how i can convert to jsm threejs

vincentfretin:
replaced by #5136

dmarcos:
close in favor of https://github.com/aframevr/aframe/pull/5136

PR #4171, opened 2019-05-17
Add component setters and getters

dmarcos:
This is just a prototype based on https://github.com/aframevr/aframe/pull/4167 using JS setters and getters

Once can set components like this:

`el.position = {x: 0, y: 1, z: 1};`

Also retrieve values:

`var position = el.position;`


dmarcos:
Don't think this is a good idea anymore

PR #4173, opened 2019-05-20
add link to throttle utility under `tick` section

kfarr:
**Description:**
small doc change to add link to throttle utility under tick section of best practices page
**Changes proposed:**
-
-
-

### DIFF
--- a/docs/introduction/best-practices.md
+++ b/docs/introduction/best-practices.md
@@ -202,7 +202,7 @@ More articles on reducing garbage collector activity:
 
 ### `tick` Handlers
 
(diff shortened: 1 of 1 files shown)


docs/introduction/best-practices.md
```diff
@@ -248,7 +250,8 @@ AFRAME.registerComponent('foo', {
 ```
 
 Again be careful what you do in tick functions, treat them as critical
-performance code because they will be run 90 times per second.
+performance code because they will be run 90 times per second. [Consider
+using `utils.throttle`][throttle] to run your code at less frequent intervals.
```
ngokevin: Thanks! I think more common is throttleTick to wrap a tick function and throttle it.

PR #4176, opened 2019-05-20
Fix wrong element reference in guide

Paspartout:
I worked through the guide and had to remove the .el indirection to make the component example work.


ngokevin:
Oops. This should be `() => {}` instead of `function ()` to get the self-binding.

Issue #4187, opened 2019-05-26
Samsung Internet for Android is identified as a VR browser, not a mobile browser


DougReeder:
**Description:**
AFRAME.utils.device.isMobile() returns true and AFRAME.utils.device.isMobileVR() returns true under Samsung Internet for Android  (Under Samsung Internet for VR, they return the same values, which are correct there.)

- A-Frame Version: 0.9.2
- Platform / Device: Android 7.0,  Samsung Galaxy S6 (SM-G920V)
- Reproducible Code Snippet or URL:
https://juniper-plot.glitch.me/

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/4188

PR #4188, opened 2019-05-27
isMobileVR distinguishes Samsung Internet for Gear VR from Samsung Internet for Android

DougReeder:
**Description:**
Fixes issue #4187
https://github.com/aframevr/aframe/issues/4187

**Changes proposed:**
Samsung Internet for Gear VR has "Mobile VR" in the useragent where Samsung Internet for Android has "Mobile", so the existing regex clause that checks for `Mobile VR` will return true for the Gear VR version and false for the Android version.
-
-


dmarcos:
The test would return false for Go / GearVR versions of Samsung Internet. We need a different way to discriminate. What's the complete userAgent string for Samsung Internet in Go / GearVR?

Side note. We don't submit A-Frame builds as part of PRs

DougReeder:
Format of userAgent matches https://developer.samsung.com/internet/user-agent-string-format for devices I can test, with $(UX RECOMMEND) being "Mobile" on Android and "Mobile VR" on Gear VR.

I don't have a Go to check what it is there.

dmarcos:
Thanks. it looks good if we can verify that the function returns true in Oculus Go. 

Can you remove the dist and index.js files from the PR?

dmarcos:
Thanks. it looks good if we can verify that the function returns true in Oculus Go for Samsung Internet.

Can you remove the dist and index.js files from the PR?

dmarcos:
Thanks

PR #4196, opened 2019-06-02
update three.js introduction docs with new api requiring targets

jhsu:
**Description:**
update docs to use new three.js methods and provide targets to prevent
warnings.

related to #3900

### DIFF
--- a/docs/introduction/developing-with-threejs.md
+++ b/docs/introduction/developing-with-threejs.md
@@ -218,13 +218,15 @@ three.js defaults `Object3D.matrixAutoUpdate` to `true`. We can use three.js's
 To get the world position of an `Object3D`:
 
 ```js
-entityEl.object3D.getWorldPosition(new THREE.Vector3);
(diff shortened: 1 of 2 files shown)


docs/introduction/developing-with-threejs.md
```diff
@@ -213,26 +213,26 @@ accomplished with matrices, but three.js provides helpers to make them easier.
 
 Normally, we'd need to call `.updateMatrixWorld ()` on parent `Object3D`s, but
 three.js defaults `Object3D.matrixAutoUpdate` to `true`. We can use three.js's
-`.getWorldPosition ()` and `.getWorldRotation ()`.
+`.getWorldPosition (vector)` and `.getWorldQuaternion (quaternion)`.
 
 To get the world position of an `Object3D`:
 
 ```js
-entityEl.object3D.getWorldPosition();
```
ngokevin: should have `()` and probably better to show if the vector/quat is created in a variable in passed in so people know to cache the objects.

Issue #4198, opened 2019-06-06
requestPresent from vrdisplayactivate is failing due to usage of wrong layer.


Artyom17:
**Description:**
The requestPresent from within the vrdisplayactivate is failing on Oculus Browser (and I bet on Chrome too) due to the wrong layer provided. The commit that introduced the issue is a3ac58e4084053c14706abc6ef37bde45086a33e. See https://github.com/aframevr/aframe/blob/b3ed182fefb883e36272401529a8b5a3b1a9ce5d/src/utils/device.js#L13: it creates a new non-gl canvas, and tries to use it as a layer, which is not going to work. Even in the nested requestPresents you must specify the ACTUAL GL canvas, not just a placeholder (https://immersive-web.github.io/webvr/spec/1.1/#dom-vrdisplay-requestpresent%E2%91%A0)

Seems like the previous version worked (or should have worked).

- A-Frame Version: 0.9.2 (master)
- Platform / Device: Oculus Browser / Chrome
- Reproducible Code Snippet or URL:
With connected Go or Quest, run this adb command:
adb shell am start -n com.oculus.vrshell/.MainActivity -d apk://com.oculus.browser -e uri "ovrweb://vr?uri=http%3A%2F%2Faframe.io%2Fexamples%2Fshowcase%2Fsky" (may replace the actual url to whatever is better for you).

Artyom17:
Changing it back to `window.addEventListener('vrdisplayactivate', function (evt) {
  // WebXR takes priority if available.
  if (navigator.xr) { return; }

  let sceneEl = document.querySelector('a-scene');
  let canvasEl = sceneEl.canvas;
  vrDisplay = evt.display;
  // Request present immediately. a-scene will be allowed to enter VR without user gesture.
  vrDisplay.requestPresent([{source: canvasEl}]).then(function () {}, function () {});
});` seems to fix the issue.

dmarcos:
@Artyom17 Thanks for taking a look, much appreciated. The placeholder canvas is passed to honor the time limit that WebVR implementations have to call requestPresent after vrdisplayactivate fires. At that point in time the scene and final canvas might not be initialized. Will initializing the webgl context with the placeholder canvas work? Does Oculus not support multiple calls to requestPresent with different canvases?

dmarcos:
Oh thanks for the suggestion. We just crossed comments and I didn’t see yours

Artyom17:
[...]

PR #4199, opened 2019-06-07
Fixing issue #4198: VR deep-linking / VR navigation doesn't work.

Artyom17:
**Description:**
Fixing issue #4198: VR deep-linking / VR navigation doesn't work.

**Changes proposed:**
- Canvas must have gl context associated with it for requestPresent to work;
- onVRDisplayPresentChange by spec is called only once for the first requestPresent (which might be called from the ondisplayactivate and can't init all the data); so, firing the second event "artificially".

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -314,11 +314,7 @@ module.exports.AScene = registerElement('a-scene', {
           // to setup everything from there. Thus, we need to emulate another vrdisplaypresentchange
           // for the actual requestPresent. Need to make sure there are no issues with firing the
           // vrdisplaypresentchange multiple times.
-          var event = new CustomEvent('vrdisplaypresentchange', {
-            detail: {
-              display: utils.device.getVRDisplay()
-            }
-          });
+          var event = new CustomEvent('vrdisplaypresentchange', {detail: {display: utils.device.getVRDisplay()}});
           window.dispatchEvent(event);
 
           self.addState('vr-mode');
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -10,6 +10,7 @@ var utils = require('../../utils/');
 // Require after.
 var AEntity = require('../a-entity');
 var ANode = require('../a-node');
+var CustomEvent = require('custom-event');
 var initPostMessageAPI = require('./postMessage');
 
 var bind = utils.bind;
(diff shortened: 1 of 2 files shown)


src/core/scene/a-scene.js
```diff
@@ -10,6 +10,7 @@ var utils = require('../../utils/');
 // Require after.
 var AEntity = require('../a-entity');
 var ANode = require('../a-node');
+var CustomEvent = require('custom-event');
```
dmarcos: I don't think this is necessary. If the linter is complaining you can just add `CustomEvent` here: https://github.com/aframevr/aframe/pull/4199/files#diff-eb9bb1084863a2f65678ea67d6e7cf5eR1

dmarcos:
It looks good to me. Just a small change and it's ready to go. Thank you very much for this and investigating the issue. Glad to have robust in-vr navigation again 👏

src/core/scene/a-scene.js
Artyom17: Got it!

src/core/scene/a-scene.js
```diff
@@ -309,6 +309,18 @@ module.exports.AScene = registerElement('a-scene', {
 
         // Callback that happens on enter VR success or enter fullscreen (any API).
         function enterVRSuccess () {
+          // vrdisplaypresentchange fires only once when the first requestPresent is completed;
+          // the first requestPresent could be called from ondisplayactivate and there is no way
+          // to setup everything from there. Thus, we need to emulate another vrdisplaypresentchange
+          // for the actual requestPresent. Need to make sure there are no issues with firing the
+          // vrdisplaypresentchange multiple times.
+          var event = new CustomEvent('vrdisplaypresentchange', {
```
dmarcos: One more thing. Styling. We can maybe combine in one line:

 `var event = new CustomEvent('vrdisplaypresentchange', {detail: {display: utils.device.getVRDisplay()}});`

Artyom17: Yeah, linter just argued about inconsistent spacing, so I decided to make sure that spacing is super consistent! ;) but no biggie, can make it one-liner.

dmarcos: our linter doesn't like space around curly braces. e.g `{default: true}`

dmarcos:
Awesome. Thank you!

Issue #4149, opened 2019-05-01
Show message on iOS to request user to enable devicemotion manually


dmarcos:
Since iOS 12.2 Safari `devicemotion` API comes disabled by default. The user has to enable manually on settings. We could detect iOS and if the API is disabled we can display a message above the enter VR button asking the user to enable.

dmarcos:
It's not trivial but not super hard to implement. We could use some help if someone is interested in contributing to A-Frame

eyexpo-dan:
How would you detect iOS safari privacy & security settings? I am listening to `devicemotion`, and it'll not trigger for either desktop, iOS 12.2+, and mobile devices without gyroscope. Is there a granular method of detection for all these possible devices?

dmarcos:
Can just one feature detect `devicemotion`? A bit ugly but maybe:

```
 var deviceMotionEvt;
 setTimeout(function () { 
    if (!deviceMotionEvt && AFRAME.utils.isIOS()) { showMessage(); };
 }, 1000);

  window.addEventListener('devicemotion', function(event) {
   deviceMotionEvt = event;
});```

eyexpo-dan:
Is aframe using VRDisplay in the background?

`    navigator.getVRDisplays().then((displays) => {
      // Do something with the available VR displays
    });
`

Would any of the following properties be useful for confirming if the permission is turned on or off in iOS?

`
isIOS: true
gyroscope { x: 0, y: 0, z: 0 }
sensor: null
errors: ... Can't find RelativeOrientationSensor
`

dmarcos:
Whatever works. In my snippet above I assumed the only way to detect devicemotion is listening to events that would always fire at page load if the API is enabled. A synchronous way would be better if there’s any.

EslamNemr:
using callback
https://aframe-ios-motion-orientation-access.glitch.me/
tried on i phone xs-max
can we use promise ?

dmarcos:
not sure If I understood. devicemotion is not promise API

EslamNemr:
I mean Async/Await to avoid timeout.

micsun-al:
iPad Air (MD785LL/A) running iOS 11.4 (15F79) laying flat on a table does not send any `devicemotion` events until the device actually moves, so this setTimeout devicemotion check fails ... wonder what other things we can check for hrmmm (time for more googling).

dmarcos:
[...]

PR #4202, opened 2019-06-08
(#4149) Instructs user how to resolve the issue

jamesdigid:
**Description:**
Detects if motion & orientation tracking is enabled from deviceorientation event, and creates an async call. This is a quick win for now and will help with any confusion.

**Changes proposed:**
Offering simple instruction overlayed on the scene explaining how to enable motion tracking for Safari or Chrome. 
-
-
-

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -60,7 +60,7 @@ if (utils.device.isBrowserEnvironment) {
 }
 
 // Check iOS motion enabled
-var isIOSMotionalEnabled = utils.device.isIOSMotionAndOritentationEnabled();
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks. This might no be necessary anymore and we probably should use `DeviceOrientationEvent.requestPermission` instead. I think you can play with if you have a device with iOS 13 beta installed

dmarcos:
@jamesdigid I don't think we want to manually show a message anymore to the user and instead use `DeviceOrientationEvent.requestPermission`. More info https://github.com/aframevr/aframe/issues/4149#issuecomment-500179345

ngokevin:
We will use requestPermission soon. Thanks!

PR #4212, opened 2019-06-25
fix(dev-deps): move extraneous

Francois-Esquire:
**Description:**

Moved three development dependencies down to `devDependencies`; unnecessarily added during prod builds and deploys. VERY minor, simply refining the package.

**Changes proposed:**

Packages moved to `devDependencies`:
- `browserify-css`
- `envify`

Packages removed entirely:
- `style-attr`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,17 +38,14 @@
     "vendor/**/*"
   ],
   "dependencies": {
-    "browserify-css": "^0.8.4",
     "custom-event-polyfill": "^1.0.6",
     "debug": "ngokevin/debug#noTimestamp",
     "deep-assign": "^2.0.0",
     "document-register-element": "dmarcos/document-register-element#8ccc532b7f3744be954574caf3072a5fd260ca90",
-    "envify": "^3.4.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -93,6 +92,7 @@
     "sinon": "^1.17.5",
     "sinon-chai": "2.8.0",
     "snazzy": "^5.0.0",
+    "style-attr": "^1.0.2",
```
ngokevin: I believe this is a hard dependency.

Francois-Esquire: Ahh, I could have missed something, however the only reference of `style-attr` that I found was commented, found here (`src/utils/styleParser.js`):

https://github.com/aframevr/aframe/blob/master/src/utils/styleParser.js#L3

ngokevin: I see, it might be able to just be removed then.

Francois-Esquire: removed it

Francois-Esquire:
I've squashed history and ready to merge. Please let me know if anything else needs doing on my end.

PR #4225, opened 2019-07-10
Add Valve Index controller support [WIP]

dmarcos:
It needs validation on [Exokit PR](https://github.com/exokitxr/exokit/pull/1295)


dmarcos:
We already have valve-index support. Not needed

Issue #4219, opened 2019-07-09
a-asset-item fails to infer correct response type if query parameters present 


Ely-S:
**Description: Adding query parameters to the src of an a-asset-item will prevent the loader from correctly parsing glb/gltf files.**

I ran into this when trying to load a GLB file from a signed S3 url.

- A-Frame Version: 0.9.0
- Platform / Device: Version 74.0.3729.169 (Official Build) Built on Ubuntu, running on Ubuntu 16.04 (64-bit)
- Reproducible Code Snippet or URL: https://codepen.io/anon/pen/gNQeRB

In the example, try removing the `?key=2` from the asset item src to demonstrate the bug.

The problem is that aframe uses the file ending to determine the file type, but doesn't take into account any query parameters that may come after the file name.

dmarcos:
Thanks for filing. Proper way to deal with this would be to determine format from file header / metadata / magic number when parsed.

PR #4228, opened 2019-07-15
Fix a-asset-item binary glTF file loading

Ely-S:
**Description:**
Fixes loading of binary glTF files by determining the file type using the glTF header. #4219 

**Changes proposed:**
- When loading resources using a-asset-item, aframe determines the response type (text or arraybuffer) by checking to see if the requested resource has a .glb/.gltf file extension. This solution failed when loading resources with query parameters at the end (e.g` /file.glb?version=1`). This fix checks for the presence of the glTF magic number in the response body to determine if it should be treated as binary glTF data or text.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -253,7 +253,7 @@ function extractDomain (url) {
 function inferResponseType (src) {
   var dotLastIndex = src.lastIndexOf('.');
   if (dotLastIndex >= 0) {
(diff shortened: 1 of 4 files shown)


dmarcos:
This breaks loading of the glTF file loading in examples: `/showcase/shopping/` and `/examples/showcase/anime-UI/`

PR #4243, opened 2019-08-02
fix ie 11 fullscreen mode bug

AlexandraWins:
exit fullscreen mode in IE correctly

**Description:**
Currently fullscreen mode is not exited correctly in IE11 on a Desktop device. It is only possible to see images in fullscreen once. As soon as fullscreen is exited, fullscreen button is gone. 
**Changes proposed:**
- add exit fullscreen mode event listener for IE


dmarcos:
@AlexandraWins There's a linter issue. Extra space at the end of the line (https://github.com/aframevr/aframe/pull/4243/files#diff-eb9bb1084863a2f65678ea67d6e7cf5eR845). We can merge after fix

AlexandraWins:
thanks for feedback @dmarcos

dmarcos:
Thanks

PR #4245, opened 2019-08-03
respect embedded on loading screen

andyps:
**Description:**
Currently regardless of existing _embedded_ attribute loading screen is always in fullscreen

**Changes proposed:**
if a scene has _embedded_ attribute it should apply to loading screen as well


dmarcos:
FYI, you can also iframe your scene as a workaround

andyps:
Anyway _embedded_ is treated like a special component and is taken into account in _exitVR_ and _resize_ methods so _a-scene_ is already dependant on it.

PR #4246, opened 2019-08-04
Fixes #3592: Fix resolution drop on exiting VR mode

### LINKED ISSUE #3592: resolution drops on mobile when exiting VR-mode
**Description:**
I’m having an issue with Aframe on mobile dramatically reducing the resolution after exiting vr-mode. Basically the resolution drops from 1080p to 360p (based on canvas.width/.height and renderer.getSize()). 

I can find other posts about issues with the resolution in VR-mode (like #3075 #3160 #3031 #2761) but for some reason I can’t find anything on this situation though it happens consistently for me on both Android and iPhone devices and for seemingly every aframe demo I can find online (not just my own app).

Image when entered into VR-mode;
before

Image after exiting VR-mode;
after

See also my stackoverflow post; https://stackoverflow.com/questions/50394308/aframe-resolution-drops-on-mobile-when-exiting-vr-mode 

- A-Frame Version: 0.8.2
- Platform / Device: Android/iPhone Firefox & Chrome
- Reproducible Code Snippet or URL: https://aframe.io/examples/showcase/helloworld/

tomegz:
**Description:**
Initial pixel ratio is set when setting up the renderer. After exiting VR mode it somehow gets reset. Additional `setPixelRatio` call on exiting VR mode fixes this issue.

**Changes proposed:**
- set correct pixel ratio on exiting VR mode

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -394,6 +394,7 @@ module.exports.AScene = registerElement('a-scene', {
           if (self.hasAttribute('embedded')) { self.removeFullScreenStyles(); }
           self.resize();
           if (self.isIOS) { utils.forceCanvasResizeSafariMobile(self.canvas); }
+          self.renderer.setPixelRatio(window.devicePixelRatio);
           self.emit('exit-vr', {target: self});
         }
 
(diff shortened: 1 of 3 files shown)


tomegz:
I see that the tests are failing. I need to check why I can't access `self.renderer.setPixelRatio`

dmarcos:
@tomegz Thanks. To fix the test you have to mock `setPixelRatio` method here https://github.com/aframevr/aframe/blob/master/tests/__init.test.js#L32

tomegz:
Done, sorry it took me so long. Please review @dmarcos

PR #4249, opened 2019-08-07
only emit cursor mouseup if mousedown originated on scene

edsilv:
**Description:**

If the A-Frame scene is embedded, using ionic-range slider controls outside the scene and releasing inside the scene causes the slider values to be 0.

ezgif com-video-to-gif

**Changes proposed:**
- only emit cursor mouseup if mousedown originated on scene

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -244,9 +244,7 @@ module.exports.Component = registerComponent('cursor', {
    *   in case user mousedowned one entity, dragged to another, and mouseupped.
    */
   onCursorUp: function (evt) {
-    if (!this.isCursorDown) {
-      return;
-    }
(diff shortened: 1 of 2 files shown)


src/components/cursor.js
```diff
@@ -242,6 +244,12 @@ module.exports.Component = registerComponent('cursor', {
    *   in case user mousedowned one entity, dragged to another, and mouseupped.
    */
   onCursorUp: function (evt) {
+    if (!this.isCursorDown) {
```
dmarcos: Can we do this a oneliner? `if (!this.isCursorDown) { return; }`

dmarcos:
Thanks!

PR #4264, opened 2019-09-07
Fixed "we'll" to "we" in "we'll won't be setting a schema"

craftecode:
Fixed "we'll" to "we" in "we'll won't be setting a schema"

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thank you very much. Super appreciated

Issue #3493, opened 2018-03-23
Don't fire click event when mouse is dragged


rockwalrus:
As mentioned in #2625, the `click` event fires when the mouse is dragged, which is inconsistent with the way `click` events normally are fired in the browser.

rockwalrus:
To be clear, it fires on mouseup even if the mouse has been dragged. Normal browser behavior is to fire click only if the mouse hasn't moved appreciably between down and up.

ngokevin:
This can be fixed by checking that the mouse movement has not moved a lot on the touchend.

Johannemoeller112:
Thanks for sharing this tip! I also experienced that it can be fixed by checking that the mouse movement has not moved that much :-)

Johanne from Denmark / [https://handyhand.dk/](url)

Tzyinc:
I'd like to contribute to this, but I need some guidance as I'm new to both vr and this repo.

I've looked at the source code and I think the file I want to change is `/src/components/cursor.js`

I also think the solution suggested by @ngokevin is to save the evt x and y in `onCursorDown` (line 226) and then check distance traveled in `onCursorUp` (line 246) right?

I did consider a timing implementation instead of a distance traveled implementation but I found that browser's eventlistener will trigger even if the user long presses, so to be consistent, distance travelled seems to be the better solution.

PR #4296, opened 2019-10-19
bug: fixed how click events are fired #3493

Tzyinc:
**Description:**
from #3493 added distance calculation for the two points mousdown and mouse up

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -26,6 +26,9 @@ var CANVAS_EVENTS = {
 
 var CANVAS_HOVER_CLASS = 'a-mouse-cursor-hover';
 
+// check cursor move over this treshold
+var CURSOR_MOVE_TRESHOLD = 50;
+
 /**
  * Cursor component. Applies the raycaster component specifically for starting the raycaster
  * from the camera and pointing from camera's facing direction, and then only returning the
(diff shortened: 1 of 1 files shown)


src/components/cursor.js
```diff
@@ -26,6 +26,9 @@ var CANVAS_EVENTS = {
 
 var CANVAS_HOVER_CLASS = 'a-mouse-cursor-hover';
 
+// check cursor move over this treshold
+var CURSOR_MOVE_TRESHOLD = 50;
```
vincentfretin: missing a h => threshold

Tzyinc: sorry for the late reply! will make a change over the weekend! also will fix the  CI errors

Tzyinc: @vincentfretin after trying to do this over this weekend, I think I still need some guidance.

I added a  check in onCursorUp, to not trigger EVENTS.CLICK  if the distance is greater than the treshold, but i'm still failing the CI as evt can be undefined. is there a better way for me to get x and y of the cursor at any given time?

dmarcos:
Closing due to inactivity. Can reopen if necessary!

PR #4298, opened 2019-10-22
Add optional initial rotation of the entity to look-controls

massimo-bortolamei-hegias:
**Description:**
I added support to `look-controls` to keep initial rotation of the entity it is attached to, instead of resetting it to `0 0 0` .
This happens as a consequence of the tick's `updateOrientation()` where 
```
object3D.rotation.x = hmdEuler.x + pitchObject.rotation.x;
object3D.rotation.y = hmdEuler.y + yawObject.rotation.y;
```
which initially is  `0 0 0` if there is no hmd

**Changes proposed:**
- add to the schema a variable to enable keeping the entity's rotation
- in the `setupMouseControls()` function, configure `pitchObject.rotation.x` and `yawObject.rotation.y` accordingly


dmarcos:
This would make magic window and headset experiences inconsistent. Depending on what you do consider making the camera a child of an entity where you apply the rotation shift

massimo-bortolamei-hegias:
Ah, I don't use magic window therefore I cannot really speak about that, but to avoid any inconsistencies I made it as a schema variable with a default configured to keep the original behavior.

dmarcos:
Sorry, I meant magic window and also mouse behavior. The recommended way is to parent the camera and add the rotation there to preserve it. Would that work for your use case?

dmarcos:
Closing due to inactivity.

PR #4303, opened 2019-11-04
Device motion permission

KevinEverywhere:
**Description:**
iOS 13 introduced new permissions requirements for device motion and orientation. This PR enables iOS 13 Safari users to use device motion.

**Changes proposed:**
- A Scene Component addition (device-motion-permission-ui)
- CSS for screen visible selector to acquire permision
- Updated package.json file was required to build aframe

**Recommended Todo**
- Ensure that iPad 13 browsers also work (only Chrome does now)
- Always display this graphic for content that requires device motion

The second recommended todo is to proactively give user feedback that their device motion will be part of their experience. Depending on its function, this could have different opacity or color.

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -33,6 +33,7 @@ require('./windows-motion-controls');
 
 require('./scene/background');
 require('./scene/debug');
+require('./scene/device-orientation-permission-ui');
 require('./scene/embedded');
 require('./scene/inspector');
 require('./scene/fog');
--- a/src/style/aframe.css
+++ b/src/style/aframe.css
@@ -289,3 +289,93 @@ a-scene audio {
   top: 0px;
   color: white;
 }
+
+.a-modal {
+  position: absolute;
+  background: rgba(0, 0, 0, 0.60);
+  background-size: 50% 50%;
+  bottom: 0;
+  font-size: 14px;
+  font-weight: 600;
+  left: 0;
+  line-height: 20px;
+  right: 0;
+  position: fixed;
+  top: 0;
+  z-index: 9999999;
+}
+
+.a-dialog {
+  position: relative;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 199995;
+  width: 300px;
+  height: 200px;
+  background-size: contain;
+  background-color: white;
+  font-family: sans-serif, monospace;
+  font-size: 20px;
+  border-radius: 3px;
+  padding: 6px;
+}
+
+.a-dialog-text-container {
+  width: 100%;
+  height: 70%;
+  align-self: flex-start;
+  display: flex;
+  justify-content: center;
+  align-content: center;
+  flex-direction: column;
+}
+
+.a-dialog-text {
+  display: inline-block;
+  font-weight: normal;
+  font-size: 14pt;
+  margin: 8px;
+}
+
+.a-dialog-buttons-container {
+  display: inline-flex;
+  align-self: flex-end;
+  width: 100%;
+  height: 30%;
+}
+
+.a-dialog-button {
+  cursor: pointer;
+  align-self: center;
+  opacity: 0.9;
+  height: 80%;
+  width: 50%;
+  font-size: 12pt;
+  margin: 4px;
+  border-radius: 2px;
+  text-align:center;
+  border: none;
+  display: inline-block;
+  -webkit-transition: all 0.25s ease-in-out;
+  transition: all 0.25s ease-in-out;
+  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.10), 0 1px 2px rgba(0, 0, 0, 0.20);
+  user-select: none;
+}
+
+.a-dialog-permission-button:hover {
+  box-shadow: 0 7px 14px rgba(0,0,0,0.20), 0 2px 2px rgba(0,0,0,0.20);
+}
+
+.a-dialog-allow-button {
+  background-color: #8ce3ba;
+}
+
+.a-dialog-deny-button {
+  background-color: #ff4b8b;
+}
+
+.a-dialog-ok-button {
+  background-color: #8ce3ba;
+  width: 100%;
+}
--- a/docs/components/device-orientation-permission-ui.md
+++ b/docs/components/device-orientation-permission-ui.md
@@ -1,0 +1,47 @@
+---
+title: device-orientation-permission-ui
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/scene/device-orientation-permission-ui.js
+---
+
+Some browsers like Safari on iOS 13 and later require sites to request user permission to access DeviceOrientation events. This component presents a permission dialog for the user to grant or deny access.
+The device-orientation-permission-ui component applies only to the [`<a-scene>` element][scene]
+
+To configure the style of the dialog one can redefine the associated css styles. To change the colors of the allow, deny and ok buttons:
+
+```css
+.a-dialog-allow-button {
+  background-color: red;
+}
+
+.a-dialog-deny-button {
+  background-color: blue;
+}
+
+.a-dialog-ok-button {
+   background-color: green;
+ }
+```
+
+## Example
+
+```html
+<a-scene device-orientation-permission-ui="enabled: false"></a-scene>
+```
+
+## Properties
+
+| Property      | Description                                                         | Default Value |
+|---------------|---------------------------------------------------------------------|---------------|
+| enabled       | Whether or not to display the dialog when required                  | true          |
+
+## Events
+
+| Event Name                           | Description                                                                                |
+|--------------------------------------|--------------------------------------------------------------------------------------------|
+| deviceorientationpermissiongranted   | User has granted access to DeviceOrientation events                                        |
+| deviceorientationpermissionrejected  | User or browser has denied access to DeviceOrientation events                              |
+| deviceorientationpermissionrequested | Application has requested permission to access DeviceOrientation events                    |
+[scene]: ../core/scene.md
--- a/src/components/scene/device-orientation-permission-ui.js
+++ b/src/components/scene/device-orientation-permission-ui.js
@@ -1,0 +1,173 @@
+/* global DeviceOrientationEvent  */
+var registerComponent = require('../../core/component').registerComponent;
+var utils = require('../../utils/');
+var bind = utils.bind;
+
+var constants = require('../../constants/');
+
+var MODAL_CLASS = 'a-modal';
+var DIALOG_CLASS = 'a-dialog';
+var DIALOG_TEXT_CLASS = 'a-dialog-text';
+var DIALOG_TEXT_CONTAINER_CLASS = 'a-dialog-text-container';
+var DIALOG_BUTTONS_CONTAINER_CLASS = 'a-dialog-buttons-container';
+var DIALOG_BUTTON_CLASS = 'a-dialog-button';
+var DIALOG_ALLOW_BUTTON_CLASS = 'a-dialog-allow-button';
+var DIALOG_DENY_BUTTON_CLASS = 'a-dialog-deny-button';
+var DIALOG_OK_BUTTON_CLASS = 'a-dialog-ok-button';
+
+/**
+ * UI for enabling device motion permission
+ */
+module.exports.Component = registerComponent('device-orientation-permission-ui', {
+  schema: {enabled: {default: true}},
+
+  init: function () {
+    var self = this;
+
+    if (!this.data.enabled) { return; }
+
+    // Show alert on iPad if Safari is on desktop mode.
+    if (utils.device.isMobileDeviceRequestingDesktopSite()) { this.showMobileDesktopModeAlert(); }
+
+    // Browser doesn't support or doesn't require permission to DeviceOrientationEvent API.
+    if (!DeviceOrientationEvent || !DeviceOrientationEvent.requestPermission) { return; }
+    this.onDeviceMotionDialogAllowClicked = bind(this.onDeviceMotionDialogAllowClicked, this);
+    this.onDeviceMotionDialogDenyClicked = bind(this.onDeviceMotionDialogDenyClicked, this);
+    // Show dialog only if permission has not yet been granted.
+    DeviceOrientationEvent.requestPermission().catch(function () {
+      self.devicePermissionDialogEl = createPermissionDialog(
+        'This immersive website requires access to your device motion sensors',
+        self.onDeviceMotionDialogAllowClicked,
+        self.onDeviceMotionDialogDenyClicked);
+      self.el.appendChild(self.devicePermissionDialogEl);
+    }).then(function () {
+      self.el.emit('deviceorientationpermissiongranted');
+    });
+  },
+
+  remove: function () {
+    // This removes the modal screen
+    if (this.devicePermissionDialogEl) { this.el.removeChild(this.devicePermissionDialogEl); }
+  },
+
+  onDeviceMotionDialogDenyClicked: function () {
+    this.remove();
+  },
+
+  showMobileDesktopModeAlert: function () {
+    var self = this;
+    var safariIpadAlertEl = createAlertDialog(
+      'Request the mobile version of this site to enjoy it in immersive mode',
+      function () { self.el.removeChild(safariIpadAlertEl); });
+    this.el.appendChild(safariIpadAlertEl);
+  },
+
+  /**
+   * Enable device motion permission when clicked.
+   */
+  onDeviceMotionDialogAllowClicked: function () {
+    var self = this;
+    this.el.emit('deviceorientationpermissionrequested');
+    DeviceOrientationEvent.requestPermission().then(function (response) {
+      if (response === 'granted') {
+        self.el.emit('deviceorientationpermissiongranted');
+      } else {
+        self.el.emit('deviceorientationpermissionrejected');
+      }
+      self.remove();
+    }).catch(console.error);
+  }
+});
+
+/**
+ * Create a modal dialog that request users permission to access the Device Motion API.
+ *
+ * @param {function} onAllowClicked - click event handler
+ * @returns {Element} Wrapper <div>.
+ */
+function createPermissionDialog (text, onAllowClicked, onDenyClicked) {
+  var buttonsContainer;
+  var denyButton;
+  var acceptButton;
+
+  buttonsContainer = document.createElement('div');
+  buttonsContainer.classList.add(DIALOG_BUTTONS_CONTAINER_CLASS);
+
+  // Buttons
+  denyButton = document.createElement('button');
+  denyButton.classList.add(DIALOG_BUTTON_CLASS, DIALOG_DENY_BUTTON_CLASS);
+  denyButton.setAttribute(constants.AFRAME_INJECTED, '');
+  denyButton.innerHTML = 'Deny';
+  buttonsContainer.appendChild(denyButton);
+
+  acceptButton = document.createElement('button');
+  acceptButton.classList.add(DIALOG_BUTTON_CLASS, DIALOG_ALLOW_BUTTON_CLASS);
+  acceptButton.setAttribute(constants.AFRAME_INJECTED, '');
+  acceptButton.innerHTML = 'Allow';
+  buttonsContainer.appendChild(acceptButton);
+
+  // Ask for sensor events to be used
+  acceptButton.addEventListener('click', function (evt) {
+    evt.stopPropagation();
+    onAllowClicked();
+  });
+
+  denyButton.addEventListener('click', function (evt) {
+    evt.stopPropagation();
+    onDenyClicked();
+  });
+
+  return createDialog(text, buttonsContainer);
+}
+
+function createAlertDialog (text, onOkClicked) {
+  var buttonsContainer;
+  var okButton;
+
+  buttonsContainer = document.createElement('div');
+  buttonsContainer.classList.add(DIALOG_BUTTONS_CONTAINER_CLASS);
+
+  // Buttons
+  okButton = document.createElement('button');
+  okButton.classList.add(DIALOG_BUTTON_CLASS, DIALOG_OK_BUTTON_CLASS);
+  okButton.setAttribute(constants.AFRAME_INJECTED, '');
+  okButton.innerHTML = 'Ok';
+  buttonsContainer.appendChild(okButton);
+
+  // Ask for sensor events to be used
+  okButton.addEventListener('click', function (evt) {
+    evt.stopPropagation();
+    onOkClicked();
+  });
+
+  return createDialog(text, buttonsContainer);
+}
+
+function createDialog (text, buttonsContainerEl) {
+  var modalContainer;
+  var dialog;
+  var dialogTextContainer;
+  var dialogText;
+
+  modalContainer = document.createElement('div');
+  modalContainer.classList.add(MODAL_CLASS);
+  modalContainer.setAttribute(constants.AFRAME_INJECTED, '');
+
+  dialog = document.createElement('div');
+  dialog.className = DIALOG_CLASS;
+  dialog.setAttribute(constants.AFRAME_INJECTED, '');
+  modalContainer.appendChild(dialog);
+
+  dialogTextContainer = document.createElement('div');
+  dialogTextContainer.classList.add(DIALOG_TEXT_CONTAINER_CLASS);
+  dialog.appendChild(dialogTextContainer);
+
+  dialogText = document.createElement('div');
+  dialogText.classList.add(DIALOG_TEXT_CLASS);
+  dialogText.innerHTML = text;
+  dialogTextContainer.appendChild(dialogText);
+
+  dialog.appendChild(buttonsContainerEl);
+
+  return modalContainer;
+}
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -65,6 +65,7 @@ module.exports.AScene = registerElement('a-scene', {
         this.setAttribute('keyboard-shortcuts', '');
         this.setAttribute('screenshot', '');
         this.setAttribute('vr-mode-ui', '');
+        this.setAttribute('device-orientation-permission-ui', '');
       }
     },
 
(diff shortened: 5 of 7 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "ghpages": "ghpages -p gh-pages/",
     "lint": "semistandard -v | snazzy",
     "lint:fix": "semistandard --fix",
-    "precommit": "npm run lint",
+    "porecommit": "npm run lint",
```
dmarcos: Typo

package.json
```diff
@@ -38,63 +38,63 @@
     "vendor/**/*"
   ],
   "dependencies": {
-    "custom-event-polyfill": "^1.0.6",
+    "custom-event-polyfill": "^1.0.7",
```
dmarcos: Better to not touch dependencies. Only upgrade when necessary.

src/components/index.js
```diff
@@ -33,6 +33,7 @@ require('./windows-motion-controls');
 
 require('./scene/background');
 require('./scene/debug');
+require('./scene/device-motion-permission-ui');
```
dmarcos: Great. I think this is the good approach

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * UI for enabling device motion permission
+ */
+ module.exports.Component = registerComponent('device-motion-permission', {
+
+  dependencies: ['canvas'],
+
+  schema: {
+    enabled: {default: true},
+    enableFunc: {default:''}
```
dmarcos: What is this for?

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,121 @@
...
+
+var DEVICE_PERMISSION_CLASS = 'a-device-motion-permission';
+var DEVICE_PERMISSION_BTN_CLASS = 'a-device-motion-permission-button';
+
+/**
+ * UI for enabling device motion permission
+ */
+ module.exports.Component = registerComponent('device-motion-permission', {
+
+  dependencies: ['canvas'],
```
dmarcos: This dependency might not be necessary for this component

package.json
KevinEverywhere: Ooops! That should have been removed. Before I updated dependencies in package.json, npm run lint was failing at several places. This has been reverted to "precommit".

package.json
KevinEverywhere: I agree; and also, if package.json is updated, it will likely require moving forward the aframe version number. That said, if you use a clean machine and git clone the repo, you will find a number of issues with the current package.json file that makes updating pieces of it problematic. Can anyone else verify with a clean machine that package.json is stable? Thanks!

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: These are in flux and are not yet implemented. They may not belong in the final PR on this. I will revisit before resubmitting.

package.json
dmarcos: Travis (our CI) machine does that per commit and haven't heard of any issues. Better to revert the package.json file to the original version and address any issues in a seperate PR

dmarcos:
Thanks for doing this.. So far the code approach looks sensible. I haven't been able to test it on device yet.

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,111 @@
...
+
+var constants = require('../../constants/');
+
+var DEVICE_PERMISSION_CLASS = 'a-device-motion-permission';
+var DEVICE_PERMISSION_BTN_CLASS = 'a-device-motion-permission-button';
+
+/**
+ * UI for enabling device motion permission
+ */
+ module.exports.Component = registerComponent('device-motion-permission', {
```
dmarcos: It should be called `device-motion-permission-ui.js`

KevinEverywhere: Good catch--corrected.

KevinEverywhere: Good catch--corrected.

KevinEverywhere:
Update on this: I found that iPad works for Firefox and Safari: in both cases, the applications had Desktop Version Requested. You either needed to set the preference in your preferences, or select the aA icon in the left of the location for Safari; or the three dots  on the right side of Firefox, to Request Mobile Site. That means that what I have done is working across the target browsers on iOS and iPad browsers; I believe that it will also work on Android devices as well (not tested on many). The controller is in the bottom left of the screen to ask permission. For android devices, you will not need to select it.

https://planetkevin.com/aframe/examples/showcase/device-motion-permission/a.html

KevinEverywhere:
The last thing remaining from development perspective is getting past Travis. It is erring out on code that was written before me which I do not believe my code impacts. Is there something I am not understanding on this?

package.json
KevinEverywhere: I've reverted the package.json and resolved all the other issues described. The big remaining question is, do we want to have a temporary or permanent visible icon on the bottom left, the way it is now; or, do we want to have a full screen interrupt, similar to 8thwall or vr-mode-ui when you are in portrait mode, to question device-motion-permission? I believe it is ready to be checked in.

dmarcos: Thanks for the hard work. I would say we want a modal dialog similar to 8thwall or vr-mode-ui. Button is easy to miss and hard to understand what it's for. We can apply and reuse vr-mode-ui  style for consistency.

KevinEverywhere:
Matching Travis requirements, pulling out identified linting errors. Last big remaining item, making the modal screen attractive. Currently, Cancel and Continue buttons are functional. My idea: a Built with Aframe logo and wording that tells users that permission is required for device motion. I can continue with this, or if anyone else has an idea of what can be included here graphically, that could also be helpful.

dmarcos:
Thanks for all the hard work! It's coming along pretty well. I made a few style changes to match linter style, `eslint-disable` should not be necessary. As mentioned the only remaining work is styling the dialog. I would keep it as unbranded as possible. We can reuse the rotation modal dialog gray background add cancel, accept buttons and some text: `This VR site requires access to your sensors. Press accept to allow`

Image from iOS

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * Structure: <div><button></div>
+ *
+ * @param {function} onClick - click event handler
+ * @returns {Element} Wrapper <div>.
+ */
+
+function createDeviceMotionPermissionWindow (onClick, obj) {
+  var wrapper, innerWrapper, aframeBuilt;
+  var cancelBtn, continueBtn;
```
dmarcos: We do one variable declaration per line and we avoid abbreviations: cancelButton and cotinueButton

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,159 @@
...
+var DEVICE_PERMISSION_CANCEL_CLASS = 'a-device-motion-permission-cancel';
+var BUILT_WITH_AFRAME_CLASS = 'a-built-with-aframe';
+
+/**
+ * UI for enabling device motion permission
+ */
+module.exports.Component = registerComponent('device-motion-permission-ui', {
+  schema: {
+    enabled: { default: true },
+    request: {default: {orientationChange:null, deviceOrientation:null}},
```
dmarcos: Does this need to be configurable via schema? Do we need to handle deviceOrientation events explicitly or we can just rely on CSS?

src/components/scene/device-motion-permission-ui.js
```diff
@@ -1,4 +1,3 @@
-/* global DeviceOrientationEvent */
```
dmarcos: This is necessary for the linter

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: No, that was inadvertently left in from something earlier. What is now in place is the single component with the potential for one or both orientation events (device orientation and orientation change). The example on my site has two simple functions attached through the attributes, orientation-change and device-orientation. The drawback that I have on their usage at this point is that they use the window object to reference the functions. I'll hopefully get this done tomorrow with your edits. Thanks for your attention.

<a-scene device-motion-permission-ui orientation-change="noProof" device-orientation="proofOfWork">

https://planetkevin.com/aframe/examples/showcase/device-motion-permission/a.html

KevinEverywhere: `
function noProof(event){
--
  | console.log(event);
  | }
  | function proofOfWork(event){
  | console.log(event);
  | }
  | </script>
  |  
  | </head>
  | <body>
  | <a-scene device-motion-permission-ui orientation-change="noProof" device-orientation="proofOfWork">
  |  

`

KevinEverywhere:
Just about complete. Has new modal screen; will add clickable to it later today (November 14th)

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,161 @@
...
+    orientationChange: {default: null},
+    deviceOrientation: {default: null},
+    deviceMotionEl: { default: '' }
+  },
+  init: function () {
+    this.deviceMotionEl = null;
+    this.onDeviceMotionClick = bind(this.onDeviceMotionClick, this);
+    this.onOrientationChangeClick = bind(this.onOrientationChangeClick, this);
+    this.grantedDeviceMotion = bind(this.grantedDeviceMotion, this);
+    if (typeof window.orientation !== 'undefined') {
```
dmarcos: This if statement can be simplified to:
```
if (DeviceOrientationEvent && DeviceOrientationEvent.requestPermission) {
      this.deviceMotionEl = createDeviceMotionPermissionDialog(this.onDeviceMotionClick, this);
      this.el.appendChild(this.deviceMotionEl);
}
```
See [suggested edits](https://github.com/dmarcos/aframe/blob/b4be06e206821cedaa37a6b4512f4bd6641be1bd/src/components/scene/device-motion-permission-ui.js#L27)

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,161 @@
...
+        'Your device and application combination do not support device motion events.'
+      );
+    }
+  },
+
+  grantedDeviceMotion: function () {
+    const orientationChange=this.el.getAttribute('orientation-change');
+    const deviceOrientation=this.el.getAttribute('device-orientation');
+    this.remove();
+      if(orientationChange){
```
dmarcos: We probably don't need handling orientation changes here

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Create a button that when clicked will provide device motion permission.
+ *
+ * Structure: <div><button></div>
+ *
+ * @param {function} onClick - click event handler
+ * @returns {Element} Wrapper <div>.
+ */
+
+function createDeviceMotionPermissionWindow (onClick, obj) {
```
dmarcos: Rename to createDeviceMotionPermissionDialog, update argument names and docs above. [Suggested changes](https://github.com/dmarcos/aframe/blob/b4be06e206821cedaa37a6b4512f4bd6641be1bd/src/components/scene/device-motion-permission-ui.js#L69)

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: Agreed--before DeviceOrientationChange was included as a global, I was working around eslint errors. I'll include this in the next push.

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,161 @@
...
+  remove: function () {
+    if (this.deviceMotionEl) {
+      this.el.removeChild(this.deviceMotionEl);
+    }
+  },
+
+  /**
+   * Enable device motion permission when clicked.
+   */
+  onDeviceMotionClick: function () {
```
dmarcos: try catch not needed and redundant check of DeviceOrientationEvent. This method can be simplified. [Suggested edits](https://github.com/dmarcos/aframe/blob/b4be06e206821cedaa37a6b4512f4bd6641be1bd/src/components/scene/device-motion-permission-ui.js#L42)

src/style/aframe.css
```diff
@@ -242,6 +242,96 @@ a-scene audio {
   display: block;
 }
 
+.a-device-motion-permission-full{
```
dmarcos: I would probably made it a [gray opaque background](https://user-images.githubusercontent.com/39342/68625009-fcb6ea80-048c-11ea-9142-bdf80987620c.png) covering the whole scene with text and two big buttons. Smaller dialog / window partially covering the scene is harder to read in particular in iframed or embedded mode.

dmarcos:
Great job! Thanks for sticking with it. I made a few change suggestions. This is getting really close.

src/components/scene/device-motion-permission-ui.js
```diff
@@ -0,0 +1,159 @@
...
+    try {
+      /*eslint-disable */
```
dmarcos: Don't use `eslint-disable`. Linter issues should be addressed

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: The reasons for the structure here follow. I have tried these in different browsers and devices, including multiple iOS versions. There are other ways to do this, but going too simple or too complex will lead to similar outcomes. Thoughts?

iOS13
 - DeviceOrientationEvent.requestPermission needs to be called, not only tested to exist;
iOS and Android
 - grantedDeviceMotion is called by mobile devices (window.orientation)
Desktop Browsers
 - Do not call grantedDeviceMotion

dmarcos: DeviceOrientationEvent.requestPermission is already called when the user clicks on the allow button:
https://github.com/aframevr/aframe/pull/4303/files/8c30aaf2ce8b732dcc8b1644322fcad5f08e7036#diff-81b7296fca054a84c9266fd8fdebee05R75

Do we need to call it twice? 

lf think if `DeviceOrientationEvent.requestPermission` is not available we don't need to call any additional methods or initialize the dialog.

src/components/scene/device-motion-permission-ui.js
dmarcos: Sorry, there must be something I'm not understanding. Why does orientation changes relate to the dialog? Is not using CSS sufficient? What are the attributes above for?

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: Done.

src/components/scene/device-motion-permission-ui.js
dmarcos: What is the grantedDeviceMotion method for? It seems to be only handling orientation changes that are already handled by A-Frame.

KevinEverywhere:
Just pushed latest updates; with comments and most all critiques included.

subelsky:
thanks @KevinEverywhere! FYI I tested out your demo on Android 9 Chrome browser (& also on iOS 13 Safari), works well!

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: @dmarcos, grantedDeviceMotion unifies Android/iOS device orientation and orientation change methods. For a developer, it gives one entry point to ensure that event triggers are able to be handled in like fashion. It may be a moot issue for non-iOS 13 device/browsers combinations; it is a place where aframe events can be added confidently. The biggest use case is, if someone attempts to include Device Orientation behavior before permission is given, it might fail without the user knowing that it has failed. It provides a single place where events can be associated with user interaction; a place where aframe events may be attached with confidence.

src/style/aframe.css
KevinEverywhere: Agreed; this is done in current commit.

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: This has been updated to omit eslint-disable from the code.

src/components/scene/device-motion-permission-ui.js
dmarcos: Sorry, I'm still missing something. Could you help me understand by describing a concrete use case scenario? If there's an issue related to `orientationchange` events I feel it should be handled somewhere else. Regarding the solution, using global variable names passed through HTML attributes not associated to any component doesn't feel consistent with rest of A-Frame. I made a few cosmetic changes to the PR and removed for now. We can reintroduce if necessary.

dmarcos:
Thanks a lot everybody for the hard work and the testing. I adjusted a bit the code style to the rest of A-Frame, removed some redundant logic and tweaked the visual style of the modal dialog. I tested on iOS 13.2 on iPhone X and Chrome m78 on a Pixel 3 with Android 10. I don't have an iPad handy at the moment to test with. If you can let me know if it works and what you think about the dialog. The PR is almost ready. We only need docs for the `device-motion-permission-ui` component and some basic test coverage.

dialog

src/components/scene/device-motion-permission-ui.js
dmarcos: If the issue is notifying an application if DeviceOrientationEvent.requestPermission has succeeded or failed component events would be the way to go. e.g: `deviceorientationeventsgranted` / `deviceorientationeventsdenied`. Would that work?

KevinEverywhere:
@dmarcos , will you want me to complete this: the readme, for instance. Also, I had one last test that is not included here. iPad OS 13 defaults to Request Desktop Website. A person with an iPad would not know that this fails without having feedback. my solution is below. I tested on a number of devices and operating systems, desktop and mobile. This took over 40 hours of time to resolve something that kept aframe content from being functional on iOS/iPad 13 devices. Thanks.

  isMobile: function() {
    return  /iPhone|iPad|iPod|Android/i.test(navigator.userAgent);
  },
  grantedDeviceMotion: function () {
    if (!this.isMobile()) {
      alert('You appear to have a mobile device, but you may have requested a Desktop Website. Check the location bar of your browser and ensure you are requesting the Mobile Website.');
    } else {
...
}

The other place this would be used is after the window.orientation check:

      // This is for desktop browsers and those who cancel or reject
      if( this.isMobile() ){
        alert('You appear to have a mobile device, but you may have requested a Desktop Website. Check the location bar of your browser and ensure you are requesting the Mobile Website.');
      } else {
        this.remove();
      }

src/components/scene/device-motion-permission-ui.js
KevinEverywhere: Both orientation change and device orientation are granted permission by the same event, DeviceMotionEvent. That is why I included them both. Regarding the global variable name; I completely agree with you. I would think that there might be a way to pass the functions that depend on access to device motion to grantedDeviceMotion. That way, it is ensured that iOS/iPad will not have issues associated with attempting to include functionality before they have permission. One way or another, and it could be deviceorientationeventsgranted / deviceorientationeventsdenied, the developer should have access to a flag to know that it is possible to add a deviceorientation or orientationchange event. Thanks, and I am glad that I have been able to contribute to something that I needed to work! :)

dmarcos:
I added an alert dialog recommending the user to enable mobile browsing if Safari for iPad is on desktop mode. If someone with an iPad + iOS 13 can verify that this is working it will be super appreciated.

KevinEverywhere:
URL?

dmarcos:
@KevinEverywhere can you fetch the PR and test it locally? 

`npm run start:https` and any of the examples will do. e.g: `https://10.1.10.146:9000/examples/boilerplate/hello-world/`

KevinEverywhere:
I have an iPad and have tested this quite a bit. It appears that testing navigator.platform and multitouch will target iPad users whether or not they have Desktop Website Requested functioning. Also, it does not detect other tablet users. Window.orientation gives you that information. If you have a direct link to the built aframe.js file I could give it a test. In the meantime, this URL should work on Android and iPad tablets, as well as mobile phones of either platform.  This was submitted 12 days ago, and I have included the relevant JavaScript.  PS: I cannot access my dev network from my iPad but I can reference a URL of built aframe.js file.

https://planetkevin.com/aframe/examples/showcase/device-motion-permission/b.html

  isMobile: function() {
    return  /iPhone|iPad|iPod|Android/i.test(navigator.userAgent);
  },
init: function () {
    if (typeof window.orientation !== 'undefined') {
// Desktop and Desktop Website Requested (iPad, Android, as well as phones)
    } else {
      if( this.isMobile() ){   // This test is for mobile devices who do not have orientation enabled
        alert('You appear to have a mobile device, but you may have requested a Desktop Website. Check the location bar of your browser and ensure you are requesting the Mobile Website.');
      } else {
        this.remove();
      }


> Update on this: I found that iPad works for Firefox and Safari: in both cases, the applications had Desktop Version Requested. You either needed to set the preference in your preferences, or select the aA icon in the left of the location for Safari; or the three dots on the right side of Firefox, to Request Mobile Site. That means that what I have done is working across the target browsers on iOS and iPad browsers; I believe that it will also work on Android devices as well (not tested on many). The controller is in the bottom left of the screen to ask permission. For android devices, you will not need to select it.
> 
> https://planetkevin.com/aframe/examples/showcase/device-motion-permission/a.html

dmarcos:
Adding iPad to `isMobile` function will result in tablets entering stereo cardboard mode vs. fullscreen that I don't think we want. Trying to find the simpler set of conditions that also preserve previous functionality. I'll get a hold of an iPad with iOS 13 later in the day and test the PR. Thanks for the help.

subelsky:
@dmarcos am grabbing my iPad now and cloning the PR to test

KevinEverywhere:
@subelsky and @dmarcos; if anyone can post the output file for testing, that would be great. If either you or anyone else can compare what already is up at the URL below to work on all target browser/devices; I think that the cardboard/fullscreen worry will be mooted. Also, you can test your iPhone against this. Currently, if your iPhone is set to Desktop Website Requested, you do not have feedback to let you know. What I have is reverse compatible and covers iOS/iPad and Android.

https://planetkevin.com/aframe/examples/showcase/device-motion-permission/b.html

subelsky:
@KevinEverywhere here's what I did:
```
HOST=0.0.0.0 npm run dev
ngrok http 9000

subelsky:
attached is the build file which I produced via: `npm run build`:
[aframe.js.gz](https://github.com/aframevr/aframe/files/3860791/aframe.js.gz)

hope that helps! this feature is going to be awesome for a couple of projects, PSYCHED

dmarcos:
I changed the `mobile device requesting desktop site` condition to what @KevinEverywhere suggested and It works on an iPhone X with iOS 13 when requesting a desktop site. Also works on Android m78 / Pixel 3 / Android 10 when requesting mobile and desktop sites. All looks good. I'm going to write some tests and docs and I'll merge later in the day. We can adjust edge cases later if needed. Thanks all for your help.

KevinEverywhere:
Looks great, @dmarcos; checked it on iPad Pro with iPad 13 installed with Mobile and Desktop Website Request and it worked as expected on both.

dmarcos:
Documented and tested! Shipping 🚢 Thanks everybody for the hard work.

Dirk-27:
@dmarcos : Is this feature coming with 0.9.3 on 6th december?

dmarcos:
Yes but you can use it today via master builds: https://cdn.jsdelivr.net/gh/aframevr/aframe@934b8d358a2c458efdc8d65fcb6946f7303f0b51/dist/aframe-master.min.js

Let us know if you have any issues

tommensink:
Works very well! Is there any chance this will be included in Aframe 0.8.2? Since the breaking change of <a-animation> in  V0.9.2 my code does not work in that version, and I see no real possiblity to upgrade it in the near future. I suspect a lot of people will stay with 0.8.2 for the coming months/years because of this breaking chance

dmarcos:
@tommensink Unfortunately, We don’t patch previous releases. It’s not sustainable to maintain versions. What’s the breaking change that it’s preventing to upgrade to 0.9.2? a-animation got replaced by the [animation component](https://aframe.io/docs/0.9.0/components/animation.html#sidebar) but the API is the same. It should be easy to upgrade.

tommensink:
Thanks, and I understand. It is just that I use <a-animation> a lot and would not know where to begin to replace it in my code, since I use a nested structure with some sort of animation sequencer that I made. You can see it in https://hh-webar.firebaseapp.com/Presentation_2019-01-26/. (Just let the head rotate on your screen to see the different transparancy sequences kicking in).

tommensink:
My personal opinion is that moves from Apple like this motion-issue (which of course also can be understood, but nevertheless) and breaking Aframe chances will throw it back for some years to be adopted by the professional market. I for one will do the major serious AR/VR projects in Unity and use Aframe only for gadgets like dancing business cards

KevinEverywhere:
As a developer, the choice between narrower proprietary (higher paying) and
open-source (open, agreed-upon standards) solutions is never-ending. I
disagree with the idea that this throws aframe back years. It merely drives
innovation and should keep us vigilant and on our toes. It will be the
community as much as the vendor that will keep technologies current.
Nothing against Unity or profit-motive companies. I just like the idea of a
large and common playing field.

tommensink:
I hope I don't hijack this thread for this discussion too much, but one last thing then: You should hear me about breaking changes in Unity or their horrific tight integration of libraries like Vuforia (who break the API every release). But having said that, I am a full time AR developer, make dozens of apps as lead developer in four AR startups, including one of the first to do actual brain surgery with a HoloLens. So I have to choose the most appropriate and stable Entity Component Framework regularly. Funny moves like this one from Apple (and we had something similar by Android beginning of the year which requires the user to change a setting https://github.com/jeromeetienne/AR.js/issues/369#issuecomment-482047211) is one reason to choose another platform that will keep running no matter the OS change. An issue like this means that I will judge again in one or two years for Aframe in critical apps, hence my "Aframe is thrown back years" by this. And if the Aframe community decides they should make breaking changes to the API like <a-animation>, then this does not work in favour of this decision. Just my opinion, do think the major companies have similar considerations, but cannot speak for them. 
But I think you guys are awesome, this solution is great Kevin (I applied your first one which works with V0.8.2) and I always support AFrame in discussions and presentations as THE best open platform.

PR #4307, opened 2019-11-05
Don't set the obj-model attribute if it's not null

### LINKED ISSUE #4306: CORS error loading local controller model
**Description:**

- A-Frame Version: 0.9.2
- Platform / Device: Windows 10, Firefox 70.0.1 (64-bit), Vive Pro

```html
<a-scene>
	<a-assets>
		<a-asset-item id="controller-obj" src="models/vr_controller_vive.obj"></a-asset-item>
		<a-asset-item id="controller-mtl" src="models/vr_controller_vive.mtl"></a-asset-item>
	</a-assets>

	<a-entity
		id="leftHand"
		laser-controls="hand: left"
		obj-model="obj: #controller-obj; mtl: #controller-mtl"
	></a-entity>
	<a-entity
		id="rightHand"
		laser-controls="hand: right"
		obj-model="obj: #controller-obj; mtl: #controller-mtl"
	></a-entity>
</a-scene>
```

- Error: `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cdn.aframe.io/controllers/vive/vr_controller_vive.mtl. (Reason: CORS request did not succeed).`

It seems like [this line](https://github.com/aframevr/aframe/blob/05406c5ed92b7a202971c87c5b7f1cce08a1033a/src/components/vive-controls.js#L125) gets executed no matter if we passed a local model or not, which seems a bit weird to me (we might not want to use the controller model hosted on aframe.io).

mindrones:
Fixes #4306

**Description:**

Tentative PR (sorry I'm new to this code base): if this is the right way to go I'll add tests. 

**Changes proposed:**

Set the attribute `obj-model` on controllers only if it does not exist already on the target element.

### DIFF
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -107,7 +107,9 @@ module.exports.Component = registerComponent('daydream-controls', {
       idPrefix: GAMEPAD_ID_PREFIX,
       orientationOffset: data.orientationOffset
     });
-    if (!this.data.model) { return; }
+    if (!this.data.model || this.el.getAttribute('obj-model') !== null) {
+      return;
+    }
     this.el.setAttribute('obj-model', {
       obj: DAYDREAM_CONTROLLER_MODEL_OBJ_URL,
       mtl: DAYDREAM_CONTROLLER_MODEL_OBJ_MTL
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks. I just provided solution on https://github.com/aframevr/aframe/issues/4306
This PR would not solve the problem because one can add custom models in multiple ways: [primitives](https://aframe.io/docs/0.9.0/components/geometry.html#sidebar), [gltf](https://aframe.io/docs/0.9.0/components/gltf-model.html#sidebar) or in children of the entity. Better to use the [explicit toggle](https://github.com/aframevr/aframe/blob/master/docs/components/laser-controls.md#properties)

PR #4331, opened 2019-11-26
Add support for WebXR gamepads module (fix #4322)

### LINKED ISSUE #4322: Use WebXR session.inputSources[i].gamepad
WebXR now sends the state of controller buttons and axes as a `gamepad` attribute for each XR input source:

https://immersive-web.github.io/webxr-gamepads-module/#xrinputsource-interface

Using Gamepad API separately via navigator.getGamepads() or similar is deprecated:

https://immersive-web.github.io/webxr-gamepads-module/#navigator-differences
> However, Gamepad instances returned by an XRInputSource's gamepad attribute MUST NOT be included in the array returned by navigator.getGamepads().

AFrame should be updated to use session.inputSources[i].gamepad accordingly.

dmarcos:
Work in progress with initial support for the gamepads module. @klausw What's the recommended way to identify the controller vendor to load the appropriate model? I see that the gamepad id is not populated and there's an entry in the `XRInputSource .profiles` that says `oculus-touch` for the Oculus Rift headset. Is there a list of vendor strings anywhere? Is the vendor name always guaranteed to be in the first position of the profiles array? Thank you very much.

### DIFF
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -1,4 +1,3 @@
-
 var registerComponent = require('../core/component').registerComponent;
 var utils = require('../utils/');
 
--- a/src/components/laser-controls.js
+++ b/src/components/laser-controls.js
@@ -1,3 +1,4 @@
+
 var registerComponent = require('../core/component').registerComponent;
 var utils = require('../utils/');
 
@@ -13,8 +14,7 @@ registerComponent('laser-controls', {
     var data = this.data;
     var el = this.el;
     var self = this;
-    var modelEnabled = this.data.model && !this.el.sceneEl.hasWebXR;
-    var controlsConfiguration = {hand: data.hand, model: modelEnabled};
+    var controlsConfiguration = {hand: data.hand, model: true};
 
     // Set all controller models.
     el.setAttribute('daydream-controls', controlsConfiguration);
@@ -24,9 +24,6 @@ registerComponent('laser-controls', {
     el.setAttribute('vive-controls', controlsConfiguration);
     el.setAttribute('vive-focus-controls', controlsConfiguration);
     el.setAttribute('windows-motion-controls', controlsConfiguration);
-
-    // WebXR doesn't allow to discriminate between controllers, a default model is used.
-    if (this.data.model && this.el.sceneEl.hasWebXR) { this.initDefaultModel(); }
 
     // Wait for controller to connect, or have a valid pointing pose, before creating ray
     el.addEventListener('controllerconnected', createRay);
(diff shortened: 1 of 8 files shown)


arpu:
squeeze* input events for WebVR and WebXR
https://github.com/mrdoob/three.js/pull/17938 

something aframe needs too?

arpu:
on Samsung Odyssey i get this profiles:

```
length: 2
0: XRInputSource
handedness: "left"
profiles: Array(3)
0: "samsung-odyssey"
1: "windows-mixed-reality"
2: "generic-trigger-squeeze-touchpad-thumbstick"
```

arpu:
the strings comes from here https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/registry/profiles

dmarcos:
@Artyom17 I'm testing WebXR gamepad module support on Quest and realized that the profile is not populated in the Oculus Browser beta as per the [registry](https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/profiles/oculus/oculus-touch-s.json#L3). Can we assume that it will have the correct values when WebXR ships? Thank you very much for your help

Artyom17:
@dmarcos Yes, it seems like we agreed to name Oculus Touch controllers using the following schema: "oculus-touch" - Rift
"oculus-touch-v2" - Quest and Rift S
"oculus-touch-v3", etc - Next gens

However, Oculus Browser 7.0 is already ALMOST shipped and this change won't make into it. Hopefully, it will be in 7.1.

arpu:
what is with the oculus Go ?

Artyom17:
@arpu Yes. https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/profiles/oculus/oculus-go.json

Artyom17:
Also, we are going to support "samsung-gearvr" (at least for some time).

dmarcos:
This is ready to go. FYI, Oculus Browser 7 doesn't yet support the [WebXR gamepad module](https://github.com/immersive-web/webxr-gamepads-module). Interactions are limited to a single button (trigger) until Oculus Browser 7.1 ships. Logic is already wired (tested on Chrome with Rift) and should work with no additional changes.

dmarcos:
I believe this is the last pending PR for 0.9.3 release. We have to do a last QA pass on all devices. Any help testing and filing issues will be appreciated.

dmarcos:
This needs more QA. Chrome Canary and Beta have issues to enter VR today using WebXR and the latest Oculus Runtime and couldn't test thorougly.

klausw:
> Work in progress with initial support for the gamepads module. @klausw What's the recommended way to identify the controller vendor to load the appropriate model? I see that the gamepad id is not populated and there's an entry in the `XRInputSource .profiles` that says `oculus-touch` for the Oculus Rift headset. Is there a list of vendor strings anywhere? Is the vendor name always guaranteed to be in the first position of the profiles array? Thank you very much.

Sorry, I had missed being pinged. I hope you got this sorted out in the meantime, but for the record the intent is that you should go through the profiles array in order, and use the first profile name for which you find a match in the [registry](https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/registry/profiles). They are ordered from most-specific to least-specific, with middle entries being reasonably-compatible matches and the last one being a generic fallback with a standard naming scheme describing the available features.

PR #4345, opened 2019-12-12
webxr controller for chrome >= 79 on android and daydream

arpu:
**Description:**
webxr controller for chrome >= 79 on android and daydream


dmarcos:
There’s a linter issue

arpu:
updated

dmarcos:
Thanks

PR #4363, opened 2019-12-19
Android magic window

dmarcos:


src/components/look-controls.js
```diff
@@ -50,6 +51,24 @@ module.exports.Component = registerComponent('look-controls', {
     if (this.el.sceneEl.is('vr-mode')) { this.onEnterVR(); }
   },
 
+  setupMagicWindowControls: function () {
+    var magicWindowControls;
+    // Only on mobile devices and only enabled if DeviceOrientation permission has been granted.
+    if (utils.device.isMobile()) {
+      magicWindowControls = this.magicWindowControls = new THREE.DeviceOrientationControls(this.magicWindowObject);
+      if (typeof DeviceOrientationEvent === 'undefined' && DeviceOrientationEvent.requestPermission) {
```
vincentfretin: I think the condition is always false. How can requestPermission exist if DeviceOrientationEvent is undefined?

src/components/look-controls.js
```diff
@@ -50,6 +51,24 @@ module.exports.Component = registerComponent('look-controls', {
...
+        magicWindowControls.enabled = false;
+        if (this.el.sceneEl.components['device-orientation-permission-ui'].premissionGranted) {
```
vincentfretin: there is a typo here premissionGranted => permissionGranted

src/components/look-controls.js
dmarcos: Thanks!

src/components/look-controls.js
dmarcos: Good catch!

PR #4377, opened 2019-12-27
Change example code system name

Luftare:
**Description:**
In the docs 'my-component' naming is used for system registrations which feels confusing. Using 'my-system' would be more precise.

**Changes proposed:**
- Rename 'my-component' to 'my-system' in case of systems

### DIFF
--- a/docs/core/systems.md
+++ b/docs/core/systems.md
@@ -35,7 +35,7 @@ If the system name matches a component name, then the component will have a
 reference to the system as `this.system`:
 
 ```js
-AFRAME.registerSystem('my-component', {
+AFRAME.registerSystem('my-system', {
   schema: {},  // System schema. Parses into `this.data`.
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. [Systems are usually paired to a specific component](https://github.com/aframevr/aframe/tree/master/src/systems) hence the `my-component` name. Systems provide shared state and functionality for a specific component type but I've seen used them differently as well. Not sure what's the most appropriate naming.

Luftare:
For the documentation and example purpose that's ok now that you gave more insight into the way systems are used.

I have made an implementation of ECS architecture where systems are more decoupled from each other and only process entities' components and the components don't have any methods: https://github.com/Luftare/Ecs

Luftare:
I'll close this now as I find the proposed changes to be irrelevant.

Issue #4388, opened 2020-01-06
hand-controls hand model rotation no longer matches oculus quest touch controller rotation


dbradleyfl:
**Description:**
The hand-controls hand model is now rotated roughly 90deg off of the proper rotation on the Oculus Quest touch controls. Seems to be related to #4331 (I will investigate)
- A-Frame Version: 1.0.1 and 1.0.2 and 1.0.3
- Platform / Device: Oculus Quest
- Reproducible Code Snippet or URL: https://glitch.com/~aframe-1-0-3-hands

dbradleyfl:
Looks like a change was made to oculus-touch-controls that might cause the hand model rotation to no longer match #4374

dala00:
Oh, this may be same for my issue. I just created a sample.

https://glitch.com/~aframe-model-test

On this sample, gun rotation differs between HTTPS (like glitch) and HTTP (local). Copy this code and save to html file and start local server, and access to it by Oculus Quest, rotation differs.

PR #4396, opened 2020-01-13
rotate hand models if webxr api is being used

dbradleyfl:
**Description:**
Between version 1.0.0 and 1.0.1 the rotation of the hand-controls model changed such that they no longer match the users physical hand positions. They are rotated 90 off of where they should be. I created #4388 and figured I'd propose a solution.

Webxr controller orientation seems to be what is causing this improper rotation on the oculus quest browser (where the webxr api is avialable). In the firefox reality browser (no webxr api currently) on quest the rotation is correct. This change fixes the hand rotation if aframe is using the webXR api.

**Changes proposed:**
- check for xr on the navigator in the hand-controls component
- if xr available the hand models should be rotated
- if xr not available the hand models will match controller orientations

closed #4389 in favor of this solution

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -180,19 +180,19 @@ module.exports.Component = registerComponent('hand-controls', {
     // Get common configuration to abstract different vendor controls.
     controlConfiguration = {
       hand: hand,
-      model: false,
-      orientationOffset: {x: 0, y: 0, z: hand === 'left' ? 90 : -90}
+      model: false
     };
 
     // Set model.
     if (hand !== previousHand) {
       this.loader.load(MODEL_URLS[hand], function (gltf) {
         var mesh = gltf.scene.children[0];
+        var handModelOrientation = hand === 'left' ? Math.PI / 2 : -Math.PI / 2;
         mesh.mixer = new THREE.AnimationMixer(mesh);
         self.clips = gltf.animations;
         el.setObject3D('mesh', mesh);
         mesh.position.set(0, 0, 0);
-        mesh.rotation.set(0, 0, 0);
+        mesh.rotation.set(0, 0, handModelOrientation);
         el.setAttribute('vive-controls', controlConfiguration);
         el.setAttribute('oculus-touch-controls', controlConfiguration);
         el.setAttribute('windows-motion-controls', controlConfiguration);


dmarcos:
Why is the reported orientation different between WebVR and WebXR? Worth investigating to discard underlaying issues

dbradleyfl:
@dmarcos I thought the different orientations reported between VR and XR apis was a known thing based on the two different modelPivotRotation values for the oculus-touch-controls [here](https://github.com/aframevr/aframe/blob/9cf4677e4be00d561ce204cd5d35b732a742b956/src/components/oculus-touch-controls.js#L39). The oculus-touch-controls modelPivotRotation value changes  depending on which api is being used, so I figured the same treatment hadn't been applied to the hand-controls yet?

dmarcos:
The proper solution for this is getting read [rotationOffset](https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L184) and [apply the rotation to the model directly](https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L195). `tracked-controls-webxr` doesn't have a rotationOffset and hence the value provided is ignore with WebXR.

dbradleyfl:
Thanks for looking into this @dmarcos. Two questions:

By "getting rid of orientationOffset" wouldn't we break the WebVR implementation that depends on that [here](https://github.com/aframevr/aframe/blob/2a7b742b62664c4823bd14f4ab7aa436d9395685/src/components/tracked-controls-webvr.js#L203)?

By "apply the rotation to the model instead" do you mean open up the gltf file and rotate it, or do the rotation to the mesh on that line like I did [here](https://github.com/dbradleyfl/aframe/blob/c4208a6e8b0cb08c3da02b5de42776547e7d7581/src/components/hand-controls.js#L196)?

dmarcos:
I meant getting rid of this line: 

https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L184

And apply the same orientation to the hand model here:

https://github.com/aframevr/aframe/blob/master/src/components/hand-controls.js#L195

src/components/hand-controls.js
```diff
@@ -175,13 +175,13 @@ module.exports.Component = registerComponent('hand-controls', {
     var controlConfiguration;
     var el = this.el;
     var hand = this.data;
+    var handModelOrientation = hand === 'left' ? Math.PI / 2 : -Math.PI / 2;
```
dmarcos: We can move this declaration inside the model load handler:

https://github.com/aframevr/aframe/pull/4396/files#diff-02099d98da3e9b5683f5a6ac5092aec4R190

`

dmarcos:
Thanks for the effort. Much appreciated.

Sandstedt:
Seems like these changes has made it into v1.0.4 (or perhaps before), but I still experiences the issue of the hand is rotated.

Tried in Firefox 79 on Windows 10, on Oculus Quest using link.

dmarcos:
Use master? https://cdn.jsdelivr.net/gh/aframevr/aframe@f4bca789d34ab7d011da8bb6d6558229842aa47b/dist/aframe-master.min.js

Sandstedt:
@dmarcos I'm using the "default" https://aframe.io/releases/1.0.4/aframe.min.js. Not sure what branch that is from.

Also tried your cdn link, but it's the same problem.

Sandstedt:
Ok, I tried it in the Quest Browser, and there it work perfect. The wrongly rotated hand is only on **Desktop** Firefox 79. (In Chrome or Edge 84 I can't even get a-frame to run in VR mode). 

But as I understand it, desktop VR is a bit of a mess right now. So maybe call this the day and come back to it in a year :D

dmarcos:
Test with the master build I linked above: https://cdn.jsdelivr.net/gh/aframevr/aframe@f4bca789d34ab7d011da8bb6d6558229842aa47b/dist/aframe-master.min.js

dbradleyfl:
Is the Firefox browser still using the WebVR api (instead of WebXR)? Originally I saw the same issue with Firefox on Quest, and had an if / else that checked webxr and applied the rotation differently, but I think we decided to remove that. I can't recall why.

dmarcos:
Yes Firefox is still on WebVR

Sandstedt:
@dmarcos Already tried that, and same result.

But maybe this is an edge case? I just use it this way while developing. But if it's the same using Rift S, then its probably a bug worth checking out.

PR #4397, opened 2020-01-13
Prevent jumping tracked controllers due to switing between 6DOF and 3DOF

zach-capalbo:
**Description:**

I've noticed a very annoying issue when using my Rift S with A-Frame applications. When the controllers are held still near the periphery of the field of view, they can jump back and forth quite a lot. The issue goes away when they are moved around a bit, but then reappears if they are held still for a while.

This is on Firefox 72.0.1 and various aframe versions, including 1.0.3

I've tracked it down, and it seems that when the controller is in the center of view or moving, the WebVR controller pose contains the `position` attribute, but when it is stationary on the periphery, it does not contain the `position` attribute. When the `position` attribute is missing, the `tracked-controls-webvr` instantly switches to using the 3DOF arm model, putting the controller in a completely different position. When the controller starts moving again, it switches back to 6DOF mode, again causing a very annoying jump as it goes back to the tracked position.

**Changes proposed:**

- Keep track of whether the controller pose ever contains the `position` attribute. If it does, then do not switch to 3DOF arm model mode if `position` goes away.

### DIFF
--- a/src/components/tracked-controls-webvr.js
+++ b/src/components/tracked-controls-webvr.js
@@ -42,7 +42,7 @@ module.exports.Component = registerComponent('tracked-controls-webvr', {
     idPrefix: {type: 'string', default: ''},
     orientationOffset: {type: 'vec3'},
     // Arm model parameters when not 6DoF.
-    armModel: {default: true},
+    armModel: {default: false},
     headElement: {type: 'selector'}
   },
 
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -19,7 +19,7 @@ module.exports.Component = registerComponent('tracked-controls', {
     idPrefix: {type: 'string', default: ''},
     orientationOffset: {type: 'vec3'},
     // Arm model parameters when not 6DoF.
-    armModel: {default: true},
+    armModel: {default: false},
     headElement: {type: 'selector'},
     iterateControllerProfiles: {default: false}
   },


zach-capalbo:
The WebXR behavior on desktop Chrome in such cases seems to be that the model disappears. I've attached a small video to demonstrate:

https://gitlab.com/zach-geek/vartiste/uploads/f6ff07cdf20d5b696b32f8ed4dcecdd8/2020-01-13_19-07-05.mkv

I don't have access to an Oculus Quest, so I can't try the Oculus browser.

It seems in the webxr case, the model visibility is set to `false`, even with `autoHide` set to `false`. However, it doesn't exhibit the position jumping behavior that this pull request is meant to address.

dmarcos:
Probably simple to set armModel to default.

dmarcos:
Let's simply do:

`armModel: {default: false},` in [tracked-controls-webvr](https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls-webvr.js#L45)

zach-capalbo:
I've set `armModel: {default: false}` in `tracked-controls-webvr`, but I've had to set it in `tracked-controls` as well, since it overwrites the `tracked-controls-webvr` data in https://github.com/aframevr/aframe/blob/67d568c263173ef0a56efc1e0d9e08a6f2d2d2a8/src/components/tracked-controls.js#L38

It looks like the controls that would need an arm model (`daydream-controls`, gearvr-controls`, etc) explicity set `armModel` to `true`, so this shouldn't cause any issues there.

dmarcos:
Thanks! I appreciate the patience and the effort. Congrats on your first contribution 🎉

PR #4402, opened 2020-01-14
[docs] Add AFRAME.utils.shouldCaptureKeyEvent() to Utils docs

dsinni:
**Description:**

Adds the useful `AFRAME.utils.shouldCaptureKeyEvent()` utility method to the documentation.

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -223,3 +223,19 @@ string.
 AFRAME.utils.getUrlParameter('testing');
 // If visiting the current page with ?testing=aframe, this will log 'aframe'.
 ```
+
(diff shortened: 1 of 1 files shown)


dmarcos:
That's an internal method. Feel free to use at your own peril 😄 but documenting comes with maintenance expectations. We can consider making it public if we see clear demand.

dsinni:
Ah, ok. Thanks, @dmarcos.

I was just about finishing up similar logic to avoid issues when entering text in a UI, and stumbled upon this method, which I ended up using. Just thought it could save some people quite a bit of time if it were documented. 

Totally understand if it's not something you want to advertise, though. I _am_ curious what your reservations are, however. Any issues you've noticed when using it?

Thanks again.

PR #4409, opened 2020-01-18
Listens to WebXR selectstart / selectstart and reemits as mousedown / mouseup

dmarcos:


dmarcos:
This is a first approach to handle `selectstart` / `selectend` WebXR. They're remitted as `mousedown` / `mouseup` for consistency with other cursor events using controllers, mouse and regular touch events. Ideally we should probably have a `touchscreen-controls` component that handles this similarly to other controllers.

PR #4414, opened 2020-01-21
Add Magic Leap Control Support. Add profile absence fallback handling.

atarng-magicleap:
**Description:**
Currently Aframe does not have support for the magic leap controller.
In fact, it will throw an exception if the current XR session (or browser) does not support webxr profiles.  

**Changes proposed:**
-This patch will fallback to "default" behaviors if profiles are not found.
- Added profile for magic leap controller.


atarng-magicleap:
ah, yes of course, I can adjust the naming. Thanks Diego!

PR #4421, opened 2020-01-23
Add hand and quality properties on hand-controls 

arturitu:
... to manage three different 3D models with different quality (toon / low / high)

### DIFF
--- a/examples/showcase/tracked-controls/index.html
+++ b/examples/showcase/tracked-controls/index.html
@@ -60,8 +60,8 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
       <!-- Hands -->
-      <a-entity hand-controls="hand: left; handModelStyle: low;" aabb-collider="objects: .cube;" grab></a-entity>
-      <a-entity hand-controls="hand: right; handModelStyle: low;" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="hand: left; handModelStyle: lowPoly;" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="hand: right; handModelStyle: lowPoly;" aabb-collider="objects: .cube;" grab></a-entity>
       <a-entity position="0 1.6 2" camera look-controls wasd-controls></a-entity>
     </a-scene>
   </body>
--- a/examples/showcase/tracked-controls/index.html
+++ b/examples/showcase/tracked-controls/index.html
@@ -60,8 +60,8 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
       <!-- Hands -->
-      <a-entity hand-controls="left" aabb-collider="objects: .cube;" grab></a-entity>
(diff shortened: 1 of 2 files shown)


arturitu:
I wanted to keep something that works with the old hand-controls="left" (so I wouldn't mess up old codes) and add quality as a new property, but I didn't know if is possible to do it.
With this PR you need to have something like hand-controls="hand: left; quality: low;"

examples/showcase/tracked-controls/index.html
```diff
@@ -60,8 +60,8 @@
          <a-entity light="type: ambient; color: #f4f4f4; intensity: 0.4;" position="-8 10 -18"></a-entity>
       </a-entity>
       <!-- Hands -->
-      <a-entity hand-controls="left" aabb-collider="objects: .cube;" grab></a-entity>
-      <a-entity hand-controls="right" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="hand: left; handModelStyle: low;" aabb-collider="objects: .cube;" grab></a-entity>
+      <a-entity hand-controls="hand: right; handModelStyle: low;" aabb-collider="objects: .cube;" grab></a-entity>
```
dmarcos: Maybe `lowPoly` instead of low?

src/components/hand-controls.js
```diff
@@ -3,8 +3,12 @@ var registerComponent = require('../core/component').registerComponent;
 
 // Found at https://github.com/aframevr/assets.
 var MODEL_URLS = {
-  left: 'https://cdn.aframe.io/controllers/hands/leftHand.glb',
-  right: 'https://cdn.aframe.io/controllers/hands/rightHand.glb'
+  toon_left: 'https://cdn.aframe.io/controllers/hands/leftHand.glb',
```
dmarcos: We don’t use underscores: toonLeft, lowPolyLeft...

src/components/hand-controls.js
```diff
@@ -44,7 +48,12 @@ EVENTS[ANIMATIONS.point] = 'pointing';
  * @property {string} Hand mapping (`left`, `right`).
  */
 module.exports.Component = registerComponent('hand-controls', {
-  schema: {default: 'left'},
+  // schema: {default: 'left'},
+  schema: {
+    hand: { default: 'left' },
+    // toon, low or high
+    handModelStyle: { default: 'low' }
```
dmarcos: We need the values in the schema `{default: ‘lowPoly’, oneOf: [‘lowPoly’, ‘highPoly’, ‘toon’]}`

src/components/hand-controls.js
```diff
@@ -185,7 +193,7 @@ module.exports.Component = registerComponent('hand-controls', {
 
     // Set model.
     if (hand !== previousHand) {
-      this.loader.load(MODEL_URLS[hand], function (gltf) {
+      this.loader.load(MODEL_URLS[handModelStyle + hand.charAt(0).toUpperCase() + hand.slice(1)], function (gltf) {
```
dmarcos: We can move this to a variable above to clean up a bit.
`var handmodel = MODEL_URLS[handModelStyle + hand.charAt(0).toUpperCase() + hand.slice(1);`

dmarcos:
Thank you! Great work and thanks for the patience!

Issue #4419, opened 2020-01-23
Docs still talk about A-Frame Registry and Awesome Aframe, both of which are discontinued


omgitsraven:
**Description:**

- A-Frame Version: master
- Platform / Device: documentation
- Reproducible Code Snippet or URL: https://aframe.io/docs/master/introduction/writing-a-component.html

At the bottom of the page, the "publishing a component" section says that "you should publish it to the ecosystem via the A-Frame Registry and awesome-aframe!"

However, https://github.com/aframevr/aframe-registry says that the A-Frame Registry is discontinued, and https://github.com/aframevr/awesome-aframe says that Awesome A-Frame is discontinued :)

The "components" link on what's left of the Awesome A-Frame readme implies that the modern approach is just to publish an NPM package with the `aframe` keyword, but I'm not sure if that's still accurate? Are there any more recent instructions about this that the official docs could be updated to include?

dmarcos:
Thanks. A PR to remove references will be appreciated

omgitsraven:
What should they be replaced with, though?

dmarcos:
I would just remove the references to the registry for now

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/4423

PR #4423, opened 2020-01-24
Remove references to discontinued sites

omgitsraven:
As discussed here: https://github.com/aframevr/aframe/issues/4419 "The A-Frame Registry" and "Awesome A-Frame" are both now discontinued. I'm not sure if there's anything that should be added in their place (how are components selected for the Inspector these days, for example?) but the last word in the Issue was "just remove the references to the registry for now", so that's what I've done. (I've also inserted the alternatives that the registry's discontinuation readme suggests in its place.)


dmarcos:
Nice! Thanks for doing this.

PR #4424, opened 2020-01-24
Add Magic Leap Controller Support

atarng-magicleap:
This pull request addresses three issues.

One, defines the Magic Leap Component to support the new XR session profiles.
Two, addresses an issue where the controllers are being cached, despite closing the XRsession, and so when requesting a new session, chromium/webxr rejects the session due to the previous, now invalid, session id stored in the controller.
Three, based on available documentation for XRHandedness, there is support for input devices that do not specify a specific "handedness". This should be accounted for and allow these input events to still send button events.

References:
https://immersive-web.github.io/webxr-reference/webxr-device-api/xrinputsource.html
https://chromium.googlesource.com/chromium/src/+/ab6c1c84a115d5a05c6827de7a33b70fbbb8e5ea/device/vr/public/mojom/vr_service.mojom

### DIFF
--- a/src/components/magicleap-controls.js
+++ b/src/components/magicleap-controls.js
@@ -1,0 +1,178 @@
+var bind = require('../utils/bind');
+var registerComponent = require('../core/component').registerComponent;
+
+var trackedControlsUtils = require('../utils/tracked-controls');
+var checkControllerPresentAndSetup = trackedControlsUtils.checkControllerPresentAndSetup;
+var emitIfAxesChanged = trackedControlsUtils.emitIfAxesChanged;
+var onButtonEvent = trackedControlsUtils.onButtonEvent;
+
+// See Profiles Registry:
+// https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/registry
+// TODO: Add a more robust system for deriving gamepad name.
+var GAMEPAD_ID_PREFIX = 'magicleap';
+var GAMEPAD_ID_SUFFIX = '-one';
+var GAMEPAD_ID_COMPOSITE = GAMEPAD_ID_PREFIX + GAMEPAD_ID_SUFFIX;
+
+var MAGICLEAP_CONTROLLER_MODEL_GLB_URL = 'https://cdn.aframe.io/controllers/magicleap/magicleap-one-controller.glb';
+
+/**
+ * Button IDs:
+ * 0 - trigger
+ * 1 - grip
+ * 2 - touchpad
+ * 3 - menu (never dispatched on this layer)
+ *
+ * Axis:
+ * 0 - touchpad x axis
+ * 1 - touchpad y axis
+ */
+var INPUT_MAPPING_WEBXR = {
+  axes: {touchpad: [0, 1]},
+  buttons: ['trigger', 'grip', 'touchpad', 'menu']
+};
+
+/**
+ * Magic Leap Controls
+ * Interface with Magic Leap control and map Gamepad events to controller
+ * buttons: trigger, grip, touchpad, and menu.
+ * Load a controller model.
+ */
+module.exports.Component = registerComponent('magicleap-controls', {
+  schema: {
+    hand: {default: 'none'},
+    model: {default: true},
+    orientationOffset: {type: 'vec3'}
+  },
+
+  mapping: INPUT_MAPPING_WEBXR,
+
+  init: function () {
+    var self = this;
+    this.controllerPresent = false;
+    this.lastControllerCheck = 0;
+    this.onButtonChanged = bind(this.onButtonChanged, this);
+    this.onButtonDown = function (evt) { onButtonEvent(evt.detail.id, 'down', self); };
+    this.onButtonUp = function (evt) { onButtonEvent(evt.detail.id, 'up', self); };
+    this.onButtonTouchEnd = function (evt) { onButtonEvent(evt.detail.id, 'touchend', self); };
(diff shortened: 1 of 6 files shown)


src/components/magicleap-controls.js
```diff
@@ -0,0 +1,192 @@
...
+ * 0 - trigger (intensity value from 0.5 to 1)
+ * 1 - grip
+ * 2 - touchpad
+ * 3 - menu (dispatch but better for menu options)
+ *
+ * Axis:
+ * 0 - touchpad x axis
+ * 1 - touchpad y axis
+ */
+var INPUT_MAPPING_WEBVR = {
```
dmarcos: Do you think we can remove WebVR references? My understanding is that there are no legacy WebVR browsers with magicleap support

atarng-magicleap: Yes, I missed this! Thanks for catching it!

src/systems/tracked-controls-webxr.js
```diff
@@ -19,7 +19,16 @@ module.exports.System = registerSystem('tracked-controls-webxr', {
   updateControllerList: function () {
     var xrSession = this.el.xrSession;
     var self = this;
-    if (!xrSession) { return; }
+    if (!xrSession) {
```
dmarcos: Maybe better:

```
if (!xrSession) { 
 if (this..oldControllersLength === 0) { return; }
 this.oldControllersLength = 0;
 this.controllers = [];
 this.el.emit('controllersupdated', undefined, false); 
}

File: src/components/tracked-controls-webxr.js [hunk 3]
```diff
@@ -62,17 +62,26 @@ module.exports.Component = registerComponent('tracked-controls-webxr', {
   },
 
   emitButtonDownEvent: function (evt) {
-    if (!this.controller || evt.inputSource.handedness !== this.data.hand) { return; }
+    if (!this.controller) { return; }
```
dmarcos: Why is this needed?

src/components/tracked-controls-webxr.js
```diff
@@ -62,17 +62,26 @@ module.exports.Component = registerComponent('tracked-controls-webxr', {
...
+
     this.selectEventDetails.state.pressed = true;
     this.el.emit('buttondown', this.selectEventDetails);
     this.el.emit('buttonchanged', this.selectEventDetails);
     this.el.emit('triggerdown');
   },
 
   emitButtonUpEvent: function (evt) {
-    if (!this.controller || evt.inputSource.handedness !== this.data.hand) { return; }
+    if (!this.controller) { return; }
```
dmarcos: Why is this needed? What's particular about magic leap with respect to other controllers?

dmarcos: Is magicleap controller always returning `none` as handness or is it user configurable too like Oculus Go?

src/components/magicleap-controls.js
```diff
@@ -0,0 +1,172 @@
...
+  },
+
+  onModelLoaded: function (evt) {
+    var controllerObject3D = evt.detail.model;
+    // our glb scale is too large.
+    controllerObject3D.scale.set(0.01, 0.01, 0.01);
+  },
+
+  onAxisMoved: function (evt) {
+    emitIfAxesChanged(this, this.mapping[this.data.hand].axes, evt);
```
atarng-magicleap: no need for handedness here (in fact, this is undefined).

src/components/tracked-controls-webxr.js
atarng-magicleap: regarding this, we currently do not specify a handedness for our controllers (although, in theory it is possible for us to recognize which hand is currently holding the controller, and set it at runtime; however, we currently do not have that planned as a feature for our controllers).

Since Chromium Code enumerates a NONE state, and the webxr reference does mention a value of "" is possible, it is my personal believe that there should be handling for such a scenario where it is neither left nor right. (I'm actually not sure why the reference page lists an empty string, since it seems like it's going to set to none).

src/components/tracked-controls-webxr.js
```diff
@@ -62,17 +62,24 @@ module.exports.Component = registerComponent('tracked-controls-webxr', {
   },
 
   emitButtonDownEvent: function (evt) {
-    if (!this.controller || evt.inputSource.handedness !== this.data.hand) { return; }
-    if (this.controller.gamepad) { return; }
+    if (!this.controller || this.controller.gamepad) { return; }
```
dmarcos: Maybe factor out this code and the one below in to a method `isControllerPresent`?

dmarcos:
In case you haven't seen it. I left a comment on https://github.com/aframevr/assets/pull/37 that might affect this PR.

dmarcos:
Just in case it's not too hard. Do you have a short video you can share of Helio and A-Frame using hand-controls to tweet about once this lands?

atarng-magicleap:
Of course! I'll go ahead prepare a video today!

src/components/tracked-controls-webxr.js
dmarcos: I think if we factor this code out the PR is ready to go. Thanks for the patience.

atarng-magicleap: okay, yes I think I can do so!

dmarcos:
Thank you! Great effort and thanks again for the patience!

Utopiah:
@atarng-magicleap can you please share a link to the video example?

PR #4425, opened 2020-01-25
link component: allows setting previewDistance

DougReeder:
**Description:**
For the link component, allows setting the distance within which the preview image is shown. The default is good for standard size portals used when virtually walking, but the larger portals required by higher speeds (or cyclopean architecture) need to let the preview image be seen at a greater distance.  See https://elfland-glider.surge.sh/yggdrasil/ 

**Changes proposed:**
- The default is 15m, as before, so this will have zero effect on links that don't set the previewDistance attribute.
- Can also be set on the a-link primitive, as preview-distance.
- In the link-traversal example, the center link has its previewDistance doubled.


DougReeder:
The tests are failing in CI because "apt-get install failed".  They all pass on my machine.

examples/showcase/link-traversal/index.html
```diff
@@ -23,7 +23,7 @@
       </a-assets>
       <a-link href="city.html" position="-3.5 1.5 -1.0" image="#thumbCity"></a-link>
       <a-link href="sunrise.html" position="0 1.5 -1.0" image="#thumbSunrise"></a-link>
-      <a-link href="mountains.html" position="3.5 1.5 -1.0" image="#thumbMountains"></a-link>
+      <a-link href="mountains.html" position="3.5 2.0 -1.0" image="#thumbMountains" scale="2 2 2" preview-distance="30"></a-link>
```
dmarcos: Can we leave example as it was?

src/components/link.js
```diff
@@ -15,6 +15,7 @@ module.exports.Component = registerComponent('link', {
     highlightedColor: {default: '#24CAFF', type: 'color'},
     href: {default: ''},
     image: {type: 'asset'},
+    previewDistance: {default: 15, min: 1},
```
dmarcos: This seems hard to understand and puts the burden on the developer. What about adapting the distance automatically based on the scale of the entity?

dmarcos:
I think if anything this could be automated instead of adding additional properties to the component

examples/showcase/link-traversal/index.html
DougReeder: Sure.  It's just hard to understand the need for this, without a larger-scaled link.

src/components/link.js
DougReeder: That sounds doable.

examples/showcase/link-traversal/index.html
DougReeder: I've removed the changes.  You can transiently set the scale of a link to see the effects now.

src/components/link.js
DougReeder: Done.

DougReeder:
Does the current PR meet your standards?

src/components/link.js
```diff
@@ -28,6 +28,7 @@ module.exports.Component = registerComponent('link', {
     this.quaternionClone = new THREE.Quaternion();
     // Store hidden elements during peek mode so we can show them again later.
     this.hiddenEls = [];
+    this.previewDistance = 15;
```
dmarcos: Add this as a constant at the top. Similar to https://github.com/aframevr/aframe/blob/master/src/components/gearvr-controls.js#L15

src/components/link.js
```diff
@@ -199,7 +205,7 @@ module.exports.Component = registerComponent('link', {
       cameraWorldPosition.setFromMatrixPosition(camera.matrixWorld);
       distance = elWorldPosition.distanceTo(cameraWorldPosition);
 
-      if (distance > 20) {
+      if (distance > this.previewDistance + 5) {
```
dmarcos: Why `+ 5` and not making previewDistance `25`?

dmarcos:
Thanks for the patience. I made a few more comments.

src/components/link.js
DougReeder: I left the functionality unchanged, when the scale is unchanged.  

The original code in master shows the preview when the camera is closer than 15m, and turns the link to face the camera when the camera is farther than 20m (you could call that the "link-lookAt distance'). In between, the center of the link just shows red, but the link isn't turned to face the camera.

I vary the preview distance, rather than the link-lookAt distance, because neither should be zero or less.  Also, the preview distance is used in several places in the code, and link-lookAt distance only one.

Any change in the relationship between these parameters or change to the behavior of links with scale 1, should be done in a different PR, I think.

src/components/link.js
DougReeder: fixed by commit 312efe07522cef3abc1b1f8a4f66dad6a1c9643f

dmarcos:
Thank you!

PR #4427, opened 2020-01-27
added +/- for clarity

dirkk0:
Initially, I wasn't aware that shadowBias could also have negative values.
For example, in a scene I needed a value of -0.001 to make the artifacts go away, as discussed in this thread:
https://aframevr.slack.com/archives/C0FABCPJQ/p1578640473070500

With the +/-, I hope to hint that people try out both positive and negative values.

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thank you!

PR #4431, opened 2020-01-31
Docs: Animation API Example Updated for 1.0.3

ChicagoDev:
**Description:**

Changed the API examples for animation to be compatible with A-Frame 1.0.3. Old examples do not work in the latest version of A-Frame. Additionally, I slightly edited the examples so the shapes can be seen animating without camera movement and in more notable colors.


dmarcos:
Thank you! Congrats on your first PR 🥇

PR #4433, opened 2020-02-03
Material dithering

Firepal:
**Description:**
Made material dithering available from the component and enabled by default as it is much more preferable than without.

Here's two images of a lit cube seen from above, with contrast edited in-post to show the difference dithering makes. 
image


dmarcos:
Thanks. We need to document the new property: https://github.com/aframevr/aframe/blob/master/docs/components/material.md

Firepal:
@dmarcos donesies

dmarcos:
Thanks!

PR #4434, opened 2020-02-05
not addState  vr-mode and emit enter-vr on requestFullscreen

arpu:
if user gets on fullscreen not emit vr-mode and enter-vr

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -333,7 +333,6 @@ module.exports.AScene = registerElement('a-scene', {
             window.dispatchEvent(event);
           }
 
-        
           // Lock to landscape orientation on mobile.
           if (!isWebXRAvailable && self.isMobile && screen.orientation && screen.orientation.lock) {
             screen.orientation.lock('landscape');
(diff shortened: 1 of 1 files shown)


dmarcos:
Why is this needed?

dmarcos:
Fullscreen is fallback when entering VR and  no headset or polyfill is available. Different ways to resolve the enter vr intent and that’s why the event is emitted.

arpu:
hm i use the enter-vr event to detect user is in vr,  on fullscreen the user experience should be the same as not in fullscreen

dmarcos:
Now that we removed the polyfill it makes sense to not add the state or emit the event when entering fullscreen mode. Needs rebase and I can merge. Thanks

mrxz:
Not sure it's a good idea to break the symmetry between `enterVR` and `exitVR` like this. It would mean that calling `enterVR` could result in a state (fullscreen) where calling `exitVR` throws an error. It would make more sense to have distinct "fullscreen" and "VR/immersive session" modes with their corresponding methods.

I'm also slightly concerned that many users might rely on the fact that the `enter-vr` event is emitted when clicking on the fullscreen/XR/AR/VR button. For example resuming the `AudioContext` or hiding a splash/menu screen. So if we're going to change this, it might be nice to have a unified enter/exit event across modalities (fullscreen or XR).

dmarcos:
Closing for now. We can revisit in the future. Thanks everyone

PR #4435, opened 2020-02-05
add powerPreference: "high-performance"  to webgl renderConfig

arpu:
from https://www.khronos.org/registry/webgl/specs/latest/1.0/

```
high-performance

Indicates a request for a GPU configuration that prioritizes rendering performance over power consumption. Developers are encouraged to only specify this value if they believe it is absolutely necessary, since it may significantly decrease battery life on mobile devices. Implementations may decide to initially respect this request and, after some time, lose the context and restore a new context ignoring the request.
```

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -589,7 +589,8 @@ module.exports.AScene = registerElement('a-scene', {
           alpha: true,
           antialias: !isMobile,
           canvas: this.canvas,
-          logarithmicDepthBuffer: false
+          logarithmicDepthBuffer: false,
+          powerPreference: "high-performance"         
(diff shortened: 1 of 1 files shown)


dmarcos:
Has this been tested to have any effect on any browser?

arpu:
looks like this works in chrome on dual gpu settings 
https://bugs.chromium.org/p/chromium/issues/detail?id=1005428&sort=-modified&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified&q=high-performance&can=2

dmarcos:
Thanks

PR #4436, opened 2020-02-06
  iterateControllerProfiles set true for windows mixed reality window…

arpu:
…s-motion-controls

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thanks. Was not this working before?

arpu:
not on the samsung odyssey it always fallback to the gerneric

dmarcos:
Thanks!

PR #4441, opened 2020-02-07
light: Add THREE.RectAreaLight as option

Firepal:
**Description:**
Area lights as an option. 
The intensity is much darker than other light types, so I multiply the intensity by 32, and also I attenuate the intensity if the area light gets larger.
I can't tell how to do the attenuation math every frame though, so it's just on initial load.

### DIFF
--- a/docs/components/light.md
+++ b/docs/components/light.md
@@ -43,11 +43,11 @@ To manually disable the defaults, without adding other lights:
 
 All light types support a few basic properties:
 
-| Property  | Description                                                     | Default Value |
-|-----------|-----------------------------------------------------------------|---------------|
-| type      | One of `ambient`, `directional`, `hemisphere`, `point`, `spot`. | directional   |
-| color     | Light color.                                                    | #fff          |
-| intensity | Light strength.                                                 | 1.0           |
+| Property  | Description                                                             | Default Value |
+|-----------|-------------------------------------------------------------------------|---------------|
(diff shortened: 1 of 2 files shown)


donmccurdy:
My impression was that the RectAreaLight type is pretty expensive – what device(s) are you using this with?

Firepal:
@donmccurdy ...Well that's a bit embarrasing. I didn't think to check performance on a HMD or a mobile device. My initial impression of one (1) RectAreaLight on my laptop seemed good, but I'll have to test with multiple RectAreaLights on my Oculus Quest.

Firepal:
Misclick. Damn phone screen. I'm still wondering how to do targeting, like for spot and directonal lights, since RectAreaLight doesn't have a .target property.

Firepal:
@donmccurdy You were correct. As opposed to my initial laptop glance, performance on Oculus Quest tested with [this edited version of Hello World](https://firepal.tilde.team/3D/hello-world/) is... not great at all. I have no idea how I didn't think for a second about performance.

I guess using a spot light with a 90° angle and a penumbra value of 1 will have to be good enough!

donmccurdy:
Ah, that's too bad. Thanks for the PR in any case! I've had some amount of luck baking rect area lights in Blender, that's another option: https://raw.githack.com/donmccurdy/three.js/feat-lightprobevolume-build/examples/?q=volume#webgl_lightprobe_volume

PR #4443, opened 2020-02-09
Missing colons for width attribute

richardjbaker:
**Description:**
Missing colons on the width attribute.  The a-entity rendered ok on codepen without it but when I used glitch.com the colon was required.


dmarcos:
Thanks! Congrats on your first contribution 👍

PR #4452, opened 2020-02-13
Cordova mobile app compatibility

benallfree:
# Description

This PR provides experimental Cordova mobile app compatibility.


dmarcos:
Any way to detect cordova implicitly?

dmarcos:
`'cordova' in window` ?

src/index.js
```diff
@@ -35,20 +12,54 @@ if (utils.isIE11) {
...
-       'HTML.');
+// Polyfill `Promise`.
+window.Promise = window.Promise || require('promise-polyfill');
+
+var isIOSOlderThan10 = require('./utils/isIOSOlderThan10');
+// Workaround for iOS Safari canvas sizing issues in stereo (webvr-polyfill/issues/102).
+// Only for iOS on versions older than 10.
+var bufferScale = isIOSOlderThan10(window.navigator.userAgent) ? 1 / window.devicePixelRatio : 1;
+
+var aframeConfig = _.merge({
```
dmarcos: I would make all this automatic and implicitly if a `cordova` environment is detected

package.json
```diff
@@ -43,6 +43,7 @@
     "deep-assign": "^2.0.0",
     "document-register-element": "dmarcos/document-register-element#8ccc532b7f3744be954574caf3072a5fd260ca90",
     "load-bmfont": "^1.2.3",
+    "lodash": "^4.17.15",
```
dmarcos: use `deep-asign`. We can move to something else in the future if needed

src/index.js
```diff
@@ -1,27 +1,4 @@
-// Polyfill `Promise`.
-window.Promise = window.Promise || require('promise-polyfill');
```
dmarcos: Order if this code is important. Don't move if not necessary.

src/index.js
```diff
@@ -35,20 +12,54 @@ if (utils.isIE11) {
...
+        'HTML.');
+  }
+
+  // Error if not using a server.
+  if (window.location.protocol === 'file:') {
+    error(
+      'This HTML file is currently being served via the file:// protocol. ' +
+      'Assets, textures, and models WILL NOT WORK due to cross-origin policy! ' +
+      'Please use a local or hosted server: ' +
+      'https://aframe.io/docs/0.5.0/introduction/getting-started.html#using-a-local-server.');
```
dmarcos: link pointing to an old docs version.

package.json
benallfree: deep-assign fails to merge null values correctly. Thoughts?

dmarcos: Avoid nulls or cycle through the object keys replacing nulls. lodash is a pretty big dependency.

benallfree: Will do. I think it doesn’t matter anyway if embedding Cordova checks is the preferred approach. Much less code changes. Will update this PR soon.

benallfree:
@dmarcos Ready for another review, but please do not merge yet. Still testing in conjunction with AR.js.

dmarcos:
What version of Chrome is Cordova shipping with? Can someone build apks for Android that use WebXR?

benallfree:
I will have to check. I know for iOS I need to use some polyfills that make native calls but mimic web APIs like getUserMedia. I think it also needs a WebVR polyfill. I’m not extremely familiar with the VR support on Android but I’ll check. 

Does the answer alter anything needed for this PR?

dmarcos:
I’m not educated on Cordova. If someone wants to package a WebXR for Rift / SteamVR / Quest we would need a way to enter VR automatically. It seems you are focused on Magic Window

benallfree:
iOS WebKit has some missing APIs that even Mobile Safari allows. So for embedded WebKit like Cordova uses, polyfills are needed even though the browser is capable. Kind of frustrating but that’s the status currently. 

I’ll keep researching though.

benallfree:
Curious. Has any research been done on A-Frame + React Native? https://github.com/supermedium/aframe-react/issues/35

dmarcos:
> 
> 
> I am curious: Has any research been done on A-Frame + React Native? [supermedium/aframe-react#35](https://github.com/supermedium/aframe-react/issues/35)

I haven't heard of any examples. In order for A-Frame to work on React Native a WebGL backend and  DeviceMotion APIs would be needed. Not sure if that's possible at the moment.

What is the status of this PR?

benallfree:
> What is the status of this PR?

I am working on getting some A-Frame demos working in a sample Cordova project living here: https://github.com/benallfree/arjs-mobile-app-samples

My ultimate goal is to get AR.js working too, but right now it appears that there are a few rendering issues with A-Frame that I am looking into.

benallfree:
@dmarcos For example, the Hello WebVR looks like this on the iPhone simulator running Cordova with an embedded WebView:

image

benallfree:
@dmarcos This is roughly what I need to do in Cordova (dynamically load A-Frame) because WebKit is not available/initialized until after Cordova's `deviceready` event: https://glitch.com/~wide-puppy-po6d3fgfy

That demo works on Mobile Safari, but does not work when embedded in Cordova.

dmarcos:
Make sure you enable https://aframe.io/docs/1.0.0/components/renderer.html#antialias

dmarcos:
Notice that `ar.js` imposes additional restrictions on top of A-Frame like access to the camera. Making A-Frame `hello-world` example work is the first step as you are doing.

benallfree:
We are getting closer. This is running in Cordova on a physical iPhone device. The iOS simulator chokes. 

You can see there are a few bordering issues and the zoom appears wrong but it's getting close. I will keep you updated in this thread :)

dmarcos:
Great work and thanks for the update! Much appreciated.

benallfree:
It looks like these lines are getting executed but Browserify is not injecting the `<script>` tags:

```javascript
if (utils.device.isBrowserEnvironment || window.cordova) {
  require('./style/aframe.css');
  require('./style/rStats.css');
}
```

Looking into it further. Any ideas?

dmarcos:
> It looks like these lines are getting executed but Browserify is not injecting the `<script>` tags:
> 
> ```js
> if (utils.device.isBrowserEnvironment || window.cordova) {
>   require('./style/aframe.css');
>   require('./style/rStats.css');
> }
> ```
> 
> Looking into it further. Any ideas?

dmarcos:
> > It looks like these lines are getting executed but Browserify is not injecting the `<script>` tags:
> > ```js
> > if (utils.device.isBrowserEnvironment || window.cordova) {
> >   require('./style/aframe.css');
> >   require('./style/rStats.css');
> > }
> > ```
> > 
> > 
> > Looking into it further. Any ideas?

Yes, those are executed by browserify at build time. What's the problem with those styles in Cordova?

benallfree:
I'm debug tracing now. It looks like Browserify does successfully insert them into `<head>`, but then they disappear from the DOM again. I'm going to see if Cordova is modifying `<head>` and somehow is removing the `<script>` tags that A-Frame inserts.

benallfree:
Success!! Now the `Hello WebVR` demo is working flawlessly. I was calling `document.write()` [after the document was loaded](https://stackoverflow.com/questions/10873942/document-write-clears-page).



Working demo in https://github.com/benallfree/arjs-mobile-app-samples/tree/master/cordova-aframe-demos/hello-webvr

dmarcos:
Awesome. Congratulations. Well done! 👏

benallfree:
@dmarcos Maybe you can help with this... as you can see, when I press the VR button it appears to pop up a video window. Do you know what causes this?

dmarcos:
No idea why a video window is showing up. I can guarantee that A-Frame is not sneaky loading a video 🙂 Cordova might get confused somehow when entering fullscreen, may be when locking orientation...

benallfree:
This is causing that strange popup https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L352

benallfree:
It's coming from https://github.com/richtr/NoSleep.js/blob/master/src/index.js#L60, which seems to be in both cardboard and webvr-polyfill.

Basically it's launching an invisible video, but on webkit that pops a modal video player.

More investigation to come...

benallfree:
Hello WebVR demo seems pretty perfect now. Any artifacts I'm noticing are also present in the browser-based demo.

benallfree:
@dmarcos I'm working on the metaverse demo. It loads assets, and since those are `file://` assets in Cordova, XHR does not work. I'm retrofitting an old XHR polyfill that intercepts `file://` requests to use Cordova's [File](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-file/index.html) plugin to read locally. The File plugin is API-compatible with [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) standard.

I think this will also help with the `DPDB_URL` piece.

Ultimately, I think this kind of thing as well as the `cordova.css` modifications are probably better suited for a `cordova-aframe-plugin` package.

So a few of the mods will be in the A-Frame core here, but  Cordova-specific CSS and polyfills will be in an official Cordova plugin that will bundle aframe-master.js.

So Cordova users will just use

```bash
% cordova plugin install cordova-aframe-plugin
```

And that will correctly configure and load the aframe bundle and related CSS/polyfills. That means Cordova users wouldn't install the `aframe` package directly, but would install and use it via the Cordova plugin. Which means we'd need to keep that plugin up to date with new aframe releases.

What do you think?

dmarcos:
It sounds good to me. Any changes still needed on the A-Frame core side?

benallfree:
Yes. I’ll update this PR. 

I’m having trouble with Metaverse demo. No animation or fonts working, even after the XHR polyfill. Maybe I need to try some raw three.js demos to see if the problem is deeper than aframe. It seems that way.

benallfree:
@dmarcos I have the first cut at the Cordova plugin and updated the description at the top of this PR. Please review the proposed core changes and also the to-do list and let me know your thoughts.

If the animation and font rendering issues turn out to be coming from a different package, then I don't expect A-Frame will need any further modification. I'll run through at least a few more of the A-Frame demos to see how they handle, but things are looking really good!

src/index.js
```diff
@@ -1,3 +1,6 @@
+// Package metadata
+var pkg = require('../package');
+
```
benallfree: Necessary to hoist this because the `file:` check uses it. (You asked me to correct the version number, I made it dynamic).

src/index.js
```diff
@@ -35,26 +42,28 @@ if (utils.isIE11) {
...
+        'HTML.');
+  }
+
+  // Error if not using a server.
+  if (window.location.protocol === 'file:') {
+    error(
+      'This HTML file is currently being served via the file:// protocol. ' +
+      'Assets, textures, and models WILL NOT WORK due to cross-origin policy! ' +
+      'Please use a local or hosted server: ' +
+      'https://aframe.io/docs/' + pkg.version + '/introduction/installation.html#use-a-local-server.');
```
benallfree: Per your request I updated this to always point to the latest version number. Not sure if this is correct :/

src/index.js
```diff
@@ -35,26 +42,28 @@ if (utils.isIE11) {
...
+  }
 }
 
 require('present'); // Polyfill `performance.now()`.
 
 // CSS.
-if (utils.device.isBrowserEnvironment) {
+if (utils.device.isBrowserEnvironment || window.cordova) {
```
benallfree: For some reason isBrowserEnvironment is `false` when running in Cordova. I can try to fix this, or we can just add a check here as I have done.

src/index.js
```diff
@@ -35,26 +42,28 @@ if (utils.isIE11) {
...
-}
-
-// Error if not using a server.
-if (window.location.protocol === 'file:') {
-  error(
-    'This HTML file is currently being served via the file:// protocol. ' +
-    'Assets, textures, and models WILL NOT WORK due to cross-origin policy! ' +
-    'Please use a local or hosted server: ' +
-    'https://aframe.io/docs/0.5.0/introduction/getting-started.html#using-a-local-server.');
+if (!window.cordova) {
```
benallfree: Only do these checks if NOT in Cordova mode.

src/index.js
```diff
@@ -19,6 +22,10 @@ if (!window.hasNativeWebXRImplementation && !window.hasNativeWebVRImplementation
     CARDBOARD_UI_DISABLED: true,
     ROTATE_INSTRUCTIONS_DISABLED: true
   };
+  if (window.cordova) {
+    polyfillConfig.DPDB_URL = null;
```
benallfree: I would like this to be configurable so we can use a `file:` db rather than `null` or a remote URL. Thoughts?

src/index.js
```diff
@@ -19,6 +22,10 @@ if (!window.hasNativeWebXRImplementation && !window.hasNativeWebVRImplementation
...
+    polyfillConfig.MOBILE_WAKE_LOCK = false;
```
benallfree: This is to avoid a mobile video player popping in Cordova mode

src/components/daydream-controls.js
```diff
@@ -6,7 +6,7 @@ var checkControllerPresentAndSetup = trackedControlsUtils.checkControllerPresent
 var emitIfAxesChanged = trackedControlsUtils.emitIfAxesChanged;
 var onButtonEvent = trackedControlsUtils.onButtonEvent;
 
-var DAYDREAM_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/google/';
+var DAYDREAM_CONTROLLER_MODEL_BASE_URL = window.AFRAME_CONFIG.ASSET_BASE_URL + 'controllers/google/';
```
dmarcos: Why is this necessary?

src/config.js
```diff
@@ -0,0 +1,40 @@
...
+    CARDBOARD_UI_DISABLED: true,
+    ROTATE_INSTRUCTIONS_DISABLED: true
+  },
+  debug: {
+    warnIfNotHead: true,
+    warnIfFileProtocol: true
+  }
+};
+
+function deepMerge (carry, toMerge) {
```
dmarcos: Use `AFRAME.utils.extendDeep`

src/components/daydream-controls.js
benallfree: Just updated PR description at the top. Relying on a network connection is undesirable for mobile apps. So, these assets are bundled in the `cordova-plugin-aframe` package and loaded statically from a `file://` URL.

src/config.js
benallfree: Will fix

benallfree:
@dmarcos I see you are reviewing already, just a heads up that I reverted back to a global config override strategy using `window.AFRAME_CONFIG` due to how Cordova needs to control the asset base URL.

Loading A-Frame in Cordova is actually more complex than imagined. First we have to wait for a `deviceready` document event so we can compute the asset base URL, then we need to load A-Frame. Then we need to load Cordova-specific CSS overrides :)

In all, it felt much cleaner to refactor into a config strategy rather than pepper `window.cordova` conditions throughout the A-Frame codebase.

src/index.js
```diff
@@ -35,20 +31,20 @@ if (utils.isIE11) {
...
 
 // Error if not using a server.
-if (window.location.protocol === 'file:') {
+if (window.AFRAME_CONFIG.debug.warnIfFileProtocol && window.location.protocol === 'file:') {
   error(
     'This HTML file is currently being served via the file:// protocol. ' +
     'Assets, textures, and models WILL NOT WORK due to cross-origin policy! ' +
     'Please use a local or hosted server: ' +
-    'https://aframe.io/docs/0.5.0/introduction/getting-started.html#using-a-local-server.');
+    'https://aframe.io/docs/' + pkg.version + '/introduction/installation.html#use-a-local-server.');
```
dmarcos: This url is not correct. There are only doc versioning for major versions.

src/index.js
```diff
@@ -121,3 +115,6 @@ module.exports = window.AFRAME = {
   utils: utils,
   version: pkg.version
 };
+
+document.dispatchEvent(new window.Event('aframe-ready'));
```
dmarcos: Why is this necessary?

src/index.js
```diff
@@ -35,20 +31,20 @@ if (utils.isIE11) {
 var error = debug('A-Frame:error');
 var warn = debug('A-Frame:warn');
 
-if (window.document.currentScript && window.document.currentScript.parentNode !==
+if (window.AFRAME_CONFIG.debug.warnIfNotHead && window.document.currentScript && window.document.currentScript.parentNode !==
```
dmarcos: Do we need configurable warnings? Can we just check if it's a cordova environment?

src/index.js
```diff
@@ -35,20 +31,20 @@ if (utils.isIE11) {
...
   warn('Put the A-Frame <script> tag in the <head> of the HTML *before* the scene to ' +
-       'ensure everything for A-Frame is properly registered before they are used from ' +
-       'HTML.');
+      'ensure everything for A-Frame is properly registered before they are used from ' +
+      'HTML.');
 }
 
 // Error if not using a server.
-if (window.location.protocol === 'file:') {
+if (window.AFRAME_CONFIG.debug.warnIfFileProtocol && window.location.protocol === 'file:') {
```
dmarcos: Do we need configurable warnings? Can we just check if it's a cordova environment?

src/config.js
```diff
@@ -0,0 +1,29 @@
...
+var isIOSOlderThan10 = require('./utils/isIOSOlderThan10');
+// Workaround for iOS Safari canvas sizing issues in stereo (webvr-polyfill/issues/102).
+// Only for iOS on versions older than 10.
+var bufferScale = isIOSOlderThan10(window.navigator.userAgent)
+  ? 1 / window.devicePixelRatio
+  : 1;
+
+var DEFAULT_CONFIG = {
+  ASSET_BASE_URL: 'https://cdn.aframe.io/',
+  polyfillConfig: {
```
dmarcos: I prefer to check internally if it's cordova instead of increasing public API surface that we would have to document, maintain and be sensible when changed.

src/index.js
dmarcos: Unrelated to this PR. We can handle separately.

src/config.js
```diff
@@ -0,0 +1,29 @@
...
+    warnIfNotHead: true,
+    warnIfFileProtocol: true
+  }
+};
+
+var extendDeep = require('./utils').extendDeep;
+
+var finalConfig = extendDeep({}, DEFAULT_CONFIG, window.AFRAME_CONFIG);
+
+console.log({ finalConfig });
```
dmarcos: Is this logging necessary?

src/index.js
benallfree: The bootstrap sequence in Cordova goes like this:

1. Wait for `deviceready` event
2. Inject `aframe-master.js` script tag
3. Wait for `aframe-master.js` to inject `aframe.css` and `rStats.css` link tags
4. Insert `cordova-aframe.css` link tag

The `aframe-ready` is how I know that A-Frame is done injecting its css tags.

What else can I do to wait for A-Frame to finish loading?

src/config.js
benallfree: Agreed, a few questions then:

1. How can I change the hard-coded `cdn.aframe.io` references to point to local resources instead?
2. How can I point polyfillConfig's `DPDB_URL` to a local resource?

src/components/daydream-controls.js
dmarcos: Is not fair to assume that most mobile devices will have internet connection? Most apps today don't work offline.

src/index.js
dmarcos: Injecting your css after A-Frame should be enough unless cordova operates differently. I don't think you need to wait or an extra event. Loading them in order should work e.g:

```
var el = document.createElement('script');
el.src = 'aframe-master.js';
el.async = false;
document.head.appendChild(el);

var link  = document.createElement('link');
link.rel  = 'stylesheet';
link.type = 'text/css';
link.href = 'http://yoururl.com/style.css';
document.head.appendChild(link);
```

benallfree: A-Frame's css doesn't get injected until the next tick. [This glitch](https://glitch.com/~turquoise-grand-eocursor) shows the situation.

src/components/daydream-controls.js
benallfree: It depends. Apple sometimes rejects apps that don't work offline. Or, I might be designing an app specifically made to work offline or just LAN. Same with desktop/Electron. I think there are valid app use cases for loading assets locally rather than hitting a CDN. Same goes for the polyfillConfig DPDB_URL and other data files loaded by A-Frame dependencies. 

If you feel like these changes don't belong in the core, another idea I had was to maintain a branch or fork, basically a custom A-Frame build, that gets bundled with [cordova-plugin-aframe](https://github.com/benallfree/cordova-plugin-aframe) since that's how people will be using the package anyway.

Do you think that would be better than what this PR is asking to do? Even with the changes I'm proposing in this PR, this package will never work in Cordova by itself. It needs other Cordova-specific dependencies like `file://`-based XHR and Cordova's File API.

src/index.js
dmarcos: Load the styles from a JS file so they're both sync. Your example modified:

https://glitch.com/edit/#!/tundra-cotton?path=cordova-aframe.js:10:22

src/components/daydream-controls.js
dmarcos: Let’s try to accommodate in A-Frame itself. I think we can converge on something surgical. 

FYI, Those models won’t be loaded on iOS. Maybe we can change those url when need arises?

benallfree: Can Android and iOS devices attach to various external controllers? I'm not familiar with that.

In any case I know `/fonts` does get loaded. That's what motivated the creation of a dynamic base URL (see `src/components/text.js`).

Also, Cordova targets Electron. Those controllers could be available there. 

Thinking in terms of Cordova compatibility, the asset base URL seems like a predictable issue that Cordova users will face.

src/index.js
benallfree: We can remove this. Depending on the asset base URL and polyfillConfig strategies you're comfortable with, I may need to maintain a custom build anyway. If that happens, I will keep this event because your solution adds a layer of indirection that feels unnecessarily complex for me as a Cordova user. I understand you want to minimize core changes though.

benallfree:
@dmarcos You've persuaded me to a very surgical approach.

The only remaining issue is that the A-Frame canvas seems to take full screen control regardless of the parent container size, but the VR button is relative to the parent container size. 

The CSS mod fixes that, but I believe it might be an appropriate mod for non-Cordova as well. [see glitch](https://glitch.com/edit/#!/admitted-befitting-wrinkle?path=index.html:40:4)

benallfree:
We will not need a separate aframe fork or even a specialized Cordova plugin at this time.

The 'offline' requirement caused a lot of complexity, thank you for resisting it. 

I still think it would be wise to consider making A-Frame's bootstrap more configurable by refactoring the asset base URL and polyfillConfig settings, but that can be another project for another time.

Please let me know what you think of the CSS issue. I don't think it belongs in this PR but there definitely appears to be an unintended side effect if the <a-scene> is nested in a <div>.

benallfree:
Ok this one is ready to be merged & closed.

Thank you for all the assistance. I created a separate issue for the CSS thing I found.

dmarcos:
Thanks for the patience. This work is super appreciated. 

Do you know what version of Chromium ships with? WebXR ships in Chrome since m79. We've seen people asking for ways to package A-Frame apps and distribute them on Steam / Oculus Store. I think we're pretty close to get something usable but not a Cordova expert

benallfree:
Interesting, do you have anyone you can put me in touch with?

I don't have an Android device so I haven't done any Android testing with this, just iOS. 

This is definitely just a first effort, I expect we will find out lots of little details once someone tries to ship an app :)

dmarcos:
I believe @arpu expressed interest in the past on packaging A-Frame apps (sorry if my memory is not accurate).

dmarcos:
Thanks for this. Sorry it took so long to merge.

Issue #4458, opened 2020-02-14
Fix problems with Magic Leap Controller


ericdacal:
**Description:**

- A-Frame Version: 1.0.4
- Platform / Device: Magic Leap
- Reproducible Code Snippet or URL:
I have test Magic Leap Controller with Magic Leap 1, and when I used I find the following bug:
![error](https://user-images.githubusercontent.com/16444893/74533079-09862700-4f31-11ea-9aec-ee55ba7c65cb.png)
My code is very simple, I think that is sometime with the controller:
![code](https://user-images.githubusercontent.com/16444893/74536807-1a3a9b00-4f39-11ea-8399-2bf924545d9b.png)

Thanks

dmarcos:
@atarng-magicleap I don’t have a device to test on. Do you have any ideas what it might be going on?

ericdacal:
It seems that WebXR don't detect Magic Leap device correctly. I think is something like this. In this link explain how to access magic leap controls from WebXR.

[https://developer.magicleap.com/learn/guides/webvr-webxr]()

atarng-magicleap:
Hi, correct. Magic Leap is currently still using a version of chromium that does not support input profiles. This should be supported in the next software update.

As far as that error it looks like the profiles field needs to be checked against, before checking against the length. (this may have been something that my previous patch neglected to do)

atarng-magicleap:
As a workaround, inside the file tracked-controls.js, changing:
`if (profiles.length === 0) { return; }`
to
`if (!profiles || profiles.length === 0) { return; }`
should remove the error.
However full input profile support will be ready with the next software release which should be sometime in the next few weeks.

dmarcos:
Thanks everybody. I would probably wait until the Magic Leap soft fix

ericdacal:
Hi, there is a new software version for Magic Leap 1 and seems that now support profiles.  
https://developer.magicleap.com/en-us/learn/guides/webxr-overview
I probe with the new version and now there is not error but the controllers don't show the model and seems that don't work.  My code is simple:

![image](https://user-images.githubusercontent.com/16444893/80472948-bf182200-8945-11ea-98fc-23ef621f0413.png)



Thanks

dmarcos:
Any errors on console?

ericdacal:
There isn't any error in console. But finally I solved the error removing an if in utils/tracked-controllers. 
[...]

PR #4471, opened 2020-02-18
Gracefully handle lack of profiles from Controller (old definition in…

atarng-magicleap:
… Mojom)

Magic Leap version 0.98 is still using an older version of WebXR that does not support the profiles field on its input sources.

This change allows it to gracefully continue on despite not having this field.

(Also, since this is an array, I think it can continue looking for a proper controller rather than returning.)

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -162,7 +162,7 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index, 
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     profiles = controller.profiles;
-    if (profiles.length === 0) { return; }
+    if (!profiles || profiles.length === 0) { continue; }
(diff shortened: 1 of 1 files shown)


atarng-magicleap:
https://github.com/aframevr/aframe/issues/4458

dmarcos:
What are the plans to support profiles in the ML browser? I would prefer to avoid hacks if possible.

atarng-magicleap:
Profiles should be supported within the next few weeks. I created this mostly dependent on if there was any urgency to resolve this issue.

dmarcos:
@atarng-magicleap Thanks for staying on top of it. Much appreciated. We can keep the PR open as a reference if someone needs a quick fix until `profiles` ship in the Magic Leap Browser.

dmarcos:
I believe this has been already fixed on Magic Leap side?

dmarcos:
closing due to inactivity.

PR #4497, opened 2020-02-25
fix exponential deceleration in wasd-controls (fixes #4241)

### LINKED ISSUE #4241: wasd-controls flies for tens of meters during framedrops with a single key press.
**Description:**

When wasd-controls became independent of frame rate, the behavior during frame drops worsened to flying at high-speeds across the world.

ntamas:
**Description:**

This PR fixes a problem with the implementation. of exponential deceleration in `wasd-controls`. The current implementation yields sudden stops when two frames are rendered in quick succession (significantly faster than the expected frame rate) and sudden jumps (in the order of several meters) in case of significant frame drops.

**Changes proposed:**

The velocity update has been fixed according to the Stack Overflow question originally linked in the source code.

See also the discussion in #4241.

### DIFF
--- a/src/components/wasd-controls.js
+++ b/src/components/wasd-controls.js
@@ -101,10 +101,10 @@ module.exports.Component = registerComponent('wasd-controls', {
     var scaledEasing = Math.pow(1 / this.easing, delta * 60);
     // Velocity Easing.
     if (velocity[adAxis] !== 0) {
-      velocity[adAxis] -= velocity[adAxis] * scaledEasing;
+      velocity[adAxis] = velocity[adAxis] * scaledEasing;
     }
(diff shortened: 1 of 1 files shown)


ntamas:
Tested experimentally with the Anime UI example.

dmarcos:
Awesome! Thanks for investigating and sharing the fix. Do you think it needs more testing?

ntamas:
I think it's fine and ready to be merged. I've tested it quite extensively in Firefox and Chrome on macOS and I doubt that it would be OS-dependent. If someone has access to Windows machines running IE / Edge, it might be worth testing it there a bit.

dmarcos:
Cool thanks! Congrats for your first contribution 🥇

PR #4500, opened 2020-02-26
Merged latest mrdoob/three.js to Superframe and A-Frame

jimdose:
Hello.  I work at ForwardXP.  We're working on a project using A-Frame.  Our gltf models uses sRGB textures, which render dark in the current three.js A-Frame uses.  To fix this I forked superframe/three.js and aframevr/aframe to merge the code and make any necessary changes, which included removing support for WebVR in AFrame.  I can understand there may be reasons you haven't merged the latest three.js (such as the WebVR change), but if not, please consider this pull request.


dmarcos:
I’m not sure I understand the purpose of this PR. Master uses THREE r114. You can use [master builds](https://cdn.jsdelivr.net/gh/aframevr/aframe@68fd0460ea8b9d471fa8ce10aaf71ac53c91de32/dist/aframe-master.min.js). Closing this for now but can reopen if necessarg

dmarcos:
Keeping WebVR is intentional. Not browsers have ship the spec and a lot of content still uses WebVR. That should not affect how your sRGB textures are rendered

jimdose:
Yeah, I figured that you might not want to remove the WebVR yet.  In my case it didn't matter since our project targets Oculus Quest.

In terms of sRGB textures, though, there is a difference between between the superframe/three.js and mrdoob/three.js in that you can set:
    renderer.outputEncoding = THREE.sRGBEncoding;

and it will render at the proper gamma.  Without that the textures are much darker.  I've just verified that setting outputEncoding to srgb does not fix the texture gamma with https://aframe.io/releases/1.0.4/aframe.min.js.

Anyway, no worries about the fork.  Thanks for your help!

dmarcos:
@jimdose FYI, master does that for you already: https://github.com/aframevr/aframe/commit/b8d66dbe550a5b410a8bc3a7a1a1fd0fa2873b2b#diff-05e4dad4b0e4792a1721dd241358c018L37

Use master-builds: https://cdn.jsdelivr.net/gh/aframevr/aframe@da6ba49fe7429d7287211151db846b297877ba71/dist/aframe-master.min.js

jimdose:
Finally got a chance to try out the master build and the link you gave me works!  I had been using https://aframe.io/releases/1.0.4/aframe.min.js in my tests previously.

Thanks for you're help! :)

PR #4509, opened 2020-03-02
Add webxr-dom-overlay system and example

klausw:
**Description:**

This system supports configuring WebXR DOM Overlay mode as an optional or required feature.

**Changes proposed:**
- add the system and an example


src/systems/renderer.js
```diff
@@ -21,7 +21,10 @@ module.exports.System = registerSystem('renderer', {
     colorManagement: {default: false},
     gammaOutput: {default: false},
     alpha: {default: true},
-    foveationLevel: {default: 0}
+    foveationLevel: {default: 0},
+    xrSessionInit: {default: {
```
dmarcos: We don't have proper support for arbitrary objects in the schema. What about doing to `properties` instead:
```
webXRrequiredFeatures: {type:'array', default: ['local-floor']},
webXROptionalFeatures: {type:'array', default: ['bounded-floor'])}
```

src/systems/webxr-dom-overlay.js
```diff
@@ -0,0 +1,21 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * WebXR DOM Overlay support.
+ */
+module.exports.System = registerSystem('webxr-dom-overlay', {
+  schema: {
+    root: {default: document.body, type: 'selector'},
+    required: {default: false, type: 'boolean'}
```
dmarcos: Maybe calle it `enabled` for consistency with other components?

klausw: This is something different - if required is true, this is added to WebXR's `requiredFeatures`, so that session creation fails if it's not supported. If false, it's added to `optionalFeatures`, so the session can proceed without a DOM Overlay if that's not supported. It's enabled (if supported) in either case.

src/systems/webxr-dom-overlay.js
```diff
@@ -0,0 +1,21 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * WebXR DOM Overlay support.
+ */
+module.exports.System = registerSystem('webxr-dom-overlay', {
+  schema: {
+    root: {default: document.body, type: 'selector'},
```
dmarcos: `rootElement` for consistency

src/systems/webxr-dom-overlay.js
```diff
@@ -0,0 +1,21 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * WebXR DOM Overlay support.
+ */
+module.exports.System = registerSystem('webxr-dom-overlay', {
```
dmarcos: maybe we can call it just `dom-overlay`? We can explain nuances in docs.

dmarcos:
This is pretty cool. Just made a few comments / suggestions about naming. We need to add some docs: pointing to the spec, saying its experimental, necessary flags, limitations...

src/systems/renderer.js
dmarcos: Also wondering if we should move this out of `renderer` component and move it to a `webxr` one. It might be cleaner.

dmarcos: e.g: `<a-scene webxr="requiredFeatures: local-floor, dom-overlay, optionalFeatures: bounded-floor, dom-overlay>`

klausw: That works for those two specific arrays, but we still need an attachment point where systems can modify the raw init argument, for example setting `domOverlay.root`.

I've updated this, what do you think?

src/systems/webxr-dom-overlay.js
klausw: Done.

src/systems/webxr-dom-overlay.js
klausw: Done, apart from the missing docs.

src/systems/renderer.js
klausw: About your example, I think it's a bit clunky and not very future-proof if scenes need to maintain an explicit list of features. Having a system such as _dom-overlay_ automatically add to the arrays may be a bit of an abstraction layer break, but I think it's a lot more convenient, and makes it easier for people to mix&match systems and components in the future.

I agree that this would be cleaner in a _webxr_ component, but maybe use names such as "requiredReferenceSpaces" and "optionalReferenceSpaces" to make it clearer that this will be used as input to creating the XR init dictionary, without implying that it'll be an exhaustive list?

dmarcos: we could have ` webxr` systems that has an internal reference to the session object that others could access for experimentation. It seems that `renderer` and `webxr` are different things and worth decoupling. Maybe something like this:

```
module.exports.System = registerSystem('webxr', {
  schema: {
     requiredFeatures: {type: 'array', default: ['local-floor']},
     optionalFeatures: {type: 'array', default: ['bounded-floor']},
     overlayElement: {type: 'selector'}
  },
 update: function () {
    var data =  this.data;
    this.sessionConfiguration = {
        requiredFeatures: data.requiredFeatures,
        optionalFeatures: data.optionalFeatures
     }

     if (data.overlayElement) {
        this.sessionConfiguration.domOverlay = {root: data.overlayElement};
     }
  }
});
```
Experimentation can happen by accessing `sceneEl.systems.webxr.sessionConfiguration`

src/systems/dom-overlay.js
```diff
@@ -0,0 +1,21 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * WebXR DOM Overlay support.
+ */
+module.exports.System = registerSystem('dom-overlay', {
+  schema: {
+    rootElement: {default: document.body, type: 'selector'},
+    required: {default: false, type: 'boolean'}
```
dmarcos: When would someone pick required or optional?

klausw: An application would use "required" if it's essentially unusable without the DOM Overlay, for example if the main information it displays is in the DOM, or if critical user interface elements only exist as DOM elements. 

It would use "optional" if it still works on a device where the overlay isn't supported, either because the functionality is secondary (i.e. a convenience "Exit AR" button), or because it has alternate input or display methods that don't need it. For example, the RC car sample I've been experimenting with uses the DOM overlay to show a screen controller, but an AR headset wouldn't need it since it uses controller input instead.

klausw: There's also a middle ground where functionality may be degraded without the overlay, but it technically still works. It's up to application developers how to handle that, but my general recommendation would be to request it as optional, and consider showing a warning about missing functionality if it's not available and users may be confused otherwise.

dmarcos: Thanks, wondering if `required` should be exposed on the schema or hide it for now and default to `optional` until we see a need to expose.

src/systems/renderer.js
klausw: Done, using your suggested webxr system.

src/systems/webxr-dom-overlay.js
klausw: I've removed this, apps now need to request it as an optional/required feature themselves if they want to use it.

klausw:
What do you think about this approach? It doesn't auto-request features, leaving it up to the application to do so, but I added some reasonably user-friendly console warnings to give developers a hint what went wrong.

dmarcos:
Linter errors

src/systems/webxr.js
```diff
@@ -0,0 +1,46 @@
+var registerSystem = require('../core/system').registerSystem;
+
+/**
+ * WebXR session initialization and XR module support.
+ */
+module.exports.System = registerSystem('webxr', {
```
dmarcos: I think this could be a component

dmarcos:
Can we add some docs for the new system/component? https://github.com/aframevr/aframe/blob/master/docs/components/

dmarcos:
Thank you!

Issue #4089, opened 2019-03-16
Inspector shortcut does not work on certain keyboard layouts


LiekeVanmulken:
**Description:**
The keyboard shortcut for the inspector does not work on the United States-international keyboard layout.

- A-Frame Version: 0.9.0
- Platform / Device: chrome and firefox, the rest is untested
- Reproducible Code Snippet or URL:

This is the default code from the [introduction in the docs](https://aframe.io/docs/0.9.0/introduction/)
```
<html>
  <head>
    <script src="https://aframe.io/releases/0.9.0/aframe.min.js"></script>
  </head>
  <body>
    <a-scene>
      <a-box position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9"></a-box>
      <a-sphere position="0 1.25 -5" radius="1.25" color="#EF2D5E"></a-sphere>
      <a-cylinder position="1 0.75 -3" radius="0.5" height="1.5" color="#FFC65D"></a-cylinder>
      <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4"></a-plane>
      <a-sky color="#ECECEC"></a-sky>
    </a-scene>
  </body>
</html>
```

It looks like when i use an English keyboard layout (United States international keyboard layout), `<ctrl> + <alt> + i`  it will produce the following character (gotten from [keyboardtester](https://www.keyboardtester.com/)): 
í
 
This then prevents the inspector from opening. This is solved when using a different keyboard layout, using the Dutch keyboard layout fixes the problem.

WouterVanmulken:
It looks like that when you hold down `<ctrl> + <alt> + i` for a solid second it works as well.

WouterVanmulken:
This actually seems to be an issue with windows turning `ctrl + alt` into `AltGr` which you can read about [here](https://superuser.com/questions/592970/can-i-make-ctrlalt-not-act-like-altgr-on-windows). I tested it and when you press `ctrl+alt` it recognizes that both are pressed but the moment you press the `i` it thinks that the `ctrl` and `alt` key are no longer pressed. When you hold it longer it will eventually recognize that both are pressed and will trigger.

The following code was used to check the keydown:
<html>
<head>
    <script>
        function handler(e) {
            var key = window.event ? e.keyCode : e.which;
            console.log(key, e.altKey, e.ctrlKey);

            //Shortcut code from aframe  src\components\scene\inspector.js line 62
[...]

PR #4516, opened 2020-03-08
Fix inspector toggle on US international keyboards

svenkonings:
Fixes issue #4089 
Related to aframevr/aframe-inspector#598

On US international keyboards Ctrl + Alt are combined into AltGraph. When this happens Ctrl and Alt are registered as unpressed, breaking the toggle expression.

This is fixed by adding an additional clause for the AltGraph combination.

Currently it only works once because a version of aframe-inspector without the fix is loaded. Combined with aframevr/aframe-inspector#598 everything works as expected.

### DIFF
--- a/src/components/scene/inspector.js
+++ b/src/components/scene/inspector.js
@@ -59,7 +59,7 @@ module.exports.Component = registerComponent('inspector', {
    * <ctrl> + <alt> + i keyboard shortcut.
    */
   onKeydown: function (evt) {
-    var shortcutPressed = evt.keyCode === 73 && evt.ctrlKey && evt.altKey;
+    var shortcutPressed = evt.keyCode === 73 && (evt.ctrlKey && evt.altKey || evt.getModifierState('AltGraph'));
     if (!shortcutPressed) { return; }
     this.openInspector();
   },


src/components/scene/inspector.js
```diff
@@ -59,7 +59,7 @@ module.exports.Component = registerComponent('inspector', {
    * <ctrl> + <alt> + i keyboard shortcut.
    */
   onKeydown: function (evt) {
-    var shortcutPressed = evt.keyCode === 73 && evt.ctrlKey && evt.altKey;
+    var shortcutPressed = evt.keyCode === 73 && (evt.ctrlKey && evt.altKey || evt.getModifierState('AltGraph'));
```
dmarcos: AltGraph is deprecated as per https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState

svenkonings: Hmm I see. There doesn't seem to be another way to detect ctrl + alt on an US international keyboard. Some sources report that `crtlKey` and `altKey` should both be true but that doesn't seem to be the case on both Firefox and Chrome.

The only other solution I can think of right now is to use `evt.key === 'í'`, as í is the key returned when pressing ctrl + alt + i on a US international keyboard. This, however, seems quite layout dependent.

svenkonings: The link has been updated and AltGraph no longer seems deprecated as far as I can see.

dmarcos:
Thanks!

PR #4522, opened 2020-03-13
Fix for missing controllers in Moon Rider

cabanier:
This issue was discovered in Moon Rider after moving to WebXR. See https://github.com/supermedium/moonrider/issues/98

Because of a timing issue, when `updateController` is called by the event, the `update` in this function didn't execute yet so no controllers were visible the first time you entered an immersive session.

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -34,6 +34,7 @@ module.exports.Component = registerComponent('tracked-controls', {
         index: data.controller,
         iterateControllerProfiles: data.iterateControllerProfiles
       });
+      el.components["tracked-controls-webxr"].updateController();
     } else {
       el.setAttribute('tracked-controls-webvr', data);
     }
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -34,7 +34,6 @@ module.exports.Component = registerComponent('tracked-controls', {
         index: data.controller,
         iterateControllerProfiles: data.iterateControllerProfiles
       });
-      el.components['tracked-controls-webxr'].updateController();
     } else {
       el.setAttribute('tracked-controls-webvr', data);
     }
--- a/src/components/tracked-controls.js
+++ b/src/components/tracked-controls.js
@@ -34,7 +34,7 @@ module.exports.Component = registerComponent('tracked-controls', {
         index: data.controller,
         iterateControllerProfiles: data.iterateControllerProfiles
       });
-      el.components["tracked-controls-webxr"].updateController();
+      el.components['tracked-controls-webxr'].updateController();
     } else {
       el.setAttribute('tracked-controls-webvr', data);
     }
--- a/src/components/tracked-controls-webxr.js
+++ b/src/components/tracked-controls-webxr.js
@@ -32,6 +32,10 @@ module.exports.Component = registerComponent('tracked-controls-webxr', {
     this.axisMoveEventDetail = {axis: this.axis, changed: this.changedAxes};
   },
 
+  update: function () {
+    this.updateController();
+  },
+
   play: function () {
     var sceneEl = this.el.sceneEl;
     this.updateController();


src/components/tracked-controls.js
```diff
@@ -34,6 +34,7 @@ module.exports.Component = registerComponent('tracked-controls', {
         index: data.controller,
         iterateControllerProfiles: data.iterateControllerProfiles
       });
+      el.components["tracked-controls-webxr"].updateController();
```
Artyom17: Should use single quote

src/components/tracked-controls.js
```diff
@@ -34,6 +34,7 @@ module.exports.Component = registerComponent('tracked-controls', {
         index: data.controller,
         iterateControllerProfiles: data.iterateControllerProfiles
       });
+      el.components['tracked-controls-webxr'].updateController();
```
dmarcos: I realized we're already doing this in the `play` method of the [tracked-controls-webxr](https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls-webxr.js#L37). This should be redundant.

cabanier: I put a breakpoint in `play` and itwas never called

cabanier: @dmarcos play only called once when the application launches. At that point, there are no controllers since they are only available once the immersive session starts.

dmarcos: Thanks for testing and the patience. `play` running once before entering immersive mode is expected and the handler for `controllersupdated` should be attached. Do you see the event being fired at all? https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls-webxr.js#L40

cabanier: > Thanks for testing and the patience. `play` running once before entering immersive mode is expected 

It's called when the page loads, not when the immersive session starts. At that point, there are no controllers.
I see that the event is called when the page loads.

dmarcos: Yes, the `controllersupdate` handler is attached on the play method. I expect the event to fire after entering immersive when controllers are available triggering the `controllersUpdate` method. that doesn’t seem to be happening. I wonder why

cabanier: It is called after entering the session, but at that point `update` (which registers the controllers in aframe) hasn't run yet

dmarcos: update should be called on component initialization just before `play`

cabanier: yes, it's called (once for each hand) when the page loads and when the immersive session starts.

cabanier: Maybe we need to talk face to face :-)

dmarcos: Proper fix would be moving the call ` this.updateController` to  an `update` method in the [tracked-controls-webxr component](https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls-webxr.js)

```
play: function () {
   this.updateController();
}
```
this way if the properties change the logic to look for a matching controller will run.

dmarcos: The issue comes from the `haptics` components used in moonrider that set [tracked-controls](https://github.com/supermedium/superframe/blob/master/components/haptics/index.js#L11) early. When `oculus-touch-controls`  change the properties the updateController logic doesn't run with the new properties.

cabanier: I just tried that and it fixed the problem in Moon Rider!
Do you want me to make a PR with that change?

dmarcos: Awesome, thanks for testing! Yeah, incorporate the change on this PR and we can merge. We’re pretty close!

cabanier:
When will this go in an aframe release?

dmarcos:
Yeah! 1.1.0 but you can use master builds now:

https://cdn.jsdelivr.net/gh/aframevr/aframe@44212208beaa0f2d13c9721549687455886587c3/dist/aframe-master.min.js

Maybe we can update the moonrider PR with the build above? https://github.com/supermedium/moonrider/pull/97/files

PR #4526, opened 2020-03-17
Add webxr referenceSpaceType schema entry and documentation

klausw:
**Description:**
Support changing the scene reference space to something other than 'local-floor'

**Changes proposed:**
- add schema entry
- use the name in webxr-tracked-controller and in three.js (for the viewer pose)
- update documentation

### DIFF
--- a/docs/components/webxr.md
+++ b/docs/components/webxr.md
@@ -26,11 +26,41 @@ them.
 
 | Property                | Description                                                    | Default Value |
 |-------------------------|----------------------------------------------------------------|---------------|
+| referenceSpaceType      | The scene's reference space type for camera and controllers.   | local-floor   |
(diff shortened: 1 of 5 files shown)


dmarcos:
There's a failing test but I think is a travis issue.

klausw:
I've changed the laser-controls test to use "local" reference space to test this, here's a live version: https://klausw.github.io/aframe/examples/test/laser-controls/

I've tried it on Chrome using Daydream View and AR on Android, and SteamVR on Windows. It looks like the head and controller positions are consistent with each other with this change included.

dmarcos:
Thanks!

dmarcos:
Awesome work! Thanks!

Issue #4520, opened 2020-03-09
Look controls hmdEnabled:false seems to be ignored


andreyrd:
**Description:**
I would expect the configuration `enabled: true; hmdEnabled: false; touchEnabled: true;` to only take touch input, but gyroscope controls are still enabled. In issue #4095, the look-controls component is *entirely* disabled. But I would like touch controls to stay enabled while having the ability to toggle on & off gyroscope controls.

The source code for look-controls will reset yaw & pitch when you set `hmdEnabled` to `false`, but it will not actually do anything else with that setting, and will still read continue to read and apply data from hmd:
https://github.com/aframevr/aframe/blob/master/src/components/look-controls.js#L79

In past versions, I would rewrite look-controls and implement `hmdEnabled` myself. Last version I did this for was 0.9.2. If I'm understanding the purpose of `hmdEnabled` correctly, and this is a bug, I can open a PR for the latest version.

- A-Frame Version: 1.0.4
- Platform / Device: iOS 13.3.1 / iPhone 8, Android Chrome 80.0.3987.132 / Android 9 / motorola one zoom
- Reproducible Code Snippet or URL: https://rust-low-camp.glitch.me

dmarcos:
yeah thanks. A PR will be appreciated. `hmdEnabled` might be also misleading. I would probably rename to `magicWindowTrackingEnabled`

PR #4528, opened 2020-03-24
Implement and rename hmdEnabled to magicWindowTrackingEnabled

andreyrd:
**Description:**
This is an implementation of the fix for #4520  

**Changes proposed:**
- Rename `hmdEnabled` to `magicWindowTrackingEnabled`
- Pass through `magicWindowTrackingEnabled` setting to `magicWindowControls.enabled`
- Reset magic window eulers when tracking is disabled. Otherwise you can get stuck looking at the ground. This was also what the original code was aiming to do, since previous versions used `pitchObject` and `yawObject` for gyroscope tracking. But the previous check `oldData && !data.hmdEnabled && !oldData.hmdEnabled` should actually have been more like `oldData && !data.hmdEnabled && oldData.hmdEnabled` in order to catch the transition from enabled to disabled.

### DIFF
--- a/src/components/look-controls.js
+++ b/src/components/look-controls.js
@@ -15,7 +15,7 @@ module.exports.Component = registerComponent('look-controls', {
 
   schema: {
     enabled: {default: true},
-    hmdEnabled: {default: true},
+    magicWindowTrackingEnabled: {default: true},
     pointerLockEnabled: {default: false},
     reverseMouseDrag: {default: false},
     reverseTouchDrag: {default: false},
@@ -51,6 +51,7 @@ module.exports.Component = registerComponent('look-controls', {
 
   setupMagicWindowControls: function () {
     var magicWindowControls;
+    var data = this.data;
 
     // Only on mobile devices and only enabled if DeviceOrientation permission has been granted.
     if (utils.device.isMobile()) {
@@ -58,10 +59,10 @@ module.exports.Component = registerComponent('look-controls', {
       if (typeof DeviceOrientationEvent !== 'undefined' && DeviceOrientationEvent.requestPermission) {
         magicWindowControls.enabled = false;
         if (this.el.sceneEl.components['device-orientation-permission-ui'].permissionGranted) {
-          magicWindowControls.enabled = true;
+          magicWindowControls.enabled = data.magicWindowTrackingEnabled;
         } else {
           this.el.sceneEl.addEventListener('deviceorientationpermissiongranted', function () {
-            magicWindowControls.enabled = true;
+            magicWindowControls.enabled = data.magicWindowTrackingEnabled;
           });
         }
       }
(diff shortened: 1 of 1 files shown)


src/components/look-controls.js
```diff
@@ -76,10 +77,15 @@ module.exports.Component = registerComponent('look-controls', {
       this.updateGrabCursor(data.enabled);
     }
 
-    // Reset pitch and yaw if disabling HMD.
-    if (oldData && !data.hmdEnabled && !oldData.hmdEnabled) {
-      this.pitchObject.rotation.set(0, 0, 0);
-      this.yawObject.rotation.set(0, 0, 0);
+    // Reset magic window eulers if tracking is disabled.
+    if (oldData && !data.magicWindowTrackingEnabled && oldData.magicWindowTrackingEnabled) {
```
dmarcos: I think this should be 
```
if (oldData && !data.magicWindowTrackingEnabled && !oldData.magicWindowTrackingEnabled) 
{
...
}
```
equivalent to old code

dmarcos:
Thanks! I thinks there's a bug. I left a comment

src/components/look-controls.js
andreyrd: I think the old code might be wrong.

My intention is that it only resets when the property **changes** from disabled -> enabled.
With the old check it will only change the _second time_ that update is called with the value disabled.

So it will still work, but on the second `update` and will reset every update afterwards.

If I go back one more commit in git history of that line, it appears the original intent was closer to what I'm thinking with my change:
https://github.com/aframevr/aframe/blame/e9ba54ebcf724d828fba90fce9adc530e40526d6/src/components/look-controls.js#L45

`!hmdEnabled && oldData && hmdEnabled !== oldData.hmdEnabled` can be logically reduced to `!hmdEnabled && oldData && false !== oldData.hmdEnabled` and then to `!hmdEnabled && oldData && oldData.hmdEnabled`

Commit fe199d changed this: https://github.com/aframevr/aframe/commit/fe199da677e4fa5c5f83e3927fb8da39ca3ed766#diff-6285039eee70ed830f70f7d2f7b7dfa5R47

andreyrd: Any thoughts on this? I can change it to be equivalent with the old code, even though I think the logic is wrong. It doesn't affect my use case of it anyway, so doesn't have to be my problem to deal with.

dmarcos: Maybe we always want to reset if it value has changed?

`if (oldData && data.hmdEnabled !== oldData.hmdEnabled)`

dmarcos:
Thanks for the patience. I think we're pretty close to merge this.

src/components/look-controls.js
dmarcos: This is pretty close. What do you think about the comment above?

andreyrd: Hey, sorry, have been quite busy lately. That makes sense. I think either way when enabling, the objects will almost immediately get overwritten from motion data, but I think it makes sense to always have a clean state when toggling.

andreyrd: Actually, never mind. Just tested. And it's actually nice to _not_ reset when enabling motion controls. That way you can set your "starting point" for motion using touch.

dmarcos:
thanks!

PR #4540, opened 2020-04-07
Fix GearVR controller detection with WebXR

vincentfretin:
**Description:**

GearVR controller is not detected with WebXR on Oculus Browser 8.1.5.203450095 based on Chromium 79.0.3945.126

**Changes proposed:**
Capture d’écran de 2020-04-07 14-40-22

The issue is in src/utils/tracked-controls.js findMatchingControllerWebXR.
We need to continue iterating over controllers instead of returning if the first controller didn't match.

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -162,7 +162,7 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index, 
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     profiles = controller.profiles;
-    if (profiles.length === 0) { return; }
+    if (profiles.length === 0) { continue; }
     if (iterateProfiles) {
       for (j = 0; j < profiles.length; j++) {
         controllerMatch = profiles[j].startsWith(idPrefix);
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -162,14 +162,13 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index, 
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     profiles = controller.profiles;
-    if (profiles.length === 0) { continue; }
     if (iterateProfiles) {
       for (j = 0; j < profiles.length; j++) {
         controllerMatch = profiles[j].startsWith(idPrefix);
         if (controllerMatch) { break; }
       }
     } else {
-      controllerMatch = profiles[0].startsWith(idPrefix);
+      controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);
     }
     if (!controllerMatch) { continue; }
     // Vive controllers are assigned handedness at runtime and it might not be always available.
--- a/src/components/daydream-controls.js
+++ b/src/components/daydream-controls.js
@@ -18,6 +18,39 @@ var GAMEPAD_ID_WEBVR = 'Daydream Controller';
 var GAMEPAD_ID_PREFIX = isWebXRAvailable ? GAMEPAD_ID_WEBXR : GAMEPAD_ID_WEBVR;
 
 /**
+ * Button indices:
+ * 0 - trackpad
+ * 1 - menu (never dispatched on this layer)
+ * 2 - system (never dispatched on this layer)
+ *
+ * Axis:
+ * 0 - trackpad x
+ * 1 - trackpad y
+ */
+var INPUT_MAPPING_WEBVR = {
+  axes: {trackpad: [0, 1]},
+  buttons: ['trackpad', 'menu', 'system']
+};
+
+/**
+ * Button indices:
+ * 0 - none
+ * 1 - none
+ * 2 - touchpad
+ *
+ * Axis:
+ * 0 - touchpad x
+ * 1 - touchpad y
+ * Reference: https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/profiles/google/google-daydream.json
+ */
+var INPUT_MAPPING_WEBXR = {
+  axes: {touchpad: [0, 1]},
+  buttons: ['none', 'none', 'touchpad', 'menu', 'system']
+};
+
+var INPUT_MAPPING = isWebXRAvailable ? INPUT_MAPPING_WEBXR : INPUT_MAPPING_WEBVR;
+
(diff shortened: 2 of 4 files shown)


vincentfretin:
Don't merge this yet. I can see the controller now but none of the button works because of some errors.
Capture d’écran de 2020-04-07 15-39-39

arpu:
maybe it needs something like this https://github.com/aframevr/aframe/commit/2001e5ff2f924fd970dd1a90abd72c791f6c2e7e#diff-bc91e5db13c01778a4dfb59ce1fd3f05 ?

vincentfretin:
`iterateControllerProfiles: true` allows to check for the second string in the profiles array, this is not the issue here. The string "samsung-gearvr" is first in the array, so iterateControllerProfiles doesn't matter. The issue is that the right controller is the second controller in the controllers array and not the first. I don't know how the controllers array looks like on Oculus Go or other headsets...

vincentfretin:
So I fixed two other issues, the mapping was wrong, we only have on controller on GearVR similar to Oculus Go. And I fixed the touchpad highlight. The issue should be in Oculus Go controller as well, I did a commit for that too. If someone can test this on Oculus Go, this would be great.
This is ready to merge for me.

vincentfretin:
The first controller on the GearVR is the "gaze" controller with the touchpad on the headset. This is the only headset with a controller on the headset. :-)
I think I'm the only one still using GearVR. :-D

vincentfretin:
I rebased from master.

src/utils/tracked-controls.js
```diff
@@ -162,7 +162,7 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index,
   for (i = 0; i < controllers.length; i++) {
     controller = controllers[i];
     profiles = controller.profiles;
-    if (profiles.length === 0) { return; }
```
dmarcos, line 165: Is this needed?

dmarcos:
GearVR is being sunset. I believe the Oculus Browser will no longer receive any updates.

src/utils/tracked-controls.js
vincentfretin: The change from `return` to `continue`? Yes, this is the main fix to detect the controller, see the screenshot and my comments. The line in itself? Yes, because we use `profiles[0]` below.

vincentfretin:
Oculus browser won't receive any more updates for GearVR, indeed. :(
https://support.oculus.com/198231711449088
https://venturebeat.com/2020/04/05/facebook-kills-software-updates-for-samsung-gear-vr/

In the meantime, it still works and this is the only device I have in hands for now.

vincentfretin:
Hi @dmarcos, if you need more info or some changes on my part for this PR to be merged, let me know.
I see from the changelog that @Artyom17 @dmarcos @arpu contributed the changes for Oculus Go and Gear VR support for WebXR. Do anyone of you have a real device to test it?

What this PR fixes:
- Fix Gear VR controller detection over WebXR when the controller is second in the list of available controllers
- Fix WebXR buttons mapping for Gear VR controller
- Fix Gear VR and Oculus Go touchpad highlight when touchpad is touched over WebXR
- Add support for Daydream controller over WebXR

I can confirm all the fixes for Gear VR.
Can anyone confirm the touchpad highlight fix for Oculus Go?
Can anyone confirm the fixes for Daydream controller over WebXR? If not I can remove the commit concerning Daydream in this PR, let me know.

arpu:
Hi i can test this on the Oculus go!  can i access your branch from https build?

src/utils/tracked-controls.js
dmarcos: I'm worried about this code to break other controllers support. It needs to be tested properly in several browsers and headsets

vincentfretin:
I made two glitch:
One without the PR: 
https://aframe-without-pr4540.glitch.me/
One with the PR:
https://aframe-pr4540.glitch.me/

Target the cube with one of your controllers and press the trigger button to switch the color between white and black.

If one of you have the Quest at hand, please confirm that it is still working.

vincentfretin:
My colleague tested it with the Quest, both controllers are working properly.

arpu:
@vincentfretin  tested both on the oculus go 

without PR trigger can set color on the cube but the touch event is not shown on the controller pad
with PR  both works just fine!

looks good on Oculus go!

vincentfretin:
Thanks for testing. So we confirmed it works on Gear VR, Oculus Go and Oculus Quest.

src/utils/tracked-controls.js
dmarcos: I think there might be a browser bug. I think the profiles array should be always populated for all controllers through WebXR. I would like to avoid workarounds browser bugs.

vincentfretin: This may be true. But in this case why did you have this check in the first place? You already had this code to workaround a browser bug I think. The check before didn't try other controllers, the new check now try the other controllers. Would you like me to remove this line and just replace 
```
controllerMatch = profiles[0].startsWith(idPrefix);
```
by
```
controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);
````
? This will work too.

Even if I create an issue about the gaze controller having an empty profiles array on the Oculus browser tracker, we won't see any update of Oculus browser for Gear VR, so I think there is no point of creating it.

dmarcos: Yes, that line was introduce by this [commit](https://github.com/aframevr/aframe/commit/9a7d7f41f9bb26addb7c405180fdbaceba265e4e). Profiles were not available when Oculus Browser shipped WebXR for the first time. We should first understand if it's a bug in the Oculus Browser implementation or expected behavior.

dmarcos: The controllers detection code is complicated and workarounds for a device that has been discontinued might not be worth the risk

vincentfretin: I can understand this. So I'm ok to remove the line completely and add an additional check before calling startsWith on profiles[0] just in case profiles[0] is undefined. This is just js best practice and not a workaround. :)

vincentfretin:
@dmarcos I removed the line and did the additional check in the last commit like I said. This still works.
I updated https://aframe-pr4540.glitch.me with the new build.

dmarcos:
Thanks!

Issue #4407, opened 2020-01-17
Support WebXR `select` event for controllers.


dmarcos:
For controllers with no Gamepad attached or no available buttons like for example Cardboard v2 and AR Module screen input.

PR #4544, opened 2020-04-09
Fix selectstart/selectend listeners never registered

vincentfretin:
I just tested it in Chrome canary in AR mode. The listeners are never registered because xrSession is undefined. In my case `this.el` is some `<a-entity cursor="fuse:false" raycaster="objects: .clickable">` as a child of `<a-camera>`, not the scene. Did you mean to write ` var xrSession = this.el.sceneEl.xrSession;` @dmarcos in #4407 ? With this change the listeners are registered and the raycaster/cursor works in AR mode.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -319,7 +319,7 @@ module.exports.Component = registerComponent('cursor', {
   },
 
   onEnterVR: function () {
-    var xrSession = this.el.xrSession;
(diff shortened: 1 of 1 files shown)


vincentfretin:
@machenmusik you may be interested by this change. I saw you are doing some conversion selectstart/selectend to touchstart/touchend yourself in aframe-ar. This may not be needed and you can just use the cursor component with this fix.

Issue #4459, opened 2020-02-14
wakelock causes NSURLErrorDomain in cordova app on iOS 13+


IntraNoctem:
- A-Frame Version: 1.0.4
- Platform / Device: iOS 13+, iPhone 7

`window.location.href = '/'` used by wakelock causes NSURLErrorDomain errors in cordova applications on iOS 13+.

In a clean cordova application it produces the error code -1101 which you only see in xcode logs.

But if you use Ionic webview plugin it ends up in an error code -999 of which the user will be informed by a system message on each wakelock interval. Therefore it's not possible to use a daydream or any other similar vr headset.

In our case removing the iOS wakelock solved the problem and vr mode works fine on iOS 12 & 13, in cordova apps and webbrowsers firefox, chrome and safari. 

Maybe it can be removed anyway?

PR #4547, opened 2020-04-11
disable wakelock on newer iOS

machenmusik:
**Description:**
Old wakelock code causes intermittent failures on iOS, see immersive-web/webvr-polyfill#354 and aframevr/aframe#4459.
 
**Changes proposed:**
Don't enable wakelock if on newer iOS.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,8 @@
 // Polyfill `Promise`.
 window.Promise = window.Promise || require('promise-polyfill');
 
+var utils = require('./utils/');
+
 // WebVR polyfill
 // Check before the polyfill runs.
 window.hasNativeWebVRImplementation = !!window.navigator.getVRDisplays ||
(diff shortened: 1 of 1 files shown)


machenmusik:
Note that is it not currently possible to override AFRAME's desired polyfill configuration.

machenmusik:
I don't know, but this was causing sporadic but eventual whiteout AR crashes on WebXR Viewer, and I don't think it works on mobile Safari anymore either.

dmarcos:
The [WebXR Viewer](https://apps.apple.com/us/app/webxr-viewer/id1295998056) has not been updated since June 2019. Probably better to use Chrome as baseline for WebXR AR features

machenmusik:
Sadly, my understanding is that it's still the only functional AR browser for iOS.
The PR is conditioned on newer iOS since those are where the wakelock is broken.

dmarcos:
Not sure how much we should accommodate since it doesn’t follow the spec progress. If project gets picked back up we can look to support.

machenmusik:
My understanding is they are picking it up via webxr branch of firefox-ios.  In any event, iOS Safari also fails with the current wakelock, so it may be more important in that regard.  Are you sure you want to close?

machenmusik:
see https://github.com/mozilla-mobile/firefox-ios/tree/webxr

PR #4548, opened 2020-04-11
use webvr-polyfill version that handles missing rotationRate

machenmusik:
**Description:**
Fixes log spam due to missing rotationRate observed on newer iOS.
**Changes proposed:**
Use fixed webvr-polyfill (which in turn uses fixed cardboard-vr-display)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,17 +39,17 @@
   ],
   "dependencies": {
     "custom-event-polyfill": "^1.0.6",
-    "debug": "ngokevin/debug#noTimestamp",
+    "debug": "github:ngokevin/debug#noTimestamp",
     "deep-assign": "^2.0.0",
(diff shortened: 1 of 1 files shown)


dmarcos:
Head orientation on stereo mode via polyfill is currently broken on latest iOS 13.4.1 Does this fix the issue?

machenmusik:
I don't know, but quite possibly - was going to suggest that on the issue and reopen it

andreyrd:
This prevents the page from crashing because of the log spam, but it does not make orientation work again.

And it also relies on referencing specific branches of a forked repository..to two levels.
Ultimately it needs to be fixed way upstream in https://github.com/immersive-web/cardboard-vr-display.

dmarcos:
The issue has been fixed upstream in the webvr-polyfill repo thanks to @jsantell. Now merged in master. Thanks everybody for investigating

Issue #4494, opened 2020-02-24
Click event is triggered twice when using both cursor and laser-controls


binzume:
**Description:**

click event fires twice in WebXR when using both cursor and laser-controls.

`<a-scene cursor="rayOrigin: mouse" ...` 

Above component seems to emit a click event when the controller button is pressed.
This only happens within a WebXR session. also it worked correctly up to v1.0.3.

- A-Frame Version: >= 1.0.4
- Platform / Device: Oculus Quest
- Reproducible Code Snippet or URL: https://gist.githack.com/binzume/8f81e06fb719bd9dede70e75673cfa63/raw/6bfbb295fe8e2a7804c6a98f4104f398a755f192/aframe-click.html

PR #4566, opened 2020-05-02
Fix an issue that click event fired twice in WebXR session (#4494)

binzume:
**Description:**
When entering VR , clear intersection of cursor component for mouse to prevent unintended click events in WebXR. Because the raycaster will never be updated in VR.

**Changes proposed:**
- Fix issue https://github.com/aframevr/aframe/issues/4494

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -319,6 +319,7 @@ module.exports.Component = registerComponent('cursor', {
   },
 
   onEnterVR: function () {
+    if (this.data.rayOrigin === 'mouse') { this.clearCurrentIntersection(true); }
     var xrSession = this.el.sceneEl.xrSession;
     var self = this;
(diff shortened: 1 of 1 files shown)


src/components/cursor.js
```diff
@@ -319,6 +319,7 @@ module.exports.Component = registerComponent('cursor', {
   },
 
   onEnterVR: function () {
+    if (this.data.rayOrigin === 'mouse') { this.clearCurrentIntersection(true); }
```
dmarcos, line 322: Can we just call `this.clearCurrentIntersection(true);`. Do we need to check for `mouse`?

binzume: The point is correct. I removed the mouse check.

dmarcos:
Thanks!

PR #4567, opened 2020-05-02
Update controller property for Oculus Quest

binzume:
**Description:**

The ray direction of oculus quest controller in the WebXR session is wrong. So adjusted direction and position.
The direction within WebVR API may will be broken, but the latest Oculus Browser no longer supports the WebVR.

**Changes proposed:**
-
-
-

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -59,14 +59,14 @@ var CONTROLLER_PROPERTIES = {
   'oculus-touch-v2': {
     left: {
       modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'gen2-left.gltf',
-      rayOrigin: {origin: {x: -0.01, y: 0, z: -0.02}, direction: {x: 0, y: -0.5, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, 0),
+      rayOrigin: {origin: {x: 0, y: 0, z: 0}, direction: {x: 0, y: 0, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0.01, 0.005, 0.03),
       modelPivotRotation: new THREE.Euler(0, 0, 0)
     },
     right: {
       modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'gen2-right.gltf',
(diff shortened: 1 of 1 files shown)


dmarcos:
what do you mean by wrong?

binzume:
It seems that the controller's ray direction adjusted for earlier version of the OculusBrowser. (It actually worked correctly). However, there is now a difference from the direction of the ray on the home screen.
When using WebXR device API, the ray direction is provided by XRInputSource, so 
the ray direction(and origin) should not be adjusted.

ray in Home:
com oculus vrshell-20200503-054620

A-Frame 1.0.4:
com oculus browser-20200503-054938

dmarcos:
We should probably use ` XRInputSource.targetRaySpace` for WebXR instead of hardcoded values

src/components/oculus-touch-controls.js
```diff
@@ -59,14 +59,14 @@ var CONTROLLER_PROPERTIES = {
   'oculus-touch-v2': {
     left: {
       modelUrl: TOUCH_GEN2_CONTROLLER_MODEL_BASE_URL + 'gen2-left.gltf',
-      rayOrigin: {origin: {x: -0.01, y: 0, z: -0.02}, direction: {x: 0, y: -0.5, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, 0),
+      rayOrigin: {origin: {x: 0, y: 0, z: 0}, direction: {x: 0, y: 0, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0.01, 0.005, 0.03),
```
dmarcos, line 63: Why does the model position need to be adjusted? I thought the problem was only with the ray position.

binzume: It seems that the model is slightly out of position. I adjusted the offset so that the model position does not change when entering/leaving VR.
However it is only tested on my Oculus Quest, so I'll also consider to reset the change.

dmarcos: Better to separate changes and put the model shift in a different PR if necessaryy. I still think that for the rayOrigin we should probably use InputSource values instead of hard coding numbers.

binzume: Sorry for late, I reverted the change for model shift.
Also, it seems that XRInputSource.targetRaySpace has already been applied to the pose of controllers. https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls-webxr.js#L115 So rayOrigin and direction may not be needed if XRInputSource is available.

PR #4570, opened 2020-05-02
Map touchpad events to trackpad events on Oculus Go

andygill:
- I've tested this on Oculus browser and Firefox, which covers both with and without WebXR.

**Description:**
Maps all touchpad events to trackpad events when using WebXR.

### DIFF
--- a/docs/components/oculus-go-controls.md
+++ b/docs/components/oculus-go-controls.md
@@ -42,15 +42,26 @@ and/or pressed buttons (trackpad, trigger).
 
 | Event Name         | Description           |
 | ----------         | -----------           |
-| trackpadchanged    | Trackpad changed.     |
-| trackpaddown       | Trackpad pressed.     |
-| trackpadup         | Trackpad released.    |
-| trackpadtouchstart | Trackpad touched.     |
-| trackpadtouchend   | Trackpad not touched. |
-| trackpadmoved      | Trackpad moved.       |
+| touchpadchanged    | Touchpad changed.     |
+| touchpaddown       | Touchpad pressed.     |
+| touchpadup         | Touchpad released.    |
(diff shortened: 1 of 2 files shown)


dmarcos:
It's probably simpler to just update the docs to use `touchpadXXX` for events that matches the [WebXR nomenclature](https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/assets/profiles) . Only Firefox remains with a WebVR implementation and it will be deprecated soon when WebXR ships. All event names will use `touchpadXXX`. In the meantime in your code you can listen two both events using the same handler:

```
el.addEventListener('touchpadchanged', yourHandler);
el.addEventListener('trackpadchanged', yourHandler);
```

docs/components/oculus-go-controls.md
```diff
@@ -42,16 +42,27 @@ and/or pressed buttons (trackpad, trigger).
...
+| touchpadtouchstart | Touchpad touched.     |
+| touchpadtouchend   | Touchpad not touched. |
+| touchpadmoved      | Touchpad moved.       |
 | triggerchanged     | Trigger changed.      |
 | triggerdown        | Trigger pressed.      |
 | triggerup          | Trigger released.     |
 
+### Legacy WebVR Browsers
+
+Legacy (WebVR-based) Oculus Go browsers use the older 'trackpadXXXX' event names, rather than the
```
dmarcos, line 57: We can say:
```Legacy WebVR browsers use the older 'trackpadXXXX' event names```

It's not exclusive to Oculus Go. It also applies to Firefox desktop for instance.

andygill: I'll fix and push a new pull request. (I did not realize this pull make it to the aframe github; I intended it to be doc changes only.)

PR #4579, opened 2020-05-07
Asset property type needs to parse attrValue when updating

paolis92:
Only in components with single-property schema, asset is not parsed when updating (for example gltf-model)
So #asset-id will not be converted in url necessary for this.data.

Maybe instead of doing this in propertyTypes we should add the possibility to force a parse, like:
`function registerPropertyType (type, defaultValue, parse, stringify, parseRequired)`
so it becomes
`registerPropertyType('model', '', assetParse, undefined, true);`
and then in the code commited above instead of checking the parse function name, we check
`if (this.schema.parseRequired)`

**Description:**

**Changes proposed:**
- add parse of attrValue for singleProperty schema for type: asset when updating component
-
-

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -356,6 +356,11 @@ Component.prototype = {
       if (this.isObjectBased) {
         parseProperty(attrValue, this.schema);
       }
+      // Assets need parsing otherwise #asset-id will not be converted in url for this.data (ex gltf-model)
+      if (this.schema.parse.name === 'assetParse') {
+        this.data = parseProperty(attrValue, this.schema);
+        return;
+      }
       // Single-property (already parsed).
       this.data = attrValue;
       return;


src/core/component.js
```diff
@@ -356,6 +356,11 @@ Component.prototype = {
       if (this.isObjectBased) {
         parseProperty(attrValue, this.schema);
       }
+      // Assets need parsing otherwise #asset-id will not be converted in url for this.data (ex gltf-model)
+      if (this.schema.parse.name === 'assetParse') {
```
dmarcos, line 360: Adding an exception for a specific parser is brittle. We would need to find a more robust way.

paolis92: Maybe we can always call parse function like it happens for components that have schema with more than one property (am I wrong here?) a few lines below the highlighted ones
`parseProperties(attrValue, this.schema, true, this.name);`

vincentfretin:
@paolis92 That PR is really old, and I see your issue is closed, so I guess you can close that PR?

Issue #4496, opened 2020-02-24
mousedown event fires twice


dirkk0:
**Description:**
On Oculus Quest, the mousedown event seems to fire twice.
- A-Frame Version:
1.0.4
- Platform / Device:
Oculus Quest
- Reproducible Code Snippet or URL:
https://curious-electric.com//w/experiments/aframe/minimal-quest-event-bug/

Possibly related to #4494 and/or #3297.

dmarcos:
Better to use glitch to be able to look and modify code and suggest solutions. https://glitch.com/~aframe 
Also adding all the info in the issue will make for an easier discussion.

dirkk0:
Of course you are right on both points, sorry for that.

Description:
On Oculus Quest, the mousedown event seems to fire twice.
I think this is somehow related to https://github.com/aframevr/aframe/issues/3297 and  https://github.com/aframevr/aframe/issues/4141

Before entering VR the button cube changes the state, but after that it doesn't since the function is executed twice.

A workaround seems to be to delay another execution like so:
```
if (this.waiting) {
  console.log("sorry, no ...");
  return;
}
          this.waiting = true;
setTimeout(() => {
  this.waiting = false;
  console.log("reset ...");
}, 100);

```

Glitch is here:
https://husky-insidious-brother.glitch.me/

binzume:
The reason mousedown fires twice is that both `generic-tracked-controller-controls` and `oculus-touch-controls` emit events at the same time.

I have fixed some controller issues using A-Frame with Oculus Quest.
I put A-Frame with the workaround in the following URL. If possible, could you make sure your issue is fixed? @dirkk0 https://binzume.github.io/aframe/dist/aframe-master.min.js

@dmarcos The cause is clear, but I'm not sure if the fix makes sense just by looking at the code... Are you worried about the above commit?

dmarcos:
@binzume Thanks for investigating. Can you open a PR?

dirkk0:
@binzume Yes, I can confirm that this works! Thanks!

NomisGnos:
I am encountering this issue as well.

PR #4580, opened 2020-05-07
Fix confliction of generic-tracked-controller-controls and other tracked-controllers (#4496)

binzume:
**Description:**

The mousedown event fires twice because `generic-tracked-controller-controls` and another tracked controls are enabled at the same time.
Since `component.addEventListeners()` is called in `checkControllerPresentAndSetup()`, check if other `tracked-controls` exists before calling.

**Changes proposed:**
- issue https://github.com/aframevr/aframe/issues/4496


dmarcos:
A lot has changed. I think this is no longer an issue?

dmarcos:
I think this is no longer an issue but can reopen if necessary. Thanks.

Issue #4499, opened 2020-02-26
Clicking and dragging in Firefox on Windows locks view to -y direction


AdaRoseCannon:
To reproduce open the scene in Firefox on Windows, click and drag on the AFrame scene, view gets locked to -y and rotates around that point.

Expected behaviour: look around with click and drag.

- A-Frame Version: Tested on all demos on the homepage
- Platform / Device: Windows, Firefox
- Reproducible Code Snippet or URL: https://aframe.io/examples/showcase/helloworld/

dmarcos:
I can't reproduce the described problem on any of the [examples](https://aframe.io/aframe/examples/). Do you have any installed extension that might be interfering with the input?

AdaRoseCannon:
Yeah it's only Firefox, it doesn't seem to be an extension, since it still breaks even in a private window, where all extensions are disabled.

I just inspected the camera rotation and it has valid values, no NaNs, they are just -PI/2 in the x axis, and it spin wildly in the y axis, making me think that some how the input is magnified but I can't place it.

dmarcos:
Mmmh I cannot get to spinning wildly as
described. Do you have more concrete steps to reproduce? What version of Firefox are you using?

AdaRoseCannon:
It's Firefox 73.0.1 (64-bit), 

It really is hard to put more steps in, for me any scene, at any point if I click and drag (in any direction) it will immediately lock onto looking at the floor, any further dragging rotating rapidly around the y axis.

AdaRoseCannon:
Perhaps leave the issue open for now in case anyone else can reproduce, I wish I could work out the distinguishing factor.

dmarcos:
For the record. Screen capture clicking / dragging on the [Hello World Example](https://aframe.io/aframe/examples/boilerplate/hello-world/) on Firefox 73.0.1 (64-bit) and Windows.

https://share.getcloudapp.com/Qwu7K8d6

It works as expected for me.

AdaRoseCannon:
Glad it's working for you, hopefully I am the only one it's weird for. It's definitely synced between both my computers.

dmarcos:
Is this still an issue?

AdaRoseCannon:
Just checked it on the demos on the main website, it is still an issue.

dmarcos:
[...]

PR #4599, opened 2020-05-26
fix screenX screenY

### LINKED ISSUE #3778: Camera spins down on Firefox
Hello all, when I go to the A-Frame site or any glitch project, if I try dragging my mouse on the scene to look around the camera starts spinning and points downwards. If I stop dragging the mouse it stays pointed down. If I try to readjust the camera view it starts spinning again.

I have tried disabling Tracking Protection, Privacy Badger, and HTTPS Everywhere, as those sometimes give me issues on other sites, but it doesn't seem to change anything here. 

It doesn't seem to happen on other WebVR things (such as [React's 360 here](https://github.com/facebook/react-360)). It also seems fine on Edge.

Let me know if there is anything else I can provide to help.

- A-Frame Version: Looking at the source for some of the A-Frame examples: 0.5.0, 0.7.1, and 0.8.2
- Platform / Device: Kubuntu 18.04 and Windows 10 on Firefox 62.0.2

AdaRoseCannon:
Fixes #3778 #4499 

**Changes proposed:**
- Instead of storing reference to previous event stores the values needed from that event
- Unifies mixed use of `evt` and `event` to use `event`


src/index.js
```diff
@@ -90,7 +90,7 @@ require('./core/a-mixin');
 require('./extras/components/');
 require('./extras/primitives/');
 
-console.log('A-Frame Version: 1.0.4 (Date 2020-05-13, Commit #ef405451)');
```
dmarcos, line 93: Can we remove this from PR?

src/components/look-controls.js
```diff
@@ -336,36 +339,36 @@ module.exports.Component = registerComponent('look-controls', {
...
+      x: event.touches[0].pageX,
+      y: event.touches[0].pageY
     };
     this.touchStarted = true;
   },
 
   /**
    * Translate touch move to Y-axis rotation.
    */
-  onTouchMove: function (evt) {
```
dmarcos, line 353: Can we keep `evt` instead of `event` for consistency with rest of the code base?

src/components/look-controls.js
```diff
@@ -290,16 +292,17 @@ module.exports.Component = registerComponent('look-controls', {
...
     var sceneEl = this.el.sceneEl;
     if (!this.data.enabled || (sceneEl.is('vr-mode') && sceneEl.checkHeadsetConnected())) { return; }
     // Handle only primary button.
-    if (evt.button !== 0) { return; }
+    if (event.button !== 0) { return; }
 
     var canvasEl = sceneEl && sceneEl.canvas;
 
     this.mouseDown = true;
-    this.previousMouseEvent = evt;
```
dmarcos, line 302: This makes some sense. We're holding a reference to the event object that somehow gets deallocated or values get overwritten by the browser. I still don't understand why it does only happen in some Firefox instances / computers. Wondering if there's a deeper issue that could affect other parts of the code base.

dmarcos:
Thanks for doing this. Super appreciated 👍

src/components/look-controls.js
AdaRoseCannon: A cursory search for `= evt;` reveals the same thing may occur in *src/components/cursor.js* but it's not an issue I have noticed.

AdaRoseCannon: Correction, in *src/components/cursor.js* the event is not kept beyond the function therefore won't present an issue.

dmarcos:
Thank you. Much appreciated.

PR #4632, opened 2020-06-28
Typo in number property type description.

lordbron:
There was, what seems to be, an extra apostrophe after the float number. That may confuse newbs (like me). LOL

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thank you and congrats on your first contribution! 🏅

PR #4640, opened 2020-07-02
Add support for WebXR layers (WIP)

dmarcos:
Work in progress support for WebXR layers.

This PR requires an Oculus Browser build with layers support with API enabled via `chrome://flags`

To test run `npm install` since it relies on additional THREE.js required changes.


dmarcos:
@cabanier I'm starting off your code at https://cabanier.github.io/WebXRLayers-samples/equirect-video.html

I realized you are using a `webgl` 2 context. Is there also a way to copy the texture to the layer using webgl 1.1? I tried to use `glCopyTexImage2D` on the layer texture with no success. I'm getting `ERROR :GL_INVALID_OPERATION : glCopyTexImage2D: texture is immutable`

cabanier:
I will take a look and get back to you.

> Thanks so much for the work. Having layers is going to be awesome.

Thank you! We're excited as well :-)

dmarcos:
@cabanier Thanks.

> I will take a look and get back to you

If you have WebGL 1.1 sample code I can look it will works as well.

cabanier:
Try https://cabanier.github.io/WebXRLayers-samples/equirect-video-2.html
I replaced `texSubImage2D` with `texImage2D`.

dmarcos:
@cabanier Thanks! I see `gl = canvas.getContext('webgl2', { xrCompatible: true, webgl2: false });` in `equirect-video-2.html`. Am I looking into the right file?

cabanier:
Oops. Let me fix thiat.

cabanier:
@dmarcos I think you found a bug in our implementation. I will discuss this with @Artyom17 and we'll fix it in an upcoming release.
Until then, you could draw the texture the hard way by setting the layer texture as framebuffer targets and draw vertices filled with the image.

dmarcos:
@cabanier I got your [WegGL 1 layer example](https://github.com/cabanier/WebXRLayers-samples/blob/master/equirect-video-2.html#L333) working on Quest like charm. I tried to adapt and [render a quad on the layer texture](https://github.com/dmarcos/WebXRLayers-samples/blob/drawQuad/equirect-video-2.html#L464) but couldn't make it work. My logic is not rendering on the layer and clearing the canvas framebuffer instead (see scene all red). I might missing something. Let me know if you can see anything obvious. I'm expecting seeing something [like this](https://glitch.com/edit/#!/hello-world-webgl?path=hello-world-webgl.js%3A26%3A0) rendered to the layer

cabanier:
> My logic is not rendering on the layer and clearing the canvas framebuffer instead (see scene all red). I might missing something. Let me know if you can see anything obvious. 

Yes, that should be working. Let me take a look and if I don't see anything obvious, Artem and I will investigate.

Artyom17:
@dmarcos Check that viewport is set correctly. You probably want to use glLayer.viewport instead of what you use there. Also, check that another layer still uses copyTexImage2D.

cabanier:
@dmarcos https://github.com/dmarcos/WebXRLayers-samples/pull/1 fixes the issue.
You forgot to remove the clearcolor :-)

dmarcos:
This is an old WIP PR. We already have A-Frame layers support

PR #4641, opened 2020-07-06
Fix Documentation - destroy method out of place

da4throux:
Just moved the destroy method description were it should, as it was in the middle of the setAttribute method description

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Good catch! Thanks! Congrats on your first contribution 🥇

PR #4646, opened 2020-07-17
Replace a-animation with animation component

danbuckland:
**Description:**
Updates the Minecraft Demo documentation to use the `animation` component instead of the deprecated `<a-animation>` entity.

**Changes proposed:**
- Replace the non-working `<a-animation>` entity in two code snippets with the `animation` component and equivalent properties


dmarcos:
Thanks and congrats on your first contribution 🏅

PR #4656, opened 2020-07-28
Update the TEXT component shader to conform GLSL 3

NikolaySuslov:
**Description:**

The latest update of A-Frame to THREE r118 uses 'webgl2' context by default. That breaks the TEXT component, as it's shader should be updated to GLSL 3  #version 300 es.

### DIFF
--- a/src/shaders/msdf.js
+++ b/src/shaders/msdf.js
@@ -16,11 +16,12 @@ module.exports.Shader = registerShader('msdf', {
   raw: true,
 
   vertexShader: [
-    'attribute vec2 uv;',
-    'attribute vec3 position;',
+    '#version 300 es',
+    'in vec2 uv;',
+    'in vec3 position;',
     'uniform mat4 projectionMatrix;',
     'uniform mat4 modelViewMatrix;',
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks. Wondering if we should set THREE back to webgl 1 as default in A-Frame. WebGL 2 might break other things and 3rd party components

NikolaySuslov:
Yes, there is a special THREE.WebGL1Renderer object in updated THREE r118. So, in module AScene setupRenderer: we could change it to use THREE.WebGL1Renderer instead of THREE.WebGLRenderer (with webgl2 fallback).
 ``` renderer = this.renderer = new THREE.WebGL1Renderer(rendererConfig); ```
What is about to allow choosing the default renderer somewhere during initialisation of a-scene?

dmarcos:
Not sure about having the rendering customizable. It will fragment the component community. I would either stay on WebGL 1 or move everyone to WebGL 2 for next version. It would draw a more clear line.

dmarcos:
Thanks. Let's go with WebGL 2 and see how it goes :)

PR #4665, opened 2020-08-16
Update rotation.md - indicate order of rotation

da4throux:
**Description:**
Add mention of the 'order' attribute for rotation, as it can have a strong impact since rotations are not commutative.

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/components/rotation.md
+++ b/docs/components/rotation.md
@@ -25,11 +25,12 @@ A-Frame uses a right-handed coordinate system. When aligning our right hand's
 thumb with a positive axis, our hand will curl in the positive direction of
 rotation.
 
-| Axis | Description                       | Default Value
-|------|-----------------------------------|---------------|
-| x    | Pitch, rotation about the X-axis. | 0             |
-| y    | Yaw, rotation about the Y-axis.   | 0             |
-| z    | Roll, rotation about the Z-axis.  | 0             |
+| Axis  | Description                           | Default Value
+|-------|---------------------------------------|---------------|
+| x     | Pitch, rotation about the X-axis.     | 0             |
+| y     | Yaw, rotation about the Y-axis.       | 0             |
+| z     | Roll, rotation about the Z-axis.      | 0             |
+| order | Order that the rotations are applied. | 'XYZ'         |
 
 ## Relative Rotation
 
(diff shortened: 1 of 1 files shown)


da4throux:
I'm all confused about this, I noticed:
- the default rotation.order for the a-scene is: 'XYZ'
- but actually for elements created in it it's: 'YXZ'

docs/components/rotation.md
```diff
@@ -25,11 +25,12 @@ A-Frame uses a right-handed coordinate system. When aligning our right hand's
...
-|------|-----------------------------------|---------------|
-| x    | Pitch, rotation about the X-axis. | 0             |
-| y    | Yaw, rotation about the Y-axis.   | 0             |
-| z    | Roll, rotation about the Z-axis.  | 0             |
+| Axis  | Description                           | Default Value
+|-------|---------------------------------------|---------------|
+| x     | Pitch, rotation about the X-axis.     | 0             |
+| y     | Yaw, rotation about the Y-axis.       | 0             |
+| z     | Roll, rotation about the Z-axis.      | 0             |
+| order | Order that the rotations are applied. | 'XYZ'         |
```
vincentfretin, line 33: XYZ order is the default in threejs (object3D.rotation is an Euler instance and the default is defined in https://github.com/mrdoob/three.js/blob/6e897f9a42d615403dfa812b45663149f2d2db3e/src/math/Euler.js#L321), but the default order if you use the rotation component is actually YXZ set here https://github.com/aframevr/aframe/blob/837023cca1a04b408b4acde4c09823e09dc58e6d/src/components/rotation.js#L14
The order is not configrable via the rotation component, so you can't add it to the table in the documentation like that.
You can change the order by setting `this.el.object3D.rotation.order = "XYZ"` though and setting directly in radians on `this.el.object3D.rotation.set(Math.PI/2, 0, 0)` for example, but be careful to not use again `this.el.setAttribute('rotation', "90 0 0")` because it will change again the order to YXZ.

docs/components/rotation.md
```diff
@@ -91,5 +92,5 @@ See also [reading position and rotation of the camera](./camera.md#reading-posit
 Transformations are applied to entities in this order:
 
 * [scale](scale.md)
-* rotation
+* rotation (in the order described by the attribute `order` (default: Pitch then Yaw then Roll)
```
vincentfretin, line 95: As I explained above, the default is actually YXZ, so Yaw Pitch Roll.

Issue #2837, opened 2017-07-03
setAttribute on text geometry not working


vrimaginar:
When trying to set setAttribute('geometry',{primitive: "plane", height: "auto", width: "auto"}); on `<a-entity text>` the plane is not being drawn. However if you change the wrap count programaticaly afterwards the plane is showing. The plane will not update it's width and height if you change the wrap count again.

- A-Frame Version: 0.6.0 master (8d60730)
- Platform / Device: PC

dmarcos:
Can you provide a link to an example that reproduces the issue?

dmarcos:
A fix for `setAttribute` has landed on master (https://github.com/aframevr/aframe/pull/2840). I cannot reproduce this bug on the text example.  Please let us know if you are still affected and can provide an example to reproduce

vrimaginar:
Hi @dmarcos, we still have the problem;

`var text = document.querySelector('#af-text');
text.setAttribute('geometry',{primitive: "plane", height: "auto", width: "auto"});`

Will not work, but when we change the e.g. wrapCount the textbox will appear.

dmarcos:
@vrimaginar are you trying with the master branch?

machenmusik:
IiRC, the support for auto width and height is on text component, not geometry component...

vrimaginar:
@dmarcos, yes I am on the master branche as per your suggestion.

@machenmusik, the auto width and height are on the geometry component see:

https://aframe.io/docs/0.6.0/components/text.html

`<a-entity
  geometry="primitive: plane; height: auto; width: auto"
  material="color: blue"
  text="width: 4; value: This text will be 4 units wide."></a-entity>`

The thing is, after creating the entity component with:

`<a-entity text="width: 4; color: #000; value: test text" id-text></a-entity>`

and adding `geometry="primitive: plane; height: auto; width: auto;" material="color: red"`
in the console nothing happens. But when you change anything in the text="" section, the plane appears. Even doing:

`text = document.querySelector('#text);
text.setAttribute('text','bla',1);`

will work.

dmarcos:
That's an error in the documentation. Geometry does not have auto height and width properties. I tested on the [text example](https://a-frobot.github.io/aframe/examples/test/text/) and it works fine to me (see the fucsia plane in the middle) by doing:
```javascript
el.setAttribute('geometry', 'primitive: plane; height: 1; width: .5')
```
[...]

PR #4670, opened 2020-08-25
Documentation: Text Component - replace auto reference by 0 for geometry auto-scaling

da4throux:
as per https://github.com/aframevr/aframe/issues/2837 'auto' is not acceptable value for geometry width/height.
For the time being '0' can be used for having the auto-scaling behavior.

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
This is already reflected in the docs. Thanks

PR #4675, opened 2020-09-06
Detect oculus controllers when using SteamVR for WebXR

zach-capalbo:
**Description:**

On Chrome WebXR with SteamVR runtime, Oculus touch controllers are reported with a different ID than `oculus-touch`

**Changes proposed:**
- Check for both Oculus WebXR and SteamVR WebXR gamepad id
- Keep track of the WebXR gamepad id used

**Possible Alternative Fix:**

A good alternative would be to let the gamepad ids be regular expressions. This would involve considerable refactoring, however.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -12,6 +12,7 @@ var isWebXRAvailable = require('../utils/').device.isWebXRAvailable;
 
 var GAMEPAD_ID_WEBXR = 'oculus-touch';
 var GAMEPAD_ID_WEBVR = 'Oculus Touch';
+var GAMEPAD_ID_STEAMVR = 'oculus-oculus-rift-s';
 
 // Prefix for Gen1 and Gen2 Oculus Touch Controllers.
 var GAMEPAD_ID_PREFIX = isWebXRAvailable ? GAMEPAD_ID_WEBXR : GAMEPAD_ID_WEBVR;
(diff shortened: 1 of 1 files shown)


dmarcos:
This looks like a Chrome bug. To my understanding, controller ids should match the [webxr profiles](https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/profiles/oculus/oculus-touch.json#L2) regardless of the runtime used. It looks like the profileId for Rift-S controllers over SteamVR is `oculus-oculus-rift-s` instead of `oculus-touch` / `oculus-touch-v2` @toji @klausw is this a known issue? Thanks!

toji:
For SteamVR we're using the following code to automatically produce the string with OpenVR: https://source.chromium.org/chromium/chromium/src/+/master:device/vr/openvr/openvr_gamepad_helper.cc;l=206;drc=36c0310725bd6a03de6c11dae1d4986e7e64bb18;bpv=1;bpt=1

What that means is that OpenVR is reporting a manufacturer of "oculus" and a model of "oculus-rift-s", which is a little weird. We can map it, but we're actually in the process of removing OpenVR support from Chrome in favor of OpenXR, which will support WMR, SteamVR, and Oculus devices. That code path DOES have a proper [device profile mapping system](https://source.chromium.org/chromium/chromium/src/+/master:device/vr/openxr/openxr_interaction_profiles.h;l=99) so it should work better overall. The biggest downside at the moment is that OpenXR's interaction profile system doesn't differentiate between the Gen 1 and Gen 2 Oculus Touch, so we end up reporting the 1st gen model all the time. This is better than falling back to a generic profile all the time, to be sure, but still not ideal.

dmarcos:
@toji Thanks for the info. Super appreciated! Is there any available timeline for the transition to OpenXR and when we can expect to launch WebXR content without [additional Chrome flags](https://immersiveweb.dev/chrome-support.html)?

toji:
At this point the release timelines are in Valve and Oculus' court. Once their OpenXR runtimes with the appropriate extension are released it should "just work". You can grab the SteamVR beta channel today and it'll work with Chrome 86+. Oculus has released an OpenXR runtime, but they need to release support for the appcontainer extension before Chrome will start to use it.

dmarcos:
@toji Thanks for the info

@zach-capalbo Can you give SteamVR beta channel and Chrome m86+ a try? Everything should work as expected with A-Frame 1.0.4 and RiftS

zach-capalbo:
@dmarcos, thanks for triaging this! SteamVR beta + Chrome 86+ does indeed detect the oculus touch controllers through OpenXR. However, the controllers are backwards and upside down...

In any case, I'll close this PR. If anyone finds it and needs a workaround until OpenXR support is released all around, I've packaged these changes up into a monkey-patch component here: https://gitlab.com/zach-geek/vartiste/-/blob/d626dd3be88b7cc6e9bd17d9f902720c1151acd7/src/fix-oculus-steamvr.js

toji:
My understanding is that the controller orientation problem is a known issue on Oculus' part with their OpenXR implementation.

dmarcos:
@zach-capalbo Thanks for reporting and testing this.

@toji The orientation: Is it an Oculus / OpenXR implementation bug or a Valve / SteamVR / Oculus integration issue? My understanding is that SteamVR uses its OpenXR implementation? Not my area of expertise.

toji:
A bug in Oculus' OpenXR implementation, as far as I can tell. Valve and Microsoft's implementations both line up the controllers well, and Chrome is using the same code to convey the poses across the board.

dmarcos:
@toji Thanks. Do you know if SteamVR talks to the Oculus OpenXR runtime underneath? I thought that the Oculus OpenXR runtime was only available on [beta channel](https://immersiveweb.dev/chrome-support.html).

PR #4678, opened 2020-09-12
Add reference to google hosted draco decoder URL

kfarr:
In gltf-model doc add a reference to hosted draco decoder libraries from google

**Description:**

As a new user it's annoying to include these draco decoder files in my project and I wouldn't have done so had I known this value accepted remote locations and there exists a Google hosted CDN version.

**Changes proposed:**
-
-
-


dmarcos:
Thank you!

PR #4686, opened 2020-09-22
[docs] Add Oculus Quest 2 to Headsets list

vikohone:
**Description:**
The new Oculus Quest 2 isn't on the [What Are Some Current Headsets?](https://aframe.io/docs/1.0.0/introduction/vr-headsets-and-webvr-browsers.html#what-are-some-current-headsets) list.

**Changes proposed:**
- Add the Oculus Quest 2 to the list
- Since [Oculus Quest is no longer available](https://www.oculus.com/quest/), should we remove it from the list? 🤔


dmarcos:
Thanks. We should probably just mention `Oculus headsets` and link to the site instead of individual models. Info won't get outdated

vikohone:
@dmarcos Thanks, that's a great suggestion! I've merged the links, does it look good?
https://github.com/aframevr/aframe/pull/4686/commits/022ed61fc3cadd309bbf128ea1941d1c7b4c6b7f

dmarcos:
Thank you!

PR #4691, opened 2020-09-30
Improve pinching accuracy (fixes #4689)

### LINKED ISSUE #4689: Pinching with hand-tracking-controls is inaccurate
**Description:**
Currently, the pinching gesture recognized by `hand-tracking-controls` is inaccurate for me. `pinchstarted` rarely fires as expected and when releasing an object that follows the pinch gesture, it unexpectedly follows the index finger (as it is separating from the thumb) until the gesture is ended.

I suggest two changes to the existing code:
- Increase the minimum distance between thumb and index finger needed to recognize a pinch (I found that increasing it from `0.01` meters to `0.02` meters significantly increases the frequency at which the gesture is recognized, probably partly because the finger tip positions are in the center of the fingers (and further apart than the pads of the fingers). This could also be a configurable property of the component.
- Use the position of the midpoint between the index fingertip and the thumb fingertip for pinch events so objects following the pinch position do not incorrectly move with the index finger as the pinch is released

If these sound good, I can submit a PR.

- A-Frame Version: master
- Platform / Device: Quest
- Reproducible Code Snippet or URL: https://bristle-curly-primula.glitch.me/ Both behaviors are exhibited here

cesmoak:
**Description:** As discussed in #4689, this PR improves pinching accuracy by 1) adjusting the distance between thumb and index finger needed to start the gesture, and by 2) moving the position pinch events report to the halfway point between the thumb and index finger.

**Changes proposed:**
- The min distance between thumb and index fingertips to start a pinch gesture is changed to be `0.015`.
- Two new properties are exposed on `hand-tracking-controls`: `startPinchDistance`, which defaults to `0.015` and `endPinchDistance`, which defaults to `0.03`.
- The `position` property for `pinchstarted`, `pinchmoved`, and `pinchended` events is moved to the halfway point between thumb and index fingertips.
- Checks are added to tests to sanity-check the pinch event position.

### DIFF
--- a/docs/components/hand-tracking-controls.md
+++ b/docs/components/hand-tracking-controls.md
@@ -25,7 +25,8 @@ Use `hand-tracking-controls` to integrate [hand tracked input][webxrhandinput] i
 | hand                 | The hand that will be tracked (i.e., right, left). | left                 |
 | modelColor          | Color of hand material.                                                                | white         |
 | modelStyle           | Mesh representing the hand or dots matching the joints        | mesh
-
+| startPinchDistance    | Maximum distance between tip of thumb and index finger before a pinch gesture is started. | 0.015 |
+| endPinchDistance    | Minimum distance between tip of thumb and index finger before a pinch gesture is ended. | 0.03 |
 
 ## Events
 
--- a/docs/components/hand-tracking-controls.md
+++ b/docs/components/hand-tracking-controls.md
@@ -25,8 +25,6 @@ Use `hand-tracking-controls` to integrate [hand tracked input][webxrhandinput] i
 | hand                 | The hand that will be tracked (i.e., right, left). | left                 |
 | modelColor          | Color of hand material.                                                                | white         |
 | modelStyle           | Mesh representing the hand or dots matching the joints        | mesh
-| startPinchDistance    | Maximum distance between tip of thumb and index finger before a pinch gesture is started. | 0.015 |
-| endPinchDistance    | Minimum distance between tip of thumb and index finger before a pinch gesture is ended. | 0.03 |
 
 ## Events
 
--- a/src/components/hand-tracking-controls.js
+++ b/src/components/hand-tracking-controls.js
@@ -221,20 +221,20 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
 
       if (distance < this.data.startPinchDistance && this.isPinched === false) {
         this.isPinched = true;
-        this.pinchEventDetail.position.copy(indexTipPose.transform.position);
+        this.pinchEventDetail.position.copy(indexTipPosition).lerp(thumbTipPosition, 0.5);
         this.pinchEventDetail.position.y += 1.5;
         this.el.emit('pinchstarted', this.pinchEventDetail);
       }
 
       if (distance > this.data.endPinchDistance && this.isPinched === true) {
         this.isPinched = false;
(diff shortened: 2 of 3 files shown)


docs/components/hand-tracking-controls.md
```diff
@@ -25,7 +25,8 @@ Use `hand-tracking-controls` to integrate [hand tracked input][webxrhandinput] i
 | hand                 | The hand that will be tracked (i.e., right, left). | left                 |
 | modelColor          | Color of hand material.                                                                | white         |
 | modelStyle           | Mesh representing the hand or dots matching the joints        | mesh
-
+| startPinchDistance    | Maximum distance between tip of thumb and index finger before a pinch gesture is started. | 0.015 |
```
dmarcos, line 28: Should we make it configurable? Is not a value that clearly works better for most cases?

tests/components/hand-tracking-controls.test.js
```diff
@@ -8,6 +8,8 @@ suite('tracked-controls-webxr', function () {
   var standingMatrix = new THREE.Matrix4();
   var index = {transform: {position: {x: 0, y: 0, z: 0}}};
   var thumb = {transform: {position: {x: 0, y: 0, z: 0}}};
+  var indexPosition = new THREE.Vector3();
+  var thumbPosition = new THREE.Vector3();
```
dmarcos, line 12: Thanks for taking care of the tests 🏅

docs/components/hand-tracking-controls.md
cesmoak: This is the best value I've found from my personal tests--happy to hardcode for that. The even better approach would use the `radius` property of the joint poses, but maybe best to wait until we have more than one example piece of hardware to determine the best number using that. If you confirm, I'll remove both configurable properties.

cesmoak: The latest commit removes the properties.

src/components/hand-tracking-controls.js
```diff
@@ -217,22 +217,22 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
 
       var distance = indexTipPosition.distanceTo(thumbTipPosition);
 
-      if (distance < 0.01 && this.isPinched === false) {
+      if (distance < 0.015 && this.isPinched === false) {
```
dmarcos, line 220: We can maybe move this to a variable at the top `var PINCH_DISTANCE = 0.015` . We use [similar pattern in other components](https://github.com/aframevr/aframe/blob/master/src/components/oculus-touch-controls.js#L13)

src/components/hand-tracking-controls.js
```diff
@@ -217,22 +217,22 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
...
         this.isPinched = true;
-        this.pinchEventDetail.position.copy(indexTipPose.transform.position);
+        this.pinchEventDetail.position.copy(indexTipPosition).lerp(thumbTipPosition, 0.5);
```
dmarcos, line 222: What is the 0.5 value for? Maybe define another variable on top?

src/components/hand-tracking-controls.js
cesmoak: Addressed in the latest commit

dmarcos:
Thanks! Congrats on your first PR 🥳

PR #4696, opened 2020-10-04
Example Added

aayushigupt:
**Description:**

**Changes proposed:**
-
-
-


dmarcos:
This kind of scene / content is better suited to the [community channels](https://aframe.io/community/) and not as a bult-in A-Frame example.

PR #4698, opened 2020-10-06
Improve pinching for moving objects (fixes #4690)

### LINKED ISSUE #4690: Add properties to hand-tracking-controls pinch events to be able to use pinch gesture to move objects
**Description:** When using pinching to move objects, knowing which hand is pinching and what the orientation of the hand is in is useful so that the grabbed object can track the appropriate hand's position and rotation changes.

In my custom version of `hand-tracking-controls`, I have added in two additional properties to the existing event details for each pinch event:
- `hand`: a reference to the hand element that initiated the pinch gesture
- `wristOrientation`: a `THREE.Quaternion` version of the the `XRHand.WRIST` pose `transform.orientation` property, which seems to be a good proxy for the orientation of the hand as a whole

I could see an argument for using the strings `'left'` or `'right'` instead of the hand element for the `hand` parameter. I could see an argument for using the orientation of the index finger or thumb (or some combination of the two) as an orientation for the hand (I haven't tried any but the wrist). What's useful for moving an object is an initial reference frame, though, so the specific choice of orientation matters less for this use-case.

Question: If the wrist pose is null, should an event be fired? I currently don't fire if the pose is null.

I can provide a PR once finalized.

- A-Frame Version: master
- Platform / Device: Quest
- Reproducible Code Snippet or URL: Can put together a PR w/ example once decided on

cesmoak:
**Description:** As discussed in #4690, this change improves pinching for use in moving objects. Specifically, it adds a wrist orientation that can be used to track hand rotation over the course of a pinch gesture, allowing a grabbed object to be rotated appropriately. In addition, for improved performance, the `pinchmoved` event is removed in favor of a `pinched` state and two new members are added to the component that mirror the properties of the `pinchstarted` and `pinchended` events.

Note that since the only additional joint tracked is the wrist, I deferred adding a property to turn on and off joint tracking. Happy to add if that is preferred. If adding, would like clarity if it would be for all pinch-related joints or just the wrist. If it is for all pinch-related joints, suggest the property is a flag like `pinchTracking` or `gestureTracking` instead of `jointPositionCalculation` or similar.

Also, was hung up on `pinched` vs `pinching` for the new state, but went with `pinched` as it is the name of the previous flag member the component used internally. Also happy to change if desired.

**Changes proposed:**
- Remove `pinchmoved` event
- Add `pinched` state
- Add wrist orientation tracking; this is exposed on the `pinchstarted` and `pinchended` events as `orientation` and on the component as a `wristWorldOrientation` member
- Add `pinchWorldPosition` member to the component, which is the same as `position` on the `pinchstarted` and `pinchended` event details
- Update tests
- Add documentation

### DIFF
--- a/src/components/hand-tracking-controls.js
+++ b/src/components/hand-tracking-controls.js
@@ -85,9 +85,10 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
     this.onModelLoaded = this.onModelLoaded.bind(this);
     this.jointEls = [];
     this.controllerPresent = false;
-    this.isPinched = false;
-    this.pinchEventDetail = {position: new THREE.Vector3()};
     this.indexTipPosition = new THREE.Vector3();
+    this.pinchWorldPosition = new THREE.Vector3();
+    this.wristWorldOrientation = new THREE.Quaternion();
+    this.pinchEventDetail = {position: this.pinchWorldPosition, orientation: this.wristWorldOrientation};
 
     this.bindMethods();
 
@@ -204,8 +205,11 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
   detectPinch: (function () {
     var thumbTipPosition = new THREE.Vector3();
     return function () {
+      var el = this.el;
       var frame = this.el.sceneEl.frame;
       var indexTipPosition = this.indexTipPosition;
+      var wristWorldOrientation = this.wristWorldOrientation;
+      var pinchWorldPosition = this.pinchWorldPosition;
       var controller = this.el.components['tracked-controls'] && this.el.components['tracked-controls'].controller;
       var trackedControlsWebXR = this.el.components['tracked-controls-webxr'];
       var referenceSpace = this.referenceSpace || trackedControlsWebXR.system.referenceSpace;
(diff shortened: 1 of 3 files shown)


src/components/hand-tracking-controls.js
```diff
@@ -204,44 +205,40 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
...
+      var wristPose = frame.getJointPose(controller.hand[XRHand.WRIST], referenceSpace);
 
-      if (!indexTipPose || !thumbTipPose) { return; }
+      if (!indexTipPose || !thumbTipPose || !wristPose) { return; }
 
       thumbTipPosition.copy(thumbTipPose.transform.position);
+      thumbTipPosition.y += 1.5;
       indexTipPosition.copy(indexTipPose.transform.position);
+      indexTipPosition.y += 1.5;
+      wristWorldOrientation.copy(wristPose.transform.orientation);
```
dmarcos, line 228: What is the use case for the wrist orientation? Is not wrist position more useful?

dmarcos:
we now have [hand-tracking-grab-controls](https://aframe.io/docs/1.6.0/components/hand-tracking-grab-controls.html#main) that I think covers the requested functionality. Closing this but can reopen if necessary

PR #4710, opened 2020-10-23
Introduce a new 'clockSynced' animation component property

Elettrotecnica:
when set to true (default is false), the animation will be syncronized
with the browsers's clock. The effect of this is that, assuming all
viewers clocks are correct, everybody will see the animation in the
same state, as if it was an event happening at the same time for all
the viewers.

The change makes use of the seek animejs api https://animejs.com/documentation/#seekAnim and the input is computed based on the epoch and the animation's duration and direction.

### DIFF
--- a/src/components/animation.js
+++ b/src/components/animation.js
@@ -122,30 +122,31 @@ module.exports.Component = registerComponent('animation', {
     config.loop = data.loop;
     config.round = data.round;
 
-    this.clockSynced = data.clockSynced;
-
     // Start new animation.
     this.createAndStartAnimation();
   },
 
   tick: function (t, dt) {
     if (!this.animationIsPlaying) { return; }
-      if (this.clockSynced) {
-          var clock = new Date().getTime();
-          var direction = this.animation.direction;
-          if (direction == "alternate") {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. This adds a lot of code for a very specific use case. Also not sure synchronizing using the clock will yield good results. It seems that a server side approach would be more robust. Probably better to experiment with this in your own component that you can derive from the animation component. If there's demand for this feature we can consider incorporating in the future. Looking forward to seeing what you make.

Elettrotecnica:
Thanks for your feedback. Still I would like to keep discussing about this, because I think this use case is actually quite typical when one deals with "shared networked environments".

I argue that, when the animation is completely deterministic, a synchronized clock is actually all you need to give the illusion that a certain event is happening for everybody at the same time. A server-side approach might indeed be more robust, but would first of all require a server, and might not scale so well with a high number of clients, depending on how much information one needs to sync and how often... but I am no expert in networked gaming... How would you achieve the use case I try to address?

I do not claim this to be a perfect solution, there might be corner cases I did not think about, but I am not in a hurry to get the change into core and I can rework it as much as needed. In the meantime I can always fork the animation component and maintain it myself as you suggest for my own purposes.

All the best

PR #4722, opened 2020-11-14
hand-controls: also emit the raw gesture event

Elettrotecnica:
In my use case, I have created a custom component that is just a pair of hands that receive their position and gesture from a network backend. My need is that of mirroring whatever gesture has been emitted by a hand-control enabled entity, send it through the network and let my avatar's "remote hands" play the gesture to my peers.

I could immagine though, that other people might find useful to handle any kind of supported gesture, and in the code I have realized there is sometimes the need to know from which gesture we come from, therefore I have also added the lastGesture in the event payload.

All the best

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -338,6 +338,9 @@ module.exports.Component = registerComponent('hand-controls', {
     // Emit event for current gesture now active.
     eventName = getGestureEventName(gesture, true);
     if (eventName) { el.emit(eventName); }
+
+    // Emit the raw gesture+lastGesture event
+    el.emit('gesture', {'gesture': gesture, 'lastGesture': lastGesture});
   },
 
   /**


dmarcos:
Thanks. Not sure yet if this should be part of `hand-controls`. It looks like it can be handled on the application side as you did. I understand the convenience for you I would like to see where this is useful beyond your use case.

dmarcos:
Another note. This is emitting a new event `gesture` on top of the gesture event. If we were to merge this I would add last gesture to the existing event.

Elettrotecnica:
Thanks for your feedback.

Something that would work out for me is also that the events triggered inside hand-controls would span through the whole set of supported gestures. Right now 3 gestures would emit (at least) an event:

```
// Map animation to public events for the API.                                                                                                                                                                                                                                           
var EVENTS = {};
EVENTS[ANIMATIONS.fist] = 'grip';
EVENTS[ANIMATIONS.thumbUp] = 'pistol';
EVENTS[ANIMATIONS.point] = 'pointing';
```

But the total number of gestures is 6:

```
// Poses.                                                                                                                                                                                                                                                                                 
var ANIMATIONS = {
  open: 'Open',
  // point: grip active, trackpad surface active, trigger inactive.                                                                                                                                                                                                                       
  point: 'Point',
  // pointThumb: grip active, trigger inactive, trackpad surface inactive.                                                                                                                                                                                                                
  pointThumb: 'Point + Thumb',
  // fist: grip active, trigger active, trackpad surface active.                                                                                                                                                                                                                          
  fist: 'Fist',
  // hold: trigger active, grip inactive.                                                                                                                                                                                                                                                 
  hold: 'Hold',
  // thumbUp: grip active, trigger active, trackpad surface inactive.                                                                                                                                                                                                                     
  thumbUp: 'Thumb Up'
};
```

The mention of "public events" makes me think there might be a reason for it, but I am no expert.

Of course I would then need to retranslate the event name to the original gesture (e.g. "pistol" -> "thumb up"), hence the original idea of emitting just the raw gesture.

For my use case I am not actually interested in the lastGesture, the current one is enough (and I can always get the last one from my custom component).

Depending on what you prefer I could come with a change that removes the lastGesture from the event, or a proposal for emitting also the remaining gestures.

All the best

Elettrotecnica:
Just checking on this PR as I am trying to move away from maintaining a fork of aframe just for a couple of features. Most of them are already in custom components, but this one in particular would be tricky to rework that way without loads of duplication.

I suggest to further simplify the commit so that the one line becomes this:

`el.emit('gesture', gesture);`

This enables my use case and IMHO makes sense in general, as gives the developer a chance to react to the full spectrum of gesture events.

Any contraindication you see with this?

All the best

Elettrotecnica:
Just to comment that also PR #4883 and #4887 acknowledge the issue that not every gesture for hand-controls is currently emitting an event.

Elettrotecnica:
I am closing this and vouch that https://github.com/aframevr/aframe/pull/4887 is merged, as this fixes the inconsistencies between documented and emitted events :-)

PR #4725, opened 2020-11-15
doc: add an example of a French character set

vincentfretin:
**Description:**

**Changes proposed:**
- From A-Frame 0.9.0 and above, you don't need to rename `*.png` to `*-msdf.png` anymore, A-Frame loads the image defined in the json file.
- Add an example of a French character set you can use


dmarcos:
Thank you!

Issue #3137, opened 2017-10-06
Model does not unload from memory after removing from DOM


Pterko:
**Description:**

A loaded model (gltf-model) doesn't unload from memory after it removed from DOM.

- A-Frame Version: 0.7.0
- Platform / Device: At least Windows/Chrome (possibly all platforms)
- Reproducible Code Snippet or URL: https://pterko.github.io/aframe/memory_bug_demo/test.html

How to reproduce:
 - Wait until page loads
 - When page loads, take a memory snapshot
 - Wait 10-15s before model will appear and loaded, take a snapshot
 - Wait until model will be deleted, take memory snapshot again

Expected behavior: amount of used memory decreases
Real behavior: amount of used memory don't change
![image](https://user-images.githubusercontent.com/3855974/31297860-475fe166-aaf0-11e7-92da-e68ae860eccf.png)

calrk:
I've noticed this in my project as well, though I use a sllightly different version of the gltf loader. When I unmount/unload any gltf object, I run it through a function as below to make sure to dispose of any material/geometry/textures.

Could add this somewhere to the a-frame source but not sure where would be appropriate or if there is a simpler way of doing it.

```
export const disposeObject3D = object => {
  object.traverse(obj => {
    if(obj.material){
      obj.material.dispose();
      if(obj.material.map){
        obj.material.map.dispose();
      }
      if(obj.material.lightMap){
        obj.material.lightMap.dispose();
      }
      if(obj.material.aoMap){
        obj.material.aoMap.dispose();
      }
      if(obj.material.emissiveMap){
        obj.material.emissiveMap.dispose();
      }
      if(obj.material.bumpMap){
        obj.material.bumpMap.dispose();
      }
      if(obj.material.normalMap){
        obj.material.normalMap.dispose();
      }
      if(obj.material.displacementMap){
        obj.material.displacementMap.dispose();
      }
      if(obj.material.roughnessMap){
        obj.material.roughnessMap.dispose();
      }
      if(obj.material.metalnessMap){
        obj.material.metalnessMap.dispose();
      }
      if(obj.material.alphaMap){
        obj.material.alphaMap.dispose();
      }
    }
    if(obj.geometry){
      obj.geometry.dispose();
    }
  });
}
```

donmccurdy:
@calrk Did your function above resolve the memory leaks for your project?

[...]

PR #4734, opened 2020-11-24
Dispose

dmarcos:
Just for illustration


vincentfretin:
I know it's for illustration purpose and the PR is quite old now, but while I'm reading really old issues, be aware that the code in this PR doesn't dispose textures. You may look at https://github.com/aframevr/aframe/issues/3137#issuecomment-1242742566 that include a link to the wiki with a better `disposeNode` function that close and dispose the textures.
Also with #5112 we have better handling of disposing the renderer.

dmarcos:
Won't merge

PR #4743, opened 2020-12-08
Fix raycaster overriding line default color

GMartigny:
**Description:**

When creating a `raycaster` (or a `laser-controls` for ease of testing), it assigns a default value to the `line`'s color.
It means that [the example](https://aframe.io/docs/1.1.0/components/laser-controls.html#customizing-the-line) doesn't works:

```html
<a-entity laser-controls line="color: red"></a-entity>
```

My fix keep the `raycaster`'s `lineColor` property for backward compatibility. But in my opinion, it should be deprecated in favor of `line`'s `color` property.

**Changes proposed:**
 - Remove default value to `raycaster`'s `lineColor` property.

**Reproduction**

```html
<a-scene>
  <a-sky color="#49ACEF"></a-sky>
  <a-entity laser-controls line="color: red; opacity: 0.2"></a-entity>
</a-scene>
```
The opacity is working but not the color.

### DIFF
--- a/docs/components/laser-controls.md
+++ b/docs/components/laser-controls.md
@@ -81,5 +81,5 @@ See [*Raycaster: Customizing the Line*][customize].
 For example:
 
 ```html
-<a-entity laser-controls line="color: red; opacity: 0.75"></a-entity>
+<a-entity laser-controls raycaster="lineColor: red; lineOpacity: 0.5"></a-entity>
 ```
--- a/docs/components/raycaster.md
+++ b/docs/components/raycaster.md
@@ -14,7 +14,7 @@ examples: []
 The raycaster component provides line-based intersection testing with a
 [raycaster][wiki-raycasting]. Raycasting is the method of extending a line from
 an origin towards a direction, and checking whether that line intersects with
-other entites.
+other entities.
 
 The raycaster component uses the [three.js raycaster][3ray]. The raycaster
 checks for intersections at a certain interval against a list of objects, and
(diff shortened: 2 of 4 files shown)


GMartigny:
For now, I just changed the source-code. If this is accepted, I guess I should update `docs` and `dist` ?

dmarcos:
Using `line` to set the raycaster color is a brittle since that depends on the internal implementation of `raycaster`. That could change in the future. Is not the `lineColor` attribute working for you to customize the raycaster?

GMartigny:
> Using line to set the raycaster color is a brittle since that depends on the internal implementation of raycaster.

I agree with you. In that case, it should be possible to change the "laser" opacity (without depending on `line`).

> Is not the lineColor attribute working for you to customize the raycaster?

Yes it does. It's juste that the ["Customizing the Line" on laser-controls](https://aframe.io/docs/1.1.0/components/laser-controls.html#customizing-the-line) and ["Customizing the Line" on raycaster](https://aframe.io/docs/1.1.0/components/raycaster.html#customizing-the-line) on the documentation are not working.
(I only tested the laser-controls, but both should fail the same way.)

dmarcos:
What about modifying this PR to add a `lineOpacity` property to the `raycaster` component? The [docs](https://github.com/aframevr/aframe/blob/master/docs/components/laser-controls.md#customizing-the-line) have to be updated accordingly. Thanks so much for looking into this.

GMartigny:
> The docs have to be updated accordingly

I don't see any update on the docs yet. I can edit them if you want.

Should I push update to the `dist` directory also ?

dmarcos:
Thanks. Can you edit the [docs](https://github.com/aframevr/aframe/blob/master/docs/components/laser-controls.md#customizing-the-line) as part of this PR? 

The new `lineOpacity` property has to be also documented in the [raycaster component](https://github.com/aframevr/aframe/blob/master/docs/components/raycaster.md#properties) below `lineColor`

Don't push changes on the `dist` directory

dmarcos:
Thanks for remembering the tests.

Merci, bien fait! 

Congrats on your first contribution! 🥳

PR #4748, opened 2020-12-12
Use uglifyify (instead of minifyify) for build

GMartigny:
**Description:**

The package [minifyify](https://github.com/ben-ng/minifyify#warning-unmaintained) is unmaintained. This could lead less performant or larger build.

In fact, changing to [uglifyify](https://github.com/hughsk/uglifyify) output a ~3% (from 1,349,116o to 1,311,362o) lighter build.

This change requires to upgrade a few other packages. This should be considered with attention. But, I didn't change any production dependencies and I run all test (and lint) with flying colors.

**Changes proposed:**
 - Use `uglifyify`
 - Update `browserify`
 - Update related dependencies
 - Remove unused `uglifyjs`


vincentfretin:
Nice, when I tried last time I gave up. Did you run the examples manually with the minified build, does it works?

GMartigny:
I just test a few of them and I can't see any issue (chrome view and Valve Index).
It may be sound to test with another headset at least.

What example cover the most of the codebase (approximately) ?

vincentfretin:
If some of the examples works it should be ok I think. I just ask because last time when I tried to use uglifyify I thought I successfully minified the build but it didn't work at all for any example. :)

GMartigny:
Why on the subject of package minification, why is the `docs` folder included in the tarball ?
I don't think users look at the documentation on their `node_modules`.

dmarcos:
Thanks for measuring the size difference. Much appreciated. The data is interesting.

In general, it's ok if a package is not maintained if it does its job and has no outstanding bugs.

There's a wrinkle with `minifyify` though. It chokes with some files. `DracoLoader` for example. It doesn't like certain combination of symbols. You can see in the build that some files are not minimized: that could explain some of the difference in size you see.

I'm open to replace `minifyify` with something else. Replacing unrelated dependencies makes me a bit anxious though 😄  

> Why on the subject of package minification, why is the docs folder included in the tarball?

docs in npm package was introduced by `123102167dc`. [A-Frame site](https://aframe.io/docs/1.1.0/introduction/) relies on those to show versioned docs. I agree it's a tad ugly and there should be a more elegant way to deal with that.

dmarcos:
Closing this for now. Thanks for submitting. 

We should figure out why `minifyify` cannot handle some of the files before swapping it with something else.

PR #4749, opened 2020-12-13
Remove lint

GMartigny:
**Description:**

I notice that `semistandard` was critically outdated. Updating it triggered a lot of new lint warnings. Especially because it moved to ES6.

I have a few questions about all this:
 - Do you care about all this ?
 - If yes, do you still want to adhere to the `semistandard` rules ?
 - Why is there a `.jshintrc` file (which is ignored by `semistandard` anyway) ?
 - Would you accept a large PR containing a huge lots of small lint removing ?

I have already replaced all `var` with `let` as an example.

**Changes proposed:**
 - Updating `semistandard`
 - Replaced `var` with `let`


dmarcos:
> Do you care about all this ?

Thanks. A-Frame code base is ES5. No plans to move to ES6 at the moment.

> If yes, do you still want to adhere to the semistandard rules ?

Yes. semistandard has served us well. Not sure if there's any reason to switch to something else at the moment

> Why is there a .jshintrc file (which is ignored by semistandard anyway) ?

It came from dev requests that use Atom / Sublime plugins relying on `.jshintrc`  to format code automatically. I don't use it and I would remove it to tidy up the repo but prefer to not disrupt existing workflows. 

> Would you accept a large PR containing a huge lots of small lint removing ?

Yes, if there's a good reason for those small changes and the tests pass 😄 

I'm closing this for now but can continue the discussion if necessary

GMartigny:
> semistandard has served us well. Not sure if there's any reason to switch to something else at the moment

It's just that it not configurable by design. With the last version of semistandard, it raised about 7500 lint warning on the aframe code-base.

> I don't use it and I would remove it to tidy up the repo but prefer to not disrupt existing workflows.

People should use the [existing plugins](https://github.com/standard/standard#are-there-text-editor-plugins) for that.

dmarcos:
Thanks for the reply.

> It's just that it not configurable by design. 

We haven't had the need to configure it so far.

> With the last version of semistandard, it raised about 7500 lint warning on the aframe code-base.

With the existing version `npm run lint` returns a clean output for me. I guess you meant the latest version published in npm. As mentioned `semistandard` is serving us well. You mentioned `critically` outdated. Are there any outstanding bugs you've seen that we should address? Thanks so much.

GMartigny:
> I guess you meant the latest version published in npm?

Indeed.

> Are there any outstanding bugs you've seen that we should address?

Not a all ! You already answered this when you said "A-Frame code base is ES5. No plans to move to ES6 at the moment."
So, as long as you don't wish to move to ES6, I don't think you'll need to update anything.

I don't know if this is a good habit or not, but I always keep every dependencies up to date. This allow me to use the latest features and remove potential vulnerabilities. But you do you ;)

dmarcos:
Yeah. I prefer to be deliberate. Bumping a dependency comes with risks of regressions, API changes and new bugs introduced. Thanks for the PR though. Much appreciated.

PR #4758, opened 2020-12-29
document use of `oneOf` prop within schema

kylebakerio:
**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/docs/core/component.md
+++ b/docs/core/component.md
@@ -178,6 +178,16 @@ schema: {type: 'string'}  // default: ''
 schema: {type: 'vec3'}  // default: {x: 0, y: 0, z: 0}
 ```
 
+#### Specifying Possible Values
+
+You can also provide an array of acceptable values to use:
+
+```
+  schema: {
+    rayOrigin: {default: 'entity', oneOf: ['mouse', 'entity']}
+  },
+```
+
 #### Custom Property Type
 
 We can also define our own property type by providing a `parse` and `stringify`


docs/core/component.md
```diff
@@ -178,6 +178,16 @@ schema: {type: 'string'}  // default: ''
 schema: {type: 'vec3'}  // default: {x: 0, y: 0, z: 0}
 ```
 
+#### Specifying Possible Values
+
+You can also provide an array of acceptable values to use:
```
dmarcos, line 183: Could this be misleading? The value range is not checked at run time. It's only a hint for the inspector to show the combo box widget. We should probably add a `#### Inspector` section explaining: the property type is used by the inspector to show the appropriate widget. `max` and `min` can be used to constrain the range of available values and `oneOf` to display a list of options.

kylebakerio:
I see. I definitely think that would be misleading, then. I can either update this text block with the correct description, or we can format it under a different section (I think that also makes sense, I've always wanted to find some full inspector documentation)... But are there objections to just adding this functionality as an option (runtime comparison of this kind)?

dmarcos:
Not sure if runtime boundary check is very useful. We could print warnings in console but comes with a performance penalty due to the additional logic and increases complexity in code that it's already complex. My preference would be to wait  and see more demand of such a feature.

Regarding docs. Maybe adding a note in the [inspector / components section](https://aframe.io/docs/1.1.0/introduction/visual-inspector-and-dev-tools.html#components-panel) explaining how property types are used to display the appropriate widget. Does it sound good?

dmarcos:
I'm closing this because it can be misleading. OneOf is just a hint for the inspector and A-Frame doesn't enforce that the values are in the set

dmarcos:
Thanks!

Issue #4759, opened 2020-12-29
cursor does not update raycaster's origin and direction when dynamically updated from 'rayOrigin: mouse' to 'rayOrigin:entity'


kylebakerio:
As I posted on slack:

> So, I've been trying to dynamically set my cursor's raycaster's origin from mouse to entity within my app. What I've noticed is that by doing so in this order, if I have touched the mouse, then my termination point (the spot that is 'clicked', the raycaster component's lineData.end is wherever my mouse was on screen as opposed to the expected behavior (and the behavior I get if I start with rayOrigin: entity without having rayOrigin: mouse first, or if I do it this way but don't touch the mouse before changing rayOrigin), where the termination of the line is directly the center of wherever I'm staring.
> I think this should be regarded as a bug, right? Probably the rayOrigin just wasn't quite designed to handle an update from mouse to entity correctly?
> Going to go look in source code and maybe attempt a pull request if I figure it out, but thought I'd ask here at the same time.
> 1 reply
> 
> kylebaker.io  29 minutes ago
> yeah, seems raycaster origin (which I'd already realized wasn't being set right and had been manually updated) as well as direction are both not being updated when the parent cursor's rayOrigin is updated. setting direction in-app with raycaster.data.direction.set(0,0,-1)  instantly fixes the issue.

- A-Frame Version: 1.1.0
- Platform / Device: desktop
- Reproducible Code Snippet or URL: 

```html
<a-entity
              id="cursor"
              cursor=" rayOrigin:mouse;"
              position="0 0 -0.1"
              raycaster="showLine: true; far: 2; interval:50; near: 0.03; objects: [gui-interactable], .raycastable;"
          >
```
```js
// later:
let cursor = document.querySelector("[cursor]")

 cursor.setAttribute('cursor',{
                fuse: true,
                rayOrigin: 'entity',
})
```

running this fixes the issue for me locally:

```js
       console.warn("PATCHING CURSOR")

       vrgc.getComponent('cursor').updateOriginAndDirection = function() {
            console.warn("called custom updateOriginAndDirection")
            vrgc.getComponent('raycaster').data.direction.set(0,0,-1)
            vrgc.getComponent('raycaster').data.origin.set(0,0,0)
        }

       vrgc.getComponent('cursor').update = function(oldData) {
            console.warn("called custom cursor update")
            if (this.data.rayOrigin !== oldData.rayOrigin) {
                if (this.data.rayOrigin === "entity") {
[...]

PR #4760, opened 2020-12-29
handle update on rayOrigin from mouse -> entity

kylebakerio:
fixes issue https://github.com/aframevr/aframe/issues/4759

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
I usually recommend adding one cursor on `<a-scene cursor="cursor="rayOrigin: mouse">` and then additional cursors in relevant entities. Would that be useful in your scenario?

kylebakerio:
That's an interesting thought. Is there documentation that describes that idea? Do you just set multiple cursors, and then disable/enable whichever ones you want active accordingly? I imagine I _could_ do it that way, but I wouldn't be sure what entities to put the cursor on (shouldn't the cursor generally be a child of the active camera? Would the only result be the origin of the raycaster being defined from whatever element you make the child a cursor of?). If I had multiple cursors enabled at the same  time, I assume that would cause multiple click events to be fired, right?

It seems like a strange model to me to create multiple cursors and keep track of their state separately than to just update the one cursor component, if only one is ever active at a time. Not sure.

kylebakerio:
Based on your suggestion before, I recently refactored to be using many cursor objects for various control schemes. I disable the raycaster for the mouse within cardboard and quest, for example, disable the fuse/gamepad view-based cursor when using desktop mouse mode, and always leave a third one running to check if the other player (for example) is blocking the player's view of the game board. I then of course add 4th and 5th raycasters through the HMD controllers when appropriate. This is a better solution.

Also, digging further, I found that '0 0 0' is treated as an implicit flag in that code block, meaning my origin seemed to be at world origin 0 0 0, which caused it to start working in some cases for me, but not all. I since have started setting x to 0.001, and finally it is working as desired everywhere, instead of in the range of the raycaster from world origin 0 0 0... but I find that code block slightly confusing and so haven't decided what the best way to do this is.

Either way, though, it doesn't seem that the cursor responds well to updates in its current form. Perhaps this should be somehow documented at least, if not updated. 

> Improvements ideas. On laser-controls we could enable / disable the raycaster on controllerconnected / controllerdisconnected to save on the raycaster tock calculations. The same for cursor="rayOrigin: mouse": we could enable / disable the raycaster on exit-vr / enter-vr events.

I am essentially doing this in my code now, I have a function for detecting the various viewport/input combinations my app runs with and setting them that I run on various events like enter-vr/exit-vr, and am designing an in-game init gui sequence to select it explicitly.

dmarcos:
A bit conflicted with this. For mouse cursor we recommend adding to the scene so no change to dynamic change to entity should be necessary. In any case the contract of a component is that it can change dynamically so this is technically a bug.

We should also probably considering adding a message to the component that for a `mouse` type of cursor it's recommended be added to the scene.

dmarcos:
Thanks!

PR #4761, opened 2020-12-29
Added optimized version tick handling code

Earl-Brown:
While reading the docs, I noticed a couple of code elements that could be improved.

Realizing that this is an introduction document, I added a code description those performance improvements.

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thanks so much! We have performance considerations in the [tick documentation](https://aframe.io/docs/1.1.0/introduction/best-practices.html#tick-handlers). Maybe better to add any missing info there?

PR #4762, opened 2020-12-30
Added a couple of performance tips

Earl-Brown:
I added the suggestion of creating variables outside of arrays.

I added the suggestion of a lookup table instead of a switch statement.

### DIFF
--- a/docs/introduction/best-practices.md
+++ b/docs/introduction/best-practices.md
@@ -118,6 +118,72 @@ an A-Frame scene:
   which will skip `.setAttribute` and animate JS values directly. 
   For example, instead of `material.opacity`, 
   animate `components.material.material.opacity`.
+- Initialize all variables possible outside of your loops.
+  In this example:
+  ```js
+    var factor = this.data.speed / distance;
+    ['x', 'y', 'z'].forEach(function (axis) {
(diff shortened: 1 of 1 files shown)


docs/introduction/best-practices.md
```diff
@@ -118,6 +118,72 @@ an A-Frame scene:
   which will skip `.setAttribute` and animate JS values directly. 
   For example, instead of `material.opacity`, 
   animate `components.material.material.opacity`.
+- Initialize all variables possible outside of your loops.
+  In this example:
+  ```js
```
dmarcos, line 123: Thanks for submitting. Super super appreciated. This is not specific to A-Frame but general JS best practices. Is there any external resource we can link from the docs to keep the docs focused on A-Frame related topics?

Earl-Brown: I can't really think of any - these are things I "picked up along the way".

JGrotex: here 2 nice links about this **general topic**, but it's nicely brought to the `a-frame` context here!

- https://www.w3.org/wiki/JavaScript_best_practices
- https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html

Issue #4764, opened 2021-01-02
a-text doesnt render when positioned behind camera


mmnoo:
**Description:**

Simple `a-text` elements dont show when positioned behind the camera.

- A-Frame Version: 1.1.0
- Platform / Device: MacOS 10.15.7, Chrome 87.0.4280.88 & Samsung Galaxy S7 Chrome 87.0.4280.101
- Reproducible Code Snippet or URL:
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Label position troubleshooting</title>
    <meta name="description" content="Hello, WebVR! • A-Frame" />
    <script src="https://aframe.io/releases/1.1.0/aframe.min.js"></script>
  </head>
  <body>
    <a-scene background="color: #333">
      <a-entity
        light="type:directional; castShadow:true;"
        position="-2 10 0"
      ></a-entity>
      <a-plane
        position="0 -2 -4"
        rotation="-90 0 0"
        width="4"
        height="4"
        color="#cccccc"
        shadow
      ></a-plane>

      <a-sphere position="0 1 -2" radius=".1" color="#EF2D5E" shadow></a-sphere>
      <a-text value="in front of camera" position=" 0 1.2 -2" color="#000" ></a-text>
      
      <a-sphere position="0 1 2" radius=".1" color="#EF2D5E" shadow></a-sphere>
     <!-- this text doesnt show -->
      <a-text value="behind camera" position=" 0 1.2 2" color="#000" ></a-text>
    </a-scene>
  </body>
</html>
```
https://glitch.com/edit/#!/sore-probable-sunscreen?path=index.html%3A8%3A9

dmarcos:
That's expected. By default 3D elements are not rendered when the back face faces the camera. Rotate the the element 180 degrees:

https://glitch.com/edit/#!/broken-coordinated-farmer?path=index.html%3A28%3A75

Notice that even if you disable backface culling you still have to rotate 180 degrees or the text will be mirrored

mmnoo:
Beautiful! Thank you @dmarcos.

PR #4765, opened 2021-01-03
Update a-text.md

mmnoo:
**Description:**
Updates to `a-text` documentation.

**Changes proposed:**
- Updating `a-text` documentation to include learnings from https://github.com/aframevr/aframe/issues/4764

### DIFF
--- a/docs/primitives/a-text.md
+++ b/docs/primitives/a-text.md
@@ -42,6 +42,11 @@ Wraps the [text component][text].
 | wrap-pixels    | text.wrapPixels    |
 | z-offset       | text.zOffset       |
 
+## Usage Notes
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks! Much appreciated! I think the info is a bit specific to your use case to be included in the docs. 

You stumbled on [Back Face Culling](https://en.wikipedia.org/wiki/Back-face_culling), something 3D developers should be aware of in general.

I recommend opening a question in [Stack Overflow](https://stackoverflow.com/questions/tagged/aframe) with your glitch illustration the issue. You can also answer your own question with the fixed glitch. It's easy to reference for future learners and gives better SEO than the docs. Thanks

mmnoo:
Ya, that makes sense.  I am 100% new to 3D anything. Thanks for providing language for it. :) 

The git hub issue existing is searchable enough for future rookies who make the same mistake I suppose. I spent an embarrassing amount of time on it.

PR #4782, opened 2021-01-29
Fix geometry-merger links

KovacsGG:
Geometry-merger component links were not parsed correctly, because of the extra `[pool]` before the link's definition.
The definition was also duplicated as `[merge]`.

**Description:**
Errors in the genereal Performance sections were fixed by editing the link definitions.

**Changes proposed:**
- Remove one line to fix syntax
- Remove a superfluous line


dmarcos:
Thanks!

PR #4795, opened 2021-02-13
Add registerShader Texture Example to Materials Documentation

mordof:
**Description:**

I found it very difficult to find information on using textures in a registerShader setup. It's mostly described with the schema option, but doesn't have anything in the documentation currently that even mentions it. I ended up googling for examples, only to run into many historical discussions about whether or not to even add this into a-frame. Ultimately it was very confused and took several hours to sort out.

**Changes proposed:**

Add a link + gif demonstrating a registerShader custom shader which uses textures, where the texture can be specified using the Asset Managment System A-Frame has.

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -587,6 +587,10 @@ And using from HTML markup:
 
 ***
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

dmarcos:
Congrats on your first contribution 🙌

PR #4797, opened 2021-02-16
Use background to set the default environment map

AdaRoseCannon:
**Description:**

Adds behavior to `background` to allow it to set the environment property on the scene element.

It automatically generates an environment map as a cube texture which can be overridden by setting the environment map on a per element basis. 

## No background:
image

## Background is hotpink, no skybox
image

## Background is still hot pink but with an image skybox
image

### DIFF
--- a/docs/components/background.md
+++ b/docs/components/background.md
@@ -13,6 +13,10 @@ frustum culling issues when `a-sky` is further than the far plane of the
 camera. There are no unexpected occlusions either with far objects that might
 be behind of the sphere geometry of `a-sky`.
 
+The background component can also generate a default envrionment cube map for all
+materials, if you find GLB models end up too dark or reflective materials don't
+reflect the environment this will provide a default reflective environment.
+
 ## Example
 
 The example below sets the background color to red.
@@ -23,7 +27,8 @@ The example below sets the background color to red.
 
 ## Properties
 
-| Property    | Description                                               | Default Value |
-|-------------|-----------------------------------------------------------|---------------|
-| color       | Color of the scene background.                            | black         |
-| transparent | Background is transparent. The color property is ignored. | false         |
+| Property            | Description                                               | Default Value |
+|---------------------|-----------------------------------------------------------|---------------|
+| color               | Color of the scene background.                            | black         |
+| transparent         | Background is transparent. The color property is ignored. | false         |
+| generateEnvironment | Whether to generate a default environment                 | true          |
--- a/docs/components/background.md
+++ b/docs/components/background.md
@@ -13,7 +13,7 @@ frustum culling issues when `a-sky` is further than the far plane of the
 camera. There are no unexpected occlusions either with far objects that might
 be behind of the sphere geometry of `a-sky`.
 
-The background component can also generate a default envrionment cube map for all
+The background component can also generate a default environment cube map for all
 materials, if you find GLB models end up too dark or reflective materials don't
 reflect the environment this will provide a default reflective environment.
 
--- a/src/components/scene/background.js
+++ b/src/components/scene/background.js
@@ -9,10 +9,8 @@ module.exports.Component = register('background', {
     generateEnvironment: {default: true}
   },
   init: function () {
-    var cubeRenderTarget = new THREE.WebGLCubeRenderTarget(128, { format: THREE.RGBFormat, generateMipmaps: true, minFilter: THREE.LinearMipmapLinearFilter });
-    var cubeCamera = new THREE.CubeCamera(1, 100000, cubeRenderTarget);
-    this.cubeRenderTarget = cubeRenderTarget;
-    this.cubeCamera = cubeCamera;
+    this.cubeRenderTarget = new THREE.WebGLCubeRenderTarget(128, { format: THREE.RGBFormat, generateMipmaps: true, minFilter: THREE.LinearMipmapLinearFilter });
+    this.cubeCamera = new THREE.CubeCamera(1, 100000, this.cubeRenderTarget);
     this.needsEnvironmentUpdate = true;
   },
   update: function () {
--- a/src/components/scene/background.js
+++ b/src/components/scene/background.js
@@ -25,6 +25,11 @@ module.exports.Component = register('background', {
       object3D.background = new THREE.Color(data.color);
     }
 
+    if (scene.environment && scene.environment !== this.cubeRenderTarget.texture) {
+      console.warn('Background will not overide predefined environment maps');
+      return;
+    }
+
     if (data.generateEnvironment) {
       scene.environment = this.cubeRenderTarget.texture;
     } else {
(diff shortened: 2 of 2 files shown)


AdaRoseCannon:
For the future this could also be updated to support the environment cube maps generated by WebXR Lighting Estimation:

https://github.com/immersive-web/lighting-estimation/blob/main/lighting-estimation-explainer.md

docs/components/background.md
```diff
@@ -13,6 +13,10 @@ frustum culling issues when `a-sky` is further than the far plane of the
 camera. There are no unexpected occlusions either with far objects that might
 be behind of the sphere geometry of `a-sky`.
 
+The background component can also generate a default envrionment cube map for all
```
vincentfretin, line 16: typo envrionment => environment

src/components/scene/background.js
```diff
@@ -15,6 +16,30 @@ module.exports.Component = register('background', {
...
+      this.needsEnvironmentUpdate = true;
+      scene.environment = cubeRenderTarget.texture;
+      this.el.sceneEl.addEventListener('loaded', function () {
+        this.needsEnvironmentUpdate = true;
+      });
+    }
+  },
+
+  tick () {
+    if (!this.data.auto && !this.needsEnvironmentUpdate) return;
```
vincentfretin, line 34: I don't see `auto` defined in schema.

vincentfretin (commented):
The pictures are nice. :)

src/components/scene/background.js
AdaRoseCannon: Thanks for pointing that out! That would be an expensive error.

src/components/scene/background.js
```diff
@@ -5,7 +5,8 @@ var COMPONENTS = require('../../core/component').components;
 module.exports.Component = register('background', {
   schema: {
     color: {type: 'color', default: 'black'},
-    transparent: {default: false}
+    transparent: {default: false},
+    generateEnvironment: {default: true}
```
dmarcos, line 9: Do we want the default to be true?

src/components/scene/background.js
```diff
@@ -15,6 +16,30 @@ module.exports.Component = register('background', {
       return;
     }
     object3D.background = new THREE.Color(data.color);
+
+    if (data.generateEnvironment) {
```
dmarcos, line 20: Maybe we can separate into a separate `initEnvironmentMap` method. I think we just need to run it once and not recreate if the component is updated. We should check if there's already an environment.

src/components/scene/background.js
```diff
@@ -15,6 +16,30 @@ module.exports.Component = register('background', {
...
+      const scene = this.el.sceneEl.object3D;
+      const cubeRenderTarget = new THREE.WebGLCubeRenderTarget(128, { format: THREE.RGBFormat, generateMipmaps: true, minFilter: THREE.LinearMipmapLinearFilter });
+      const cubeCamera = new THREE.CubeCamera(1, 100000, cubeRenderTarget);
+      this.cubeCamera = cubeCamera;
+      this.needsEnvironmentUpdate = true;
+      scene.environment = cubeRenderTarget.texture;
+      this.el.sceneEl.addEventListener('loaded', function () {
```
dmarcos, line 27: I believe this event handler is not needed.

dmarcos:
Nice job! It's pretty close. Just a few changes. 

This is going to be really useful. Some people got confused with models looking dark and we didn't have a quick solution to give them

src/components/scene/background.js
AdaRoseCannon: I think it's pretty safe and not a huge performance hit as it only renders the once. It will even fix materials unexpectedly not looking right because they expect an environment map. So it's a nice feature, that you only need to type 'background` to get.

I'd even be tempted to take it a step further and make `background="transparent:true"` a default component on `<a-scene>` or at least add `background` to the hello-world boilerplate just to get the better behaviour by default.

src/components/scene/background.js
```diff
@@ -15,6 +16,30 @@ module.exports.Component = register('background', {
...
+        this.needsEnvironmentUpdate = true;
+      });
+    }
+  },
+
+  tick () {
+    if (!this.needsEnvironmentUpdate) return;
+    const scene = this.el.object3D;
```
dmarcos, line 35: replace with var (A-Frame is ES5)

AdaRoseCannon: Haha oh wow blast from the past!

I should also change `tick () {}` too as well!

src/components/scene/background.js
AdaRoseCannon: Thanks for the spot this should've gone in init() not update!!

src/components/scene/background.js
AdaRoseCannon: I think you are right,

On a related note, do you know if there is an event handler for things like an object or image has finished loading asynchronously, which would allow this to still work nicely even if someone is loading their skybox etc without using `<a-assets>`?

dmarcos: mmh not sure if there's a way to universally know it all assets have loaded if `<a-assets>` is not used. Components can do all sort of arbitrary stuff.

AdaRoseCannon:
I think this now resolves the issues.

src/components/scene/background.js
AdaRoseCannon: I guess if someone wanted this behaviour they could do `background.needsEnvironmentUpdate=true` on this component from their custom components.

It might be nice if components like `material` would fire a bubbling event once they had loaded an image asynchronously so components like this can update the texture in a few common situations. Although this feature may be the carrot needed to get more developers to use `<a-assets>`

src/components/scene/background.js
```diff
@@ -5,16 +5,48 @@ var COMPONENTS = require('../../core/component').components;
...
     color: {type: 'color', default: 'black'},
-    transparent: {default: false}
+    transparent: {default: false},
+    generateEnvironment: {default: true}
+  },
+  init: function () {
+    var cubeRenderTarget = new THREE.WebGLCubeRenderTarget(128, { format: THREE.RGBFormat, generateMipmaps: true, minFilter: THREE.LinearMipmapLinearFilter });
+    var cubeCamera = new THREE.CubeCamera(1, 100000, cubeRenderTarget);
+    this.cubeRenderTarget = cubeRenderTarget;
+    this.cubeCamera = cubeCamera;
```
dmarcos, line 15: These two can be in the same line
```
 var cubeRenderTarget = this.cubeRenderTarget  = ...
 var cubeCamera = this.cubeCamera  = ...
```

AdaRoseCannon:
that's done now :-)

dmarcos:
It looks great! Thanks so much! 🥇

AdaRoseCannon:
Since Lighting Estimation has landed in Samsung Internet, I'm going to update this to handle lighting estimation by default.

Issue #4798, opened 2021-02-17
Getting "magic window" mode to trigger on Android tablets is difficult OR impossible.


PlumCantaloupe:
I am huge fan of the vastly underrated "magic window" mode that WebXR/A-Frame allows for mobile devices (and my initial research suggests that many others agree with me), but I have found that when using an Android tablet running Chrome A-Frame cannot get access to the accelerometers/gyrometers, even in "mobile site" mode. Switching to request "desktop site" or "request mobile site" both create the same modal pictured below on Android Chrome. 

However one can use Firefox on Android and "request mobile device" to access the magic window, but unfortunately Firefox is _very_ slow. On an iPad one can use either Safari, Firefox, or Chrome and "request mobile site" and it works; but **having to find the "request mobile site" feature makes for a terrible user experience.**

- Assuming this isn't an A-Frame bug, can we force the mobile site to load somehow to make this easier for WebXR users to use magic window mode?
- If on Android Chrome, this is a bug with their mobile site request is a there a workaround to forcing device access?

Thank you. Looking forward to some discussion on this.

![IMG_3743](https://user-images.githubusercontent.com/264985/108233679-07d9db00-7112-11eb-9038-85c1d4e08c9a.jpg)

- A-Frame Version: current
- Platform / Device: Android Tablet running Chrome (don't have an Android phone to test)
- Reproducible Code Snippet or URL: Aframe.io examples :)

dmarcos:
Unfortunately most tablets today default to desktop mode and websites cannot override that setting. Only the user can change it and hence the need of the prompt on the A-Frame side. On desktop mode mobile browsers don't have the devicemotion API enabled that it's necessary for magic window mode. Not sure why some Android tablets have also devicemotion API disabled in mobile mode. It looks like an overlook or a bug on the browser side.

PlumCantaloupe:
Thanks for the response. This is tricky as it appears to affect all browsers built from Chromium i.e., Samsung browser. I’ll keep poking around. I am sorry to poke, but any  suggestions @donmccurdy?

PlumCantaloupe:
Okay, some progress. I will look at creating. PR in a bit but it seems the problem lies with A-Frame asking if a tablet "isMobile" in look-controls.js and is asking for "DeviceOrientationEvent.requestPermission()" in 'device-orientation-permission-ui.js' which doesn't seem to exist on Chrome. Chrome just allows it in either desktop or "mobile" mode (which is interesting).

[...]

PR #4803, opened 2021-02-19
Fixing Magic Window Mode on Android Tablets

PlumCantaloupe:
**Description:**
This addresses #4798 by adding to the "isTablet()" regex in [utils]device.js in to find "Android" but ignoring "Mobile" in the the useragent string. ["Android+Mobile" is a check for Android phones whereas just "Android" is a check for tablets in mobile site mode](https://developer.chrome.com/docs/multidevice/user-agent/).

**Changes proposed:**
- See above
- **For the future**, I think we should discuss whether it is important to even check for whether a device is mobile/tablet before trying to access accelerometers/gyrometers. As most tablets default to "Desktop mode" and I can imagine phones will soon, _asking the user to find "Request Mobile Site" is an extra confusing step_ to access the more interesting [default??] "magic window" mode.

**Tested On:**
- iPad Pro 10.5" (2017) (Chrome, Firefox, Safari)
- Samsung Galaxy Tab S6 Lite (Chrome, Firefox)
- iPhone XR (Safari)


dmarcos:
Thanks. I need to think about it a bit.

FWIW, I don't think phones will default to desktop mode. Tablets got more powerful and with larger displays. Many are used today with a keyboard attached most of the time. Apple in particular is pushing iPads as alternative to laptops for some people. It makes sense they want to offer a "desktop like" browsing experience.

dmarcos:
The ideal solution for us would be if browsers offered `DeviceOrientationEvent` in both desktop and mobile modes. Not sure if there's a way to influence browser vendors to consider.

PlumCantaloupe:
Yeah I think so. 

Some food for  that thought: Maybe it didn't use to work but I did a few quick tests with Safari, Firefox, and Android Chrome and it appears that _DeviceOrientationEvent_ is available and allows "magic window" in desktop mode if we remove the isMobile() condition and desktop modal pop-up ...

kylebakerio:
Hey, pretty sure this is critical to get support for ipads... We have a project where we switch from 8th wall with aframe to pure aframe. On an ipad, the device motion stuff is used just fine by 8th wall, but as soon as we transition to vanilla aframe, we get a broken magic window.

In other words, this can be fixed, and without this, the default is indeed broken. 

I could see an argument be made for another modal that asks a tablet user if they want magic window or desktop experience? But as it is now, magic window being broken on a tablet is a pretty big bummer. Will try to use a hand-patched version of aframe from this pull request and see how it goes for now, but definitely would like to see something like this merged.

dmarcos:
@kylebakerio What part is broken? Did you set the browser to mobile mode? Just tested on my iPad and if the browser is set to desktop mode (default) the below message is displayed. When the browser is set to mobile then magic window works as expected. 

IMG_0015

Tested with https://aframe.io/aframe/examples/showcase/anime-UI/

and iPad with iPad OS 14.8

The PR seems to address Android tablets where the behavior might be different. Does the PR work for you on iPad?

kylebakerio:
That's correct, but when a website is designed for magic window, and an ipad user is not using their tablet as a desktop but just sitting around using it as a tablet, they just get the impression that the page doesn't work. Switching between the mobile and desktop version of a page is not something we can reliably expect users to know, and it seems like it makes more sense to default to using sensor data that is available rather than default to assuming users will be using a tablet as a desktop.

I had a client look at a project I built on their ipad and decalre it was broken. I eventually did the research, ending up here, and told him to try accessing it through mobile mode. Yes, that worked, but if he hadn't had tech support to call (me), he would just just left the page and assumed it didn't work.

kylebakerio:
I did check the pull request and notice it was anroid only, that's why I mentioned "something like this". It should be expanded to include ipad, imo.

dmarcos:
I know the situation is frustrating but it's Apple decision and policy. I think the case remaining the same. DeviceMotion (necessary for magic window) is not available in desktop mode which is now the default. Only the user can switch to mobile mode. Pages can only inform the user of the situation. Open to ideas but I don't think there's anything better that can be done at the moment.

PlumCantaloupe:
Just a note, noted above, that the DeviceOrientationEvent is available in Safari Desktop mode on mobile devices. Thus it seems that with an HTML canvas the desktop mode vs. mobile mode appears is merely semantic. Perhaps the modal should just ask if they want mobile or finger-drag look-controls.

In fact, some of my recent user-studies using mobile look-controls had some users wanting to be able to switch between gyro and finger-drag controls at will (novelty and intuitiveness versus comfort for longer sessions). I will hopefully attack such a variation of look-controls soon; but we would still need to figure out this mobile/desktop modal bit ...

kylebakerio:
https://github.com/aframevr/aframe/issues/4798#issuecomment-781761778

> One thing I am interesting in getting opinions about is whether it is even necessary to check if something "isMobile" or "isTablet" or "isDesktop" before checking for features such as DeviceOrientationEvent?

So, as mentioned here previously--is there a reason to only even hide the check if orientation is available behind `isMobile` ? One thing that does come to mind is that on iOS, to even request gyro sensors, I believe you need to tap to get a user interaction token allowing the iOS permissions request to show to the user. But a-frame already has a modal for this, I think it just needs to allow it to request on more devices.

An interesting question is how you know whether to make the request. I guess A-Frame was just using mobile mode as a shortcut for this, but the point is that this seems to be preventing access to gyro on devices that don't need to be prevented.

Because A-Frame already shows the 'request with mobile mode' modal, it sounds like that could just be extended to allow _trying_ to request gyro permissions. Certainly that would be an easy way to test. It could say 'request motion controls' vs. 'continue with desktop mode', basically (worded better).

cameronValtech:
@dmarcos The fixes in 5a21337 seem to work as expected on the Galaxy Tab, without my PR above. That said, I've improved the `isTablet` regex so we don't continue to misidentify on these devices.

At the very least can we create a new version tag so we can let npm do its thing?

dmarcos:
Magic window should be now working on master. I think this PR is no longer necessary. Thanks so much everybody for investigating the issue. I'm closing this but can reopen if necessary.

PR #4804, opened 2021-02-22
Update building-a-minecraft-demo.md

AriaMoKr:
Added scaled normal offset position to fix new voxel placement issue when clicking on the left and back faces of a voxel in Aincraft (new voxel gets same position of the voxel that was clicked on in those cases).
Also added pull request to intersection-spawn.js
https://github.com/supermedium/superframe/pull/288

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thanks! Congrats on your first contribution 🥇

PR #4809, opened 2021-02-26
fix: threejs reference

oneWaveAdrian:
**Description:** Updated threejs reference link

**Changes proposed:**
- reference link to threejs object3D was exchanged to the correct one


dmarcos:
Good catch! Thank you! Congrats on your first contribution 🥳

PR #4811, opened 2021-03-02
Changing permissions modal style to allow bigger messages

antoninklopp:
**Description:**
Changing the permission modal to allow users to add bigger messages. 

Today, width and height are hardcoded, which creates a strange effect when the message is too long. 
before-pc-alert-dialog

**Changes proposed:**
- Making dialogs flex
- Keeping today's dialog width and button height, the modal will only change height depending on the content size

I actually don't know how you would like to handle too long messages (maybe with a text overflow?, or let the user test the message size?). I'll leave this decision up to you. 

Here is a screenshot with the changes : 
pc-alert-dialog

You can also test the changes [with this glitch](https://aframe-modal-update.glitch.me/)

### DIFF
--- a/src/style/aframe.css
+++ b/src/style/aframe.css
@@ -308,34 +308,31 @@ a-scene audio {
 }
 
 .a-modal {
-  position: absolute;
+  display: flex; 
   background: rgba(0, 0, 0, 0.60);
   background-size: 50% 50%;
-  bottom: 0;
   font-size: 14px;
   font-weight: 600;
-  left: 0;
   line-height: 20px;
-  right: 0;
-  position: fixed;
-  top: 0;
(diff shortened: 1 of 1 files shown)


dmarcos:
Notice that you could also override the styles in your own CSS

antoninklopp:
Yes but I thought that maybe the change could benefit to other people having the same problem.

oneWaveAdrian:
I agree this is important to make multi language support easier. Plus it does not seem to have any disadvantages to users who just use the standard messages.

dmarcos:
We haven't seen demand for bigger modal messages. Closing this for now but can reevaluate in the future. Thanks

Issue #4810, opened 2021-02-28
Animations don't occur when the same animation is requested twice in a row.


diarmidmackenzie:
**Description:**

If you invoke the same animation twice in a row, the 2nd time, it does not occur, and does not trigger an animationcomplete event.

See example here:
https://successful-giddy-chime.glitch.me/

One use case this severely impacts is where an application tries to chain a series of animations together, using the animationcomplete event to trigger the next animation.  When the animationcomplete event does not occur, this leads to a complete blockage of the chain of animations.

The reason for the problem appears to be that when the value set on the animation component is unchanged, the update method of the animation component doesn't get called by A-Frame, so the animation component is never invoked at all for the duplicate animation request.

While in general it makes sense not to invoke update functions when properties are unschanged, in this specific case, it's perfectly reasonable for an application to want to trigger the same animation twice in a row, and so the animation component *should* be invoked in this case.

One workaround which I have tested, and works, is to add a tiny additional delta to the animation duration (I used a random float between 0 & 1).  While this has no practical impact on the animation duration, it's enough to cause A-Frame to call the animation component's update function, and everything seems fine from that point on.

- A-Frame Version: 1.2.0
- Platform / Device: All
- Reproducible Code Snippet or URL:
https://successful-giddy-chime.glitch.me/

It is possible that this issue was the cause of what was reported here: https://github.com/aframevr/aframe/issues/3806

However the code linked to in that issue is no longer available, so it's not possible to confirm that this was in fact the same problem.

kylebakerio:
Responded on slack, copying here:

kylebaker.io  5 minutes ago
I think I know why this is happening

kylebaker.io  5 minutes ago
basically, you're setting exactly the same values

kylebaker.io  5 minutes ago
when that happens, no update is performed, it doesn't run again

kylebaker.io  4 minutes ago
this is true of any component--the 'update' method is only called when the 'newData' isn't the same as the 'oldData' .

kylebaker.io  3 minutes ago
so if you add this line `          this.el.removeAttribute('animation')`
to the top of your animateRotation method, it works fine. See:
https://abstracted-perfect-sunset.glitch.me/

kylebaker.io  1 minute ago
this 'feature' of components helps them run performantly when they're getting dozens of update calls per frame, not updating when they get identical input. But they does mean that sending values to a component is not like calling a pure function and getting an idempotent result.

diarmidmackenzie:
Agree your workaround is better than mine & simple to apply.

[...]

PR #4812, opened 2021-03-02
Explain pitfalls with using setAttribute to trigger animations.

diarmidmackenzie:
**Description:**

As detailed (at some length) in this issue...
https://github.com/aframevr/aframe/issues/4810

**Changes proposed:**

Updates to animation documentation, to:
- correct default value for autoplay (docs are currently plain wrong here)
- to offer some examples of how animations can be started using setAttribute method calls, and highlight the potential pitfall with duplicate animations that can occur here.

### DIFF
--- a/docs/components/animation.md
+++ b/docs/components/animation.md
@@ -114,24 +114,7 @@ Accessed as `el.components.animation.<MEMBER>`.
 
 ### Controlling Animations using setAttribute
 
-Like any A-Frame component, the animation component can be configured from JavaScript by calling setAttribute() on an element.
-
-The second parameter can either be a string, or an object, detailing the properties to set.
-
-For example, either:
-
-```
-el.setAttribute('animation', 'property: rotation; from 0 0 0; to 0 180 0; dur: 1000');
-```
-
-or:
-
-```
-el.setAttribute('animation', {'property': 'rotation',
-                              'from': {'x' : 0, 'y' : 0, 'z': 0},
-                              'to': {'x' : 0, 'y' : 180, 'z': 0},
-                              'dur': 1000});
-```
+Like any A-Frame component, the animation component can be configured from JavaScript by calling [setAttribute()]( https://aframe.io/docs/1.2.0/introduction/javascript-events-dom-apis.html#updating-a-component-with-setattribute) on an element.
 
 By default, the animation will begin playing immediately (autoplay is true by default).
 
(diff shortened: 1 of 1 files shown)


docs/components/animation.md
```diff
@@ -112,6 +112,38 @@ Accessed as `el.components.animation.<MEMBER>`.
 | animation | anime.js object.           |
 | config    | Config passed to anime.js. |
 
+### Controlling Animations using setAttribute
+
+Like any A-Frame component, the animation component can be configured from JavaScript by calling setAttribute() on an element.
+
+The second parameter can either be a string, or an object, detailing the properties to set.
```
dmarcos, line 119: We should probably link to the relevant docs to avoid duplication https://aframe.io/docs/1.2.0/introduction/javascript-events-dom-apis.html#updating-a-component-with-setattribute

docs/components/animation.md
```diff
@@ -112,6 +112,38 @@ Accessed as `el.components.animation.<MEMBER>`.
...
+By default, the animation will begin playing immediately (autoplay is true by default).
+
+However, care should be taken when using the interface in this way.  If a finite (i.e. non-looping) animation is requested twice in a row, with identical parameters, the second animation will not play, and neither will it generate an animationcomplete event.
+
+This is because A-Frame sees the second request as a duplicate request for configuration already applied to the element, and so the second request never reaches the animation component.
+
+To ensure an animation is played, you can do either of the following:
+
+- explicitly remove any previous animation attribute using removeAttribute(), prior to calling setAttribute();
+- or start the animation using an event, rather than using autoplay (see the next section)
```
dmarcos, line 145: We can maybe [link the docs](https://aframe.io/docs/1.2.0/core/entity.html#emit-name-detail-bubbles) or add an example of explicitly triggering an animation via `el.emit`.

docs/components/animation.md
```diff
@@ -112,6 +112,38 @@ Accessed as `el.components.animation.<MEMBER>`.
...
+
+By default, the animation will begin playing immediately (autoplay is true by default).
+
+However, care should be taken when using the interface in this way.  If a finite (i.e. non-looping) animation is requested twice in a row, with identical parameters, the second animation will not play, and neither will it generate an animationcomplete event.
+
+This is because A-Frame sees the second request as a duplicate request for configuration already applied to the element, and so the second request never reaches the animation component.
+
+To ensure an animation is played, you can do either of the following:
+
+- explicitly remove any previous animation attribute using removeAttribute(), prior to calling setAttribute();
```
dmarcos, line 144: This would work but not sure if I recommend that way. Another alternative depending on the use case is having [multiple animations](https://aframe.io/docs/1.2.0/core/component.html#multiple) defined on the entity with different parameters and trigger them selectively via events

docs/components/animation.md
diarmidmackenzie: ok - fixed.

docs/components/animation.md
diarmidmackenzie: Agreed.  Done both.

diarmidmackenzie:
Sorry for slow reply on this one.  Blame David Goggins... :-)

Latest version should have addressed all your comments.

dmarcos:
Awesome! Thanks for the patience!

PR #4821, opened 2021-03-13
set sRGBEncoding for perspective screenshots

kfarr:
**Description:**
- set sRGBEncoding for perspective screenshots  (ctrl-alt-s)

**Screenshots**
Before
old-without-srgbencoding

After
with-srgbencoding


dmarcos:
Thanks! Only the changes in screenshot.js are necessary. Dist, version number, etc are generated automatically when we release a new version

kfarr:
Ok I'll resubmit with just the source change

Issue #4817, opened 2021-03-07
Screenshot created in 'perspective' projection mode is darker than expected


kfarr:
**Description:**
Using A-Frame's built-in keyboard activated screenshot the output of perspective screenshots is darker than expected. It is also darker than equirectangular screenshots from the same scene.
 
To reproduce:
- Go to https://kfarr.github.io/3dstreet/#https://streetmix.net/kfarr/3
- Press ctrl + alt + s
- Press ctrl + alt + shift + s
- Compare the resulting output

Suggested fix:
- Note that equirectangular projection uses sRGBEncoding encoding ([link](https://github.com/aframevr/aframe/blob/master/src/components/scene/screenshot.js#L155))
- I think you can insert `       encoding: THREE.sRGBEncoding,` [before line 93](https://github.com/aframevr/aframe/blob/master/src/components/scene/screenshot.js#L93)

Normal ticket questions:
- A-Frame Version: 1.1.0
- Platform / Device: Mac OS Chrome Desktop
- Reproducible Code Snippet or URL: See above

Images for comparison
![screenshot--1615104581133](https://user-images.githubusercontent.com/470477/110233616-e7fc3280-7ed9-11eb-897b-4886a3385944.png)
![screenshot--1615104250109](https://user-images.githubusercontent.com/470477/110233621-eaf72300-7ed9-11eb-89c4-e6f2f7c7c725.png)

dmarcos:
Nice thanks. If you have some bandwidth, a PR will be super appreciated

kfarr:
Yes can do as soon as I fix this which is blocking upgrade to latest a-frame on my project:
https://github.com/supermedium/superframe/issues/290

kfarr:
Ok just submitted a PR which works on my local machine but seems to trigger a build error on Travis CI
https://github.com/aframevr/aframe/pull/4821

The error is `    ✖ can use spherical env maps` but I don't understand how that test would be affected by the changes in the PR.

dmarcos:
Fixed by https://github.com/aframevr/aframe/commit/65f02cc81d8e73b6e36a41763a214ecacbb054c3

Tim-Laeufer:
In Version 1.3.0 I get this bug too.
In the following pictures there is a clear color difference. I'm not sure if it may be the same problem.

![screenshot-1657290602100](https://user-images.githubusercontent.com/89784833/178012842-b46cd873-99b6-43e6-86bc-9409db0979aa.png)

[...]

PR #4822, opened 2021-03-14
add sRGB encoding to perspective screenshot

kfarr:
**Description:** add sRGB encoding to perspective screenshot, fixing #4817

### DIFF
--- a/src/components/scene/screenshot.js
+++ b/src/components/scene/screenshot.js
@@ -85,6 +85,7 @@ module.exports.Component = registerComponent('screenshot', {
 
   getRenderTarget: function (width, height) {
     return new THREE.WebGLRenderTarget(width, height, {
+      encoding: THREE.sRGBEncoding,
       minFilter: THREE.LinearFilter,
       magFilter: THREE.LinearFilter,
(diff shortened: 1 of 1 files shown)


dmarcos:
I fixed the broken tests on master. If you rebase this is ready to merge. Thanks so much for the patience.

kfarr:
rebased and no worries, I'm always learning new things working with you and the community!

dmarcos:
Thank you!

diarmidmackenzie:
As per a thread in the slack community, I'm concerned that this fix has had a bad effect on colors in some screenshots.

This is what a screenshot of teh HelloWebVR page looked like in 1.2.0.
screenshot--1645741701810

This is what it looks like in 1.3.0.
screenshot--1645741354272

This glitch shows that if you revert the changes made in this PR, you get back to the previous coloring (i.e. this change is the cause of the difference).
https://glitch.com/edit/#!/sly-sixth-topaz?path=index.html%3A19%3A9

I'll raise a separate issue to track this, but thought it made sense to flag here as well.

kfarr:
Thanks, replied here: https://github.com/aframevr/aframe/issues/5014

PR #4823, opened 2021-03-15
Alow HP Reverb G2 Controllers to show up as Oculus Touch

zach-capalbo:
Allow iterating when setting up oculus-touch controllers. Allows ReverbG2 controllers to be found as Oculus touch controllers.

**Description:**

On Windows Chrome desktop, HP Reverb G2 controllers do not appear in experiences utilizing `tracked-controls` or derived components. 

The input source profiles supplied by Chrome are: "hp-mixed-reality", "oculus-touch", "generic-trigger-squeeze"

**Changes proposed:**
- Set `iterateControllerProfiles` to true by default in `tracked-controls` to allow for secondary profiles to be found at all
- Pass `iterateControllerProfiles` to `checkControllerPresentAndSetup` in `oculus-touch-controls` to allow HP Reverb G2 controllers to by initialized by their secondary profile.

I've tested these changes using the following workaround pasted into the developer console on several A-Frame experiences, including the "tracked controls showcase":

```
document.querySelectorAll('*[hand-controls]').forEach(el => {
    el.setAttribute('tracked-controls', 'iterateControllerProfiles', true); 
    AFRAME.utils.trackedControls.checkControllerPresentAndSetup(el.components['oculus-touch-controls'], 'oculus-touch', 
        {hand: el.getAttribute('hand-controls').hand, iterateControllerProfiles: true})
})
```

PR #4845, opened 2021-04-11
Add Reverb G2 Support

zach-capalbo:
**Description:**

Adds the Reverb G2 controller as a supported controller

From https://github.com/aframevr/aframe/pull/4823#issuecomment-799624075

- [X]  Copy magicleap-controls and rename to hp-mixed-reality-controls
- [X]  Replace controller id with hp-mixed-reality as indicated in webxr profile
- [X]  Point URL to appropriate model. Take models from the repo and submit to the A-Frame assets repo. Model scale adjustment might be necessary. I think we need to account for handennes. See oculus touch support for an example
    - Pending review: https://github.com/aframevr/assets/pull/39
- [x] Map buttons and axes according to the profile

Additional Changes Needed: 
- [x] Add to necessary tracked control locations to be auto-configured (laser controls, etc)
- [x] Adjust copy-pasted code for handedness
- [x] Adjust model offset and rotation to match physical orientation

### DIFF
--- a/src/components/hp-mixed-reality-controls.js
+++ b/src/components/hp-mixed-reality-controls.js
@@ -13,7 +13,7 @@ var GAMEPAD_ID_PREFIX = 'hp';
 var GAMEPAD_ID_SUFFIX = '-mixed-reality';
 var GAMEPAD_ID_COMPOSITE = GAMEPAD_ID_PREFIX + GAMEPAD_ID_SUFFIX;
 
-var HP_MIXEDL_REALITY_MODEL_GLB_BASE_URL = 'https://cdn.aframe.io/controllers/hp-mixed-reality/';
+var HP_MIXEDL_REALITY_MODEL_GLB_BASE_URL = 'https://cdn.aframe.io/controllers/hp/mixed-reality/';
 
 /**
  * Button IDs:
@@ -57,10 +57,10 @@ module.exports.Component = registerComponent('hp-mixed-reality-controls', {
     this.controllerPresent = false;
     this.lastControllerCheck = 0;
     this.onButtonChanged = bind(this.onButtonChanged, this);
-    this.onButtonDown = function (evt) { onButtonEvent(evt.detail.id, 'down', self); };
-    this.onButtonUp = function (evt) { onButtonEvent(evt.detail.id, 'up', self); };
-    this.onButtonTouchEnd = function (evt) { onButtonEvent(evt.detail.id, 'touchend', self); };
-    this.onButtonTouchStart = function (evt) { onButtonEvent(evt.detail.id, 'touchstart', self); };
+    this.onButtonDown = function (evt) { onButtonEvent(evt.detail.id, 'down', self, self.data.hand); };
+    this.onButtonUp = function (evt) { onButtonEvent(evt.detail.id, 'up', self, self.data.hand); };
+    this.onButtonTouchEnd = function (evt) { onButtonEvent(evt.detail.id, 'touchend', self, self.data.hand); };
+    this.onButtonTouchStart = function (evt) { onButtonEvent(evt.detail.id, 'touchstart', self, self.data.hand); };
     this.previousButtonValues = {};
     this.rendererSystem = this.el.sceneEl.systems.renderer;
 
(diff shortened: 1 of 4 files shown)


dmarcos:
Thanks so much. This is the correct approach! Well done. I left a comment in the PR of the models.

src/components/hp-mixed-reality-controls.js
```diff
@@ -0,0 +1,180 @@
...
+    buttons: ['trigger', 'grip', 'none', 'thumbstick', 'xbutton', 'ybutton']
+  },
+  right: {
+    axes: {touchpad: [2, 3]},
+    buttons: ['trigger', 'grip', 'none', 'thumbstick', 'abutton', 'bbutton']
+  }
+};
+
+/**
+ * Magic Leap Controls
```
dmarcos, line 41: Note. Update comment

src/components/hp-mixed-reality-controls.js
```diff
@@ -0,0 +1,180 @@
...
+
+  onControllersUpdate: function () {
+    // Note that due to gamepadconnected event propagation issues, we don't rely on events.
+    this.checkIfControllerPresent();
+  },
+
+  /**
+   * Rotate the trigger button based on how hard the trigger is pressed.
+   */
+  onButtonChanged: function (evt) {
```
dmarcos, line 156: Note. animate trigger and buttons on press if possible. If model is prepared for it.

src/components/hp-mixed-reality-controls.js
zach-capalbo: I've updated this comment.

src/components/hp-mixed-reality-controls.js
zach-capalbo: Would you be willing to accept the `hp-mixed-reality-controls` without button animations initially, and then add button animations at a later time? That way Reverb G2 users will at least have a usable experience, rather than being stuck unable to use most a-frame apps.

dmarcos:
Thanks for sticking with this. Much appreciated. Have you checked if virtual and real controllers align? When integrating new controllers I test alignment by wearing headset and peeking inside / outside quickly to make sure the virtual and real controllers align?

zach-capalbo:
> 
> 
> Thanks for sticking with this. Much appreciated. Have you checked if virtual and real controllers align? When integrating new controllers I test alignment by wearing headset and peeking inside / outside quickly to make sure the virtual and real controllers align?

Yes, the latest commit adds a position and rotation offset which aligns the model to both the WMR controller display and to the physical controller (at least within the margin of error allowed by the G2 tracking)

src/components/hp-mixed-reality-controls.js
```diff
@@ -0,0 +1,191 @@
...
+      model: this.data.model,
+      rayOrigin: new THREE.Vector3(0, 0, 0)
+    });
+  },
+
+  onAxisMoved: function (evt) {
+    emitIfAxesChanged(this, this.mapping.axes, evt);
+  },
+
+  updateModel: function (buttonName, evtName) {},
```
dmarcos, line 187: We can remove if not implemented

src/components/hp-mixed-reality-controls.js
```diff
@@ -0,0 +1,191 @@
...
+
+  setButtonColor: function (buttonName, color) {}
```
dmarcos, line 189: We can remove

src/components/hp-mixed-reality-controls.js
```diff
@@ -0,0 +1,191 @@
...
+var checkControllerPresentAndSetup = trackedControlsUtils.checkControllerPresentAndSetup;
+var emitIfAxesChanged = trackedControlsUtils.emitIfAxesChanged;
+var onButtonEvent = trackedControlsUtils.onButtonEvent;
+
+// See Profiles Registry:
+// https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/registry
+// TODO: Add a more robust system for deriving gamepad name.
+var GAMEPAD_ID_PREFIX = 'hp';
+var GAMEPAD_ID_SUFFIX = '-mixed-reality';
+var GAMEPAD_ID_COMPOSITE = GAMEPAD_ID_PREFIX + GAMEPAD_ID_SUFFIX;
```
dmarcos, line 15: Perhaps we don't need to have suffix / prefix on this one and can have just `var GAMEPAD_ID`

zach-capalbo: Collapsed this into `GAMEPAD_ID`

src/components/hp-mixed-reality-controls.js
zach-capalbo: Removed these placeholders

src/components/hp-mixed-reality-controls.js
zach-capalbo: Removed placeholders

dmarcos:
Thank you!

PR #4864, opened 2021-05-21
Use proper THREE.Cache API in a-assets.js

### LINKED ISSUE #4861: Is THREE.Cache handling in a-assets.js correct?
**Description:**

- A-Frame Version: v1.2.0

I found two direct `THREE.Cache` accesses in `a-assets.js`.

```javascript
// Set in cache because we won't be needing to call three.js loader if we have.
// a loaded media element.
THREE.Cache.files[imgEls[i].getAttribute('src')] = imgEl;
```

https://github.com/aframevr/aframe/blob/v1.2.0/src/core/a-assets.js#L43-L45

```javascript
// Set in cache because we won't be needing to call three.js loader if we have.
// a loaded media element.
// Store video elements only. three.js loader is used for audio elements.
// See assetParse too.
if (el.tagName === 'VIDEO') {
    THREE.Cache.files[el.getAttribute('src')] = el;
}
```

https://github.com/aframevr/aframe/blob/v1.2.0/src/core/a-assets.js#L163-L169

They give me two questions.

1. Why don't we use [`THREE.Cache.add(key, file)`](https://threejs.org/docs/#api/en/loaders/Cache.add) method? I think using `Cache.add()` method is more proper API rather than directly accessing `Cache.files`.

2. From the comment "Set in cache because we won't be needing to call three.js loader if we have." I supposed we have like the following code

```javascript
let el = THREE.Cache.get(url);
if (!el) {
  const loader = new THREE.FooLoader();
  el = await loader.load(url);
}
```

but I couldn't find such codes although I ran `$ find ./src -print0 | xargs -0 grep "THREE.Cache"`. So I'm wondering if these elements cache are working expectedly.

**Suggestion**

I would like to suggest to use `THREE.Cache.add()` method or simply we can remove the lines.

takahirox:
Fixes #4861

**Description:**

Refer to #4861 for the detail.

**Changes proposed:**
- Replace `THREE.Cache.files[url] = file` with likely more proper API `THREE.Cache.add(url, file)`.

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -42,7 +42,7 @@ module.exports = registerElement('a-assets', {
           loaded.push(new Promise(function (resolve, reject) {
             // Set in cache because we won't be needing to call three.js loader if we have.
             // a loaded media element.
-            THREE.Cache.files[imgEls[i].getAttribute('src')] = imgEl;
+            THREE.Cache.add(imgEls[i].getAttribute('src'), imgEl);
             imgEl.onload = resolve;
             imgEl.onerror = reject;
           }));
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #4874, opened 2021-06-21
docs: fix default values

oneWaveAdrian:
**Description:**
Edited values according to source code https://github.com/aframevr/aframe/blob/v1.2.0/src/extras/primitives/primitives/a-sky.js


dmarcos:
Thank you!

PR #4884, opened 2021-07-12
fix appearance of raycaster (fixes aframevr#4882)

Clicky02:
**Description:** Fixes issue aframevr#4882

**Changes proposed:**
- make unitLineEndVec3 in raycaster.js store the normalized direction instead of the normalized direction plus origin
- make the raycaster's line's endpoint be equal to unitLineEndVec3 times the length plus the origin

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -101,9 +101,9 @@ module.exports.Component = registerComponent('raycaster', {
     if (data.showLine &&
         (data.far !== oldData.far || data.origin !== oldData.origin ||
          data.direction !== oldData.direction || !oldData.showLine)) {
-      // Calculate unit vector for line direction. Can be multiplied via scalar to performantly
-      // adjust line length.
-      this.unitLineEndVec3.copy(data.origin).add(data.direction).normalize();
+      // Calculate unit vector for line direction. Can be multiplied via scalar and added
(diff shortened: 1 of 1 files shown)


mikemainguy (commented):
Can we get this into 1.4.2?  This is more prominent in new release.  I actually think only the second half is necessary (line 381).  I tested just that change locally and it fixes the problem for me.  I'm not sure about the changes on lines 104-106.

mikemainguy:
https://user-images.githubusercontent.com/554154/212378120-e477f493-69bf-4999-93f6-d778570ae8ed.mp4

Attached example of problem

Clicky02:
> Can we get this into 1.4.2? This is more prominent in new release. I actually think only the second half is necessary (line 381). I tested just that change locally and it fixes the problem for me. I'm not sure about the changes on lines 104-106.

Just want to add that I'm pretty sure both changes are necessary to fully fix the issue. Just making the second the second change will make it look better in certain scenarios (i.e. when the origin is close to <0,0,0> or when the origin is displaced in the raycaster's direction), but it won't fix it in all cases.

I'd also definitely like to see this get merged. I brought up this bug initially because the raycaster for the controller I was using would be noticeably inaccurate whenever it started intersecting with something close to it, which looks like the same thing you're experiencing.

PR #4886, opened 2021-07-14
Updated look-controls to work with AR (fixes aframevr#4885)

Clicky02:
**Description:** Updated the look-controls component so that it would update the position, rotation, and scale of the camera when in AR mode (see aframevr#4885).

**Changes proposed:**
-When the component checks to see if the scene is in vr-mode, the component will now check if it is in vr-mode or ar-mode.


dmarcos:
Thank you!

dmarcos:
Congrats on your first contribution 🎖️ 🥳 🍾

Clicky02:
Thanks!

PR #4890, opened 2021-07-19
Background lighting estimation

AdaRoseCannon:
**Description:**

This updates the environment mapping capabilities to include mapping the real world for Augmented Reality using the lighting estimation part of the lighting estimation API.

ezgif com-gif-maker (1)

Code example:

```
    <a-scene
       webxr="optionalFeatures: light-estimation;"
       background='directionalLightEl:a-light[type="directional"];probeLightEl:a-light[type="probe"];'>

		<a-light type="probe"></a-light>
		<a-light light="castShadow:true" type="directional" position="1 2 1"></a-light>

```

The background element will update `.environment` property of scene to be the real environment in AR just like it does for virtual environments already. It will update the two lights as well.

The code for this was adapted from:
https://storage.googleapis.com/chromium-webxr-test/r886480/proposals/lighting-estimation.html

*Notes:*

It may be unorthodox but I think it should create it's own lights so that users don't need to provide their own lights. In addition restoring the light to their original configuration upon leaving XR will be tricky and unreliable.

**Changes proposed:**
- [x] Generate the Environment Map in background
- [x] Add a new `"probe"` `light` type
- [x] Let this light type use predefined envMaps for Flat & VR
- [x] Use WebXR Ligthing estimation to update Probe lights
- [x] Use WebXR Ligthing estimation to update Directional lights
- [x] Create own Directional and Probe Lights
- [x] Turn off other scene lights

Update the docs for:
- [x] Background
- [x] a-light
- [x] light


AdaRoseCannon:
I've marked this as draft since the core functionality is in but this still needs some work to complete. I will aim to have it completed by the end of the week.

AdaRoseCannon:
I think I should let the developer provide a directional light which can be controlled that way they can still do shadow maps and the like.

src/components/scene/background.js
```diff
@@ -1,49 +1,229 @@
...
       scene.environment = this.cubeRenderTarget.texture;
+      this.needsEnvironmentUpdate = true;
     } else {
       scene.environment = null;
     }
+
+    if (this.updateInterval) {
+      clearInterval(this.updateInterval);
+    }
+    if (this.data.environmentUpdateFrequency > 0) {
```
dmarcos, line 120: We can do this in tick with [delta](https://aframe.io/docs/1.2.0/core/component.html#tick-time-timedelta) instead of using setInterval

src/components/scene/background.js
```diff
@@ -1,49 +1,229 @@
...
+
+      var gl = renderer.getContext();
+
+      this.needsLightProbeUpdate = false;
+      this.glBinding = new XRWebGLBinding(xrSession, gl);
+      gl.getExtension('EXT_sRGB');
+      gl.getExtension('OES_texture_half_float');
+
+      xrSession.requestLightProbe()
+        .then(function (lightProbe) {
```
dmarcos, line 182: We van maybe factor out some of this code into component methods, the tick function is getting a tad big

src/components/scene/background.js
```diff
@@ -1,49 +1,229 @@
...
+      ) {
+        self.startLightProbe();
+      }
+    });
+
+    this.el.sceneEl.addEventListener('exit-vr', function () {
+      self.stopLightProbe();
+    });
+
+    var directionalLight = document.createElement('a-light');
```
dmarcos, line 35: Should we make this conditional and only add this lights if / when background estimation is enabled to keep the scene clean?

dmarcos: Maybe add a method `setupProbeLights` that it's invoked from the `startLightProbe` method. It would be no-op if the lights have been already added.

dmarcos:
Cool stuff! Thanks so much. I left a few comments. Still understanding the code and approach.

AdaRoseCannon:
Thanks so much for the great comments!

AdaRoseCannon:
Btw if you want an easy way to test it here is the HTML file i've been using to test it:

```
<html>
  <head>
    <script src="./aframe/build/aframe.js"></script>
	<script src="https://unpkg.com/aframe-environment-component@1.2.0/dist/aframe-environment-component.min.js"></script>
	<script src="https://aframe.io/aframe/examples/js/hide-on-enter-ar.js"></script>
  </head>
  <body>
    <a-scene webxr="optionalFeatures: light-estimation;" background="directionalLight:a-light;">
		<a-camera position="0 1.6 2"></a-camera>

		<a-assets>
		  <a-cubemap id="pisa">
			<img src="https://threejs.org/examples/textures/cube/pisa/px.png">
			<img src="https://threejs.org/examples/textures/cube/pisa/nx.png">
			<img src="https://threejs.org/examples/textures/cube/pisa/py.png">
			<img src="https://threejs.org/examples/textures/cube/pisa/ny.png">
			<img src="https://threejs.org/examples/textures/cube/pisa/pz.png">
			<img src="https://threejs.org/examples/textures/cube/pisa/nz.png">
		  </a-cubemap>
		</a-assets>

		<a-light type="directional" light="shadowMap:true;" intensity="0.6" position="1 1 1"></a-light>
		<a-light type="probe" envMap="#pisa"></a-light>

		<!-- model by Snooze https://sketchfab.com/3d-models/low-poly-table-b940256ec5994e26a9e71289d1211b19 -->
		<a-entity shadow gltf-model="table.glb" scale="0.7 0.7 0.7" position="0 0 -1">
			<a-entity scale="0.3 0.3 0.3" position="0 1.05 0">
				<a-cylinder shadow position="1 0.75 1" radius="0.5" height="1.5" color="#FFC65D" material="roughness: 0.1; metalness: 0.2;"></a-cylinder>
				<a-box shadow position="-1 0.5 1" rotation="0 45 0" color="#4CC3D9" material="roughness: 0.1; metalness: 0.2;"></a-box>
				<a-sphere shadow position="0 1.25 -1" radius="1.25" color="#FFFFFF" material="roughness: 0; metalness: 0;"></a-sphere>
				<a-torus-knot position="0 3 0" material="metalness: 1; roughness: 0.17" geometry="radius: 0.45; radiusTubular: 0.09" animation__rotate="easing: linear; from: 0 0 0; loop: true; property: rotation; to: 0 0 360; dur: 3000;"></a-torus-knot>
			</a-entity>
		</a-entity>
		<a-entity hide-on-enter-ar position="-5 0 0" environment="lighting: false; skyType: gradient; skyColor: skyblue; preset: forest; shadow: true; lightPosition: 0 8.4 -0.2;"></a-entity>
    </a-scene>
  </body>
</html>
```

src/components/scene/background.js
AdaRoseCannon: Making a `setupProbeLights` method makes sense.

A lot of the setup is done in the tick function because the "enter-vr" event is often fired before the session is started.

I will probably move the things which are setup on the first tick after the session starts into a separate setupProbe method which will invoke setupProbeLights

AdaRoseCannon:
Your feedback cleaned it up a bit thanks!

I like having the creation of the lights together it annoyed me it was separate.

docs/components/background.md
```diff
@@ -13,22 +13,35 @@ frustum culling issues when `a-sky` is further than the far plane of the
...
-| color               | Color of the scene background.                            | black         |
-| transparent         | Background is transparent. The color property is ignored. | false         |
-| generateEnvironment | Whether to generate a default environment                 | true          |
+| Property                   | Description                                               | Default Value   |
+|----------------------------|-----------------------------------------------------------|-----------------|
+| color                      | Color of the scene background.                            | black           |
+| transparent                | Background is transparent. The color property is ignored. | false           |
+| generateEnvironment        | Whether to generate a default environment.                | true            |
+| environmentUpdateFrequency | How often to update the environment in seconds, 0 is off. | 0               |
+| sceneLights                | Query Selector for the lights to turn off when AR starts. | a-light,[light] |
```
dmarcos, line 46: Does this need to be exposed on the component? It looks like it should stay internal?

AdaRoseCannon: The only lights that you want to be turned off are the environment lights. Lights that are part of the scene such as a car's headights, a building's internal lights or some glowing features on a model should probably stay lit.

dmarcos: I see. In this case all the lights that use `a-light` and `light` component will be disabled. Some of them could belong to the scene. Maybe we just want to disabled the env lights that A-Frame adds by default?

AdaRoseCannon: I think lots of people use their own lights rather than relying on the default ones and this even works with the environment component. That someone will have models which include lights which need to be kept on is an edge case. 

Although I guess if a gltf model included lights they would be unaffected by this anyway since they won't get their own html element.

AdaRoseCannon: Oh!  Maybe it should only disable lights which effect infinite areas such as ambient, directional, probe and hemisphere and leave all the others? 

That should leave all Point and Spot lights which only do small areas and are most likely to be the lights you want to leave on?

AdaRoseCannon:
I took at a few alternatives to scene lights:

* No control, simpler but user has to manually turn off their own lights when they enter AR which will be a pain. They could use the `hide-in-ar` component but it's not a standard AFrame component, I could include it in this PR though.
* User defines lights to control on background element should work for almost all scenes without additional configuration. Downside is that user may need to make a complex selector to exclude lights
* Don't provide control but automatically turn off all lights except point and spot lights. This breaks if for some reason the developer is using a spot light to provide the main light for their scene. 

I'm going to go with the first one, it brings in a much needed component simplifies the code and configuration for this component it even works on the environment component!

AdaRoseCannon:
I think I prefer this it feels more robust and passes control to the user so shouldn't have surprise lighting edge cases.

src/components/scene/background.js
```diff
@@ -1,49 +1,212 @@
...
+    });
+  },
+  stopLightProbe: function () {
+    var data = this.data;
+    var scene = this.el.sceneEl.object3D;
+    this.xrLightProbe = null;
+
+    this.probeLight.components.light.light.intensity = 0;
+    if (this.__ownDirectionalLight) {
+      this.__ownDirectionalLight.components.light.light.intensity = 0;
```
dmarcos, line 43: can remove the double underscore `__` for consistency. Code base doesn't use variable naming for visibility

src/components/scene/background.js
```diff
@@ -1,49 +1,212 @@
...
+      this.data.environmentUpdateFrequency > 0 && // should update in general?
+      this.timeSinceUpdate > (this.data.environmentUpdateFrequency * 1000) // should update this tick?
+    ) {
+      this.timeSinceUpdate = 0;
+      this.needsEnvironmentUpdate = true;
+    }
+
+    // Update Light Probe
+    // source: view-source:https://storage.googleapis.com/chromium-webxr-test/r886480/proposals/lighting-estimation.html
+    if (frame && this.xrLightProbe) {
```
dmarcos, line 154: We can maybe factor out this part to an `updateLightProbe` method?

src/components/scene/background.js
```diff
@@ -1,49 +1,212 @@
...
+      // wait until the XR Session has started before trying to make
+      // the light probe
+      if (!frame) {
+        return;
+      }
+      this.needsLightProbeUpdate = false;
+      this.setupLightProbe();
+    }
+
+    if (this.xrLightProbe) {
```
dmarcos, line 195: Maybe an `updateCubemap` method?

dmarcos:
Thanks so much for the patience and sticking with it! It's pretty close. 

Do you think we can quickly add a simple example to illustrate?

AdaRoseCannon:
Shall I drop in this as an example? If #4892 gets merged it's a nice example of both light-estimation and hit-test and very simple.

https://ada.is/aframe-ar-boilerplate/index.html

dmarcos:
> Shall I drop in this as an example? If #4892 gets merged it's a nice example of both light-estimation and hit-test and very simple.
> 
> https://ada.is/aframe-ar-boilerplate/index.html

Yeah! It's really cool

docs/components/hide-on-enter-ar.md
```diff
@@ -0,0 +1,29 @@
...
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/hide-on-enter-ar.js
+examples: []
+---
+
+When the user enters AR this component will hide the component by toggling it's `visible` state.
+
+This is used to hide background elements such as floors, sky boxes, environments and other large elements. Letting the user comfortable fit the remaining visible elements into their physical space.
```
dmarcos, line 12: Nice description

AdaRoseCannon:
Feedback actioned

src/components/scene/background.js
dmarcos: I think this is still missing in the PR

dmarcos:
It would be great to incorporate the example to this PR if possible. Pretty close 👍 Thank you

src/components/scene/background.js
AdaRoseCannon: I took out the light updating logic and moved it to a function at the top I left the checks in place.

AdaRoseCannon:
Sure I'll do it now.

AdaRoseCannon:
I added the example I showed you earlier and manually included the hit-test stuff as a seperate JS file which can be removed if the hit-test PR gets merged.

The hit-test stuff is nice to have because it lets you move the model around to test the lighting.

src/components/scene/background.js
```diff
@@ -1,49 +1,222 @@
...
+      var estimate = frame.getLightEstimate(this.xrLightProbe);
+
+      if (estimate) {
+        updateLights(
+          estimate,
+          this.probeLight.components.light.light,
+          this.directionalLight.components.light.light
+        );
+      } else {
+        console.log('light estimate not yet available');
```
dmarcos, line 194: That's because the browser doesn't support the API or because an estimate has not yet been calculated?

AdaRoseCannon: It hasn't yet been calculated, I find I normally get this message for 6 frames and then it's working as expected.

dmarcos: I see. it looks that estimation has some sort of calibration / bootstrap process? Is it worth to display a message or is it noisy? Maybe a note for the developer would suffice? 

```
 // Light estimation might not be available: initialization, calculation time...
  if (estimate) {
         updateLights(
           estimate,
           this.probeLight.components.light.light,
           this.directionalLight.components.light.light
         );
       } 
```

AdaRoseCannon: Sure. No point adding unneeded console noise.

AdaRoseCannon:
I've removed the logs for the cubemap and the light estimation

I left comments behind so that future developers can see the situations which would cause them to fail:

lighting estimation warming up and cubemaps not being supported respectively.

src/components/scene/background.js
```diff
@@ -1,49 +1,219 @@
...
+      var estimate = frame.getLightEstimate(this.xrLightProbe);
+
+      if (estimate) {
+        updateLights(
+          estimate,
+          this.probeLight.components.light.light,
+          this.directionalLight.components.light.light
+        );
+      }
+      // else { light estimate not yet available, it takes a few frames to start working }
```
dmarcos, line 192: We can add that comment above

```
 // light estimate not yet available, it takes a few frames to start working
  if (estimate) {
         updateLights(
           estimate,
           this.probeLight.components.light.light,
           this.directionalLight.components.light.light
         );
       }
```

dmarcos:
Awesome work! Thanks so much! 🍾

PR #4892, opened 2021-07-27
AR hit test

AdaRoseCannon:
**Description:**

Add support for hit-test for Augmented Reality headsets and handsets. I haven't been able to test on real headset hardware so please test there if you can, it should work but I've only tested in an emulator.

This is dependent on #4890 to be merged first here is a link to the comparison between the branches:

https://github.com/AdaRoseCannon/aframe/compare/background-lighting-estimation...AdaRoseCannon:ar-hit-test?expand=1

**Changes proposed:**
- Add a new ar-hit-test component.

Additionally, if you want, I could add support for letting it do hit-testing in VR by doing raytracing against the scene to allow developers to build a uniform AR/VR experience but I am not sure how you'd feel about that.


dmarcos:
Awesome thanks! Once we merge #4890 I look into this.

dmarcos:
I merged #4890. I think needs a rebase.

src/shaders/shadow.js
```diff
@@ -0,0 +1,29 @@
+var registerShader = require('../core/shader').registerShader;
+var THREE = require('../lib/three');
+
+/**
+ * Flat shader using THREE.ShadowMaterial.
+ */
+module.exports.Shader = registerShader('shadow', {
```
dmarcos, line 7: Is this shader used?

AdaRoseCannon: I just updated the sample to include a ground shadow in AR. This sample uses the special shadow shader to accomplish this as it can't be done using regular materials.

dmarcos:
We should replace the existing [ar-hit-test component](https://github.com/aframevr/aframe/blob/master/examples/showcase/model-viewer/ar-hit-test.js) of the model viewer example with the new one

AdaRoseCannon:
Note to self: test how well the reticle auto-sizing works with a really off centre model.

AdaRoseCannon:
Note to self: Add anchor support

AdaRoseCannon:
Note to self: See if the hit-test bbox calculations can be leveraged to reposition the directional light in the helper

AdaRoseCannon:
The reticle texture is now automatically generated from the model being moved it only renders the texture when the model is changed

Here is a video of it in action: https://twitter.com/AdaRoseCannon/status/1422986340178747404

Needs:
- [ ] Cleaning up
- [ ] Options for whether to do this behaviour
- [ ] Controls for how much of the model to use

AdaRoseCannon:
Anchors are working well, I have found an issue with the Spinosaurus Demo where the rotation of the model is not preserved so the feet of the dinosaur don't match footprints going to identify next

AdaRoseCannon:
The issue was that the camera was rotated by 90 degrees which wasn't visible for my other model because it had 90 degrees rotational symmetry! 

The issue is fixed now: https://twitter.com/AdaRoseCannon/status/1425051509679661058

AdaRoseCannon:
I just updated the model-viewer example it works well.

AdaRoseCannon:
Because of the nested objects to support the rotation it wasn't generating the model footprints well so I set it to use the image version instead.

dmarcos:
It might need rebase. I think I see commits from previous [PR](https://github.com/aframevr/aframe/pull/4890)

AdaRoseCannon:
Rebased.

docs/components/ar-hit-test.md
```diff
@@ -0,0 +1,48 @@
...
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/hit-test.js
+examples: []
+---
+
+[hit-test]: https://immersive-web.github.io/hit-test/
+
+This component uses the WebXR hit-test API to position virtual objects in the real world. Remember to request the `hit-test` optional feature to allow it work.
```
dmarcos, line 12: Can we perhaps automate this? The `ar-hit-test` component adds `hit-test` to WebXR/optionalFeatures?

src/components/scene/ar-hit-test.js
```diff
@@ -37,6 +61,36 @@ HitTest.prototype.sessionStart = function sessionStart (options) {
   }
 };
 
+HitTest.prototype.anchorFromLastHitTestResult = function (options) {
+  var hitTest = this.lastHitTest;
+
+  if (!hitTest) {
```
dmarcos, line 67: Can do one liner `if (!hitTest) { return; }`

src/components/scene/ar-hit-test.js
```diff
@@ -37,6 +61,36 @@ HitTest.prototype.sessionStart = function sessionStart (options) {
...
+    return;
+  }
+
+  var object3D = options.object;
```
dmarcos, line 71: Can define top of the function for consistency

docs/components/ar-hit-test.md
```diff
@@ -0,0 +1,48 @@
...
+	<a-entity id="myobject"></a-entity>
+```
+
+## Properties
+
+| Property       | Description                                                  | Default Value |
+|----------------|--------------------------------------------------------------|---------------|
+| target         | The object to move around.                                   | null          |
+| enabled        | Whether to do hit-testing or not                             | true          |
+| map            | Image to use for the reticle                                 | ""            |
```
dmarcos, line 33: I believe this is called `src` in the code. Do you think we should expose it now or keep it non-configurable for simplicity until the need arise?

src/components/scene/ar-hit-test.js
```diff
@@ -70,16 +126,77 @@ HitTest.prototype.doHit = function doHit (frame) {
...
+    this.textureCache = new Map();
+
+    this.orthoCam = new THREE.OrthographicCamera();
+    this.orthoCam.layers.set(CAM_LAYER);
+    this.textureTarget = new THREE.WebGLRenderTarget(512, 512, {});
+    this.basicMaterial = new THREE.MeshBasicMaterial({
+      color: 0x000000
+    });
+    this.canvas = document.createElement('canvas');
+    this.ctx = this.canvas.getContext('2d');
```
dmarcos, line 193: call `context` instead of `ctx` for consistency.

src/components/scene/ar-hit-test.js
```diff
@@ -91,13 +208,19 @@ AFRAME.registerComponent('ar-hit-test', {
       this.hasPosedOnce = false;
       this.bboxMesh.visible = true;
 
+      if (!hitTestCache) {
```
dmarcos, line 211: Can do one liner `if (!hitTestCache) { hitTestCache = new Map(); }`

src/components/scene/ar-hit-test.js
```diff
@@ -37,6 +61,36 @@ HitTest.prototype.sessionStart = function sessionStart (options) {
   }
 };
 
+HitTest.prototype.anchorFromLastHitTestResult = function (options) {
```
dmarcos, line 64: We can perhaps make the parameters explicit instead of passing an object. It makes the function more legible / easier to use?

src/components/scene/ar-hit-test.js
```diff
@@ -37,6 +61,36 @@ HitTest.prototype.sessionStart = function sessionStart (options) {
...
+    if (entryObject === object3D) {
+      anchorToObject3D.delete(anchor);
+      anchor.delete();
+    }
+  });
+
+  if (hitTest.createAnchor) {
+    hitTest.createAnchor()
+    .then(function (anchor) {
+      anchorToObject3D.set(anchor, options);
```
dmarcos, line 86: Is better to use a map vs. a regular object here? object3D has an id that could be used as object key. I don't have much experience working with Maps.

src/components/scene/ar-hit-test.js
```diff
@@ -1,5 +1,29 @@
...
+  object3D.position.copy(pose.transform.position);
+  object3D.quaternion.copy(pose.transform.orientation);
+
+  tempVec3.copy(offset);
+  tempQuaternion.copy(pose.transform.orientation);
+  tempVec3.applyQuaternion(tempQuaternion);
+  object3D.position.sub(tempVec3);
+}
 
 function HitTest (renderer, options) {
```
dmarcos, line 28: Maybe better to replace options with explicit parameters for clarity. Also adding a description on top of HitTest to document what the object does would be useful

src/components/scene/ar-hit-test.js
```diff
@@ -1,5 +1,29 @@
...
+var tempVec3 = new THREE.Vector3();
+var tempQuaternion = new THREE.Quaternion();
+var fakePose = {
+  transform: {
+    orientation: new THREE.Quaternion(),
+    position: new THREE.Vector3()
+  }
+};
+
+function applyPose (pose, object3D, offset) {
```
dmarcos, line 18: maybe tmp variables can be wrapped in closure?

var applyPose = (function () {
  var tempVec3 = new THREE.Vector3();
  var tempQuaternion = new THREE.Quaternion();
  return function (pose, object3D, offset) {
    ....
  }
})()

src/components/scene/ar-hit-test.js
```diff
@@ -1,5 +1,29 @@
...
+var register = require('../../core/component').registerComponent;
+var THREE = require('../../lib/three');
+var CAM_LAYER = 21;
+var previousFrameAnchors = new Set();
```
dmarcos, line 6: Maybe previousFrameAnchors and anchorToObject3D can be part of `HitTest.prototype`?

src/components/scene/ar-hit-test.js
```diff
@@ -1,5 +1,29 @@
...
+};
+
+function applyPose (pose, object3D, offset) {
+  object3D.position.copy(pose.transform.position);
+  object3D.quaternion.copy(pose.transform.orientation);
+
+  tempVec3.copy(offset);
+  tempQuaternion.copy(pose.transform.orientation);
+  tempVec3.applyQuaternion(tempQuaternion);
+  object3D.position.sub(tempVec3);
```
dmarcos, line 25: Is `sub` more clear than `add` in this context? I usually associate `offset` with `addition` but don't fully understand yet the full context here.

dmarcos:
Looking good 👍 Left a few comments. Thanks so much

docs/components/ar-hit-test.md
AdaRoseCannon: Yes I was thinking that but no other components do it.

If we do this then I think also:
* `webxr` should add `DOM-Overlay` if required
* `background` should add lighting-estimation
* this should also add `anchors` as well as `hit-test`

is that okay to do in this PR?

docs/components/ar-hit-test.md
AdaRoseCannon: Nice catch i'll update the doc.

I would like to keep it in case people want to replace it with something branded or product specific such as an image with dimensions.

src/components/scene/ar-hit-test.js
AdaRoseCannon: Sure, this is a hangover from when I had destructuring before.

src/components/scene/ar-hit-test.js
AdaRoseCannon: We're already using Maps and they make more semantic sense in this case since it is how they are supposed to be used.

src/components/scene/ar-hit-test.js
AdaRoseCannon: Sure, this is a hangover from when I had destructuring before.

Describing HitTest is a good idea since it is a wrapper around the WebXR HitTest API to handle the difference between WebXR Input Types

src/components/scene/ar-hit-test.js
AdaRoseCannon: I think I am also using some of the temp variables elsewhere but this would be neater, i'll figure something out. Maybe just have 2 sets of temp variables for each use case.

src/components/scene/ar-hit-test.js
AdaRoseCannon: There are multiple HitTest objects, one for for each space used for casting hit-test rays i.e. viewer, transient input sources, left and right hand or controllers. This Set and Map are designed to be global so that each object has only a single associated anchor regardless of the HitTest result that was originally used to create it.

I could make it as a static property of HitTest i.e. `HitTest.previousFrameAnchors` but not on `HitTest.prototype`

src/components/scene/ar-hit-test.js
AdaRoseCannon: Ah i'm not adding an additional offset.

At the beginning I calculate the offset the object's position versus the center of the calculated reticle.
I subtract this offset when the user places the object so that it lines up optically with the reticle

AdaRoseCannon:
Thanks for the feedback, i'll fix most of those and left comments for the ones I think should stay.

I am on holiday for the next few days so i'll update the PR with fixes on Monday.

docs/components/ar-hit-test.md
dmarcos: Yeah I think it should work. the `init` method of `ar-hit-test` handles the setup

src/components/scene/ar-hit-test.js
dmarcos: Static property could work but probably better to use prototype. The Map will be shared across all instances too and can be referenced using `this`. A static property would have to be qualified `HitTest.previousFrameAnchors` e.g: 

```
 var Test = function () {};
 Test.prototype.previousFrameAnchors = new Set();
 var t1 = new Test();
 var t2 = new Test();
 t1.previousFrameAnchors === t2.previousFrameAnchors
```
> Test.kk = new Set();
< Set {} (0)
> t1.kk
< undefined
> var t2 = new Test();
< undefined

AdaRoseCannon: Oh I had never thought about having objects on the prototype like that before, that would work.

docs/components/ar-hit-test.md
AdaRoseCannon: What is the best way to do this?

```
var features = new Set(this.el.systems.webxr.data.optionalFeatures);
features.add('hit-test');
features.add('anchors');
this.el.setAttribute('webxr', 'optionalFeatures', Array.from())
```

Doesn't work, I think Systems don't handle the 3 argument setAttribute like components do.

IS it okay to update `this.el.systems.webxr.data.optionalFeatures` directly?

AdaRoseCannon:
Addressed all other comments, just need feedback [on this one](https://github.com/aframevr/aframe/pull/4892#discussion_r689523265) regarding the best way to update the webxr system.

docs/components/ar-hit-test.md
dmarcos: I don't love much the API we have know (array typed properties are rare) but you can do.

```
  var optionalFeatures = sceneEl.getAttribute('webxr').optionalFeatures;
  optionalFeatures.push('hit-test');
  sceneEl.setAttribute('webxr', optionalFeatures);
```

AdaRoseCannon: And that won't set the Overlay element and required features to null right?

AdaRoseCannon: It was setting the OverlayElement to null so I did the following to ensure all the settings were preserved.

```javascript
    // Update WebXR to support light-estimation
    var webxrData = this.el.getAttribute('webxr');
    var optionalFeaturesArray = webxrData.optionalFeatures;
    if (!optionalFeaturesArray.includes('light-estimation')) {
      optionalFeaturesArray.push('light-estimation');
      this.el.setAttribute('webxr', webxrData);
    }
```

AdaRoseCannon:
WebXR system now updates OptionalDependencies automatically which is a lot more convenient to users and something which has caught me out many times before.

dmarcos:
@AdaRoseCannon Thanks for the patience. Code looks good. I don't have an Android / AR phone to test at the moment. Is this ready to go?

AdaRoseCannon:
I think so, I am currently writing a more advanced demo using it thoroughly which will hopefully unearth anything weird but I think it's fine

AdaRoseCannon:
Found some issues regarding dynamically changing `ar-hit-test.target` fixing now

AdaRoseCannon:
It doesn't handle the case very well where the object you are trying to position is inside another object which has scale. I'll look into fixing it but it might be too flaky and might be easier to tell people to just not do that.

dmarcos:
Thanks so much! This is awesome. Love the work and patience with me 😄

PR #4900, opened 2021-08-11
Make Listening for Events section inclusive of custom controllers

evansjohnson:
**Description:**
As I understand, this section is really about listening for controller events, which won't necessarily be for buttons and axes. On an initial read through the docs, I had just read about creating custom controllers, so felt it may be more clear to make this section inclusive of the custom controllers just covered. Open to any suggested edits, I believe my statements are backed up from what I've seen so far but am generally unfamiliar with the library so need to be checked!

At a minimum, I'd like to fix up the "For each button, every time a button is pressed down, released, or for some cases, even touched." sentence fragment!

**Changes proposed:**
- Make section about listening to controller events, not specific to A-Frame controllers
- Link tracked-controls-component instead of relisting events

### DIFF
--- a/docs/introduction/interactions-and-controllers.md
+++ b/docs/introduction/interactions-and-controllers.md
@@ -501,13 +501,10 @@ A-Blast][a-blast].
 
 ## Listening for Controller Events
 
-The events emitted by the A-Frame and vendor specific controllers (e.g Quest,
-Index...) build on the base level events emitted by the
-[tracked-controls component]((#tracked-controls-component)) to the specific
-events of each controller. To handle these events, we look for the event name
-in the controller component's documentation pages at the event tables, then
-register event handlers how we want. Here are the events emitted by A-Frame's
-controller components:
+To handle controller events, we look for the event name in the controller
+component's documentation pages in the event tables, then register event
+handlers how we want. Here are the events emitted by A-Frame's controller
+components:
 
 - [daydream-controls events](../components/daydream-controls.md#events)
 - [gearvr-controls events](../components/gearvr-controls.md#events)
@@ -535,6 +532,10 @@ Then attach the component:
 ```html
 <a-entity oculus-touch-controls x-button-listener></a-entity>
 ```
+
+See [creating custom controllers](#creating-custom-controllers) for more
+details on how events are defined by building abstractions on the base level
+events emitted by the [tracked-controls component]((#tracked-controls-component)).
 
 ## Adding Laser Interactions for Controllers
 
--- a/docs/introduction/interactions-and-controllers.md
+++ b/docs/introduction/interactions-and-controllers.md
@@ -501,14 +501,13 @@ A-Blast][a-blast].
 
 ## Listening for Controller Events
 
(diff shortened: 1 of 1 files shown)


docs/introduction/interactions-and-controllers.md
```diff
@@ -499,14 +499,16 @@ For advanced examples on real applications, see the [paint-controls component
...
-
-Controllers have many buttons and emit many events. For each button, every time
-a button is pressed down, released, or for some cases, even touched. And for
-each axis (e.g., trackpad, thumbstick), an event is emitted every time it is
-touched. To handle buttons, look for the event name in the respective
-controller component documentation pages at the event tables, then register
-event handlers how we want:
+## Listening for Controller Events
+
+Controllers may have many buttons, many axes, and may emit many events. The
```
dmarcos, line 504: Thanks. Some alt rewording?

The events emitted by the vendor specific controllers (e.g Quest, Index...) map the lower events emitted by [tracked-controls component]((#tracked-controls-component)) to the specific buttons of each physical controller. To handle input events, we look for the event name...

evansjohnson: thanks for the suggestion! I think it really helped to break away from the original wording there. I did one more round, happy to keep iterating if something feels off.

Notes on changes to suggestion:
- this is for the A-Frame and vendor specific controllers
- when I see "map" I'm thinking about mapping over an array and it's not necessarily a 1:1 transformation so felt "build on the base level events" captured this a bit better
- "specific buttons" -> "specific events" - not limited to buttons
- "each physical controller" -> "each controller" - I believe this refers to the "controllers" earlier in the sentence which are the software a-frame controllers and may be levels of abstraction away from a physical controller.
- To handle input events -> To handle events - I would imagine controllers don't have to exclusively emit input events. Ex a controller "random-interval" that emits a signal at a configurable random interval to power some dynamic content.

evansjohnson:
thank you for your work with this library @dmarcos! it was really cool reading about how it builds on HTML and the event system already built into browsers

this is ready for another look if you get a chance and want to get it merged

docs/introduction/interactions-and-controllers.md
dmarcos: Not 100% if talking about `tracked controls` helps the avg user. It should be an implementation detail. Probably more clear to keep it high level. As you mentioned the section [create custom controllers](https://aframe.io/docs/1.2.0/introduction/interactions-and-controllers.html#creating-custom-controllers) exposes lower level detail. Maybe including a link to that section would be enough for those that want to dig deeper?

evansjohnson: good point, I updated to make it a footnote for the section. will let you decide if it's worth the backlinks, they do link to anchor tags on the same page so perhaps this is trying to do too much and is noisy but I found myself looking for this info when I read this section

dmarcos:
Not sure if the rewording adds much value. It omits details that were important for beginners. And custom controllers seems like and advanced option that doesn't come up very often. Thanks so much. I'm closing this but can reopen if necessary

PR #4901, opened 2021-08-11
Add some default styling to DOM-Overlay

AdaRoseCannon:
Styling DOM Overlay so that it is full screen and can still let events pass through to the scene underneath is really tricky.

This will add some default styling and a class name so that there are sensible defaults for the DOM Overlay element.


dmarcos:
Thank you!

PR #4906, opened 2021-08-18
Fix docs 'window.THREE' -> 'AFRAME.THREE'

jlecordier:
**Description:**

The doc ask to use `window.THREE` but I think it should be `AFRAME.THREE`

**Changes proposed:**

- Replace `window.THREE` by `AFRAME.THREE` in the doc


dmarcos:
using the global `window.THREE` or just `THREE` (window global is implicit). It's what examples and most of the community components use. Thanks so much for the suggestion though. Super appreciated.

PR #4910, opened 2021-08-24
[gltf-model] Support glTF+Meshopt (EXT_meshopt_compression)

### LINKED ISSUE #4909: Support for glTF+Meshopt assets ("EXT_meshopt_compression")
Would we want to support glTF models with Meshopt compression? See https://github.com/google/model-viewer/issues/1164 for some discussions of the benefits, and https://github.com/google/model-viewer/pull/2706 for an example implementation; mostly we just have to load a 6 KB decoder script (e.g. from unpkg) similar to what we do for the Draco decoder. I'd be happy to make a PR here if we're interested.

donmccurdy:
**Description:**

Adds support for glTF models compressed with `EXT_meshopt_compression` (e.g. by gltfpack or glTF-Transform tools).

**Changes proposed:**
- Adds `meshoptDecoderPath` property to `gltf-model` system.
- Fixes #4909

### DIFF
--- a/docs/components/gltf-model.md
+++ b/docs/components/gltf-model.md
@@ -104,35 +104,39 @@ If you want to use the animations from your glTF model, you can use the [animati
 <a-entity gltf-model="#monster" animation-mixer></a-entity>
 ```
 
-## Geometry compression with Draco
+## Using compression
 
 [draco]: https://google.github.io/draco/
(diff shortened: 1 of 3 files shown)


dmarcos:
Yeee! Thanks so much!

PR #4913, opened 2021-08-24
Fix anchors having multiple anchors per model and that the reticle is visible before tracking has started.

AdaRoseCannon:
Really sorry this is a last-minute fix for a few minor issues (and one major one, the anchors one)

**Description:**

Placing an object multiple times generates multiple anchors which is bad, this fixes that.

**Changes proposed:**
- Fir `ar-hit-test.js` to rectify that


dmarcos:
Thank you!

PR #4916, opened 2021-08-26
Fix ar-hit-test footprints

AdaRoseCannon:
**Description:**

AR Hit Test was trying to minimise the bounding box by removing rotations on the target object this led to odd behaviours unless under specific circumstances. This removes that which cleans up a reusable Quaternion which is nice.

It also makes the footprint render material double sided so that convex objects aren't invisible when the orthographic camera's clip plane sits inside them.

It also tweaks the rotation of the reticle mesh to fit the rotation of the orthographic camera so that it fits the content better. Previously it was rotated by 90 degrees. These weren't picked up in testing because the silhouette of the tested objects had 90 degrees rotational symmetry.

It now works well and tested with asymmetrical objects.

image

### DIFF
--- a/src/components/scene/ar-hit-test.js
+++ b/src/components/scene/ar-hit-test.js
@@ -199,7 +199,6 @@ HitTest.updateAnchorPoses = function (frame, refSpace) {
 };
 
 var hitTestCache;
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you very much!

PR #4921, opened 2021-09-03
fix hit-test error handling

AdaRoseCannon:
**Description:**

The current version of Edge on Hololens doesn't support hit-test which revealed some issues with the error handling in ar-hit-test

This was a hangover from when it was ported from modern JS to ES5.


dmarcos:
Thank you!

PR #4927, opened 2021-09-06
Disable XR when rendering reticle because the XR cameras ignore the layers property

AdaRoseCannon:
This fixes a bug where when the reticle tries to update in XR it includes unexpected objects because it is using the XR camera not the specified camera. The XR camera ignores layers.

### DIFF
--- a/src/components/scene/ar-hit-test.js
+++ b/src/components/scene/ar-hit-test.js
@@ -383,7 +383,7 @@ module.exports.Component = register('ar-hit-test', {
     var tempImageData;
     var renderer = this.el.sceneEl.renderer;
     var oldRenderTarget, oldBackground;
-    var isXR = renderer.xr.enabled;
+    var isXREnabled = renderer.xr.enabled;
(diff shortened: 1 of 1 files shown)


src/components/scene/ar-hit-test.js
```diff
@@ -383,6 +383,7 @@ module.exports.Component = register('ar-hit-test', {
     var tempImageData;
     var renderer = this.el.sceneEl.renderer;
     var oldRenderTarget, oldBackground;
+    var isXR = renderer.xr.enabled;
```
dmarcos, line 386: `var isXREnabled` for consistency

dmarcos:
Thanks. Just a small comment

AdaRoseCannon:
Thanks! Updated. Sorry for all these tiny PRs, I keep coming across tiny edge cases.

dmarcos:
> Thanks! Updated. Sorry for all these tiny PRs, I keep coming across tiny edge cases.

No problem at all! Tiny PRs are easy to review. Thanks for thoroughly testing the code.

PR #4931, opened 2021-09-20
Ar hit test fixes and docs

AdaRoseCannon:
**Description:**

* Fixes an issue in the texture not loading on the reticle if no target is selected
* Fixes an error in firing the event if no reticle is selected
* Fixes the wrong title and path in the ar-hit-test docs
* Adds the ability to set the reticle size


dmarcos:
Thanks!

PR #4932, opened 2021-09-21
Distortion map texture properties

AdaRoseCannon:
**Description:**

Distortion maps were using the default Wrapping so that when a normal map and a map were provided and there was repeating 
the texture would end up cut off e.g.:

image

**Changes proposed:**
- Move the material's setTextureProperties out of the system and into the utils
- use it to set the properties on the distortion maps too

image

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -10,6 +10,36 @@ var COLOR_MAPS = new Set([
 ]);
 
 /**
+ * Set texture properties such as repeat and offset.
+ *
+ * @param {object} data - With keys like `repeat`.
+*/
+module.exports.setTextureProperties = function (texture, data) {
+  var offset = data.offset || {x: 0, y: 0};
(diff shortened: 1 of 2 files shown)


src/utils/material.js
```diff
@@ -118,6 +148,9 @@ module.exports.updateDistortionMap = function (longType, shader, data) {
     if (texture && COLOR_MAPS.has(slot)) {
       rendererSystem.applyColorCorrection(texture);
     }
+    if (texture) {
+      module.exports.setTextureProperties(texture, data);
```
dmarcos, line 152: I think we can name the function above:

```module.exports.setTextureProperties = function setTextureProperties(texture, data) {```

and call it here:

```
 if (texture) { setTextureProperties(texture, data); }
```

A bit cleaner

dmarcos:
Thanks!

PR #4940, opened 2021-09-28
Samsung tablets were not recognized as tablets; updated user agent st…

cameronValtech:
**Description:**

Samsung tablets (at least the Galaxy Tab S7+) were being mis-identified as desktop browsers, thus displaying a user dialog to switch viewing modes. This is the same dialog seen in #4798 (see below). This PR includes Samsung-specific checks that are not addressed by  #4803

Screen Shot 2021-09-28 at 4 17 42 PM

**Changes proposed:**

- Update the `isTablet` regular expression to include user agent matching for Chrome and Samsung Browser on the Galaxy tablet
- Added relevant unit tests for Galaxy tablets

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -120,7 +120,7 @@ module.exports.isMobile = isMobile;
  */
 function isTablet (mockUserAgent) {
   var userAgent = mockUserAgent || window.navigator.userAgent;
-  return /ipad|Nexus (7|9)|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent);
+  return /ipad|Nexus (7|9)|xoom|sch-i800|SamsungBrowser|sm-[a-z]+[a-z0-9]+|playbook|tablet|kindle/i.test(userAgent);
 }
 module.exports.isTablet = isTablet;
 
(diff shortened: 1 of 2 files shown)


dmarcos:
Some tablets now default to desktop mode where DeviceMotion API might no be available. The user has to manually switch to mobile mode or magic window mode won't work. Have you test a master build without this patch?

https://cdn.jsdelivr.net/gh/aframevr/aframe@7ac01b7b61c2d8fc530a9dffb26e7dc0408915b4/dist/aframe-master.min.js

I made a couple of adjustments there. Safari for iOS now has DeviceMotion API available while in the desktop mode. I don't know if for your combo of tablet / browser is also the case.

cameronValtech:
Yes - I just tested your commit at master@HEAD / e0ed068 - and it actually looks like my issue in particular has been resolved (prior I was only looking at the most recent tag... whoops!). Feel free to reject this PR if you think it might be problematic.

dmarcos:
Cool thanks for giving it a try. I'm closing this because the deeper issue was addressed by https://github.com/aframevr/aframe/commit/5a2133799d58d8a26af8c8352dcd52dddae844d1

cameronValtech:
Do you think it's worth posting a 1.2.1 or 1.3.0 tag so npm users can make use a little bit easier?

dmarcos:
Yes we will release 1.3.0 soon but not ETA yet. Master needs QA. Feel free to use master in the meantime. Sorry for the inconvenience.

PR #4946, opened 2021-10-15
Fix environment generation

AdaRoseCannon:
**Description:**

Fixes the environment map generation from `background` due to changes in THREE.js 132

These new changes require `PMREMGenerator` which is slow so if it is triggered the reflections disappear for a short while before returning. So I removed the ability to run periodic updates of the reflections because it looks terrible because of this.

**Changes proposed:**
- Use PMREM Generator on environment cube maps
- Remove periodic background environment updates

### DIFF
--- a/src/components/scene/background.js
+++ b/src/components/scene/background.js
@@ -33,9 +33,10 @@ module.exports.Component = register('background', {
     var self = this;
 
     this.pmremGenerator = new THREE.PMREMGenerator(this.el.renderer);
+    this.pmremGenerator.compileCubemapShader();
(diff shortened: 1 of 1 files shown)


AdaRoseCannon:
This needs a rebase sorry I'll get on that

dmarcos:
Thanks for looking into this. This is ready to merge? You mentioned rebase. Just making sure there's no pending work. Thanks again

AdaRoseCannon:
This is ready to merge, sorry it doesn't fix lighting estimation yet that
is waiting on a change to THREE.

PR #4953, opened 2021-11-07
Fixes to raycasting examples.

diarmidmackenzie:
**Description:**

Alfonso complained on slack that the raycasting examples were broken.
https://aframevr.slack.com/archives/C0GG937RN/p1636245295015900

These fixes address the points that he raised.

**Changes proposed:**

index.html
- Replace use of raycast-helper component with showLine.  raycast-helper no longer works because it depends on THREE.Geometry (deprecated).  But it is no longer needed because raycast now supports raycast lines natively via the showLine attribute.

simple.html
- Add a brief description of what the expected behavior is.
- Added config so that the example doesn't generate warnings.
- Reduce the raycast timer so that performance is more responsive (making it less likely that someone will think it is not working).


dmarcos:
Thanks. We can probably remove the helper: 

https://github.com/aframevr/aframe/blob/master/examples/test/raycaster/components/raycaster-helper.js

dmarcos:
Thank you!

PR #4955, opened 2021-11-10
Update building-a-360-image-gallery.md

h-nasu:
id of the template should be "plane"

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Sorry for the delay. Good catch!

PR #4956, opened 2021-11-10
Update building-a-minecraft-demo.md

h-nasu:
**Description:**
I'm not sure which page was needed to refer for "Read more about using JavaScript and DOM APIs" but the previous one redirect to 404.

**Changes proposed:**
- changed link


dmarcos:
Great catch! Thanks! Congrats on your first contribution!

Issue #4965, opened 2021-11-17
Raycasting to objects that are not the closest


diarmidmackenzie:
**Description:**

I've been hitting various circumstances in which I want to access raycast entities that are not the *closest* to the camera.

Specific cases are:
- In VR, I have a control system where I can grab an object with the raycaster and move it around.  While grabbed, I want the raycast line to always point to the grabbed object, even when it passes behind a closer raycastable object.
- I have some small objects that appear always on top (using depthTest: false).  I want raycasting experience with these to match what the user sees, so when they can see an object, they can interact with it, even when it is further away than another object.
- Plus one more case involving a transparent plane that I want to be able to grab with a raycast, but also want to be able to raycast through.

So far I have been working around this in my application code:
- I have a customized version of "raycaster-line" that picks the correct item to draw a line to.
- I have code to grab objects that steps through the list of raycast matches to find the top priority one.

However I'm now needing to solve a similar problem in a situation where I use cursor.js.

I could make some modifications to cursor.js, but it occurred to me that a more elegant solution would be to allow raycaster to have a configurable object priority order to use, instead of distance order.

This would mean that I could define my precedence order in one place, and avoid having custom code in each of:
- raycaster-line
- cursor
- my own application code.

I have made some code changes to raycaster.js, and it's all working fine apart from one fix I needed to make to cursor as well.

PR to follow soon...

dmarcos:
> In VR, I have a control system where I can grab an object with the raycaster and move it around. While grabbed, I want the raycast line to always point to the grabbed object, even when it passes behind a closer raycastable object.

Have you played with the [objects](https://aframe.io/docs/1.2.0/components/raycaster.html#properties_objects) filter to determine the entities that get raycasted? Once you have grabbed the object you can change the filter to only track the entity of interest.

> I have some small objects that appear always on top (using depthTest: false). I want raycasting experience with these to match what the user sees, so when they can see an object, they can interact with it, even when it is further away than another object.

Similar the objects filter might be helpful in this case. You can perhaps add a class to those entities that have depthTest false? [Multiple raycasters](https://aframe.io/docs/1.2.0/core/component.html#multiple) to track different objects could also be an option I think and prioritize hits from one over the other.

> Plus one more case involving a transparent plane that I want to be able to grab with a raycast, but also want to be able to raycast through.

Iterating over the list of raycasted objects is not good enough? 

Better to make sure the current API cannot cover your use case before introducing changes.

[...]

PR #4966, opened 2021-11-17
Raycaster priorities

diarmidmackenzie:
**Description:**

This PR addresses use cases described here:
https://github.com/aframevr/aframe/issues/4965

**Changes proposed:**

- New property on raycast, which allows a user to specify a sortOrder to use, in place of the default distance order.  
- This sortOrder is provided in the form of another attribute on the objects, that takes values indicating the desired order.
- I've also had to make a fix to cursor.js to fully trust the order of objects it gets from raycast, rather than making its own decisions based on relative distances.  As far as I can tell the code I have removed here was superfluous.  However this is probably the biggest risk with the PR in terms of possible regressions.

For more details o how this works see the additions to the documentation, and the example included in the PR.


diarmidmackenzie:
Made a further fix to cursor.js, because raycaster was not behaving as expected.
https://github.com/aframevr/aframe/issues/4974

The problem scenario was when a small object was behind a large object of equal priority.  The small object was getting selected due to the fact that raycaster was not reporting the intersection with the larger object, since it was not new.

PR #4967, opened 2021-11-18
Fix blank scene after link navigation

### LINKED ISSUE #4964: Random blank scene after link navigation
When navigating between links in the link traversal showcase, sometimes the scene fails to load, showing a black scene instead. I get the following error in the developer console:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'parent')
    at WebGLRenderer.render (three.js:19705)
    at HTMLElement.value (a-scene.js:798)
    at bound (bind.js:12)
    at onAnimationFrame (three.js:18489)
    at XRSession.onAnimationFrame (three.js:9450)
```

I suspect that the scene is being rendered before the camera is attached. As a workaround, I added this condition to the a-scene render function:

```
// Don't render until we have a camera
if (this.camera !== undefined)
{
  renderer.render(this.object3D, this.camera);
}
```

This seemed to fix it, but I suspect there may be a better way of handling the underlying race condition. The scene can start rendering from two different places: after the 'loaded' event and after entering vr. When using the link navigation, it seems you can enter vr before the scene is loaded.

- A-Frame Version: 1.2.0
- Platform / Device: Oculus Quest 2 browser
- Reproducible Code Snippet or URL: examples/showcase/link-traversal/index.html

wtodd1:
The scene would sometimes fail to load after link navigation. If you enter
vr before the scene has loaded, it starts the render loop early, which can
cause an exception if the camera has not been injected yet.

Fixes #4964

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -378,7 +378,6 @@ module.exports.AScene = registerElement('a-scene', {
             requestFullscreen(self.canvas);
           }
 
-          self.renderer.setAnimationLoop(self.render);
           self.resize();
           if (resolve) { resolve(); }
         }


dmarcos:
Thanks.

dmarcos:
Congrats on your first contribution 🥳

PR #4980, opened 2021-12-11
New throttle function for the componentchanged event. 

### LINKED ISSUE #4972: Robustness issues with componentchanged event due to throttling
**Description:**

- A-Frame Version: 1.2.0.
- Platform / Device: All
- Reproducible Code Snippet or URL: None (can create if required).

In my application, I have some state that is (for usability reasons) controlled by multiple different mechanisms.  One of these is a toggle switch displayed in the VR scene.

The master record of the state is on an entity A, component C.

I want the toggle switch T (a distinct entity from A) to reflect the state of C.

To do this, T listens for the component changed event on A.  When this event fires, it checks the state of component C and updates the toggle switch state.

This works well most of the time.  However I have hit some issues due to throttling on componentchanged event.

There are cases where I want to reset things in my scene, and a convenient way to do this is to toggle the state of C off & on again in consecutive lines of code.

When I do this, the following happens:
- Turn C off
- componentchanged event triggers on T.  T updates to reflect off state.
- Turn C on again.
 - component changed event does not fire again, because it is throttled on a 200msec timer.  (see: https://github.com/aframevr/aframe/blob/b164623dfa0d2548158f4b7da06157497cd4ea29/src/core/component.js#L75).
 - Now T is stuck indefinitely, in an incorrect "off" state, not reflecting the master state of C.
 
 I recognize that there are good performance reasons for throttling componentchanged events.  However it seems that the consequence of this throttling is a lack of robustness in the componentchanged event, which means it can't be depended on.
 
My understanding is that the intent of componentchange is to provide a way for monitoring for changes on a component *without* having the cost of running a permanent tick to monitor it.  But if component changed can't be relied on to fire when a component changes, then it seems unusable for this kind of use case, and I'll end up having to revert to monitoring on a tick...
 
 It seems to me that a slightly more sophisticated form of throttling could solve this problem, while keeping the performance benefits of throttling:
 
When a component is changed within 200 msecs of a componentchanged event, rather than dropping further notifications, I suggest that A-Frame should instead set a flag indicating that a further component change has occurred, and then 200 msecs later, generate an additional componentchanged event (only one would be required, no matter how many changes there had been in the 200msec period).

I'm happy to help progress this further:
- I can create a simple glitch repro if needed.
- I'm happy to attempt to implement a fix.

Before that, though, I'd like to check whether there'd be interest in moving forward with a chage along these lines?

dmarcos:
One extra call might be issued at the end of the last interval to broadcast the last state of the component after a burst of component changes (fix #4972)

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -66,6 +66,41 @@ module.exports.throttle = function (functionToThrottle, minimumInterval, optiona
 
 /**
  * Returns throttle function that gets called at most once every interval.
+ * If there are multiple calls in the last interval we call the function one additional
+ * time.
+ * It behaves like a throttle except for the very last call that gets deferred until the end of the interval.
+ *
+ * @param {function} functionToThrottle
+ * @param {number} minimumInterval - Minimal interval between calls (milliseconds).
+ * @param {object} optionalContext - If given, bind function to throttle to this context.
+ * @returns {function} Throttled function.
+ */
+module.exports.throttleComponentChanged = function (functionToThrottle, minimumInterval, optionalContext) {
+  var lastTime;
+  var deferTimer;
+  if (optionalContext) {
+    functionToThrottle = module.exports.bind(functionToThrottle, optionalContext);
+  }
+  return function () {
+    var time = Date.now();
+    var sinceLastTime = typeof lastTime === 'undefined' ? minimumInterval : time - lastTime;
+    var args = arguments;
+    if (typeof lastTime === 'undefined' || sinceLastTime >= minimumInterval) {
+      clearTimeout(deferTimer);
+      lastTime = time;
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -64,6 +64,41 @@ module.exports.throttle = function (functionToThrottle, minimumInterval, optiona
   };
 };
 
+/**
+ * Returns throttle function that gets called at most once every interval.
+ * If there are multiple calls in the last interval we call the function one additional
+ * time.
+ * It behaves like a throttle except for the very last call that gets deffered until the end of the interval.
```
diarmidmackenzie, line 71: Typo in comment: deffered -> deferred.

PR #4983, opened 2022-01-03
Add user-controlled camera and canvas to cursor component.

diarmidmackenzie:
**Description:**

For an application I've been building, I have had to make some changes to the cursor component.  I believe these increase flexibility of the component and would be generally useful.

Where there are multiple cameras & multiple canvases used in an A-Frame scene, these changes allow cursor components to be used on both canvases.

See below for more details on the use case that drove this.

This PR is to yet ready to merge.  It needs further testing, and examples to illustrate how this can be used.  I'm raising it now to share the concept, and confirm that it's worth progressing with this testing and examples.

Feedback on whether this PR could be considered for A-Frame, once supporting examples are available, would be appreciated.

**Changes proposed:**

Changes are all to cursor component & accompanying docs.  In summary:
- new configuration options on the cursor component to determine whether or not user has control of camera & canvas.
- if not (auto), function is completely unchanged.
- if yes (user), the cursor component exposes the camera and canvas, the user can set them to whichever camera and canvas is desired, and the cursor component will use this camera & canvas in interpreting mouse movements.

**Use case that motivates this**

We are using A-Frame for a collaborative model viewer that supports both VR and desktop users.

For the desktop users, we want to offer a view selector widget in the corner, like this one in AutodeskMeshMixer.
image

This widget
- is 3D, so it's desirable to implement it using A-Frame
- has to be rendered using a different camera and canvas
- requires a cursor component, so that different sub-elements of the widget can be selected by mouse.

A-Frame's primary purpose is as a VR library, so we don't expect native support for multiple cameras, multiple canvases etc. (as per this discussion: https://github.com/aframevr/aframe/issues/916), and and we are building out our own code to handle these functions.

However, we want to use A-Frame as much as possible.

This PR doesn't add any new native functionality to cursor, so it is not extending the scope of A-Frame.  It simply refactors the code in a way that makes it much easier for an application to customize the function of cursor as needed (and does so with minimum risk to pre-existing function, by keeping the default behaviour identical).

I hope this wil be considered to be well-aligned with the A-Frame philosophy to provide a small common core of code focused on VR use cases, but provide users with the flexibility to extend functionality using custom THREE.js code as required.

** Examples**

I would like to provide an example to show the use of this in an application.  I think this be helpful both to appreciate the value of the PR, and to help future users to understand how they can take advantage of it.

I am working on getting permission to open-source the view-controller widget as an example.  However if I don't get agreement to do that, I'll build a simple demo app that illustrates the concept.

I don't think this PR should be merged until the example is in place.

However I'm raising the PR now, as I don't want to put more time into building such an example if there is actually zero chance of the PR being accepted.


src/components/cursor.js
```diff
@@ -196,14 +213,11 @@ module.exports.Component = registerComponent('cursor', {
   },
 
   onMouseMove: (function () {
-    var direction = new THREE.Vector3();
```
dmarcos, line 199: These variables not meant to store state but used for local calculations and cached to prevent unnecessary instantiations. Why do they need to be public now?

src/components/cursor.js
```diff
@@ -64,9 +66,18 @@ module.exports.Component = registerComponent('cursor', {
     this.canvasBounds = document.body.getBoundingClientRect();
     this.isCursorDown = false;
 
+    // expose camera and cursor to user if required.
+    if (this.data.camera === 'user') {
+      this.camera = this.el.sceneEl.camera;
+    }
+    if (this.data.canvas === 'user') {
```
dmarcos, line 73: This API is strange. Public variables are set or not depending on the value of a component property. Purpose is not clear either reading the schema or the docs. Probably feels to ad-hoc to the need of your use case.

src/components/cursor.js
dmarcos: The more variables exposed, the larger the API surface area, the greater maintenance burden and increased possibility of misuse.

diarmidmackenzie: They don't need to be public.  The issue is that they are not only used transiently, but actually passed by reference into the raycaster and used there.

[`this.el.setAttribute('raycaster', rayCasterConfig);`](https://github.com/aframevr/aframe/blob/5da874943f6aa871120f6b865fe9dcd7619e0cea/src/components/cursor.js#L236)

Result is that if you have two separate cursor components on the same scene, they share these variables, and end up scribblng over each others' raycaster parameters.  

It was a bit of a surprise to me that this happened, but it definitely did (took me quite a long time to figure out what was going on, as it was so unexpected - can create a repro if you want to take a look as exactly what happens here).

This was a simple solution that definitely works, because it gives you a unique instance of origin & direction per cursor instance.

I don't know of another way to solve this that doesn't have perf implications (e.g. copying the value of the origin & dircetion vectors to pass them into the raycaster).  Can you suggest a better way to solve this?

src/components/cursor.js
diarmidmackenzie: Yes, it is a bit ad hoc.

The reason for adding this to the interface was a desire to maintain complete back compatibility.

An example of a problem that could be caused by not putting this control on the interface is the following:

- An A-Frame app that switches between 2 different cameras, e.g. orthographic & perspective.
- With the existing cursor, cursor will auto-adapt to the current camera (as of[ this fix](https://github.com/aframevr/aframe/pull/4936)).
- After my change, it would become necessary to manually update the cursor's view of the camera when switching between ortho & perspective.
- I didn't want to risk breaking existing applications, hence the default setting of "auto" and the new mode needing to be explicitly activated in config.
- Not strictly necessary for 2 different parameters, but they are logically different functions so it seemed better.

In terms of the documentation, I think it would help a lot if I had an example to link to.  As per notes above, I'm hoping to re-use the component I have built as an example, but still TBD whether the owners of that code will be willing to open source it.

In short, I think a config option is necessary if we are to avoid breaking existing apps.  OTOH, the fix to make cursor work with an orthographic camera isn't even released yet (in master but not 1.2.0), so another option might be to simply document that fact that if you want to switch which camera is used for a cursor, you need to explicitly update `this.camera` in the cursor component.

diarmidmackenzie:
Stepping back, the fact you have reviewed the code suggests to me that if we can solve the issues you have flagged, you might be willing to take this into A-Frame.

Hence I'm thinking it would be worth pushing ahead with an example that illustrates how this can be used, to complete the PR.

Any comment on that?

src/components/cursor.js
diarmidmackenzie: Would giving them names that indicate they are internal, e.g. `this._orgin` and `this._direction` help?

Or can we make them private?  this.#origin, this.#direction?  (apparently this is new in ES2019, I only just found out about this option)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

diarmidmackenzie:
FInally got the OK to publish some examples of how this can be used.

In particular see:
https://diarmidmackenzie.github.io/aframe-multi-camera/examples/viewpoint-selector-basic.html
and
https://diarmidmackenzie.github.io/aframe-multi-camera/examples/multi-screen-with-cursor.html

dmarcos:
This feel a bit ad-hoc. Closing for now but can reopen if necessary. Thanks

Issue #3257, opened 2017-11-27
Material attributes related with map doesn't update with setAttribute


arturitu:
If you try to update offset or repeat parameter of a material don't update correctly.

- A-Frame Version: 0.7.0
- Platform / Device: All
- Reproducible Code Snippet or URL: https://spiffy-pear.glitch.me/

```

        self.el.setAttribute('material', 'offset', {
          x: 0,
          y: 0
        });
        self.el.setAttribute('material', 'repeat', {
          x: 1,
          y: 1
        });

        self.el.setAttribute('material', 'opacity', 1);

        // Remove this code comment to see how it works
        // var material = self.el.object3D.children[0].material;
        // material.map.offset.x = 0;
        // material.map.offset.y = 0;
        // material.map.repeat.x = 1;
        // material.map.repeat.y = 1;

```

Vamoss:
Did you find a solultion for this issue?

jan-thoma:
Can confirm this issue

hurda21:
Having the same issue. Any suggestions?

jan-thoma:
I can't find the issue in the source. I wrote a material component from scratch, maybe this helps.

````
AFRAME.registerComponent('mesh-standard-material',
{
    schema:
    {
        /* Shared */

        flatShading: {type: 'boolean', default: false},
        depthTest: {type: 'boolean', default: true},
        depthWrite: {type: 'boolean', default: true},
        wireframe: {type: 'boolean', default: false},
        side: {type: 'string', default: 'front'},
        fog: {type: 'boolean', default: true},
        opacity: {type: 'number', default: 1},

        /* Shader */

        color: {type: 'color', default: '#cccccc'},
        emissive: {type: 'color', default: '#000000'},
        metalness: {type: 'number', default: 0},
        roughness: {type: 'number', default: 0},

        /* Texture */

        map: {type: 'string'},
        mapRepeat: {type: 'vec2', default: {x: 1, y: 1}},
        mapOffset: {type: 'vec2', default: {x: 0, y: 0}},

        /* Normal map */

        normalMap: {type: 'string'},
        normalMapRepeat: {type: 'vec2', default: {x: 1, y: 1}},
        normalMapOffset: {type: 'vec2', default: {x: 0, y: 0}},
        normalScale: {type: 'vec2', default: {x: 1, y: 1}},

        /* displacementMap */

        displacementMap: {type: 'string'},
        displacementScale: {type: 'number', default: 1},
[...]

PR #4985, opened 2022-01-04
Proposed fix for issue 3257

diarmidmackenzie:
**Description:**

A potential fix for issue 3257, an old bug which came up again recently on the A-Frame slack.

As per @donmccurdy's comment, the problem is with this line of code:
https://github.com/aframevr/aframe/issues/3257#issuecomment-374825324

https://github.com/aframevr/aframe/blob/d27fd57e23cd14de9fb5400542d42590dbe02433/src/utils/material.js#L70
// Don't process if material src hasn't changed.
if (src === shader.materialSrcs[materialName]) { return; }

This glitch shows that removing this line appears to solve the issue, therefore I am proposing we simply remove this line, to fix this problem.
https://glitch.com/edit/#!/texture-offset-with-aframe-fix?path=index.html%3A9%3A59

My concern is whether there are circumstances in which this line of code was delivering valuable optimization.

My expectation is that the material update() processing won't be invoked unless there has been a change to attributes, and therefore there will be few (if any) situations where that line of code was actually doing the right thing.

It should be possible to write a more elaborate fix where we explicitly check for changes to the various parameters (offset, repeat etc.), and return if none of them have been changed.  However I believe that code is superfluous and would create a maintenance over head.

Hence proposing the simple fix, but happy to implement a more elaborate fix if there are good reasons to do so.


diarmidmackenzie:
Hmm.  Looking at this a bit more, it seems like a better fix might be to call

setTextureProperties()

in this case....

That ought to achieve what needs doing, but without the costs of reloading the texture...

diarmidmackenzie:
Based on the suggestion above, I think I have improved the fix.  Still seems to work ion the problem case we had, and should avoid unecessary repeat loading of textures.

dmarcos:
Not necessary to commit `dist` and index.js files

dmarcos:
Thanks! I think it looks reasonable. I was also concerned with doing extra work when unnecessary.

Perhaps a further improvement would be to [check if the values of the texture](https://github.com/aframevr/aframe/blob/master/src/utils/material.js#L17) have changed before setting them again

dmarcos:
If we remove changes in dist / index file this one is pretty close. Thanks

diarmidmackenzie:
I probably didn't do that in the most efficient way, but I think I got there.  Changes to index.js and dist files now removed from the PR.

diarmidmackenzie:
I didn't understand the intent of this comment: 
"Perhaps a further improvement would be to check if the values of the texture have changed before setting them again"

The things we are setting here are just vec2s?  Doesn't seem to deliver any real efficiency gain to check for diffs before setting vs. just setting the new values...?

dmarcos:
Thanks for the patience! Great effort!

PR #4986, opened 2022-01-06
Add info about binding event listeners (docs update)

diarmidmackenzie:
**Description:**

I answered a question on A-Frame slack where someone hadn't realized they needed to bind event listeners in order to access this inside them.

I looked for an explanation of this in the docs, but there wasn't one.  I guess I figured this out from examples + generic JS documentation on `this`.  But it's a pretty fundamental to using event listeners in A-Frame, and I think deserves some explanation in the core docs.

**Changes proposed:**

- Added a section to the docs intro to event listeners explaining that `this` is not accessible by default, and how to get access to it.

### DIFF
--- a/docs/introduction/javascript-events-dom-apis.md
+++ b/docs/introduction/javascript-events-dom-apis.md
@@ -502,6 +502,37 @@ entityEl.addEventListener('physicscollided', collisionHandler);
 entityEl.removeEventListener('physicscollided', collisionHandler);
 ```
 
+### Binding Event Listeners
+
+By default, event listeners do not have access to the component's `this`.  Instead, `this` will be bound to `window`, so you won't see errors if you reference `this` inside an event Listener, but you are likely to get unexpected results.
+
+In order for the component's `this` to be accessible inside an event listener, [it must be bound](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this).
+
+There are several ways you can do this:
+
+1. By using an arrow function to define the event listener.  Arrow functions automatically bind `this`.
+
+```entityEl.addEventListener('physicscollided', function (event) {
+entityEl.addEventListener('click', (event) => {
+  console.log(this.el.id);
+});
+```
+
+2. By binding the function as you add the event listener.  You will also need to do the same when [removing the event listener.](#removing-an-event-listener-with-removeeventlistener)
+
+```entityEl.addEventListener('physicscollided', (event) => {
+entityEl.addEventListener('click', this.clickListener.bind(this));
+```
+
+3. By creating another function, which is the bound version of the function.
+
+   ```
+   this.listeners = {
+     clickListener: this.clickListener.bind(this);
+   }
+   entityEl.addEventListener('click', this.listeners.clickListener);
+   ```
+
 ## Caveats
 
 [faq]: ./faq.md#why-is-the-html-not-updating-when-i-check-the-browser-inspector
--- a/docs/introduction/javascript-events-dom-apis.md
+++ b/docs/introduction/javascript-events-dom-apis.md
@@ -504,7 +504,7 @@ entityEl.removeEventListener('physicscollided', collisionHandler);
 
 ### Binding Event Listeners
 
-By default, event listeners do not have access to the component's `this`.  Instead, `this` will be bound to `window`, so you won't see errors if you reference `this` inside an event Listener, but you are likely to get unexpected results.
+By default, [Javascript execution context rules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) binds `this` to the global context (`window`) for any independent function, meaning that these functions won't have access to the component's `this` by default.
 
 In order for the component's `this` to be accessible inside an event listener, [it must be bound](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this).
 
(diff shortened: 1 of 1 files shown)


dmarcos:
As you mentioned this is JS general knowledge. Not sure how much the A-Frame docs should teach JS vs being focus on A-Frame itself.

dmarcos:
This info could be perhaps collected in Stack Overflow. One can open and self answer questions.

dmarcos:
Ok. I see this is in the JS / DOM section. Concern is that A-Frame docs become a JS tutorial that I think is out of scope. We can merge this. Looking forward, we should probably reemphasize that knowledge of HTML / JS / DOM are prerequisites to use A-Farme.

docs/introduction/javascript-events-dom-apis.md
```diff
@@ -502,6 +502,41 @@ entityEl.addEventListener('physicscollided', collisionHandler);
 entityEl.removeEventListener('physicscollided', collisionHandler);
 ```
 
+### Binding Event Listeners
+
+By default, event listeners do not have access to the component's `this`.  Instead, `this` will be bound to `window`, so you won't see errors if you reference `this` inside an event Listener, but you are likely to get unexpected results.
```
dmarcos, line 507: It's not specific to event listeners but applies to any function. We should probably start with that. Maybe:

By default, [Javascript execution context rules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) binds `this` to the global context (`window`) for any independent function. To access the component object from an event listener there are several options:

diarmidmackenzie: Updated with your suggestion.

docs/introduction/javascript-events-dom-apis.md
```diff
@@ -502,6 +502,41 @@ entityEl.addEventListener('physicscollided', collisionHandler);
...
+
+By default, [Javascript execution context rules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) binds `this` to the global context (`window`) for any independent function, meaning that these functions won't have access to the component's `this` by default.
+
+In order for the component's `this` to be accessible inside an event listener, [it must be bound](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this).
+
+There are several ways you can do this:
+
+1. By using an arrow function to define the event listener.  Arrow functions automatically bind `this`.
+
+```entityEl.addEventListener('physicscollided', function (event) {
```
dmarcos, line 515: This example might be a bit convoluted. Attaching an event listener within an event listener. Perhaps we should simplify or add more context of what this example illustrates. I myself got confused.

docs/introduction/javascript-events-dom-apis.md
```diff
@@ -502,6 +502,41 @@ entityEl.addEventListener('physicscollided', collisionHandler);
...
+entityEl.addEventListener('click', (event) => {
+  console.log(this.el.id);
+});
+```
+
+2. By defining your event listener within the events object of the component (this will also handling adding and removing the listener automatically)
+
+   See the explanation [here](../core/component.html#events).
+
+3. By binding the function as you add the event listener.  You will also need to do the same when [removing the event listener.](#removing-an-event-listener-with-removeeventlistener)
```
dmarcos, line 525: Are 2, 3 and 4 redundant? `explicit binding`

dmarcos:
Added some comments on the examples. Adding event listener within an event listener is perhaps a rare scenario that might obscure what the docs are trying to teach: `this` pointer and how to bind it to functions / methods.

docs/introduction/javascript-events-dom-apis.md
diarmidmackenzie: I don't understand why you describe this as "adding an event listener within an event listener".

This just shows addign an event listener that makes use of "this" inside the event listener (makign a console log, which is about the simplest thing I could think to do.

Maybe `this.el.addEventListener` is beter than `entityEl.addEventListener` as `this.el` is always defined and it may not be obvious what `entityEl` is...?

docs/introduction/javascript-events-dom-apis.md
diarmidmackenzie: I don't think 2 is redundant.  It's a good solution - probably the most efficient overall (i.e. least code to write), but because it obscures what's going on under the covers, I didn't think it should be the only solution offered.

3 & 4 are both approaches I have seen used very widely, but arguably there is very little difference between them.  I'd be quite happy to drop one or other of these.  I presonally usually use 3 as it's more compact, but have seeen a lot of examples that use 4.

I think it is good to show both `bind(this)` and arrow functions.

docs/introduction/javascript-events-dom-apis.md
diarmidmackenzie: RIght - have been looking at / editing this in Typora, where it looked like this:

image

Just looked at this in "source code" mode, and I see there is a load of extra code in there that definitely shouldn't be there.  (looks like this is because I had failed to include a newline after the ``` in MD).

Have fixed this up...

docs/introduction/javascript-events-dom-apis.md
dmarcos: Yeah 3 seems too specific to a use case. I would remove. Thanks

diarmidmackenzie: Removed #3 and renumbered # 4 to #3.

dmarcos:
Thanks so much for the patience! 🥇

PR #4991, opened 2022-01-14
Bug fixes for WebXR layers:

snagy:
Request "local-floor" reference space to match three.js default.
Init loading screen images for quad layer
Check mono layout instead of monocubemap to ensure we ask for the proper subimage type.

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -60,6 +60,8 @@ module.exports.Component = registerComponent('layer', {
 
     this.visibilityChanged = false;
     if (!this.layer) { return; }
+    if (type !== 'monocubemap' && type !== 'stereocubemap') { return; }
+
     if (!src.complete) {
       this.pendingCubeMapUpdate = true;
     } else {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks so much! I just made a comment to clarify

src/components/layer.js
```diff
@@ -73,7 +72,7 @@ module.exports.Component = registerComponent('layer', {
       this.loadingScreen = false;
     }
 
-    if (type === 'monocubemap') {
+    if (this.layer.layout === 'mono') {
```
dmarcos, line 75: Was this triggering a bug? Trying to understand because the layer type in the component properties should correspond to `this.layer.layout === 'mono'`. They should be equivalent or we're doing something wrong somewhere else.

snagy: I would occasionally see it get into loadCubeMapImages() from the tick function, and it would wind up there with the type of 'quad', which would have a mono layout.   It might be worth looking at the overall flow to preventing it from getting there, but this change fixed the error being thrown because of that.

dmarcos: I see that function shouldn't run for a `quad` layer type. I added a commit with an alternative logic. Could you verify that also fixes the issue? Sorry to add extra work. Not with my Quest at the moment. Thanks so much!

snagy (commented):
Tested on my Quest 2, looks good.

dmarcos:
Thank you!

Issue #4992, opened 2022-01-14
Refactor reflection logic out of background component


dmarcos:
Previous conversation https://github.com/aframevr/aframe/issues/4989

dmarcos:
fixed by https://github.com/aframevr/aframe/pull/4995

PR #4995, opened 2022-01-20
Reflection refactor 

AdaRoseCannon:
**Description:**

This brings out the environment generation from background into it's own file/component so that developers can turn on reflection as needed by adding the reflection component to <a-scene>.

**Changes proposed:**
- Remove environment map functionality from `background`
- Create new environment map component
- Fix lighting estimation from the new version three (can't get working yet)

### DIFF
--- a/docs/components/reflection.md
+++ b/docs/components/reflection.md
@@ -7,10 +7,12 @@ source_code: src/components/reflection.js
 examples: []
 ---
 
-The reflection component can also generate a default environment cube map for all
+The reflection component generates a default environment cube map for all
 materials, this is useful in case you find GLB models end up too dark or reflective
 materials don't look right because they are not reflecting the environment this
 will provide a default reflective environment.
+
+![Left hand side has objects with no reflection, the objects on the right reflect the environment](https://user-images.githubusercontent.com/4225330/151032019-1f14a079-604a-4c5f-b377-ea30a4e2b098.png)
 
 ## Scene Lighting and Lighting Estimation
 
--- a/docs/components/reflection.md
+++ b/docs/components/reflection.md
@@ -1,0 +1,39 @@
+---
+title: reflection
+type: components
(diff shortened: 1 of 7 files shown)


AdaRoseCannon:
Fixes #4992

AdaRoseCannon:
I think there is a bug in THREE preventing this from working https://discourse.threejs.org/t/updating-a-textures-webgltexture-works-for-normal-textures-but-not-for-scene-environment/33948

AdaRoseCannon:
I'm now going to make the docs for it.

AdaRoseCannon:
Documentation added. Ready for merging unless you find issues.

docs/components/reflection.md
```diff
@@ -0,0 +1,39 @@
+---
+title: reflection
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/reflection.js
+examples: []
+---
+
+The reflection component can also generate a default environment cube map for all
```
dmarcos, line 10: `can also` or `The reflection component generates...`?

docs/components/reflection.md
```diff
@@ -0,0 +1,39 @@
...
+materials, this is useful in case you find GLB models end up too dark or reflective
+materials don't look right because they are not reflecting the environment this
+will provide a default reflective environment.
```
dmarcos, line 13: Maybe a `before / after` screenshot here will help illustrate

dmarcos:
This looks great. Thanks so much. Just a few comments about the docs. Having background and reflection in separate components is so much cleaner! 👍

AdaRoseCannon:
Hmm i've noticed it's not getting all of the detail, the Octohedron on the left is what reflections should look like but they actually look like the sphere on the right, i'm looking into it further

Screenshot_20220125-104510_Samsung Internet Beta

PR #5001, opened 2022-01-26
Add the phong shader to allow performant materials

AdaRoseCannon:
**Description:**

Add the THREE.js Phong material to let developers use a more perfomant material for large non-flat objects.

**Changes proposed:**
- Fixes an issue in `layers.js` that was preventing it from being merged
- Fixes a bug in reflection.js where there was a circular texture dependency
- Phong material code based on standard material but tweaked for phong parameters
- Phong envMap defaults to a getter to the scene.environment because phong materials don't automatically use the scene.environment

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -147,6 +147,30 @@ For example, for a tree bark material, as an estimation, we might set:
 </a-entity>
 ```
 
+#### Phong-Based Shading
+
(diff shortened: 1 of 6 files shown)


AdaRoseCannon:
This has the weird side effect of actually handling the environment maps from the lighting estimation API better than the standard material

dmarcos:
Thank you!

PR #5005, opened 2022-02-02
don't try to use Hit Test in VR

AdaRoseCannon:
**Description:**

When the AR Hit Test component is used and the user tries to enter VR on a device using the WebXR Emulator it get stuck in a loop.

**Changes proposed:**
- Don't try to start the Hit-Test functionality in a non-ar session


dmarcos:
Thanks so much!

Issue #4973, opened 2021-11-29
Silent failure where components not initialized if they are modified immediately after object creation.


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.2.0
- Platform / Device: Chrome browser
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/modifying-attributes-after-creation?path=index.html%3A34%3A7

A problem that I have hit, which may be due to A-Frame restrictions around entity creation, but it would be good to at least get a clear error message when making unsupported API calls, rather than (as currently happens) getting a silent failure to initialize the component.

I have a very simple bit of code that is creating an entity and applying a component to it.  However, after appending the entity to the scene (before it has initialized), it may further modify the attributes on the entity, as per the example below.

(this example is simplified: in my case the last 2 lines of code occur in a different context, where it's less obvious the component may still be initializing).

When it does this, no error messages are generated, but the component initialization never actually happens at all.

```
          const box = document.createElement('a-box');
          box.setAttribute('make-red', '');
          this.el.appendChild(box);
          box.removeAttribute('make-red');
          box.setAttribute('make-red', '');
```

It would be ideal if these further attributes updates could be processed and applied when the component initializes.  However I recognize this may not be possible.

I note the docs here do say that operations after appendChild() and before initialization completes, are limited:
https://aframe.io/docs/1.2.0/introduction/javascript-events-dom-apis.html#adding-an-entity-with-appendchild

> "we can’t do many operations on the entity (such as calling .getAttribute()"

Assuming that setAttribute() is forbidden as well as getAttribute(), it would be great to do a couple of things:

- Could we generate an explicit error message on this setAttribute() call, rather than just silently failing?  That would have helped me to track down the problem a lot faster.
- Can we update the docs to clarify that setAttribute() is disallowed, as well as getAttribute().

I'd be happy to work on a PR for the above, if you agree it would be a worthwhile change to make.

arpu:
For me this looks more a problem in the box component not  use the update() https://github.com/aframevr/aframe/blob/master/src/geometries/box.js

diarmidmackenzie:
@arpu, I'm not sure I understand.  There is no box component in the example.

There is a box geometry (which you linked to - this doesn't need an update() method) and the a-box primitive.

I'm pretty sure this same problem would reproduce with an <a-entity> just the same.

diarmidmackenzie:
Looking into why this *does* fail, it looks like the issue is that, in the case above
[...]

PR #5006, opened 2022-02-05
Possible fix for 4973

diarmidmackenzie:
**Description:**

A fix for https://github.com/aframevr/aframe/issues/4973.

In brief...
- if called before an entity is created, removeAttribute() completes asynchronously.
- this leaves a window in which a subsequent call to setAttribute() has no effect - it gets overwritten by the async completion of removeAttribute()
- the fix handles this window.

**Changes proposed:**
- An additional state flag on the component to track the "pending removal" state of a component that occurs when removeAttrbute() completes asynchronously.  Only go through with removal if this flag is still set.
- If setAttribute() is called before removeAttribute() completes...
-- clear this flag (so that the removal is cancelled)
-- recreate the component's corresponding attribute in the DOM (since this will have been removed asynchronously by removeAttribute().

As per #4973 there is a sample glitch that I have used to verify the fix here:
https://glitch.com/edit/#!/modifying-attributes-after-creation?path=index.html%3A1%3A0

Not sure whether this can be incorporated into A-Frame as an additional test case?

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -387,9 +387,13 @@ var proto = Object.create(ANode.prototype, {
 
       // Wait for component to initialize.
       if (!component.initialized) {
+        component.pendingRemoval = true;
         this.addEventListener('componentinitialized', function tryRemoveLater (evt) {
           if (evt.detail.name !== name) { return; }
-          this.removeComponent(name, destroy);
+          if (component.pendingRemoval) {
+            this.removeComponent(name, destroy);
+            component.pendingRemoval = false;
+          }
           this.removeEventListener('componentinitialized', tryRemoveLater);
         });
         return;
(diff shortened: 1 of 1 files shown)


dmarcos:
Good catch. Asynchronous code bites us once again 😄 

I have an alternative that might work and also be simpler.

in the [removeComponent function](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L390) we can set `component.pendingRemoveComponent = true;` and make the [async callback a no-op](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L392) `if (!component.pendingRemoveComponent) { return; }. In [update component](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L481) we can do `component.pendingRemoveComponent = false`

diarmidmackenzie:
I haven't had a chance to test yet, but I think your suggested fix won't work.

As I noted in #4973 (should have mentioned above as well), removeComponent also (synchronously) deletes the DOM Attribute.

The DOM Attribute doesn't get added back in on the update, because that's only done when the component does not exist.

So with your suggested fix, I think we would end up without the attribute in the DOM.

I considered delaying the removal of the DOM attribute to the async phase of removal, but that felt riskier, as it introduces more change to mainline paths.

diarmidmackenzie:
This fix is not good-to-go yet.   Some not-yet-checked-in UTs have thrown up some further issues that the fix doesn't yet address.

PR #5026, opened 2022-03-18
iOS15 ios orientation blank bug

levmillr:
**Description:**
This is causing issues for me on ios15 on safari.  The scene disappears from view - goes black.  I'll try and recreate the issue in a codepen or something.  If anything it doesn't seem to be needed anymore for iOS15(at least 15.4) no white space. There seems to be a different issue on other aframe examples where the camera points straight down on orientation change

**Changes proposed:**
- check for ios15 before adding the listener

with aframe 1.3.0
https://codepen.io/lmiller152/pen/WNdGYOP

1.3.0 with my ios check
https://codepen.io/lmiller152/pen/ExogJgw

@dmarcos  I did a little more digging and found it was a styles issue using `embedded` on `a-scene` with wrappers.  Probably more user error with our setup - but strange the issue only happens on iOS15/safari.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -14,12 +14,13 @@ var initPostMessageAPI = require('./postMessage');
 
 var bind = utils.bind;
 var isIOS = utils.device.isIOS();
+var isIOS15 = utils.device.isIOS15();
 var isMobile = utils.device.isMobile();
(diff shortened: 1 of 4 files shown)


dmarcos:
Also not needed to commit index.js changes

levmillr:
> Also not needed to commit index.js changes

I will get that removed!

PR #5029, opened 2022-03-21
Add exposure to AFrame

AdaRoseCannon:
**Description:**

This adds exposure to the renderer settings and allows flat materials to opt out of it so that they can still look like light emitting surfaces

**Changes proposed:**
- exposure and toneMapping to renderer system
- toneMapped to flat shader

TODO: Docs

### DIFF
--- a/src/systems/renderer.js
+++ b/src/systems/renderer.js
@@ -16,6 +16,8 @@ module.exports.System = registerSystem('renderer', {
     maxCanvasWidth: {default: 1920},
     maxCanvasHeight: {default: 1920},
     physicallyCorrectLights: {default: false},
+    exposure: {default: 1, if: {type: ['ACESFilmic', 'Linear', 'Reinhard', 'Cineon']}},
+    toneMapping: {default: 'No', oneOf: ['No', 'ACESFilmic', 'Linear', 'Reinhard', 'Cineon']},
     precision: {default: 'high', oneOf: ['high', 'medium', 'low']},
     sortObjects: {default: false},
     colorManagement: {default: false},
@@ -31,6 +33,7 @@ module.exports.System = registerSystem('renderer', {
     var renderer = sceneEl.renderer;
     renderer.sortObjects = data.sortObjects;
     renderer.physicallyCorrectLights = data.physicallyCorrectLights;
+    renderer.toneMapping = THREE[this.data.toneMapping + 'ToneMapping'];
 
     if (data.colorManagement || data.gammaOutput) {
       renderer.outputEncoding = THREE.sRGBEncoding;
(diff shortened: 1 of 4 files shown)


AdaRoseCannon:
Preview of change

See how the colours are more muted and less graphic generally. ALso noticed how one of the torus-knots has toneMapped:false so that it has it's original pure colour.

src/systems/renderer.js
```diff
@@ -16,6 +16,8 @@ module.exports.System = registerSystem('renderer', {
     maxCanvasWidth: {default: 1920},
     maxCanvasHeight: {default: 1920},
     physicallyCorrectLights: {default: false},
+    exposure: {default: 1, if: {type: ['ACESFilmic', 'Linear', 'Reinhard', 'Cineon']}},
```
dmarcos, line 19: Can name values with lowercase if not an acronym for consistency: `linear`, `reinhard`, `cineon`

Can capitalize first letter below:

`string.charAt(0).toUpperCase() + string.slice(1);`

src/systems/renderer.js
```diff
@@ -48,6 +51,13 @@ module.exports.System = registerSystem('renderer', {
     }
   },
 
+  update: function () {
+    var data = this.data;
+    var sceneEl = this.el;
+    var renderer = sceneEl.renderer;
+    renderer.toneMappingExposure = data.exposure;
```
dmarcos, line 58: I have no experience with exposure or tone mapping so this might be dumb. Are there use cases for different values of tone mapping and exposure? Should we always match them automatically?

src/systems/renderer.js
AdaRoseCannon: "1" is the THREE.js default of no change. 

https://threejs.org/docs/#api/en/renderers/WebGLRenderer.toneMappingExposure

dmarcos: Probably can replace 1 with something more user friendly and translate internally.

dmarcos: Maybe `no` as `toneMapping` for consistency?

AdaRoseCannon: Well it's a float it should be set to something like 0.5, or 1.1

AdaRoseCannon: Like "intensity" on the light component

dmarcos: Ah I see that the list is an `if` not a `oneOf` Misread. Sorry

src/systems/renderer.js
```diff
@@ -16,6 +16,8 @@ module.exports.System = registerSystem('renderer', {
     maxCanvasWidth: {default: 1920},
     maxCanvasHeight: {default: 1920},
     physicallyCorrectLights: {default: false},
+    exposure: {default: 1, if: {type: ['ACESFilmic', 'Linear', 'Reinhard', 'Cineon']}},
+    toneMapping: {default: 'No', oneOf: ['No', 'ACESFilmic', 'Linear', 'Reinhard', 'Cineon']},
```
dmarcos, line 20: We can do:

```
toneMapping: {default: 'no', oneOf: ['no', 'ACESFilmic', 'linear', 'reinhard', 'cineon']},
```

AdaRoseCannon: Yeah I'll do that.

src/systems/renderer.js
dmarcos: This should be:


toneMapping instead of type
```
    exposure: {default: 1, if: {toneMapping: ['ACESFilmic', 'Linear', 'Reinhard', 'Cineon']}},
```

AdaRoseCannon: Thank you! Great spot!

dmarcos:
This is awesome. Thank you!

PR #5031, opened 2022-03-22
Add alphaMap to Standard material schema.

jmschrack:
**Description:**

Adds `alphaMap`, `alphaTextureOffset`, and `alphaTextureRepeat` to Standard Material schema.  This is already supported by Super-Three

**Changes proposed:**
- Add `alphaMap`, `alphaTextureOffset`, and `alphaTextureRepeat` to Standard Material schema.  
- Update Material#Standard docs to match

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -100,6 +100,9 @@ These properties are available on top of the base material properties.
 
 | Property                      | Description                                                                                                                                     | Default Value |
 |-------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|---------------|
+| alphaMap                      | Alpha map. Controls the opacity of the mesh. Can either be a selector to an `<img>`, or an inline URL.  Requires `transparent=true`. | None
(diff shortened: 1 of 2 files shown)


withyousoft1983 (approved):

dmarcos:
@mrxz thoughts?

Still needs rebase.

mrxz:
The `MeshStandardMaterial` in three supports it, so we could indeed expose it. It just raises the question if it should be added to `flat` and `phong` as well. And whether or not we should expose other/all properties available on the respective materials.

Personally I don't really see the need for `alphaMap` as it can be encoded as the alpha channel in the diffuse texture already. That's also how it's done in glTF, and is generally more efficient. If added, I'd probably advice not using it, unless really needed for a specific use-case. But in those cases people can always write their own component or shader that sets the `alphaMap`.

dmarcos:
Closing for now. can reevaluate if we see more
demand. thanks

Issue #5032, opened 2022-03-22
Asset loading hangs forever when errors hit loading an asset.


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: All
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/assets-dont-timeout-on-error?path=index.html%3A29%3A92

When a URL referred to by an asset is not resolvable, A-Frame scene loading just hangs forever.

See linked Glitch, which is a modified version of this https://aframe.io/examples/showcase/modelviewer/ but with the GLTF URL set to an invalid value.

One reason why it would be nice to initialize the scene, rather than hang forever, is that it would allow application-specific error handling.

I have a potential fix for this that seems to work, which I'll present as a PR.

vincentfretin:
Actually I looked at #2737 and #1155 those two should be closed really.

vincentfretin:
If it's the big glb representing the whole scene that gave an error, I probably want to show the user a message "something went wrong, please try again later" instead of showing a white canvas. If it's smaller glb that may not be so important for the experience,  I may want to still enter the experience, yes, but this is a case by case thing.

kylebakerio:
yes, you can put whatever handling you want there. you can see what element threw the error and handle accordingly. a basic option would be set `localStorage.retryCount = true;' and then `location.reload()`. then, if it happens again, you also have a check; if you see that you've already reloaded, then you just throw an error message to user saying "site is down for maintenance" or something.

and/or just ignore if file is small, which you can see by checking `el.src`.

vincentfretin:
Ok, the choice of showing a message to the user or retry downloading an asset that gave an error can be done in your app. An external library that implement another loader can help with that too.
For aframe core, the logic is currently to load the scene after the timeout, so it actually hide loading screen, and some other glb may finish and appear after the timeout if understand it, at least the is the behavior I see currently.
To be consistent, it should indeed load the scene even if there was an error, which it doesn't do currently, the PR #5033 fixes that.

dmarcos:
Fixed by https://github.com/aframevr/aframe/commit/5189765cbbd81fc4eefe880045662e33aebd138f

PR #5033, opened 2022-03-22
Possible fix for #5032: Don't hang forever on asset loading error

diarmidmackenzie:
**Description:**

Currently, if a bad URL is specified for an asset, A-Frame just shows the loading screen indefinitely (regardless of the asset timeout configured).

This prevents applications from providing application-specific handling (e.g. their own error message) when this might happen.
(in terms of why this might happen, suppose the HTML content is generated by server-side rendering on a back-end, and a bug emerges in the backend).

**Changes proposed:**

Update a-node processing when awaiting all children loading, so that:
- it acknowledges the possibility of errors, and handles them
- goes ahead and declares items "loaded" so that the scene can render, even if there have been errors.

I also added a small example showing the problem:
/examples/test/gltf-model-error/

Not sure if you'll want this included as part of the PR, but it was handy for reproducing the problem & validating the fix.


src/core/a-node.js
```diff
@@ -128,6 +129,13 @@ module.exports = registerElement('a-node', {
           self.emit('loaded', undefined, false);
         }).catch(function (err) {
           error('Failure loading node: ', err);
+
+          // Failed to load a node, but things won't get any better by waiting around.
+          // So get on with rendering the scene by signaling that this node has loaded.
+          // An "error" event has already been fired, so we don't need to fire another one.
+          warn('Rendering scene with errors on node: ', self);
```
dmarcos, line 136: Don't you think this can be misleading? Worried that we're masking errors that the dev might want to stop and fix.

examples/test/gltf-model-error/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>glTF Model</title>
+    <meta name="description" content="glTF Model - A-Frame">
```
dmarcos, line 6: I know it's misleading the test examples are not actually tests but simple examples that test very specific features. For this we should probably write a [unit test](https://github.com/aframevr/aframe/tree/master/tests)

src/core/a-node.js
diarmidmackenzie: There's already a console error about the failure to load a file.
So it didn't seem to me like another error was helpful, and a warning was the appropriate error level here.

Is there anything we can do apart from console warnings & errors to draw a Dev's attention to the fact that something is not right?

(apart from failing to render the scene at all, which as per #5032 I don't think that's a good solution).

examples/test/gltf-model-error/index.html
diarmidmackenzie: OK - I will have a go at a Unit Test.

diarmidmackenzie:
Sorry for the slow progress on this.

I've now added some test cases.

It took me a long time to realize that Mocha is very sensitive to detecting unhandeld errors, and will fail tests whenever your code emits an event named "error", and your test script does not immediately call done(), or stop event propagation.

The error message for this is rather opaque...
```
Error: [object CustomEvent] (undefined:undefined)
```

I have also adjusted the fix itself.  Previously, as soon as a single faulty asset was found, it would jump into the scene.  Now it proceeds with the rest of asset loading on the normal pattern, and logs warnings about the failed assets at the end.

I think that's better behaviour: as per the original issue, we want to mitigate the impact of one incorrect asset URL, but that doesn't mean we want to jump into rendering the scene before other, available, assets have had a chance to load.

diarmidmackenzie:
I've left the gltf_more_error.html test file in for now, as I found it pretty useful while working on this, so it could be useful in future?

But happy to remove it if you prefer...

src/core/a-node.js
```diff
@@ -116,18 +115,26 @@ module.exports = registerElement('a-node', {
...
             child.addEventListener('loaded', resolve);
+            child.addEventListener('error', reject);
           });
         });
 
-        Promise.all(childrenLoaded).then(function emitLoaded () {
+        Promise.allSettled(childrenLoaded).then(function emitLoaded (results) {
+          results.forEach(function checkResultForError (result) {
+            if (result.status === 'rejected') {
+              // An "error" event has already been fired by THRE.js loader,
```
vincentfretin, line 128: THRE.js => THREE.js

src/core/a-node.js
```diff
@@ -116,18 +115,26 @@ module.exports = registerElement('a-node', {
...
+              // so we don't need to fire another one.
+              // A warning explaining teh consequences of the error is sufficient.
```
vincentfretin, line 130: teh => the

src/core/a-node.js
diarmidmackenzie: Thanks, fixed.

diarmidmackenzie:
@dmarcos, with the UTs done, can this be merged now?  Or are there other concerns to address?

@kylebakerio has reached out to ask whether I can push ahead and get this merged.

tests/core/a-assets.test.js
```diff
@@ -246,12 +265,40 @@ suite('a-asset-item', function () {
     assetItem.setAttribute('src', 'doesntexist');
     assetItem.addEventListener('error', function (evt) {
       assert.ok(evt.detail.xhr !== undefined);
+      evt.stopPropagation();
```
vincentfretin, line 268: OMG this line! This is a good find @diarmidmackenzie 
Without this line, it reduced silently the number of tests that was run from 1121 to 559 while I was working on #5123

vincentfretin:
@dmarcos you should probably merge this first, then I will rebase #5123 to resolve a conflict.

vincentfretin:
@diarmidmackenzie can you please rebase your PR, you need to fix the conflict around `evt.stopPropagation();` since #5123 was merged first.

examples/test/gltf-model-error/index.html
diarmidmackenzie: Unit Tests now added.

diarmidmackenzie: Can we leave this in as well, or do you prefer me to delete it?

diarmidmackenzie:
> @diarmidmackenzie can you please rebase your PR, you need to fix the conflict around `evt.stopPropagation();` since #5123 was merged first.

Done.

diarmidmackenzie:
I believe this is now all up-to-date with latest master again.

A couple of conversation ponts where I'm not 100% sure whether you were happy with what I'd proposed or not..
https://github.com/aframevr/aframe/pull/5033#discussion_r984745019
https://github.com/aframevr/aframe/pull/5033#discussion_r833226414

diarmidmackenzie:
Hmm - seing a test failure like this:
SUMMARY:
✔ 734 tests completed
ℹ 3 tests skipped
✖ 1 test failed

FAILED TESTS:
  ✖ "after each" hook for "does not try to enter VR if already in VR"
  
  I'll look into it.

diarmidmackenzie:
OK, I don't know why this test issue has only come to light now, but it is due to inadequate mocking of the canvas element.

This leads to an exception in the `look-controls` component when it tries to add to the canvas classList.

```
TypeError: Cannot read properties of undefined (reading 'add')
      at enableGrabCursor (build/commons.js:17060:32)
      at NewComponent.updateGrabCursor (build/commons.js:17074:7)
      at NewComponent.update (build/commons.js:16723:12)
      at NewComponent.initComponent (build/commons.js:26161:10)
      at NewComponent.updateProperties (build/commons.js:26130:12)
      at AEntity.updateComponent (build/commons.js:24971:17)
      at AEntity.updateComponents (build/commons.js:24948:12)
      at entityLoadCallback (build/commons.js:24739:12)
      at emitLoaded (build/commons.js:25673:9)
```

I believe the solution (which I've just added as a fix into this PR) is to extend the mock canvas object so that `classList.add` and `classList.remove` can be invoked without causing exceptions.

I don't have a good understanding of why this didn't cause problems prior to this fix, but the exception does occur in a loading callback, and we have made changes in the handling of such exceptions, so it's plausible that for some reason this exception didn't cause problems for the tests previously.

This particular failng test is not intended to test exceptions, and the intent of the test will be better met if we avoid these exceptions, which is what the extension to the canvas mock object achieves.  So I think that is the correct fix, and I don't see any evidence of a problem in the code under test.

kylebakerio:
Vincent has made many changes asnd improvements to the tests in recent months, part of which allowed many tests that were being skipped to now be run. _Could_ be related to that, I didn't look into those changes in too much detail.

Good catch and fix.

vincentfretin:
I created #5176 with a slightly modified version of the fix @diarmidmackenzie

diarmidmackenzie:
@dmarcos Do you want to take this fix?  Or give more feedback on the discussion items above?

tests/core/scene/a-scene.test.js
```diff
@@ -162,6 +162,10 @@ suite('a-scene (without renderer)', function () {
 
       // mock canvas
       sceneEl.canvas = {
+        classList: {
+          add: function () {},
+          remove: function () {}
+        },
```
vincentfretin, line 168: @diarmidmackenzie now that #5176 is merged, you should remove first those 4 lines and then merge changes from master to fix the conflict.

diarmidmackenzie: Of course -  sorry, didn't immediately register the implications of your comment.

diarmidmackenzie: Those updates are done.

examples/test/gltf-model-error/index.html
dmarcos: I would remove this so `examples/test` is consistent. It serves the purpose to test visually specific features. Your unit test handles the failure case.

dmarcos:
We can merge this once we remove the example. Thanks for sticking with it.

examples/test/gltf-model-error/index.html
diarmidmackenzie: OK - removed it.

diarmidmackenzie:
Ok - removed that file from examples\test so I think we're good now.

dmarcos:
Thanks so much. Super appreciated

PR #5040, opened 2022-04-12
Add ability to use grip instead of targetRay space

msub2:
**Description:**
`tracked-controls-webxr` currently uses `targetRaySpace` to report controller pose with no option to change it. This PR adds the ability to specify `gripSpace` over `targetRaySpace`, while still keeping `targetRaySpace` as the default.

**Changes proposed:**
- Adds a `space` property on both `tracked-controls` and `tracked-controls-webxr` to allow the user to specify `gripSpace` over `targetRaySpace`.


dmarcos:
Thanks. What’s the use case?

msub2:
I was working on a physics hands demo where I was using just tracked-controls to get controller position and wanted some cubes to stand in for hands, but they'd automatically be centered at the target ray point instead of just being in the center of my hands.  It would also be useful if you're simulating something like the player holding a sword or other tool, which would make less sense being based around the target ray point. I just think this is easier than having to manually get references to the refSpace and frame, calling getPose on that, and *then* setting it or having to apply some manual offset to the controller position.

dmarcos:
Thanks!

PR #5050, opened 2022-05-02
Clear velocity on previous WS / AD axis when changed to a new value

vulture-boy:
In 1.3.0, if the WS or AD axes is changed during runtime, the old velocity associated with the last axis 'sticks' to the entity, resulting in often unwanted, constant movement. This change clears that velocity when the WS or AD axes are adjusted to a new value.

The script now monitors if the WS or AD axis is dynamically changed (via previous value variable) and clears the velocity value associated with its previous axis when a change occurs during an update. The nature of this stop is not smoothed, however... leaving that resolution to the future.

### DIFF
--- a/src/components/wasd-controls.js
+++ b/src/components/wasd-controls.js
@@ -33,6 +33,8 @@ module.exports.Component = registerComponent('wasd-controls', {
     // To keep track of the pressed keys.
     this.keys = {};
     this.easing = 1.1;
+	this.lastWSAxis = wsAxis;
+	this.lastADAxis = adAxis;
 
     this.velocity = new THREE.Vector3();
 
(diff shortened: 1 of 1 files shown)


src/components/wasd-controls.js
```diff
@@ -90,6 +92,16 @@ module.exports.Component = registerComponent('wasd-controls', {
 
     adAxis = data.adAxis;
     wsAxis = data.wsAxis;
+	
+	// If a control axis was changed, reset its old velocity
```
dmarcos, line 96: This would be better handled in an [update method](https://aframe.io/docs/1.3.0/core/component.html#update-olddata) by checking the oldData parameter:

```
  if (oldData.adAxis !== this.data.adAxis) { velocity[oldData.adAxis] = 0; }
```

dmarcos:
Thanks so much for submitting. I suggested a slightly different approach.

vulture-boy:
Makes sense! Didn't realize there was an 'oldData' that could be accessed haha.

vulture-boy:
Your approach is implemented! ✔️

dmarcos:
Thanks. Just edited to reformat and reword comment. Congrats on your first contribution 🎖️

PR #5052, opened 2022-05-05
Allow animating vec3 object3D raw position

slecorne:
**Description:**
Allow animating vec3 position directly on Object3D at once.
Currently we can animate object3D position axis using `property: object3D.position.z` but not the complete vec3 at once.
The goal is to be able to animate the vec3 position directly through object3D without using `setAttributes`.
The current alternative to animate position directly through object3D is to perform 3 animations which has important performance issue.

**Changes proposed:**
- Adding `type: position` similar to raw color animation detection
- In that case animate directly vec3 `object3d.position`
- Tests + documentation update

Feedback welcome

### DIFF
--- a/tests/components/animation.test.js
+++ b/tests/components/animation.test.js
@@ -104,6 +104,27 @@ suite('animation', function () {
   });
 
   suite('direct object3D value animation', () => {
+    test('can animate object3D vec3 position directly', function () {
+      el.setAttribute('animation', {
+        property: 'object3D.position',
+        dur: 1000,
+        from: '1 1 1',
+        to: '0 0 0',
+        type: 'position'
+      });
+      component.tick(0, 1);
+      assert.equal(el.object3D.position.x, 1);
+      assert.equal(el.object3D.position.y, 1);
+      assert.equal(el.object3D.position.z, 1);
+      component.tick(0, 500);
+      assert.ok(el.object3D.position.x > 0);
+      assert.ok(el.object3D.position.x < 1);
+      component.tick(0, 500);
+      assert.equal(el.object3D.position.x, 0);
+      assert.equal(el.object3D.position.y, 0);
+      assert.equal(el.object3D.position.z, 0);
+    });
+
     test('can animate object3D value directly', function () {
       el.setAttribute('animation', {
         property: 'object3D.position.x',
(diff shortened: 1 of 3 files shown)


tests/components/animation.test.js
```diff
@@ -104,6 +104,27 @@ suite('animation', function () {
   });
```
captain-hoity-toity, line 104: Here are some recommended changes for this PR commit.

1. The `setTimeout` function is used in the `remove` suite. This is not recommended because the timeout duration is arbitrary and may lead to flaky tests. Instead, you can use the `done` function in combination with event listeners to ensure that the tests only proceed once the necessary conditions have been met.

2. The `stopRelatedAnimations` test suite could benefit from some additional comments explaining the purpose and functionality of each test. This will make it easier for other developers to understand the intention of the tests and maintain them in the future.

3. In the `color animation` suite, the test `can animate color object directly` has a complex logic. It would be better to break this test down into smaller, more manageable tests. This would make it easier to identify specific issues if the test fails.

4. In the `basic animation` suite, the `handles non-truthy from value (i.e., 0)` test could benefit from a more descriptive test name. The current name does not clearly convey the purpose of the test.

5. In the `event emissions` suite, the `emits animationcomplete event twice` test could be simplified by using a counter variable instead of a boolean flag. This would make the test logic more straightforward and easier to understand.

6. In the `tick` suite, the test `only calls animejs animation.tick if playing` could benefit from a more descriptive test name. The current name does not clearly convey the purpose of the test.

7. The `vec3 animation` suite could benefit from some additional comments explaining the purpose and functionality of each test. This will make it easier for other developers to understand the intention of the tests and maintain them in the future.

slecorne: thanks for the comment, I did not touch most of the test suites you are mentioning and I am not sure I fill confident to modify them.
regarding the vec3 animation, I'll try to update with more comments if time but I did not touch aframe code since that PR more than a year ago so not sure I'll manage that soon

mrxz:
I might be missing something here, but is there a reason to prefer `object3D.position` over `position` for the animation component? It already had special treatment for `position`, `scale` and `rotation` that skipped `setAttribute` and directly set the values on the underlying `object3D`.

The documentation and examples all use `rotation`, `position` and `scale` and only resort to `object3D.position.x` when animating a single value. It'd be more idiomatic if the `animation` component would start supporting `position.x` instead of `object3D.position`.

Either way, with https://github.com/aframevr/aframe/commit/4f711a4efabd68bb36ffacef3718f7e22135c559 there is a bug when not specifying a `from` value. This works for `position` but fails for `object3D.position`.

dmarcos:
"position" and "object3D.position" the same. It's nice both work as expected.

I fixed the issue in https://github.com/aframevr/aframe/commit/204a8db7a00c586191b28d6ee1133ff52bd2516b. thanks

dmarcos:
We can also add support for `position.x`, `scale.x` and `rotation.x`

PR #5064, opened 2022-06-22
degToRad fix in vendor

AdaRoseCannon:
**Description:**

Some missing Math -> MathUtils changes

**Changes proposed:**
- Fix `vendor/DeviceOrientationControls.js`


dmarcos:
Great catch. Thank you!

PR #5065, opened 2022-06-22
XR Cursor

AdaRoseCannon:
**Description:**

Allows the cursor component to be used with generic XR inputs without needing to be specifically attached to anything.

In particular this lets it work screen based AR

**Changes proposed:**
- Add an xrselect rayOrigin which sets the origin to the targetRaySpace where select events happen
- Fix mouse constant fusing on mobile
- built on top of #5064 

Demo:
https://www.youtube.com/watch?v=j1SNRmsgfOo

### DIFF
--- a/docs/components/cursor.md
+++ b/docs/components/cursor.md
@@ -88,7 +88,7 @@ AFRAME.registerComponent('cursor-listener', {
 | fuse               | Whether cursor is fuse-based.                                                                                                                      | false on desktop, true on mobile |
 | fuseTimeout        | How long to wait (in milliseconds) before triggering a fuse-based click event.                                                                     | 1500                             |
 | mouseCursorStylesEnabled | Whether to show pointer cursor in `rayOrigin: mouse` mode when hovering over entity.                                                               | true                             |
-| rayOrigin          | Where the intersection ray is cast from (i.e.,entity or mouse). `rayOrigin: mouse` is extremely useful for VR development on a mouse and keyboard. | entity
+| rayOrigin          | Where the intersection ray is cast from (i.e. xrselect ,entity or mouse). `rayOrigin: mouse` is extremely useful for VR development on a mouse and keyboard. | entity
 | upEvents           | Array of additional events on the entity to *listen* to for triggering `mouseup` (e.g., `trackpadup` for daydream-controls).                       | []                               |
 
 To further customize the cursor component, we configure the cursor's dependency
@@ -183,3 +183,16 @@ pick up event with the `begin` attribute:
 
 To play with an example of a cursor with visual feedback, check out the [Cursor
 with Visual Feedback example on CodePen][cursor-codepen].
+
+## XR Select Cursor
+
+When an XR `"selectstart"` event happens the raycaster picks an element based upon it's current location.
+This works for handheld AR, and headmounted VR and AR. This works well with the mouse `rayOrigin` too.
+
+```html
+<a-scene
+  cursor__mouse="rayOrigin: mouse"
+  cursor__xrselect="rayOrigin: xrselect"
+  raycaster="objects:#objects *;"
+>
+```
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -65,6 +65,7 @@ module.exports.Component = registerComponent('cursor', {
     this.intersectedEl = null;
     this.canvasBounds = document.body.getBoundingClientRect();
     this.isCursorDown = false;
+    this.activeXRInput = null;
 
     // Debounce.
     this.updateCanvasBounds = utils.debounce(function updateCanvasBounds () {
@@ -86,6 +87,19 @@ module.exports.Component = registerComponent('cursor', {
   update: function (oldData) {
     if (this.data.rayOrigin === oldData.rayOrigin) { return; }
     this.updateMouseEventListeners();
+  },
+
+  tick: function () {
+    // Update on frame to allow someone to select and mousemove
+    var frame = this.el.sceneEl.frame;
+    var inputSource = this.activeXRInput;
+    if (this.data.rayOrigin === 'xrselect' && frame && inputSource) {
+      this.onMouseMove({
+        frame: frame,
+        inputSource: inputSource,
+        type: 'fakeselectevent'
+      });
+    }
   },
 
   play: function () {
@@ -210,6 +224,12 @@ module.exports.Component = registerComponent('cursor', {
       var point;
       var top;
 
+      var frame;
+      var inputSource;
+      var referenceSpace;
+      var pose;
+      var transform;
+
       camera.parent.updateMatrixWorld();
 
       // Calculate mouse position based on the canvas element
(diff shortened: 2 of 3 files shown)


src/components/cursor.js
```diff
@@ -223,7 +225,16 @@ module.exports.Component = registerComponent('cursor', {
       mouse.x = (left / bounds.width) * 2 - 1;
       mouse.y = -(top / bounds.height) * 2 + 1;
 
-      if (camera && camera.isPerspectiveCamera) {
+      if (this.data.rayOrigin === 'xrselect' && evt.type === 'selectstart') {
+        const frame = evt.frame;
```
dmarcos, line 229: use var for consistency with the code base

docs/components/cursor.md
```diff
@@ -183,3 +183,16 @@ pick up event with the `begin` attribute:
...
 with Visual Feedback example on CodePen][cursor-codepen].
+
+## XR Select Cursor
+
+When an XR `"selectstart"` event happens the raycaster picks an element based upon it's current location.
+This works for handheld AR, and headmounted VR and AR. This works well with the mouse `rayOrigin` too.
+
+```html
+<a-scene
+  cursor__mouse="rayOrigin: mouse"
```
dmarcos, line 194: multiple components for this might feel like an overkill. wondering if we can simplify somehow. maybe `rayOrigin: mouse` also considers selectstart? Not a fan either because makes `rayOrigin: mouse` feel deceiving (not really only about mouse anymore)

src/components/cursor.js
dmarcos: How is `mousemove` triggered in an `xrselect` context?

dmarcos:
Thanks for the hard work and patience. Left a few comments questions. Let me know what you think

docs/components/cursor.md
AdaRoseCannon: I can think of a few situations in which a developer may prefer to handle all of their own XR select behaviours but still take advantage of the mouse cursor.

[My XR Input Handling library](https://github.com/AdaRoseCannon/handy-work/blob/main/README-AFRAME.md) in particular makes it really simple to start building your x-platform own XR cursor using the `entity` method of the cursor component component.

src/components/cursor.js
AdaRoseCannon: Currently I am just handling mousedown and mouseup but I can add mousemove, there are two option that jump to mind:

* only mousemove on an input when it is between a selectstart and selectend
* mousemove all the time for all inputs (obviously transient inputs will only mouse move when an input is pressed because they only exist when the user is touching the screen)

docs/components/cursor.md
dmarcos: Any other ideas to avoid the multiple component approach?

vendor/DeviceOrientationControls.js
```diff
@@ -85,13 +85,13 @@ THREE.DeviceOrientationControls = function ( object ) {
 
     if ( device ) {
 
-      var alpha = device.alpha ? THREE.Math.degToRad( device.alpha ) + scope.alphaOffset : 0; // Z
+      var alpha = device.alpha ? THREE.MathUtils.degToRad( device.alpha ) + scope.alphaOffset : 0; // Z
```
dmarcos, line 88: Probably these Math/MathUtils can go in a separate PR? Unrelated to the AR cursor

AdaRoseCannon: hmm these were in another PR that got merged they should disappear I will rebase

docs/components/cursor.md
AdaRoseCannon: have rayOrigin be an array of strings?

vendor/DeviceOrientationControls.js
dmarcos: ok cool thanks

docs/components/cursor.md
dmarcos: array would weird too I think. probably even weirder. Let's go with the multiple components for now but I don't feel super comfortable. Thanks for the patience

vendor/DeviceOrientationControls.js
dmarcos: If we rebase this is pretty close to merge

AdaRoseCannon: I'm still working on the new tick method for mousemove support for xr and entity rayOrigins

src/components/cursor.js
AdaRoseCannon: My mistake here was that I forgot that the raycaster updates eah frame automatically, adding this support was a lot easier than I had expected

src/components/cursor.js
```diff
@@ -250,6 +261,24 @@ module.exports.Component = registerComponent('cursor', {
...
+      this.el.components.raycaster.checkIntersections();
+
+      // if something was tapped on don't do ar-hit-test things
+      if (
+        this.el.components.raycaster.intersectedEls.length &&
+        this.el.sceneEl.components['ar-hit-test'] !== undefined &&
+        this.el.sceneEl.components['ar-hit-test'].data.enabled
+      ) {
+        // Cancel the ar-hit-test behaviours and disable the ar-hit-test
+        this.el.sceneEl.components['ar-hit-test'].data.enabled = false;
```
dmarcos, line 275: We shouldn't access data directly but using getAttribute:

`this.el.sceneEl.getAttribute('ar-hit-test).enabled`

src/components/cursor.js
```diff
@@ -250,6 +261,24 @@ module.exports.Component = registerComponent('cursor', {
...
     }
 
+    if (this.data.rayOrigin === 'xrselect' && evt.type === 'selectstart') {
+      this.onMouseMove(evt);
+      this.el.components.raycaster.checkIntersections();
+
+      // if something was tapped on don't do ar-hit-test things
+      if (
+        this.el.components.raycaster.intersectedEls.length &&
+        this.el.sceneEl.components['ar-hit-test'] !== undefined &&
```
dmarcos, line 271: Should the component `ar-hit-test` be set on `init` if rayOrigin `xrselect`? For this to work it depends on the user to also set `ar-hit-test` separately

src/components/cursor.js
```diff
@@ -269,14 +298,19 @@ module.exports.Component = registerComponent('cursor', {
     var data = this.data;
     this.twoWayEmit(EVENTS.MOUSEUP);
 
+    if (this.reenableARHitTest === true) {
+      this.el.sceneEl.components['ar-hit-test'].data.enabled = true;
```
dmarcos, line 302: Same as a above. t`his.el.sceneEl.setAttribute('ar-hit-test', 'enabled', true);`

src/components/cursor.js
AdaRoseCannon: it's not so much a dependency just that things can get messy if both are being used so if the user is tapping on a 3d object then it shouldn't also interact with the real world behind it

src/components/cursor.js
```diff
@@ -250,6 +261,24 @@ module.exports.Component = registerComponent('cursor', {
...
+
+      // if something was tapped on don't do ar-hit-test things
+      if (
+        this.el.components.raycaster.intersectedEls.length &&
+        this.el.sceneEl.components['ar-hit-test'] !== undefined &&
+        this.el.sceneEl.components['ar-hit-test'].data.enabled
+      ) {
+        // Cancel the ar-hit-test behaviours and disable the ar-hit-test
+        this.el.sceneEl.components['ar-hit-test'].data.enabled = false;
+        this.el.sceneEl.components['ar-hit-test'].hitTest = null;
```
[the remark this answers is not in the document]
AdaRoseCannon: Oh I like that, i'm a stupid, that is so much better.

AdaRoseCannon:
This now includes a fix for ar-hit-test to end any active hit testing if it's disabled mid hit test

src/components/scene/ar-hit-test.js
```diff
@@ -366,6 +366,11 @@ module.exports.Component = register('ar-hit-test', {
     this.makeBBox();
   },
   update: function () {
+    // If it is disabled it's cleaned up
+    if (this.data.enabled === false) {
+      this.hitTest = null;
+      this.bboxMesh.visible = false;
+    }
```
dmarcos, line 373: Do we have to restore those if the component is reenabled?

AdaRoseCannon: They get reassigned when an xr selectstart happens 

the mesh is made visible  
and the hit-test set to the one corresponding to that input

AdaRoseCannon:
I think this should only work when the XR Input is Screen Based Input OR the Immersive Mode is AR

src/components/cursor.js
```diff
@@ -250,6 +284,23 @@ module.exports.Component = registerComponent('cursor', {
       evt.preventDefault();
     }
 
+    if (this.data.rayOrigin === 'xrselect' && evt.type === 'selectstart') {
```
dmarcos, line 287: Can this `if` be include in the [if above](https://github.com/aframevr/aframe/pull/5065/files#diff-0e82e2c917e7127027721c5ab28da273907fdf1ddebc653d68cda0a9f97b17a7R248)

```
  if (evt.type === 'selectstart') {
     this.activeXRInput = evt.inputSource;
      this.onMouseMove(evt);
      this.el.components.raycaster.checkIntersections();
      ....
  } 

```

PR #5074, opened 2022-07-12
Don't use the cursor component if you need to render the controllers yourself

AdaRoseCannon:
This is a small addition to the cursor component that I was thinking about but was a bit slow adding it to the last PR. I still need to test to make sure it behaves as expected in headsets.

**Description:**

This disables the cursor component when controllers need to be rendered by the user (tracked controllers on opaque displays) because in these situations the user is probably using one of AFrame's controller inputs to render a controller and should put an additional cursor directly on those inputs.

**Changes proposed:**
- Early return for xr select events for tracked controllers on opaque displays


dmarcos:
This needs rebase.

AdaRoseCannon:
rebased

AdaRoseCannon:
This is related as well https://github.com/aframevr/aframe/pull/5076

AdaRoseCannon:
This issue is okay to be merged I think but I will close 5076 for now.

Issue #5078, opened 2022-07-15
device-orientation-permission-ui incorrectly emits deviceorientationpermissiongranted too early


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: iPhone
- Reproducible Code Snippet or URL: https://premature-granted-event.glitch.me/

Open this on an iPhone, and accept permissions for device orientation.

Logs show that events are created in the following order:
- Granted (this is a bogus event, should not be generated).
- Requested
- Granted

The bug is [here](https://github.com/aframevr/aframe/blob/fb9beae5143e22c537f09bdd230b37b1c30d9fde/src/components/scene/device-orientation-permission-ui.js#L55-L65)

The code is structued as

```
requestPermission.catch().then()
```

This means that the "then" branch gets executed even when an exception is caught in the requestPermission.  Which results in the behaviour above.

Fix is to switch the catch() and then() branches.  I'll submit a PR for this.

dmarcos:
Thanks!

vincentfretin:
You can close the issue, the PR has been merged.

dmarcos:
fixed by https://github.com/aframevr/aframe/pull/5079

PR #5079, opened 2022-07-15
device-orientation-permisson-ui: Don't emit "granted" event until permissions are actually granted.

diarmidmackenzie:
**Description:**

See https://github.com/aframevr/aframe/issues/5078

**Changes proposed:**

Code is currently structured as:

```
requestPermission.catch(<create dialog>).then(<mark permissions enabled>)
```

This is flawed, because when the `catch` branch is executed, the `then` branch then gets executed immediately as well.

The fix is to restructure the code as follows:
```
requestPermission.then(<mark permissions enabled>).catch(<create dialog>)
```

I've also made a small UT fix to check for the non-emission of events at this stage in the process.  This change made the script fail without the fix, and it succeeds with the fix in place.

UTs for this component could benefit from some significant extensions, but I'm not sure how soon I'd find time for that, and I didn't want to hold up making this fix available.  So I've only implemented this minimal update to the UT alongside this change.

### DIFF
--- a/src/components/scene/device-orientation-permission-ui.js
+++ b/src/components/scene/device-orientation-permission-ui.js
@@ -52,7 +52,10 @@ module.exports.Component = registerComponent('device-orientation-permission-ui',
     this.onDeviceMotionDialogAllowClicked = bind(this.onDeviceMotionDialogAllowClicked, this);
     this.onDeviceMotionDialogDenyClicked = bind(this.onDeviceMotionDialogDenyClicked, this);
     // Show dialog only if permission has not yet been granted.
-    DeviceOrientationEvent.requestPermission().catch(function () {
+    DeviceOrientationEvent.requestPermission().then(function () {
+      self.el.emit('deviceorientationpermissiongranted');
+      self.permissionGranted = true;
+    }).catch(function () {
       self.devicePermissionDialogEl = createPermissionDialog(
         self.data.denyButtonText,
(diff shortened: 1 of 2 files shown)


dmarcos:
It seems I don't fully understand promises 😄 If the `catch` function doesn't explicitly return a value the `then` function shouldn't run, right? It looks you're seeing something different and I don't understand promises.

diarmidmackenzie:
[Here's](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) the documentation I can find on `catch()`

> The Promise returned by catch() is rejected if onRejected throws an error or returns a Promise which is itself rejected; otherwise, it is fulfilled.

I infer from this that if the function provided in `catch()` returns nothing, it returns a Promise that is fulfilled, hence the `.then()` is invoked.

Alternatives to the fix I proposed could be:
- throwing an error in the `catch()` function 
- explicitly returning a rejected Promise

I've tested these, and both work - specifically:

either adding this line at the end of the `catch()` function

```
throw 'need to showing permissions dialog';
```

or (more convoluted, and I think potentially rather confusing, but it does seem to work)....

```
const promiseRef = DeviceOrientationEvent.requestPermission().catch(function () {
```
and then at the end of the `catch()` function:
```
return promiseRef
```

dmarcos:
Thanks for the explanation!

PR #5081, opened 2022-07-19
Fix bug with removal of line with __suffix component name (fix #5080)

### LINKED ISSUE #5080: Hit error when removing "line__xxx" component
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: All
- Reproducible Code Snippet or URL: None provided.

Create a line components on an entity with a name like line__2, as per this example.
https://aframe.io/docs/1.3.0/components/line.html

Remove the line component.

You get this console error.
core:a-entity:warn Tried to remove `Object3D` of type:  line which was not defined.

Very simple bug.

[This line](https://github.com/aframevr/aframe/blob/2c4509aa1a9a3c447c08be9facfea0c7d6a9e156/src/components/line.js#L69) assumes the Object3D name used for the mesh is "line".  In fact it will be "line__2"

```
    this.el.removeObject3D('line', this.line);
```

Correct code should be:

```
    this.el.removeObject3D(this.attrName, this.line);
```

(matching [the earlier call to setObject3D](https://github.com/aframevr/aframe/blob/2c4509aa1a9a3c447c08be9facfea0c7d6a9e156/src/components/line.js#L31))

I will submit a PR for this.

diarmidmackenzie:
**Description:**

Fix for bug described in issue 5080

**Changes proposed:**

- Use correct name when removing "line" Object3D
- Added Unit Test to cover line__suffix style component.

### DIFF
--- a/src/components/line.js
+++ b/src/components/line.js
@@ -66,7 +66,7 @@ module.exports.Component = registerComponent('line', {
   },
 
   remove: function () {
-    this.el.removeObject3D('line', this.line);
+    this.el.removeObject3D(this.attrName, this.line);
   }
 });
 
--- a/tests/components/line.test.js
+++ b/tests/components/line.test.js
@@ -4,15 +4,23 @@ var entityFactory = require('../helpers').entityFactory;
 suite('line', function () {
   var el;
   var component;
+  var componentSuffix;
 
   setup(function (done) {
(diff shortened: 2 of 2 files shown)


dmarcos:
Great catch. Thanks so much! And 🥇 for adding a test.

PR #5086, opened 2022-08-01
update some github repositories to their new location

vincentfretin:
Some repositories moved to a different organization, even if the old url redirects to the new url, it's better to point it directly to the new one.


dmarcos:
Thank you!

Issue #3833, opened 2018-10-29
Cursor component: events should contain mouse button


naraesk:
**Description:**

I've added event listeners for the mousedown/mouseup/mouseenter/mouseleave events of the [cursor component](https://github.com/aframevr/aframe/blob/master/docs/components/cursor.md). The event object contains no information about which button was pressed to trigger this event, i.e. left, right, or middle mouse button. It is possible to get the information from the javascript mouse events and combine that information with the aframe cursor events, but this is needless complicated. It would be much easier if the events from the cursor components would contain an attribute button with value 1, 2, or 3.

- A-Frame Version: 0.8.2
- Platform / Device: Desktop/Chromium
- Reproducible Code Snippet or URL: https://codesandbox.io/s/n035zom4o4

gftruj:
Hi, i've modified [src/components/cursor.js](https://github.com/gftruj/aframe/blob/feat-cursor-button-id/src/components/cursor.js) a bit, using `rayOrigin: mouse` adds the mouse button value to the event detail. 

The changes are within the [update](https://github.com/gftruj/aframe/blob/feat-cursor-button-id/src/components/cursor.js#L78-L83), [onCursorDown](https://github.com/gftruj/aframe/blob/feat-cursor-button-id/src/components/cursor.js#L228-L229), and [onCursorUp](https://github.com/gftruj/aframe/blob/feat-cursor-button-id/src/components/cursor.js#L261-L263) functions, but it seems to be working with `mouseenter` and `mouseleave` events as well.

You can check it out [here](https://jsfiddle.net/rvemwxkz/)

This feature doesn't seem VRy, still if it's worthwhile, I'd gladly make a proper PR.

ngokevin:
If you are making a 2D/mouse-y app, then I think manually listening to mouse clicks, and then grabbing in the handler the cursor.intersectedEl or raycaster.getIntersection is good. Don't need we need to add to cursor.js. Thanks!

naraesk:
@gftruj: Thanks, perfect! Would be glad to see it merged. 

@ngokevin: Not sure I understood. Do you mean to have something like a normal javascript document.onclick eventhandler and from there accessing the cursor.intersectedEl? That doesn't sound very robust. When the cursor moves fast enough cursor.intersectedEl will give me the wrong element or am I missing something?

ngokevin:
[...]

PR #5088, opened 2022-08-12
Make MouseEvent and TouchEvent available on cursor events

diarmidmackenzie:
**Description:**

When using cursor on desktop, it can be complicated to distinguish between left clicks & right clicks.  This is because the cursor events for click, mouseup and mousedown have much less information than the MouseEvent and TouchEvents.

This issue has been raised previously here (in 2018), but was closed without any changes being made.
https://github.com/aframevr/aframe/issues/3833

Recent discussion on the A-Frame discord channel with @dmarcos led to a proposed solution to this problem, which this PR implements (full discussion pasted below for reference)

**Changes proposed:**

- When a MouseEvent or TouchEvent is the immediate cause of a cursor event (mousedown, mouseup or click), include a reference to the MouseEvent or TouchEvent object in event.detail as event.detail.originalEvent
- Unit tests for this change
- Documentation updates
- A simple example (at /examples/test/cursor/mouse-events.html) that changes color of a block based on left, right & middle button clicks.

** Full discussion from discord **

> diarmid — 08/04/2022
> I'd love cursor events triggered by the mouse to include data telling me what mouse button was pressed (left, middle or right).  For desktop apps it's pretty common to have different mouse buttons do different functions.
> 
> This has been requested before (back in 2018), and various offers of PRs made to make this extension, but offers have been rejected on the basis that "it's not needed".
> 
> https://github.com/aframevr/aframe/issues/3833
> 
> @dmarcos Would you feel any differently about this today?  I'd be happy to do a PR for this, just to add data about which mouse button was pressed to the click, mouseup & mousedown events generated by cursor.  I agree with the analysis in the issue that this can be handled in the application, but would be far simpler to be able to use cursor events directly.
> 
> dmarcos — 08/04/2022
> I always worry abstracting away DOM APIs too much. Prefer people to use the standard APIs as much as possible. Once you commit to abstraction things get complex and hard to maintain very quickly 
> an example that people can copy and paste might be more appropriate. can add to the docs
> what do you think?
> fan of copy pasta in general. reusable code without hiding the details
> if simple is easy to adapt and tweak
> a high level abstract API makes you more dependent of maintainers
> 
> diarmid — 08/04/2022
> That makes sense, but the cursor component already has an API that generates custom events (click, mouseup, mousedown), which are missing some important detail vs. what's on the DOM APIs for the equivalent events.
> 
> These events do provide a number of conveniences over the standard DOM events (e.g. you can listen for them on a specific target object), so I'd not be in favour of deprecating them.
> 
> What about including a reference to the original MouseEvent in the cursor's CustomEvent?  That would seem to avoid the maintainability concerns, and allow people to use the standard API to get whatever additional info they might want about the mouse event that triggered the cursor event.
> My problem with a copy-paste example, is that it still leaves us with 2 completely different patterns for handling cursor events - one way if you care about differentiating left click from right click, another way if you don't.  And people will develop apps that handle left & right click indiscriminately, and then decide later that they want to handle them differently and have to rewrite a bunch of code from one pattern into another.
> 
> No matter how good the sample code is, that's still a pain to have to deal with.  The basic way of handling cursor events should ideally be easily extensible to more advanced use cases without having to do substantial rewrites.
> 
> dmarcos — 08/05/2022
> yeah passing the mousevent might be simple enough. thanks for the convo
> 
> diarmid — 08/05/2022
> Ok, if you'd consider that change, I'll have a go at a PR some time next week.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -301,7 +301,7 @@ module.exports.Component = registerComponent('cursor', {
       }
     }
 
-    this.twoWayEmit(EVENTS.MOUSEDOWN);
+    this.twoWayEmit(EVENTS.MOUSEDOWN, evt);
     this.cursorDownEl = this.intersectedEl;
   },
 
@@ -318,7 +318,7 @@ module.exports.Component = registerComponent('cursor', {
     this.isCursorDown = false;
 
     var data = this.data;
-    this.twoWayEmit(EVENTS.MOUSEUP);
+    this.twoWayEmit(EVENTS.MOUSEUP, evt);
 
     if (this.reenableARHitTest === true) {
       this.el.sceneEl.setAttribute('ar-hit-test', 'enabled', true);
@@ -334,7 +334,7 @@ module.exports.Component = registerComponent('cursor', {
 
     if ((!data.fuse || data.rayOrigin === 'mouse' || data.rayOrigin === 'xrselect') &&
         this.intersectedEl && this.cursorDownEl === this.intersectedEl) {
-      this.twoWayEmit(EVENTS.CLICK);
+      this.twoWayEmit(EVENTS.CLICK, evt);
     }
 
     // if the current xr input stops selecting then make the ray caster point somewhere else
@@ -475,7 +475,7 @@ module.exports.Component = registerComponent('cursor', {
   /**
    * Helper to emit on both the cursor and the intersected entity (if exists).
    */
-  twoWayEmit: function (evtName) {
+  twoWayEmit: function (evtName, originalEvent) {
     var el = this.el;
     var intersectedEl = this.intersectedEl;
     var intersection;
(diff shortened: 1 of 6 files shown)


src/components/cursor.js
```diff
@@ -475,19 +475,25 @@ module.exports.Component = registerComponent('cursor', {
...
+    if (originalEvent) {
+      this.eventDetail.originalEvent = originalEvent;
+    }
     el.emit(evtName, this.eventDetail);
 
     if (!intersectedEl) { return; }
 
     this.intersectedEventDetail.intersection = intersection;
+    if (originalEvent) {
+      this.intersectedEventDetail.originalEvent = originalEvent;
```
dmarcos, line 495: maybe calling `touchEvent` or `mouseEvent` depending on the case instead of generic `originalEvent` is more clear. Uses the Web standard nomenclature

diarmidmackenzie: Yes, I like that suggestion.  I was aiming for minimal change to the interface, but I agree this will be more usable.  Application developers need to consider whether hey are reading a mouse event or a touch event as they have different properties, and this will help steer them in the correct direction.

PR updated with this change.

dmarcos:
Thanks so much for the patience. Awesome work

PR #5089, opened 2022-08-14
Fix remaining occurrences of THREE.Math

vincentfretin:
Replace remaining occurrences of `THREE.Math.degToRad` by `THREE.MathUtils.degToRad`.


dmarcos:
Thank you!

PR #5090, opened 2022-08-14
Fix bad link in documentation

vincentfretin:
`npm run test:docs` didn't work anymore because the referenceRegex regexp wrongly matching `[Blender][blender]: Draco compression` when searching for `[blender]: https://www.blender.org/` type of reference. Instead of trying to modify the regexp, I just changed the doc to avoid the colon.
With that fixed, the script found a bad link caused by the renaming of webvr to webxr of the `vr-headsets-and-webxr-browsers.md` file.


dmarcos:
Thanks so much!

PR #5095, opened 2022-08-17
5094 firefox tests

diarmidmackenzie:
**Description:**

Core code + test scripts must account for the fact that Firefox doesn't support TouchEvent.

**Changes proposed:**

Check TouchEvent exists before using it
Don't run tests that depend on TouchEvent when TouchEvent does not exist


dmarcos:
Thank you!

vincentfretin:
@dmarcos I think the build bot is down again. There was no build after this PR was merged.

PR #5096, opened 2022-08-22
Fixes dynamically created a-scene attributes

rockwalrus:
**Description:**
document.createElement("a-scene").setAttribute doesn't work
correctly when there is already a registered system and component
with the same name.

**Changes proposed:**
- Call the `ANode` implementation if there are no `systemNames` yet.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -550,7 +550,12 @@ module.exports.AScene = registerElement('a-scene', {
           system.updateProperties(value);
           return;
         }
-        AEntity.prototype.setAttribute.call(this, attr, value, componentPropValue);
+        if (this.systemNames.length) {
+          AEntity.prototype.setAttribute.call(this, attr, value, componentPropValue);
+        } else {
+          // system registration isn't completed yet, so just set as a regular node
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks. Can you provide a concrete example of expected vs. what you get behavior?

rockwalrus:
Well, the test case shows the issue pretty well. What happens if you create your `a-scene` after A-Frame initializes is that in certain cases some of the attributes on the element never get set, leading to misconfigured systems. I noticed this with `gltf-model` but it's true at least for all attributes that correspond to systems with the same name as their associated component.

My expectation is that A-Frame should handle system attributes correctly regardless of whether `a-scene` is created before or after A-Frame initializes. Note that I'm only ever creating one `a-scene` and this is the only apparent problem with dynamically creating it.

dmarcos:
Can you provide some code that reproduces the issue?

rockwalrus:
It's at https://aframe-dynamic-system-bug.glitch.me . Check the console logs to see the problem.

dmarcos:
This works as expected in 1.6.0. This fix is no longer necessary. Thanks

Issue #4919, opened 2021-08-31
Support for glTF+KTX2/Basis textures ("KHR_texture_basisu")


donmccurdy:
The KTX2 (".ktx2") texture format uses Basis Universal compression to provide several benefits:

1. Much lower in-memory size for textures.
2. Much faster GPU upload for textures.
3. (Sometimes) Lower file/transmission size for textures.

In combination these make it possible to use more texture detail at lower performance cost in a scene, and these textures can also be embedded in glTF files.

Adding support in A-Frame would be straightforward (similar to https://github.com/aframevr/aframe/pull/4910) except for one issue —  KTX2Loader.js is in `examples/jsm` and there's no `examples/js` equivalent. Does A-Frame's three.js fork have an existing way of pulling in `examples/jsm` files? Or should we ask to have the `examples/js` version created on the three.js repository?

dmarcos:
Thanks so much. Not sure what's the best way to handle. An `examples/js` version would be great is possible. Would that be something reasonable for the THREE folks?

donmccurdy:
Let's give it a try! Opened https://github.com/mrdoob/three.js/pull/22485. If that gets merged we'll need to import `THREE.KTX2Loader` and its dependency, `THREE.WorkerPool`.

dmarcos:
> Let's give it a try! Opened [mrdoob/three.js#22485](https://github.com/mrdoob/three.js/pull/22485). If that gets merged we'll need to import `THREE.KTX2Loader` and its dependency, `THREE.WorkerPool`.

It seems it got merged. Yeeeeah! Thanks so much

XrayBravoGolf:
Is there any updates

> Let's give it a try! Opened [mrdoob/three.js#22485](https://github.com/mrdoob/three.js/pull/22485). If that gets merged we'll need to import `THREE.KTX2Loader` and its dependency, `THREE.WorkerPool`.

Were there any work-in-progress of the KEX2Loader being put into A-Frame right now? As of today my build out of the master branch (latest commit 9e0cea) still gives this warning:
> components:gltf-model:warn THREE.GLTFLoader: setKTX2Loader must be called before loading KTX2 textures 
, thus I assume it has not been made part of this project yet.

XrayBravoGolf:
By the way, is the fact that A-Frame is using the super-three instead of the THREE directly from mrdoob/three.js going to affect this KTX2 loader issue?

dmarcos:
[...]

PR #5101, opened 2022-08-25
Add support for KTX2 texture compression to gltf-model

jameskane05:
**Description:**

This addresses https://github.com/aframevr/aframe/issues/4919 by adding THREE.KTX2Loader to the A-Frame gltf-model system. It works just like the DracoLoader, in that you have to provide it with a link to a hosted transcoder library folder, which must contain two files: `basis_transcoder.wasm` and `basis_transcoder.js`, as documented in the `gltf-model.md`. These files are available from the three.js repository, under `examples/js/libs/basis`.

This also extends an example scene to include an asset with a compressed texture (made with `gltf-transform etc1s grip.glb grip-compressed.glb` command). This is a small asset from my upcoming game, which is linked to from Glitch.com hosting. Let me know if there is a preferred asset to use for this example or if it should be moved to within the project somehow instead of remotely hosted. Also let me know if there are other stylistic or convention things I missed. Otherwise looking forward to having access to this feature!

image

### DIFF
--- a/docs/components/gltf-model.md
+++ b/docs/components/gltf-model.md
@@ -165,7 +165,7 @@ be included. A Google-hosted version of the Draco decoder libraries saves you fr
     basis_transcoder.wasm — WebAssembly transcoder.
 
 These files are available from the three.js repository, under
-[examples/js/libs/bsdid][basis]. A Google-hosted version of the Basis transcoder libraries saves you from needing to include these libraries in your own project: set `https://www.gstatic.com/draco/v1/decoders/` as the value for `dracoDecoderPath`.
+[examples/js/libs/basis][basis].
 
 
 `meshoptDecoderPath` path should be the complete file path (including filename) for a Meshopt decoder, typically named `meshopt_decoder.js`. Meshopt requires WebAssembly support. A CDN-hosted, versioned decoder is available at `https://unpkg.com/meshoptimizer@0.16.0/meshopt_decoder.js`, or you may download copies from the [meshoptimizer GitHub repository][meshopt-decoder].
--- a/docs/components/gltf-model.md
+++ b/docs/components/gltf-model.md
@@ -164,8 +164,7 @@ be included. A Google-hosted version of the Draco decoder libraries saves you fr
     basis_transcoder.js — JavaScript wrapper for the WebAssembly transcoder.
     basis_transcoder.wasm — WebAssembly transcoder.
 
-These files are available from the three.js repository, under
-[examples/js/libs/basis][basis].
+These files are available from the three.js repository in [`/examples/js/libs/basis`](https://github.com/mrdoob/three.js/tree/master/examples/js/libs/basis).
 
 
 `meshoptDecoderPath` path should be the complete file path (including filename) for a Meshopt decoder, typically named `meshopt_decoder.js`. Meshopt requires WebAssembly support. A CDN-hosted, versioned decoder is available at `https://unpkg.com/meshoptimizer@0.16.0/meshopt_decoder.js`, or you may download copies from the [meshoptimizer GitHub repository][meshopt-decoder].
--- a/docs/components/gltf-model.md
+++ b/docs/components/gltf-model.md
@@ -136,7 +136,7 @@ When using glTF models compressed with Draco, KTX2 or Meshopt, you must configur
 
 ```html
 <a-scene gltf-model="dracoDecoderPath: path/to/decoder/;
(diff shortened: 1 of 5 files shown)


docs/components/gltf-model.md
```diff
@@ -157,6 +159,15 @@ These files are available from the three.js repository, under
 automatically choose whether to use a WASM or JavaScript decoder, so both should
 be included. A Google-hosted version of the Draco decoder libraries saves you from needing to include these libraries in your own project: set `https://www.gstatic.com/draco/v1/decoders/` as the value for `dracoDecoderPath`.
 
+`ktx2TranscoderPath` path must be a folder containing two files:
+
+    basis_transcoder.js — JavaScript wrapper for the WebAssembly transcoder.
+    basis_transcoder.wasm — WebAssembly transcoder.
+
+These files are available from the three.js repository, under
+[examples/js/libs/basis][basis].
```
vincentfretin, line 168: I didn't see `[basis]` defined in the document, similar to `[blender]`. I guess you want to include something like
`[basis]: https://github.com/mrdoob/three.js/tree/dev/examples/js/libs/basis`

jameskane05: Good call, bad syntax. Corrected w/ link to that file folder.

dmarcos:
Thanks! How common is KTX2 these days? My only concern is adding this as a hard dependency if it's not used much.

jameskane05:
Yeah I noticed this comment above all the `require()`, I wonder how that could be achieved?

image

As for how common it is, the Mozilla Hubs team has implemented KTX2 and even ships the .wasm with their client (while this PR relies on optional remotely hosted .wasm and .js files). I know it's popular and relied-on by their team and community. As noted above, `gltf-transform` supports it, as well. I can also say my team uses KTX2 on most any diffuse texture or lightmap - the gains are really great, often 4-6x texture compression, and it stays (almost entirely) compressed when loaded in GPU memory, so if you're shipping high-def models to a WebXR headset it's pretty invaluable. I am sure adoption would grow if A-Frame supported it! Maybe Don could speak to its use at NYT?

donmccurdy:
I can't guess at number of models in the wild using KTX2, certainly models found in sources like Sketchfab won't include it. But KTX2 represents a 4-8x GPU memory savings compared to JPEG, PNG, or WebP — that's a big help in avoiding out-of-memory crashes on mobile devices. Size on disk/network is more variable, but often also improves. We do use KTX2 at the NYT, more for reasons of memory and GPU upload speed.

dmarcos:
Wonder if there's a way we can come up with to lazy load the dependencies when KTX2 is used

donmccurdy:
The larger WASM binary is already lazy-loaded, but the JS parts that are not so easy to fetch I expect.

Exairnous:
I'm an artist in the Hubs community and I try to use KTX2 whenever I can.  It's very helpful!

paradowskicreative:
Yeah, since it's the developer's decision to include/exclude `basisTranscoderUrl` in their code at runtime, I'm not sure how we could determine at build time whether or not these loader scripts should be included w/ the A-Frame dist.

Anyway, A-Frame includes Draco in the same way, and to me KTX2 is just as useful, if not moreso b/c it optimizes memory use, not just network/disk use. Supporting the most popular geometry and texture compression libraries side for gltf by side makes a lot of sense, and it can't grow in popularity if major frameworks don't support it. If something better and more standardized comes along this can always be removed in its favor. The cost of a few dormant lines of JS when not in use seems pretty low. Just my two cents!

dmarcos:
Thanks! Great work!

vincentfretin:
This PR was working with super-three r141, but in the meantime aframe master switched to super-three r144 and the UMD build of `examples/js/KTX2Loader.js` was removed in r144, see https://github.com/mrdoob/three.js/pull/24398 :sob: 
I think we really need to switch to a maintained build tool that is supporting ESM.

donmccurdy:
Also see https://github.com/mrdoob/three.js/issues/24593#issuecomment-1238550059 — more UMD+CJS files may be removed by the end of this year. I believe an ESM-based build process would be ideal looking forward.

donmccurdy:
Great work here, thank you @jameskane05!

vincentfretin:
I spent 9 hours today :sob: to switch aframe to webpack #5116 to fix master that is currently broken. Please merge it @dmarcos :pray:

kylebakerio:
Thanks so much for doing that work veincent. We're all in debt to you!

Issue #5067, opened 2022-06-24
Oculus Quest 2 animations on trigger / grip / button press don't work


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: Oculus Quest 2
- Reproducible Code Snippet or URL: Any A-Frame Example that uses controllers.  E.g. https://aframe.io/examples/showcase/responsiveui/

When I press the trigger or grip, the trigger & grip on the controller models as viewed in VR.  Similarly, changes to button colors (as configured via buttonColor, buttonHighlightColor, buttonTouchColor) are not applied.

I hadn't realized this was expected function, but looking at the code in [oculus-touch-controls](https://github.com/aframevr/aframe/blob/master/src/components/oculus-touch-controls.js) there is code to animate the trigger & grip & change colors of buttons.  It looks like it would work for the Oculus Quest 1, but hasn't been updated to work for Oculus Quest 2.

The key bit of code is [here](https://github.com/aframevr/aframe/blob/a7d1154146e915b0e739bcd761457ff9ed097776/src/components/oculus-touch-controls.js#L302-L310), which uses getObjectByName to access the relevant sub-components of the controller 3D model.

I've been looking at the controller models in a GLTF viewer.

the "gen2" model (used for Quest 1) has subcomponents with names that match the code.
![image](https://user-images.githubusercontent.com/16045703/175561716-ff686227-4339-42bf-ab9f-0758664c6349.png)

the "v3" model (used for Quest 2) has a different set of names for the subcomponents (as well as a much more complicated hierarchy).
![image](https://user-images.githubusercontent.com/16045703/175562589-a7f95ba1-0c6a-4538-985b-e6be216c31ba.png)

Because the names don't match, the calls to getObjectByName all fail, and the component has no references in buttonMeshes, for example [here](https://github.com/aframevr/aframe/blob/a7d1154146e915b0e739bcd761457ff9ed097776/src/components/oculus-touch-controls.js#L281-L288) and [here](https://github.com/aframevr/aframe/blob/a7d1154146e915b0e739bcd761457ff9ed097776/src/components/oculus-touch-controls.js#L337-L341)

I haven't looked into what's involved in fixing yet.  It might be a matter of just using the alternate names on the calls to getObjectByName.  Or it's possible that the details of exactly what rotations & position offsets need to be applied also need to be adjusted to reflect the different controller model.

[...]

PR #5103, opened 2022-08-27
Quest 2 controller fixes (fixes issue #5607)

kylebakerio:
See [issue 5607](https://github.com/aframevr/aframe/issues/5067).

**Features:**
All Oculus controllers:
- fixes bug where all buttons would light up if any button was pressed
- untouched color state is now mesh's original color

Quest 2 controllers:
- all of the above
- correct joystick model movement
- moving a/b/x/y buttons

**Description:**
- moved onButtonChanged to bindMethods
- added onThumbstickMoved to support thumbstick model updates
- added recycled euler for trigger updates, which are performed as a multi-dimensional rotation for v3 controller model
- added thumbstickMoved event listener, and matching handlers
- slight comment cleanup in a few places
- added onButtonChangedV3 and onModelLoadedV3 for handling non-conforming model cleanly while maintaining backwards compatibility unambiguously
- added multiMeshFix to fix issue that all oculus controlers have had, where all buttons light up if any button is pressed
- update to use model default color as button color when not being touched

I've tested all four controller models using the this pre-merge version of the WebXR Emulator Extension: https://github.com/MozillaReality/WebXR-emulator-extension/pull/278

I have run a build with this branch and use that built version of A-Frame that includes this change here:
https://glitch.com/~quest-2-controller-responsive-buttons

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -470,7 +470,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 });
 
 /**
- * A-Frame's material component makes a single unified shared material for all meshes.
+ * A-Frame's material component makes a single unified shared material for all child meshes.
  * In order to be able to color individual buttons, we need to give each sub-mesh its own
  * THREE material. Without this, changing the color on any individual button mesh changes
  * the color on every button mesh.
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -172,7 +172,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.previousButtonValues = {};
     this.rendererSystem = this.el.sceneEl.systems.renderer;
     this.bindMethods();
-    this.triggerEuler = new THREE.Euler;
+    this.triggerEuler = new THREE.Euler();
   },
 
   addEventListeners: function () {
@@ -275,10 +275,9 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
   onButtonChanged: function (evt) {
     // move the button models
-    if (this.isV3) { 
+    if (this.isV3) {
       this.onButtonChangedV3(evt);
-    }
-    else {  
+    } else {
       var button = this.mapping[this.data.hand].buttons[evt.detail.id];
       var buttonMeshes = this.buttonMeshes;
       var analogValue;
@@ -300,7 +299,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     this.el.emit(button + 'changed', evt.detail.state);
   },
 
-  clickButtons: ['xbutton','ybutton','abutton','bbutton','thumbstick'],
+  clickButtons: ['xbutton', 'ybutton', 'abutton', 'bbutton', 'thumbstick'],
   onButtonChangedV3: function (evt) {
     var button = this.mapping[this.data.hand].buttons[evt.detail.id];
     var buttonObjects = this.buttonObjects;
@@ -309,22 +308,19 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
     analogValue = evt.detail.state.value;
     analogValue *= this.data.hand === 'left' ? -1 : 1;
-    
+
     if (button === 'trigger') {
       this.triggerEuler.copy(this.buttonRanges.trigger.min.rotation);
       this.triggerEuler.x += analogValue * this.buttonRanges.trigger.diff.x;
       this.triggerEuler.y += analogValue * this.buttonRanges.trigger.diff.y;
       this.triggerEuler.z += analogValue * this.buttonRanges.trigger.diff.z;
       buttonObjects.trigger.setRotationFromEuler(this.triggerEuler);
-    }
-    else if (button === 'grip') {
+    } else if (button === 'grip') {
       buttonObjects.grip.position.x = buttonObjects.grip.minX + analogValue * 0.004;
+    } else if (this.clickButtons.includes(button)) {
+      buttonObjects[button].position.y = analogValue === 0 ? this.buttonRanges[button].unpressedY : this.buttonRanges[button].pressedY;
     }
-    else if (this.clickButtons.includes(button)) {
-      buttonObjects[button].position.y = analogValue === 0 ? 
-        this.buttonRanges[button].unpressedY : this.buttonRanges[button].pressedY;
-    }
-  },  
+  },
 
   onModelLoaded: function (evt) {
     if (this.isV3) {
@@ -352,7 +348,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       this.multiMeshFix(button);
     }
 
-    this.applyOffset(evt.detail.model)
+    this.applyOffset(evt.detail.model);
 
     this.el.emit('controllermodelready', {
       name: 'oculus-touch-controls',
@@ -360,16 +356,16 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       rayOrigin: this.displayModel[this.data.hand].rayOrigin
     });
   },
-  
-  applyOffset(model) {
+
+  applyOffset (model) {
     model.position.copy(this.displayModel[this.data.hand].modelPivotOffset);
     model.rotation.copy(this.displayModel[this.data.hand].modelPivotRotation);
   },
-  
-  multiMeshFix(button) {
+
+  multiMeshFix (button) {
     if (!this.buttonMeshes[button]) return;
     this.buttonMeshes[button].traverse((node) => {
-      if (node.type != "Mesh") return;
+      if (node.type !== 'Mesh') return;
       let originalMaterial = node.material;
       this.buttonMeshes[button].naturalColor = originalMaterial.color;
       node.material = new THREE.MeshStandardMaterial(); // {color: originalMaterial.color}
@@ -380,37 +376,37 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
           // because A-Frame uses the same single material object for all material nodes, it also uses the
           // same color. This means they use the same color object--we have to break that link. Other links
           // likely also exist, but they don't matter to us, because color is the only one we change.
-          node.material[key] = key !== 'color' ? originalMaterial[key] :  originalMaterial[key].clone();
+          node.material[key] = key !== 'color' ? originalMaterial[key] : originalMaterial[key].clone();
         }
-      })
+      });
       originalMaterial.dispose();
-    })
+    });
   },
 
-  onModelLoadedV3(evt) {
+  onModelLoadedV3 (evt) {
     let controllerObject3D = this.controllerObject3D = evt.detail.model;
-    
+
     if (!this.data.model) { return; }
-    
-    let buttonObjects = this.buttonObjects = {},
-        buttonMeshes = this.buttonMeshes = {},
-        buttonRanges = this.buttonRanges = {};
-    
+
+    let buttonObjects = this.buttonObjects = {};
+    let buttonMeshes = this.buttonMeshes = {};
+    let buttonRanges = this.buttonRanges = {};
+
     buttonMeshes.grip = controllerObject3D.getObjectByName('squeeze');
     buttonObjects.grip = controllerObject3D.getObjectByName('xr_standard_squeeze_pressed_value');
     buttonRanges.grip = {
       min: controllerObject3D.getObjectByName('xr_standard_squeeze_pressed_min'),
-      max: controllerObject3D.getObjectByName('xr_standard_squeeze_pressed_max'),
-    }
+      max: controllerObject3D.getObjectByName('xr_standard_squeeze_pressed_max')
+    };
     buttonObjects.grip.minX = buttonObjects.grip.position.x;
-    
+
     buttonMeshes.thumbstick = controllerObject3D.getObjectByName('thumbstick');
     buttonObjects.thumbstick = controllerObject3D.getObjectByName('xr_standard_thumbstick_pressed_value');
     buttonRanges.thumbstick = {
       min: controllerObject3D.getObjectByName('xr_standard_thumbstick_pressed_min'),
       max: controllerObject3D.getObjectByName('xr_standard_thumbstick_pressed_max'),
-      originalRotation: this.buttonObjects.thumbstick.rotation.clone(),
-    }
+      originalRotation: this.buttonObjects.thumbstick.rotation.clone()
+    };
     buttonRanges.thumbstick.pressedY = buttonObjects.thumbstick.position.y;
     buttonRanges.thumbstick.unpressedY =
       buttonRanges.thumbstick.pressedY + Math.abs(buttonRanges.thumbstick.max.position.y) - Math.abs(buttonRanges.thumbstick.min.position.y);
@@ -419,57 +415,57 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     buttonObjects.trigger = controllerObject3D.getObjectByName('xr_standard_trigger_pressed_value');
     buttonRanges.trigger = {
       min: controllerObject3D.getObjectByName('xr_standard_trigger_pressed_min'),
-      max: controllerObject3D.getObjectByName('xr_standard_trigger_pressed_max'),
-    }
+      max: controllerObject3D.getObjectByName('xr_standard_trigger_pressed_max')
+    };
     buttonRanges.trigger.diff = {
       x: Math.abs(buttonRanges.trigger.max.rotation.x) - Math.abs(buttonRanges.trigger.min.rotation.x),
       y: Math.abs(buttonRanges.trigger.max.rotation.y) - Math.abs(buttonRanges.trigger.min.rotation.y),
-      z: Math.abs(buttonRanges.trigger.max.rotation.z) - Math.abs(buttonRanges.trigger.min.rotation.z),      
-    }
-    
-    let btn1 = this.data.hand === 'left' ? 'x' : 'a',
-        btn2 = this.data.hand === 'left' ? 'y' : 'b' ;
-    
-    buttonMeshes[btn1+"button"] = controllerObject3D.getObjectByName(btn1+'_button');
-    buttonObjects[btn1+"button"] = controllerObject3D.getObjectByName(btn1+'_button_pressed_value');
-    buttonRanges[btn1+"button"] = {
-      min: controllerObject3D.getObjectByName(btn1+'_button_pressed_min'),
-      max: controllerObject3D.getObjectByName(btn1+'_button_pressed_max'),
-    }
+      z: Math.abs(buttonRanges.trigger.max.rotation.z) - Math.abs(buttonRanges.trigger.min.rotation.z)
+    };
 
-    buttonMeshes[btn2+"button"] = controllerObject3D.getObjectByName(btn2+'_button');
-    buttonObjects[btn2+"button"] = controllerObject3D.getObjectByName(btn2+'_button_pressed_value');
-    buttonRanges[btn2+"button"] = {
-      min: controllerObject3D.getObjectByName(btn2+'_button_pressed_min'),
-      max: controllerObject3D.getObjectByName(btn2+'_button_pressed_max'),
-    }    
-    
-    buttonRanges[btn1+"button"].unpressedY = buttonObjects[btn1+"button"].position.y;
-    buttonRanges[btn1+"button"].pressedY = 
-      buttonRanges[btn1+"button"].unpressedY + Math.abs(buttonRanges[btn1+"button"].max.position.y) - Math.abs(buttonRanges[btn1+"button"].min.position.y); 
-    
-    buttonRanges[btn2+"button"].unpressedY = buttonObjects[btn2+"button"].position.y;
-    buttonRanges[btn2+"button"].pressedY = 
-      buttonRanges[btn2+"button"].unpressedY - Math.abs(buttonRanges[btn2+"button"].max.position.y) + Math.abs(buttonRanges[btn2+"button"].min.position.y); 
+    let btn1 = this.data.hand === 'left' ? 'x' : 'a';
+    let btn2 = this.data.hand === 'left' ? 'y' : 'b';
+
+    buttonMeshes[btn1 + 'button'] = controllerObject3D.getObjectByName(btn1 + '_button');
+    buttonObjects[btn1 + 'button'] = controllerObject3D.getObjectByName(btn1 + '_button_pressed_value');
+    buttonRanges[btn1 + 'button'] = {
+      min: controllerObject3D.getObjectByName(btn1 + '_button_pressed_min'),
+      max: controllerObject3D.getObjectByName(btn1 + '_button_pressed_max')
+    };
+
+    buttonMeshes[btn2 + 'button'] = controllerObject3D.getObjectByName(btn2 + '_button');
+    buttonObjects[btn2 + 'button'] = controllerObject3D.getObjectByName(btn2 + '_button_pressed_value');
+    buttonRanges[btn2 + 'button'] = {
+      min: controllerObject3D.getObjectByName(btn2 + '_button_pressed_min'),
+      max: controllerObject3D.getObjectByName(btn2 + '_button_pressed_max')
+    };
+
+    buttonRanges[btn1 + 'button'].unpressedY = buttonObjects[btn1 + 'button'].position.y;
+    buttonRanges[btn1 + 'button'].pressedY =
+      buttonRanges[btn1 + 'button'].unpressedY + Math.abs(buttonRanges[btn1 + 'button'].max.position.y) - Math.abs(buttonRanges[btn1 + 'button'].min.position.y);
+
+    buttonRanges[btn2 + 'button'].unpressedY = buttonObjects[btn2 + 'button'].position.y;
+    buttonRanges[btn2 + 'button'].pressedY =
+      buttonRanges[btn2 + 'button'].unpressedY - Math.abs(buttonRanges[btn2 + 'button'].max.position.y) + Math.abs(buttonRanges[btn2 + 'button'].min.position.y);
   },
-  
+
   onAxisMoved: function (evt) {
     emitIfAxesChanged(this, this.mapping[this.data.hand].axes, evt);
   },
-  
-  onThumbstickMoved: function(evt) {
-    if (!this.isV3 || !this.buttonMeshes || !this.buttonMeshes.thumbstick) {return;}
+
+  onThumbstickMoved: function (evt) {
+    if (!this.isV3 || !this.buttonMeshes || !this.buttonMeshes.thumbstick) { return; }
     for (let axis in evt.detail) {
-      this.buttonObjects.thumbstick.rotation[this.axisMap[axis]] = 
-        this.buttonRanges.thumbstick.originalRotation[this.axisMap[axis]] - 
-        (Math.PI*1/8) *
-        evt.detail[axis] * 
+      this.buttonObjects.thumbstick.rotation[this.axisMap[axis]] =
+        this.buttonRanges.thumbstick.originalRotation[this.axisMap[axis]] -
+        (Math.PI * 1 / 8) *
+        evt.detail[axis] *
         (axis === 'y' || this.data.hand === 'right' ? -1 : 1);
     }
   },
   axisMap: {
     y: 'x',
-    x: 'z',
+    x: 'z'
   },
 
   updateModel: function (buttonName, evtName) {
@@ -481,11 +477,11 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   updateButtonModel: function (buttonName, evtName) {
     var button;
     var color = (evtName === 'up' || evtName === 'touchend') ? this.buttonMeshes[buttonName].naturalColor || this.data.buttonColor : evtName === 'touchstart' ? this.data.buttonTouchColor : this.data.buttonHighlightColor;
-    
+
     if (this.buttonMeshes && this.buttonMeshes[buttonName]) {
       button = this.buttonMeshes[buttonName];
       button.material.color.set(color);
       this.rendererSystem.applyColorCorrection(button.material.color);
     }
-  },
+  }
 });
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -242,7 +242,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       }
     }
     var modelUrl = this.displayModel[data.hand].modelUrl;
-    this.isV3 = this.displayModel === CONTROLLER_PROPERTIES['oculus-touch-v3'];
+    this.isOculusTouchV3 = this.displayModel === CONTROLLER_PROPERTIES['oculus-touch-v3'];
     this.el.setAttribute('gltf-model', modelUrl);
   },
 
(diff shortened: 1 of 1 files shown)


vincentfretin:
I looked quickly at your code changes, I don't have anything to say actually code wise, the changes looks good.

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
   },
 
+  applyOffset (model) {
+    model.position.copy(this.displayModel[this.data.hand].modelPivotOffset);
+    model.rotation.copy(this.displayModel[this.data.hand].modelPivotRotation);
+  },
+
+  multiMeshFix (button) {
+    if (!this.buttonMeshes[button]) return;
+    this.buttonMeshes[button].traverse((node) => {
```
dmarcos, line 367: use regular function. A-Frame is ES5

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+      if (node.type !== 'Mesh') return;
+      let originalMaterial = node.material;
```
dmarcos, line 369: use `var` for consistency

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    buttonMeshes[btn2 + 'button'] = controllerObject3D.getObjectByName(btn2 + '_button');
+    buttonObjects[btn2 + 'button'] = controllerObject3D.getObjectByName(btn2 + '_button_pressed_value');
+    buttonRanges[btn2 + 'button'] = {
+      min: controllerObject3D.getObjectByName(btn2 + '_button_pressed_min'),
+      max: controllerObject3D.getObjectByName(btn2 + '_button_pressed_max')
+    };
+
+    buttonRanges[btn1 + 'button'].unpressedY = buttonObjects[btn1 + 'button'].position.y;
+    buttonRanges[btn1 + 'button'].pressedY =
+      buttonRanges[btn1 + 'button'].unpressedY + Math.abs(buttonRanges[btn1 + 'button'].max.position.y) - Math.abs(buttonRanges[btn1 + 'button'].min.position.y);
```
dmarcos, line 445: can consolidate this in a variable:

```
 var idButton1 = button1 + 'button';
```
Also for consistency use complete words. btn1 -> button1

kylebakerio:
Thanks for the review, I've made the requested changes.

Out of curiousity: is there any chance of A-Frame allowing ES6 in the future?

dmarcos:
> Thanks for the review, I've made the requested changes.
> 
> Out of curiousity: is there any chance of A-Frame allowing ES6 in the future?

Thanks for the patience. I'll do another pass as soon as I can.

No plans to move to ES6 at the moment. I like to keep the code consistent and like the simplicity of ES5.

kylebakerio:
I see. I wonder at what point knowing what counts as ES5 vs what counts as ES6 will be considered more esoteric knowledge than knowing ES6. I imagine all newbies these days learn javascript with ES6 just taught as a normal part of the language without knowing that we had to live without it before 2017~.

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,130 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+
+  multiMeshFix: function (button) {
+    if (!this.buttonMeshes[button]) return;
+    this.buttonMeshes[button].traverse((node) => {
+      if (node.type !== 'Mesh') return;
+      var originalMaterial = node.material;
+      this.buttonMeshes[button].naturalColor = originalMaterial.color;
+      node.material = new THREE.MeshStandardMaterial();
+
+      // preserve details; not iterable, so we have to use Object.keys()
```
vincentfretin, line 373: Why are you saying it's not iterable? What did you try?
You gave me some doubt here so I add to verify again my own code. :D 
I have other code that does this:
```
for (const propertyName in material) {
    const texture = material[propertyName];
    if (texture instanceof THREE.Texture) {
      const image = texture.source.data;
      if (image instanceof ImageBitmap) {
        image.close && image.close();
      }
      texture.dispose();
    }
  }
```
It's definitely enumerable.

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,130 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+      Object.keys(originalMaterial).forEach(key => {
+        if (originalMaterial[key] !== node.material[key]) {
+          // because A-Frame uses the same single material object for all material nodes, it also uses the
+          // same color. This means they use the same color object--we have to break that link. Other links
+          // likely also exist, but they don't matter to us, because color is the only one we change.
+          node.material[key] = key !== 'color' ? originalMaterial[key] : originalMaterial[key].clone();
+        }
+      });
+      originalMaterial.dispose();
+    });
```
vincentfretin, line 383: So in the end you just want to clone the material, using the same textures, but have a different color instance, right?
Material has a copy method that do just this, so you can use it and just clone the color property afterwards. This should work:
```
node.material = new THREE.MeshStandardMaterial();
node.material.copy(originalMaterial);
node.material.color = originalMaterial.color.clone();
originalMaterial.dispose();
```

src/components/oculus-touch-controls.js
kylebakerio: You're right. I'm not sure how that happened, but my first attempt using for-in didn't sreem to show the keys I needed, so I switched to this and it worked. It may have just been a quick caching issue and a resulting assumption? But retrying this, you're definitely correct.

That said, for-in loops are ES6... a bit confusing, since for-in is recommended as a best practice in the aframe docs if I'm not mistaken, as it avoids an unnecessary array generation? Not terribly relevant here since this runs once, of course.

And all that said, seems the copy method you suggest will be the best option. This iteration worked, but I'm definitely glad to find a more efficient way to do this! I was riffing on an example I found elsewhere--I tried to make it more efficient and fit to my example, then things broke, then I fixed the things that broke. 

While I think this will be slightly less efficient, it is cleaner code, and also should 'fix' any other 'links' between the child materials--not relevant right now, but makes this more future proof and makes the code a better example.

src/components/oculus-touch-controls.js
kylebakerio: I opted for a variation on this approach:
```js
      let newMaterial = node.material.clone();
      this.buttonMeshes[button].naturalColor = node.material.color;
      node.material.dispose();
      node.material = newMaterial;
```

No special handling is needed for color now, since we're calling `clone()` it seems to recursivley clone all the child elements as well, I imagine / it seems. Again, probably slight efficiency reduction, but definitely better code.

kylebakerio:
(that last commit is not just a linter commit, it's also the fix in response to vincent's comment--I then had to make some linter recommended changes and didn't catch that my prior commit that triggered the linter didn't go through.)

src/components/oculus-touch-controls.js
vincentfretin: for-in is ES5, for-of is ES6.
https://hacks.mozilla.org/2015/04/es6-in-depth-iterators-and-the-for-of-loop/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

Aframe actually use some ES6 features in some places and those changes were actually introduced by @dmarcos ;-). You can easily see where because the minify we use doesn't support ES6 syntax. I created #5106 for this.

kylebakerio:
@dmarcos just giving a quick ping on this

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,121 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
   },
 
+  applyOffset: function (model) {
+    model.position.copy(this.displayModel[this.data.hand].modelPivotOffset);
+    model.rotation.copy(this.displayModel[this.data.hand].modelPivotRotation);
+  },
+
+  multiMeshFix: function (button) {
+    if (!this.buttonMeshes[button]) return;
+    this.buttonMeshes[button].traverse((node) => {
```
dmarcos, line 367: no arrow function for consistency

dmarcos: use regular function and `var self = this;` pattern

src/components/oculus-touch-controls.js
```diff
@@ -220,25 +224,25 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+      } else {
+        // WebXR
         controllerId = CONTROLLER_DEFAULT;
         controllerId = controller.profiles.indexOf('oculus-touch-v2') !== -1 ? 'oculus-touch-v2' : controllerId;
         controllerId = controller.profiles.indexOf('oculus-touch-v3') !== -1 ? 'oculus-touch-v3' : controllerId;
         this.displayModel = CONTROLLER_PROPERTIES[controllerId];
       }
     }
     var modelUrl = this.displayModel[data.hand].modelUrl;
+    this.isV3 = this.displayModel === CONTROLLER_PROPERTIES['oculus-touch-v3'];
```
dmarcos, line 245: rename to `isOculusTouchV3` to make it easier to understand

src/components/oculus-touch-controls.js
```diff
@@ -270,48 +274,81 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+      buttonObjects.grip.position.x = buttonObjects.grip.minX + analogValue * 0.004;
+    } else if (this.clickButtons.includes(button)) {
+      buttonObjects[button].position.y = analogValue === 0 ? this.buttonRanges[button].unpressedY : this.buttonRanges[button].pressedY;
+    }
+  },
 
-    buttonMeshes = this.buttonMeshes = {};
+  onModelLoaded: function (evt) {
+    if (this.isV3) {
+      this.onModelLoadedV3(evt);
```
dmarcos, line 327: onOculusTouchV3ModelLoaded

src/components/oculus-touch-controls.js
```diff
@@ -270,48 +274,81 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    } else {
+      var controllerObject3D = this.controllerObject3D = evt.detail.model;
```
dmarcos, line 329: Probably can factor out this part to a separate function. onOculusTouchV1ModelLoaded?

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,121 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     });
   },
 
+  applyOffset: function (model) {
+    model.position.copy(this.displayModel[this.data.hand].modelPivotOffset);
+    model.rotation.copy(this.displayModel[this.data.hand].modelPivotRotation);
+  },
+
+  multiMeshFix: function (button) {
```
dmarcos, line 365: What does this function do? Maybe we can find a better name? changeButtonColor?

kylebakerio: This function patches the A-Frame mesh material issue, Diarmid mentions it here: https://github.com/aframevr/aframe/issues/5067#issuecomment-1223931781

Basically, in A-Frame, if you try to color any 'child' mesh, all the meshes receive that color. This replaces all meshes with raw THREE meshes so that they have separate color properties.

src/components/oculus-touch-controls.js
kylebakerio: Well, to be clear, this handles three versions of oculus headsets that came before "v3". Seems if you call it v1 is might be confusing?

src/components/oculus-touch-controls.js
kylebakerio: will do

src/components/oculus-touch-controls.js
kylebakerio: will do

src/components/oculus-touch-controls.js
kylebakerio: will do

src/components/oculus-touch-controls.js
kylebakerio: (I'm also interested in looking into making a fix for that underlying issue, but haven't looked into that part of A-Frame's source yet.)

kylebakerio: I wouldn't call it "changeButtonColor" because button color changes happen in `updateButtonModel()`. This function just applies a one-time patch to swap out the mesh materials to enable it.

kylebakerio:
@dmarcos made requested changes on 3 out of 5, responses on the other two.

src/components/oculus-touch-controls.js
diarmidmackenzie: I'd suggest a more descriptive function name might be

`createOneMaterialPerMesh`

or maybe

`createUniqueMaterialForEachMesh`

diarmidmackenzie: I don't understand why this code acts on `this.buttonMeshes[button]` rather than just `button`...?

I think it would be cleaner for this bit of code to make no assumptions about `this.buttonMeshes` and just act on whatever object3D it is passed?

src/components/oculus-touch-controls.js
dmarcos: I see. Maybe a comment in the branch will help clarify

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,122 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
   },
 
-  updateButtonModel: function (buttonName, state) {
+  updateButtonModel: function (buttonName, evtName) {
+    // update the button mesh colors
     var button;
-    var color = (state === 'up' || state === 'touchend') ? this.data.buttonColor : state === 'touchstart' ? this.data.buttonTouchColor : this.data.buttonHighlightColor;
+    var color = (evtName === 'up' || evtName === 'touchend') ? this.buttonMeshes[buttonName].naturalColor || this.data.buttonColor : evtName === 'touchstart' ? this.data.buttonTouchColor : this.data.buttonHighlightColor;
     var buttonMeshes = this.buttonMeshes;
-    if (!this.data.model) { return; }
```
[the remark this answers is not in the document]
kylebakerio: `updateButtonModel` is called by `updateModel()` which already does this check. (`updateModel()` is the external-facing function that is called by the trackedControls utils).

src/components/oculus-touch-controls.js
kylebakerio: > I don't understand why this code acts on `this.buttonMeshes[button]` rather than just `button`...?
> 
> I think it would be cleaner for this bit of code to make no assumptions about `this.buttonMeshes` and just act on whatever object3D it is passed?

Sure. That's a good idea. This was just the result of me slowly wittling down a proof of concept piece by piece, and just designed it for my needs here. One step further would generalize it completely.

> yep. probably we can add this function to utils if it's generic enough.

Yeah, with that change it should be, and that would probably be a good idea. Should I add that to this pull request and switch to using that?

src/components/oculus-touch-controls.js
kylebakerio: Do you mean just a line between 328 and 329 like
`// All oculus headset controller models prior to the Quest 2 (i.e., Oculus Touch V3) used a consistent format that is handled here` ?

Or something else / somewhere else?

dmarcos: yep

src/components/oculus-touch-controls.js
dmarcos: probably I would make it an independent function at the bottom of the component like [this one](https://github.com/aframevr/aframe/blob/master/src/components/material.js#L263). Make it operate over the object3D, add a descriptive name (cloneMeshMaterials?) and comment / document the purpose

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,122 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+  },
+  axisMap: {
+    y: 'x',
+    x: 'z'
+  },
+
   updateModel: function (buttonName, evtName) {
-    if (!this.data.model) { return; }
-    this.updateButtonModel(buttonName, evtName);
+    if (this.data.model) {
```
dmarcos, line 465: Keep early return for style consistency (less indentation, more readable code) `if (!this.data.model) { return; }`

src/components/oculus-touch-controls.js
```diff
@@ -270,48 +274,81 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
 
-    buttonMeshes = this.buttonMeshes = {};
+  onModelLoaded: function (evt) {
+    if (this.isOculusTouchV3) {
+      this.onOculusTouchV3ModelLoaded(evt);
+    } else {
+      var controllerObject3D = this.controllerObject3D = evt.detail.model;
+      var buttonMeshes;
+
+      if (!this.data.model) { return; }
```
dmarcos, line 332: This line could be moved at the top of the function

src/components/oculus-touch-controls.js
```diff
@@ -320,20 +357,122 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+      self.buttonMeshes[button].naturalColor = node.material.color;
+      node.material.dispose();
+      node.material = newMaterial;
+    });
+  },
+
+  onOculusTouchV3ModelLoaded: function (evt) {
+    var controllerObject3D = this.controllerObject3D = evt.detail.model;
+
+    if (!this.data.model) { return; }
```
dmarcos, line 380: this one could be removed if we do https://github.com/aframevr/aframe/pull/5103/files#r966498190

src/components/oculus-touch-controls.js
kylebakerio: I agree! I saw a pattern where it seemed whoever wrote this _really_ liked having var declarations at the top of functions, but I am in favor of that change and have no such attachment if you don't.

src/components/oculus-touch-controls.js
kylebakerio: Yup, also noticed that, and agreed.

src/components/oculus-touch-controls.js
kylebakerio: Ok. I felt like this change improved readability, but I understand that that's subjective and that the other is indeed a pattern. Will change it back.

src/components/oculus-touch-controls.js
kylebakerio: Sounds good.

kylebakerio:
@dmarcos have made the requested changes.

kylebakerio:
I'm not sure what to make of that fail...?

```
/usr/bin/bash /home/runner/work/_actions/GabrielBB/xvfb-action/v1/cleanup.sh
Killing the following xvfb processes: 2112
Error: The process '/usr/bin/xvfb-run' failed with exit code 1
```

Looks like the cleanup process failed.

While this test supposedly should be about firefox, I went as far as getting a full build of aframe at this state, modifying the glitch to use that directly, installing the webxr emulator extension linked at the top of this pull request into firefox, and testing it. It worked and reported no console errors in firefox. Seems like a build script that needs to be fixed.
Screenshot from 2022-09-10 18-29-31
Screenshot from 2022-09-10 18-29-20

src/components/oculus-touch-controls.js
```diff
@@ -320,24 +360,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+ * In order to be able to color individual buttons, we need to give each sub-mesh its own
+ * THREE material. Without this, changing the color on any individual button mesh changes
+ * the color on every button mesh.
+ */
+
+function cloneMaterialToTHREE (object3d) {
+  object3d.traverse(function (node) {
+    if (node.type !== 'Mesh') return;
+    let newMaterial = node.material.clone();
+    object3d.naturalColor = node.material.color;
```
dmarcos, line 483: I would call it originalColor

src/components/oculus-touch-controls.js
```diff
@@ -320,24 +360,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
     if (buttonMeshes && buttonMeshes[buttonName]) {
       button = buttonMeshes[buttonName];
       button.material.color.set(color);
       this.rendererSystem.applyColorCorrection(button.material.color);
     }
   }
 });
+
+/**
+ * A-Frame's material component makes a single unified shared material for all meshes.
```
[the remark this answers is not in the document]
kylebakerio: Hm, personally, I'm not sure; Diarmid suggested this was a characteristic of A-Frame itself (perhaps a characteristic of the GLTF loader?), but I took his word on that and just worked backwards from there to come up with this fix. I haven't personally deeply investigated what A-Frame does, or whether this is something that is encoded in the GLTF.

I do know that I can color button mesh materials individually in a babylon sandbox viewer, which suggested to me that it was about the gltf loader/ a-frame, and not something with the model.

Happy to update the comment with more correct information if you have any insight into what is going on more precisely here.

src/components/oculus-touch-controls.js
kylebakerio: sounds equally valid to me, done.

src/components/oculus-touch-controls.js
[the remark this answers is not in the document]
kylebakerio: @diarmidmackenzie if I've worded this poorly or incorrectly, or if you have more insight into explaining the issue you've encountered here, let me know.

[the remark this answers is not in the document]
kylebakerio: Also of note: all oculus models had this problem where all button meshes would get re-colored if you colored any of them.

> fixes bug where all buttons would light up if any button was pressed

Perhaps all oculus controllers' models had the same issue? there are other controllers with other components with related code that _did_ work, but I haven't checked, maybe those were the older ones that weren't gltf.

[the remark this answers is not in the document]
vincentfretin: The comment is definitely wrong in this context.
@kylebakerio I think the comment references the issue described in this comment https://github.com/aframevr/aframe/issues/5067#issuecomment-1224095532 ? But I think this is a complete different issue.
The aframe geometry or material component are not even used here. The model is loaded via the gltf-model component that just use THREE.GLTFLoader. I don't think there is an issue with GLTFLoader either. If you open https://cdn.aframe.io/controllers/oculus/oculus-touch-controller-gen2-right.gltf in https://gltf-viewer.donmccurdy.com/ you will see 7 draw calls and 1 material, meaning 7 meshes and 1 material. If you import it in blender you will see that the same material is used in 7 meshes. Same issue with oculus-touch-controller-v3-right.glb where 1 material is used by 6 meshes.
The GearVR controller is an obj/mtl and has 3 meshes, each mesh with its own material.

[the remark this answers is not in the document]
kylebakerio: I see. So the 'problem' is the model, not the issue with a-frame mentioned there (if it actually exists?), then. Will update comment to reflect this, then.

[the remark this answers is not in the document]
diarmidmackenzie: Seems it is not relevant here but there is a limitation in the A-Frame material component.

Specifically, a Three JS Mesh can have multiple materials specified as an array, as described here.

https://threejs.org/docs/index.html#api/en/objects/Mesh.material

When this happens, they are applied in turn to the different groups in the geometry.

The A-Frame material component doesn't support this, and always a single material for the whole object, which is applied to all the groups in the geometry.

However all this is only relevant when working with geometry and materials, so it sounds like it's not relevant in this context.

src/components/oculus-touch-controls.js
```diff
@@ -320,24 +360,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
 });
+
+/**
+ * A-Frame's material component makes a single unified shared material for all child meshes.
+ * In order to be able to color individual buttons, we need to give each sub-mesh its own
+ * THREE material. Without this, changing the color on any individual button mesh changes
+ * the color on every button mesh.
+ */
+
+function cloneMaterialToTHREE (object3d) {
```
dmarcos, line 479: `cloneMeshMaterials` probably a more descriptive name

src/components/oculus-touch-controls.js
```diff
@@ -220,25 +224,25 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     if (!data.model) { return; }
     // Set the controller display model based on the data passed in.
     this.displayModel = CONTROLLER_PROPERTIES[data.controllerType] || CONTROLLER_PROPERTIES[CONTROLLER_DEFAULT];
-    // If the developer is asking for auto-detection, see if the displayName can be retrieved to identify the specific unit.
-    // This only works for WebVR currently.
```
dmarcos, line 224: this comment is gone

src/components/oculus-touch-controls.js
```diff
@@ -220,25 +224,25 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    // If the developer is asking for auto-detection, use the retrieved displayName to identify the specific unit.
     if (data.controllerType === 'auto') {
       var trackedControlsSystem = this.el.sceneEl.systems['tracked-controls-webvr'];
-      // WebVR
       if (trackedControlsSystem && trackedControlsSystem.vrDisplay) {
+        // WebVR
```
dmarcos, line 231: this comment is repositioned.

src/components/oculus-touch-controls.js
```diff
@@ -220,25 +224,25 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
         var displayName = trackedControlsSystem.vrDisplay.displayName;
-        // The Oculus Quest uses the updated generation 2 inside-out tracked controllers so update the displayModel.
         if (/^Oculus Quest$/.test(displayName)) {
           this.displayModel = CONTROLLER_PROPERTIES['oculus-touch-v2'];
         }
-      } else { // WebXR
+      } else {
+        // WebXR
```
dmarcos, line 237: this should move to the line above as it was before

src/components/oculus-touch-controls.js
```diff
@@ -320,24 +360,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
+    x: 'z'
+  },
+
   updateModel: function (buttonName, evtName) {
     if (!this.data.model) { return; }
     this.updateButtonModel(buttonName, evtName);
   },
 
-  updateButtonModel: function (buttonName, state) {
+  updateButtonModel: function (buttonName, evtName) {
```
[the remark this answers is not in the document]
kylebakerio: Well, this was a conscious change. If you'll humor me, this feels like a mistake.

### tldr:
- updateModel calls this argument evtName
- this is indeed only a name, and not state
- onButtonChanged follows a different flow and actually _does_ use `evt.detail.state`, which is unavailable here
- calling this state incorrectly feels distinctly like an oversight

### Detailed argument:
- oculus-touch-controls.js adds onButtonDown (etc.) event listeners, have them point to utils/tracked-controls.js/onButtonEvent()
- tracked-controls.js emits 'buttondown'
- oculus-touch-controls.js picks that up, and strips away the event and only passes the detail.id and the word 'down'; state is ignored
- utils/tracked-controls.js/onButtonEvent() calls oculus-touch-controls.js/updateModel() directly
- updateModel() calls the argument it receives `evtName` (correctly, as it does not include state)
- but then, updateModel() calls updateButtonModel(), which now calls that `state` instead of `evtName`

This is particularly confusing, because evt.detail.state actually is a thing that exists, but it isn't available here.

There is a different flow where state is made accessible:
- oculus-touch-controls.js adds onButtonChanged event listener
- this is different than the onButton Down / Up / TouchStart / TouchEnd events; it doesn't use that utils function, but instead passes the raw event, _which includes event.detail.state_, to its own function (_not_ the utils onButtonEvent() function)
- evt.detail.state.value is then used to precisely adjust the mesh movement

Based on all that, it seemed like an oversight that this argument was called state instead of evtName, being that both are clearly defined concepts in this file and outside of this file.

src/components/oculus-touch-controls.js
kylebakerio: based on the conversation above, I agree. Will change.

src/components/oculus-touch-controls.js
kylebakerio: 1. This does not only work for WebVR, we do it for both WebVR and WebXR. This comment was out of date.
2. This comment _is_ still in the code: `     // If the developer is asking for auto-detection, use the retrieved displayName to identify the specific unit.`, line 227

src/components/oculus-touch-controls.js
kylebakerio: I found having one to the side, and one above, to make the code consuing to read. Given that they are symmetrical comments describing symmetrical code blocks, this was a way to make them consistent.

The other way to make them consistent would be to have them both to the side

```js
if { // webvr
// ...
} else { // webxr
// ...
```

but that feels messier. It also feels like a comment outside of the entire `if` block describes the entire `if/else` chain, where as a comment _inside_ the `if` block describes what is inside that block (which is more accurate).

As someone who was working on this code and having to grok those comments, are you willing to reconsider this reposition? I personally really feel it is an improvement. If you insist, though, I can put them back I suppose.

src/components/oculus-touch-controls.js
```diff
@@ -320,24 +360,128 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
...
     if (buttonMeshes && buttonMeshes[buttonName]) {
       button = buttonMeshes[buttonName];
       button.material.color.set(color);
       this.rendererSystem.applyColorCorrection(button.material.color);
     }
   }
 });
+
+/**
+ * A-Frame's material component makes a single unified shared material for all child meshes.
```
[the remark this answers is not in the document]
kylebakerio: agreed, will change.

src/components/oculus-touch-controls.js
kylebakerio: (see below)

src/components/oculus-touch-controls.js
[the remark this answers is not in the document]
dmarcos: It's better not to change anything unrelated to the PR. It makes reviews more complicated

dmarcos:
This is really close. Just needs to change back some lines / code unrelated to the PR. Thanks for the patience and the effort. So glad we have this working again! Awesome job to everyone that participated

kylebakerio:
@dmarcos do you have any insight into why that build would fail? When I run that test locally it passes. 

`npm test -- --browsers Firefox` looks like what failed, running that I just get 352 tests completed with a checkmark:

image

This is the error in the test:

```sh
13 09 2022 21:42:49.196:ERROR [framework.browserify]: bundle error
13 09 2022 21:42:49.197:ERROR [framework.browserify]: Error: Can't walk dependency graph: Cannot find module 'super-three/examples/js/loaders/KTX2Loader' from '/home/runner/work/aframe/aframe/src/lib/three.js'
    required by /home/runner/work/aframe/aframe/src/lib/three.js
```

Is that related to some change on master, perhaps...?

kylebakerio:
Looks like https://github.com/supermedium/three.js/tree/super-r144/examples/js/loaders i smissing KTX2loader.js, while my npm install was on 141 before the bump we had a couple days ago to 144.

kylebakerio:
Here's the change: https://github.com/supermedium/three.js/commit/d5b300eae4ef13e6e3edd5d1042273a812e401ac

vincentfretin:
@kylebakerio https://github.com/aframevr/aframe/pull/5101#issuecomment-1246492779

kylebakerio:
Just followed vincent's comment all the way up the chain. I look forward to us switching to a more modern build tool in the form of webpack for the future health of the project. I think this is still safe to merge as-is, since it passes locally and was passing before the last non-substantial changes, but if it needs to wait until the upgrade to webpack and it starts passing again, so be it.

Thanks (again) for doing that work Vincent.

dmarcos:
I think we can merge this after https://github.com/aframevr/aframe/pull/5103#issuecomment-1246878413. Thanks

dmarcos:
I addressed a couple of nits. Thanks so much for the patience and being accommodating. This is great work. So glad we have this working again. This kind of details matter a lot to make a great experience. Great job! 🥳 🏅

PR #5102, opened 2022-08-26
correcting master branch for IOS VR

shabarito:
**Description:**
VR is not working on IOS. The changes below seems to fix the issues.

**Changes proposed:**
- Creating a numerable object with EventDispatcher class to allow correct assignment with WebVRManager
- Adding back camera to xr.getCamera() to fix WebVRManager.

PR #5110, opened 2022-09-07
Set WebXR's fixedFoveation property based on foveationLevel (fixes #5108)

### LINKED ISSUE #5108: Foveation can't be configured for WebXR
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: Any platform/device with support for WebXR and `fixedFoveation`

The documentation mentions the `foveationLevel` property on the `render` component. However, this property is only used in case of WebVR ([a-scene.js#L334](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L334)). WebXR now also supports setting a foveation level, and Three.js exposes this since r131 (https://github.com/mrdoob/three.js/pull/22162). By default it is set to 1 in three (the maximum). This means that:

- Setting `<a-scene renderer="foveationLevel: 0">` currently has no effect when using WebXR
- Since A-Frame version 1.3.0 foveation is on by default, whereas it used to be off by default (and can't be disabled)
- The values for WebVR are `0, 1, 2, 3` but for WebXR in the interval `[0.0, 1.0]`

I can make a PR so that the foveation can be configured for WebXR, but would like to discuss the desired approach first. None of the options seem without their drawbacks as the value ranges are inherently incompatible and adding a new property might lead to confusion.

mrxz:
**Description:**
These changes make `foveationLevel` property have an effect on WebXR by setting the `foveation` property through Three.js. In contrast to the `foveationLevel` of WebVR, this can be changed on the fly. 

One thing I noticed is that the async `WebXRManager.setSession` is not awaited. Setting the `foveation` becomes a no-op unless the base layer is initialized, which happens during `setSession`. I now simply chain the setting of the `foveation` to it, but it feels like the promise returned by `enterVR` should not resolve before setSession is done in the first place. But I might be missing something and didn't want to change more than is needed.

**Changes proposed:**
- Let `foveationLevel` set the `foveation` of WebXR (and no longer for WebVR)
- Changes default value for `foveationLevel` to `1` (which it effectively already was in case of WebXR)

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -293,6 +293,7 @@ module.exports.AScene = registerElement('a-scene', {
 
         // Has VR.
         if (this.checkHeadsetConnected() || this.isMobile) {
+          var rendererSystem = self.getAttribute('renderer');
           vrManager.enabled = true;
 
           if (this.hasWebXR) {
@@ -309,7 +310,9 @@ module.exports.AScene = registerElement('a-scene', {
                 function requestSuccess (xrSession) {
                   self.xrSession = xrSession;
                   vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1;
-                  vrManager.setSession(xrSession);
+                  vrManager.setSession(xrSession).then(function () {
+                    vrManager.setFoveation(rendererSystem.foveationLevel);
+                  });
                   xrSession.addEventListener('end', self.exitVRBound);
                   enterVRSuccess(resolve);
                 },
@@ -328,10 +331,8 @@ module.exports.AScene = registerElement('a-scene', {
               enterVRSuccess();
               return Promise.resolve();
             }
-            var rendererSystem = this.getAttribute('renderer');
             var presentationAttributes = {
-              highRefreshRate: rendererSystem.highRefreshRate,
-              foveationLevel: rendererSystem.foveationLevel
+              highRefreshRate: rendererSystem.highRefreshRate
             };
 
             return vrDisplay.requestPresent([{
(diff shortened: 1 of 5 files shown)


dmarcos:
Thank you! This is awesome. Congrats on your first contribution 👏

src/core/scene/a-scene.js
```diff
@@ -309,7 +310,9 @@ module.exports.AScene = registerElement('a-scene', {
                 function requestSuccess (xrSession) {
                   self.xrSession = xrSession;
                   vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1;
-                  vrManager.setSession(xrSession);
+                  vrManager.setSession(xrSession).then(function () {
+                    vrManager.setFoveation(rendererSystem.foveationLevel);
```
vincentfretin, line 314: This broke Firefox, we need to check if vrManager.setFoveation is defined before calling it. On Firefox vrManager is WebVRManager that we still maintain in the super-three fork, not WebXRManager.

vincentfretin: Instead of adding a condition in aframe code, we could maybe add a dummy `setFoveation` function in `WebVRManager `to fix that @dmarcos?

mrxz: I don't think it's this invocation that causes issues, since it's inside the `this.hasWebXR` clause. It's probably the one added in `renderer.js#update`, as that one is called unconditionally. Good catch, in Firefox I always have `dom.vr.webxr.enabled` set to true, so it didn't fail for me during testing.

In any case, I also think a (dummy) `setFoveation` in `WebVRManager` is the cleanest solution. The less difference there is between the `WebVRManager` and `WebXRManager` in terms of interface, the less likely bugs like these are.

vincentfretin: You're right. We need to do some other fixes in WebVRManager, see https://github.com/aframevr/aframe/pull/5102#issuecomment-1228735590 so I guess it's one more. :)

Issue #3146, opened 2017-10-10
memory not released after scene is removed


ebhoren:
Memory is not released after a scene is removed from DOM.

- A-Frame Version: 0.7.0
- Platform / Device: iPad/iPhone
- Reproducible Code Snippet or URL: https://codepen.io/ebhoren/pen/zEjYRO

Make a memory snapshot before adding a scene. 
Another memory snapshot after adding a scene.
Remove the scene and take another memory snapshot.

You will see the memory is not released and just keep increasing each time you remove a scene and add a new one.

Thanks

dmarcos:
What is the use case for adding / removing scenes? It will be probably better to empty a scene and repopulate.

ebhoren:
I'm adding/removing scenes because i splitted my app logic in 3 big scenes. 
Each scenes load around 30-40mb on textures and videos.

How can i achieve this by emptying the scene, and his assets, and repopulate it with new elements+assets in the assets manager?

ngokevin:
Asset manager is just for blocking the scene on first load, it does not have any benefit after that. You can use inline URLs and have the same effect. But emptying/modifying the scene is the way to go, and we won't much support this use cases of attaching/reattaching scene. If you have any specific places where A-Frame can free memory, feel free to file an issue for that. Thanks!

donmccurdy:
one quick comment i'll pull from #3137 - we should probably call WebGLRenderer.dispose() when a scene detaches

PR #5112, opened 2022-09-10
Properly dispose renderer to stop the animation loop

vincentfretin:
**Description:**
In aframe 1.3.0 and master (three r141), when you are removing the a-scene from the DOM, it disposes the WebXRManager/WebVRManager (`renderer.xr`) on `detachedCallback` but not the renderer itself (WebGLRenderer), so it doesn't stop properly the animation loop and it continues executing some tick methods.
You can test with `document.body.removeChild(AFRAME.scenes[0])`
Before the PR:
Capture d’écran du 2022-09-10 17-42-52

After the PR:
Capture d’écran du 2022-09-10 18-33-41

This is known issue since a long time, see https://github.com/aframevr/aframe/issues/3146#issuecomment-335682180 and https://stackoverflow.com/questions/54930875/aframejs-how-to-completely-destroy-a-scene and I also documented it in https://aframe.wiki/en/memory

**Changes proposed:**
- This PR properly call `renderer.dispose()` that execute `renderer.xr.dispose()` and `animation.stop()` and also dispose other things, see https://github.com/supermedium/three.js/blob/super-r141/src/renderers/WebGLRenderer.js#L592-L621

Properly disposing the renderer in `detachedCallback` seems right because in `attachedCallback` that is calling `this.setupRenderer()` we're creating a new WebGLRenderer always.
But note that [detaching or reattaching the scene is not a supported use case](https://github.com/aframevr/aframe/issues/3146#issuecomment-335660380), it gives currently an error before and after this PR anyway on the `/examples/boilerplate/3d-model/` example where I tested that and I'm no gonna investigate that.

```
scene=AFRAME.scenes[0]
document.body.removeChild(scene)
document.body.appendChild(scene)
core:a-entity:warn Tried to remove `Object3D` of type:  mesh which was not defined.
2
browser.js:111 core:a-node:error Failure loading node:   TypeError: Cannot read properties of undefined (reading 'constructor')
    at extendProperties (component.js:736:1)
    at NewComponent.callUpdateHandler (component.js:416:1)
    at NewComponent.updateProperties (component.js:300:1)
    at HTMLElement.value (a-entity.js:490:1)
    at HTMLElement.value (a-entity.js:456:1)
    at entityLoadCallback (a-entity.js:249:1)
    at emitLoaded (a-node.js:127:1)
browser.js:111 components:gltf-model:warn Unexpected token '<', "<!DOCTYPE "... is not valid JSON 
```

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -214,7 +214,7 @@ module.exports.AScene = registerElement('a-scene', {
         window.removeEventListener('vrdisplaypointerrestricted', this.pointerRestrictedBound);
         window.removeEventListener('vrdisplaypointerunrestricted', this.pointerUnrestrictedBound);
         window.removeEventListener('sessionend', this.resize);
-        this.renderer.xr.dispose();
+        this.renderer.dispose();
       }
     },
 
--- a/tests/__init.test.js
+++ b/tests/__init.test.js
@@ -41,6 +41,7 @@ setup(function () {
       dispose: function () {},
       enabled: false
     },
+    dispose: function () {},
     getContext: function () { return undefined; },
     setAnimationLoop: function () {},
     setSize: function () {},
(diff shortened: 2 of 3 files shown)


dmarcos:
Thanks!

PR #5115, opened 2022-09-14
Update writing-a-component.md

janarthnk:
I found the code in the 'Handling Property Updates' to be a little unclear. `addEventListener` would be called even if another property (eg `message`) was updated. As it is the code works fine, since `addEventListener`s arguments are unchanged. Calling `addEventListener` again with unchanged parameters has no effect.

I thought it would be a bit clearer if the changes to the event property were handled separately.

**Description:**

**Changes proposed:**
-
-
-


dmarcos:
Thanks! Congrats on your first contribution 🥳

PR #5116, opened 2022-09-14
Switch to webpack to support ESM

### LINKED ISSUE #5084: Self Signed SSL error
On Windows 10, after installing the latest OpenSSL v3, `npm run start:https` resulted in this error:

```
node:events:505
      throw er; // Unhandled 'error' event
      ^

Error: RSA PRIVATE KEY not found from openssl output:
---stdout---
-----BEGIN PRIVATE KEY-----
...
```

Basically the output text from OpenSSL seems to have changed, hence the self signed cert cannot be parsed.  A quick hackish fix for this is a one line change to `node_modules/pem/lib/pem.js`, line 77:

FROM:
   `openssl.exec(params, 'RSA PRIVATE KEY', function (sslErr, key) {`

TO:   
   `openssl.exec(params, 'PRIVATE KEY', function (sslErr, key) {`
	 	 
After that, it starts right up and is useable on my VR Headset now that it has SSL in place for WebXR. 
 Hopefully this saves someone else some time too!  :)

- v1.3
- Windows 10
- OpenSSL v3.05 (Lite) installed from https://slproweb.com/products/Win32OpenSSL.html

vincentfretin:
**Description:**

Browserify doesn't support ESM (ES6 modules), the PR #5101 broke master after it has been merged because of three r144 that removed the  `examples/js/KTX2Loader.js` file as I said in my [comment](https://github.com/aframevr/aframe/pull/5101#issuecomment-1246492779)
and [three.js will remove all the umd builds from their repo by the end of the year](https://github.com/aframevr/aframe/pull/5101#issuecomment-1246799159)
I don't think you want to maintain umd builds yourself in the super-three repo @dmarcos so I think it's time to move away from browserify/budo. In this PR I propose to switch to webpack.

**Changes proposed:**

I kept the changes to the minimum, so I kept polyfilling `Buffer` and `process.browser` and for dev server I kept the same behavior for the host, port, watching files.

For the `Buffer` polyfill, it's needed by the `buffer-equal` package, dependency of `node_modules/load-bmfont/package.json`
and `Buffer` is used in `node_modules/load-bmfont/lib/is-binary.js`
`load-bmfont` is used in `src/components/text.js`

Webpack is using ESM first if found, then fallback to UMD. For the super-animejs package that supports both, this means we need to replace `require('super-animejs')` by `require('super-animejs').default`. This is how imported ESM build behave when importing with `require()`.

For `lib/three.js`, to not change to `require('lib/three.js').default` everywhere I did the following:
I created a `lib/three.module.js` file where I now import the jsm versions instead of js versions of all the loaders and in `lib/three.js` I use `require('./three.module.js').default;` so that's the only place where we use `.default` like that.

Important part, import from "three" that are included in jsm files are aliased to super-three.

Although we use `babel-loader`, we don't do any transpiling, I didn't add the `@babel/preset-env` package.

I tested:
- `npm run start`
- `npm run start:https` generates a certificate on Ubuntu 22.04 successfully, this closes #5084
- editing a file in src or example will reload the page
- `TEST_ENV=ci npm run test:chrome` and open the coverage tests/coverage/report/index.html
- `npm run build` create the non minified and minified build (1.3MB, webpack is using terser with -c -m options I guess) with sourcemaps working


vincentfretin:
- style-loader is the equivalent of browserify-css
- webpack with mode=production does the minification with terser by default. Terser is an indirect dependency of webpack, so I removed the direct dependency
- sourcemaps are in a separate file automatically.

tests/karma.conf.js
```diff
@@ -14,15 +16,20 @@ if (process.env.TEST_FILE) {
     }
   });
 } else {
-  FILES.push('tests/**/*.test.js');
+  glob.sync('tests/**/*.test.js').forEach(function (filename) {
+    FILES.push(filename);
+  });
```
vincentfretin, line 21: I didn't quite understand why the previous code didn't work anymore, it gave me 16 tests instead of 355 tests. Doing the glob myself like it's done when you use `TEST_FILE=atestfile npm run test:chrome` (the if above) fixed it.

tests/karma.conf.js
```diff
@@ -69,7 +67,6 @@ if (process.env.TEST_ENV === 'ci') {
     ]
   };
   karmaConf.reporters.push('coverage');
-  karmaConf.preprocessors['src/**/*.js'] = ['coverage'];
```
vincentfretin, line 72: Removing this line is on purpose, this is not an error, https://github.com/istanbuljs/babel-plugin-istanbul#karma is saying to not add the coverage preprocessor.
I compared the report coverage before and after this PR, this is working properly.

webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+    static: {
+      directory: 'examples'
+    }
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.INSPECTOR_VERSION': JSON.stringify(
+        process.env.INSPECTOR_VERSION
+      )
+    }),
```
vincentfretin, line 29: If you want to know more about DefinePlugin: https://webpack.js.org/plugins/define-plugin/

webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+  devtool: 'source-map',
+  mode: 'development',
+  devServer: {
+    host: 'local-ip',
+    port: process.env.PORT || 9000,
+    hot: false,
+    liveReload: true,
+    static: {
+      directory: 'examples'
+    }
```
vincentfretin, line 22: For you to understand, before you had a specific build for budo and replaced /dist by /build in the script. In webpack-dev-server, the build is done in memory when you are accessing `/dist/aframe-master.js`, all the other files are served from the examples directory.

webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.INSPECTOR_VERSION': JSON.stringify(
+        process.env.INSPECTOR_VERSION
+      )
+    }),
+    new webpack.ProvidePlugin({
+      process: 'process/browser',
+      Buffer: ['buffer', 'Buffer']
+    })
```
vincentfretin, line 33: https://webpack.js.org/plugins/provide-plugin/

I found the solution for `process.browser` and `Buffer` here https://www.alchemy.com/blog/how-to-polyfill-node-core-modules-in-webpack-5
Note that webpack gave me this information at the beginning:
```
ERROR in ./node_modules/buffer-equal/index.js 1:13-37
Module not found: Error: Can't resolve 'buffer' in '/home/vincentfretin/workspace/aframe/node_modules/buffer-equal'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "buffer": require.resolve("buffer/") }'
	- install 'buffer'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "buffer": false }
```

but using `resolve.fallback` didn't work because `buffer-equal` doesn't even use `require('buffer')`, it just uses `Buffer` directly.

webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+  ],
+  resolve: {
+    alias: {
+      three: 'super-three'
```
vincentfretin, line 37: This one is super important otherwise you get three not found when importing the jsm files.

webpack.prod.config.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = merge(commonConfiguration, {
+  output: {
+    library: 'AFRAME',
+    libraryTarget: 'umd',
+    path: path.resolve(__dirname, 'dist'),
+    publicPath: '/dist/',
+    filename: 'aframe-master.min.js'
+  },
+  mode: 'production'
+});
```
vincentfretin, line 14: This file exists just to change the filename and the mode to production so that we have a minified build.
https://webpack.js.org/configuration/mode/

webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+        use: {
+          loader: 'babel-loader'
+          // options: {
+          //   presets: ['@babel/preset-env'],
+          // },
+        }
+      },
+      {
+        test: /\.css$/,
+        use: ['style-loader', 'css-loader']
```
vincentfretin, line 53: style-loader will inject two style tags in the DOM corresponding to the two css files we import, similar to browserify-css

tests/__init.test.js
```diff
@@ -19,13 +19,14 @@ navigator.getVRDisplays = function () {
...
 
-require('index');
+const AFRAME = require('index');
 var AScene = require('core/scene/a-scene').AScene;
 // Make sure WebGL context is not created since Travix CT runs headless.
 // Stubs below failed once in a while due to asynchronous tesst setup / teardown.
 AScene.prototype.setupRenderer = function () {};
 
 setup(function () {
+  window.AFRAME = AFRAME;
```
vincentfretin, line 29: This one gave me a hard time. I'm not sure how this worked with browserify with karma before, I think karma-browserify injected the `AFRAME` variable in each file but I'm not really sure. With webpack with karma `AFRAME` was an empty object in some tests, defining as a global variable in `setup` fixed it.

package.json
```diff
@@ -5,12 +5,10 @@
   "homepage": "https://aframe.io/",
   "main": "dist/aframe-master.js",
   "scripts": {
-    "browserify": "browserify src/index.js -s 'AFRAME' -p browserify-derequire",
-    "build": "shx mkdir -p build/ && npm run browserify -- --debug -t [ envify --INSPECTOR_VERSION dev ] -o build/aframe.js",
-    "dev": "npm run build && cross-env INSPECTOR_VERSION=dev node ./scripts/budo -t envify",
-    "dist": "node scripts/updateVersionLog.js && npm run dist:min",
-    "dist:max": "npm run browserify -s -- --debug | exorcist dist/aframe-master.js.map > dist/aframe-master.js",
-    "dist:min": "npm run dist:max && terser dist/aframe-master.js -c --source-map 'content=dist/aframe-master.js.map,url=aframe-master.min.js.map' -o dist/aframe-master.min.js",
+    "dev": "cross-env INSPECTOR_VERSION=dev webpack serve",
```
vincentfretin, line 8: The webpack command comes from the webpack-cli package. The `serve` argument is available when you have the webpack-dev-server package. `webpack serve` use the `devServer` configuration in `webpack.config.js`

package.json
```diff
@@ -20,7 +18,7 @@
     "prepush": "node scripts/testOnlyCheck.js",
     "prerelease": "node scripts/release.js 1.2.0 1.3.0",
     "start": "npm run dev",
-    "start:https": "cross-env SSL=true npm run dev",
+    "start:https": "npm run dev -- --server-type https",
```
vincentfretin, line 21: I chose to use the parameter here but we could have used an environment variable and set it in the config https://webpack.js.org/configuration/dev-server/#devserverserver

webpack.config.js
vincentfretin: https://webpack.js.org/configuration/resolve/

src/index.js
```diff
@@ -99,7 +99,7 @@ module.exports = window.AFRAME = {
   AComponent: require('./core/component').Component,
   AEntity: AEntity,
   ANode: ANode,
-  ANIME: require('super-animejs'),
+  ANIME: require('super-animejs').default,
```
vincentfretin, line 102: For you to understand webpack by default is importing the package from the field 'browser', 'module', 'main' in this order of priority by default if those fields are defined in `package.json`. super-animejs has a module field defined pointing to an ESM build.
https://webpack.js.org/configuration/resolve/#resolvemainfields

dmarcos:
Great! Much appreciated. Thanks for making a surgical as possible.

tests/__init.test.js
```diff
@@ -19,13 +19,14 @@ navigator.getVRDisplays = function () {
   return Promise.resolve([mockVRDisplay]);
 };
 
-require('index');
+const AFRAME = require('index');
```
vincentfretin, line 22: Sorry for the `const` here :) you can modify it to `var` directly on master if you want.

vincentfretin:
I didn't expect you to merge it so quickly, I expected at least one question lol. I guess you didn't have any because I added a lot of comments to explain the changes. ;-)

arpu:
this is fantastic!

vincentfretin:
@dmarcos you will see that webpack generates `dist/aframe-master.js.LICENSE.txt` and `dist/aframe-master.min.js.LICENSE.txt`, you can add them to the repository.
Can you please create a build manually and push it, or make the bot working again? Thanks.

vincentfretin:
Note that in the dist there is now usage of `class`, but really nowadays it's safe to use https://caniuse.com/?search=class
Also three is targeting browsers >1%, see [their babel config ](https://github.com/mrdoob/three.js/blob/916ed1e848a47d4fc7065ff23333b21f925621bb/utils/build/.babelrc.json) that they use with [rollup](https://github.com/mrdoob/three.js/blob/bd61eb4469f1244fb3c4f7eac75b30c192be3f15/utils/build/rollup.config.js) to make their build.

And super-animejs is also using [rollup](https://github.com/juliangarnier/anime/blob/master/build.js) to convert the code to ES6 (ES6 and ES2015 is the same thing) via [@rollup/plugin-buble](https://www.npmjs.com/package/@rollup/plugin-buble).
Aframe is using mainly ES5, ES6 for-of and now ES6 import/export, we are sure we only use supported [ES6 syntax](https://caniuse.com/es6) in our build, so we don't need to bother to use `@babel/preset-env` ourself unless you decide to use [class properties](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) or some other ES proposal in the code base.

vincentfretin:
Actually there was already usage of `class` in the previous build with the imported loaders, so yeah it doesn't change anything.

Issue #5105, opened 2022-08-30
Running a-frame without internet access seems impossible


gertjanal:
**Description:**
For a stand-alone setup, I used a laptop hosting an a-frame webpage and an Oculus Quest.
The loading of the page takes ages when no (real) internet is available.
It occurred to me that for a-text elements, fonts are being downloaded from the web. Also the oculus controller models don't load when no internet is available.

I think a better way of hosting these can be found, or a way to run a-frame without internet should be documented

- A-Frame Version: Not sure, the version that was available in July 2022
- Platform / Device: Oculus Quest (with Firefox Reality Browser or Wolvic Reality Browser)
- Reproducible Code Snippet or URL:
`https://cdn.aframe.io/fonts/Roboto-msdf.json`

dmarcos:
For text you could point the [text component](https://aframe.io/docs/1.3.0/components/text.html#properties_font) to an URL you host locally. 

For the controllers the URL of the built-in models is not currently exposed in the [component schema](https://aframe.io/docs/1.3.0/components/oculus-touch-controls.html#value). It's a bit complicated because we support Rift / Oculus 1 and Oculus 2 models. 

If you're offline I'm assuming you are not pulling A-Frame from the CDN but serving your own custom builds. You can modify the base URL from [oculus-touch-controls](https://github.com/aframevr/aframe/blob/master/src/components/oculus-touch-controls.js#L19) and serve the models locally.

Utopiah:
FWIW as I need an offline session few weeks ago I updated the URL from the build. Not clean but worked for both text and controllers.

Utopiah:
IMHO documenting this should be sufficient.

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5118

PR #5118, opened 2022-09-19
Documenting offline use

Utopiah:
Pointing out FONT_BASE_URL and MODEL_URLS and the process.

**Description:** clarifying what is needed to work offline

**Changes proposed:**
- font and model hardcoded URLs, discussed in https://github.com/aframevr/aframe/issues/5105


dmarcos:
Thank you!

PR #5125, opened 2022-09-24
Fix reflection component removal

vincentfretin:
**Description:**

We get a `Uncaught TypeError: this.removeChild is not a function` error when removing the scene if we had the reflection component on it and we didn't enter vr mode.

**Changes proposed:**
- Be sure `this.probeLight` exists before trying to remove it, also use `this.el.removeChild` instead of non existing `this.removeChild`
- Add a test that doesn't fail on this error

### DIFF
--- a/src/components/scene/reflection.js
+++ b/src/components/scene/reflection.js
@@ -32,6 +32,7 @@ module.exports.Component = register('reflection', {
     this.cubeCamera = new THREE.CubeCamera(0.1, 1000, this.cubeRenderTarget);
     this.lightingEstimationTexture = (new THREE.WebGLCubeRenderTarget(16)).texture;
     this.needsVREnvironmentUpdate = true;
+    this.probeLight = null;
 
     // Update WebXR to support light-estimation
     var webxrData = this.el.getAttribute('webxr');
@@ -142,7 +143,7 @@ module.exports.Component = register('reflection', {
       this.needsVREnvironmentUpdate = false;
       this.cubeCamera.position.set(0, 1.6, 0);
       this.cubeCamera.update(renderer, scene);
-      this.el.object3D.environment = this.cubeRenderTarget.texture;
+      scene.environment = this.cubeRenderTarget.texture;
     }
 
     if (this.needsLightProbeUpdate && frame) {
@@ -155,6 +156,9 @@ module.exports.Component = register('reflection', {
 
   remove: function () {
     this.el.object3D.environment = null;
-    this.removeChild(this.probeLight);
+    if (this.probeLight) {
+      this.el.removeChild(this.probeLight);
+      this.probeLight = null;
+    }
   }
 });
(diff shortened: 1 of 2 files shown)


src/components/scene/reflection.js
```diff
@@ -32,6 +32,7 @@ module.exports.Component = register('reflection', {
     this.cubeCamera = new THREE.CubeCamera(0.1, 1000, this.cubeRenderTarget);
     this.lightingEstimationTexture = (new THREE.WebGLCubeRenderTarget(16)).texture;
     this.needsVREnvironmentUpdate = true;
+    this.probeLight = null;
```
vincentfretin, line 35: I prefer to declare the variable in init like this. Also I check for it in the test.

src/components/scene/reflection.js
```diff
@@ -142,7 +143,7 @@ module.exports.Component = register('reflection', {
       this.needsVREnvironmentUpdate = false;
       this.cubeCamera.position.set(0, 1.6, 0);
       this.cubeCamera.update(renderer, scene);
-      this.el.object3D.environment = this.cubeRenderTarget.texture;
+      scene.environment = this.cubeRenderTarget.texture;
```
vincentfretin, line 146: We use `scene.environment = null` three line above, we should use the scene variable here for symmetry :)

src/components/scene/reflection.js
```diff
@@ -155,6 +156,9 @@ module.exports.Component = register('reflection', {
 
   remove: function () {
     this.el.object3D.environment = null;
-    this.removeChild(this.probeLight);
+    if (this.probeLight) {
+      this.el.removeChild(this.probeLight);
+      this.probeLight = null;
```
vincentfretin, line 161: `this.probeLight = null;` is not mandatory, but I find it good code hygiene to remove any reference to a DOM element explicitly. This can help the GC too.

dmarcos: This should probably be undefined. null usually represent a missing object. In this case we're resetting the variable. Also for style consistency

src/components/scene/reflection.js
dmarcos: Do we need to declare it?

src/components/scene/reflection.js
vincentfretin: I removed the line. This doesn't add much. The component is removed anyway, so the DOM element is removed as well.

dmarcos:
Thank you!

src/components/scene/reflection.js
vincentfretin: About null vs undefined, please note that in the aframe code base we are currently setting some variable to null in remove like I suggested. In the text component for example:
https://github.com/aframevr/aframe/blob/579473b83772b67c76e1e75513b754defef5c32c/src/components/text.js#L126-L135

PR #5128, opened 2022-10-03
Remove Google Blocks (poly) from FAQ

daniel-schroeder-dev:
**Description:**

Google Poly was shut down on June 30, 2021, so it is no longer a viable option for getting 3D models. See https://support.google.com/poly/answer/10192635 for details.

**Changes proposed:**
- Remove the link from the FAQ


dmarcos:
Thanks!

PR #5139, opened 2022-10-19
Add suppport for Meta Quest Touch Pro controllers (fix #5138)

### LINKED ISSUE #5138: Add suport for touch controller pro
Meta shipped a new device with new controllers.
Aframe is hardcoding controller names and doesn't have the name for this controller. This is causing experiences to render without controllers.

Aframe should add support for this controller and it should also use the fallback controller if it can't find the top one.

cc @dmarcos

dmarcos:
I noticed there's not a button mapping profile on the [webxr repository](https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/registry/profiles/oculus). I'm assuming the mapping is the same as the previous touch controllers.

To test you can use the example below:

`npm start` and load `http://localhost:9000/showcase/ui/`

I haven't look at the model but if parts are separated we could also animate / highlight the buttons, trigger when pressed in a separate PR.

@cabanier Let me know if it works for you. Thanks so much for testing and congrats on the Quest Pro release. It looks awesome.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -218,7 +218,8 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
   checkIfControllerPresent: function () {
     checkControllerPresentAndSetup(this, GAMEPAD_ID_PREFIX, {
-      hand: this.data.hand
+      hand: this.data.hand,
+      iterateControllerProfiles: true
     });
   },
 
@@ -252,10 +253,9 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       } else { // WebXR
         controllerId = CONTROLLER_DEFAULT;
         var controllersPropertiesIds = Object.keys(CONTROLLER_PROPERTIES);
-        for (var i = 0; i < controllersPropertiesIds.length; i++) {
-          var controllerPropertyId = controllersPropertiesIds[i];
-          if (controller.profiles.indexOf(controllerPropertyId)) {
-            controllerId = controllerPropertyId;
+        for (var i = 0; i < controller.profiles.length; i++) {
+          if (controllersPropertiesIds.indexOf(controller.profiles[i]) !== -1) {
+            controllerId = controller.profiles[i];
             break;
           }
         }
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -276,7 +276,8 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
       id: id,
       hand: data.hand,
       orientationOffset: data.orientationOffset,
-      handTrackingEnabled: false
+      handTrackingEnabled: false,
+      iterateControllerProfiles: true
     });
     this.loadModel(controller);
   },
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -17,6 +17,7 @@ var GAMEPAD_ID_PREFIX = isWebXRAvailable ? GAMEPAD_ID_WEBXR : GAMEPAD_ID_WEBVR;
 
 // First generation model URL.
 var TOUCH_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/oculus/oculus-touch-controller-';
+var META_CONTROLLER_MODEL_BASE_URL = 'https://cdn.aframe.io/controllers/meta/';
 
 var OCULUS_TOUCH_WEBVR = {
   left: {
@@ -79,6 +80,20 @@ var CONTROLLER_PROPERTIES = {
       rayOrigin: {origin: {x: -0.015, y: 0.005, z: 0}, direction: {x: 0, y: 0, z: -1}},
       modelPivotOffset: new THREE.Vector3(-0.01, -0.01, 0.05),
       modelPivotRotation: new THREE.Euler(Math.PI / 4, 0, 0)
+    }
+  },
+  'meta-quest-touch-pro': {
+    left: {
+      modelUrl: META_CONTROLLER_MODEL_BASE_URL + 'quest-touch-pro-left.glb',
+      rayOrigin: {origin: {x: 0.015, y: 0.005, z: 0}, direction: {x: 0, y: 0, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0, 0, 0),
+      modelPivotRotation: new THREE.Euler(0, 0, 0)
+    },
+    right: {
+      modelUrl: META_CONTROLLER_MODEL_BASE_URL + 'quest-touch-pro-right.glb',
+      rayOrigin: {origin: {x: -0.015, y: 0.005, z: 0}, direction: {x: 0, y: 0, z: -1}},
+      modelPivotOffset: new THREE.Vector3(0, 0, 0),
+      modelPivotRotation: new THREE.Euler(0, 0, 0)
     }
   }
 };
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -251,8 +251,14 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
         }
       } else { // WebXR
         controllerId = CONTROLLER_DEFAULT;
-        controllerId = controller.profiles.indexOf('oculus-touch-v2') !== -1 ? 'oculus-touch-v2' : controllerId;
-        controllerId = controller.profiles.indexOf('oculus-touch-v3') !== -1 ? 'oculus-touch-v3' : controllerId;
+        var controllersPropertiesIds = Object.keys(CONTROLLER_PROPERTIES);
+        for (var i = 0; i < controllersPropertiesIds.length; i++) {
+          var controllerPropertyId = controllersPropertiesIds[i];
+          if (controller.profiles.indexOf(controllerPropertyId)) {
+            controllerId = controllerPropertyId;
+            break;
+          }
+        }
         this.displayModel = CONTROLLER_PROPERTIES[controllerId];
       }
     }


cabanier:
> I noticed there's not a button mapping profile on the [webxr repository](https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/registry/profiles/oculus). I'm assuming the mapping is the same as the previous touch controllers.

The new controller's buttons are a superset of the old touch controller so that should be fine

> I haven't look at the model but if parts are separated we could also animate / highlight the buttons, trigger when pressed in a separate PR.

@felixtrz noticed that aframe has code to animate those. Yes, it would be awesome if we can add that. Will you (or someone you know) have access to a Quest Pro

> @cabanier Let me know if it works for you. Thanks so much for testing and congrats on the Quest Pro release. It looks awesome.

Thanks!

cabanier:
Can we also add logic to use the fallback controller? That way if someone comes out with a new controller, aframe can still use an older controller that is compatible.

cabanier (changes requested):
Can we also add logic to use the fallback controller? That way if someone comes out with a new controller, aframe can still use an older controller that is compatible.

dmarcos:
The [laser-controls](https://aframe.io/docs/1.3.0/components/laser-controls.html#sidebar) component used in the example I pointed above `http://localhost:9000/showcase/ui/` has the fallback to a generic controller.

dmarcos:
> 
> The new controller's buttons are a superset of the old touch controller so that should be fine
>

Are there any new buttons we should map?
 
> @felixtrz noticed that aframe has code to animate those. Yes, it would be awesome if we can add that. Will you (or someone you know) have access to a Quest Pro

Not easily but can ask around. If there's a device loaner program at Meta it will make it a bit easier for this and future work.

dmarcos:
Never mind. I found the button mapping [profile in the meta directory](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/registry/profiles/meta/meta-quest-touch-pro.json). It looks it's the same than other Touch controllers. We should be good

cabanier:
> > The new controller's buttons are a superset of the old touch controller so that should be fine
> 
> Are there any new buttons we should map?

No, the other "buttons" are for drawing your index finger and thumb

> Not easily but can ask around. If there's a device loaner program at Meta it will make it a bit easier for this and future work.

I will ask. :-)

felixtrz:
Tested http://localhost:9000/showcase/ui/ on Quest Pro, the controllers are picked up and working (trigger events are registering, target ray seems correct), but no quest pro controller model is rendered, only a sphere, is that expected? 

https://user-images.githubusercontent.com/11973041/197017044-4fd4b353-2e9f-43c5-8829-3d50cc5018fd.mp4

felixtrz:
@dmarcos I did some digging and found where the problem is. The isControllerPresent() in tracked-controls.js is checking and categorizing the type of controller by idPrefix, which is why meta-quest-touch-pro never gets matched as oculus-touch-controls, and that's why it gets matched into the generic control every time. You might need to make an exception in the isControllerPresent to allow meta-quest-touch to be matched as oculus-touch as well.

dmarcos:
@felixtrz Sorry for the inconvenience. Thanks for testing and patience. 

I generalized the logic that loads the model. I think it should work now. If it doesn't. Can you add a [breakpoint here](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R257) and let me know what profiles you see in `controller.profiles`? You'll need to enter VR so controllers show up.

Srile:
@cabanier and I did some digging and figured out the cause of the bug. 

The issue is caused by the `GAMEPAD_ID_WEBXR` variable for the meta devices changing from `oculus-touch` to `meta-quest-touch`, which fails in the `findMatchingControllerWebXR` function.

dmarcos:
> @cabanier and I did some digging and figured out the cause of the bug.
> 
> The issue is caused by the `GAMEPAD_ID_WEBXR` variable for the meta devices changing from `oculus-touch` to `meta-quest-touch`, which fails in the `findMatchingControllerWebXR` function.

As per [WebXR profiles](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/registry/profiles/meta/meta-quest-touch-pro.json) the profile id should be `meta-quest-touch-pro`. This is what this PR assumes.

felixtrz:
> @felixtrz Sorry for the inconvenience. Thanks for testing and patience.
> 
> I generalized the logic that loads the model. I think it should work now. If it doesn't. Can you add a [breakpoint here](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R257) and let me know what profiles you see in `controller.profiles`? You'll need to enter VR so controllers show up.

The problem is in [checkIfControllerPresent](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R220), this always returns false if we don't make changes to isControllerPresentWebXR() in tracked-controls.js. The breakpoint does not run because of this, but I can give you the profiles captured in tracked-controls.js: ['meta-quest-touch-pro', 'oculus-touch-v2', 'oculus-touch', 'generic-trigger-squeeze']

dmarcos:
> > @felixtrz Sorry for the inconvenience. Thanks for testing and patience.
> > I generalized the logic that loads the model. I think it should work now. If it doesn't. Can you add a [breakpoint here](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R257) and let me know what profiles you see in `controller.profiles`? You'll need to enter VR so controllers show up.
> 
> The problem is in [checkIfControllerPresent](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R220), this always returns false if we don't make changes to isControllerPresentWebXR() in tracked-controls.js. The breakpoint does not run because of this, but I can give you the profiles captured in tracked-controls.js: ['meta-quest-touch-pro', 'oculus-touch-v2', 'oculus-touch', 'generic-trigger-squeeze']

Logic works across touch for rift, Quest 1 and Quest 2. Wonder what's different about Quest Pro. No changes should be needed to checkIfControllerPresent. The profile id is passed to that logic from [here](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R271). Can you add a breakpoint there and let me know what `id` is passed to the component? Thanks for the help and the patience.

Srile:
The `id` passed is always `oculus-touch`, which does not match with the prefix for the meta quest pro, as the prefix was changed due to the rebranding.

felixtrz:
> > > @felixtrz Sorry for the inconvenience. Thanks for testing and patience.
> > > I generalized the logic that loads the model. I think it should work now. If it doesn't. Can you add a [breakpoint here](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R257) and let me know what profiles you see in `controller.profiles`? You'll need to enter VR so controllers show up.
> > 
> > 
> > The problem is in [checkIfControllerPresent](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R220), this always returns false if we don't make changes to isControllerPresentWebXR() in tracked-controls.js. The breakpoint does not run because of this, but I can give you the profiles captured in tracked-controls.js: ['meta-quest-touch-pro', 'oculus-touch-v2', 'oculus-touch', 'generic-trigger-squeeze']
> 
> Logic works across touch for rift, Quest 1 and Quest 2. Wonder what's different about Quest Pro. No changes should be needed to checkIfControllerPresent. The profile id is passed to that logic from [here in line 275](https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd275). Can you add a breakpoint there and let me know what `id` is passed to the component? Thanks for the help and the patience.

I've put up my fix that have the quest controller show up here: https://github.com/aframevr/aframe/pull/5140

dmarcos:
@felixtrz I made a change to iterate through the profiles instead of stopping at first one:

https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R280

Can you let me know if it now works as expected? Thanks so much

felixtrz:
> @felixtrz I made a change to iterate through the profiles instead of stopping at first one:
> 
> https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R280
> 
> Can you let me know if it now works as expected? Thanks so much

Still does not work, in your fix, checkIfControllerPresent still returns false in oculus-touch-controls

felixtrz:
> @felixtrz I made a change to iterate through the profiles instead of stopping at first one:
> 
> https://github.com/aframevr/aframe/pull/5139/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R280
> 
> Can you let me know if it now works as expected? Thanks so much

Would it be helpful if we jump on a call to debug together live? Happy to do so if you are available

felixtrz:
> available

yes, I will shoot you an email at your gmail address for a quick zoom call

dmarcos:
> > available
> 
> yes, I will shoot you an email at your gmail address for a quick zoom call

Perfect, thanks!

dmarcos:
@felixtrz Thanks so much for jumping and helping debug so quickly. I just incorporated your changes. Let me know if it works now. Fingers crossed!

PR #5140, opened 2022-10-20
Add suppport for Meta Quest Touch Pro controllers (fix #5138)

### LINKED ISSUE #5138: Add suport for touch controller pro

felixtrz:
A patch for https://github.com/aframevr/aframe/pull/5139

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -171,7 +171,11 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index, 
           if (controllerMatch) { break; }
         }
       } else {
-        controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);
+        if (idPrefix === 'oculus-touch') {
+          controllerMatch = profiles.length > 0 && (profiles[0].startsWith(idPrefix) || profiles[0].startsWith('meta-quest-touch'));
+        } else {
+          controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);
+        }
       }
     }
     if (!controllerMatch) { continue; }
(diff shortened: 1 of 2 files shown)


src/utils/tracked-controls.js
```diff
@@ -171,7 +171,11 @@ function findMatchingControllerWebXR (controllers, idPrefix, handedness, index,
           if (controllerMatch) { break; }
         }
       } else {
-        controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);
+        if (idPrefix === 'oculus-touch') {
+          controllerMatch = profiles.length > 0 && (profiles[0].startsWith(idPrefix) || profiles[0].startsWith('meta-quest-touch'));
```
dmarcos, line 175: profile ids should not be hard coded here. Something is not working higher in the call chain

felixtrz: Please feel free to commandeer this and make changes or just reference the changes made here in #5139. Hard coding it definitely is not a good solution, but I wasn't sure what you would like to do here to handle the two prefix mapping to one control problem

felixtrz: > profile ids should not be hard coded here. Something is not working higher in the call chain

Happy to test out any new changes you make!

dmarcos: I made changes to https://github.com/aframevr/aframe/pull/5139 Can you give it another try? Thanks

dmarcos:
Closing in favor of https://github.com/aframevr/aframe/pull/5139

PR #5141, opened 2022-10-21
Remove quest 2 controller offsets

felixtrz:
Remove quest 2 controller offsets now that the controller are directly getting transform data from gripSpace instead of targetRaySpace. (based on https://github.com/aframevr/aframe/commit/88e8db12deeb501bfaeb03fe015284197a1942e2) @dmarcos 

https://user-images.githubusercontent.com/11973041/197260871-c63d0ce4-bb6c-4ec1-a038-48c7e44fc757.mp4

We will still need to address the ray orientation problem separately


felixtrz:
Re-oriented target rays in correct position and direction

https://user-images.githubusercontent.com/11973041/197266896-95d2a2fa-4a76-4944-92ac-88687596c44d.mp4

dmarcos:
Thanks so much! 🥳 Merging this. For the ray we should eventually move to use the pose and `targetRaySpace` vs hardcoding the values.

Noticed the Quest Pro controllers render super dark. Maybe something we can address also in the future. 

Have a great weekend!

PR #5149, opened 2022-11-09
Fix triggerchanged event not firing from touch v3

msub2:
**Description:**

I found that the `triggerchanged` event was not firing with the current A-Frame master build, as `button` was undefined in `onButtonChanged` if the V3 code path was taken.

**Changes proposed:**
- Since the assignment to `button` is identical in both functions, it can be moved out and passed along to the V3 function when necessary. This puts it back in scope to be referenced for the event emit.

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -309,14 +309,14 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   onButtonChanged: function (evt) {
+    var button = this.mapping[this.data.hand].buttons[evt.detail.id];
+    if (!button) { return; }
     // move the button meshes
     if (this.isOculusTouchV3) {
-      this.onButtonChangedV3(evt);
+      this.onButtonChangedV3(button, evt);
     } else {
-      var button = this.mapping[this.data.hand].buttons[evt.detail.id];
       var buttonMeshes = this.buttonMeshes;
       var analogValue;
-      if (!button) { return; }
 
       if (button === 'trigger' || button === 'grip') { analogValue = evt.detail.state.value; }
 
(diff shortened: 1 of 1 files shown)


src/components/oculus-touch-controls.js
```diff
@@ -309,14 +309,14 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   onButtonChanged: function (evt) {
+    var button = this.mapping[this.data.hand].buttons[evt.detail.id];
+    if (!button) { return; }
     // move the button meshes
     if (this.isOculusTouchV3) {
-      this.onButtonChangedV3(evt);
+      this.onButtonChangedV3(button, evt);
```
dmarcos, line 316: can we define a variable inside `onButtonChangedV3` to decouple branches?

src/components/oculus-touch-controls.js
```diff
@@ -309,14 +309,14 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   onButtonChanged: function (evt) {
+    var button = this.mapping[this.data.hand].buttons[evt.detail.id];
+    if (!button) { return; }
```
dmarcos, line 313: Not sure I fully understand the bug now. Since we were already checking the button existence in `onButtonChangedV3` and early returing. this and old code should be equivalent

msub2: This jsfiddle is the simplest demonstration of what I mean https://jsfiddle.net/1xgnzv58/3/

That is what the current code does. If I have a touch v3 controller, it won't fire the `triggerchanged` event.

dmarcos: thanks!

PR #5150, opened 2022-11-10
[Docs] Fixed broken links on text components

Descent098:
**Description:** On the text component documentation it looks like about 1/3 of the examples 404. It seems like they reference a no longer existing git repo.

**Changes proposed:**
- Update links to valid locations for examples

### DIFF
--- a/docs/components/text.md
+++ b/docs/components/text.md
@@ -49,13 +49,13 @@ See more examples to see configuration of alignments, anchors, baselines,
 scaling, and auto-sizing.
 
 [exampletext]: https://aframe.io/aframe/examples/test/text/index.html
-[codetext]: https://aframe.io/aframe/examples/test/text/index.html
+[codetext]: https://github.com/aframevr/aframe/blob/master/examples/test/text/index.html
 [exampleanchors]: https://aframe.io/aframe/examples/test/text/anchors.html
-[codeanchors]: https://aframe.io/aframe/examples/test/text/anchors.html
-[examplescenarios]: https://aframe.io/aframe/examples/test/textt/scenarios.html
-[codescenarios]: https://aframe.io/aframe/examples/test/text/scenarios.html
+[codeanchors]: https://github.com/aframevr/aframe/blob/master/examples/test/text/anchors.html
+[examplescenarios]: https://aframe.io/aframe/examples/test/text/scenarios.html
+[codescenarios]: https://github.com/aframevr/aframe/blob/master/examples/test/text/scenarios.html
 [examplesizes]: https://aframe.io/aframe/examples/test/text/sizes.html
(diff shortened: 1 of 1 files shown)


docs/components/text.md
```diff
@@ -48,14 +48,14 @@ configuration.
...
-[exampletext]: https://aframevr.github.io/aframe/examples/test/text/index.html
-[codetext]: https://github.com/aframevr/aframe/blob/master/examples/test/text/index.html
-[exampleanchors]: https://aframevr.github.io/aframe/examples/test/text/anchors.html
-[codeanchors]: https://github.com/aframevr/aframe/blob/master/examples/test/text/anchors.html
-[examplescenarios]: https://aframevr.github.io/aframe/examples/test/text/scenarios.html
-[codescenarios]: https://github.com/aframevr/aframe/blob/master/examples/test/text/scenarios.html
-[examplesizes]: https://aframevr.github.io/aframe/examples/test/text/sizes.html
-[codesizes]: https://github.com/aframevr/aframe/blob/master/examples/test/text/sizes.html
+[exampletext]: https://aframe.io/aframe/examples/test/text/index.html
+[codetext]: https://aframe.io/aframe/examples/test/text/index.html
```
dmarcos, line 52: The links to [codeXXX] should point to the code otherwise they're redundant

dmarcos: What I mean is that we had two links before. One pointing to the example and the other to the code. See below

[examplesizes]: https://aframevr.github.io/aframe/examples/test/text/sizes.html
[codesizes]: https://github.com/aframevr/aframe/blob/master/examples/test/text/sizes.html

2nd link is still working. In your PR now both links point to the same URL in the examples which is redundant

Descent098: My bad, I should have been clearer in my last response. I saw what you mean and updated those links with a follow-up commit to resolve that problem. 

I made the initial commit quickly late at night when I ran into the example 404's and missed the code ones, so when I replaced them I hadn't checked. I have since fixed that in [this commit](https://github.com/aframevr/aframe/pull/5150/commits/c3ce16c1962af8991d232588f22b915bcb7ac40e) which happened after your Nov 15th comment. The code on this branch includes that fix as of the 16th (should be visible below this comment thread in the github UI)

If you look at the commit from Nov 16 (the current code in this branch) it resolves that problem so that now the links are correct for example:

```
[examplesizes]: https://aframe.io/aframe/examples/test/text/sizes.html
[codesizes]: https://github.com/aframevr/aframe/blob/master/examples/test/text/sizes.html
```

Let me know if there are any other concerns!

dmarcos:
Thanks so much for the patience and congrats on your first contribution 🏅

Issue #5144, opened 2022-11-02
Propose renaming throttleComponentChanged to throttledDebounce


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: All
- Reproducible Code Snippet or URL: N/A

Since 1.3.0, we have a new utility function:

[`AFRAME.utils.throttleComponentChanged`](https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/utils/index.js#L78)

This was introduced by PR #4980 to solve #4972

This truns out to be a generically useful utility function, but it doesn't have a generic name.

Recently in an A-Frame app I wanted to do exactly the same thing with the objectChange event from THREE.TransformControls.

But using a utility function called `throttleComponentChanged` to do a debounced throttle on an event called `objectChange` could have been confusing...!

I propose renaming this utility function to a name that reflects it's generic capability, namely to debounce an event in a throttled manner, which puts a cap on the rate at which the event fires, but also guarantees a correct end state.

My suggested name is `AFRAME.utils.throttledDebounce`

I'd be happy to do the PR to do all of this, assuming I can get agreement this is a good idea.

I expect we need to keep `AFRAME.utils.throttleComponentChanged`  for back compatibility, but it can just be mapped to `AFRAME.utils.throttledDebounce`.

vincentfretin:
Having throttle and debounce in the name is confusing to me.
See for example the definition of throttle and debounce functions in lodash
https://lodash.com/docs/4.17.15#throttle and https://lodash.com/docs/4.17.15#debounce
The current implementation here is really a throttle, not debounce.

diarmidmackenzie:
I don't think it is that simple.

We already have a throttle [here](https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/utils/index.js#L52)

And a debounce [here](https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/utils/index.js#L133)

Prior to #4980, componentchanged was throttled.  This created issues as described in #4972, because in a scenario where there is a rapid burst of changes, the final state may not be relayed.

With a debounce, the final state is always relayed reliably, but because a debounce ["waits before triggering the function execution, until the events stop happening so rapidly"](https://css-tricks.com/debouncing-throttling-explained-examples/#aa-leading-edge-or-immediate), you can end up with a long period of no updates at all.

[...]

PR #5151, opened 2022-11-12
5144 rename throttle component changed

diarmidmackenzie:
**Description:**

PR as per #5144

**Changes proposed:**
- name throttleComponentChanged -> throttleLeadingAndTrailing
- leave throttleComponentChanged utility function for 1.3.0 back-compatibility
- add documentation for throttleLeadingAndTrailing in utils API docs.

### DIFF
--- a/docs/core/utils.md
+++ b/docs/core/utils.md
@@ -212,6 +212,22 @@ AFRAME.registerComponent('foo', {
 });
 ```
 
+### `AFRAME.utils.throttleLeadingAndTrailing (function, minimumInterval [, optionalContext])`
+
+[lodash]: https://lodash.com/docs/#throttle
+
+Returns a throttled function that is called at most once every `minimumInterval` milliseconds, but ensures that the very last call of a burst gets deferred until the end of the interval.  This is useful when an event is used to trigger synchronization of state, and there is a need to converge to the correct final state following a burst of events.
+
+Example use cases:
+
+ * synchronizing state based on the componentchanged event
+ * following a mouse pointer using the mousemove event
+ * integrating with [THREE.TransformControls](https://threejs.org/docs/#examples/en/controls/TransformControls), via the objectChange event.
+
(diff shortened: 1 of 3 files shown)


vincentfretin (commented):
This looks good to me.

src/utils/index.js
```diff
@@ -99,6 +106,12 @@ module.exports.throttleComponentChanged = function (functionToThrottle, minimumI
   };
 };
 
+/**
+ * Identical to throttleLeadingAndTrailing.
+ * Exists for back-compatibility with 1.3.0.
+ */
+module.exports.throttleComponentChanged = module.exports.throttleLeadingAndTrailing;
```
dmarcos, line 113: I think we can just rename. Don't think many were relying on this. We can add a deprecation note in next release notes.

docs/core/utils.md
```diff
@@ -212,6 +212,22 @@ AFRAME.registerComponent('foo', {
...
+
+Example use cases:
+
+ * synchronizing state based on the componentchanged event
+ * following a mouse pointer using the mousemove event
+ * integrating with [THREE.TransformControls](https://threejs.org/docs/#examples/en/controls/TransformControls), via the objectChange event.
+
+A context such as `this` can be provided to handle function binding for convenience. 
+
+The same as [lodash's`throttle`][lodash], with `leading` and `trailing` options both set to `true`
```
dmarcos, line 229: wonder if this reference is a liability long term in case lodash API changes.

diarmidmackenzie: Maybe, but the existing docs already reference loadash throttle (the same link)

See: https://aframe.io/docs/1.3.0/core/utils.html#aframe-utils-throttle-function-minimuminterval-optionalcontext

This doesn't make the situation any worse, I think.

dmarcos:
Thank you!

Issue #5014, opened 2022-02-24
1.3.0 Hellow WebVR screenshot looks washed out


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: Chrome browser
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/sly-sixth-topaz?path=index.html%3A19%3A9

This is what a screenshot of the Hello Web VR app (Ctrl-Alt-S) used to look like
![screenshot--1645741701810](https://user-images.githubusercontent.com/16045703/155619220-33c2a8dd-2077-4892-a7f1-aa7a00c3074b.png)

This is what it looks like in 1.3.0 - to my eye this looks far too washed out.
![screenshot--1645741354272](https://user-images.githubusercontent.com/16045703/155619234-5b421fd0-c37f-448e-b3e5-555605feaebc.png)

This glitch uses a small JS script to run a version of 1.3.0 with the fix for PR4822 reverted, and gives the same screenshot output as 1.2.0 did.

I understand that PR4822 was added because screenshots were coming out too dark in some circumstances, but it seems that the way it's been fixed has just traded one problem for another.

I don't have any expertise around the various WebGLRenderTarget settings, so no idea what the "correct" fix is - but it looks like we still have issues here in getting the screenshot colors right in all cicumstances.

diarmidmackenzie:
Wild guess, from reading 4915 and 4917, is it a question of whether colorManagement is enabled?

The Hello WebVR example doesn't use colorManagement, whereas maybe the cases being looked at in 4915 and 4917 did?

diarmidmackenzie:
Here's a glitch with a totally unmodified "Hello WebVR" except for the addition of the "screenshot" attribute on the scene.
https://fire-vine-family.glitch.me/

Ctrl-Alt-S on this will reproduce the washed out screenshot.

kfarr:
Thanks for raising @diarmidmackenzie . I just took a look at the commit (it's been a while). I agree with your analysis.

- I think the user expectation is "the way my scene looks in the browser is the way the screenshot should look"
- A proposed solution would be for the screenshot code to respect the scene's `colorManagement` attribute to properly handle either case. If `true` then use the current method, if `false` then omit the line that sets sRGB.

dmarcos:
Yeah we should match color management on screen and screenshot component. Thanks

zackdove:
Is there an easy fix for this in the meantime?

[...]

PR #5157, opened 2022-11-19
Encoding should match scene's color management settings

diarmidmackenzie:
**Description:**

Fix for #5014

**Changes proposed:**
- encoding setting for screenshot is read from renderer, rather than assuming sRGB or Linear
- unit tests to check interaction between colorManagement setting on scene, and encoding used in screenshots.


dmarcos:
Thank you so much! 🏆 🥳

PR #5159, opened 2022-11-23
[docs] Update building-a-basic-scene.md

oroju1:
**Description:**
I went through the Building a Basic Scene tutorial for the first time and found it very helpful. I also found some areas that could be improved.

**Changes proposed:**
1. Adding Audio: Browsers do not like the "autoplay" attribute and the audio wont always play automatically. Instead, I added audio to the scene so that on clicking the scene the audio plays.

2.  Using the Asset Management System: This section does not describe how to use the asset folder to upload and use your own assets, so I added some instructions on how to do this.


dmarcos:
We should probably remove audio from a basic scene tutorial. It's a bit more advanced and can be frustrating

PR #5160, opened 2022-11-23
[docs] Update geometry.md

oroju1:
**Description:**
I noticed the icosahedron primitive was missing from the built-in geometries section.

**Changes proposed:**
- Added a section for the icosahedron primitive.


dmarcos:
Thanks

PR #5161, opened 2022-11-23
Update a-box.md

oroju1:
**Description:**
I noticed that the definitions in many of the primitives pages were not described in a similar way and sometimes there was not a definition. I also noticed that the link would not take you directly to the geometry.

Here, I updated a-box.

**Changes proposed:**
- I standardized the text for the primitive to match the formatting style of other primitives and to match the definition given in the geometry component page. 

- I also changed the geometry component page link to the anchor link in the geometry component page.

### DIFF
--- a/docs/primitives/a-box.md
+++ b/docs/primitives/a-box.md
@@ -9,7 +9,7 @@ section_order: 5
 source_code: src/extras/primitives/primitives/meshPrimitives.js
 ---
 
-The box primitive creates shapes such as boxes, cubes, or walls.
+The box primitive creates a quadrilateral using the [geometry component] with type set to `box`.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


docs/primitives/a-box.md
```diff
@@ -9,7 +9,7 @@ section_order: 5
 source_code: src/extras/primitives/primitives/meshPrimitives.js
 ---
 
-The box primitive creates shapes such as boxes, cubes, or walls.
+The box primitive creates a quadrilateral using the [geometry component] with type set to `box`.
```
dmarcos, line 12: I think this is more technical and confusing than the original for beginners. Also quadrilateral is usually applied for 2D shapes.

PR #5162, opened 2022-11-23
[docs] Update a-circle.md

oroju1:
**Description:**
I noticed that the definitions in many of the primitives pages were not described in a similar way and sometimes there was not a definition. I also noticed that the link would not take you directly to the geometry.

Here, I updated a-circle.

**Changes proposed:**
- I standardized the text for the primitive to match the formatting style of other primitives and to match the definition given in the geometry component page. 

- I also changed the geometry component page link to the anchor link in the geometry component page.

### DIFF
--- a/docs/primitives/a-circle.md
+++ b/docs/primitives/a-circle.md
@@ -6,8 +6,7 @@ parent_section: primitives
 source_code: src/extras/primitives/primitives/meshPrimitives.js
 ---
 
-The circle primitive creates circles surfaces using the [geometry][geometry]
-component with the type set to `circle`.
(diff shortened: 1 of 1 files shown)


docs/primitives/a-circle.md
```diff
@@ -6,8 +6,7 @@ parent_section: primitives
 source_code: src/extras/primitives/primitives/meshPrimitives.js
 ---
 
-The circle primitive creates circles surfaces using the [geometry][geometry]
-component with the type set to `circle`.
+The circle primitive creates a flat two-dimensional circle using the [geometry component] with type set to `circle`.
```
dmarcos, line 9: Same as my previous comment about a-box. This description while technically correct is less colloquial and beginner friendly.

Issue #5021, opened 2022-03-10
Broken Hyperlink on Aframe website - videosphere section


jadegamesuk:
Broken hyperlink on your Docs page. 
- A-Frame Version:1.3.0
- Platform / Device:
- Reproducible Code Snippet or URL: https://aframe.io/docs/1.3.0/primitives/a-videosphere.html#attributes - click on the common attributes link and you'll see it doesn't link anywhere. Please fix this.

msub2:
I don't believe this page ever existed. Even if you browse the repo [in the state it was at when that line was first added](https://github.com/aframevr/aframe/blob/b3837ddd93ef9bd1ddf21a70392a63b99fb1cbd8/docs/primitives/a-videosphere.md) it's not there.

dmarcos:
PR welcome to correct.

vincentfretin:
That link was removed, it doesn't exist in https://aframe.io/docs/1.7.0/primitives/a-videosphere.html#attributes
You can close.

PR #5175, opened 2022-11-24
Update a-videosphere.md

jadegamesuk:
Removed "common-attribute" link that has historically never actually existed

**Description:**
Removed link that has historically never existed. #5021 
**Changes proposed:**
-
-
-


dmarcos:
This fall through the cracks. Sorry for the wait and congrats on your first contribution!

PR #5176, opened 2022-12-01
Fix failing tests with firefox by mocking canvas.classList

vincentfretin:
**Description:**

Fix failing tests with firefox on master, see https://github.com/aframevr/aframe/actions/runs/3520429062/jobs/5901336232

**Changes proposed:**
- Don't create canvas element when we mock it completely afterwards
- Mock classList as well to fix tests on firefox (fix initially proposed by @diarmidmackenzie on https://github.com/aframevr/aframe/pull/5033#issuecomment-1320322715)
- Run tests on ubuntu-20.04 where firefox is still present. ubuntu-latest is now using 22.04 and gives an error saying it can't find the firefox binary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   test:
     name: Test Cases
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


vincentfretin:
All green. Good to merge @dmarcos

dmarcos:
Thank you!

PR #5186, opened 2022-12-19
Detach unused pool elements from THREE.js scene

diarmidmackenzie:
**Description:**

A PR based on the performance enhancement described here:
https://github.com/diarmidmackenzie/moonrider/issues/3#issuecomment-1356447670

The scope of this has grown a little.  I know you are pretty conservative about adding new components & API changes for A-Frame, so if you prefer me to dial back the scope of changes here, let me know.

To explain how I got from my original fix to this PR...

- Detaching entities from the scene & re-attaching might be valuable outside of the `pool` component, so better to provide a clean API on `a-entity` and handle the fiddly details (like window conditions around initialization) inside  `a-entity`

- There are existing use cases for which the new API is actually a great fit.  Use case described in `visible` component where you can toggle a group of entities in / out of the scene is actually better met by the new API.  Making such elements invisible cuts any cost of the GPU, but still leaves some CPU cost for updateMatrixWorld() and raycasting.  The new API would get rid of the CPU cost as well, as well as raycasting false positives, which have caused me problems in the past when usign this pattern.

- Support for this use case requires a component, so that detachment from THREE.js scene can be encoded declaratively in HTML.  Hence new "attached" component.

- It's important for raycasting to ignore object3Ds that are not part of the main scene graph: they are orphaned, and matrixWolrld values are not meaningful.  Hence updates to raycaster to explicitly ignore them.

- For raycaster auto-refresh to work, new events were needed to indicate detachment from scene & re-attachment to it.  I had a slight concern about perf impacts here, but I think any impacts that do occur can be mitigated by disabling autoRefresh on the raycaster component, as already described [here](https://aframe.io/docs/1.3.0/components/raycaster.html#manually-refreshing-the-target-entities-of-the-raycaster) 

**Changes proposed:**

- `a-entity` offers `detachFromScene()` and `attachToScene()` API calls
- `a-entity` emits new events, `attached-to-scene` and `detached-from-scene` indicating when these new operations complete.
- `pool` component uses these to ensure that unused pool entities do not appear in the scene graph (gives major perf boost for Moonrider)
- New `attached` component gives HTML-l;evel contol of detachment from THREE,js scene graph, very similar to `visible` component.
- `raycaster` updated to ignore all objects that are not rooted in the main THREE.js scene graph (including auto-updates)
- Documentation & UT for the above.  Documentation for toggling on/off a part of the HTML scene graph now points to `attach` component rather than `visible` component.

### DIFF
--- a/docs/components/attached.md
+++ b/docs/components/attached.md
@@ -7,15 +7,20 @@ source_code: src/components/attached.js
 examples: []
 ---
 
+[visible]: ./visible.md
+[pool]: ./pool.md
+[attach]: ../core/entity.md#attachtoscene-
+[detach]: ../core/entity.md#detachfromscene-
+
 The attached component determines whether an entity is attached to the THREE.js scene graph at all.
 
 All entities are attached by default.  Detaching an entity from the scene means that the entity and its descendants will have no interactions in the 3D scene at all: it is not rendered, it will not interact with raycasters.
 
-This is similar to the [visible](visible) component, but places even more limitations on the entity.
+This is similar to the [visible][visible] component, but places even more limitations on the entity.
 
 Invisible entities are not rendered, but their position in space is still updated every frame, and they can interact with raycasters, be checked for collisions etc.
 
-In contrast, when an entity is detached from the scene, by setting `attached="false"`, even these interactions do not occur, which will improve performance even further vs. just making an entity invisible.  For example, entity pools implemented by the [pool](pool) component detach entities in the pool when they are not in use.
+In contrast, when an entity is detached from the scene, by setting `attached="false"`, even these interactions do not occur, which will improve performance even further vs. just making an entity invisible.  For example, entity pools implemented by the [pool][pool] component detach entities in the pool when they are not in use.
 
 It's a common pattern to create container entities that contain an entire group of entities that you can flip on an off with `attached`.
 
@@ -36,7 +41,7 @@ It's a common pattern to create container entities that contain an entire group 
 
 ## Updating Attachment
 
-It is slightly faster to control attachment to the THREE.js scene using direct calls to [`attachToScene()`](../core/entity#attachtoscene) and [`detachFromScene()`](../core/entity#detachfromscene):
+It is slightly faster to control attachment to the THREE.js scene using direct calls to [`attachToScene()`][attach] and [`detachFromScene()`][detach]:
 
 ```js
 // direct use of entity interface
--- a/docs/components/attached.md
+++ b/docs/components/attached.md
@@ -1,0 +1,51 @@
+---
+title: attached
+type: components
+layout: docs
+parent_section: components
+source_code: src/components/attached.js
+examples: []
+---
+
+The attached component determines whether an entity is attached to the THREE.js scene graph at all.
+
+All entities are attached by default.  Detaching an entity from the scene means that the entity and its descendants will have no interactions in the 3D scene at all: it is not rendered, it will not interact with raycasters.
+
+This is similar to the [visible](visible) component, but places even more limitations on the entity.
+
+Invisible entities are not rendered, but their position in space is still updated every frame, and they can interact with raycasters, be checked for collisions etc.
+
+In contrast, when an entity is detached from the scene, by setting `attached="false"`, even these interactions do not occur, which will improve performance even further vs. just making an entity invisible.  For example, entity pools implemented by the [pool](pool) component detach entities in the pool when they are not in use.
+
+It's a common pattern to create container entities that contain an entire group of entities that you can flip on an off with `attached`.
+
+
+## Example
+
+```html
+<a-entity attached="false"></a-entity>
+```
+
+## Value
+
+| Value | Description                                                                            |
+|-------|----------------------------------------------------------------------------------------|
+| true  | The entity will be rendered and visible; the default value.                            |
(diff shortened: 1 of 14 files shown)


diarmidmackenzie:
Just been reviewing [this THREE.js thread](https://github.com/mrdoob/three.js/issues/14360), and wondering whether it would be simpler to:
- leave "detached" object3Ds in the scene graph
- add a new special flag (e.g. "ignore") to them
- update updateMatrixWorld() in super-three.js to return when it sees the "ignore" flag on an Object3D.

Per [this comment](https://github.com/mrdoob/three.js/issues/14360#issuecomment-402951981) it may be a little smoother / less performance-impactful to set a flag, rather than removing object3Ds from the scene graph...?

dmarcos:
Maybe we can do this more specific / self contained to the `pool` component. It handles internally attaching / detaching of entities to the scene graph. We might not need modifying a-entity, attach components, new events...

dmarcos:
> Just been reviewing [this THREE.js thread](https://github.com/mrdoob/three.js/issues/14360), and wondering whether it would be simpler to:
> 
> * leave "detached" object3Ds in the scene graph
> * add a new special flag (e.g. "detached") to them
> * update updateMatrixWorld() in super-three.js to return when it sees the "detached" flag on an Object3D.
> 
> Per [this comment](https://github.com/mrdoob/three.js/issues/14360#issuecomment-402951981) it may be a little smoother / less performance-impactful to set a flag, rather than removing object3Ds from the scene graph...?
> 
> EDIT: On further reflection, less convinced this is any better. THREE.js rendering code would also need updating to treat "detached" like "visible". Overall, feeling like a _more_ complex solution, rather than less complex. Propose leaving the PR as-is.

Maybe a more specific `skipMatrixUpdate` flag? Other scenarios where we might want to skip besides detached objects?

diarmidmackenzie:
> Maybe a more specific skipUpdateMatrix flag? Other scenarios where we might want to skip besides detached objects?

There is a pretty simillar existing flag in THREE.js `matrixWorldAutoUpdate`, which can be set to false.

Differences from what we'd need:
- this can be overridden using the "force" option on updateMatrixWorld.  I can't see any circumstances in which we'd want the new flag overridden
- it doesn't prevent rendering, only matrixUpdate calculations.

But having given this some more thought, I'm against the flag-based approach.
- It doesn't reduce code complexity - it just pushes complexity from A-Frame into super-three, where it's actually harder to manage, as code merges with evolving THREE.js code would have to be handled manually.
- I also suspect a flag would make things worse, rather than better from an overall performance point of view....

The new flag would need to be checked twice every frame for every object (once on updateMatrixWorld(), one on rendering). 
 Better way to minimize perf impact of these objects is to not have them in the tree at all.

Looking at the code for [`Object3D.add` ](https://github.com/mrdoob/three.js/blob/03349e98aa1f57a97433c0a3f17697594fdc0997/src/core/Object3D.js#L306) and [`Object3D.remove`](https://github.com/mrdoob/three.js/blob/03349e98aa1f57a97433c0a3f17697594fdc0997/src/core/Object3D.js#L350), for the case where you are adding / removing a single object, they basically boil down to an `Array.push()` for add, and an `Array.splice()` for remove, plus a single event emitted in each case.  I don't see how that is going to amount to a significant performance impact.

dmarcos:
What's your recommendation for this PR then? Still worth it?

diarmidmackenzie:
Yes, absolutely I think the PR is worth it.

Just think we should stick with the code as-is, deleting from / re-adding to the object tree, rather than trying to add new flags.

dmarcos:
Can we make this more self contained to the pool component? No new events or modifications to a-entity? If it turns useful we can later generalize

diarmidmackenzie:
Happy to drop "attached" component.

Dropping the events will leave some poor functionality around raycasting auto-updayes.

Detached objects will still be scanned for raycasting (perf impact) and will have outdated matrixWorlds (functional impact: raycasting triggering in the wrong places).

In order to keep the events, I think we need to put the changes into a-entity, not just the pool component (in theory I guess the pool component could emit the events, but that feels really messy vs. emitting them from a-entity).

dmarcos:
To keep it simple I think we can query for the raycaster entities in the pool component and call `setDirty` there instead of relying on an event.

diarmidmackenzie:
Feels like a bit of a hack, but yes it can work.

As well as this, I'll also create an "attached" component external to A-Frame, and see whether the community finds it useful.

dmarcos:
Yes, It couples the pool and raycaster component but relying to much on events makes code harder to understand too. In this case not super worried since the raycaster API has been stable for a while. If the attach / reattach approach is adopted and yields noticeable results we can revisit and generalize to something like your original PR. Thanks for being flexible

dmarcos:
We can squeeze this in 1.4.0 if you have some time. Thanks!

diarmidmackenzie:
Looking at it now...

diarmidmackenzie:
Alternative fix now available at #5188.  However I'm concerned about the perf impact of `querySelectorAll('[raycaster]')` every time we add to / remove from a pool.

Personally I think that using events is safer in terms of performance regression risks, but awaiting your feedback on this point...

diarmidmackenzie:
#5189 offered as an alternative, about which I have fewer perf concerns, in case it is preferred.  We should close whichever of #5188 or #5189 is not taken.

PR #5188, opened 2022-12-24
Reduced-scope version of detaching unused pool elements from scene graph

diarmidmackenzie:
Updated set of changes as per feedback in PR #5186.

However I am concerned about the performance characteristics of updating raycasters in this way.

A querySelectorAll() call on an attribute selector `[raycaster]` is potentially an expensive operation as it requires traversal of the entire HTML graph.

However the alternative of caching the set of raycaster entities in the scene is also problematic, as you'd have to figure out when to refresh that cache...

Hence I think it would be better to stick with emitting events, as per #5186, to keep the performance overhead of adding to / removing from pools to a minimum.

@dmarcos - for now, leaving the code as per your request in #5186, pending your feedback.  Happy to update this PR to include emitting the events if you agree that would be better from a performance pov.  Let me know...

### DIFF
--- a/docs/components/pool.md
+++ b/docs/components/pool.md
@@ -15,6 +15,8 @@ entities in dynamic scenes. Object pooling helps reduce garbage collection pause
 Note that entities requested from the pool are paused by default and you need 
 to call `.play()` in order to activate their components' tick functions.
 
+For performance reasons, unused entities in the pool are detached from the THREE.js scene graph, which means that they are not rendered, their matrices are not updated, and they are excluded from raycasting.
+
 ## Example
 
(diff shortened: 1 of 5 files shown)


diarmidmackenzie:
#5189 offered as an alternative, about which I have fewer perf concerns, in case it is preferred.  We should close whichever of #5188 or #5189 is not taken.

dmarcos:
Thanks so much for the hard work. The time you put researching and profiling is super super appreciated. It can be really tedious, on headset in particular. I know this PR is not ideal. Just trying to contain the added code complexity. Enjoy the holydays!

PR #5189, opened 2022-12-24
Deatch unused pool elements - reduced version but with events, no querySelectorAll('[raycaster]')

diarmidmackenzie:
Variant of #5188, addressing the perf concern I have about making a call to `querySelectorAll('[raycaster]')`

I'll have limited availability later, so offering this as an alternative now in case it is preferred.


diarmidmackenzie:
Close this, as we took #5188 instead.

PR #5190, opened 2022-12-24
Fix quest 1 controller and ray offsets

### LINKED ISSUE #5183: Quest 1 wrong controller and ray offsets
The Quest 1 controllers offset and ray offset are completely off when using laser-controls. Issue similar to Quest 2 that was fixed in #5141 by @felixtrz 

- A-Frame Version: master Dec 7, 2022 threejs r147
- Platform / Device: Meta Quest 1 on latest OS Version 47.0.0.282.341.428237303, Meta browser 24.4.0.22.60.426469926
- Reproducible Code Snippet or URL: `npm run start:https` and open in the Quest https://192.168.1.x:8080/showcase/ui/

I have no idea how to get the correct offsets to fix it.
It's been a long time this is like this, but I don't use laser-controls anymore in my projects, I'm using [handy-work](https://github.com/AdaRoseCannon/handy-work) that is using the webxr-input-profiles information and models where I don't have the issue.
I created the issue so you are aware of it before releasing aframe 1.4.0. If somebody could look at it to fix it or give me some instructions how to get the correct values, it would be great.

vincentfretin:
**Description:**

I tweaked the values by trial and error.
This close #5183 

**Changes proposed:**
- Modify modelPivotOffset modelPivotRotation rayOrigin for oculus-touch-v2 that are used for Quest 1 controllers

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -57,15 +57,15 @@ var CONTROLLER_PROPERTIES = {
   'oculus-touch-v2': {
     left: {
       modelUrl: TOUCH_CONTROLLER_MODEL_BASE_URL + 'gen2-left.gltf',
-      rayOrigin: {origin: {x: -0.01, y: 0, z: -0.02}, direction: {x: 0, y: -0.5, z: -1}},
-      modelPivotOffset: new THREE.Vector3(0, 0, 0),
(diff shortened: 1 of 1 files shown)


dmarcos:
Merging this for now. We need to get rid of these harcoded values in the future. I believe we can retrieve the orientation and position of the laser from the WebXR API itself.

Issue #4814, opened 2021-03-03
Disable cardboard mode for iPads


dmarcos:
There are no humongous cardboard holders as far as I know :)

We should perhaps offer a fullscreen button instead.

vincentfretin:
With the changes in #5102
On iPad when you are in desktop mode (the default), the VR button make it fullscreen, but somewhat stretched horizontally. I'm not going to investigate that myself.
If you switch to mobile mode and click the VR button, you have the stereo working properly, but yeah not so useful with a tablet.

dmarcos:
I'm committing some code to disable cardboard mode by default and start phasing it out. Cardboard holders are produced less and less and hard to come by the one that works for your device and screen size. The experience is not great, usage low and getting it right is becoming an intractable problem. The new behavior I'm aiming towards is:

- Desktop on a browser with WebXR. Show VR button
- Desktop on a browser without WebXR. Show fullscreen button (I'll implement this in a follow up commit)
- Mobile on a browser with WebXR support. Show VR / AR buttons when applicable.
- Mobile on a browser without WebXR support. Don't show any buttons.

vincentfretin:
@dmarcos I don't know if you know it, your changes in commit https://github.com/aframevr/aframe/commit/042a3d6b7087a632c5165227b14bc37573375cde indeed disabled the VR button for Safari iPad but the button only disappear if you select mobile version of the page from the url bar. Desktop version is the default since years now on iPad, so we still can see the VR button by default here (and we get just a black render on desktop version). I'm not sure if we can better detect Safari iPad in desktop mode to hide the VR button. I don't have an iPhone, is the mobile version the default there? Does your changes hide the button on iPhone?
On Chome Android, we still see the VR button.

dmarcos:
On Chrome Android WebXR is available hence the VR button is displayed. On iPad I will test but VR button should be expected until I replace the icon with a fullscreen one (same as regular desktop). Behavior should be entering fullscreen but not sure what's the expected Safari for iPad fullscreen situation anymore. There should not be any buttons displayed on iPhone

vincentfretin:
Ok, all good then for Chrome Android.
For iPhone, Cardboard is working again since we fixed WebVRManager in #5102 but it may break again in the future if nobody is actively testing it and fixing it that's for sure so I understand we want to disable it by default.
[...]

PR #5191, opened 2022-12-25
Hide fullscreen button in desktop mode on iPad

vincentfretin:
**Description:**

As discussed in #4814 hide the fullscreen button on iPad in desktop mode when `cardboardModeEnabled` is `false`, it just rendered a black screen.

**Changes proposed:**
- Using `utils.device.isMobileDeviceRequestingDesktopSite()` to detect desktop mode on iPad

### DIFF
--- a/src/components/scene/vr-mode-ui.js
+++ b/src/components/scene/vr-mode-ui.js
@@ -135,7 +135,7 @@ module.exports.Component = registerComponent('vr-mode-ui', {
     var sceneEl = this.el;
     if (!this.enterVREl) { return; }
     if (sceneEl.is('vr-mode') ||
-       (sceneEl.isMobile && !this.data.cardboardModeEnabled && !utils.device.checkVRSupport())) {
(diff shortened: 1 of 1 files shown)


dmarcos:
Is there any error in console? Fullscreen API not working on iPads? I tried quickly on one and the browser is entering some form of fullscreen mode (notice the x button on the top / left corner). I would investigate a bit more why we're rendering black

vincentfretin:
I don't know if there is an error or not, I have no idea how to get console.log on iPad Safari. I tried using VorlonJS last time but couldn't make it work anymore. I don't have any macOS.

dmarcos:
I couldn't see any errors in console. Safari behavior is really bizarre. The fullscreen seems to engage: denoted by the X button on the top-left corner. Nothing renders but one can tap the x to exit the mode. I couldn't find any info anywhere about why this behavior. If fullscreen is really not available on iPad I would expect an error. As usual Safari is frustrating. Gonna merge this but hopefully we find a better way to handle this.

PR #5195, opened 2022-12-27
Don't generate aframe-master.min.js.LICENSE.txt file

### LINKED ISSUE #5192: Don't generate license file
How can we disable the generation of `dist/aframe-master.min.js.LICENSE.txt`? I was not able to figure out. Thanks! @vincentfretin

vincentfretin:
**Description:**

Don't generate aframe-master.min.js.LICENSE.txt file.
This closes #5192

See documentation:
https://github.com/webpack-contrib/terser-webpack-plugin#remove-comments
and default config was:
https://github.com/webpack/webpack/blob/9fcaa243573005d6fdece9a3f8d89a0e8b399613/lib/config/defaults.js#L1160-L1174

### DIFF
--- a/webpack.prod.config.js
+++ b/webpack.prod.config.js
@@ -1,6 +1,7 @@
 var path = require('path');
 var merge = require('webpack-merge').merge;
 var commonConfiguration = require('./webpack.config.js');
+var TerserPlugin = require('terser-webpack-plugin');
 
 module.exports = merge(commonConfiguration, {
   output: {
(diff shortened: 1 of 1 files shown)


vincentfretin:
I verified, it produces the same minified output.

dmarcos:
Thanks!

vincentfretin:
There is just a single difference in the produced file, the first line `/*! For license information please see aframe.min.js.LICENSE.txt */` is removed. If you remove it manually, the file has the same md5.

PR #5196, opened 2022-12-27
Fix prerelease command (fix #5193)

### LINKED ISSUE #5193: `npm run prerelease` no longer works as it used to with webpack
This line doesn't work as it used to:

https://github.com/aframevr/aframe/blob/master/scripts/release.js#L21

Change the following command to a new "old version" and "new version" numbers:

https://github.com/aframevr/aframe/blob/master/package.json#L19

Then run `npm run prerelease`. We should see the "old version" removed "aframe-v1.3.0.js" and replaced with "aframe-v1.4.0.js" with the "new version". Now it just removes the old version and generates new builds with name "aframe-master.js". The file name doesn't get replaced.

@vincentfretin

vincentfretin:
**Description:**
Fix the `npm run prerelease` command to properly remove old version and generate new version.
This closes #5193

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -15,11 +15,11 @@ if (!prevVersion || !nextVersion) {
 let distMin;
 let distMax;
 if (process.env.FOR_RELEASE) {
-  distMin = pkg.scripts['dist:min'].replace(/-master/g, '');
-  distMax = pkg.scripts['dist:max'].replace(/-master/g, '');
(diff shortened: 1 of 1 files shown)


dmarcos:
We use `process.env.FOR_RELEASE` in the release process https://github.com/aframevr/aframe/wiki/Release-Process

vincentfretin:
Oh I didn't know this was documented on the wiki, thanks. I reintroduced it.

dmarcos:
Awesome! Thanks so much!

Issue #5137, opened 2022-10-17
Crash when exiting VR on Master Branch #0fe1f286


wyzwon:
**Description:**
Upon exiting VR, I get the following error in the browser console:
Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener') at XRSession.onSessionEnd (three.js:18792:1)
This error leaves the page entirely unresponsive and does not show up in release 1.3.0

- A-Frame Version: Master build #0fe1f286
- Platform / Device: gen 1 HTC Vive / Windows 10 / Steam VR / Chrome 106.0.5249.119
- Reproducible Code Snippet or URL:
Enter VR mode then hit Esc to exit VR. The scene should then become un-responsive.
Any A-frame scene using build #0fe1f286. Such as from:
<script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@c27d4a2156bde2ec249600fca4967ba49fecfdc3/dist/aframe-master.min.js"></script>

dmarcos:
Chrome WebXR support on desktop has had a lot quirks since transition to OpenXR. It would be worth trying on Quest / Oculus browser and see it the issue reproduces consistently.

wyzwon:
I was able to exit VR on the Quest 2 without any fatal crash on both the standalone Quest browser and via a desktop link. I also reconfirmed the issue still persists for the Vive in Chrome version 107.0.5304.88

dmarcos:
Chrome WebXR on desktop has a lot of quirks. Might want to read https://github.com/aframevr/aframe/issues/4709 and add your experience if not covered

wyzwon:
I have narrowed down the introduction of the bug to between c937479ed9f4e70c1cadacc4e14af6a9a82e3cef and d8de4098e6985f604e36002e00bd3b6cc37c8385. The only code changes between these two builds appear to be:

1. An update to wasd-controls (480a303)
2. A super-three bump from 0.137.0 to 0.141.0 (837023c)

I am also confirming that the error still persists in version 1.4.0 running Chrome 108.0.5359.125/Steam VR beta 1.25.2

I did notice that even in older versions, the relaunched game is a little twitchy, but I suspect it's unrelated to the bug at hand.

vincentfretin:
I remember I saw this error on Quest 1 Meta browser sometimes while I was developing. I didn't really care at that time because I just refreshed the page between each VR test. So I don't know if the page was also unresponsive or not in my case.
[...]

PR #5198, opened 2022-12-30
Fix random error when exiting VR

vincentfretin:
**Description:**
Fix random error when exiting VR that could make the page unresponsive.
`Uncaught TypeError: Cannot read properties of null (reading 'removeEventListener') at XRSession.onSessionEnd (three.js:18792:1)`
See discussion in #5137

**Changes proposed:**
- Remove setSession(null) call on exiting VR, threejs is already setting session to null in its onSessionEnd listener


vincentfretin:
I tested the change on Quest 1 on https://192.168.1.15:8080/showcase/ui/
enter VR, exit VR, enter VR. This still works.

dmarcos:
You're awesome. Thank you

PR #5199, opened 2023-01-02
Correct documentation for a-cylinder radius

diarmidmackenzie:
Unlike THREE.js CylinderGeometry, the A-Frame `cylinder` geometry does not have distinct radii for top & bottom.

Distinct radii for top & bottom are only provided by the A-Frame `cone` geometry (which also uses the THREE.js CylinderGeometry).

The docs for `a-cylinder` are currently wrong, referring to `radius-bottom` and `radius-top` attributes that don't exist, and not mentioning the `radius` attribute that does exist.

This change fixes them.


dmarcos:
Thank you!

Issue #5203, opened 2023-01-04
Passthrough of attributes on primitives such as a-sphere seems to be broken in 1.4.0


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.4.0
- Platform / Device: All 
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/primitive-attribute-passthrough?path=index.html%3A32%3A7

Example shows a couple of different ways of setting the color & radius of an a-sphere.

```
      AFRAME.registerComponent('big-blue-1', {
        init() {
          this.el.setAttribute('geometry', 'radius: 0.75')
          this.el.setAttribute('material', 'color: blue')
        }
      })
```

```
      AFRAME.registerComponent('big-blue-2', {
        init() {
          this.el.setAttribute('radius', '0.75')
          this.el.setAttribute('color', 'blue')
        }
      })
```

In 1.3.0 both big-blue-1 and big-blue-2 work as they should.

In 1.4.0, big-blue-1 works OK, but big-blue-2 no longer alters the sphere's color or radius.

Found when updating aframe-physics-system examples to 1.4.0.  I haven't yet looked at why this is.  For now I am working around by updating my examples to the alternative approach that still works.

dmarcos:
On https://aframe.glitch.me/ changing color works for me. bug gotta be more nuanced

![change color](https://p199.p4.n0.cdn.getcloudapp.com/items/jkuOEx8v/1879413b-761d-418e-bd52-44d48189d453.jpg?source=viewer&v=31ec8f88b4d260e162fb269d5c51d146)

dmarcos:
What about if you set `material` and `geometry` as component dependencies?

diarmidmackenzie:
Looks like some sort of race condition so only affects attributes set on entity creation.  E.g. this works OK.

```
      AFRAME.registerComponent('big-blue-2', {
        init() {
          setTimeout(() => {
              this.el.setAttribute('radius', '0.75')
              this.el.setAttribute('color', 'blue')
          }, 0)
        }
      })
 ```
 
 However, adding dependencies as you suggested doesn't fix things - this still doesn't work
 ```
       AFRAME.registerComponent('big-blue-2', {
        dependencies: ['geometry', 'material'],
        init() {
          this.el.setAttribute('radius', '0.75')
          this.el.setAttribute('color', 'blue')
        }
      })
```

Other observations:
- re-ordering the components in the `<a-sphere>` doesn't make any difference:
```
[...]

PR #5206, opened 2023-01-06
Make sure attribute mutation observers are in place before initializing components on an entity.

diarmidmackenzie:
**Description:**

See #5203 for background.  This PR is a proposed fix for that issue, as described [here](https://github.com/aframevr/aframe/issues/5203#issuecomment-1373811602) 

**Changes proposed:**

Set up mutation obervers for attributes *before* initializing components on an entity.
This is important in the case where the component sets attributes that need obsevers in place to be set correctly - a particular example being attributes like `radius` and `color` on `a-sphere`, which map to `geometry.radius` and `material.color`.

### DIFF
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -140,8 +140,8 @@ class ANode extends HTMLElement {
       });
 
       self.hasLoaded = true;
-      if (cb) { cb(); }
       self.setupMutationObserver();
+      if (cb) { cb(); }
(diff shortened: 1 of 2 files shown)


vincentfretin (approved):
LGTM. Thanks for adding the tests!

dmarcos:
Thanks starting the year kicking ass

Issue #5207, opened 2023-01-07
Outdated links & example code in documentation


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.4.0
- Platform / Device: all
- Reproducible Code Snippet or URL: See below.

There are a range of very out-of-date links in the documentation, especially the getting started guide, that it would be great to fix.

Since upgrading to 1.4.0, aframe-teleport-controls has been broken.  I was concerned about this as I knew it was referenced from the docs.  Discussing with @vincentfretin we agreed that it would be a good idea to change the docs to refer to [blink-controls](https://github.com/jure/aframe-blink-controls/), which is an actively maintained teleport component.

I started working on this [here](https://github.com/diarmidmackenzie/aframe/tree/teleport-controls-to-blink-controls), but in the process of updating the docs, realized that there are a heap of similar issues that have stacked up.

Just considering the pages I've been editing, and a little bit of additional digging, the other issues are:

1. Various outdated references to the A-Frame Registry.  Easy to fix these to refer to the new Component Directory on the wiki, so I did that on the branch above.

2. aincraft source code built in the example is also referred to as being in GitHub [here](https://github.com/supermedium/superframe/tree/csstricks/scenes/aincraft/) in CodePen [here](https://production-assets.codepen.io/assets/embed/ei.js), and in Glitch [here]( - all substantially out of date.  Suggest we update the GitHub version to match the code in the tutorial, and retire the CodePen version.

Not sure about the Glitch version - I think it does add value over & above GitHub + GH Pages due to the remix option, so probably worth keeping.  I assume there is already a set of Glitch demos that do get updated on a per-release basis, so shouldn't be too much overhead to add this to that list?  (once we have an up-to-date working version on GitHub).  I'm guessing only @dmarcos has access to the A-Frame Glitch account?

3. The [ECS tutorial page](https://aframe.io/docs/1.3.0/introduction/entity-component-system.html) makes extensive use of the aframe-particle-system component, which has had a string of recent compatibility issues... @vincentfretin and I recently discussed bringing this into c-frame and under better maintenance.  If we do that, we could update this page to refer to that version.

There's also a Glitch link here - again, probably worth keeping if we can, once we have updated working code.

4. The Building a 360 image gallery probably needs a bunch of updates similar to the Minecraft Demo, needs Glitch updated etc.
https://aframe.io/docs/master/guides/building-a-360-image-gallery.html
Currently it runs at 1.0.4

[...]

Issue #48, opened 2015-10-14
Add sphere template


jcarpenter:
Enables devs to add cylinders to their scenes. 

Dev can specify:
- Radius (default .5)
- Horizontal segments (default 36)
- Height segments (default 18)

PR #5211, opened 2023-01-11
Bring "Community components" example into A-Frame Examples.

diarmidmackenzie:
**Description:**

1st PR for #5207.  This updates the example intended to show use of community components.

A few notes on some choices I took here:

- gradient-sky has been broken since 0.8.0, so I moved to a more current [working sky component](https://www.npmjs.com/package/aframe-simple-sun-sky).  There is a [sun-sky component in superframe](https://github.com/supermedium/superframe/tree/master/components/sun-sky) that I'd have liked to have used, but it's also not working at the moment.  Hopefully we can repair that (along with the rest of superframe), and then maybe move to that.  For now, this seems the best option.

- I've standardized on jsdlivr rather than unpkg URLs.  jsdelivr is more flexible as it can deliver resources from github as well as NPM, and seems to have more widespread usage in the community these days.

- The docs have a long section explainign unpkg CDN.  I'd like to rework this to cover multiple CDNs including jsdelivr, but I don't want to complicate this PR.  So for now I have left the unpkg explanation as it is, and just added a brief intro to jsdelivr, so that the URLs in the example can be understood by a reader.

As mentioned in #5207, this is the 1st of 3 PRs to bring external content referenced in the documentation into the scope of the A-Frame Examples.


docs/introduction/entity-component-system.md
```diff
@@ -451,27 +451,37 @@ Now we can include it into our HTML:
...
 
 ```html
 <html>
   <head>
-    <script src="https://aframe.io/releases/1.4.0/aframe.min.js"></script>
-    <script src="https://unpkg.com/aframe-animation-component@3.2.1/dist/aframe-animation-component.min.js"></script>
-    <script src="https://unpkg.com/aframe-particle-system-component@1.0.x/dist/aframe-particle-system-component.min.js"></script>
-    <script src="https://unpkg.com/aframe-extras.ocean@%5E3.5.x/dist/aframe-extras.ocean.min.js"></script>
-    <script src="https://unpkg.com/aframe-gradient-sky@1.4.0/dist/gradientsky.min.js"></script>
+    <title>Community Components Example</title>
```
dmarcos, line 476: We now have duplicated code here and the example that have to be in sync. It would be cool if there's a markdown way to display html from the file hosted on gihub

docs/introduction/entity-component-system.md
```diff
@@ -451,27 +451,37 @@ Now we can include it into our HTML:
...
+    <meta name="description" content="Community Components Example">
+    <script src="../../../dist/aframe-master.js"></script>
```
dmarcos, line 478: For embedded code in docs we can point to the A-Frame version. That gets updated automatically. If we point to dist/aframe-master once cannot copy / paste the code.  

One of the features I once requested to glitch is having a glitch backed up by a git directory. These way we could have all the glitches in a single repo and update them all on a new release.

vincentfretin: FYI for networked-aframe, I update the naf-examples glitch by opening the glitch terminal and git pull from github, see my procedure https://github.com/networked-aframe/networked-aframe/blob/master/docs/release-instructions.md
Each glitch is actually a git repository where it autocommit the changes.

diarmidmackenzie: I'm proposing these lines in each example:

```
    <!-- If running this example without a local copy of A-Frame, replace this next line with:
    <script src="https://aframe.io/releases/<release_number>/aframe.min.js"></script>
    -->
    <script src="../../../dist/aframe-master.js"></script>   
```

Not perfect, but feels like a reasonable compromise?

Same applies to #5213, #5214, #5215.

docs/introduction/entity-component-system.md
diarmidmackenzie: One option:
https://github.com/zakhenry/embedme

A limitation is that code fragments are indexable only by line number.  If line numbers change, extracts will get out of sync.  That's discussed (together with some possible solutions) [here](https://github.com/zakhenry/embedme/issues/48) but there doesn't seem to be a good solution yet (and momentum has stalled since 2020).

How widespread an issue do you think this is in the docs?  Is there enough benefit to be worth the overhead of adding new tools like embedme to the pipeline?  Is it worth doing if snippets have to be by line number, rather than tag-based as #48 will enable?

diarmidmackenzie:
Moved from examples\showcase to examples\docs.  Also fixed some bad links left over from #5215 where aincraft demo was moved from examples\showcase to examples\docs.

diarmidmackenzie:
Following up on this PR & also #5213

I believe both can be merged.  Unresolved issues that came up in review were:
- auto-sync to glitch.  Technically complicated.  [Separate issue](https://github.com/aframevr/aframe/issues/5222) tracks the options here: 
- whether to refer to `../../../dist/aframe-master.js` or `aframe.io/releases/[release]/aframe.min.js`  - we agreed the best option is to refer to dist, with a comment indicating how to use aframe.io when copy/pasting the code.  5214 and 5215 already merged using this pattern.  If we implement ato-sync to glitch, it will be possible to adjust the code as part of the sync operation so that it works on glitch without a local aframe library.
- [desire to avoid duplication of code between documentation and example](https://github.com/aframevr/aframe/pull/5211#discussion_r1068829180) - unfortunately it doesn't look like any good options available here.  As per [this comment](https://github.com/aframevr/aframe/pull/5211#discussion_r1069218528), embedme is promising, but will create maintenance pain as code snippets have to be referenced by line number, rather than by tags.

While it would be nice to solve some of the above issues, none is simple to solve, and I don't see a good case for keeping outdated example code in the meantime.

So on the basis that we merged #5124 and #5215 already, I think we should be OK to merge this one & #5213

docs/introduction/entity-component-system.md
dmarcos: I think this embedded code should point to the latest A-Frame version [as it was before](https://github.com/aframevr/aframe/pull/5211/files#diff-91dd1cf1fe2f714a59bcfe0a8260e0d2b28b1a4363e224280ce654124d2bf6a3L467). The doc update script updates the version automatically upon publication. Does it make sense?

dmarcos:
Thanks for the patience. Commented. Just a nit.

docs/introduction/entity-component-system.md
diarmidmackenzie: OK, makes sense.  Changed back to 1.4.0 version.  Quoted code in the docs won't exactly match the code in the linked example, but the difference is hopefully not going to be too confusing to anyone...

dmarcos:
Super thanks!

PR #5213, opened 2023-01-12
Documentation: Guide update: Updates to "Basic Scene" Guide

diarmidmackenzie:
**Description:**

2nd of what will in fact be 4 updates to bring external glicth examples up-to-date & into the A-Frame repo, per #5207.

- Bring code from Glitch into A-Frame Examples
- Code up-to-date with latest versions
- Code and documentation match


examples/showcase/basic-scene/index.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Basic Scene with Environment - A-Frame</title>
+    <meta name="description" content="Basic Scene with Environment - A-Frame">
+    <script src="https://aframe.io/releases/1.4.0/aframe.min.js"></script>
```
vincentfretin, line 6: You need to use `../../../dist/aframe-master.js` here like the other showcases.

examples/showcase/basic-scene/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <script src="https://unpkg.com/aframe-environment-component@1.3.1/dist/aframe-environment-component.min.js"></script>
```
vincentfretin, line 7: update to environment 1.3.2

docs/guides/building-a-basic-scene.md
```diff
@@ -229,7 +228,7 @@ First, include the environment component using a script tag after A-Frame:
 ```html
 <head>
   <script src="https://aframe.io/releases/1.4.0/aframe.min.js"></script>
-  <script src="https://unpkg.com/aframe-environment-component/dist/aframe-environment-component.min.js"></script>
+  <script src="https://unpkg.com/aframe-environment-component@1.3.1/dist/aframe-environment-component.min.js"></script>
```
vincentfretin, line 231: update to environment 1.3.2

examples/showcase/basic-scene/index.html
vincentfretin: If you use a versioned url here for education purpose so that the user can copy paste the code, that's another story but be aware that currently the `npm run prerelease` command that use the `scripts/release.js` script only update the version in `*.md` files, not in `*.html` files.

diarmidmackenzie: Ok, hadn't appreciated that limitation.

I'll point to master and add a comment explaining what to do if you want to run the code in a different context.

vincentfretin: Yes. It's a tradeoff. Also as the release manager, before a release I want all the examples to use latest aframe-master to test for any regression.

diarmidmackenzie:
Moved from examples\showcase to examples\docs.

diarmidmackenzie:
Following up on this PR & also https://github.com/aframevr/aframe/pull/5211

I believe both can be merged. Unresolved issues that came up in review were:

auto-sync to glitch. Technically complicated. https://github.com/aframevr/aframe/issues/5222 tracks the options here:
whether to refer to ../../../dist/aframe-master.js or aframe.io/releases/[release]/aframe.min.js - we agreed the best option is to refer to dist, with a comment indicating how to use aframe.io when copy/pasting the code. 5214 and 5215 already merged using this pattern. If we implement sync-to-glitch, it will be possible to adjust the code as part of the sync operation so that it works on glitch without a local aframe library.
https://github.com/aframevr/aframe/pull/5211#discussion_r1068829180 - unfortunately it doesn't look like any good options available here. As per https://github.com/aframevr/aframe/pull/5211#discussion_r1069218528, embedme is promising, but will create maintenance pain as code snippets have to be referenced by line number, rather than by tags.
While it would be nice to solve some of the above issues, none is simple to solve, and I don't see a good case for keeping outdated example code in the meantime.

So on the basis that we merged https://github.com/aframevr/aframe/pull/5124 and https://github.com/aframevr/aframe/pull/5215 already, I think we should be OK to merge this one & https://github.com/aframevr/aframe/pull/5211

docs/guides/building-a-basic-scene.md
diarmidmackenzie: fixed

examples/showcase/basic-scene/index.html
diarmidmackenzie: fixed

dmarcos:
Ugh. sorry. this fell through the cracks and is a good contribution. I'm all about removing glitches from the docs.

If you can rebase I'll merge! Thanks so much

diarmidmackenzie:
Rebased.

I also reworked this a bit to deal with the "optional" part of the tutorial, so now we have 2 examples, one with the pre-canned environment, and one with the custom environment.

dmarcos:
Thanks so much for jumping back on it so quickly

PR #5217, opened 2023-01-13
Implement low / high refresh rates on WebXR devices

diarmidmackenzie:
**Description:**

The renderer component offers a property `highRefreshRate`.
https://aframe.io/docs/1.4.0/components/renderer.html#properties_highrefreshrate

Currently this is only implemented for WebVR.(sets the `highRefreshRate` value in `presentationAttributes`)

This means that in WebXR, A-Frame does not influence frame rate, meaning expriences run at the frame rate preferred by the browser.  On Quest 2 that is 90fps, which is arguably too high (and often results in experiences missing frames & hitting 45fps).

**Changes proposed:**

This PR provides a WebXR implementation of the `highRefreshRate` for devices/browsers that support the WebXR API to control frame rate:
https://www.w3.org/TR/webxr/#dom-xrsession-updatetargetframerate

Proposed behaviour:
1 - if device does not support 90fps, low -> 60, high -> 72
2 - if device supports 90fps, low -> 72, high -> 90

Currently 1 = Quest 1; 2 = Quest 2 & Quest Pro.

Other browsers / devices do not yet support the WebXR spec linked above, but when they do this seems like sensible default behaviour - it can be further tuned when additional info is known about what frame rates these browsers support.

**Other notes**

This PR is not ready to merge yet.  Sharing for an initial view of what the code / function should be.  If we get agreement on that, I'll add UTs & updated documentation to reflect the changes.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -301,6 +301,19 @@ class AScene extends AEntity {
               vrManager.setSession(xrSession).then(function () {
                 vrManager.setFoveation(rendererSystem.foveationLevel);
               });
+              const rates = xrSession.supportedFrameRates;
+              if (rates && xrSession.updateTargetFrameRate) {
+                let targetRate;
+                if (rates.includes(90)) {
+                  targetRate = rendererSystem.highRefreshRate ? 90 : 72;
+                } else {
+                  targetRate = rendererSystem.highRefreshRate ? 72 : 60;
+                }
+                xrSession.updateTargetFrameRate(targetRate).catch(function (error) {
+                  console.warn('failed to set target frame rate of ' + targetRate + '. Error info: ' + error);
+                });
+              }
+
               xrSession.addEventListener('end', self.exitVRBound);
               enterVRSuccess(resolve);
             },
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -301,19 +301,7 @@ class AScene extends AEntity {
               vrManager.setSession(xrSession).then(function () {
                 vrManager.setFoveation(rendererSystem.foveationLevel);
               });
-              const rates = xrSession.supportedFrameRates;
-              if (rates && xrSession.updateTargetFrameRate) {
-                let targetRate;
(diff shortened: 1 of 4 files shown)


src/core/scene/a-scene.js
```diff
@@ -301,6 +301,19 @@ class AScene extends AEntity {
               vrManager.setSession(xrSession).then(function () {
                 vrManager.setFoveation(rendererSystem.foveationLevel);
               });
+              const rates = xrSession.supportedFrameRates;
+              if (rates && xrSession.updateTargetFrameRate) {
+                let targetRate;
+                if (rates.includes(90)) {
```
diarmidmackenzie, line 307: Usually I prefer to avoid "mmagic numbers" in code, but I'm not sure that replacing these numbers with constants would do anything to aid maintenance or readability.

If you'd like me to use constants like FRAME_RATE_90 etc., please point me to where you'd like them defined.  Top of this module?  constants/index.js?  Or somewhere else?

src/core/scene/a-scene.js
```diff
@@ -301,6 +301,19 @@ class AScene extends AEntity {
               vrManager.setSession(xrSession).then(function () {
                 vrManager.setFoveation(rendererSystem.foveationLevel);
               });
+              const rates = xrSession.supportedFrameRates;
```
dmarcos, line 304: no const / let only var

src/core/scene/a-scene.js
dmarcos: can this be factored out to a function? `setImmersiveModeRefreshRate`. May be the code should be part of the renderer component?

diarmidmackenzie: OK, good suggestion - I made this a function on the renderer component & added unit tets for that

src/core/scene/a-scene.js
diarmidmackenzie: Fixed.

diarmidmackenzie:
Docs & UTs up to date, so I think this is now mergeable.

dmarcos:
Thank you!

Issue #5220, opened 2023-01-18
oculus-touch-controls.js:347 Uncaught TypeError: Cannot read properties of undefined (reading 'trigger')


mikemainguy:
**Description:**

- A-Frame Version: a-frame master
- Platform / Device: oculus quest 2
- Reproducible Code Snippet or URL:  https://feather-glistening-jaw.glitch.me

Navigate to above site:

oculus-touch-controls.js:347 Uncaught TypeError: Cannot read properties of undefined (reading 'trigger')
    at NewComponent.onButtonChangedV3 (oculus-touch-controls.js:347:48)
    at NewComponent.onButtonChanged (oculus-touch-controls.js:316:12)
    at AEntity.bound (bind.js:12:17)
    at AEntity.emit (a-node.js:287:10)
    at NewComponent.handleButton (tracked-controls-webxr.js:134:13)
    at NewComponent.updateButtons (tracked-controls-webxr.js:115:12)
    at NewComponent.tick (tracked-controls-webxr.js:81:12)
    at AScene.tick (a-scene.js:696:30)
    at AScene.render (a-scene.js:742:32)
    at bound (bind.js:12:17)

vincentfretin:
Probably a button is pressed before the model is loaded, we'll need to verify that `this.buttonRanges` exists before accessing it I guess
https://github.com/aframevr/aframe/blob/529eee87fc84dcaeca7eab7591826801b77145c1/src/components/oculus-touch-controls.js#L347
There is also a similar issue in #5212 
If anyone having a Quest 2 have one of those issues, those are really simple changes to make your first PR.

vincentfretin:
Oh I didn't see it was also you who reported the other one. Can you try making a PR?

dmarcos:
PRs def welcome

vincentfretin:
It should be fixed in #5223, if you can test @mikemainguy that would be awesome.

vincentfretin:
@dmarcos you can close this issue now that #5223 is merged.

dmarcos:
Thanks. Fixed by https://github.com/aframevr/aframe/pull/5223

What I usually do to close issues automatically is to add the issue to the commit message. "blah blah blah (fix #5220)" I'll point it out in future reviews.

vincentfretin:
[...]

PR #5223, opened 2023-01-21
Fix errors when buttons are pressed on Quest 2 controllers before the model is loaded

### LINKED ISSUE #5212: Uncaught TypeError: Cannot read properties of undefined (reading 'originalColor')
**Description:**

- A-Frame Version: 1.4.1
- Platform / Device: Oculus Quest 2
- Reproducible Code Snippet or URL:  https://dev.immersiveidea.com/local
- launch that site, go into immersive mode, and hit b button on controller

local:15 Uncaught TypeError: Cannot read properties of undefined (reading 'originalColor')
    at NewComponent.updateButtonModel (aframe-master.js:18543:1)
    at NewComponent.updateModel (aframe-master.js:18538:1)
    at module.exports.onButtonEvent (aframe-master.js:34703:1)
    at AEntity.onButtonTouchEnd (aframe-master.js:18275:1)
    at AEntity.nrWrapper (local:15:29524)
    at AEntity.emit (aframe-master.js:25830:1)
    at NewComponent.handleTouch (aframe-master.js:22514:1)
    at NewComponent.handleButton (aframe-master.js:22440:1)
    at NewComponent.updateButtons (aframe-master.js:22426:1)
    at NewComponent.tick (aframe-master.js:22385:1)

I don't observe functional change (button events still fire and my app works), I'm not sure if there's supposed to be other things happen (my buttons and thumbsticks are all black and hard to see...IDK if I've misconfigured something or something changed when we added the button color changes....which I really like BTW.

vincentfretin:
**Description:**
Fix errors when buttons are pressed on Quest 2 controllers before the model is loaded.
I don't have a Quest 2, please someone test the changes.
I tested on Quest 1, I didn't break it. :)

**Changes proposed:**
- Check if buttonObjects or buttonMeshes is defined before accessing it, this fixes Quest 2 buttons highlighting errors when the model is not loaded yet (fix #5212 and #5220)-

### DIFF
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -340,6 +340,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     var buttonObjects = this.buttonObjects;
     var analogValue;
 
+    if (!buttonObjects) { return; }
     analogValue = evt.detail.state.value;
     analogValue *= this.data.hand === 'left' ? -1 : 1;
 
@@ -491,11 +492,13 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
 
   updateButtonModel: function (buttonName, state) {
     // update the button mesh colors
-    var button;
-    var color = (state === 'up' || state === 'touchend') ? this.buttonMeshes[buttonName].originalColor || this.data.buttonColor : state === 'touchstart' ? this.data.buttonTouchColor : this.data.buttonHighlightColor;
     var buttonMeshes = this.buttonMeshes;
+    var button;
+    var color;
 
-    if (buttonMeshes && buttonMeshes[buttonName]) {
+    if (!buttonMeshes) { return; }
+    if (buttonMeshes[buttonName]) {
+      color = (state === 'up' || state === 'touchend') ? buttonMeshes[buttonName].originalColor || this.data.buttonColor : state === 'touchstart' ? this.data.buttonTouchColor : this.data.buttonHighlightColor;
       button = buttonMeshes[buttonName];
       button.material.color.set(color);
       this.rendererSystem.applyColorCorrection(button.material.color);
--- a/src/components/oculus-touch-controls.js
+++ b/src/components/oculus-touch-controls.js
@@ -309,25 +309,24 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
   },
 
   onButtonChanged: function (evt) {
+    var buttonMeshes = this.buttonMeshes;
     var button = this.mapping[this.data.hand].buttons[evt.detail.id];
     if (!button) { return; }
+    if (!buttonMeshes) { return; }
     // move the button meshes
     if (this.isOculusTouchV3) {
       this.onButtonChangedV3(evt);
     } else {
-      var buttonMeshes = this.buttonMeshes;
       var analogValue;
 
       if (button === 'trigger' || button === 'grip') { analogValue = evt.detail.state.value; }
 
-      if (buttonMeshes) {
-        if (button === 'trigger' && buttonMeshes.trigger) {
-          buttonMeshes.trigger.rotation.x = this.originalXRotationTrigger - analogValue * (Math.PI / 26);
-        }
-        if (button === 'grip' && buttonMeshes.grip) {
-          analogValue *= this.data.hand === 'left' ? -1 : 1;
-          buttonMeshes.grip.position.x = this.originalXPositionGrip + analogValue * 0.004;
-        }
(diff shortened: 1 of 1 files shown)


src/components/oculus-touch-controls.js
```diff
@@ -340,6 +340,7 @@ module.exports.Component = registerComponent('oculus-touch-controls', {
     var buttonObjects = this.buttonObjects;
     var analogValue;
 
+    if (!buttonObjects) { return; }
     analogValue = evt.detail.state.value;
     analogValue *= this.data.hand === 'left' ? -1 : 1;
```
dmarcos, line 346: Thinking this code block doesn't belong here (function should be model agnostic) and probably we should listen to a `triggermoved` listener event like here:

https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2L214

vincentfretin: I don't understand what you want to change here. `onButtonChangedV3` is called from `onButtonChanged` that is a listener of `buttonchanged` already. If you have a refactoring in mind, you should probably merge this PR first and do a refactoring in another PR?

dmarcos: This line https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R343 `if (!buttonObjects) { return; }` is needed because the model is accessed here https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R352 to update the trigger / grip. Those updates should be done in a different method than `clickButtons` that should remain model agnostic. 

My proposal is that this check should not be necessary https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2R343 and grip / trigger and button should be updated in a separate method similar to how we do with thumbstick in [onThumbStickMoved](https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2L472)

dmarcos: Alternatively move the logic to [updateButtonModel](https://github.com/aframevr/aframe/pull/5223/files#diff-2ee48b207a426377ea32a1749d345fd8c9b1adbe0b9d971eccb7860e1fb41cd2L492). Goal is to consolidate model updates in fewer places. Less redundant checks and easier to maintain.

vincentfretin: `clickButtons` is not a method, it's just an array, I think you meant `onButtonChangedV3` here?
`onThumbstickMoved` is for `thumbstickmoved` event
`onButtonChange`/`onButtonChangedV3` is for `buttonchanged` event.
`onThumbstickMoved` and `onButtonChange` have similar check for `buttonMeshes`.
You probably want to keep a separate listener for each event, right?
You want to remove `onButtonChangedV3` and put the code back to `onButtonChange` and slightly refactor it, not sure what can be refactored exactly though? Is that what you want?
Or simply do the `if (buttonMeshes)` check before calling `this.onButtonChangedV3(evt)` and remove it from `onButtonChangedV3`?. I'll do the latter that in a commit now.

`updateButtonModel` is called from `updateModel` which is called by `onButtonEvent`, used to define listeners `onButtonDown` `onButtonUp` `onButtonTouchStart` `onButtonTouchEnd`, but not for `onButtonChanged` currently, not sure what is particular about it currently and if we can define `onButtonChanged` to use `onButtonEvent` instead and move all the current logic of `onButtonChanged`/`onButtonChangedV3` to `updateButtonModel`. If this is what you want, I really prefer we do it in a separate PR, because it's not a trivial change, I don't want to break anything in this PR.

vincentfretin:
I tried to put the condition in onButtonChanged but I got a failing test
```
 oculus-touch-controls
    buttonchanged
      ✖ can emit triggerchanged
```
I thought it was a simple change though I don't understand why it failed. I reverted it.
Feel free to merge it as is or not. I'll try doing a refactor in two weeks maybe, not right now.

vincentfretin:
@dmarcos you should merge it really, it's a minimal set of changes to fix the errors of the two linked issues.
I don't know when I'll have the time to do a refactoring like you suggested, but I think it should be in a separate PR anyway.

dmarcos:
Super thanks!

PR #5226, opened 2023-01-26
Remove registerElement in doc

vincentfretin:
**Description:**

`AFRAME.registerElement` was removed in A-Frame 1.4.0 but the doc still list it.

**Changes proposed:**
- remove reference to registerElement


dmarcos:
Thanks

PR #5227, opened 2023-02-01
Webxr unit tests

diarmidmackenzie:
**Description:**

As noted in #5217, and one reason that bug #5221 slipped through - the Enter / Exit VR tests for a-scene are still entirely focussed on WebVR rather than WebXR.

This PR adds some tests for WebXR

**Changes proposed:**

I approached this as follows;
- Duplicate the existing WebVR tests
- Add additional stubbing code etc. to make them work
- Remove the tests that seem not to be relevant (some tests depend on events like `vrdisplaypointerrestricted` and `vrdisplaypresentchange` that seem to be WebVR only).

I doubt this gives us perfect coverage of WebXR, but it should give good coverage of a broad range of key scenarios around entry to / exit from VR.

I have also confirmed that this set of tests would have caught the bug fixed under #5221.

Some very minor changes to a-scene also made, to allow simulation of WebXR availability via the `el.hasWebXR` flag.

### DIFF
--- a/tests/__init.test.js
+++ b/tests/__init.test.js
@@ -37,9 +37,7 @@ setup(function () {
       getDevice: function () { return {requestPresent: function () {}}; },
       isPresenting: function () { return true; },
       setDevice: function () {},
-      setSession: function () {
-        return Promise.resolve();
-      },
+      setSession: function () { return Promise.resolve(); },
       setFoveation: function () {},
       setPoseTarget: function () {},
       dispose: function () {},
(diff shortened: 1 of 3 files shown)


tests/__init.test.js
```diff
@@ -37,9 +37,13 @@ setup(function () {
       getDevice: function () { return {requestPresent: function () {}}; },
       isPresenting: function () { return true; },
       setDevice: function () {},
+      setSession: function () {
```
dmarcos, line 40: can do one liner

diarmidmackenzie: fixed

dmarcos:
Thank you very much!

PR #5233, opened 2023-02-08
[docs] format markdown tables | rename Animation component to animation

foobar404:
**Description:**
The only ui change was renaming the Animation component to animation.


dmarcos:
Why do the tables need reformatting?

foobar404:
@dmarcos It was just an effort to make the tables easier to read in the markdown files, its hard to tell that in the pull request though.

dmarcos:
This makes code harder to maintain have to add remove dashes each time the length of a column changes. It doesn't make a difference when markdown is rendered.

dmarcos:
I also noticed there's a change in the titles markdown.

I can merge the component rename if you remove other changes or create a separate PR

PR #5243, opened 2023-02-14
Single property schema array

### LINKED ISSUE #5242: Single property schema where the property is an array does not work (broken since 1.0.0)
**Description:**

- A-Frame Version: 1.0+
- Platform / Device: All 
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/single-property-schema-array-141?path=index.html%3A8%3A23

Working version using A-Frame 0.8.0 here:
https://glitch.com/edit/#!/single-property-schema-array?path=index.html%3A8%3A23

Found when updating superframe emplate examples to 1.4.1.

Linked Glitch is a very simple example which uses a single property schema, where the property is an array, and outputs the data to screen via the 'text' property.

This works OK at 0.8.0.  From 1.0.0, including at 1.4.1, this no longer works.

[This branch](https://github.com/diarmidmackenzie/aframe/tree/single-property-schema-array) adds a Unit Test that exhibits the problem, but I haven't figured out a good fix for this yet.

Diagnosis of the problem so far:

The component does not get classified as `isObjectBased` (I don't know for sure whether that's correct or not) because this expression is not true:

```
isObject(parseProperty(undefined, this.schema))
```
[here](https://github.com/aframevr/aframe/blob/be856f835974d9da321b098c8f80aa5d3858d339/src/core/component.js#L45)

Then, in `updateCachedAttrValue`, `this.attrValue` gets set up correctly, but the array entries it points to have been copied in by reference, not value, and so get changed to `undefined` on the call to `clearObject`

The relevant code is [here](https://github.com/aframevr/aframe/blob/be856f835974d9da321b098c8f80aa5d3858d339/src/core/component.js#L229)

A puzzling earlier line of code is the use of `value instanceof Object` [here](https://github.com/aframevr/aframe/blob/be856f835974d9da321b098c8f80aa5d3858d339/src/core/component.js#L207), rather than using `isObject`.

For an array, `isObject` is false, but `value instanceof Object` is true, [because an Array is an instance of an Object.](https://stackoverflow.com/questions/26222604/why-is-array-instanceof-object)

diarmidmackenzie:
**Description:**

Small change to fix #5242 and allow an array to be used as the property in a single property schema.
Unit tests for this.

**Changes proposed:**

I don't have much confidence in this fix, so needs review by someone who understands the code better.  It may be that an altogether different approach is needed.

Following up on my previous analysis [here](https://github.com/aframevr/aframe/issues/5242#issue-1583933484), I experimented with various changes.

What seemed to work best was skipping the object pool altogether for this array property, by updating this test:
```
if ((value instanceof Object) && !(value instanceof window.HTMLElement))
```
to this:
```
if ((value instanceof Object) &&
        !(value instanceof Array) &&
        !(value instanceof window.HTMLElement))
```

I tried this simple option:
```
if (this.isObjectBased && !(value instanceof window.HTMLElement))
```
but that broke some cursor UTs, presumably because of they use Vector3s, for which it's also the case that `instanceof Object` is true, but `isObject` is false...  I guess we don't want to make any changes for Vector3s, just arrays...?

A build including this fix does resolve the problem.
https://glitch.com/edit/#!/single-property-schema-array-fix?path=index.html%3A8%3A23

There's still lots I don't understand about the code I have modified, exactly when it is invoked, and what it should do.

Things I don't understand:
- Why does the problem we have with Arrays not also happen for Vector3s?  Components like `position` and `rotation` use a single property schema, with a Vector3, and they clearly work fine.
- Why did we only have a problem on entity creation, and not on subsequent `setAttribute()` calls?

Regarding the new UTs...

I have reasonable confidence that the 2 x new UTs are valid, but not sure they belong in the `component` suite: although they are testing code in `src/core/component` they are more like end-to-end tests, rather than unit tests of this module.

Maybe there's a better place for these?

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -204,7 +204,9 @@ Component.prototype = {
       return;
     }
 
-    if (value instanceof Object && !(value instanceof window.HTMLElement)) {
+    if ((value instanceof Object) &&
+        !(value instanceof Array) &&
+        !(value instanceof window.HTMLElement)) {
       // If value is an object, copy it to our pooled newAttrValue object to use to update
       // the attrValue.
       tempObject = this.objectPool.use();
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -833,10 +833,23 @@ suite('Component', function () {
         schema: {list: {default: ['a']}}
       });
       var component = new TestComponent(this.el);
+      assert.deepEqual(component.data.list, ['a']);
       component.update = updateStub;
       component.updateProperties({list: ['b']});
       component.updateProperties({list: ['b']});
       sinon.assert.calledOnce(updateStub);
+      assert.deepEqual(component.data.list, ['b']);
+    });
+
+    test('supports array property in single property schema', function () {
+      registerComponent('dummy', {
+        schema: {
+          schema: {default: ['a']}
+        }
+      });
+      var el = document.createElement('a-entity');
+      el.setAttribute('dummy', ['b']);
+      assert.deepEqual(el.components.dummy.data, ['b']);
     });
 
     test('emit componentchanged when update calls setAttribute', function (done) {
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -841,15 +841,34 @@ suite('Component', function () {
       assert.deepEqual(component.data.list, ['b']);
     });
 
-    test('supports array property in single property schema', function () {
+    test('supports array property on entity creation', function (done) {
+      entityFactory();
       registerComponent('dummy', {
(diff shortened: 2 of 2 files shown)


dmarcos:
This needs rebase. I haven't stumbled much in the wild on arrays for single property schemas. 

@vincentfretin @mrxz thoughts?

vincentfretin:
It seems you can close that PR. The issue was fixed in #5500 that included those tests.

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5500

PR #5261, opened 2023-03-17
hand-controls: rotates models to match WebXR definition (fix #4848)

### LINKED ISSUE #4848: Hand tracking controls model needs update
The hand tracking WebXR API has changed on how it reports the joint pose and the [model needs to be updated](https://github.com/aframevr/assets/tree/master/controllers/oculus-hands/unity).

Information I collected from 3rd party

> The left hand joints were rotated around the X axis by 180 degrees and then around the Y axis by -90. The right hand joints were rotated around the Y axis by 90.

DougReeder:
**Description:**
Fixes #4848 
The WebXR standard now defines the grip space such that a cylinder held in a closed hand points along the Z axis.  The models currently have such a cylinder point along the X-Axis.

**Changes proposed:**
There are two approaches
1. rotate the mesh in code, as done here
2. edit the GLB files

I'm fine with either, but I lack knowledge of 3D tools (such as Blender) to implement the second approach.

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -206,7 +206,7 @@ module.exports.Component = registerComponent('hand-controls', {
         var handMaterial = mesh.children[1].material;
         handMaterial.color = new THREE.Color(handColor);
         mesh.position.set(0, 0, 0);
-        mesh.rotation.set(0, 0, handModelOrientation);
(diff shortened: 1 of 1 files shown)


DougReeder:
Documentation: https://developer.mozilla.org/en-US/docs/Web/API/WebXR_Device_API/Inputs#facing_and_targeting

dmarcos:
Thank you!

Issue #5293, opened 2023-05-23
Warning about physicallyCorrectLights that was renamed to useLegacyLights (three r152)


vincentfretin:
**Description:**

```
THREE.WebGLRenderer: the property .physicallyCorrectLights has been removed. Set renderer.useLegacyLights instead.
```

I commented about this in https://github.com/aframevr/aframe/pull/5210#issuecomment-1520194017

- A-Frame Version: master on three r152
- Platform / Device: Chrome desktop
- Reproducible Code Snippet or URL: https://aframe.io/aframe-inspector/examples/

dmarcos:
Thanks. PRs welcome

PR #5297, opened 2023-06-05
Set renderer.useLegacyLights instead of renderer.physicallyCorrectLights

mrxz:
**Description:**
Fixes https://github.com/aframevr/aframe/issues/5293

**Changes proposed:**
- Set `renderer.useLegacyLights` instead of the deprecated `renderer.physicallyCorrectLights`


dmarcos:
Thanks so much!

PR #5298, opened 2023-06-05
Use `fillPoses` and `fillJointRadii` instead of `getJointPose`

mrxz:
**Description:**
Changes the logic in `hand-tracking-controls` to use `fillPoses` and `fillJointRadii` instead of calling `getJointPose` many times over. The downside of `getJointPose` is that it allocates temporary objects, which quickly becomes inefficient.

As described in the [WebXR hand input explainer](https://github.com/immersive-web/webxr-hand-input/blob/main/explainer.md#efficiently-obtaining-hand-poses)
> Each use of `getPose()` allocates one short-lived `XRPose` object, one `XRRigidTransform` object, and at least one `DOMPointReadOnly` or `Float32Array`. For 25 joints per hand and two hands, this is 150-250 objects created per frame. This can have noticeable performance implications, especially around garbage collection.

Also removed the offset on the reference space. It didn't seem to serve any purpose and only required the rest of the code to apply a counter offset to get the actual position of the hand joints.

**Changes proposed:**
- No longer apply an offset to the reference space
- Per tick call `fillPoses` and `fillJointRadii` once
- Let `updateHandMeshModel`, `updateHandDotsModel` and `detectPinch` use these poses and radii.

### DIFF
--- a/src/components/hand-tracking-controls.js
+++ b/src/components/hand-tracking-controls.js
@@ -1,4 +1,4 @@
-/* global THREE, XRRigidTransform */
+/* global THREE */
 var registerComponent = require('../core/component').registerComponent;
 var bind = require('../utils/bind');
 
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks! I didn't know about fillPoses and fillJointRadii or didn't exist when I first wrote the code

PR #5303, opened 2023-06-09
Update "colorManagement" default in renderer documentation

donmccurdy:
Documentation is out of sync with the code – the default is true.


dmarcos:
Thank you!

PR #5308, opened 2023-06-14
ar-hit-test: uses 1st tracked controller instead of headset orientation (fix #5315)

### LINKED ISSUE #5315: ar-hit-test fails if controller button released before reticle is placed
**Description:**

- A-Frame Version: 1.4.2 and master
- Platform / Device: Any headset with AR capability and one or more controllers
- Reproducible Code Snippet or URL: https://aframe.io/examples/showcase/modelviewer/

1. click AR button to start AR mode
2. While ensuring the headset is never pointed toward a hit-test surface, press and release the select button (usually the trigger).

Expected result: reticle will still appear when headset is then pointed toward a hit-test surface
Actual result: reticle does **not** appear when headset is then pointed toward a hit-test surface

DougReeder:
**Description:**

When used with a headset capable of AR, the ar-hit-test component currently projects its ray from the headset, which is suboptimal UX when tracked controllers are available.

**Changes proposed:**

This change adds a new HitTest component, using the first tracked controller's `targetRaySpace`.

hit-testing has only been tested with a Quest 1 headset


DougReeder:
Also fixes bug #5315

DougReeder:
hit testing now also tested under Android 13 (on a Pixel 5)

hi-liang:
This indirectly addresses #5314 as well, the undefined reference error would now be in the `try` block 👍

gydence:
I can confirm that this fixes https://github.com/aframevr/aframe/issues/5315 for me on multiple Android phones.

dmarcos:
This needs rebase. Sorry for the wait. Should we use the controller that trigger the selectstart event vs the first we found?

dmarcos:
Thanks!

DougReeder:
I meant to add that the event is a controllerchange event, so it's not obvious what the relationship is to any selectstart event.  It might be possible to figure out which controller was involved, if it was triggered by a controller.

PR #5309, opened 2023-06-16
Add anisotropy property to renderer and material for configuring it globally and per material respectively

mrxz:
**Description:**
When viewing textures under a steep angle it can quickly become "blurry" and "greyish". In those cases anisotropic filtering can make a huge difference in visual quality. See for example the effect it has on the aframe-environment-component: https://github.com/supermedium/aframe-environment-component/pull/95

Meta [sets this on all textures for loaded models](https://github.com/meta-quest/ProjectFlowerbed/blob/6d617d008444e032264d175724a7b4da1ecbab59/src/js/systems/core/SceneCreationSystem.js#L153) in their Project Flowerbed and in [a blog post](https://developer.oculus.com/blog/common-rendering-mistakes-how-to-find-them-and-how-to-fix-them/) they summarize it as:
> On your environment textures: Turn on anisotropic filtering

Three.js provides two ways to configure this, one setting a global default and the other on a per-texture basis. This PR exposes both of them; on the renderer system and material component respectively. For the material it will apply to _all_ texture maps for that material, but in practice you'd want them in sync anyway. The code doesn't have to worry about unsupported values or [invalid combinations](https://github.com/mrdoob/three.js/pull/25068) as Three.js handles that already :-)

**Changes proposed:**
- Add an `anisotropy` property on the `renderer` system for setting `THREE.Texture.DEFAULT_ANISOTROPY`
- Add an `anisotropy` property on the `material` component

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -54,6 +54,7 @@ Here is an example of using an example custom material:
 
 [flat]: #flat
 [standard]: #standard
+[renderer]: ./renderer.md
 
 The material component has some base properties. More properties are available
 depending on the material type applied.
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "homepage": "https://aframe.io/",
   "main": "dist/aframe-master.js",
   "scripts": {
-    "dev": "cross-env INSPECTOR_VERSION=dev webpack serve --port 8080",
+    "dev": "cross-env INSPECTOR_VERSION=dev webpack serve --port 4000",
```
vincentfretin, line 8: I guess you committed that change by mistake.

mrxz: Oops.. :sweat_smile:

dmarcos:
Thank you so much!

PR #5310, opened 2023-06-16
Remove attributes array and cleanup Shader

mrxz:
**Description:**
The property combination `is: "attribute"` doesn't really do anything any more as attributes don't have to be manually specified for `(Raw)ShaderMaterial`. This PR removes it from the code and cleans up the Shader file and corresponding test cases.

**Changes proposed:**
- Remove `attributes` and special handling of `is: "attribute"` in shader.js

### DIFF
--- a/src/core/shader.js
+++ b/src/core/shader.js
@@ -50,10 +50,8 @@ Shader.prototype = {
    * Called during shader initialization and is only run once.
    */
   init: function (data) {
-    this.attributes = this.initVariables(data, 'attribute');
-    this.uniforms = this.initVariables(data, 'uniform');
+    this.uniforms = this.initUniforms();
     this.material = new (this.raw ? THREE.RawShaderMaterial : THREE.ShaderMaterial)({
-      // attributes: this.attributes,
       uniforms: this.uniforms,
       glslVersion: this.raw || this.glsl3 ? THREE.GLSL3 : null,
       vertexShader: this.vertexShader,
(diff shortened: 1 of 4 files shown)


src/core/shader.js
```diff
@@ -50,29 +50,27 @@ Shader.prototype = {
...
-  initVariables: function (data, type) {
+  initUniforms: function () {
     var key;
     var schema = this.schema;
     var variables = {};
     var varType;
 
     for (key in schema) {
-      if (schema[key].is !== type) { continue; }
+      if (schema[key].is !== 'uniform') { continue; }
```
nightgryphon, line 69: Should 'is' property be removed from schema and all shaders updated accordingly?

mrxz: I did contemplate that as well, but I think it's safer to leave it. It can also serve some purpose. Say for example you rename or deprecate a uniform on your shader. Now you can add a property to the schema under the old name, but not mark it as a uniform. By overloading the `update` method you can then print a warning message and delegate to the old behaviour without it ending up as a uniform.

hexiaokang (approved):

dmarcos:
Looking at this now. Nothing pending here?

mrxz:
@dmarcos Rebased on master, the `msdf` and `sdf` shaders started using `initVariables` (5409), so updated those.

dmarcos:
Thanks!

Issue #4279, opened 2019-10-02
Text not rendering when logarithmicDepthBuffer is enabled


Kershawj:
Creating a scene with just a <a-text> entity in but then using logarithmicDepthBuffer set to true in the renderer settings works fine so like:

`<a-scene renderer="logarithmicDepthBuffer: true;">
            <a-text value="Some text please" color="black" position="0 0 -5"></a-text>
        </a-scene>`

However adding a plane behind the text whilst logarithmicDepthBuffer set to true causes text to not render over the plane.

`<a-scene renderer="logarithmicDepthBuffer: true;">
            <a-plane height="1" width="3" color="red" position="0 0 -5.2"></a-plane>
            <a-text value="Some text please" color="black" position="0 0 -5"></a-text>
        </a-scene>`

Switching to `logarithmicDepthBuffer: false;` then works as expected.
I have a scene which really needs this set to true for lots of other entities to render correctly and to reduce z-fighting but also need some generated text in there so I'm kinda stuck.
I'm guessing the text shader isn't aware of how to handle this correctly?

- A-Frame Version: 0.9.2
- Platform / Device: Chrome/Macbook

Kershawj:
Just to say after more testing, it works as expected when using v0.8.0
`<script src="https://aframe.io/releases/0.8.0/aframe.min.js"></script>`

arpu:
does it work with the master version? 
``` <script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@ab9821761340fd6ee28171fc38f85a45c2ade514/dist/aframe-master.min.js"></script>```

Kershawj:
No, same result.
I do get some additional warnings in the console log, here's both as a comparison. First using 0.9.2:
<img width="432" alt="0 9 2_log" src="https://user-images.githubusercontent.com/9318275/66109385-6e616600-e5bc-11e9-87c1-def16c9c9ffd.png">

Then with master:
<img width="424" alt="master_log" src="https://user-images.githubusercontent.com/9318275/66109398-74574700-e5bc-11e9-82f6-9daa7c110d99.png">

arpu:
sorry i have no idea here best waiting for @dmarcos

dzoech:
Any updates on this issue?  
I'm experiencing the same issues with AR.js that text is not rendered in front of other entities independent from their z-indices.

[...]

Issue #4261, opened 2019-09-03
Scene fog not working on a-text


marcelpi:
**Description:**

Fog applies to most entities I've tested it on but not to a-text. Surface remains flat.

- A-Frame Version: 0.9.2
- Platform / Device: macOS High Sierra, Macbook Pro
- Reproducible Code Snippet or URL: http://www.pirosca.com/a-frame-example/fog/fog.html

dmarcos:
Probably the SDF shader doesn't take fog into consideration

marcelpi:
Any way this can be resolved in the current version?

arpu:
you can use https://github.com/lojjic/aframe-troika-text

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5409

PR #5328, opened 2023-07-06
FIX sdf/msdf shaders, add Shader.glslVersion

nightgryphon:
**Description:**
Fix shaders according to THREE changes. Fixes Text component.

**Changes proposed:**
- added `Shader.glslVersion`. Shaders should use this to specify version instead of shader `#version` string
- fixed sdf/msdf shaders used in Text component

### DIFF
--- a/src/core/shader.js
+++ b/src/core/shader.js
@@ -45,6 +45,8 @@ Shader.prototype = {
       'gl_FragColor = vec4(1.0, 0.0, 1.0, 1.0);' +
     '}',
 
+  glslVersion: null,
+
   /**
    * Init handler. Similar to attachedCallback.
    * Called during shader initialization and is only run once.
@@ -56,7 +58,8 @@ Shader.prototype = {
       // attributes: this.attributes,
       uniforms: this.uniforms,
       vertexShader: this.vertexShader,
-      fragmentShader: this.fragmentShader
+      fragmentShader: this.fragmentShader,
+      glslVersion: this.glslVersion
     });
     return this.material;
   },
--- a/src/shaders/msdf.js
+++ b/src/shaders/msdf.js
@@ -1,4 +1,5 @@
 var registerShader = require('../core/shader').registerShader;
+var THREE = require('../lib/three');
 
 var isWebGL2AVailable = !!document.createElement('canvas').getContext('webgl2');
 
@@ -15,7 +16,6 @@ var VERTEX_SHADER_WEBGL1 = [
 ].join('\n');
 
 var VERTEX_SHADER_WEBGL2 = [
-  '#version 300 es',
   'in vec2 uv;',
   'in vec3 position;',
   'uniform mat4 projectionMatrix;',
@@ -74,7 +74,6 @@ var FRAGMENT_SHADER_WEBGL1 = [
 ].join('\n');
 
 var FRAGMENT_SHADER_WEBGL2 = [
-  '#version 300 es',
   'precision highp float;',
   'uniform bool negate;',
   'uniform float alphaTest;',
@@ -135,5 +134,7 @@ module.exports.Shader = registerShader('msdf', {
 
   vertexShader: VERTEX_SHADER,
 
-  fragmentShader: FRAGMENT_SHADER
+  fragmentShader: FRAGMENT_SHADER,
+
+  glslVersion: isWebGL2AVailable ? THREE.GLSL3 : null
 });
--- a/src/shaders/sdf.js
+++ b/src/shaders/sdf.js
@@ -1,4 +1,5 @@
 var registerShader = require('../core/shader').registerShader;
+var THREE = require('../lib/three');
 
 var isWebGL2AVailable = !!document.createElement('canvas').getContext('webgl2');
 
@@ -15,7 +16,6 @@ var VERTEX_SHADER_WEBGL1 = [
 ].join('\n');
 
 var VERTEX_SHADER_WEBGL2 = [
-  '#version 300 es',
   'in vec2 uv;',
   'in vec3 position;',
   'uniform mat4 projectionMatrix;',
@@ -114,7 +114,6 @@ var FRAGMENT_SHADER_WEBGL1 = [
 ].join('\n');
 
 var FRAGMENT_SHADER_WEBGL2 = [
-  '#version 300 es',
   'precision highp float;',
   'uniform float alphaTest;',
   'uniform float opacity;',
@@ -213,5 +212,7 @@ module.exports.Shader = registerShader('sdf', {
 
   vertexShader: VERTEX_SHADER,
 
-  fragmentShader: FRAGMENT_SHADER
+  fragmentShader: FRAGMENT_SHADER,
+
+  glslVersion: isWebGL2AVailable ? THREE.GLSL3 : null
 });


src/core/shader.js
```diff
@@ -45,6 +45,8 @@ Shader.prototype = {
       'gl_FragColor = vec4(1.0, 0.0, 1.0, 1.0);' +
     '}',
 
+  glslVersion: null,
```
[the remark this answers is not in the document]
mrxz: Depends on if we want to support WebGL1 or not. (Though I'm not sure if A-Frame works with a WebGL1 context at the moment as I've never tried it)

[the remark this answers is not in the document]
dmarcos: This should not be part of the public API. Should check if the [WebGLRenderer is webgl1 or not](https://github.com/mrdoob/three.js/blob/95fbc39b07f677a3f550f185dd2d3d18d4a52c93/src/renderers/WebGL1Renderer.js#L5). Still I think is moot because as you said A-Frame doesn't offer a WebGL1 option

src/core/shader.js
```diff
@@ -56,7 +58,8 @@ Shader.prototype = {
       // attributes: this.attributes,
       uniforms: this.uniforms,
       vertexShader: this.vertexShader,
-      fragmentShader: this.fragmentShader
+      fragmentShader: this.fragmentShader,
+      glslVersion: this.glslVersion
```
dmarcos, line 62: probably passing here `THREE.GLSL3` is all the changes we need if I'm not missing anything. thoughts?

src/core/shader.js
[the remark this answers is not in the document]
nightgryphon: This definitely depends of decision to support WebGL1. If we drop WebGL1 support than sdf/msdf code should also be stripped/cleaned to remove GL1 shaders (which are differs from GL2 ones) and some startup checks for WebGL version has to be added to warn user of unsupported WebGL version.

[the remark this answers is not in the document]
nightgryphon: This THREE functionality looks a bit not clean as it is THREE who injects extra code in to shader so it is THREE job to inject code right way instead of forcing users to fill in the additional version parameter

src/core/shader.js
nightgryphon: Hardcoding version instead of making default value for parameter can be not the best solution. What if somewhere in future we will get GLSL4,5,6 or sort of custom extensions which require to pass different version string? For me the right way was is respect original user shader code but THREE did this...

src/core/shader.js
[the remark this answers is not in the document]
mrxz: > This should not be part of the public API. Should check if the [WebGLRenderer is webgl1 or not](https://github.com/mrdoob/three.js/blob/95fbc39b07f677a3f550f185dd2d3d18d4a52c93/src/renderers/WebGL1Renderer.js#L5).

It's a bit more subtle than that. The `WebGLRenderer` supports both WebGL 1 and WebGL 2 and will automatically pick one. The `WebGL1Renderer` class only exists to _force_ WebGL 1. In other words, just checking `.isWebGL1Renderer` isn't sufficient, as you can have a normal `WebGLRenderer` with a WebGL 1 context (in case your browser/device doesn't support WebGL 2).

What we're discussing however is the **glslVersion**. This is tied to the WebGL version in the sense that you can only use ES 3.0 shaders with WebGL 2, but you can use ES 1.0 shaders on both WebGL 1 and 2. There are differences in syntax so you can't simply tag on a `#version 300 es` on an ES 1.0 shader.

So if we hide this and default to `THREE.GLSL3` we're basically deciding two things at the same time:

- A-Frame will only supports WebGL2 (which it might already depend in other areas of the code?)
- A-Frame will only supports ES 3.0 shaders

Which I don't view as a problem, but is worth having clear in case someone wants to incorporate a (raw) ES 1.0 shader. And as @nightgryphon points out, if we do opt for this approach we should cleanup the sdf/msdf code and add an explicit WebGL2 check.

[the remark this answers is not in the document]
mrxz: > This THREE functionality looks a bit not clean as it is THREE who injects extra code in to shader so it is THREE job to inject code right way instead of forcing users to fill in the additional version parameter

Totally agree with you on this. I don't mind THREE.js injecting things into shaders, but it's surprising that they now also do this for `RawShaderMaterial`...

src/core/shader.js
mrxz: Hardcoding in this instance is not too bad. There's almost certainly never going to be a newer version for WebGL with WebGPU on the horizon.

It all boils down to two separate decisions (see other comment thread): do we want to support ES 1.0 GLSL shaders with A-Frame and do we want to support WebGL1 (if that even works right now?).

dmarcos: We can hardcode for now or can also determine from the renderer (WebGL1, WebGL2, WebGPU) instead of a public API. sounds good? thanks for the patience

mrxz: Did some testing and it seems A-Frame can (still) work fine with WebGL1, so I'd rather not break core components and try to postpone that till Three.js drops support (r163). Also found out that hardcoding it, actually breaks non-raw custom shaders as the GLSL 3.0 shader conversion that Three.js does ends up behaving slightly different :-/

Automatically determining it from the renderer won't be an option as the shader _is_ written in a specific shading language. So either we expose a way for users to tell which language/version they used, or we restrict what A-Frame supports.

Basically the following support matrix exists:
|         | WebGL 1.0 | WebGL 2.0 | WebGPU |
|-------|------------------|-----------------|--------------|
|ShaderMaterial (GLSL 1.0 /w 3.0 conversion) | ✅/❌ (2)| ✅ | ❌ |
| GLSL 1.0 | ✅ | ✅ | ❌ |
| GLSL 3.0 | ❌ | ⚠️ (1) | ❌ |
| NodeMaterial | ? | ✅ | ✅ |

 1. Before the Three.js update the shader code itself could indicate the GLSL version in its source. Now that three.js injects other things at the start, it _must_ also handle injecting the version directive. By extension this means that we _must_ tell three.js which version it is.
 2. Depends on which GLSL features and keywords are used. The GLSL 3.0 conversion in Three.js only handles a small subset.

### **tl;dr** I'd suggest the following:

- For now set `glslVersion: raw ? THREE.GLSL3 : undefined`, fixing the problem for the majority of the users.
- As a separate/follow-up: change the sdf and msdf shaders to non-raw shaders. This would remove the need to keep two almost identical copies around (GLSL1 and GLSL3 versions), restores WebGL1 compatibility, makes it easier to fix fog interaction (#4261), logarithmic depth (#4279) and color management.

dmarcos: let's go for the simple solution. thanks

> For now set glslVersion: raw ? THREE.GLSL3 : undefined, fixing the problem for the majority of the users.
> As a separate/follow-up: change the sdf and msdf shaders to non-raw shaders. This would remove the need to keep two almost identical copies around (GLSL1 and GLSL3 versions), restores WebGL1 compatibility, makes it easier to fix fog interaction (https://github.com/aframevr/aframe/issues/4261), logarithmic depth (https://github.com/aframevr/aframe/issues/4279) and color management.

dmarcos:
@nightgryphon can we incorporate these changes:

> For now set glslVersion: raw ? THREE.GLSL3 : undefined, fixing the problem for the majority of the users.
> As a separate/follow-up: change the sdf and msdf shaders to non-raw shaders. This would remove the need to keep two almost identical copies around (GLSL1 and GLSL3 versions), restores WebGL1 compatibility, makes it easier to fix fog interaction (Scene fog not working on a-text #4261), logarithmic depth (Text not rendering when logarithmicDepthBuffer is enabled #4279) and color management.

nightgryphon:
@dmarcos this will make all raw shaders treated as GLSL3. Didn't this cause backwards compatibility issues for anyone who uses raw shaders? 
For thouse with GLSL3 shaders they will get duplicate version string which will be clearly visible in error log so not so bad. 
For GLSL1... I'm not sure if GLSL1 shaders will compile with GLSL3 version string but if so than for GLSL1 shaders this will produce the unexpected shaders behavior without any error message which is awful

Regarding sdf/msdf shaders this will not fix them and text component. The minimum is to remove version string from GLSL3 shader code. But it is better also to drop GLSL1 shaders just to be sure using right shader code in any case to prevent hard to detect bugs.

As an alternative we can just extract version from shader fixing shader code at the same time. Should work for most cases. Updating sdf/msdf is not required and can be postponed.
```
    // THREE r154 require '#version' as standalone value in glslVersion
    var version = undefined;
    for (var name of ['vertexShader', 'fragmentShader'] ) {
        var Lines = this[name].split("\n");
        var pos = Lines[0].indexOf('#version');

        if (pos >= 0) {
            version = Lines.shift().substr(pos+9);
            this[name] = Lines.join("\n");
        }
    }

```
here is the commit i can make PR from https://github.com/aframevr/aframe/commit/a6d735c8acd6fb42af12eb1b799ea6b186697cff

mrxz:
@nightgryphon The idea is indeed both `glslVersion: raw ? THREE.GLSL3 : undefined` and removing the `#version 300 es` lines from the sdf/msdf shaders. Apologies for not making that clear.

This does indeed mean that third party raw shaders will require changes, though I think you'll find that not many people use _raw_ shaders in the first place. For GLSL3 shaders they will simply have to remove the `#version 300 es` line (which is a change people using THREE.js will need to make anyway). For GLSL1 shaders, they'll either have to rewrite them or override `init()` and construct their own `RawShaderMaterial`.

I'd rather not introduce workarounds or special behaviour into A-Frame for the purpose of retaining old Three.js behaviour. While it's annoying that many of the breaking changes in A-Frame come from (breaking) changes in Three.js, going against it would result in workarounds and fixes piling up over time. And given that A-Frame offers virtually no abstraction over the Three.js shaders I don't think it makes sense to maintain this old behaviour indefinitely.

nightgryphon:
Hm. Okay. At least that fixes the current problem with text.
So this comes to be like this
https://github.com/aframevr/aframe/compare/master...nightgryphon:aframe-local:FIX_Shaders3
Should i create PR ?

dmarcos:
@nightgryphon can you modify this PR?

nightgryphon:
A bit messy but done. Huh...

src/core/shader.js
```diff
@@ -50,11 +50,16 @@ Shader.prototype = {
    * Called during shader initialization and is only run once.
    */
   init: function (data) {
+    if (this.raw) {
```
dmarcos, line 53: Not sure this message is very valuable. Perhaps better making a note in the docs. I'm also happy to rip the band-aid and perhaps starting to say that WebGL1 is deprecated and WebGL2 is the way to go.

nightgryphon: This definitely has to be added to docs. But i think keeping such warning just for a few versions can save some headache to those who still using old shaders and has to migrate.

dmarcos: Not sure warning is super descriptive for the avg. dev. It is extra code to maintain and eventually have to remember to remove when obsolete.

What we usually do is to add `deprecations` in the release notes for those migrating https://github.com/aframevr/aframe/releases That's evergreen info.

mrxz: The warning isn't particular useful as it will always be printed. It might even confuse people seeing the warning when everything is correct. The fact that a big error is printed by Three.js down the line is probably enough. Anyone writing their own raw shaders will either understand the error or look it up and find the relevant Three.js and A-Frame issues. Those using older shaders from others will most likely open an issue about _that_ error message anyway.

Documenting this change and mentioning it in the release notes is key, as that will help anyone updating and can easily be linked/pointed to when people encounter this.

nightgryphon:
Ok. I've removed the warning

nightgryphon:
@mrxz Oh. That's makes things clear. Thanks!

dmarcos:
Thanks so much for the patience and being so accomodating!

Issue #5119, opened 2022-09-19
GDPR compliance for websites using AFrame


Olm-e:
**Description:**
In the current state of the core code, and in many examples, it is made call to github cdn or other websites for assets and eventually components. 
In Europe since the the publication of the GDPR rules and more recently the Schrems II EUCJ ruling, the transfer of a website users data to a third party and more precisely ones located outside EU (and particularly in the US or under US-based company control) without specific informed consent from the user is outlawed.
So for an unknowing person/organisation in Europe making use of the current distribution of Aframe without modification is liable and not "GDPR Compliant" because they cannot control the flow of data on their website for their users. 
This has _huge_ consequence many don't (want to) see actually, but the good news is Aframe is easily fixable with a few things :  

Distribute the /controller and /fonts directory from the assets rep inside a /asset directory in Aframe directly, and replace any occurrence of links to  https://cdn.aframe.io and copy the https://dpdb.webvr.rocks/dpdb.json in the same asset dir while also replacing it's https://dpdb.webvr.rocks link in aframe(.min).js by the same /asset/ dir ... 

this way Aframe is self contained, can work locally and does not leak user information to other possibly unwanted parties [edit] and also work without problem for people using "no-script" browser addons that restrict the use of external javascript ... 

- A-Frame Version: all using hardcoded external links to cdn/websites for assets 
- Platform / Device: all
- Reproducible Code Snippet or URL:

dmarcos:
@Utopiah has generously written down [instructions to make A-Frame work offline / self-hosted](https://github.com/aframevr/aframe/pull/5118/files#diff-a8cb57a43e2782b7d0b5dd92825a33b98824f5a1d77eafbdd75c685cea49c071R344)

Olm-e:
when you access an image, the server is still getting some info from the download (thus the use of 1 pixel images by some company to track users) so it is still a leak of personal info (IP is one, like your street, address, phone number else, even if multiple person could use the same)
Bundling the controllers and fonts is indeed a weight but not much : as it will not be used so often, the bw cost is linear to users use, and only limited to the webserver, as the users has always to dl them ... 
The ideas is assets are not downloaded through cdn but together with aframe are hosted on the website directly or it's controlled subsidiaries (and not hardcoded with uncontrollable sources)
I'm not using npm generaly and have hand made my own distro as in the issue comment above, maybe will try a better how-to description ... ;)

dmarcos:
[...]

PR #5333, opened 2023-07-09
Use global variable for CDN root directory

diarmidmackenzie:
**Description:**

As per #5119 

**Changes proposed:**

Replace hardcoded instances of `https://cdn.aframe.io/` with a global variable AFRAME_ROOT_CDN.

Based on @vincentfretin's suggestion [here](https://github.com/aframevr/aframe/issues/5119#issuecomment-1382817113)

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -2,5 +2,6 @@ module.exports = {
   AFRAME_INJECTED: 'aframe-injected',
   DEFAULT_CAMERA_HEIGHT: 1.6,
   DEFAULT_HANDEDNESS: 'right',
-  keyboardevent: require('./keyboardevent')
+  keyboardevent: require('./keyboardevent'),
+  AFRAME_CDN_ROOT: window.AFRAME_CDN_ROOT || 'https://cdn.aframe.io/'
 };
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -1,7 +1,7 @@
 module.exports = {
+  AFRAME_CDN_ROOT: window.AFRAME_CDN_ROOT || 'https://cdn.aframe.io/',
   AFRAME_INJECTED: 'aframe-injected',
   DEFAULT_CAMERA_HEIGHT: 1.6,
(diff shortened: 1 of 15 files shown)


src/constants/index.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   AFRAME_INJECTED: 'aframe-injected',
   DEFAULT_CAMERA_HEIGHT: 1.6,
   DEFAULT_HANDEDNESS: 'right',
-  keyboardevent: require('./keyboardevent')
+  keyboardevent: require('./keyboardevent'),
+  AFRAME_CDN_ROOT: window.AFRAME_CDN_ROOT || 'https://cdn.aframe.io/'
```
vincentfretin, line 6: This list seems to be sorted, you should put the new constant first, before AFRAME_INJECTED.

vincentfretin (commented):
Thanks for doing this work.

src/constants/index.js
diarmidmackenzie: Yes, well spotted.  Now alphabetized.

diarmidmackenzie:
> The text in https://github.com/aframevr/aframe/blob/master/docs/introduction/faq.md#can-i-use-a-frame-offline-or-self-hosted would also need to be modified.

I'd like to handle that as a separate PR.

I don't think the text as it stands *needs* to change for this PR to be OK - everything it says is still true, even after this code change.

More work needed to fully solve #5119, but I'd like to tackle this one bit at a time.

nightgryphon:
it also be cool to add an assets archive for loacal deployment. Including fonts, controllers models etc. I has to manualy figure out which files are required for local deployment and download one by one...

nightgryphon:
@vincentfretin Ohh! I've missed that repo. It can be a good idea to add a comment line with repo url near the constant and later in the docs. Also it is good to note which folders are actually required.

dmarcos:
Thanks!

PR #5335, opened 2023-07-12
Remove Chrome bug workaround from a-asset-item as it's no longer needed

mrxz:
**Description:**
The referenced Chrome bug has long been fixed in a patch version of Chrome 53. Affected browser usage is <0.01% based on https://caniuse.com/usage-table 

**Changes proposed:**
- Remove Chrome bug workaround that is no longer needed


dmarcos:
Thank you!

PR #5340, opened 2023-07-19
A more robust fix that doesn't depend on hardcoded indices

diarmidmackenzie:
**Description:**

Background here:
https://github.com/aframevr/aframe/commit/701477decf6dedef341a87b4c0d94bd7d8b97032

I thought this was a bug, so I prespared a fix.  Looks like this was actually due to an issue with my test environment.  Neverthless, I think this version of the code is more robust and should be more resilient to future changes in how GLB models get loaded.

**Changes proposed:**

Having hardcoded assumptions about the order that Bones vs. SkinnedMesh will appear in the three.js object tree makes the code vulnerable to changes in future.  This version of the code doesn't make such assumptions, and so should be more robust in future.

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -204,7 +204,7 @@ module.exports.Component = registerComponent('hand-controls', {
         mesh.mixer = new THREE.AnimationMixer(mesh);
         self.clips = gltf.animations;
         el.setObject3D('mesh', mesh);
-        mesh.traverse((o) => {
+        mesh.traverse(function (o) {
           if (o.isMesh) {
             o.material.color = new THREE.Color(handColor);
           }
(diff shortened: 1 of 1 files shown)


vincentfretin (commented):
@dmarcos doesn't want arrow function in the code base :p

diarmidmackenzie:
ofc, sorry.

Frustrating that we don't lint for this.

src/components/hand-controls.js
```diff
@@ -204,9 +204,11 @@ module.exports.Component = registerComponent('hand-controls', {
         mesh.mixer = new THREE.AnimationMixer(mesh);
         self.clips = gltf.animations;
         el.setObject3D('mesh', mesh);
-
-        var handMaterial = mesh.children[0].material;
-        handMaterial.color = new THREE.Color(handColor);
+        mesh.traverse(function (o) {
```
dmarcos, line 207: for consistency. no single letter and choose descriptive variable names.

diarmidmackenzie: upadted to `object` which is consistent with uses of `traverse()` elsewhere in the code.

dmarcos:
Thanks so much!

Issue #5332, opened 2023-07-08
A-Frame sort order flexibillity can be improved


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.4.2
- Platform / Device: All
- Reproducible Code Snippet or URL: None

I've been doing a deep dive into A-Frame / Three.js sortObjects, prompted by some difficulties I had getting a ["hider material"](https://github.com/8thwall/web/blob/57185d7b811e57f4b6fb5cc33429c55904f7301b/xrextras/src/aframe/xr-components.js#L986) to work for a Mixed Reality scene I was working on.

A-Frame sortObjects is described [here](https://aframe.io/docs/1.4.0/components/renderer.html#sortobjects) as follows:

> Sorting is used to attempt to properly render objects that have some degree of transparency. Due to various limitations, proper transparency often requires some amount of careful setup. By default, objects are not sorted, and the order of elements in the DOM determines order of rendering. Re-ordering DOM elements provides one way of forcing a consistent behavior, whereas use of renderer="sortObjects: true" may cause unwanted changes as the camera moves.

However this description fails to mention some important details about how the THREE.js sortObjects option actually works.

When `sortObjects` is enabled, sorting of objets happens according to the following logic (see code [here](https://github.com/mrdoob/three.js/blob/3e0a5739d948597f29af47597c034402ceddf125/src/renderers/webgl/WebGLRenderLists.js#L1))

- 1st factor is any [`renderOrder`](https://threejs.org/docs/index.html?q=object3D#api/en/core/Object3D.renderOrder) value set on any Group the Object3D is a member of.
- 2nd factor is any [`renderOrder`](https://threejs.org/docs/index.html?q=object3D#api/en/core/Object3D.renderOrder) value set on the Object3D itself
- 3rd factor is the id of the Object3D's material (this is done for performance reasons - see this [checkin](https://github.com/mrdoob/three.js/commit/140dba7529ace34965f688e5118e334da99e4513))
- 4th factor (only if ordering can't be decided by any of the above) is z-distance of the object from the camera (the object's center, of course, which may or may not be useful - e.g it's very much not useful for a sky...)

The A-Frame sortObjects option + docs were added under #3424, following discussion under #666.

#666 includes various reports of the render order changing as the camera moves around the scene (e.g. [this](https://github.com/aframevr/aframe/issues/666#issuecomment-348024713)) - which explains the warning about sortOrder that was added to the docs.

[...]

Issue #5218, opened 2023-01-14
Update linting to enforce ES5


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.4.1
- Platform / Device: All
- Reproducible Code Snippet or URL: N/A

The project uses [semistandard](https://github.com/standard/semistandard) for linting.

However the project also aims to use ES5 only, and does not permit ES6 features like const/let, arrow functions etc.  Having rules for code submissions that are not checked automatically (when they could be) results in time wasted & unnecessary iterations on PRs.

Proposal
- move from [semistandard](https://github.com/standard/semistandard) to [standardx](https://github.com/standard/standardx), which allows overriding of specific standard rules
- update config to disallow JS features after 2015.
(based on standard's [eslint config](https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json) this probably just requires setting 
```
"parserOptions": {
    "ecmaVersion": 2015
   }
```
and
```
  "env": {
    "es2015": true
    } 
```
- fix up any pre-existing ES2016 violations (policing has been manual, so there may be some instances that slipped through).

Alternative Proposal:
- Change policy to allow modern JS in A-Frame.  However I believe that's been suggested before and ruled out...

vincentfretin:
Recent comments on that was https://github.com/aframevr/aframe/pull/5103#issuecomment-1230841914 and https://github.com/aframevr/aframe/issues/5106#issuecomment-1233354515
I don't know if those learning javascript today knows about the different behavior of var hoisting compared to let/const:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#variable_hoisting
Anyway that was already been discussed.
FYI three.js repository uses javascript features up to ES2018 and removed babel from their build tool recently.

Current aframe code base uses ES6/ES2015 module only in `src/lib/three.module.js` to be able to import from jsm folder from three package, otherwise all the code is using `require`, no template literal, no `const/let` in `src` folder (except one let in `src/components/oculus-touch-controls.js` that slipped through in a recent PR and `src/components/layer.js`), in `tests` we use `var/const/let`.

diarmidmackenzie:
Hit this again working on #5340 

I might look at a PR here, as getting as many of the project coding conventions established as linting rules would cut reviewing cycles for PRs.

[...]

PR #5341, opened 2023-07-20
Improve sort order

diarmidmackenzie:
**Description:**

A PR to implement the changes proposed in #5332

**Changes proposed:**

- Set sortObjects to true in all cases
- Override THREE.js default sort functions with A-Frame sort functions
- replace `renderer` `sortObjects` property with `sortTransparentObjects` property that sets a different A-Frame sort function for transparent objects
- documentation and Unit Tests.


docs/components/renderer.md
```diff
@@ -82,13 +82,20 @@ Controls the amount of foveation which renders fewer pixels near the edges of th
...
 
-### sortObjects
+### sortTransparentObjects
 
-Sorting is used to attempt to properly render objects that have some degree of transparency.
-Due to various limitations, proper transparency often requires some amount of careful setup.
-By default, objects are not sorted, and the order of elements in the DOM determines order of
-rendering. Re-ordering DOM elements provides one way of forcing a consistent behavior, whereas
-use of `renderer="sortObjects: true"` may cause unwanted changes as the camera moves.
+In may cases, the order in which objects is rendered doesn't matter much, but there are a few cases where sorting is important:
```
diarmidmackenzie, line 87: This is documentation of general A-Frame object sorting behaviour, rather than this particular property.

Not sure if there's a better place for this in the docs, but I couldn't see an obvious one, and didn't want to restructure docs too much.

src/core/scene/sortFunctions.js
```diff
@@ -0,0 +1,56 @@
+// Custom A-Frame sort functions.
```
diarmidmackenzie, line 1: I think putting these into a new module is right.  Not sure whether this is the right place in the directory structure, or if there's a more suitable place.

I didn't think these were `utils` as they aren't part of the exposed A-Frame API.  COuldn't see another obvious directory to put them in, so put them here.

tests/systems/renderer.test.js
```diff
@@ -41,11 +41,40 @@ suite('renderer', function () {
     document.body.appendChild(sceneEl);
   });
 
-  test('change renderer sortObjects', function (done) {
+  test('change renderer sortTransparentObjects', function (done) {
     var sceneEl = createScene();
-    sceneEl.setAttribute('renderer', 'sortObjects: true;');
+
```
diarmidmackenzie, line 46: The fact that Unit Tests mock out the `setupRenderer` function means that I was limited in how much of the code changes I could Unit Test in a straightforward way.

Should we be looking for a way to improve UT coverage of the `setupRenderer` function?

tests/__init.test.js
```diff
@@ -46,10 +46,12 @@ setup(function () {
     },
     dispose: function () {},
```
diarmidmackenzie, line 47: reordered this list to be alphabetic

dmarcos:
@mrxz What are your thoughts? Thanks!

tests/systems/renderer.test.js
mrxz: It seems that the configuration of the renderer is a bit split between both `setupRenderer` and the `init` function of the renderer system. It might be easier to move the setting of `sortObjects` and sorting functions into the `init` function as well. That way the tests here can:

- Verify that the default values are set properly (currently the test `'default initialization'` asserts that sortObjects defaults to false and doesn't fail) 
- Verify that changing the `sortTransparentObjects` result in the correct sorting functions being set

This unit test now also tests that the sorting _behaviour_ is correct, but I would split that off into a new `sortFunctions.test.js` that specifically tests them and let this unit test here cover only the responsibility of the renderer system of setting the right sort functions.

Finding a way to improve test coverage of `setupRenderer` would still be nice, but I think with the above the need for it in the scope of this change is mostly alleviated

docs/components/renderer.md
mrxz: Small typo: may -> many

The text seems a bit out of place compared to the rest as it's considerably more in-depth and technical. While I personally don't mind it too much, it's usefulness to new(er) users is probably limited as a result. Surprisingly the original text still mostly holds (as it already pertained to transparent objects). Touching that up and linking to a different page or resource describing it more in-depth (preferably with concrete pros and cons) would IMHO be ideal.

src/systems/renderer.js
```diff
@@ -1,6 +1,7 @@
 var registerSystem = require('../core/system').registerSystem;
 var utils = require('../utils/');
 var THREE = require('../lib/three');
+var {aframeSortTransparentDefault, aframeSortTransparentSpatial} = require('../core/scene/sortFunctions');
```
mrxz, line 4: I believe destructuring is an ES6 feature.

src/core/scene/sortFunctions.js
```diff
@@ -0,0 +1,56 @@
...
+    return b.z - a.z;
+  } else {
+    return 0;
+  }
+}
+
+module.exports = {
+  aframeSortOpaqueDefault,
+  aframeSortTransparentDefault,
+  aframeSortTransparentSpatial
```
mrxz, line 55: These shorthands are also an ES6 features, I'm afraid...
(Why are we torturing ourselves with this ES5-only policy, again? :sweat: )

mrxz:
Looking good, tested it on a couple of scenes and works as expected. Left a couple of comments.

docs/components/renderer.md
diarmidmackenzie: What you suggest is the approach I wanted to take, but I couldn't find a good place in the docs to slip in a section on render order.  It's too detailed for me to want to put it into the core docs sections that work as a guide for newcomers.

Have looked again and I think the FAQ https://aframe.io/docs/1.4.0/introduction/faq.html is somewhere that could probably accommodate something.

Probably a question "What order does A-Frame render objects in?" though it could also be "Why are my transparent objects blocking other objects?".

Sound OK?

mrxz: Sounds okay to me, I would go for the "Why are my transparent objects blocking other objects?" or something along those lines. It's closer to the question people likely have when facing the issue, especially if they don't know it has to do with sorting. (And those that do know it has to do with sorting will likely find the `sortTransparentObjects` property first which links to the relevant FAQ question).

tests/systems/renderer.test.js
diarmidmackenzie: Pushed some changes to address these comments...

> It might be easier to move the setting of sortObjects and sorting functions into the init function as well.

The renderer `init` function will only be called if `renderer` system is initialized.  It does appear that's the case today, but I'm not sure it's good to bake that assumption into core code.  So I'd argue these properties should be set by the `AScene` class, and `setupRenderer()` seems like the right place to do it.

With the updated code, `renderer.sortObjects` is always `true`, and the `renderer` component has no bearing on it at all, so I'd argue it is cleaner to not have the `renderer` code touch `renderer.sortObjects`.

Sorting functions are set in the renderer `update` function, as well as in `setupRenderer()`.   No reason to move them to `init`, which just restricts them from being changed later.

> currently the test 'default initialization' asserts that sortObjects defaults to false and doesn't fail

Removed that check from the test.  `renderer.sortObjects` no longer has anything to do with the `renderer` component, so that check doesn't belong here any more.

> This unit test now also tests that the sorting behaviour is correct, but I would split that off into a new sortFunctions.test.js that specifically tests them and let this unit test here cover only the responsibility of the renderer system of setting the right sort functions.

That's fair.  Since the since the sort functions are internal to A-frame, I hadn't twigged that I can just import them, and then compare the functions directly.  Have now done that and it works fine.

A full UTs of the sort functions is also a sensible idea - I have added.

src/core/scene/sortFunctions.js
diarmidmackenzie: I'll fix.

FYI, I have been working on a proposal to add linting rules to enforce ES5:
https://github.com/aframevr/aframe/issues/5218

Blocked on the fact that A-Frame now uses `class` in several places, which is an ES6 feature.

tests/systems/renderer.test.js
mrxz: > The renderer init function will only be called if renderer system is initialized. It does appear that's the case today, but I'm not sure it's good to bake that assumption into core code. So I'd argue these properties should be set by the AScene class, and setupRenderer() seems like the right place to do it.

I kind of see it the other way around. It's the AScene class that is effectively overstepping its boundaries by depending on both the `camera` and `renderer` systems. Conceptually in an ECS if the renderer system would be left out, you'd be left with a headless instance. From a pragmatic and legacy perspective I can see how this situation came to be and it's now unfeasible to every change (lots of code depends on the scene holding the renderer and camera).

That does leave us in the situation where renderer initialization is effectively split in two parts `AScene.setupRenderer` and `renderer.init`. The logical split I see at the moment is that `setupRenderer` only sets the properties needed at `WebGLRenderer` construction time and `renderer.init` sets anything that can change 'on the fly'. The object sorting is the only odd one out (even before this PR).

That being said, the actual `sortObjects` property can be left default (= true) so neither needs to assign anything. Unless we explicitly want to assert this default (and potentially shield us against three.js ever changing that default).

src/systems/renderer.js
mrxz: Calling `require` more than once on the same module gives the same result, but I think it's a tad cleaner to assign it to an intermediate, i.e. `var sortFunctions = require('../core/scene/sortFunctions')`

docs/components/renderer.md
diarmidmackenzie: I've left it as "what order..." - there are other ordering issues that are nothing to do with transparency (e.g. the hider material issues that got me into looking at this in the 1st place, and a bunch of the content has nothing to do with transparency).

tests/systems/renderer.test.js
diarmidmackenzie: Not sure I completely agree, as most of the `renderer` settings can't be changed on the fly (according to the docs, at least).

https://aframe.io/docs/1.4.0/components/renderer.html

But happy this is the right direction of travel, so I made the changes you suggested.

mrxz: Ah, you're right, seems I overlooked the few that are set by renderer but not changeable on the fly. The docs and the schema do paint a clear picture that all these settings do belong to the renderer system, yet a good chunk of them are handled (parsed even) by the `AScene`, which in an ideal setup wouldn't have been the case, but alas.

Anyway, the PR looks good. Just two more ES6 destructuring assignments in [renderer.test.js](https://github.com/aframevr/aframe/pull/5341/files#diff-4f5dc6eb19ddb970dd8594098448beffadc70d68317a24d00051a0d7682946b8R2-R4) and [sortFunctions.test.js](https://github.com/aframevr/aframe/pull/5341/files#diff-3029e71ddd0f460cba008f65476fe6b42a4ef168c7b1649b793b64f2e422d7edR3) and then it's all good to go IMO.

diarmidmackenzie: My understanding was that the ES5 restriction only applies to /src.

Plenty of existing code under /test uses let, const, arrow functions etc. 

See e.g. [animation.test.js](https://github.com/aframevr/aframe/blob/master/tests/components/animation.test.js)

mrxz: Well, then everything is good to go IMO :-)

src/core/scene/sortFunctions.js
dmarcos: I would put them in the renderer file where they are used. see `setupCanvas` function in a-scene https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L863 No need to prepend `aframe` to the function name

dmarcos: We can think of modularizing later if the functions turn to be useful somewhere else (like those in utils)

diarmidmackenzie: Are you sure about this?  They are used in both a-scene, and renderer.  Defining them in renderer means that a core module (a-scene) will have a dependency on a systems module (renderer).

That seems like the wrong direction for a dependency to me.

It might be preferable to define them in a-scene?

Agree with removing `aframe` from the function names, and have made that change.

dmarcos: Sorry missed they're used in a-scene as well. Do we need them in both places? Can we consolidate in just one?

src/core/scene/sortFunctions.js
```diff
@@ -0,0 +1,56 @@
...
+// Variations of Three.js default sort orders here:
+// https://github.com/mrdoob/three.js/blob/ebbaecf9acacf259ea9abdcba7b6fb25cfcea2ab/src/renderers/webgl/WebGLRenderLists.js#L1
+// See: https://github.com/aframevr/aframe/issues/5332
+
+// Default sort for opaque objects:
+// - respect groupOrder & renderOrder settings
+// - sort front-to-back by z-depth from camera (this should minimize overdraw)
+// - otherwise leave objects in default order (object tree order)
+
+function sortOpaqueDefault (a, b) {
```
dmarcos, line 11: Function names are a bit cryptic. `sortOpaqueFirst` maybe? Default has not meaning without more context

src/core/scene/sortFunctions.js
dmarcos: I don't see the functions used in a-scene. I might be missing something?

diarmidmackenzie: Sorry, my mistake. They were used in a-scene until I did some refactoring proposed by @mrxz .  I'll fix as you suggest now.

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+    return a.renderOrder - b.renderOrder;
+  } else if (a.z !== b.z) {
+    return b.z - a.z;
+  } else {
+    return 0;
+  }
+}
+
+// exports needed for Unit Tests
+module.exports.sortOpaqueDefault = sortOpaqueDefault;
```
dmarcos, line 152: I appreciate the tests. I'll merge but in general I lean towards just unit testing public APIs. Internal functions are implementation details and tests add friction to modify, refactor, improve the code.

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+// Variations of Three.js default sort orders here:
+// https://github.com/mrdoob/three.js/blob/ebbaecf9acacf259ea9abdcba7b6fb25cfcea2ab/src/renderers/webgl/WebGLRenderLists.js#L1
+// See: https://github.com/aframevr/aframe/issues/5332
+
+// Default sort for opaque objects:
+// - respect groupOrder & renderOrder settings
+// - sort front-to-back by z-depth from camera (this should minimize overdraw)
+// - otherwise leave objects in default order (object tree order)
+
+function sortOpaqueDefault (a, b) {
```
dmarcos, line 110: Can we find a bit more descriptive names for these functions? I had suggested `sortOpaqueObjectsFirst` but not sure if good enough. can suggest other names.

dmarcos: If we can make this function names a bit more descriptive this is almost ready to merge. Thanks!

diarmidmackenzie: Sorry for slow response on this.

I just pushed a change that updated:
`sortOpaqueDefault` -> `sortSpatialFrontToBack`
`sortTransparentDefault` -> `sortRenderOrderOnly`
`sortTransparentSpatial` -> `sortSpatialBackToFront`

This describes what the functions implement rather than what they are used for (glossing over the detail that the spatial sorts also sort by `renderOrder` first, in the interest of keeping the function names at a sensible length).

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+// Variations of Three.js default sort orders here:
+// https://github.com/mrdoob/three.js/blob/ebbaecf9acacf259ea9abdcba7b6fb25cfcea2ab/src/renderers/webgl/WebGLRenderLists.js#L1
+// See: https://github.com/aframevr/aframe/issues/5332
+
+// Default sort for opaque objects:
+// - respect groupOrder & renderOrder settings
+// - sort front-to-back by z-depth from camera (this should minimize overdraw)
+// - otherwise leave objects in default order (object tree order)
+
+function sortSpatialFrontToBack (a, b) {
```
dmarcos, line 110: What does spatial means in this context? Maybe redundant? `sortFrontToBack` maybe simpler?

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+  if (a.groupOrder !== b.groupOrder) {
+    return a.groupOrder - b.groupOrder;
+  } else if (a.renderOrder !== b.renderOrder) {
+    return a.renderOrder - b.renderOrder;
+  } else if (a.z !== b.z) {
+    return a.z - b.z;
+  } else {
+    return 0;
```
dmarcos, line 118: I think this statement might be redundant. If one does `a.z - b.z` if they are equal they will return 0 as well.

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+// - otherwise leave objects in default order (object tree order)
+function sortSpatialBackToFront (a, b) {
+  if (a.groupOrder !== b.groupOrder) {
+    return a.groupOrder - b.groupOrder;
+  } else if (a.renderOrder !== b.renderOrder) {
+    return a.renderOrder - b.renderOrder;
+  } else if (a.z !== b.z) {
+    return b.z - a.z;
+  } else {
+    return 0;
```
dmarcos, line 147: similarly:
```
return (a.groupOrder - b.groupOrder) || (a.renderOrder - b.renderOrder) || (b.z - a.z);
```

src/systems/renderer.js
dmarcos: similarly:

```
return (a.groupOrder - b.groupOrder) || (a.renderOrder - b.renderOrder) || (a.z - b.z);

```

mrxz: Wouldn't it make sense to retain the ID tie-breaker (`return a.id - b.id`) used by Three.js as well? That should ensure that, everything else being equal, the order is consistent.

As for compacting it to a single expression, I personally prefer the current more verbose version. It's easier to see what's going on and it resembles the referenced original Three.js code, making it easy to compare. I did experiment with some very crude benchmarking and while both perform about the same, a significant difference forms once some of the properties are (mostly) consistent. It seems the if-else chain allows for better branch-prediction:

- [0 groupOrder, 1% renderOrder, random Z](https://perf.link/#eyJpZCI6Im04NHh5amJxOXE2IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGFycmF5ID0gbmV3IEFycmF5KDEwMDAwKTtcbmZvcihsZXQgaSA9IDA7IGkgPCBhcnJheS5sZW5ndGg7IGkrKykge1xuXHRhcnJheVtpXSA9IHtcblx0XHRncm91cE9yZGVyOiAwLFxuXHRcdHJlbmRlck9yZGVyOiBNYXRoLnJhbmRvbSgpIDwgMC4wMSA%2FIE1hdGguZmxvb3IoTWF0aC5yYW5kb20oKSoxMDApIDogMCxcblx0XHR6OiBNYXRoLnJhbmRvbSgpICogMTAwMDAwIC0gNTAwMDAsXG5cdH07XG59IiwidGVzdHMiOlt7Im5hbWUiOiJGaW5kIGl0ZW0gMTAwIiwiY29kZSI6ImFycmF5LnNvcnQoKGEsYikgPT4ge1xuXHRpZihhLmdyb3VwT3JkZXIgIT09IGIuZ3JvdXBPcmRlcikge1xuXHRcdHJldHVybiBhLmdyb3VwT3JkZXIgLSBiLmdyb3VwT3JkZXI7XG5cdH0gZWxzZSBpZihhLnJlbmRlck9yZGVyICE9PSBiLmdyb3VwT3JkZXIpIHtcblx0XHRyZXR1cm4gYS5yZW5kZXJPcmRlciAtIGIucmVuZGVyT3JkZXI7XG5cdH0gZWxzZSB7XG5cdFx0cmV0dXJuIGEueiAtIGIuejtcblx0fVxufSkiLCJydW5zIjpbMjI1MCwyNjI1LDIwMDAsMTYyNSwyNzUwLDIyNTAsMjI1MCwyMTI1LDIxMjUsMjAwMCwyMzc1LDIwMDAsMjYyNSwyMjUwLDE1MDAsMjUwMCwyMjUwLDE4NzUsMjEyNSwxODc1LDIxMjUsMjI1MCwxODc1LDE3NTAsMjAwMCwyMTI1LDIzNzUsMjc1MCwxODc1LDE4NzUsMTUwMCwxODc1LDE3NTAsMTc1MCwyNzUwLDIyNTAsMTc1MCwyMTI1LDE4NzUsMTg3NSwxODc1LDIwMDAsMTg3NSwyMTI1LDIyNTAsMjAwMCwyNTAwLDIyNTAsMjM3NSwyMDAwLDI1MDAsMTg3NSwyMDAwLDE3NTAsMjI1MCwyMTI1LDIyNTAsMjEyNSwyMDAwLDIxMjUsMjAwMCwyMjUwLDIyNTAsMjEyNSw1MDAsMTc1MCwxNjI1LDI1MDAsMjI1MCwyMjUwLDIwMDAsMjM3NSwyMzc1LDIyNTAsMjUwMCwyMTI1LDE3NTAsMjYyNSwyMzc1LDE4NzUsMTYyNSwyMDAwLDE4NzUsMjUwMCwxMzc1LDIwMDAsMTg3NSwyMzc1LDIwMDAsMjM3NSwxODc1LDE4NzUsMTYyNSwxNzUwLDEzNzUsMTg3NSwyMDAwLDE2MjUsMTYyNSwyMDAwXSwib3BzIjoyMDU1fSx7Im5hbWUiOiJGaW5kIGl0ZW0gMjAwIiwiY29kZSI6ImFycmF5LnNvcnQoKGEsYikgPT4gKGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcikgfHwgKGEucmVuZGVyT3JkZXIgLSBiLnJlbmRlck9yZGVyKSB8fCAoYS56IC0gYi56KSkiLCJydW5zIjpbMTEyNSwxMTI1LDEzNzUsMTc1MCwxNzUwLDE2MjUsMjAwMCwxNTAwLDE2MjUsMTEyNSwzNzUsMTUwMCwxNTAwLDEyNTAsMTAwMCwxMjUwLDg3NSwyMDAwLDEwMDAsMTM3NSwxMjUwLDExMjUsMTUwMCwxMzc1LDE1MDAsMTM3NSwyMTI1LDE4NzUsMTc1MCwxNzUwLDE2MjUsMTM3NSwyMDAwLDE3NTAsMTg3NSwxNTAwLDE1MDAsMTM3NSwxNjI1LDEyNTAsMTUwMCwxNjI1LDEwMDAsMTUwMCwyMDAwLDE4NzUsMTc1MCwxNTAwLDE3NTAsMTM3NSwxNTAwLDEwMDAsMTYyNSwxNzUwLDEzNzUsMTYyNSwxMzc1LDExMjUsMTI1MCwxNjI1LDEzNzUsMTUwMCwxNjI1LDIwMDAsMTc1MCwxNzUwLDEzNzUsMTc1MCwxODc1LDE1MDAsMTYyNSwxNjI1LDEyNTAsMTUwMCwxNjI1LDE2MjUsMTUwMCwxNjI1LDExMjUsMTEyNSwxNjI1LDE1MDAsMTI1MCwxMzc1LDE3NTAsMTc1MCwxNTAwLDE1MDAsMTI1MCwxMDAwLDE1MDAsMTYyNSwxMDAwLDE1MDAsMTEyNSwxMTI1LDEzNzUsMjAwMCwxMzc1LDEyNTBdLCJvcHMiOjE0ODJ9XSwidXBkYXRlZCI6IjIwMjMtMDgtMDhUMDc6NDE6NTMuNTI4WiJ9): If-else chain faster on Chrome/Quest 2, slower on Firefox
- [random groupOrder, random renderOrder, random z](https://perf.link/#eyJpZCI6IjcyMm44ZWo1YjQzIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGFycmF5ID0gbmV3IEFycmF5KDEwMDAwKTtcbmZvcihsZXQgaSA9IDA7IGkgPCBhcnJheS5sZW5ndGg7IGkrKykge1xuXHRhcnJheVtpXSA9IHtcblx0XHRncm91cE9yZGVyOiBNYXRoLmZsb29yKE1hdGgucmFuZG9tKCkgKiAxMDAwMDApLFxuXHRcdHJlbmRlck9yZGVyOiBNYXRoLmZsb29yKE1hdGgucmFuZG9tKCkgKiAxMDAwMDApLFxuXHRcdHo6IE1hdGgucmFuZG9tKCkgKiAxMDAwMDAgLSA1MDAwMCxcblx0fTtcbn0iLCJ0ZXN0cyI6W3sibmFtZSI6IkZpbmQgaXRlbSAxMDAiLCJjb2RlIjoiYXJyYXkuc29ydCgoYSxiKSA9PiB7XG5cdGlmKGEuZ3JvdXBPcmRlciAhPT0gYi5ncm91cE9yZGVyKSB7XG5cdFx0cmV0dXJuIGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcjtcblx0fSBlbHNlIGlmKGEucmVuZGVyT3JkZXIgIT09IGIuZ3JvdXBPcmRlcikge1xuXHRcdHJldHVybiBhLnJlbmRlck9yZGVyIC0gYi5yZW5kZXJPcmRlcjtcblx0fSBlbHNlIHtcblx0XHRyZXR1cm4gYS56IC0gYi56O1xuXHR9XG59KSIsInJ1bnMiOlsxODAwLDIyMDAsMTYwMCwyMjAwLDIwMDAsMjQwMCwxNjAwLDI0MDAsMTQwMCwxODAwLDIyMDAsMTgwMCwyNjAwLDE0MDAsMjIwMCwyMjAwLDI0MDAsMjQwMCwyNDAwLDIwMDAsMTQwMCwxODAwLDI0MDAsMjAwMCwxNjAwLDE4MDAsMjQwMCwyMjAwLDIwMDAsMjIwMCwyNDAwLDE2MDAsMjAwMCwxNjAwLDIyMDAsMjQwMCwyMDAwLDI0MDAsMjQwMCwyNDAwLDIwMDAsMjIwMCwyMDAwLDI0MDAsMTQwMCwyMjAwLDIyMDAsMjQwMCwyNDAwLDI0MDAsMTYwMCwxODAwLDE4MDAsMTgwMCwyMjAwLDI0MDAsMjQwMCwyMDAwLDE4MDAsMjQwMCwxNjAwLDI2MDAsMjIwMCwyNjAwLDIwMDAsMjAwMCwxNjAwLDE4MDAsMjQwMCwyMDAwLDI0MDAsMjAwMCwxMjAwLDIwMDAsMjIwMCwyMDAwLDE2MDAsMjIwMCwxODAwLDIwMDAsMjQwMCwyNDAwLDI0MDAsMjAwMCwyMDAwLDI0MDAsMjQwMCwyMjAwLDI0MDAsMjAwMCwyNDAwLDI0MDAsMTYwMCwxODAwLDE2MDAsMjIwMCwxODAwLDE4MDAsMTYwMCwyNDAwXSwib3BzIjoyMDY0fSx7Im5hbWUiOiJGaW5kIGl0ZW0gMjAwIiwiY29kZSI6ImFycmF5LnNvcnQoKGEsYikgPT4gKGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcikgfHwgKGEucmVuZGVyT3JkZXIgLSBiLnJlbmRlck9yZGVyKSB8fCAoYS56IC0gYi56KSkiLCJydW5zIjpbMjIwMCwyNDAwLDE2MDAsMjAwMCwyMjAwLDIwMDAsMjAwMCwyNDAwLDIwMDAsMjQwMCwxODAwLDI0MDAsMjQwMCwyNDAwLDE4MDAsMjQwMCwyNDAwLDE4MDAsMjIwMCwyNDAwLDE2MDAsMTgwMCwyMjAwLDIwMDAsMTgwMCwyNDAwLDI0MDAsMjQwMCwyMDAwLDI0MDAsMjQwMCwyMDAwLDIyMDAsMjAwMCwyMDAwLDIwMDAsMjQwMCwyNDAwLDIyMDAsMTYwMCwyNDAwLDE4MDAsMjQwMCwyMDAwLDI2MDAsMjIwMCwyMjAwLDIwMDAsMjYwMCwyNDAwLDIyMDAsMjIwMCwyNDAwLDI0MDAsMTgwMCwyMDAwLDI0MDAsMjQwMCwyMDAwLDI0MDAsMTYwMCwyNDAwLDIyMDAsMjQwMCwyNDAwLDIwMDAsMTQwMCwyNDAwLDE4MDAsMjQwMCwyMjAwLDI0MDAsMjQwMCwyNDAwLDIwMDAsMjQwMCwyNDAwLDE0MDAsMjQwMCwxODAwLDIyMDAsMjAwMCwyNDAwLDI0MDAsMTgwMCwxODAwLDIwMDAsMjQwMCwyMjAwLDIwMDAsMTgwMCwyNDAwLDI0MDAsMjAwMCwyNDAwLDIwMDAsMjIwMCwyMjAwLDIyMDAsMjIwMF0sIm9wcyI6MjE2MH1dLCJ1cGRhdGVkIjoiMjAyMy0wOC0wOFQwNzo0NToxMy40MTRaIn0%3D): If-else chain slightly slower on Chrome and Firefox
- [1 groupOrder, 0 renderOrder, random z](https://perf.link/#eyJpZCI6ImN4c3h5cXY0aXl5IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGFycmF5ID0gbmV3IEFycmF5KDEwMDAwKTtcbmZvcihsZXQgaSA9IDA7IGkgPCBhcnJheS5sZW5ndGg7IGkrKykge1xuXHRhcnJheVtpXSA9IHtcblx0XHRncm91cE9yZGVyOiAxLFxuXHRcdHJlbmRlck9yZGVyOiAwLFxuXHRcdHo6IE1hdGgucmFuZG9tKCkgKiAxMDAwMDAgLSA1MDAwMCxcblx0fTtcbn0iLCJ0ZXN0cyI6W3sibmFtZSI6IkZpbmQgaXRlbSAxMDAiLCJjb2RlIjoiYXJyYXkuc29ydCgoYSxiKSA9PiB7XG5cdGlmKGEuZ3JvdXBPcmRlciAhPT0gYi5ncm91cE9yZGVyKSB7XG5cdFx0cmV0dXJuIGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcjtcblx0fSBlbHNlIGlmKGEucmVuZGVyT3JkZXIgIT09IGIuZ3JvdXBPcmRlcikge1xuXHRcdHJldHVybiBhLnJlbmRlck9yZGVyIC0gYi5yZW5kZXJPcmRlcjtcblx0fSBlbHNlIHtcblx0XHRyZXR1cm4gYS56IC0gYi56O1xuXHR9XG59KSIsInJ1bnMiOls0NzE0LDQyODUsNDAwMCw0NDI4LDQwMDAsMzU3MSwzNTcxLDQyODUsNDU3MSw0Mjg1LDQ1NzEsMjcxNCwzNzE0LDM1NzEsNDI4NSw0NTcxLDQxNDIsNDU3MSw0NTcxLDQ0MjgsMzg1Nyw0MTQyLDQwMDAsNDU3MSw0NDI4LDQ1NzEsNDU3MSw0Mjg1LDQ3MTQsNDU3MSw0NDI4LDQ3MTQsNDQyOCw0NTcxLDQxNDIsMzg1NywzODU3LDQ0MjgsMzg1NywzODU3LDM0MjgsMzcxNCwzNTcxLDQ0MjgsMzg1Nyw0MTQyLDQyODUsNDE0MiwzODU3LDQ3MTQsNDU3MSw0NzE0LDM3MTQsNDE0Miw0MDAwLDM3MTQsMzg1NywzNzE0LDM1NzEsMzg1Nyw0MTQyLDQwMDAsMzg1Nyw0NDI4LDQyODUsNDU3MSwzODU3LDQyODUsNDcxNCw0NDI4LDM3MTQsMzg1NywzODU3LDM0MjgsNDU3MSwzODU3LDM4NTcsMzU3MSwzNTcxLDM3MTQsMzcxNCwzNzE0LDM1NzEsMzQyOCwzNDI4LDQyODUsNDE0MiwzODU3LDQyODUsNDU3MSw0NzE0LDQ0MjgsNDU3MSw0NDI4LDQwMDAsNDU3MSw0Mjg1LDM4NTcsNDAwMCw0MTQyXSwib3BzIjo0MTEyfSx7Im5hbWUiOiJGaW5kIGl0ZW0gMjAwIiwiY29kZSI6ImFycmF5LnNvcnQoKGEsYikgPT4gKGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcikgfHwgKGEucmVuZGVyT3JkZXIgLSBiLnJlbmRlck9yZGVyKSB8fCAoYS56IC0gYi56KSkiLCJydW5zIjpbMTQyOCwxMjg1LDg1NywxMjg1LDEwMDAsMjg1LDg1NywxMTQyLDcxNCwxMTQyLDE1NzEsMTE0MiwxMjg1LDEyODUsMTE0MiwxMTQyLDU3MSw0MjgsODU3LDU3MSwxMjg1LDEwMDAsODU3LDEyODUsMTE0MiwxMDAwLDcxNCwxNTcxLDg1NywxMTQyLDEwMDAsNzE0LDE1NzEsODU3LDg1Nyw1NzEsMTAwMCwxMDAwLDcxNCwxMTQyLDcxNCwxMTQyLDg1Nyw4NTcsMTAwMCwxMTQyLDExNDIsNzE0LDEyODUsMTI4NSwxNTcxLDg1NywxMDAwLDExNDIsNDI4LDEwMDAsODU3LDEyODUsMTE0MiwxNTcxLDExNDIsODU3LDg1NywxNDI4LDQyOCwxMTQyLDQyOCw1NzEsMTcxNCw4NTcsMTE0Miw4NTcsODU3LDEwMDAsNDI4LDEwMDAsNzE0LDI4NSwyODUsODU3LDQyOCwxMDAwLDg1Nyw3MTQsNTcxLDcxNCwxMjg1LDcxNCw4NTcsMTE0MiwxNzE0LDE0MjgsNzE0LDg1NywxMjg1LDE1NzEsNTcxLDg1NywxMjg1LDExNDJdLCJvcHMiOjk3OH1dLCJ1cGRhdGVkIjoiMjAyMy0wOC0wOFQwNzo0NzozNS42NTBaIn0%3D) If-else completely dominates the compacted expression, both on Chrome and Firefox
- [0 groupOrder, 0 renderOrder, random z](https://perf.link/#eyJpZCI6IjZtMjl5eDI0Mm5hIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGFycmF5ID0gbmV3IEFycmF5KDEwMDAwKTtcbmZvcihsZXQgaSA9IDA7IGkgPCBhcnJheS5sZW5ndGg7IGkrKykge1xuXHRhcnJheVtpXSA9IHtcblx0XHRncm91cE9yZGVyOiAwLFxuXHRcdHJlbmRlck9yZGVyOiAwLFxuXHRcdHo6IE1hdGgucmFuZG9tKCkgKiAxMDAwMDAgLSA1MDAwMCxcblx0fTtcbn0iLCJ0ZXN0cyI6W3sibmFtZSI6IkZpbmQgaXRlbSAxMDAiLCJjb2RlIjoiYXJyYXkuc29ydCgoYSxiKSA9PiB7XG5cdGlmKGEuZ3JvdXBPcmRlciAhPT0gYi5ncm91cE9yZGVyKSB7XG5cdFx0cmV0dXJuIGEuZ3JvdXBPcmRlciAtIGIuZ3JvdXBPcmRlcjtcblx0fSBlbHNlIGlmKGEucmVuZGVyT3JkZXIgIT09IGIuZ3JvdXBPcmRlcikge1xuXHRcdHJldHVybiBhLnJlbmRlck9yZGVyIC0gYi5yZW5kZXJPcmRlcjtcblx0fSBlbHNlIHtcblx0XHRyZXR1cm4gYS56IC0gYi56O1xuXHR9XG59KSIsInJ1bnMiOlsxODc1LDEzNzUsMTEyNSwxNTAwLDE1MDAsMTYyNSwxMzc1LDE1MDAsMTYyNSwxMTI1LDE2MjUsMTc1MCwxMDAwLDIwMDAsMjAwMCwxNjI1LDE4NzUsMTc1MCwxMjUwLDEzNzUsMTI1MCwxMjUwLDExMjUsMTUwMCwxMDAwLDg3NSwxMzc1LDExMjUsMTEyNSwyMDAwLDE2MjUsMTUwMCwxNzUwLDE4NzUsMTg3NSwxNzUwLDE2MjUsMTg3NSwxODc1LDExMjUsMTc1MCwxMjUwLDE1MDAsMTg3NSwyMTI1LDE1MDAsMTEyNSwxMzc1LDEyNTAsMTUwMCwxMjUwLDIxMjUsMTg3NSwxODc1LDE2MjUsMTM3NSwxNzUwLDEzNzUsMTAwMCwxNzUwLDE3NTAsMTI1MCw4NzUsMTc1MCwxNzUwLDE2MjUsMTg3NSwxNjI1LDE1MDAsMTUwMCwxNTAwLDE2MjUsMTI1MCwxMjUwLDEyNTAsMTc1MCwyMDAwLDExMjUsMTg3NSwxNTAwLDE4NzUsMTUwMCwxODc1LDE3NTAsMTM3NSwxMzc1LDEzNzUsMTUwMCwyMDAwLDE3NTAsMTI1MCwxNjI1LDE2MjUsMjAwMCwxNTAwLDE1MDAsMTUwMCwyMDAwLDIxMjUsMTM3NV0sIm9wcyI6MTU1MX0seyJuYW1lIjoiRmluZCBpdGVtIDIwMCIsImNvZGUiOiJhcnJheS5zb3J0KChhLGIpID0%2BIChhLmdyb3VwT3JkZXIgLSBiLmdyb3VwT3JkZXIpIHx8IChhLnJlbmRlck9yZGVyIC0gYi5yZW5kZXJPcmRlcikgfHwgKGEueiAtIGIueikpIiwicnVucyI6WzE2MjUsMTM3NSwyMDAwLDE1MDAsMTYyNSwxMzc1LDEzNzUsMTM3NSwxNjI1LDE4NzUsMjAwMCwxNTAwLDExMjUsMjAwMCwxNTAwLDExMjUsMjUwLDE4NzUsMTUwMCwxMzc1LDE3NTAsMTYyNSwxMDAwLDE2MjUsMTg3NSwxMzc1LDYyNSwxMzc1LDExMjUsMTg3NSwxMzc1LDEzNzUsMTYyNSwxNzUwLDEzNzUsMTYyNSwxMzc1LDEyNTAsMTM3NSwxODc1LDE3NTAsMTYyNSwxMzc1LDE2MjUsMTUwMCwxNjI1LDEwMDAsMTAwMCwxMDAwLDE3NTAsMTUwMCwxNjI1LDExMjUsMTM3NSwxMjUwLDEzNzUsMTEyNSwxNzUwLDEzNzUsMjEyNSwxNjI1LDE1MDAsMTc1MCwxNTAwLDE1MDAsMTYyNSwxMzc1LDE1MDAsMTI1MCwxMTI1LDE4NzUsMTYyNSwxNTAwLDExMjUsMTUwMCwxNTAwLDExMjUsMTUwMCwxNjI1LDEzNzUsMTI1MCwzNzUsMTYyNSwxMzc1LDE3NTAsMTYyNSwxODc1LDE4NzUsMTYyNSwxNjI1LDE1MDAsMTYyNSwxNzUwLDE3NTAsMTM3NSwxMzc1LDE2MjUsMTEyNSwzNzUsMTYyNV0sIm9wcyI6MTQ2N31dLCJ1cGRhdGVkIjoiMjAyMy0wOC0wOFQwNzo1MDo1My41NDhaIn0%3D): About equal, if-else slightly faster on Chrome, but slightly slower on Firefox

Result obviously may vary and I don't think the performance of this sorting function would really matter at the end of the day as most VR experiences tend to have relatively low object counts anyway.

diarmidmackenzie: > Wouldn't it make sense to retain the ID tie-breaker (return a.id - b.id) used by Three.js as well? That should ensure that, everything else being equal, the order is consistent.

That would sort the objects in creation order, rather than DOM order.  My understanding is that returning 0 will leave order as-is, which implies DOM order, which matches current behaviour, is explicable in A-Frame terms, and gives the user a level of control by inserting items earlier in the DOM.

IMO that's superios to object id / creation order.

diarmidmackenzie: > As for compacting it to a single expression, I personally prefer the current more verbose version. 

Thanks for doing the perf testing, looks to me like the more verbose version is both easier to read, and probably also more performant (I don't think random data is particularly realistic, and even in that case, if-else seems to be only marginally slower).

So I think that weighs in favour of the code as-is.  Which also matches the equivalent THREE.js code.

There may be a case for eliminating the final if test and simply returning `a.z - b.z` though, so I'll make that change.

src/systems/renderer.js
```diff
@@ -83,3 +96,59 @@ module.exports.System = registerSystem('renderer', {
...
+// Default sort for transparent objects:
+// - respect groupOrder & renderOrder settings
+// - otherwise leave objects in default order (object tree order)
+function sortRenderOrderOnly (a, b) {
+  if (a.groupOrder !== b.groupOrder) {
+    return a.groupOrder - b.groupOrder;
+  } else if (a.renderOrder !== b.renderOrder) {
+    return a.renderOrder - b.renderOrder;
+  } else {
+    return 0;
```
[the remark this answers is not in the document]
diarmidmackenzie: Per above, I believe this change reduces readability, diverges unnecessarily from equivalent THREE.js code, and doesn't improve (and may even harm) performance.  So I'd prefer to leave as-is.

src/systems/renderer.js
diarmidmackenzie: Per above, I believe this change reduces readability, diverges unnecessarily from equivalent THREE.js code, and doesn't improve (and may even harm) performance.  So I'd prefer to leave as-is.

src/systems/renderer.js
diarmidmackenzie: agree FrontToBack implies Spatial, so it's redundant.  I'll fix.

dmarcos:
Thanks so much for the patience! This is awesome

PR #5342, opened 2023-07-22
Improve linting to check for ES5 conformance (with exceptions to allow use of ES6 classes in certain modules)

diarmidmackenzie:
**Description:**

The A-Frame codebase is intended to only use ES5 JavaScript - with the exception of ES6 classes, which are used sparingly in a few places.

Currently, the enforcement of this is manual during PR reviews, which is (a) not completely reliable, and (b) often creates unecessary reviewing cycles, particular for PRs from contributors who are used to using ES6 features.

As proposed in #5218, this PR updates linting to enforce the rule that code in /src should only use ES5 JavaScript.

In order to permit the use of ES6 classes where required, this PR exempts the small number of modules that use ES6 classes from the ES5 checks.  While that's not perfect, it still enables ES5 checking for 90% of the /src tree, which should be valuable in helping contributors avoid accidentally including ES6 constructs in PRs.

**Changes proposed:**

- Update from semistandard v9 (6 years old) to standardx v17 (latest).  standardx allows us to use the semistandard config as a base, but fine-tune exactly which rules we want to apply.  This also brings us up to a modern version of ESLint.

- Parse all code in /src as ES5, so linting errors will occur if ES6 constructs are used.  A small number of core modules that use ES6 classes are excepted, and parsed as ES6.  Note that we are unable to detect ES6 usage in these modules.

- Bring code into line with latest v17 semistandard rules, with just 3 classes of exceptions:
1. rules that require ES6 constructs
2. rules that require non-trivial code change.  We can adopt these rules, but better handled as separate PRs.
3. rules that require widespread code changes (even if trivial).  We can adopt these rules, but we should handle these with care as separate PRs due to the risk of creating merge issues with other PRs that may be in progress.

To conform with the new rules, there's some fairly trivial changes to ~25 modules, reflecting [changes over the last 6 years to the semistandard conventions](https://standardjs.com/changelog.html).  I don't believe any of these will be controversial.


diarmidmackenzie:
@dmarcos Any thoughts on this one?

Personally, this would be a big help when contributing to A-Frame, in helping to ensure my code is ES5-compliant.  Are there any downsides you are concerned about?

src/components/hand-tracking-controls.js
```diff
@@ -172,7 +172,7 @@ module.exports.Component = registerComponent('hand-tracking-controls', {
       if (!controller || !this.mesh) { return; }
       this.mesh.visible = false;
       if (!this.hasPoses) { return; }
-      for (var inputjoint of controller.hand.values()) {
+      controller.hand.values().forEach(function (inputjoint) {
```
dmarcos, line 175: This two are not equivalent. this in the `for` not the same as this in the function scope. In general I favor `for` loops over callbacks that introduce new scopes and are potentially confusing. not a fan of arrow function either since it introduces multiple ways to define a function with different implicit scoping rules. Super confusing specially for newbies.

dmarcos:
This needs rebase. Sorry for the wait.

FYI, ES6 classes is not part of A-Frame code style. Forced on the code base because of the custom elements API. I don't think it matters for this PR though.

examples/showcase/ui/info-panel.js
```diff
@@ -4,7 +4,7 @@ AFRAME.registerComponent('info-panel', {
     var buttonEls = document.querySelectorAll('.menu-button');
     var fadeBackgroundEl = this.fadeBackgroundEl = document.querySelector('#fadeBackground');
 
-    this.movieImageEl;
```
dmarcos, line 7: we can probably remove this line all together

src/components/hand-tracking-controls.js
diarmidmackenzie: Thanks for catching this!

Have used a classic for loop instead for this one.

examples/showcase/ui/info-panel.js
diarmidmackenzie: Sure - have done so.

diarmidmackenzie:
Rebased against latest master.  Should be OK to merge now.

dmarcos:
Thanks so much for the patience and quick turnaround. You rock!

PR #5357, opened 2023-10-09
Update text plane size following text mutation (fixes #2764).

### LINKED ISSUE #2764: text not updating geometry/layout when text is updated
**Description:**

Given a text container `<a-entity text geometry="primitive: plane; height: auto" material>`, continually update `text.value` with new lines. The height of the plane should automatically update, but it seems it might be stuck at the initial height of the text.

- A-Frame Version: 0.5.0 + master
- Platform / Device: desktop

brycethomas:
The plane width/height was being calculated when the text value was first set, but would not update when the text changed.

Documentation has also been updated to advise using `0` rather than `auto` for autoscaling behavior.  This is based on the the comment [here](https://github.com/aframevr/aframe/issues/2837#issuecomment-315923754) suggesting that `auto` is not a valid value for geometry (NaN is falsy, so it happened to still work).

### DIFF
--- a/docs/components/text.md
+++ b/docs/components/text.md
@@ -241,7 +241,7 @@ geometry component's `width`, do not specify a `width` for the text component:
 
 ```html
 <a-entity
-  geometry="primitive: plane; width: 4; height: auto"
+  geometry="primitive: plane; width: 4; height: 0"
   material="color: blue"
   text="value: This text will be 4 units wide."></a-entity>
 ```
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks for taking the time. Have been a while I touched this code. Your knowledge is more fresh. At first look I would expect the update method to be called when updating the text and this logic to change the plane size to execute:

https://github.com/aframevr/aframe/pull/5357/files#diff-59e71534a2828d71b6a8c63a750a375c722e58636a0717b4afef195333e773d3L305

Why is not working as expected?

brycethomas:
The old code checks whether the geometry width/height is falsy, and if so, sets the geometry width/height.  On the next run through update the width/height will no longer be falsy (it was set on the prior update), and so the width/height doesn't get adjusted when the text changes.

dmarcos:
Thanks for the tests!

Issue #5364, opened 2023-10-18
Missing a way to set the start time of sound components


JonathannJacobs:
Hi, well first thing, it's not a bug but a missing feature.

I wish to have a way to play a sound entirely at first loop and then rewind at some specific start loop value different than zero.
So far i had the following code : 

this.el.removeAttribute('sound')
this.el.setAttribute('sound', 'src', 'assets/some.mp3');
this.el.components.sound.playSound()
this.el.addEventListener('sound-ended', () => {
  // Missing here a way to set the next playSound start value
  this.el.components.sound.playSound()
});

And that is sad because i could totally do it with a regular audio html element that lacks positionnal audio and pool.

Currently my best shot is to split my file into two mp3: one for the intro and another for the looping part.

- A-Frame Version: 1.4.2
- Platform / Device: All

mrxz:
It would be nice for the built-in sound component to expose a `loopStart` and `loopEnd`, but currently it doesn't. Luckily you can quite easily reach into the sound component and get to the `THREE.Audio` instance or even further to the underlying WebAudio Nodes.

For example:
```js
const soundComponent = this.el.components['sound'];
const audio = soundComponent.pool.children[0]; // THREE.Audio
audio.setLoopStart(/* start time in seconds */);
audio.setLoopEnd(audio.buffer.duration)
```

It can be a bit tricky when to execute this, as it needs to take place _after_ the sound component has initialized (and the sound loaded, for the duration), but before playing the audio (otherwise the `loopStart` and `loopEnd` of THREE.Audio don't apply). Should be good to handle it in `sound-loaded`:  `this.el.addEventListener('sound-loaded', e => {})`

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5365

PR #5365, opened 2023-10-18
#5364 Bugfix : Sound : processSound callback saved and used with loaded + mustPlay

JonathannJacobs:
The processSound callback given to playSound() before the sound was loaded will now be called when loaded and not ignored anymore.

### DIFF
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -12,6 +12,8 @@ module.exports.Component = registerComponent('sound', {
     autoplay: {default: false},
     distanceModel: {default: 'inverse', oneOf: ['linear', 'inverse', 'exponential']},
     loop: {default: false},
+    loopStart: {default: 0},
+    loopEnd: {default: 0},
     maxDistance: {default: 10000},
     on: {default: ''},
     poolSize: {default: 1},
@@ -59,6 +61,8 @@ module.exports.Component = registerComponent('sound', {
         sound.setRolloffFactor(data.rolloffFactor);
       }
       sound.setLoop(data.loop);
+      sound.setLoopStart(data.loopStart);
+      sound.setLoopEnd(data.loopEnd);
       sound.setVolume(data.volume);
       sound.isPaused = false;
     }
--- a/src/components/sound.js
+++ b/src/components/sound.js
@@ -62,7 +62,13 @@ module.exports.Component = registerComponent('sound', {
       }
       sound.setLoop(data.loop);
       sound.setLoopStart(data.loopStart);
-      sound.setLoopEnd(data.loopEnd);
+
+      // With a loop start specified without a specified loop end, the end of the loop should be the end of the file
+      if(data.loopStart != 0 && data.loopEnd == 0){ 
+        sound.setLoopEnd(sound.buffer.duration);
+      }
+      else sound.setLoopEnd(data.loopEnd);
+
       sound.setVolume(data.volume);
       sound.isPaused = false;
     }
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. Can you elaborate a scenario that illustrates the bug? Not sure I understand the fix.

JonathannJacobs:
Hi dmarcos ! First, thank you for aframe, i love it.
We already talked of the issue [here](https://github.com/aframevr/aframe/issues/5364)

Summary : 
playSound have one argument, a process callback. If playSound is not yet loaded, there is a mechanic with mustPlay and some loaded events that will call playSound later, when it can run. However the given callback argument was lost here.

With the fix it is kept and used as expected.

I also added like mrzx suggested directly the loopStart and loopEnd API bindings.

src/components/sound.js
```diff
@@ -58,6 +61,14 @@ module.exports.Component = registerComponent('sound', {
         sound.setRolloffFactor(data.rolloffFactor);
       }
       sound.setLoop(data.loop);
+      sound.setLoopStart(data.loopStart);
+
+      // With a loop start specified without a specified loop end, the end of the loop should be the end of the file
+      if(data.loopStart != 0 && data.loopEnd == 0){ 
+        sound.setLoopEnd(sound.buffer.duration);
+      }
+      else sound.setLoopEnd(data.loopEnd);
```
dmarcos, line 70:
```
else {
  sound.setLoopEnd(data.loopEnd);
}
```
we always use brackets for if statements

src/components/sound.js
```diff
@@ -208,6 +219,9 @@ module.exports.Component = registerComponent('sound', {
     if (!this.loaded) {
       warn('Sound not loaded yet. It will be played once it finished loading');
       this.mustPlay = true;
+      if(processSound){
```
dmarcos, line 222: single liner:
```
 if (processSound) { this.processSound = processSound; }
```

src/components/sound.js
```diff
@@ -32,6 +34,7 @@ module.exports.Component = registerComponent('sound', {
     this.pool = new THREE.Group();
     this.loaded = false;
     this.mustPlay = false;
+    this.processSound = undefined; // Saved callback for the mustPlay mechanic
```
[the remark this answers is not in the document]
dmarcos: Where / when is this callback called? If a callback variable should be named accordingly: `playSoundCallback`?

src/components/sound.js
dmarcos: not sure if the if statement is necessary. probably can just do `this.processSound = processSound`

src/components/sound.js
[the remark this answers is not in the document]
JonathannJacobs: Used at line 96 : if (self.data.autoplay || self.mustPlay) { self.playSound(this.processSound); }

Yes i understand it looks useless to explicitely declare it undefined but it was giving me a chance to add a comment here. I'll remove it.

src/components/sound.js
JonathannJacobs: I see, will change.

src/components/sound.js
JonathannJacobs: True, will do.

dmarcos:
Thanks so much!

arpu:
hi , after this change i get in the situation 

```
sound.js:86 Uncaught TypeError: Cannot read properties of undefined (reading 'processSound')
    at eval (sound.js:86:1)
    at AudioContext.eval (three.module.js:47396:6)
eval @ sound.js:86
```
on line https://github.com/aframevr/aframe/blob/master/src/components/sound.js#L94

mrxz:
@arpu Could you test if the changes in this PR https://github.com/aframevr/aframe/pull/5414 fixes the issue?

PR #5377, opened 2023-11-02
Update hand-tracking-grab-controls.js

addy1997:
changed function name from **tock** to **tick**


dmarcos:
Is there any problem with the tock function?

vincentfretin:
Why? In case you didn't know, it's not typo, you can create a tock method on a component, that is invoked after the scene has rendered.

addy1997:
Sorry! @dmarcos. I thought that tick was the default function. Hence, this PR. I'll close it

dmarcos:
No problem. To be honest my `tock` is a bit sloppy. Started with a `tick` but was getting some object position updates lagging behind by one frame. `tock` seemed to fix it but didn't fully understand why. So your PR might make sense in the future :)

PR #5379, opened 2023-11-04
Remove remaining link to gearvr-controls in laser-controls.md

vincentfretin:
After https://github.com/aframevr/aframe/commit/c981203ceef93f6f1c9177f83f6aa06ac0d5a38a one link to gearvr-controls was still there, remove it. Also remove tracked-controls link ref that is not used.


dmarcos:
Thanks

PR #5382, opened 2023-11-14
enterVR: reject promise if xr.requestSession fails

JL-Vidinoti:
**Description:**

When calling enterVR (or enterAR), the function `navigator.xr.requestSession` may fail (e.g. if the user denied the required permissions). In this case, an error is thrown but it is impossible to catch it.

**Changes proposed:**
- Instead of simply throwing the error, the returned promise is rejected with the error.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -308,7 +308,7 @@ class AScene extends AEntity {
             function requestFail (error) {
               var useAR = xrMode === 'immersive-ar';
               var mode = useAR ? 'AR' : 'VR';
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. Just curious. In what context did you find this an issue? Might inform other decisions. Thanks!

dmarcos:
Congrats on your first contribution!

JL-Vidinoti:
I discovered the issue while doing some tests in AR. When denying the Chrome permission, the AR mode was not starting and no error message was displayed.
Now I will be able to catch the error and show a message for the user.

PR #5386, opened 2023-11-16
Remove outdated skinning property to prevent a warning

avdynut:
When hands are used a warning appears that the skinning property is not part of material


dmarcos:
Thanks. This seems was needed before. Is there anything else we should do instead besides removing the property?

avdynut:
I think it can be safely removed, because SkinnedMesh is used. For reference see https://github.com/mrdoob/three.js/commit/8c21891045d6c7db07d515133f9c2c98bd4913a9

dmarcos:
Thanks and congrats for your first contribution 🏅

PR #5390, opened 2023-11-19
Correct default value for new XRMode field

diarmidmackenzie:
**Description:**

1.5.0 docs say the default value for the new field is `vr, ar, xr`.  In fact the default value is `vr`, and `ar` & `xr` are other possible values.

The relevant line of the schema is:
```
XRMode: {default: 'vr', oneOf: ['vr', 'ar', 'xr']}
```

**Changes proposed:**

Update docs to indicate the actual default value.


dmarcos:
Thanks!

PR #5398, opened 2023-11-29
[docs] Fix wrong default value link.md

Samu01Tech:
**Description:**
The default value for the property `visualAspectEnabled` is `false`

**Changes proposed:**
Set it to `true`


dmarcos:
Thanks so much!

PR #5399, opened 2023-11-29
Only enable shadowMap if an entity uses the shadow component and the shadow system is enabled

mrxz:
**Description:**
The shadow system has logic for only enabling once an entity uses shadows. However, this logic is broken since https://github.com/aframevr/aframe/pull/4040, resulting in `renderer.shadowMap.enabled` to always default to true. This means that by default every A-Frame scene renders an empty shadowMap and compiles the materials with `USE_SHADOWMAP` defined.

On top of that, while the system would update `renderer.shadowMap.enabled` when changing the system's enabled property, this alone isn't enough to actually toggle the shadows in Three.js. It's needed for materials to be updated whenever `shadowMap.enabled` changes ([How to update things](https://threejs.org/docs/index.html#manual/en/introduction/How-to-update-things)).

This PR addresses both issues, restoring the original intended behaviour and ensuring enabling and disabling the shadow system on the fly works as well.

**Changes proposed:**
- Use `this.data.enabled` for if the shadow _system_ is enabled and `this.shadowMapEnabled` for if a shadowMap is needed (e.g. an entity uses the `shadow` component). As a result `renderer.shadowMap.enabled = this.data.enabled && this.shadowMapEnabled`
- Mark all materials as `needsUpdate` whenever the actual value of `renderer.shadowMap.enabled` changes
- Update unit tests to reflect the above changes

### DIFF
--- a/src/systems/shadow.js
+++ b/src/systems/shadow.js
@@ -26,16 +26,13 @@ module.exports.System = registerSystem('shadow', {
 
     this.shadowMapEnabled = false;
 
-    if (!sceneEl.renderer) { return; }  // For tests.
-
     sceneEl.renderer.shadowMap.type = SHADOW_MAP_TYPE_MAP[data.type];
     sceneEl.renderer.shadowMap.autoUpdate = data.autoUpdate;
-    this.setShadowMapEnabled(this.shadowMapEnabled);
   },
 
   update: function (prevData) {
     if (prevData.enabled !== this.data.enabled) {
-      this.setShadowMapEnabled(this.data.enabled);
+      this.setShadowMapEnabled(this.shadowMapEnabled);
     }
   },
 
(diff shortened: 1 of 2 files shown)


src/systems/shadow.js
```diff
@@ -44,10 +41,25 @@ module.exports.System = registerSystem('shadow', {
...
-    this.shadowMapEnabled = this.data.enabled && enabled;
-    if (renderer) {
-      renderer.shadowMap.enabled = this.shadowMapEnabled;
+
+    this.shadowMapEnabled = enabled;
+    var newEnabledState = this.data.enabled && this.shadowMapEnabled;
+    if (renderer && newEnabledState !== renderer.shadowMap.enabled) {
+      renderer.shadowMap.enabled = newEnabledState;
+
+      // If scene has already rendered, materials must be updated.
```
dmarcos, line 52: Maybe this can be factored out to an independent function ([example](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L859)) at the bottom of the file. `updateAllMaterials(sceneEl)`

mrxz: Done

dmarcos:
Thanks so much. You rock 🎸

PR #5404, opened 2023-12-09
Remove epsilon for scale as Three.js handles zero scale

mrxz:
**Description:**
Three.js gracefully handles objects with zero scale for quite a while now (https://github.com/mrdoob/three.js/issues/17926, https://github.com/mrdoob/three.js/pull/18885). As a result it's no longer needed for A-Frame to use an epsilon value to prevent scale from becoming zero along any or all axes.

For the animation component this also allows the animation to go from positive to negative or vice versa, which is currently prevented.

**Changes proposed:**
- Remove epsilon value when scaling to 0 (with `scale` or `animation` component)

### DIFF
--- a/src/components/animation.js
+++ b/src/components/animation.js
@@ -376,12 +376,6 @@ module.exports.Component = registerComponent('animation', {
         return function (anim) {
           var value = anim.animatables[0].target;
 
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

PR #5405, opened 2023-12-09
Update hand-tracking-grab-controls.md

ogreenaway:
**Description:**
The schema for [hand-tracking-grab-controls](https://github.com/aframevr/aframe/blob/v1.5.0/src/components/hand-tracking-grab-controls.js) uses the property "color" instead of "handColor" as mentioned in the documentation.

**Changes proposed:**
- rename property in the documentation to match the JavaScript implementation


dmarcos:
Thanks. We should probably rename the property to `handColor` instead. Seems more accurate. What do you think?

dmarcos:
Thanks!

dmarcos:
Congrats on your first contribution! 🎉

PR #5410, opened 2023-12-11
Add support for offerSession

cabanier:
`offerSession` is a recent addition to WebXR that makes it more obvious that a page can enter WebXR.
Quest browser got feedback that the "Enter VR/AR/XR" buttons on a lot of sites is hard to find. 
Often the button is small or scrolled of the bottom of the page. To remedy this, we worked with the W3C Immersive group on a new API that adds a button to the URL bar. This button will launch a pending immersive session.

This change will call `offerSession` when the page is opened or when the immersive session exits.

cc @dmarcos

### DIFF
--- a/src/components/scene/xr-mode-ui.js
+++ b/src/components/scene/xr-mode-ui.js
@@ -150,6 +150,7 @@ module.exports.Component = registerComponent('xr-mode-ui', {
     } else {
       if (!utils.device.checkVRSupport()) { this.enterVREl.classList.add('fullscreen'); }
       this.enterVREl.classList.remove(HIDDEN_CLASS);
+      sceneEl.enterVR(false, true);
     }
   },
 
@@ -161,6 +162,7 @@ module.exports.Component = registerComponent('xr-mode-ui', {
       this.enterAREl.classList.add(HIDDEN_CLASS);
     } else {
       this.enterAREl.classList.remove(HIDDEN_CLASS);
+      sceneEl.enterVR(true, true);
     }
   },
 
--- a/src/components/scene/xr-mode-ui.js
+++ b/src/components/scene/xr-mode-ui.js
@@ -120,6 +120,8 @@ module.exports.Component = registerComponent('xr-mode-ui', {
 
     this.orientationModalEl = createOrientationModal(this.onModalClick);
     sceneEl.appendChild(this.orientationModalEl);
+
+    this.updateEnterInterfaces();
   },
 
   remove: function () {
--- a/src/components/scene/xr-mode-ui.js
+++ b/src/components/scene/xr-mode-ui.js
@@ -120,8 +120,6 @@ module.exports.Component = registerComponent('xr-mode-ui', {
 
     this.orientationModalEl = createOrientationModal(this.onModalClick);
     sceneEl.appendChild(this.orientationModalEl);
-
-    this.updateEnterInterfaces();
   },
 
   remove: function () {
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -56,6 +56,7 @@ class AScene extends AEntity {
     self.systems = {};
     self.systemNames = [];
     self.time = self.delta = 0;
+    self.usedOfferSession = false;
 
     self.behaviors = {tick: [], tock: []};
     self.hasLoaded = false;
@@ -270,14 +271,15 @@ class AScene extends AEntity {
    * @param {bool?} useAR - if true, try immersive-ar mode
    * @returns {Promise}
    */
-  enterVR (useAR, useOfferSession = false) {
+  enterVR (useAR, useOfferSession) {
     var self = this;
     var vrDisplay;
     var vrManager = self.renderer.xr;
     var xrInit;
 
     // Don't enter VR if already in VR.
-    if (navigator.xr.offerSession === undefined) { return Promise.resolve('OfferSession is not supported.'); }
+    if (useOfferSession && navigator.xr.offerSession === undefined) { return Promise.resolve('OfferSession is not supported.'); }
+    if (self.usedOfferSession) { return Promise.resolve('OfferSession was already called.'); }
     if (this.is('vr-mode')) { return Promise.resolve('Already in VR.'); }
 
     // Has VR.
@@ -296,9 +298,11 @@ class AScene extends AEntity {
         xrInit = this.sceneEl.systems.webxr.sessionConfiguration;
         return new Promise(function (resolve, reject) {
           var requestFunction = useOfferSession ? navigator.xr.offerSession.bind(navigator.xr) : navigator.xr.requestSession.bind(navigator.xr);
+          self.usedOfferSession |= useOfferSession;
           requestFunction(xrMode, xrInit).then(
             function requestSuccess (xrSession) {
               self.xrSession = xrSession;
+              self.usedOfferSession = false;
 
               vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1;
               vrManager.setSession(xrSession).then(function () {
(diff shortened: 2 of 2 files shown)


src/components/scene/xr-mode-ui.js
```diff
@@ -150,6 +150,7 @@ module.exports.Component = registerComponent('xr-mode-ui', {
     } else {
       if (!utils.device.checkVRSupport()) { this.enterVREl.classList.add('fullscreen'); }
       this.enterVREl.classList.remove(HIDDEN_CLASS);
+      sceneEl.enterVR(false, true);
```
dmarcos, line 153: We should probably handle this a bit differently. Give control to the dev and decide if the page is auto-entering VR. Maybe a property in this component `enterImmersiveModeAtPageLoad` a tad long but descriptive. open to a shorter alternative. We can then probably handled the sceneEl.enterVR here https://github.com/aframevr/aframe/pull/5410/files#diff-8eae4e9d86bc7d08e7db133212411b40a655722a0d8c42c1c99c45fccef72f74L124 before deciding to display the UI

src/core/scene/a-scene.js
```diff
@@ -270,13 +270,15 @@ class AScene extends AEntity {
...
-  enterVR (useAR) {
+  enterVR (useAR, useOfferSession = false) {
     var self = this;
     var vrDisplay;
     var vrManager = self.renderer.xr;
     var xrInit;
 
     // Don't enter VR if already in VR.
+    if (navigator.xr.offerSession === undefined) { return Promise.resolve('OfferSession is not supported.'); }
+    if (this.is('has-offer-session')) { return Promise.resolve('Already has offerSession.'); }
```
dmarcos, line 281: Probably don't need this branch here line below should catch if we're already in immersive mode. I think we might not need the `has-offer-session` state

https://github.com/aframevr/aframe/pull/5410/files#diff-ea7d53491f815a744ea89a9a512d67374cb7b8741284b7ed041fde95e53610e1L280

cabanier: `toggleEnterVRButtonIfNeeded` is called multiple times when starting up. The previous pending promise then becomes failed which is flagged as an runtime error by aframe.

src/core/scene/a-scene.js
```diff
@@ -294,9 +296,15 @@ class AScene extends AEntity {
         var xrMode = useAR ? 'immersive-ar' : 'immersive-vr';
         xrInit = this.sceneEl.systems.webxr.sessionConfiguration;
         return new Promise(function (resolve, reject) {
-          navigator.xr.requestSession(xrMode, xrInit).then(
+          var requestFunction = useOfferSession ? navigator.xr.offerSession.bind(navigator.xr) : navigator.xr.requestSession.bind(navigator.xr);
```
dmarcos, line 299: can just call `requestSession`

src/core/scene/a-scene.js
dmarcos: toggleEnterVRButtonIfNeeded shouldn't be called multiple times. We're doing something wrong I think :)

src/components/scene/xr-mode-ui.js
```diff
@@ -120,8 +120,6 @@ module.exports.Component = registerComponent('xr-mode-ui', {
 
     this.orientationModalEl = createOrientationModal(this.onModalClick);
     sceneEl.appendChild(this.orientationModalEl);
-
-    this.updateEnterInterfaces();
```
cabanier, line 124: This caused `updateEnterInterfaces` to be called multiple times.

dmarcos, line 124: I think this needs to be called so [AR], [VR] or both buttons are displayed based on dev configuration

src/core/scene/a-scene.js
```diff
@@ -270,13 +270,14 @@ class AScene extends AEntity {
...
    */
-  enterVR (useAR) {
+  enterVR (useAR, useOfferSession = false) {
     var self = this;
     var vrDisplay;
     var vrManager = self.renderer.xr;
     var xrInit;
 
     // Don't enter VR if already in VR.
+    if (navigator.xr.offerSession === undefined) { return Promise.resolve('OfferSession is not supported.'); }
```
dmarcos, line 280: Should we check the useOfferSession parameter here as well?

src/core/scene/a-scene.js
cabanier: no, we need to call offerSession here to show the button

src/core/scene/a-scene.js
```diff
@@ -270,13 +270,14 @@ class AScene extends AEntity {
    * @param {bool?} useAR - if true, try immersive-ar mode
    * @returns {Promise}
    */
-  enterVR (useAR) {
+  enterVR (useAR, useOfferSession = false) {
```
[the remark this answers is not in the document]
cabanier: fixed

src/core/scene/a-scene.js
cabanier: fixed

src/components/scene/xr-mode-ui.js
dmarcos: we can restore this

src/core/scene/a-scene.js
dmarcos: Sorry, I meant renaming `requestFunction` variable to `requestSession`

src/core/scene/a-scene.js
```diff
@@ -294,9 +297,13 @@ class AScene extends AEntity {
         var xrMode = useAR ? 'immersive-ar' : 'immersive-vr';
         xrInit = this.sceneEl.systems.webxr.sessionConfiguration;
         return new Promise(function (resolve, reject) {
-          navigator.xr.requestSession(xrMode, xrInit).then(
+          var requestFunction = useOfferSession ? navigator.xr.offerSession.bind(navigator.xr) : navigator.xr.requestSession.bind(navigator.xr);
+          self.usedOfferSession |= useOfferSession;
```
dmarcos, line 301: can just do assign here

cabanier: Since this same function is called for regular requestSession in response to a button click, we need to take some extra steps here.
This code is also not correct. I will update in the next diff.

src/core/scene/a-scene.js
```diff
@@ -294,9 +297,16 @@ class AScene extends AEntity {
...
         return new Promise(function (resolve, reject) {
-          navigator.xr.requestSession(xrMode, xrInit).then(
+          var requestSession = useOfferSession ? navigator.xr.offerSession.bind(navigator.xr) : navigator.xr.requestSession.bind(navigator.xr);
+          self.usedOfferSession |= useOfferSession;
+          requestSession(xrMode, xrInit).then(
             function requestSuccess (xrSession) {
               self.xrSession = xrSession;
+
+              if (useOfferSession) {
+                self.usedOfferSession = false;
```
dmarcos, line 307: Probably can always set this to false here?

dmarcos: without checking useOfferSession

cabanier: I believe the regular requestSession API will then also set this to false. At that point we'll be in a bad state because the offersession promise is still pending.

src/core/scene/a-scene.js
```diff
@@ -270,13 +271,15 @@ class AScene extends AEntity {
...
    */
-  enterVR (useAR) {
+  enterVR (useAR, useOfferSession) {
     var self = this;
     var vrDisplay;
     var vrManager = self.renderer.xr;
     var xrInit;
 
     // Don't enter VR if already in VR.
+    if (useOfferSession && (navigator.xr === undefined || navigator.xr.offerSession === undefined)) { return Promise.resolve('OfferSession is not supported.'); }
```
dmarcos, line 281: can shorten to `if (useOfferSession && (!navigator.xr || !navigator.xr.offerSession))`

Issue #5411, opened 2023-12-12
cursor should not trigger mouseenter when in VR


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.5.0
- Platform / Device: Quest 2, Pro or 3

When building an experience that can operate in desktop or VR mode, it's common to build desktop interactions using a `cursor`, but then use a different interaction system in VR (controller-based or hand-tracking).

In VR mode, we don't expect the `cursor` to fire a "mouseenter" event, but sometimes it does.

The reason is that the raycaster associated with the cursor remains in the scene on entering VR, pointing in whatever directon the last mouse pointer position was detected prior to entering VR.

If this ray intersects an object, the raycaster will fire a `raycaster-intersection` event, which will lead to `cursor` emitting a `mousenter` event [here](https://github.com/aframevr/aframe/blob/7856a058266f0aab94c262284516a6f123abdb11/src/components/cursor.js#L423)

This can trigger unexpected mouse hover effects to appear in applications that haven't anticipated for the possibility of receiving a mouseenter event in VR mode.

I'm not entirely sure how to fix this.  First idea was to check for the combination of `rayOrigin: mouse` & vr mode, in `setInteraction()` and do an early return

```
if (this.data.rayOrigin === 'mouse' && 
    this.el.sceneEl.is('vr-mode')) return;
```

However there may be some cases where a mouse is a valid input mechanism in VR mode.  E.g. on desktop if you enter full screen, you are "in vr" (`this.el.sceneEl.is('vr-mode')` returns `true`) and yet mouse remains a valid input mechanism.

I don't think there's a simple query we can do to determine whether we are on a platform where mouse input is valid or not.  In the absence of such a query a fix might be:

- on entering VR, set a flag `this.mouseAvailable` to `false`
- on detection of a mouse move or click event, set `this.mouseAvailable` to `true`
- use `this.mouseAvailable` to determine whether ot not to so an early return on detection of a raycaster intersection.

I'll draft a PR based on this concept, as a basis for further discussion.

dmarcos:
> However there may be some cases where a mouse is a valid input mechanism in VR mode. E.g. on desktop if you enter full screen, you are "in vr" (this.el.sceneEl.is('vr-mode') returns true) and yet mouse remains a valid input mechanism.

Not sure that's a very common use case or even makes sense in immersive mode (no 2D screen available)

```
[...]

PR #5412, opened 2023-12-12
Proposed fix for 5411

diarmidmackenzie:
**Description:**

See #5411

**Changes proposed:**

In VR mode, when no mouse is available, prevent `cursor` from firing a  "mouseenter" event, as follows:

- on entering VR, set a flag this.mouseAvailable to false
- on detection of a mouse move or click event, set this.mouseAvailable to true
- use this.mouseAvailable to determine whether ot not to so an early return on detection of a raycaster intersection.

This is intended to allow use of a mouse in "VR mode" for e.g. cases such as desktop full screen.

This PR is still a draft.  I haven't yet:
- run UTs
- added UTs
- done any live testing

If a fix along these lines is wanted, I'll take care of the above, and submit for review.  For now awaiting feedback either here or in #5411.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -66,6 +66,7 @@ module.exports.Component = registerComponent('cursor', {
     this.canvasBounds = document.body.getBoundingClientRect();
     this.isCursorDown = false;
     this.activeXRInput = null;
+    this.mouseAvailable = true;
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Closing this for now. I could not reproduce the issue

PR #5414, opened 2023-12-15
Fix bug in handling processSound callback when sound hadn't loaded yet

mrxz:
**Description:**
The changes made in #5365 can result in an error (see https://github.com/aframevr/aframe/pull/5365#issuecomment-1856510842). The code incorrectly uses `this` instead of `self`. This PR fixes it and adds a unit test covering this as well.

**Changes proposed:**
- Replace the incorrect `this` with `self`
- Introduce a unit test that verifies that the `processSound` callback given to `playSound` is correctly invoked upon the sound loading.


dmarcos:
Thank you very much!

PR #5415, opened 2023-12-17
Fix bug when loading models on child entities of controllers

mrxz:
**Description:**
The various controls components handle the `model-loaded` event to configure the controller models. However, they don't check where the `model-loaded` event originates from. In case a child entity loads a model, the controls components would incorrectly assume it's the controller model that just loaded in.

This actually causes issues in A-Painter where the `model-loaded` event of the brush-tip would trigger the `onModelLoaded` for the `oculus-touch-controls` (and others).

**Changes proposed:**
- Include a guard in `onModelLoaded` checking that the event didn't originate from any child element


dmarcos:
Thanks!

PR #5417, opened 2023-12-19
Set `.background` null when removing `background` component

mrxz:
**Description:**
The `background` component set the scene background to the string value `black` when removed. Not only is this an invalid value for the background property in Three.js, it also doesn't match the default value from before adding the `background` component.

**Changes proposed:**
- Set `sceneEl.object3D.background` to `null` when removing `background` component


dmarcos:
Thanks!

PR #5418, opened 2023-12-19
Fix fog removal from scene

mrxz:
**Description:**
When removing the `fog` component, the fog wouldn't actually be removed but instead rendered "ineffective" by adjusting the `near` and `far` values. This however has two issues:

- Fog type `exponential` doesn't use `near` and `far` meaning the fog would still be visible after removal
- Shaders would still perform fog calculations for no reason

**Changes proposed:**
- Instead of adjusting fog when removing, outright set it to `null` ([Three.js default value](https://threejs.org/docs/index.html?q=scene#api/en/scenes/Scene.fog))
- Remove `updateMaterials()` method as it was only used by the `fog` component and is not needed since https://github.com/mrdoob/three.js/pull/20135
- Update unit tests to match

### DIFF
--- a/src/components/scene/fog.js
+++ b/src/components/scene/fog.js
@@ -30,7 +30,6 @@ module.exports.Component = register('fog', {
     // (Re)create fog if fog doesn't exist or fog type changed.
     if (!fog || data.type !== fog.name) {
       el.object3D.fog = getFog(data);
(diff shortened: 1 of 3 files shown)


dmarcos:
Thank you!

Issue #4038, opened 2019-02-24
importing Aframe breaks build


wjessup:
using the latest dist at https://github.com/aframevr/aframe/tree/master/dist

whenever I try to use es6 modules to import aframe:

`import './aframe.min.js'`

I get this error:

`aframe.min.js:19685 Uncaught TypeError: systems[name] is not a constructor
    at HTMLElement.value (aframe.min.js:19685)
    at HTMLElement.value (aframe.min.js:19670)
    at HTMLElement.value (aframe.min.js:19603)
    at HTMLElement.wrappedMethod (aframe.min.js:18416)
    at module.exports.registerElement (aframe.min.js:18325)
    at Object.104.../../lib/three (aframe.min.js:19538)
    at o (aframe.min.js:1)
    at aframe.min.js:1
    at Object.147.../package (aframe.min.js:22136)
    at o (aframe.min.js:1)`

wjessup:
simple jsfiddle w/ the `type="module"` error. https://jsfiddle.net/wdzsbgvL/

bknill:
This is working in the fiddle - did you find a solution? I'm having the same problem.

dakotaJang:
Was ES6 module working before?
Seems fine in Chrome but when I switch to Firefox, Edge, or Safari, I get `not a constructor` error.
It would be nice if I can bundle my code in one module file that imports the A-frame and my src code.

subelsky:
In case this is helpful: I got this error when using the `defer` attribute on a `<script>` tag including Aframe. Went away when I took it off. The same issue could occur if using the `async` attribute...

mattrossman:
Scripts that have `type="module"` are automatically assigned the defer property, so they will essentially run at the end of `<body>`. However, A-Frame requires that the script is included *before* the body/scene.

I'm not sure why this is the case, as [according to this](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements), "the script that registers [custom components] has to be loaded after the DOM is parsed".

Related: #4242.

Type1J:
[...]

PR #5419, opened 2023-12-21
Allow async loading of aframe library and interoperation with ES module libraries

chrisirhc:
**Description:**
This allows interop with ES module libraries such as webxr.
ES modules are a part of the Web Component spec but it's a bit surprising that aframe doesn't play well with other Web Components that rely on ES modules.

It allows aframe to be loaded as a module or deferred on page load.
Currently, aframe needs to be loaded synchronously. Adding `async` or `defer` or `type="module"` can cause `aframe` library to fail in many different ways such as:
* `a-scene` is registered and connectedCallback hook does not find the camera component as it hasn't been registered
* `a-entity`'s connectedCallback hook unable to find that `a-scene` is initialized

These proposed changes allow the following code snippet to run:
Note that `new LookingGlassWebXRPolyfill()` needs to run before importing aframe in order to polyfill/override webxr for looking glass capabilities.

```html
    <script type="module">
      import { LookingGlassWebXRPolyfill } from "@lookingglass/webxr"
      new LookingGlassWebXRPolyfill()
      window.AFRAME_ASYNC = true;
      await import("aframe")
      // add any components
      await import("https://quadjr.github.io/aframe-gaussian-splatting/index.js")
      window.AFRAME.ready();
    </script>
```

Example of this working is at https://chrisirhc.github.io/looking-glass-viewer-experiments/aframe.html .

**Changes proposed:**
- Delay running `doConnectedCallback` until `ready` is called if `window.AFRAME_ASYNC` is set to a truthy value
- Reorder the initialization in index.js so that default components and system are registered before connectedCallback is called.

This is a strawman implementation. I'm not sure how to name these API surfaces and open to suggestions.

**Alternatives considered**
- Change the order or delay order in which the customElements are registered.
  - Not that this doesn't work well because a-scene element creates a-entity elements during its connectedCallback lifecycle hook, so if a-entity isn't already registered, then

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -11,6 +11,17 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
 var OBJECT3D_COMPONENTS = ['position', 'rotation', 'scale', 'visible'];
 var ONCE = {once: true};
 
+var _resolveReadyPromise;
+var isReady = false;
+var isReadyPromise = new Promise(function (resolve) {
+  _resolveReadyPromise = resolve;
+});
+
+function ready () {
+  isReady = true;
+  _resolveReadyPromise();
+}
+
 /**
  * Entity is a container object that components are plugged into to comprise everything in
  * the scene. In A-Frame, they inherently have position, rotation, and scale.
@@ -67,6 +78,12 @@ class AEntity extends ANode {
       return;
     }
 
+    if (!isReady) {
+      isReadyPromise.then(
+        AEntity.prototype.doConnectedCallback.bind(this)
+      );
+      return;
+    }
     AEntity.prototype.doConnectedCallback.call(this);
   }
 
@@ -868,6 +885,7 @@ function getRotation (entityEl) {
 
 AEntity.componentsUpdated = [];
 AEntity.singlePropUpdate = {};
+AEntity.ready = ready;
 
 customElements.define('a-entity', AEntity);
 
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -191,7 +191,6 @@ class AScene extends AEntity {
    */
   initSystem (name) {
     if (this.systems[name]) { return; }
-    if (!systems[name]) { return; }
     this.systems[name] = new systems[name](this);
     this.systemNames.push(name);
   }
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -191,6 +191,7 @@ class AScene extends AEntity {
    */
   initSystem (name) {
     if (this.systems[name]) { return; }
+    if (!systems[name]) { return; }
     this.systems[name] = new systems[name](this);
     this.systemNames.push(name);
   }
--- a/src/index.js
+++ b/src/index.js
@@ -62,8 +62,6 @@ if (utils.device.isBrowserEnvironment) {
   require('./style/rStats.css');
 }
 
-// Required before `AEntity` so that all components are registered.
-var AScene = require('./core/scene/a-scene').AScene;
 var components = require('./core/component').components;
 var registerComponent = require('./core/component').registerComponent;
 var registerGeometry = require('./core/geometry').registerGeometry;
@@ -81,8 +79,16 @@ require('./components/index'); // Register standard components.
 require('./geometries/index'); // Register standard geometries.
 require('./shaders/index'); // Register standard shaders.
 require('./systems/index'); // Register standard systems.
+// Required before `AEntity` so that all components are registered.
+var AScene = require('./core/scene/a-scene').AScene;
 var ANode = require('./core/a-node').ANode;
 var AEntity = require('./core/a-entity').AEntity; // Depends on ANode and core components.
+setTimeout(function () {
+  if (window.AFRAME_ASYNC) {
+    return;
+  }
+  AEntity.ready();
+}, 0);
 
 require('./core/a-assets');
 require('./core/a-cubemap');
@@ -106,6 +112,7 @@ module.exports = window.AFRAME = {
   components: components,
   coreComponents: Object.keys(components),
   geometries: require('./core/geometry').geometries,
+  ready: AEntity.ready.bind(AEntity),
   registerComponent: registerComponent,
   registerGeometry: registerGeometry,
   registerPrimitive: registerPrimitive,


src/core/a-entity.js
```diff
@@ -11,6 +11,17 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
 var OBJECT3D_COMPONENTS = ['position', 'rotation', 'scale', 'visible'];
 var ONCE = {once: true};
 
+var _resolveReadyPromise;
```
dmarcos, line 14: Can name just `resolveReadyPromise`. We don't use ampersand

vincentfretin:
The changes looks ok. I don't really have better.
Do we need a similar fix in `src/core/scene/a-scene.js` to delay executing of `this.initSystems()` if you imported a module that registered a system?

I guess this PR fixes this old issue #4038

src/core/scene/a-scene.js
```diff
@@ -191,6 +191,7 @@ class AScene extends AEntity {
    */
   initSystem (name) {
     if (this.systems[name]) { return; }
+    if (!systems[name]) { return; }
```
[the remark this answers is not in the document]
vincentfretin: Same question. Probably a mistake. You probably added this while testing when the camera system wasn't registered yet.

dmarcos:
Just curious how a complete example would look like?

`await import("aframe")` I imagine 'aframe' has to defined somehow previously.

I'm still getting familiar with ES6 modules.

vincentfretin:
You need to define the importmap. In his linked example, this is defined:
```html
    <script type="importmap">
      {
          "imports": {
              "@lookingglass/webxr": "./lib/webxr.module.js",
              "aframe": "./lib/aframe-master.js"
          }
      }
    </script>
```

You can replace the path by a jsdelivr url.

dmarcos:
Wouldn't be better to tie the promise to A-Frame loading as a whole vs when the first entity connects? For example here: https://github.com/aframevr/aframe/blob/master/src/index.js#L95

src/core/scene/a-scene.js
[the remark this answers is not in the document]
chrisirhc: Yep, will remove. Added this while testing.

chrisirhc:
> Just curious how a complete example would look like?
> 
> `await import("aframe")` I imagine 'aframe' has to defined somehow previously.

It'd look like the full JS url:
```javascript
await import("https://aframe.io/releases/1.5.0/aframe.min.js")
```

The `import()` and `await` statements are to ensure the side effects of the modules are run in the order in the code. See 
[reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#:~:text=When%20the%20module%20being%20imported%20has%20side%20effects%2C%20and%20you%20do%20not%20want%20those%20side%20effects%20unless%20some%20condition%20is%20true.%20(It%20is%20recommended%20not%20to%20have%20any%20side%20effects%20in%20a%20module%2C%20but%20you%20sometimes%20cannot%20control%20this%20in%20your%20module%20dependencies.)) .

> Wouldn't be better to tie the promise to A-Frame loading as a whole vs when the first entity connects? For example here: https://github.com/aframevr/aframe/blob/master/src/index.js#L95

Ah, the change I'm proposing is to run the entity connectedCallback after A-Frame loading has completed.
The `setTimeout` achieves this since it waits until all the synchronous loading/execution has completed.

That said, the problem it works around is that the connectedCallback expect that the A-Frame loaded has completed, but these connectedCallback are being run during registerElement because the DOM is already available during the async loading.

Typically, the steps look like:
1. Load script synchronously via  `<script src=https://aframe.io/releases/1.5.0/aframe.min.js></script>` in the head. This includes running `customElement.register` but just registers the element and doesn't call the callbacks. 
2. Parse the DOM including the body that contains `<a-scene>` and other elements. During the parse, constructor and callbacks (connectedCallback) are called for each element as they're encountered.

However, in the async/module flow, the steps look like:
1. Parse the DOM, including the body that contains `<a-scene>` and other elements. Does *not* call custom element constuctors because the module hasn't loaded yet.
2. Then while loading + executing the aframe script:
3. As it encounters the line `customElement.register('a-scene', …)` code, it runs the constructors and connectedCallback. 
4. Continues to execute the rest of the aframe code.

The problem is that in Step 3, not all of the aframe code has executed. So things like the camera weren't registered. This is only a problem because the connectedCallback was called during parsing/execution of the aframe script.
It's not a problem if all of the aframe script has already executed.

Hope this provides an explanation of what's going on.
This is what I gather was happening as I was working through this issue.

mrxz:
There is already logic in place to defer the initialization till the `document.readyState` is complete. I feel it would be cleaner to generalize this logic instead of adding additional flows. Instead of listening for the `readystatechange` it could listen to a dedicated event, which by default would still be based on the ready state.

As for the magic `AFRAME_ASYNC` variable, I don't really like it as it's easily something someone can forget or get wrong. Perhaps it's possible for A-Frame to detect it's being loaded asynchronously and automatically switch to expecting a ready signal from the user?
Ideally there would then be a warning logged to the console if this signal isn't given, but not sure if there is a logical moment to detect this apart from a timeout. Likewise an error if the ready signal is given more than once, or if components/systems are registered _after_ the signal has been given.

> Example of this working is at https://chrisirhc.github.io/looking-glass-viewer-experiments/aframe.html .

This example doesn't work in Firefox for me, but does on Chrome. Didn't dive into the cause, but it seems there is some difference in execution order between the two.

As an aside, the way I generally workaround the issue is by placing the scene in a template and adding it to the DOM after loading and registering all dependencies:
```HTML
<html>
    <head>
        <script type="importmap">
            (...)
        </script>
        <script type="module">
            await import("aframe");

            // Load dependencies

            const sceneTemplate = document.getElementById('scene');
            document.body.appendChild(sceneTemplate.content)
        </script>
    </head>
    <body>
        <template id="scene">
            <a-scene>
                <a-box id="box" color="red" position="0 1.6 -2"></a-box>
            </a-scene>
        </template>
    </body>
</html>
```

src/core/a-entity.js
```diff
@@ -11,6 +11,17 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
...
+var isReady = false;
+var isReadyPromise = new Promise(function (resolve) {
+  _resolveReadyPromise = resolve;
+});
+
+function ready () {
+  isReady = true;
+  _resolveReadyPromise();
+}
```
chrisirhc, line 23: Looking at this again, it might make sense for this to live in its own file/module that just governs whether the entire aframe library is ready. I can tidy this up.

This initial PR was only to gather and understand if there's interest/appetite for such a change.

dmarcos: sounds good. thanks. In an ideal world we would have a single build that can be loaded via modules (export, import) or script tag. As far as I know that's not possible, right?

dmarcos:
> As an aside, the way I generally workaround the issue is by placing the scene in a template and adding it to the DOM after loading and registering all dependencies:

Nice workaround. Hopefully we can find a way for A-Frame to work seamlessly with both modules and script tags. I would like to also avoid two separate builds (aframe.js, aframe.module.js) but not sure if possible. Biggest downside of ES6 modules is that how it seems to split JS space in two: pre and post modules. Also I think there's still valuable simplicity in being able to load JS with just a <script> tag without additional incantations.

arpu:
not sure where the problem is  at the moment what still works is 
https://glitch.com/edit/#!/ink-zealous-pudding

arpu:
i think a good example is model-viewer https://modelviewer.dev/

dmarcos:
Let me know if you need help or any questions to move this along. Thanks!

vincentfretin:
@arpu In your glitch example you don't async load any new component, so of course you don't have any issue.
I didn't understand the reference to modelviewer, this is a module js yes, just a single web component. For aframe the issue is component on a web component, the component is not yet loaded when the web component is initialized.

@chrisirhc on my comment https://github.com/aframevr/aframe/pull/5419#issuecomment-1866578435 need to test to async load a system.

arpu:
Thanks @vincentfretin  you are right not trying to load second componet, and now i see the problem 
for the modelviewer refrence was more for @dmarcos  to see how easy it is today with es module and script tag

chrisirhc:
> @chrisirhc on my comment https://github.com/aframevr/aframe/pull/5419#issuecomment-1866578435 need to test to async load a system.

Hi, do you mind sharing an example on how I can do this?

vincentfretin:
You can try loading the uvscroll.js as a module, that includes a uv-scroll system and uv-scroll component, see https://github.com/c-frame/sponsorship/issues/15

dmarcos:
There are some pending changes here

vincentfretin:
I created a glitch with the same example as @chrisirhc and the uv-scroll system/component, it's working properly as far as I can tell.
https://glitch.com/edit/#!/vfretin-async-aframe

dmarcos:
@chrisirhc There are few pending changes from the review (e.g underscore on name variables). Do you want to finish it up? Thanks so much!

chrisirhc:
Hi there, apologies for the radio silence. 
I likely won't be able to get to this any time soon.

I was stuck on a blurness issue ( https://github.com/mkkellogg/GaussianSplats3D/issues/65#issuecomment-1905118603 ) after finding out that aframe using a pretty old/forked version of ThreeJS so I've had to look elsewhere despite hacking this fix together.

I might not get back to this in the near future. Feel free to close this out.

dmarcos:
@chrisirhc FYI A-Frame master is on latest THREE r162. Can use builds linked in commits https://github.com/aframevr/aframe/commits/master/

vincentfretin:
Also you can close this PR now that #5481 is merged.

chrisirhc:
Yep, thanks @mrxz for getting the change to the finish line!
I'll be able to test things out with the main branch now, with the above suggestions. 👍

PR #5424, opened 2024-01-06
Various fixes in example

mrxz:
**Description:**
Found several small mistakes in the examples.

**Changes proposed:**
- Update usages of `vr-mode-ui` to `xr-mode-ui`
- Omit `crossorigin="anonymous"` on `<a-asset-item>` tags as this has no effect
- Remove components/attributes that no longer exist or don't do anything (`auto-enter-vr` and `opacity` on `<a-entity>`)
- Change `filter` to `filterLabels` as the former doesn't exist on the `real-world-meshing` component
- Fix typo accidentally using `,` as property delimiter instead of `;`

### DIFF
--- a/examples/boilerplate/webxr-ar-lighting/index.html
+++ b/examples/boilerplate/webxr-ar-lighting/index.html
@@ -3,7 +3,7 @@
 <head>
 	<script src="../../../dist/aframe-master.js"></script>
 	<script src="ar-shadow-helper.js"></script>
-	<script src="https://unpkg.com/aframe-environment-component@1.3.3/dist/aframe-environment-component.min.js"></script>
+	<script src="https://unpkg.com/aframe-environment-component@1.3.4/dist/aframe-environment-component.min.js"></script>
 	<title>AFRame Total WebXR AR Demo</title>
 	<style>
 		body {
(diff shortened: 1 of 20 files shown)


examples/boilerplate/webxr-ar-lighting/index.html
```diff
@@ -1,10 +1,9 @@
 <html>
 
 <head>
-	<script src="../../../build/aframe.js"></script>
-	<!-- <script src="../../../dist/aframe-master.js"></script> -->
+	<script src="../../../dist/aframe-master.js"></script>
 	<script src="ar-shadow-helper.js"></script>
-	<script src="https://unpkg.com/aframe-environment-component@1.3.1/dist/aframe-environment-component.min.js"></script>
```
dmarcos, line 7: we now have 1.3.4

dmarcos:
Thanks so much. just a small change

mrxz:
Didn't know a new version of the environment component was already out. I took the liberty to update all instances in the examples as it's needed for proper colour management. Did end up removing the script from the `mixed-reality/anchor` example as it didn't use the environment in any way.

dmarcos:
Thanks!

PR #5426, opened 2024-01-07
Fix bug when checking schema properties for schemaChange

mrxz:
**Description:**
Components that have one or more properties marked as `schemaChange` will check if any of these properties change to trigger an `updateSchema`. While checking the properties in an update, it incorrectly assumes each property to also exist, which might not be the case.

Reproduction (https://glitch.com/edit/#!/beaded-basalt-church?path=index.html):
```HTML
<script>
    AFRAME.registerComponent('bug', {
      init: function() {
        this.el.setAttribute('material', {
            invalidProperty: 'causes an error',
        })
      }
    })
</script>
(...)
<a-scene>
    <a-entity material bug></a-entity>
</a-scene>
```

**Changes proposed:**
- Ensure a property is part of the schema before checking its `schemaChange` flag

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -364,7 +364,7 @@ Component.prototype = {
     // Check if we need to update schema.
     if (this.schemaChangeKeys.length) {
       for (key in attrValue) {
-        if (this.schema[key].schemaChange) {
+        if (key in this.schema && this.schema[key].schemaChange) {
           mayNeedSchemaUpdate = true;
           break;
         }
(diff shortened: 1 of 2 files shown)


dmarcos:
Is it possible to cover the fix with a test?

mrxz:
Added a couple unit tests covering schemaChange/udpateSchema (there were non before) including one covering the fixed behaviour.

dmarcos:
Thanks!

vincentfretin:
@mrxz There is a related error in the aframe-inspector demo page. I didn't have time to investigate but it's exactly on the same line you just changed.
https://github.com/aframevr/aframe-inspector/issues/701

PR #5429, opened 2024-01-12
Stop the light probe only if this was set up in the first place

Elettrotecnica:
Following #5384 and https://github.com/aframevr/aframe/pull/5384#issuecomment-1812849138 in particular

In this proposed change we stop the light probe only when this has been set up in the first place.

All the best

### DIFF
--- a/src/components/scene/reflection.js
+++ b/src/components/scene/reflection.js
@@ -51,7 +51,9 @@ module.exports.Component = register('reflection', {
     });
 
     this.el.addEventListener('exit-vr', function () {
-      self.stopLightProbe();
+      if (self.xrLightProbe) {
+        self.stopLightProbe();
+      }
     });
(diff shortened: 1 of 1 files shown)


src/components/scene/reflection.js
```diff
@@ -51,7 +51,9 @@ module.exports.Component = register('reflection', {
     });
 
     this.el.addEventListener('exit-vr', function () {
-      self.stopLightProbe();
+      if (self.xrLightProbe) {
```
dmarcos, line 54: can do one liner
```
if (self.xrLightProbe) { self.stopLightProbe(); }
```

Elettrotecnica: Of course, done

dmarcos:
Thank you!

PR #5430, opened 2024-01-13
Make sure that hasLoaded flag is only true when the loaded event has fired. Alternative implementation of commit 924dc00 (fix #5425)

### LINKED ISSUE #5425: Regression in component initialization
**Description:**
Commit 924dc00ecf483c1cd21c01bcadaf53d95911155d changed the initialization behaviour for components. It postpones setting `hasLoaded` on an entity after its components have initialized. However, this introduces a state where `hasLoaded` is false, but the component _is_ supposed to initialize. The updated guard `if(!el.hasLoaded && !el.sceneEl)` ([component.js#L279-L281](https://github.com/aframevr/aframe/blob/master/src/core/component.js#L279-L281)) can't distinguish between a legitimate initialization and one triggered by a `setAttribute` (e.g. one originating from the init or update of another component). 

This can result in component initialization taking place before it should, followed by an update from `updateComponents`. In case of primitives this also means the initialization might lack properties.

Glitch demonstrating an issue arising from this change in behaviour: https://glitch.com/edit/#!/mint-available-newsboy?path=index.html
(The error itself is a different bug, but note that it surfaced due to a change in behaviour caused by the mentioned commit)

Personally I'm not entirely sure what issue the commit in question tries to resolve. The old behaviour was clearly intended, as is also evident from the comment in component.js which still indicates that component initialization _follows_ after the entity has loaded. Components and experiences will already have been written assuming this behaviour. For cases where a component depends on other components on the same entity, the `dependencies` array can be used or the component can ensure it's dependencies are met like how the various controls handle it (e.g. `oculus-touch-controls` -> `tracked-controls` -> `tracked-controls-webvr/xr`)

- A-Frame Version: master
- Platform / Device: any
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/mint-available-newsboy?path=index.html

dmarcos:


dmarcos:
@mrxz this should cover all the issues of https://github.com/aframevr/aframe/issues/5425

I introduced an `isLoading` flag different that `hasLoaded` (only set before 'loaded' fires) and also a `loaded-private` (don't love the name) event analog to `loaded` but "private" so A-Frame can handle internal stuff before the public event fires.

dmarcos:
Glitch with build for convenience:

https://glitch.com/edit/#!/lemon-magnificent-badge?path=index.html%3A1%3A0

src/core/component.js
```diff
@@ -278,7 +278,7 @@ Component.prototype = {
 
     // Just cache the attribute if the entity has not loaded
     // Components are not initialized until the entity has loaded
```
mrxz, line 280: Perhaps update this comment to something like "... until the entity starts loading"

dmarcos:
@mrxz Thanks for testing it. Just one request. I copied your glitch test case from https://github.com/aframevr/aframe/issues/5425#issuecomment-1881475628 verbatim to the tests include in this PR. They are a tad complicated / convoluted and will be hard to understand and maintain. Do you think you can simplify them a bit? Thanks

dmarcos:
@mrxz just ping in case you have some bandwidth to simplify your test cases so I can incorporate them to the unit tests. Thanks so much

mrxz:
@dmarcos Gave it a shot. The test cases included visual output, but as unit tests that's not relevant. However they do all rely on specific order of execution (things taking place _during_ `isLoading`), so I'm afraid they still remain somewhat complex.

See my attempt at simplifying them: https://github.com/mrxz/aframe/commit/2ff8fac273b8eeed8e42f24e470eb1ad148920fc

In summary:

- The double `update` test now uses the (built-in) `visible` component, as the erroneous behaviour occurs with any component, so no need to create a test-specific one
- Remove `geometry` from the mixin test and changed test name to align with surrounding tests
- Simplified primitive and updated naming to better describe what the component does.

dmarcos:
Thanks. I cherry picked your commit

PR #5435, opened 2024-01-22
Fix for Wolvic + Pico4 controllers

IamTonio:
**Description:**
Pico4 controllers works well in default browser but in Wolvic they were visible but triggers were not working. 

**Changes proposed:**
- Updated laser-controls config


dmarcos:
Thanks!

Issue #5413, opened 2023-12-14
Reconsider default `maxCanvasWidth` and `maxCanvasHeight` values


mrxz:
**Description:**
By default the renderer resolution is capped at 1920x1920. This was introduced back in 2018 with https://github.com/aframevr/aframe/pull/3641. There are a couple of downsides to this:

- Screen resolutions have gone up, making it likely that this cap is hit in everyday configurations
- Exceeding the cap might go unnoticed, or the resulting blurriness can be difficult to attribute to this cap
- The width and height take `devicePixelRatio` into account, which is inconsistent with Three.js ("device pixels" vs "CSS pixels")

I would propose changing the default values to `-1` making it uncapped by default. Users can always opt-in to capping the size if they have a specific need for it.
Otherwise it wouldn't hurt log a warning in the console once the cap is exceeded and possible increase the default cap.

- A-Frame Version: 1.5.0
- Platform / Device: all

dmarcos:
Yeah we can revisit the values. In general not super fan of using -1 values to indicate no value. Wonder if it makes sense to keep the limit at all. I personally don't use it and even forgot that it exist :)

mrxz:
Well, only suggesting -1 since that's already how it works ([docs](https://aframe.io/docs/1.4.0/components/renderer.html#properties)):
https://github.com/aframevr/aframe/blob/17f06f6ec71ea914aa889a38b2a0935c0b7d4571/src/core/scene/a-scene.js#L783

Personally I think the limit can be removed entirely. Most people probably don't expect this cap and only learn about it when they or one of their users runs into it. Not to mention that it conflates the canvas size with the render resolution. It's a lot cleaner to have the _size_ be the canvas element size in CSS pixels and use the pixelRatio to adjust the effective render resolutions. That's how I see it done in a lot of Three.js apps nowadays.

That also makes it easier in case users _do_ want to limit it. They can either limit the actual canvas size (when 'embedded') or adjust the renderer's pixelRatio.

If agreed I can create a PR that removes the logic (thus changing the default to not cap the size).

dmarcos:
fixed

PR #5440, opened 2024-01-26
Change default maxCanvasWidth/Height to -1 (no max)

mrxz:
**Description:**
As mentioned in #5413 the current default max width/height of 1920 is counter-intuitive. Most users are likely unaware of this default limit and they (or their users) will likely run into it once they notices a blurry and/or pixelated output on high resolution or oddly dimensioned displays. Especially with >1920x1080 resolutions becoming commonplace the limit is better removed IMO.

This PR only changes the default, the mechanism for limiting the resolution is still in place. Ultimately the logic might be removed entirely, but in its place we should probably expose the renderer's `pixelRatio` option to allow users to reduce the render resolution.

**Changes proposed:**
- Change the default `maxCanvasWidth/Height` to -1


dmarcos:
Thanks!

PR #5442, opened 2024-02-01
setup canvas on first capture screenshot.js

arpu:
**Description:**
The  default Screenshot Component always creates a shader and canvas on init()
this PR change it on first usage of screenshot capture.

**Changes proposed:**
-set canvas on first capture usage

### DIFF
--- a/src/components/scene/screenshot.js
+++ b/src/components/scene/screenshot.js
@@ -77,7 +77,6 @@ module.exports.Component = registerComponent('screenshot', {
       self.ctx = self.canvas.getContext('2d');
       el.object3D.add(self.quad);
       self.onKeyDown = self.onKeyDown.bind(self);
-      self.setupDone = true;
     }  
   },
 
--- a/src/components/scene/screenshot.js
+++ b/src/components/scene/screenshot.js
@@ -49,34 +49,29 @@ module.exports.Component = registerComponent('screenshot', {
     height: {default: 2048},
     camera: {type: 'selector'}
   },
-
-  init: function () {
-  },
-
+  
   setup: function () {
     var el = this.el;
-    var self = this;
-    if (!self.canvas) {
-      var gl = el.renderer.getContext();
-      if (!gl) { return; }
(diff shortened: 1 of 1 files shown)


src/components/scene/screenshot.js
```diff
@@ -51,36 +51,32 @@ module.exports.Component = registerComponent('screenshot', {
   },
 
   init: function () {
+    this.setupDone = false;
```
dmarcos, line 54: don't need an extra flag. Can just check if there's a `self.canvas` in the setup function

src/components/scene/screenshot.js
```diff
@@ -178,6 +174,9 @@ module.exports.Component = registerComponent('screenshot', {
    * Maintained for backwards compatibility.
    */
   capture: function (projection) {
+    if (!this.setupDone) {
```
dmarcos, line 177: can just do `this.setup()` and check for `this.canvas` inside setup

src/components/scene/screenshot.js
```diff
@@ -51,16 +51,12 @@ module.exports.Component = registerComponent('screenshot', {
   },
 
   init: function () {
```
dmarcos, line 53: We can probably remove the init function now

src/components/scene/screenshot.js
```diff
@@ -178,6 +174,7 @@ module.exports.Component = registerComponent('screenshot', {
    * Maintained for backwards compatibility.
    */
   capture: function (projection) {
+    this.setup();
```
dmarcos, line 177: Can add this call after variables declaration

dmarcos:
Thank you!

mrxz:
@arpu This PR causes a regression when people use the `screenshot` component in the following way:

```JS
document.querySelector('a-scene').components.screenshot.getCanvas('perspective');
```
The `getCanvas` is part of its public API and documented in case the user doesn't want to automatically download the file, but do something else with it. See https://aframe.io/docs/1.5.0/components/screenshot.html#methods

Issue #5447, opened 2024-02-06
State of polyfills and browser support


mrxz:
**Description:**
There are several polyfills present in A-Frame, many of which are already quite dated. Several of them can probably be removed, but it mostly depends on the relevancy of the (old) browsers/devices that are still intended to be supported. AFAIK there is no set [browserlist](https://github.com/browserslist/browserslist), but that might be a good idea going forwards.

Here's an overview of the polyfills:
| Polyfill | API(s) | Caniuse % | Notes |
|-----------|----------|----------------|-----------|
| `@ungap/custom-elements` | Custom Elements |  [78.83% + 18.5% = 97.33% ](https://caniuse.com/custom-elementsv1) | only for IE (Safari doesn't support customized built-in elements, but A-Frame does not require this) |
| `custom-event-polyfill` | CustomEvent |  [98.05% + 0.49% = 98.54%](https://caniuse.com/customevent) | only for IE 9+ |
| `object-assign` | Object.assign | [96.61%](https://caniuse.com/mdn-javascript_builtins_object_assign) | ponyfill, only for IE |
| `present` | Performance.now | [97.22%](https://caniuse.com/mdn-api_performance_now) | :warning: Also polyfilled by `rStats.js` |
| `promise-polyfill` | Promise | [97.89%](https://caniuse.com/promises) | only for IE |
| `webvr-polyfill` | WebVR |  [2.92% + 2.49% = 5.41%](https://caniuse.com/webvr) | contrary to what caniuse reports, FF has it disabled by default and requires specific flag to enable |

Three.js has already dropped support for IE 11, so simply following suit, virtually all above mentioned polyfills can be removed as well. If a user does want to go through the pain of targetting IE 11, they can always load the needed polyfills themselves.

As for WebVR (polyfill), it would be nice to be able to get rid of it. There's quite a bit of duplicate code logic/paths due to supporting both WebVR and WebXR. I wouldn't be surprised if WebVR support is (subtly) broken in places. 

Thoughts?

danineuss:
I would like to bump this ticket (as I did its PR). I believe I am currently experiencing issues where my Firefox (beta) for Android mobile is breaking because of the PolyFill.

```
Failed to recalculate device parameters. webvr-polyfill.js:1945:12
    Aframe 7
        recalculateDeviceParams_
        ge
        Xe
        getPolyfillDisplays
        getVRDisplays
        c
        5579
    Webpack 5
        n
        <anonymous>
        <anonymous>
[...]

PR #5448, opened 2024-02-06
Remove various polyfills

mrxz:
**Description:**
This PR removes the polyfills that should only really impact IE11 (or other older browsers). See #5447 for context.

**Changes proposed:**
- Remove `@ungap/custom-elements`, `present`, `promise-polyfill` and `custom-event-polyfill`
- Remove `String.startsWith` polyfill
- Replace `object-assign` ponyfill with `Object.assign`


dmarcos:
@WebReflection Do you think the we should still keep the custom elements polyfill around? I'm sure he has a more complete picture than me.

WebReflection:
> @WebReflection Do you think the we should still keep the custom elements polyfill around? I'm sure he has a more complete picture than me.

If you don’t use builtin extends and you don’t use the argument in whenDefined resolved promise I think the time to drop that polyfill might be right but I’d rather check dependent projects as they might assume a-frame provides the polyfill and might trust those features so that it might break others.

dmarcos:
Thanks both for the work and the feedback 👏

WebReflection:
> It probably suffices to mention it in the release notes, along with instructions on including the polyfill manually for those affected by the change.

Absolutely, my comment was rather about the fact I *think* dropping polyfills *should* be considered a breaking change or nothing less than a minor release (but as it could break, maybe breaking change is more appropriate, semver speaking).

If you think the risk is minimal though and this library users won't be affected, a note is likely all you need 👍

Issue #5449, opened 2024-02-06
Refactor texture caching to share Sources between Textures


mrxz:
**Description:**
A-Frame currently caches Textures in order to prevent duplicate uploads. The cache takes both the source (e.g. url) and some texture properties into account. However it fails to correctly handle cases where these properties change after the fact. See #5441, #1372, #5120.

I started on refactoring parts of the texture handling to cache [`Source`](https://threejs.org/docs/index.html?q=Source#api/en/textures/Source) instead. Three.js can gracefully handle only uploading textures as needed when sharing one Source between multiple `Texture` instances. The changeset for this is becoming quite large, so it's probably best to split it up in several smaller PRs. What doesn't help is that there are places that sidestep the centralized texture handling, so I'll try to unify code-paths first.

This issue is to track progress:

- [x] Finally remove `sphericalEnvMap` as it's broken/incorrect for >3 years (https://github.com/aframevr/aframe/pull/4747, deprecation in #5454 )
- [x] Remove iOS 10 specific HLS workaround, as it now causes new issues (https://github.com/aframevr/aframe/issues/4134, https://github.com/aframevr/aframe/pull/5450)
- [x] Unify `updateMapMaterialFromData`, `updateMap` and `updateDistortionMap` (#5453)
- [x] Unify `envMap` loading in both `phong.js` and `standard.js` and let it re-use texture cache (#5454 )
- [ ] Cache `Source` instead of `Textures` (#5455 )

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5455

PR #5450, opened 2024-02-06
Remove iOS 10 specific HLS workaround

### LINKED ISSUE #4134: iOS 12.2 video HLS Wrong Color / Occasional black screen
Recently iOS version 12.2 was pushed to everyone and it seemed to break video playback of aframe.

Problem 1: Color not mapped correctly
iOS seemed fixed the color BRGA problem since iOS 11 for native HLS, [link](https://bugs.webkit.org/show_bug.cgi?id=179417) . However, aframe still apply the **ios10hls** shader for playback on iOS native hls. This is messing up the colors to a wrong one. Using standard "flat" shader should be good enough for color. 

Problem 2: Native HLS occasional black screen upon video Playback.
I have provided a sample code snippet. There is a certain chance such that canvas would display black screen. One finding is that when you change device orientation or enter / exit VR mode, the playback would become normal again. Could anyone advice how to fix this? Thanks!

Another thing that I found is that, some videos cannot play on iOS 12.2, like this one: https://github.com/mrdoob/three.js/raw/dev/examples/textures/sintel.mp4. While the example of aframe still works: https://aframe.io/examples/showcase/videosphere/. This is similar to Problem 2 that orientation change or enter / exit VR mode could put the playback back to normal. Could anyone advice? 

- A-Frame Version: 0.9.1
- Platform / Device: iOS 12.2 Safari

```
<html>
  <head>
    <script src="https://aframe.io/releases/0.9.1/aframe.min.js"></script>
    <style>
      #overlay {
        position: absolute;
        z-index: 1;
        top: 0;
        left: 0;
        width: 100%;
        height:100%;
        display: flex;
        align-items: center;
        justify-content: center;
        opacity: 1;
        background-color: #000000;
        color: #ffffff;
      }
      #overlay > div {
        text-align: center;
      }
      #overlay > div > button {
        height: 20px;
        width: 100px;
        background: transparent;
        color: #ffffff;
        outline: 1px solid #ffffff;
        border: 0px;
        cursor: pointer;
      }
      #overlay > div > p {
        color: #777777;
        font-size: 12px;
      }
    </style>
  </head>
  <body>
    <a-scene>
      <assets>
        <video id="video" loop crossorigin="anonymous" webkit-playsinline src="https://bitmovin.com/player-content/playhouse-vr/m3u8s/105560.m3u8">
        </video>
      </assets>
      <a-videosphere src="#video" rotation="0 -90 0"></a-videosphere>
    </a-scene>

    <div id="overlay">
      <div>
        <button id="startButton">Click to Play</button>
        <p>Automatic video playback with audio requires a user interaction.</p>
      </div>
    </div>

    <script>
      var overlay = document.getElementById('overlay');
      var startButton = document.getElementById( 'startButton' );
      var video = document.getElementById('video');
      startButton.addEventListener( 'click', function () {
        video.play();
        overlay.style = 'display:none';
      }, false );
    </script>

  </body>
</html>
```

mrxz:
**Description:**
Part of #5449, resolves #4134. It seems that the iOS 10 specific HSL is causing issues with iOS >12 versions. Basing on the [details](https://github.com/mrdoob/three.js/issues/9754#issuecomment-380659697) from the Three.js issue regarding iOS and HSL, it seems that iOS 10 needed this fix, iOS 11.0-11.2 don't work and iOS 12 and later should _not_ need it.

**Changes proposed:**
- Remove iOS 10 specific HLS workaround


dmarcos:
Thank you!

PR #5453, opened 2024-02-08
Unify texture loading and map property fixes

mrxz:
**Description:**
Preparation work for #5449. Both `utils.materials.updateMap` and `utils.materials.updateDistortionMap` are used for setting texture maps. These methods had similar, but slightly different implementations. This PR unifies them by letting `updateDistortionMap` call `updateMapMaterialFromData`.

Removing these maps or changing texture properties are now also correctly handle. Before only the diffuse map could be removed when changing the map properties of the `standard` and `phong` shader. In case texture properties are changed that require a texture update, the `needsUpdate` flag is now also set. Note: textures are still shared based on their **_initial_** properties, this will be addressed in upcoming PRs.

**Changes proposed:**
- Use `updateMapMaterialFromData` in `updateDistortionMap` aligning it with `updateMap`
- Always call texture map update methods to allow removing textures (`standard` and `phong` shaders)
- Let `setTextureProperties` set `needsUpdate` on the texture when a property changes that requires it
- Add unit tests for testing unsetting maps and changing texture property requiring `needsUpdate`

### DIFF
--- a/src/shaders/phong.js
+++ b/src/shaders/phong.js
@@ -53,19 +53,17 @@ module.exports.Shader = registerShader('phong', {
    */
   init: function (data) {
     this.materialData = { color: new THREE.Color(), specular: new THREE.Color(), emissive: new THREE.Color() };
-    this.textureSrc = null;
     getMaterialData(data, this.materialData);
     this.material = new THREE.MeshPhongMaterial(this.materialData);
-    utils.material.updateMap(this, data);
   },
 
   update: function (data) {
     this.updateMaterial(data);
     utils.material.updateMap(this, data);
-    if (data.normalMap) { utils.material.updateDistortionMap('normal', this, data); }
-    if (data.displacementMap) { utils.material.updateDistortionMap('displacement', this, data); }
-    if (data.ambientOcclusionMap) { utils.material.updateDistortionMap('ambientOcclusion', this, data); }
-    if (data.bump) { utils.material.updateDistortionMap('bump', this, data); }
+    utils.material.updateDistortionMap('normal', this, data);
+    utils.material.updateDistortionMap('displacement', this, data);
+    utils.material.updateDistortionMap('ambientOcclusion', this, data);
+    utils.material.updateDistortionMap('bump', this, data);
     this.updateEnvMap(data);
   },
 
--- a/src/shaders/standard.js
+++ b/src/shaders/standard.js
@@ -69,11 +69,11 @@ module.exports.Shader = registerShader('standard', {
   update: function (data) {
     this.updateMaterial(data);
     utils.material.updateMap(this, data);
-    if (data.normalMap) { utils.material.updateDistortionMap('normal', this, data); }
-    if (data.displacementMap) { utils.material.updateDistortionMap('displacement', this, data); }
-    if (data.ambientOcclusionMap) { utils.material.updateDistortionMap('ambientOcclusion', this, data); }
-    if (data.metalnessMap) { utils.material.updateDistortionMap('metalness', this, data); }
-    if (data.roughnessMap) { utils.material.updateDistortionMap('roughness', this, data); }
+    utils.material.updateDistortionMap('normal', this, data);
+    utils.material.updateDistortionMap('displacement', this, data);
+    utils.material.updateDistortionMap('ambientOcclusion', this, data);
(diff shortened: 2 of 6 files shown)


src/shaders/standard.js
```diff
@@ -69,11 +69,11 @@ module.exports.Shader = registerShader('standard', {
   update: function (data) {
     this.updateMaterial(data);
     utils.material.updateMap(this, data);
-    if (data.normalMap) { utils.material.updateDistortionMap('normal', this, data); }
```
dmarcos, line 72: This guards no longer needed?

src/utils/material.js
```diff
@@ -16,30 +16,43 @@ function setTextureProperties (texture, data) {
...
 
-  // Don't bother setting repeat if it is 1/1. Power-of-two is required to repeat.
+  // Set wrap mode to repeat only if repeat isn't 1/1. Power-of-two is required to repeat.
   if (repeat.x !== 1 || repeat.y !== 1) {
-    texture.wrapS = THREE.RepeatWrapping;
-    texture.wrapT = THREE.RepeatWrapping;
-    texture.repeat.set(repeat.x, repeat.y);
-  }
-  // Don't bother setting offset if it is 0/0.
-  if (offset.x !== 0 || offset.y !== 0) {
```
dmarcos, line 36: guard no longer needed?

src/shaders/phong.js
```diff
@@ -53,19 +53,17 @@ module.exports.Shader = registerShader('phong', {
...
-    this.textureSrc = null;
     getMaterialData(data, this.materialData);
     this.material = new THREE.MeshPhongMaterial(this.materialData);
-    utils.material.updateMap(this, data);
   },
 
   update: function (data) {
     this.updateMaterial(data);
     utils.material.updateMap(this, data);
-    if (data.normalMap) { utils.material.updateDistortionMap('normal', this, data); }
```
dmarcos, line 65: checks no needed?

src/utils/material.js
```diff
@@ -16,30 +16,43 @@ function setTextureProperties (texture, data) {
...
+    wrapT = THREE.RepeatWrapping;
   }
 
-  // Only set anisotropy if it isn't 0, which indicates that the default value should be used.
-  if (anisotropy !== 0) {
+  // Apply texture properties
+  texture.offset.set(offset.x, offset.y);
+  texture.repeat.set(repeat.x, repeat.y);
+
+  if (texture.anisotropy !== anisotropy) {
```
dmarcos, line 44: can we combine this statement and the one below?

```
   if (texture.wrapS !== wrapS || texture.wrapT !== wrapT ||
      texture.magFilter !== magFilter || texture.minFilter !== minFilter) {...}
```

src/shaders/standard.js
mrxz: The guards prevented these textures from being removed. Similar to how the main (diffuse) map is handled, the implementation checks if the texture is added, changed or removed, or if only the texture properties are updated.

src/shaders/phong.js
mrxz: Same as for `standard.js`

src/shaders/phong.js
```diff
@@ -53,19 +53,17 @@ module.exports.Shader = registerShader('phong', {
    */
   init: function (data) {
     this.materialData = { color: new THREE.Color(), specular: new THREE.Color(), emissive: new THREE.Color() };
-    this.textureSrc = null;
     getMaterialData(data, this.materialData);
     this.material = new THREE.MeshPhongMaterial(this.materialData);
-    utils.material.updateMap(this, data);
```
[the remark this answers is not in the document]
mrxz: Indeed, this is in the `init` method, which is followed by `update` which also calls `updateMap`. Both `flat` and `standard` let the update handle this and `phong` was the odd one out.

src/utils/material.js
mrxz: The guard actually prevented offset from being set back to 0/0 after being set to something else. On top of that, changing the offset doesn't require a texture update, so there's nothing gained by skipping this update.

src/utils/material.js
mrxz: Done

dmarcos:
Thanks!

PR #5454, opened 2024-02-09
Unify envmap textures and deprecate sphericalEnvMap

mrxz:
**Description:**
Follow-up of #5453, also preparation work for #5449. In A-Frame environment maps were handled separately in both the `standard` and `phong` shader. Besides the code-duplication this also meant that there was no texture re-use between them or other textures.

This PR moves the envMap logic into `utils/material` and `systems/material` in a similar vein to other textures. The eventual texture loads now also go through the texture cache. The possible values for `envMap` has been broadened to support both cubemaps and equirectangular maps. The latter allows a fallback for the currently incorrect behaviour of `sphericalEnvMap`. There is one small breaking change, the `envMap` value now behaves closer to `asset` property types, meaning it now only supports ID based selectors and would otherwise assume a URL in case it's unwrapped (e.g. breaking `envMap="#someElement > a-cubemap.active"`).

**Changes proposed:**
- Move loading of `envMap` to a common implementation in `utils/material` and `systems/material` making use of existing texture loading and caching.
- Expand the possible values for `envMap` to support both cubemap (six images/`<a-cubemap>`) as well as Equirectangular images.
- Deprecate `sphericalEnvMap` as Three.js dropped support for them long ago and A-Frame incorrectly uses them as Equirectangular images. Fallback is in place, logging a warning, but otherwise ensuring those expecting the equirectangular behaviour won't notice a regression.
- Added unit tests for `utils/src-loader` and `shaders/phong`

### DIFF
--- a/src/core/a-cubemap.js
+++ b/src/core/a-cubemap.js
@@ -4,7 +4,7 @@ var debug = require('../utils/debug');
 var warn = debug('core:cubemap:warn');
 
 /**
- * Cubemap element that handles validation and exposes list of URLs.
+ * Cubemap element that handles validation and exposes list of six image sources (URL or <img>).
  * Does not listen to updates.
  */
 class ACubeMap extends HTMLElement {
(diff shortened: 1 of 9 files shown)


dmarcos:
Thanks!

Issue #5441, opened 2024-01-30
Shared texture when multiple entities use the same video as material map


dmarcos:
If multiple entities share the same video as material map. Only one texture is created and shared across all materials. It's not possible to configure each material independently like for example setting different [texture offsets](https://aframe.io/docs/1.5.0/components/material.html#properties_offset).

We can change how this works in multiple ways. We could generate [separate hashes for each texture](https://github.com/aframevr/aframe/blob/master/src/systems/material.js#L256) and create multiple textures if the material data differs. this wouldn't work as expected if all materials sharing the same video map start with the same material values and then change. Changing the material values will result in changes on the shared texture.

We could completely forgo caching video textures. Shared video materials are unusual vs. shared texture images. This is the simplest solution. We would increase resource usage when a single video is used in different materials but not common in practice.

An example to play with:

https://glitch.com/edit/#!/zippy-bush-quill?path=index.html%3A43%3A37

Can see that if both entities share same video element cannot set texture offsets independently since they modify the same shared texture. 

This is something we have to address but not sure yet about the best solution.

mrxz:
Uploading the same texture multiple times should always be avoided, especially in the case of video textures. While the use-case of having a single video with various segments/regions might be uncommon, displaying the same video texture several times in a scene is probably a lot more common (e.g. TV screens).

The current texture handling and caching in A-Frame has quite a few problems. When using the same image it's easy to end up with multiple unnecessary copies/uploads (e.g. when only using different UV offsets). At the same time textures can get shared when their initial properties match, but won't get "unshared" when changing these properties on an instance. This can result in both unintended changes across all usages, or changes not taking effect (see #1372, #5120).

It's clear that the material system predates many of the improvements made in Three.js. Luckily this means the solution is relatively simple. Instead of caching Textures, A-Frame should focus on caching [Sources](https://github.com/mrdoob/three.js/pull/22846) and let Three.js determine when textures should be uploaded and copied.

dmarcos:
We definitively should revisit old design / assumptions

dmarcos:
@mrxz Is this fixed by https://github.com/aframevr/aframe/pull/5455 ?

mrxz:
@dmarcos Yes, textures can now share the same source, including video sources. :-)

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5455

Issue #5120, opened 2022-09-23
material: repeat doesn't work, when repeat has default or '1 1' set


Dirk-27:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: Any
- Reproducible Code Snippet or URL:

```
<!DOCTYPE html>
<html>
	<head>
		<meta charset="utf-8">
		<title>A-Frame - material texture repeat issue</title>
		<meta name="description" content="A-Frame">
		<script src="https://aframe.io/releases/1.3.0/aframe.min.js"></script>
	</head>
	<body>
		<a-scene>
			<a-plane src="https://cdn.vrdoro.com/data/vrdoro/images/11/1663758749743.jpeg" position="0 1.6 -1" rotation="0 0 0"></a-plane>
		</a-scene>
	</body>
</html>

```

Now do in console of browser inspector: `document.querySelector('a-plane').setAttribute('material', 'repeat:2 2');`

- Result: Texture is not repeated correct.
- Expected: Texture is repeated twice.
- Interesting: `document.querySelector('a-plane').getObject3D('mesh').material.map.wrapS;` (and wrapT) is 1001 but should be 1000 (=THREE.RepeatWrapping)

Workaround (choose one):
1. Set `material="repeat: 1.001 1.001"` as initial value
1. call `document.querySelector('a-plane').getObject3D('mesh').material.map.needsUpdate = true;` once

dmarcos:
FWIW, wrapS and wrapT values are correct. I think we need to call `needsUpdate` on the map. [Example on glitch](https://glitch.com/edit/#!/ubiquitous-absorbing-gem?path=index.html%3A11%3A86)

PR #5455, opened 2024-02-10
Cache Source instead of Texture instances in material system

mrxz:
**Description:**
Final part of #5449, resolving #5441, #5120, #1372. Instead of caching textures and trying to do reference counting, the underlying Sources are cached instead. Actual reference counting is done by Three.js internally, which allows the same texture source to be shared with multiple textures, even if they differ in `offset` or `repeat` (which before would result in uploads). 

This PR also improves the disposing behaviour, which previously only disposed when the material unregistered itself from the material system, missing any textures that were no longer assigned to that material. This also means that since Texture instances aren't shared, they can be changed by users without affecting other textures, while still using the same underlying image.

**Changes proposed:**
- Let `systems/material` cache `Source` instances instead of Textures
- Let helpers in `utils/material` request these `Source` instances and create/update the material's `Texture` instance

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -1,3 +1,4 @@
+/* global HTMLCanvasElement, HTMLImageElement, HTMLVideoElement */
 var THREE = require('../lib/three');
 var srcLoader = require('./src-loader');
 var debug = require('./debug');
@@ -14,7 +15,7 @@ var COLOR_MAPS = new Set([
  * Set texture properties such as repeat and offset.
  *
  * @param {object} data - With keys like `repeat`.
-*/
+ */
 function setTextureProperties (texture, data) {
   var offset = data.offset || {x: 0, y: 0};
   var repeat = data.repeat || {x: 1, y: 1};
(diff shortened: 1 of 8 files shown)


src/utils/material.js
```diff
@@ -97,24 +98,55 @@ module.exports.updateMapMaterialFromData = function (materialName, dataName, sha
...
-    if (texture && COLOR_MAPS.has(materialName)) {
-      rendererSystem.applyColorCorrection(texture);
+    // Nothing to do if texture is the same
+    if (material[materialName] === texture) {
+      return;
+    }
+
+    // Dispose old texture if present
+    if (material[materialName]) {
+      material[materialName].dispose();
```
mrxz, line 146: Before this PR, the texture wouldn't be disposed when switching, effectively 'leaking' them. This did have the benefit of keeping it on the GPU in case the texture would be used later on. For example when a user repeatedly switches between two textures.

With this change, upon disposal the texture will be freed from the GPU (unless it's still being used elsewhere). So when reintroducing the texture in the scene, an upload takes place. The repeatedly switching would then incur a texture upload every time it switches.

The new behaviour is definitely more correct since A-Frame has no other way to track the (intended) lifecycle of a texture. But we might need to provide users with ways to keep textures 'alive'. Do note, however, that it's about the specific "texture source + texture properties". One such way might be to keep materials alive, see #5457

dmarcos:
This need rebase

mrxz:
@dmarcos Rebased

dmarcos:
Thanks. So much good work.

PR #5458, opened 2024-02-11
Recycle nextData and previousOldData back into the objectPool upon component destroy

mrxz:
**Description:**
In the Component constructor four objects are requested from the object pool. Upon deletion only two of those were recycled back into the pool. This PR adds the missing two recycle calls, preventing them from leaking memory.

**Changes proposed:**
- Recycle `nextData` and `previousOldData` upon component destroy.


dmarcos:
Thanks!

PR #5459, opened 2024-02-11
Don't use object pool for initialOldData, and properly recycle tempObject

mrxz:
**Description:**
A temporary object was never released to the objectPool, causing a leak. After fixing this leak, some unit tests regressed. It turns out that the `initialOldData` provided to a component's `update(oldData)` is also leased from this pool. The pool however makes no guarantees about the state of the object. It might very well not be an empty object. (And due to the various leaks, it likely was an empty object)

This PR correctly clears and returns the `tempObject` to the pool and uses a constant frozen empty object for the `initialOldData`. This is a slight change in behaviour in case users mutate this object, but that was never correct and will now give an error.

**Changes proposed:**
- Correctly clear and recycle `tempObject` in Component update logic (`updateCachedAttrValue`)
- No longer use object pool for `initialOldData` for a Component's first `update` call

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -19,6 +19,7 @@ var upperCaseRegExp = new RegExp('[A-Z]+');
 
 // Object pools by component, created upon registration.
 var objectPools = {};
+var emptyInitialOldData = Object.freeze({});
 
 /**
  * Component class definition.
@@ -223,7 +224,7 @@ Component.prototype = {
     }
     utils.objectPool.clearObject(this.attrValue);
     this.attrValue = extendProperties(this.attrValue, newAttrValue, this.isObjectBased);
-    utils.objectPool.clearObject(tempObject);
+    this.objectPool.recycle(tempObject);
   },
 
   /**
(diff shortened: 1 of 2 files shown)


src/core/component.js
```diff
@@ -203,6 +204,7 @@ Component.prototype = {
       // If value is an object, copy it to our pooled newAttrValue object to use to update
       // the attrValue.
       tempObject = this.objectPool.use();
+      utils.objectPool.clearObject(tempObject);
```
dmarcos, line 207: So I understand. `use` function should already calls [clearObject](https://github.com/aframevr/aframe/blob/master/src/utils/object-pool.js#L26). Is not this redundant?

src/core/component.js
```diff
@@ -324,9 +326,8 @@ Component.prototype = {
 
     // For oldData, pass empty object to multiple-prop schemas or object single-prop schema.
     // Pass undefined to rest of types.
-    initialOldData = this.isObjectBased ? this.objectPool.use() : undefined;
+    initialOldData = this.isObjectBased ? emptyInitialOldData : undefined;
```
dmarcos, line 329: Same here. My understanding is that the `use` function calls `clearObject`. You mention 

> The pool however makes no guarantees about the state of the object. It might very well not be an empty object

What Am I missing?

src/core/component.js
mrxz: Oops, seems when cleaning up the PR I got `clearObject` and `removeUnusedKeys` mixed up in this place. The object pool does indeed "clear" the object, but the keys remains. Looking at the code again in more detail, it is actually safe here to have superfluous keys on `tempObject` in the subsequent `utils.extend` and `extendProperties` calls.

dmarcos:
Thanks!

PR #5461, opened 2024-02-12
Use native bind instead of custom implementation

mrxz:
**Description:**
Throughout the code base both the native `Function.prototype.bind` and a custom `AFRAME.utils.bind` are used (about 50-50). The latter was introduced as a faster alternative (#1762). However, it seems that this is no longer the case and in fact performs worse:

- https://github.com/vuejs/vue/issues/7408 (2018, showing native being faster in most browsers)
- https://jsperf.app/jquery-proxy-vs-custom-bind (shows native being the fastest)
- https://github.com/nathan7/fast-bind (reran benchmark on my machine)
  ```
  Function#bind (native) x 217,139,280 ops/sec ±0.16% (97 runs sampled)
  Function#bind (slice) x 8,253,466 ops/sec ±0.79% (94 runs sampled)
  Function#bind (concat) x 4,808,219 ops/sec ±0.46% (95 runs sampled)
  Function#bind (loop) x 9,885,830 ops/sec ±0.74% (93 runs sampled)
  ```

This PR removes it and introduces a small wrapper of the native `bind` in the place of `AFRAME.utils.bind` to avoid breaking any third party code using this utility, while still given them the same perf benefit from switching to native bind.

**Changes proposed:**
- Replace usages of `utils/bind` with native `bind`
- Introduce wrapper for `AFRAME.utils.bind` that uses native `bind`


dmarcos:
Thanks

PR #5462, opened 2024-02-13
Fix typos and spelling

mrxz:
**Description:**
Let a spellchecker run on the code and fixed various typos and spelling mistakes.

I did briefly test integrating [spellchecking in eslint](https://www.npmjs.com/package/@cspell/eslint-plugin), and while it works, the precommit hook becomes too slow and there are many false-positives (e.g. event names like `xbuttonup`). It might still be worthwhile to consider setting up some spellchecking tools/configuration along with an A-Frame specific wordlist.

**Changes proposed:**
- Fixed various typos and spelling mistakes

### DIFF
--- a/examples/mixed-reality/watch/hand-menu-button.js
+++ b/examples/mixed-reality/watch/hand-menu-button.js
@@ -96,7 +96,7 @@ AFRAME.registerComponent('hand-menu-button', {
 });
 
 /*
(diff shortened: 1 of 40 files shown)


dmarcos:
Thanks. Wouldn't add it as a hook maybe something optional that can be run manually.

dmarcos:
Thanks

PR #5465, opened 2024-02-15
Introduce sceneOnly flag for components

mrxz:
**Description:**
There are a couple of components intended to be used on `<a-scene>` exclusively. These reside in `src/components/scene`, but little prevents people from (accidentally) using them on other entities. Only the `fog` component performed a check that it was applied to the scene.

Instead of adding similar checks to all these components, this PR introduces a `sceneOnly` flag that components can set in their definition. Similar to the `multiple` flag this will throw when a `sceneOnly` component is being added to an entity that isn't the scene.

**Changes proposed:**
- Introduce `sceneOnly` flag in property definition
- Verify that `sceneOnly` components are only added to `<a-scene>` otherwise throw an Error

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -310,6 +310,11 @@ class AEntity extends ANode {
     // Initialize dependencies first
     this.initComponentDependencies(componentName);
 
+    // If component is sceneOnly check the entity is the scene element
+    if (!this.isScene && COMPONENTS[componentName].sceneOnly) {
+      throw new Error('Component `' + componentName + '` can only be applied to <a-scene>');
+    }
+
     // If component name has an id we check component type multiplicity.
     if (componentId && !COMPONENTS[componentName].multiple) {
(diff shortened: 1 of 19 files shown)


src/core/a-entity.js
```diff
@@ -310,6 +310,11 @@ class AEntity extends ANode {
     // Initialize dependencies first
     this.initComponentDependencies(componentName);
 
+    // If component is sceneOnly check the entity is the scene element
```
dmarcos, line 313: Wonder if it would be better to add this logic in the component constructor. Also the multiplicity check below. to keep things tidier here

mrxz: The component does have all the information needed to perform these checks, so I moved the checks. Inside `component.js` is now also the only place the `sceneOnly` and `multiple` flag are read, which is nice.

dmarcos:
Thank you!

PR #5466, opened 2024-02-15
Remove misspelled and deprecated getComputedAttrib

mrxz:
**Description:**
Encountered this misspelled method in `a-scene.js`. Since it was intended to provide a deprecation message to those using the old `getComputedAttribute` method, there's little use in fixing the spelling or keeping it around. It's highly unlikely that someone started using this misspelled variant (due to the deprecation warning it logs), so it should be safe to remove.

**Changes proposed:**
- Remove deprecated and misspelled `getComputedAttrib`


dmarcos:
Thanks!

PR #5467, opened 2024-02-15
Use AFRAME.utils.split in getComponentPropertyPath fixing caching behaviour

mrxz:
**Description:**
Noticed that that `AFRAME.utils.entity.getComponentPropertyPath` didn't properly use its cache. It constructed a cache and populated it, but never checked if strings were already cached. Meaning it would always recompute (= split), place in cache and _then_ immediately return that value from the cache.

This PR simply replaces the broken caching mechanism with `AFRAME.utils.split`, since the split utility also uses a cache internally. Added benefit is that the cache is now also shared between the two.

**Changes proposed:**
- Fix property path caching by using `AFRAME.utils.split` which internally caches

### DIFF
--- a/src/utils/entity.js
+++ b/src/utils/entity.js
@@ -1,27 +1,25 @@
+var split = require('./split').split;
+
 /**
  * Split a delimited component property string (e.g., `material.color`) to an object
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks!

Issue #5181, opened 2022-12-08
Vectors on schema with explciit properties are used by reference not by value.


diarmidmackenzie:
**Description:**

- A-Frame Version: 1.3.0
- Platform / Device: All 
- Reproducible Code Snippet or URL:  https://glitch.com/edit/#!/vector-by-reference-not-value?path=index.html%3A38%3A9

Slightly complex to explain.  Let's stat with this code.

```
          vec = new THREE.Vector3(-1, 0, -4)
          box = document.getElementById("box1")
          box.setAttribute("position", vec)
          vec.set(1, 0, -4)
          box.setAttribute("position", vec)
```

Where does the box end up?  At position 1, 0, 4 - as expected

But what about if we define a component like this:

```
      AFRAME.registerComponent("position-component", {
        schema: {
          position: { type: "vec3" }
        },
        update() {
          this.el.setAttribute("position", vec)
        }
      })
```

and use it in the exact same way...

```
          vec = new THREE.Vector3(-1, 0, -4)
          box = document.getElementById("box2")
          box.setAttribute("position-component", {position: vec})
          vec.set(1, 0, -4)
          box.setAttribute("position-component", {position: vec})
```

Now, the box ends up at position -1, 0, 4.... !

This happens for reasons descrbed in [this comment](https://github.com/aframevr/aframe/issues/5075#issuecomment-1186240023).

The root cause is [this code](https://github.com/aframevr/aframe/blob/2c4509aa1a9a3c447c08be9facfea0c7d6a9e156/src/core/component.js#L800-L803)

```
function isObjectOrArray (value) {
  return value && (value.constructor === Object || value.constructor === Array) &&
         !(value instanceof window.HTMLElement);
}
```
this function `isObjectOrArray` is used to determine whether to copy or deep clone an object. It does not classify Vector3s as objects (because value.constructor is set to `Vector3`, not `Object` or `Array`)

This means that when A-Frame copies data to `oldData`, it just assigns the `Vector3` objects by reference, rather than doing a deep copy.

In the case of the `position` attribute, which uses an unnamed single property in its schema, this doesn't happen.  When we come to that code, the top-level `vec3` object passed into `setAttribute()` has already been decomposed into x, y & z components.

But the problem does occur whenever a vec3 is used as a named property in a schema.

Examples of this causing problems with real components include:

- [line component](https://aframe.io/docs/1.3.0/components/line.html) (`start` & `end` properties)
[...]

PR #5474, opened 2024-02-26
Rework component update logic avoiding memory allocations

### LINKED ISSUE #3442: Default value returns `NaN` for numeric, single-property component schemas in 0.7.* and 0.8.0
**Default value returns `NaN` for numeric, single-property component schemas in A-Frame `0.7.*` and `0.8.0`.**

I noticed that single-property components may be deprecated in future releases according to #2536, however they don't currently function as documented: https://aframe.io/docs/0.8.0/core/component.html#single-property-schema

- A-Frame Version: `0.7.*` and `0.8.0`
- Platform / Device: Bug reproduced on all devices tested: Firefox, Chrome, Edge, Samsung Internet on Windows and Android.

```
AFRAME.registerComponent( 'test-component', {

  // Bug exists for types 'number' and 'int', possibly others.
  schema: { type: 'number', default: 10 },

  init: function() {
    // Returns NaN
    console.log( this.data ); 
  }

});
```
Demo: https://codepen.io/dansinni/pen/vRYJQy

### LINKED ISSUE #3681: Updating a line component using a cached THREE.Vector3 fails
A `line` component will not update if using a cached `THREE.Vector3` to call `setAttribute`. I believe this is because `deepEqual` is used in component to determine if an attribute was actually updated or not. As such, this probably affects other components that similarly allow (or at least don't explicitly disallow) the use of THREE vectors. 

Aside from updating the `deepEqual` method, another option may be to allow forcing an update via another attribute like `needsUpdate`. 

```
        init() {
          this.end = new THREE.Vector3();
        },
        
        tick() {
          this.end.set(Math.random(),  Math.random(),  Math.random());
          this.el.setAttribute("line", {end: this.end}); //will not update after the first tick
        }
```
- A-Frame Version: 0.8.2
- Platform / Device: any
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/heavenly-piranha

### LINKED ISSUE #4014: FlushToDom causing the environment to crash in Aframe 0.9.0
**Description:**

- A-Frame Version: 0.9.0
- Platform / Device: Chrome Browser
- Reproducible Code Snippet or URL:   https://glitch.com/edit/#!/phase-station

### LINKED ISSUE #4018: Single-prop value not parsed when using utils.entity.setComponentProperty
**Description:** `AFRAME.utils.entity.setComponentProperty` seems to be bypassing the property's parser for single property components. I found this when using aframe-state-component to set a `gltf-model` value. The result is the asset element ID is passed into the THREE loader instead of the asset's `src` url (because of the hash the loader humurously ends up requesting the page's HTML and trying to parse a gltf from that)

- A-Frame Version: v0.9.0
- Platform / Device: desktop chrome
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/wheat-animal?path=index.html:14:9
Note the parsing error and logged property value.

### LINKED ISSUE #5242: Single property schema where the property is an array does not work (broken since 1.0.0)

mrxz:
**Description:**
--

The component update logic is very complicated due to the many cases it needs to take care of. There are **multi-property**, **object-based single-property** and **single-property** components. The schema for multi-property components may be dynamic. There are three methods used to update (or as a reaction to an update): `updateProperties`, `resetProperty` and `handleMixinUpdate`. The `updateProperties` accepts both CSS-style strings as objects, and optionally clobbers. Combined this means that there are 6 update interactions on 4 different component types, resulting in 24 different cases. And that's not even taking into account the case of updates happening before/after the entity has initialized.

The goal of this PR was two-fold: making the component update logic more readable and reducing memory allocations. As a nice side-effect the overall performance of `setAttribute` has improved a fair bit.
|                       | set-attribute test | set-attribute test (w/ `type: 'vec3'`) |
|-----------------------|--------------------|-------------------------------|
| Firefox 122 (before)  | 7.28 µs/op         | 25.91 µs/op                   |
| Firefox 122 (after)   | 4.94 µs/op         | 6.18 µs/op                    |
| Chrome 120 (before)   | 7.84 µs/op         | 29.72 µs/op                   |
| Chrome 120 (after)    | 6.20 µs/op         | 6.67 µs/op                    |

The three methods mentioned above now all effectively follow the following pattern:
```js
    // Perform changes and recompute (either specific properties or entire data)
    this.recomputeData();
    // Handle schema changes
    this.updateSchemaIfNeeded();
    // Call update in case a change is detected
    this.callUpdateHandler();
```

**Memory changes**
--

- Reduce the number of objects from the objectPool a component needs:
  - Remove the need for `parsingAttrValue`. This intermediate object held the (raw) values that would end up on `attrValue`. Instead the parsing is done onto `attrValue` (through a proxy object)
  - Remove the need for `previousOldData`. Instead `oldData` is passed and if-and-only-if a recursive update occurs, a new `oldData` object is retrieved from the objectPool and the old one recycled once done, ensuring it remains valid for the duration of the `update` call and supports arbitrary nesting (though recursive `setAttribute` isn't ideal)
 - The inheritance chain (schema defaults, mixin values, attrValue) in `buildData` was handled through the use of `copyData` and `extendProperties` both of which `clone` objects by default, even if said values might not end up in the data. Instead, `getComputedProperty` handles the same inheritance chain, but always returns the reference, leaving it up to `recomputeProperty` to perform the copy/clone (if needed)
 - Alter the semantics of `parse(Property)` to allow passing in an (optional) target value that can be re-used. This allows the `parse` method to handle parsing, cloning and copying in one.
   - :warning: This is a change in behaviour, of the built-in parsers only the coordinates didn't behave as-such. Users might however have parse functions that do not behave as such. If they only support parsing (`string -> output`) and cloning (`output -> clone`) it would still work. If instead of cloning it becomes the identity (`output -> output`), this only works if re-using by reference isn't in issue for said type. Otherwise it'll be a breaking change
   - This also allows efficient copying from `data` into `oldData` and from the computed value into `data`

Note: Effectively the `data` and `oldData` will hold and own object based properties. The `parse` method can both allocate a new one (initially) or re-use the one already present. This is versatile approach as the exact behaviour is up to the propertyType. This also (partially) addresses #5181 as coordinates can interpret Vectors just fine and will copy from them, instead of using references. I believe this was also the original behaviour but broke when Three.js switched to classes.

**Performance changes**
--

- Changes to data are no longer performed through `deep-equal`, instead the schema is used to pick the corresponding `equals` function per property.
  -  :warning: This is a change in behaviour and won't pick up on changes made directly to `data` outside of it's known properties, e.g. adding a `this.el.components.dummy.data.someVector.myField = true` won't be picked up as a change (nor appear in `oldData`)
- Avoid calling `updateSchema` unless needed. Since the changes are detected on-write, we know when a `schemaChange` property has changed

**Bug fixes**
--

- Both `resetProperty` and `handleMixinUpdate` did not take schema changes into account, now they do
- `resetProperty` would reset the property to its schema default, ignoring any mixed in values (inconsistent inheritance)
- Due to the re-use of `parsingAttrValue`, using `setAttribute` with a CSS-style string could often lead to `updateSchema` being called even when no `schemaChange` properties were included or changed
- There are various workarounds to avoid double parsing, but these were inconsistent between single-property and multi-property components. 
- Fixed inconsistent logging of "unknown property" warnings when schema changed (double logging, false positives). 
- Fixes #3442 
- Fixes #3681
- Fixes #4014 
- Fixes #4018 
- Fixes #5242 

**Next steps**
--

This PR ended up being quite large and I can imagine hard to review. In the coming days I'll try to create a separate PR that introduces a bunch of tests (including ones failing on master) to document both the fixes and changes in behaviour. 

There are bound to be more behavioural changes when tested in the wild. Especially for users that directly mutate `.data` or even use internal methods or fields. 

Following this PR, there are a couple of other changes that could follow:
- Splitting the implementations for single- and multi-property components (no runtime overhead and easier to read).
- Unify `registerSystem` and `registerComponent`, effectively turning systems into components.

### DIFF
--- a/tests/core/a-mixin.test.js
+++ b/tests/core/a-mixin.test.js
@@ -1,5 +1,6 @@
 /* global assert, setup, suite, test */
 var helpers = require('../helpers');
+var components = require('index').components;
 var registerComponent = require('index').registerComponent;
 
 suite('a-mixin', function () {
@@ -214,6 +215,44 @@ suite('a-mixin', function () {
           });
         });
       });
+    });
+  });
+
+  suite('parseComponentAttrValue', function () {
+    var mixinEl;
+
+    setup(function () {
+      components.dummy = undefined;
+      mixinEl = document.createElement('a-mixin');
+    });
+
+    test('parses single value component', function () {
+      registerComponent('dummy', {
+        schema: {default: '0 0 1', type: 'vec3'}
+      });
+      var componentObj = mixinEl.parseComponentAttrValue(components.dummy, '1 2 3');
+      assert.deepEqual(componentObj, {x: 1, y: 2, z: 3});
+    });
+
+    test('parses component using the style parser for a complex schema', function () {
+      registerComponent('dummy', {
+        schema: {
+          position: {type: 'vec3', default: '0 0 1'},
+          color: {default: 'red'}
+        }
+      });
+      var componentObj = mixinEl.parseComponentAttrValue(components.dummy, {position: '0 1 0', color: 'red'});
+      assert.deepEqual(componentObj, {position: '0 1 0', color: 'red'});
+    });
+
+    test('does not parse properties that parse to another string', function () {
+      registerComponent('dummy', {
+        schema: {
+          url: {type: 'src', default: ''}
+        }
+      });
+      var componentObj = mixinEl.parseComponentAttrValue(components.dummy, {url: 'url(www.mozilla.com)'});
+      assert.equal(componentObj.url, 'url(www.mozilla.com)');
     });
   });
 });
--- a/tests/core/component.test.js
+++ b/tests/core/component.test.js
@@ -27,7 +27,7 @@ suite('Component', function () {
     });
   });
 
-  suite('buildData', function () {
+  suite('recomputeData', function () {
     setup(function () {
       components.dummy = undefined;
     });
@@ -41,7 +41,8 @@ suite('Component', function () {
       });
       const el = document.createElement('a-entity');
       el.setAttribute('dummy', '');
-      const data = el.components.dummy.buildData({}, null);
+      el.components.dummy.recomputeData();
+      var data = el.components.dummy.data;
       assert.equal(data.color, 'blue');
       assert.equal(data.size, 5);
     });
@@ -52,7 +53,8 @@ suite('Component', function () {
       });
       var el = document.createElement('a-entity');
       el.setAttribute('dummy', '');
-      var data = el.components.dummy.buildData(undefined, null);
+      el.components.dummy.recomputeData();
+      var data = el.components.dummy.data;
       assert.equal(data, 'blue');
     });
 
@@ -66,7 +68,8 @@ suite('Component', function () {
       });
       var el = document.createElement('a-entity');
       el.setAttribute('dummy', '');
-      var data = el.components.dummy.buildData(undefined, null);
+      el.components.dummy.recomputeData();
+      var data = el.components.dummy.data;
       assert.shallowDeepEqual(data.list, [1, 2, 3, 4]);
       assert.equal(data.none, null);
       assert.equal(data.string, '');
@@ -85,7 +88,8 @@ suite('Component', function () {
       mixinEl.setAttribute('dummy', 'color: blue; size: 10');
       el.mixinEls = [mixinEl];
       el.setAttribute('dummy', '');
-      data = el.components.dummy.buildData({}, null);
+      el.components.dummy.recomputeData();
+      data = el.components.dummy.data;
       assert.equal(data.color, 'blue');
       assert.equal(data.size, 10);
     });
@@ -102,7 +106,8 @@ suite('Component', function () {
       mixinEl.setAttribute('dummy', 'blue');
       el.mixinEls = [mixinEl];
       el.setAttribute('dummy', '');
-      data = el.components.dummy.buildData(undefined, null);
+      el.components.dummy.recomputeData();
+      data = el.components.dummy.data;
       assert.equal(data, 'blue');
     });
 
@@ -120,7 +125,8 @@ suite('Component', function () {
       mixinEl.setAttribute('dummy', 'color: blue; size: 10');
       el.mixinEls = [mixinEl];
       el.setAttribute('dummy', '');
-      data = el.components.dummy.buildData({color: 'green', size: 20}, 'color: green; size: 20');
+      el.components.dummy.recomputeData({color: 'green', size: 20});
+      data = el.components.dummy.data;
       assert.equal(data.color, 'green');
       assert.equal(data.size, 20);
     });
@@ -135,7 +141,8 @@ suite('Component', function () {
       mixinEl.setAttribute('dummy', 'blue');
       el.mixinEls = [mixinEl];
       el.setAttribute('dummy', '');
-      data = el.components.dummy.buildData('green', 'green');
+      el.components.dummy.recomputeData('green');
+      data = el.components.dummy.data;
       assert.equal(data, 'green');
     });
 
@@ -146,7 +153,8 @@ suite('Component', function () {
       });
       var el = document.createElement('a-entity');
       el.setAttribute('dummy', '');
-      data = el.components.dummy.buildData('red');
+      el.components.dummy.recomputeData('red');
+      data = el.components.dummy.data;
       assert.equal(data, 'red');
     });
 
@@ -181,7 +189,8 @@ suite('Component', function () {
       });
       var el = document.createElement('a-entity');
       el.setAttribute('dummy', {color: 'blue', depthTest: false});
-      data = el.components.dummy.buildData({color: 'red'});
+      el.components.dummy.recomputeData({color: 'red'});
+      data = el.components.dummy.data;
       assert.equal(data.depthTest, false);
       assert.equal(data.color, 'red');
     });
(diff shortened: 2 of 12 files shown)


tests/extras/primitives/primitives.test.js
```diff
@@ -340,31 +340,6 @@ suite('registerPrimitive (using innerHTML)', function () {
     });
   });
 
-  test('handles primitive created and updated in init method', function (done) {
```
mrxz, line 343: This test was specifically introduced to cover a regression, however I now believe the covered behaviour is actually a bug. The partial combining of object-based single-property data only occurs when applied _before_ the entity is initialized, not after. This inconsistent behaviour isn't ideal, so either it should always combine or never.

If backwards compatibility to this extend is needed, I can introduce a quirk that explicitly causes the old behaviour when updating a object-based single property before initialization.

tests/core/component.test.js
```diff
@@ -1036,13 +990,39 @@ suite('Component', function () {
       el.setAttribute('dummy', direction);
       sinon.assert.calledTwice(updateStub);
       // oldData passed to the update method.
-      assert.equal(updateStub.getCalls()[0].args[0].x, undefined);
-      assert.equal(updateStub.getCalls()[0].args[0].y, undefined);
-      assert.equal(updateStub.getCalls()[0].args[0].z, undefined);
-      assert.deepEqual(updateStub.getCalls()[1].args[0], {x: 1, y: 1, z: 1});
```
mrxz, line 1042: This test has been re-arranged since the captured values are by reference. Properties on oldData are only valid for the duration of `update`. Now the relevant assertions are taken during the `update`.

On master this tests 'works' because the referenced object hasn't been changed yet. Making the chain longer (say three updates), it also shows the same problem. Now this is a small behavioural change, but users shouldn't be keeping references to either data or oldData even before this change.

tests/core/component.test.js
```diff
@@ -494,7 +510,7 @@ suite('Component', function () {
       });
 
       el.hasLoaded = true;
-      el.setAttribute('dummy', '');
+      el.setAttribute('dummy', 'color: red');
```
mrxz, line 513: The value of `attrValue` is only set _once_ a value is set. Without setting any properties, `attrValue` isn't allocated (saving memory). This test inspects the `attrValue` which is internal and thus this change

Furthermore the test only cares about whether or not the selector property `el` is being cloned or not. So setting a different property to ensure `attrValue` is allocated has no bearing on the tested behaviour.

tests/core/component.test.js
```diff
@@ -27,7 +27,7 @@ suite('Component', function () {
     });
   });
 
-  suite('buildData', function () {
+  suite('recomputeData', function () {
```
mrxz, line 30: The outcome of `recomputeData` is roughly the same as `buildData`. One difference being that `recomputeData` doesn't return the updated data value, requiring these test cases to be changed accordingly.

tests/core/a-mixin.test.js
```diff
@@ -216,6 +217,44 @@ suite('a-mixin', function () {
       });
     });
   });
+
+  suite('parseComponentAttrValue', function () {
```
mrxz, line 221: Moved and adapted from `component.test.js` where it tested the `parseAttrValue` method. Since this PR causes it to only be used by `a-mixin`, the tests were moved as well.

dmarcos:
needs rebase

dmarcos:
needs rebase

dmarcos:
Is still the plan to break this PR on smaller ones or should I take it as is?

dmarcos:
This is done?

mrxz:
> This is done?
The PR is in a complete state, so you could start reviewing it. I know I said I would open a PR with additional tests, but sadly haven't gotten around to it. That shouldn't change the implementation, though, mostly cover the bugfixes listed above.

dmarcos:
Thanks! Great work!

diarmidmackenzie:
Haven't had a chance to look in detail, but this looks like a substantial improvement to an area of A-Frame that was rather patchy and inconsistent (and hence frequently hard to work with)

Thank you @mrxz.  Awesome vision and execution!

mrxz:
Thanks, I'm quite pleased with how it turned out. It's still a complex part of A-Frame, simply by the fact that there a so many possible code paths when updating components, but the code should be easier to follow and behave more consistent. Some inconsistency (e.g. between single- and multi-prop) have explicitly been added for backwards compatibility and now "stick out" instead of being unintended side-effects of the different implementations.

In any case, if you do happen to look at it in more detail or notice any changes or regressions after updating, let me know.

PR #5475, opened 2024-02-27
Fix duplicate check in registerPropertyType

mrxz:
**Description:**
The check preventing duplicate `propertyTypes` from being registered incorrectly tested the string literal `"type"` instead of the type argument.

**Changes proposed:**
- Fix the duplicate check in `registerPropertyType`
- Throw an error instead of logging an error (in line with `registerComponent`, `registerSystem`, etc...)


dmarcos:
Thanks!

PR #5476, opened 2024-02-27
Avoid allocating functions and use at most one timer per interval in throttleLeadingAndTrailing

mrxz:
**Description:**
The implementation of `throttleLeadingAndTrailing` was very inefficient when many updates took place in short succession. Virtually all invocations would result in clearing a timer, allocating a new function and creating a new timer. This impacted both the performance and memory usage of `setAttribute`. 

Using the `performance/set-attribute` and `performance/animation-set-attribute` benchmarks showed a clear performance benefit (when congested) cutting the duration in half (~9.16µs/op -> ~4.04µs/op) and reduced memory consumption:
image

Instead of allocating a new function for each timer, one is created in advance. In case calls take place within the `minimumInterval` only one timer is created and re-used until it expires ensuring at most one timer per interval (if needed).

**Changes proposed:**
- Pre-allocate timeout callback function
- Avoid clearing and recreating timeout and instead create and re-use one timeout per interval (if needed)

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -89,20 +89,30 @@ module.exports.throttleLeadingAndTrailing = function (functionToThrottle, minimu
   if (optionalContext) {
     functionToThrottle = functionToThrottle.bind(optionalContext);
   }
+  var args;
+  var timerExpired = function () {
+    // Reached end of interval, call function
+    lastTime = Date.now();
+    functionToThrottle.apply(this, args);
+    deferTimer = undefined;
+  };
+
   return function () {
     var time = Date.now();
     var sinceLastTime = typeof lastTime === 'undefined' ? minimumInterval : time - lastTime;
-    var args = arguments;
-    if (typeof lastTime === 'undefined' || sinceLastTime >= minimumInterval) {
+    if (sinceLastTime >= minimumInterval) {
+      // Outside of minimum interval, call throttled function.
+      // Clear any pending timer as timeout imprecisions could otherwise cause two calls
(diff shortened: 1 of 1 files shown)


src/utils/index.js
```diff
@@ -89,20 +89,32 @@ module.exports.throttleLeadingAndTrailing = function (functionToThrottle, minimu
...
-      functionToThrottle.apply(null, args);
+      functionToThrottle.apply(null, arguments);
     } else {
-      clearTimeout(deferTimer);
-      deferTimer = setTimeout(function () {
-        lastTime = Date.now();
-        functionToThrottle.apply(this, args);
-      }, minimumInterval - sinceLastTime);
+      // Inside minimum interval, create timer if needed.
+      if (typeof deferTimer === 'undefined') {
```
dmarcos, line 113: can check for just `deferTimer === undefined` for consistency with rest of code base

src/utils/index.js
```diff
@@ -89,20 +89,32 @@ module.exports.throttleLeadingAndTrailing = function (functionToThrottle, minimu
...
+        deferTimer = setTimeout(timerExpired, minimumInterval - sinceLastTime);
```
dmarcos, line 114: can do a single line 

```
deferTimer = deferTime || setTimeout(timerExpired, minimumInterval - sinceLastTime);
```

mrxz: Done

dmarcos:
Thanks!

PR #5477, opened 2024-02-27
Make test teardown in __init.test.js synchronous

mrxz:
**Description:**
Removing the timeout from the `teardown` in `__init.test.js` speeds up test execution significantly on my machine. According to the accompanying comment this timeout is there to let `detachCallbacks` clean themselves up before considering the teardown complete. However, custom elements v1 uses `disconnectedCallback` which is (effectively) synchronous, and therefore the `setTimeout` should have no bearing on them.

Removing the timeout doesn't seem to have a negative impact. All tests are still green after making this change.

**Changes proposed:**
- Make `teardown` synchronous in `__init.test.js` speeding up tests

### DIFF
--- a/tests/__init.test.js
+++ b/tests/__init.test.js
@@ -82,7 +82,7 @@ window.addEventListener('unhandledrejection', event => {
   pendingErrorNotice = 'An uncaught promise rejection occurred between tests';
 });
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks

PR #5478, opened 2024-02-28
Allow components to define before/after constraints

### LINKED ISSUE #4164: Component ticks behave differently in Chrome vs Firefox when entities are removed

### LINKED ISSUE #5400: Animation hiccups when an object is removed from the DOM

mrxz:
**Description:**
The scene calls all the tick and tock methods in the order they were added. This can cause problems when components use the output of other components. They either get called after, in which case they have up-to-date info, or before, in which case they have 1 frame stale data. The longer the chain is the more frames of 'latency' this might add. This is often an issue with components that inspect or use the position of another element (e.g. collision detection before updated controller pose, nav-mesh constraints before locomotion, etc...). This can be seen in Ada's [aframe-xr-boilerplate](https://github.com/AdaRoseCannon/aframe-xr-boilerplate):

[update-order-PR-vid.webm](https://github.com/aframevr/aframe/assets/8823461/3c0f6338-ce7d-4bd0-9467-67b389b486ab)

This PR expand the component definition with `after` and `before`. A components `tick` will be called after **_all_** `tick` methods of the components listed in `after` have taken place and before **_all_** `tick` methods of the components listed in `before`. The scene updates per component, the order among components of the same type is still undefined.

This PR also includes an alternative implementation of #5403 (fixes #5400, fixes #4164) . Since calls take place per component type, the odds that a component adds/removes components of the same type is unlikely, meaning that a happy path can be taken most of the time (directly add or remove behaviour). In case it does happen, the behaviours are marked for removal and removed _after_ the scene is done with that particular component.

**Changes proposed:**
- Allow components to define before/after constraints
- Let the scene call `tick` and `tock` per component respecting the given constraints
- Correctly handle `removeBehavior` when called from a `tick` or `tock` method
- Use before/after constraints to fix a bug in the tracked-controls example, where quickly waving your controllers while having grabbed a cube could show the cube without highlight for brief periods.


src/utils/order.js
```diff
@@ -0,0 +1,75 @@
+var debug = require('./debug');
+var warn = debug('utils:order:warn');
+
+/**
+ * Derives an ordering from the elements with before and after constraints.
```
dmarcos, line 5: order no descriptive. not sure it needs to be a separate utils file. we can just include this function in a-scene

src/core/scene/a-scene.js
```diff
@@ -139,6 +142,12 @@ class AScene extends AEntity {
     });
 
     this.initSystems();
+    // Compute component order
+    this.componentOrder = solveOrder(components, this.componentOrder);
+    this.addEventListener('componentregistered', function () {
+      // Recompute order
+      self.componentOrder = solveOrder(components, self.componentOrder);
```
dmarcos, line 149: `solveOrder` not super descriptive. `sortComponents`?

mrxz: It's not sorting, the `components` object isn't mutated in the process. It just returns an order which it arrives at by solving the set of `before/after` constraints, hence `solveOrder`.

Perhaps something like `determineOrder` (or `determineComponentOrder`) could work?

src/utils/order.js
mrxz: The `a-scene.js` file is already pretty lengthy. I'd rather not append this to it, especially given that the implementation details aren't relevant from the POV of the scene.

Definitely open for (file) name suggestions. I left out `component` as it can also be applied to systems down the line, but if it helps clarity now, it can be included in the name.

dmarcos: 900 LOC is not super big and purpose of `utils` is factor out code used from multiple files or exported. When understanding a code base having tons of files that don't know how to relate to one another a bigger problem than big files. I would just incorporate this code to `a-scene`

src/core/scene/a-scene.js
dmarcos: `sortComponentsByTikTokOrder` ? It's not a sort in place but still a sort

src/core/scene/a-scene.js
```diff
@@ -721,9 +772,23 @@ class AScene extends AEntity {
     var systems = this.systems;
 
     // Components.
-    for (i = 0; i < this.behaviors.tock.length; i++) {
-      if (!this.behaviors.tock[i].el.isPlaying) { continue; }
-      this.behaviors.tock[i].tock(time, timeDelta, camera);
+    for (var c = 0; c < this.componentOrder.length; c++) {
```
dmarcos, line 775: wonder if this code and above can be factored out into a function. `callBehavior('tick') / callBehavior ('tock')`

dmarcos:
This one is pretty close

src/core/scene/a-scene.js
mrxz: Not trying to be pedantic, but it really isn't sorting, neither in-place nor as its result. The resulting order could be used for sorting, but isn't. Naming it as such feels out of place to me.

src/utils/order.js
mrxz: Don't really agree, but ultimately a subjective point, so added it to `a-scene.js`.

dmarcos:
Thanks! Great work!

PR #5480, opened 2024-02-28
Update webpack to latest supported version

vincentfretin:
**Description:**

Update build tools dependencies.

**Changes proposed:**
- Update webpack and associated loaders
- Update webpack-dev-server, this requires now at least node 18.12.0 that everyone should have, older versions of node aren't supported anymore. Current LTS is node 20.x
- Remove webpack-merge dependency, webpack now supports extending another configuration with "extends"


dmarcos:
Are any of these updates needed? In general I avoid updating dependencies unless there's a reason to do so. If things work no reason to get exposed to API changes, regressions, new bugs

vincentfretin:
@dmarcos now that the bot is using node 22.7.0 I don't see any reason to not merge this.

dmarcos:
@vincentfretin is this a one time thing or we will have to continuously update webpack and stuff. Would prefer to avoid get on the train of updating packages just for the shake of it. Can of course update if there's a reason to do so

vincentfretin:
That's a one time thing to remove a dependency, otherwise I wouldn't have bothered to do this PR.

dmarcos:
Thanks!

PR #5481, opened 2024-02-28
Defer initialization until end of aframe script or manual ready signal

mrxz:
**Description:**
Updated and improved version of #5419. All `<a-node>` derived custom elements now wait for an `aframeready` event. This event is emitted near the end of the aframe bundle _OR_ once the `document.readyState` becomes `complete`, whichever occurs last. In case the user needs to load more dependencies after this point, they can set the `window.AFRAME_ASYNC` flag and manually provide the ready signal using `AFRAME.ready()`

The following configurations are now all supported.

**Classic script**
```HTML
<head>
    <script src="aframe-master.js"></script>
    <script src="components-and-systems.js"></script>
</head>
```

**Deferred script** (resolving #4038)
```HTML
<head>
    <script src="aframe-master.js" defer></script>
    <script src="components-and-systems.js" defer></script>
</head>
```

**(Async) Inline Module**
```HTML
<head>
    <script type="importmap">(...)</script>
    <script type="module" async> <!-- works with and without async -->
        import 'aframe';
        import 'components-and-systems';
    </script>
</head>
```

**(Async) Module script file**
```HTML
<head>
    <script type="importmap">(...)</script>
    <script type="module" src="index.mjs" async></script> <!-- works with and without async -->
</head>
```

**Inline Module (w/ dynamic imports)** (original problem #5419, see [updated glitch](https://glitch.com/edit/#!/dour-sassy-stone))
```HTML
<head>
    <script type="importmap">(...)</script>
    <script type="module">
        window.AFRAME_ASYNC = true;
        await import("aframe")
        await import("components-and-systems");
        window.AFRAME.ready();
    </script>
</head>
```

You only really need to use `AFRAME_ASYNC` + `AFRAME.ready()` if the aframe main bundle is loaded _after_ the `document.readyState` has already reached `completed` **and** you asynchronously register new components or systems. This means asynchronously compared to the aframe main bundle, so the following works:
```JS
import 'aframe';
AFRAME.registerComponent('new-component', { ... });
```
Whereas this would fail:
```JS
import 'aframe';
setTimeout(() => {
    AFRAME.registerComponent('new-component', { ... });
});
```

The way it fails is interesting in that it doesn't give errors and the components/system will work fine for the most part. But since they are registered _after_ the scene has initialized, any attributes that weren't recognized as components or systems won't trigger component initialization. Effectively the runtime loading of components and systems could be improved. However, users should probably opt for `AFRAME_ASYNC` unless they know what they are doing, as cross component dependencies could still cause issues (say a library registers both `foo` and `bar`, where `foo` depends on `bar`, in which case initializing `foo` before `bar` is registered causes issues)

**Changes proposed:**
- Let all `<a-node>` derived elements wait on `aframeready` event instead of `readystatechanged`
- Emit `aframeready` once the main bundle is done or the `document.readyState` is `complete` (whichever comes last)
- Introduce `window.AFRAME_ASYNC` flag to allow users to manually signal A-Frame that everything is ready (through `AFRAME.ready()`)

### DIFF
--- a/src/core/a-node.js
+++ b/src/core/a-node.js
@@ -1,5 +1,6 @@
 /* global customElements, CustomEvent, HTMLElement, MutationObserver */
 var utils = require('../utils/');
+var readyState = require('./readyState');
 
 var warn = utils.debug('core:a-node:warn');
 
@@ -32,19 +33,13 @@ class ANode extends HTMLElement {
     this.mixinEls = [];
   }
 
-  onReadyStateChange () {
-    if (document.readyState === 'complete') {
-      this.doConnectedCallback();
-    }
-  }
-
   connectedCallback () {
-    // Defer if DOM is not ready.
-    if (document.readyState !== 'complete') {
-      document.addEventListener('readystatechange', this.onReadyStateChange.bind(this));
+    // Defer if not ready to initialize.
+    if (!readyState.canInitializeElements) {
+      document.addEventListener('aframeready', this.connectedCallback.bind(this));
       return;
     }
-    ANode.prototype.doConnectedCallback.call(this);
+    this.doConnectedCallback();
   }
 
   doConnectedCallback () {
--- a/src/index.js
+++ b/src/index.js
@@ -59,6 +59,7 @@ var shaders = require('./core/shader').shaders;
 var systems = require('./core/system').systems;
 // Exports THREE to window so three.js can be used without alteration.
 var THREE = window.THREE = require('./lib/three');
+var readyState = require('./core/readyState');
 
 var pkg = require('../package');
 
@@ -82,6 +83,11 @@ console.log('THREE Version (https://github.com/supermedium/three.js):',
             pkg.dependencies['super-three']);
 console.log('WebVR Polyfill Version:', pkg.dependencies['webvr-polyfill']);
 
+// Wait for ready state, unless user asynchronously initializes A-Frame.
+if (!window.AFRAME_ASYNC) {
+  readyState.waitForDocumentReadyState();
+}
+
 module.exports = window.AFRAME = {
   AComponent: require('./core/component').Component,
   AEntity: AEntity,
@@ -104,6 +110,7 @@ module.exports = window.AFRAME = {
   schema: require('./core/schema'),
   shaders: shaders,
   systems: systems,
+  emitReady: readyState.emitReady,
   THREE: THREE,
   utils: utils,
   version: pkg.version
--- a/docs/introduction/faq.md
+++ b/docs/introduction/faq.md
@@ -344,6 +344,39 @@ Phones with Adreno 300 series GPUs are notoriously problematic. Set [renderer pr
 
 Using A-Frame online sometimes is not possible or inconvenient, like for instance when traveling or during public events with poor Internet connectivity. A-Frame is mostly self-contained so including the build (aframe.min.js) in your project will be sufficient in many cases. Some specific parts are lazy loaded and only fetched when used. This is for example the case of the fonts for the text component and the 3D models for controllers. In order to make an A-Frame build work either offline or without relying on A-Frame hosting infrastructure (typically cdn.aframe.io), you can monitor network requests on your browser console. This will show precisely what assets are being loaded and thus as required for your specific experience. Fonts can be found via FONT_BASE_URL in the whereas controllers via MODEL_URLS. Both can be modified in the source and included in your own [custom build](https://github.com/aframevr/aframe#generating-builds)
 
+## Can I load A-Frame as an ES module?
+
+You can load A-Frame as an ES module using a [side effect import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only). A-Frame will then initialize any `<a-scene>` in the document. It's still important to register any components or systems you need before this happens:
+
+```HTML
+<head>
+  <script type="importmap">
+    {
+        "imports": {
+            "aframe": "https://aframe.io/releases/1.5.0/aframe.min.js",
+        }
+    }
+  </script>
+  <script type="module">
+    import 'aframe';
+    AFRAME.registerComponent('my-component', {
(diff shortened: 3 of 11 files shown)


src/core/a-node.js
```diff
@@ -1,5 +1,6 @@
 /* global customElements, CustomEvent, HTMLElement, MutationObserver */
 var utils = require('../utils/');
+var ready = require('./ready');
```
dmarcos, line 3: `ready` is not a very descriptive word. Can we think of something better?

mrxz: Perhaps `readyState` analogous to the `document.readyState`?

dmarcos: aframeReadyState?

src/index.js
```diff
@@ -104,6 +110,7 @@ module.exports = window.AFRAME = {
   schema: require('./core/schema'),
   shaders: shaders,
   systems: systems,
+  ready: ready.ready,
```
dmarcos, line 113: Do we need to make this API public?

src/core/a-node.js
```diff
@@ -32,19 +33,13 @@ class ANode extends HTMLElement {
...
-
   connectedCallback () {
     // Defer if DOM is not ready.
-    if (document.readyState !== 'complete') {
-      document.addEventListener('readystatechange', this.onReadyStateChange.bind(this));
+    if (!ready.isReady) {
+      document.addEventListener('aframeready', this.connectedCallback.bind(this));
       return;
     }
-    ANode.prototype.doConnectedCallback.call(this);
```
dmarcos, line 47: Any reason why this call has changed?

src/index.js
mrxz: Yes, see the **Inline Module (w/ dynamic imports)** case. Basically we defer the initialization till after A-Frame is fully loaded + anything synchronous after that. However, if the user loads additional components/systems asynchronously after that point, they'll run into issues. In that case, and that case only, they'll need a way to signal when everything is ready. 

Note that this is also the original example motivating #5419.

src/core/a-node.js
mrxz: As the local var name that works, though as file name I'd say `readyState.js` suffices as the aframe prefix won't make things clearer IMHO.

src/core/a-node.js
mrxz: Yeah, the inheritance was a bit wonky. The `doConnectedCallback` function was essentially setup as a "virtual function" with subclasses overriding it and calling the base implementation (`super.doConnectedCallback()`), yet `connectedCallback` explicitly called the `ANode` implementation, forcing all subclasses to also override `connectedCallback` and replicating the same logic. From an OOP standpoint this didn't really make sense. Having one shared `connectedCallback` implementation is the cleanest solution IMO.

src/index.js
dmarcos: `ready` sounds like a flag. should we use a verb for a function. `init`, `start`?

src/core/a-node.js
```diff
@@ -32,19 +33,13 @@ class ANode extends HTMLElement {
...
-    if (document.readyState === 'complete') {
-      this.doConnectedCallback();
-    }
-  }
-
   connectedCallback () {
     // Defer if DOM is not ready.
-    if (document.readyState !== 'complete') {
-      document.addEventListener('readystatechange', this.onReadyStateChange.bind(this));
+    if (!ready.isReady) {
```
dmarcos, line 38: should be `isDOMReady` but not sure because isReady involves more than DOM now

mrxz: Yeah, it now encompasses both DOM (= all elements are parsed) and A-Frame (everything registered, incl. third-party components/systems/etc...) being ready. Not sure what an apt name would be. 

Perhaps the name could be changed to not reflect what needs to be ready, but that this action can take place. So something like `canInitializeElements`?

dmarcos:
We should add to docs and maybe README (if we can keep it simple) the different ways to load A-Frame listed in this PR description

src/core/a-node.js
dmarcos: `isAnodeReadyToLoad`? too specific?

mrxz: Feels a tad specific, but it also reads as if "Anode" _is ready_, instead of everything being ready "for Anode" to load. And we'd probably want to avoid using "load" here as the `doConnectedCallback` implementations do more than just invoke `load`.

dmarcos:
How common would this pattern be? I don't have enough experience with modules

```
<head>
    <script type="importmap">(...)</script>
    <script type="module">
        window.AFRAME_ASYNC = true;
        await import("aframe")
        await import("components-and-systems");
        window.AFRAME.ready();
    </script>
</head>

```

mrxz:
@dmarcos This PR address more than just the dynamic imports. Without this PR none of the module based approaches work, nor classic scripts marked `defer`. Literally the only way to load A-Frame would be a blocking synchronous classic script. The dynamic import case in question only requires the `emitReady` method being exposed and the `window.AFRAME_ASYNC` condition in `index.js`. 

While dynamic imports aren't the most common in my experience, #5419 raises a valid usage. On top of that more complex module setups (e.g. using top-level awaits) might also need `window.AFRAME_ASYNC` + `emitReady` despite not using dynamic imports.

Having this escape hatch is useful in those situations. Another option would be to treat module based imports differently by always requiring the user to manually signal everything is ready. That would eliminate the need for the magic `AFRAME_ASYNC`, but it would break any app out there that already got A-Frame working this way with their own workarounds (which, given the various issues on this subject over time, are probably more than a few).

src/index.js
```diff
@@ -104,6 +110,7 @@ module.exports = window.AFRAME = {
   schema: require('./core/schema'),
   shaders: shaders,
   systems: systems,
+  emitReady: readyState.ready,
```
dmarcos, line 113: I think it should be `readyState.emitReady`

dmarcos:
Can we add some docs?

dmarcos:
Thanks!

Issue #3211, opened 2017-11-01
document how to disable default lights on dynamically inserted scenes


bitonic:
**Description:**

I could not find a reliable way to reliably disable default lights on an `a-scene` which is inserted from JavaScript rather than already present in the HTML. The closest I could get is https://glitch.com/edit/#!/substantial-latex?path=index.html:14:0 , and that does not seem to work on Chrome.

- A-Frame Version: 0.7.0
- Platform / Device: Linux, Firefox developer edition and Chrome 61
- Reproducible Code Snippet or URL: https://glitch.com/edit/#!/substantial-latex?path=index.html:14:0

donmccurdy:
From docs on the [`light` component and system](https://aframe.io/docs/0.7.0/components/light.html), you're looking for:

```html
<a-scene light="defaultLightsEnabled: false">
  <!-- ... -->
</a-scene>
```

bitonic:
That does not work for dynamically inserted elements, which is why I opened this ticket. It only works for elements which are already in the HTML.

donmccurdy:
Hm I see. I would have expected this to work...

```js
var scene = document.createElement("a-scene");
scene.setAttribute('light', 'defaultLightsEnabled:false');
scene.setAttribute('data-foo', 'bar');
document.body.appendChild(scene);
```

... but it does not. Only the `data-foo` attribute remains after initialization.

JP-Dorman:
"Whenever we add any lights, A-Frame removes the default lights from the scene." - Could you add your own ambient and directional lights with intensity 0 by default, then in JS remove them on page load?

PR #5482, opened 2024-02-29
Fix bug in a-scene's setAttribute in case of uninitialized systems

mrxz:
**Description:**
The `setAttribute` method in `<a-scene>` prioritizes systems over components and other attributes. However, it did this by checking if the attribute name was present in the object with _initialized_ systems. When programmatically creating an `a-scene` these might not have been initialized yet.

This PR changes the condition to first check if the attribute name matches a _registered_ system.

Fixes: #3211

**Changes proposed:**
- Check if attribute name matches a registered system in `setAttribute`
- Introduce unit tests for checking `<a-scene>`'s `setAttribute`

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -515,10 +515,15 @@ class AScene extends AEntity {
    * setAttribute.
    */
   setAttribute (attr, value, componentPropValue) {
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks!

PR #5483, opened 2024-02-29
Handle missing mixins and applying mixin attributes to mappings of primitives

### LINKED ISSUE #4455: Inconsistent behaviour for missing mixin
**Description:**
When a missing mixin is encountered for a-sphere, WebApp crashes. However, for a-entity with line, the missing mixin doesn't trigger crash.

Either one of the behaviour could be valid, but needs to be consistent.

- A-Frame Version: 1.0.4
- Platform / Device: Win10/Firefox
	```
	<a-entity id="axes">
			<a-entity id="origin_line" line="start: 0, 0, 0; end: 20000 0    0; color: red;   opacity:1" mixin="missing_mixin_no_error"><a-entity>
			<!--a-sphere mixin="missing_mixin_throws_error"-->	</a-sphere><!--uncomment to trigger crash-->
		</a-entity>
```

### LINKED ISSUE #4717: Color property not applying to primitive as expected.
Probably due to collision or application precedence between primitive shorthand `color="red"` attribute and associated component expansion `material="color: red"`. [Example to reproduce](https://boom-alabaster-tungsten.glitch.me/). Cylinder should be red and it's white instead

### PR DESCRIPTION
**Pull request by mrxz**
**Description:**
Primitives handled mixins differently compared to `a-node`. In case a mixin was referenced that didn't exist, the primitive would throw an error while the `a-node` silently ignored it. Now a warning is logged in both cases. Primitives now also check if attributes on mixins correspond to any of their mappings, allowing things like the following to work:
```HTML
<a-assets>
    <a-mixin id="red" color="red"></a-mixin>
</a-assets>

<a-box mixin="red"></a-box>
```

Fixes #4455
Fixes #4717 

**Changes proposed:**
- Log warning when mixin can't be found (instead of throwing an error or silently continuing).
- Apply mapping in primitive for any mixin attributes that correspond with defined mappings.

### DIFF
--- a/src/core/a-mixin.js
+++ b/src/core/a-mixin.js
@@ -8,11 +8,13 @@ var MULTIPLE_COMPONENT_DELIMITER = '__';
 /**
  * @member {object} componentCache - Cache of pre-parsed values. An object where the keys
  *         are component names and the values are already parsed by the component.
+ * @member {object} rawAttributeCache - Cache of the raw attribute values.
  */
 class AMixin extends ANode {
   constructor () {
     super();
     this.componentCache = {};
+    this.rawAttributeCache = {};
     this.isMixin = true;
   }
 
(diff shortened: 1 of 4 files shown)


dmarcos:
Thanks!

PR #5486, opened 2024-03-04
Move a-cubemap.test.js into the right folder

mrxz:
**Description:**
The test file for `a-cubemap.js` is currently located under `tests/extra/primitives/primitives/` whereas `a-cubemap.js` itself is located at `src/core/a-cubemap.js`. This seems like a small oversight, so this PR moves the tests to the matching folder.

**Changes proposed:**
- Move `a-cubemap.test.js` to `tests/core/a-cubemap.test.js`


dmarcos:
Thanks!

PR #5487, opened 2024-03-04
Use `window.isSecureContext` instead of manual check for localhost and/or https

### LINKED ISSUE #4672: Replace the HTTPS check with `window.isSecureContext`?
**Description:**
- A-Frame Version: latest
- Platform / Device: all
- Reproducible Code Snippet or URL: /

**Bug:**
When running an A-Frame app locally on `localhost.myapp:8080` on a browser that supports it e.g. Chrome, the alert saying that sensor APIs are not available is displayed. 
Expected behaviour: no alert, because `localhost.myapp:8080` **is** a `SecureContext`, so sensor APIs are available.

**Analysis/Assumption:**
- According to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Sensor_APIs), device sensors require a `SecureContext`. I.e. `localhost` (= `127.0.0.1`), or an app running on `https`.
- However, right now, the [check](https://github.com/aframevr/aframe/blob/7d289e8fb28f2c2895ba49200f502257110d5050/src/components/scene/device-orientation-permission-ui.js#L29) to display an alert is made "heuristically" / by looking at the URL as a proxy to determine whether this is a `SecureContext`.

**Other issue:**
- The alert text says "Access this site over HTTPS to enter VR mode and grant access to the device sensors" when in fact device sensor access is not about HTTPS (in my understanding!). Rather, it's about `SecureContext`s, which is a bit different.

**Suggestion:**
- Replace the [check](https://github.com/aframevr/aframe/blob/7d289e8fb28f2c2895ba49200f502257110d5050/src/components/scene/device-orientation-permission-ui.js#L29) with the dedicated call `window.isSecureContext`?
and
- Replace the alert text?

Let me know if this makes sense! Especially: is the [MDN list](https://developer.mozilla.org/en-US/docs/Web/API/Sensor_APIs) comprehensive, or am I omitting a range of sensor APIs that do need `HTTPS`?

mrxz:
**Description:**
As detailed in #4672, the check performed in the `device-orientation-permission-ui` component is attempting to determine if the user is accessing the page in a secure context. Using `window.isSecureContext` for this should be more reliable and avoids false negatives, as the browser _knows_ whether this is the case.

Fixes #4672

**Changes proposed:**
- Replace manual check for localhost and/or https with `window.isSecureContext`

### DIFF
--- a/src/components/scene/device-orientation-permission-ui.js
+++ b/src/components/scene/device-orientation-permission-ui.js
@@ -37,9 +37,7 @@ module.exports.Component = registerComponent('device-orientation-permission-ui',
 
     if (!this.data.enabled) { return; }
 
-    if (location.hostname !== 'localhost' &&
-        location.hostname !== '127.0.0.1' &&
-        location.protocol === 'http:') {
+    if (!window.isSecureContext) {
       this.showHTTPAlert();
     }
 


dmarcos:
Thanks so much!

PR #5489, opened 2024-03-07
Fixed XR Layer rendering race condition with session initialization and RenderState Update

zjm-meta:
**Description:**
The aframe comicbook example was broken with exceptions of 

```
TypeError: Failed to execute 'updateRenderState' on 'XRSession': Failed to read the 'layers' property
from 'XRRenderStateInit': Failed to convert value to 'XRLayer'.
```

and 

```
Uncaught InvalidStateError: Failed to execute 'getSubImage' on 'XRWebGLBinding': Layer must be in the session's RenderState.
```

The issues are related to the race condition.  The`!this.el.sceneEl.xrSession` check in [this](https://github.com/aframevr/aframe/blob/master/src/components/layer.js#L214) line doesn't guarantee the full initialization of the xr session. We should actually use the `this.el.sceneEl.renderer.xr.isPresenting` value which is set to `true` at the end of XR session initialization [here](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webxr/WebXRManager.js#L366).

Another race condition is that the `this.xrGLFactory.getSubImage` call should happen AFTER the layer is added to the sessions layers array (related W3C [doc](https://www.w3.org/TR/webxrlayers-1/?fbclid=IwAR27yKqbCs571mbvJjx11QzGeI2bocMuieoMQ8-Psr4ai3TD-foT84WciIM#dom-xrwebglbinding-getsubimage)). This happens in the next frame after `sceneEl.renderer.xr.addLayer` is called. Thus, we shouldn't force a redraw in `onEnterVR`. Instead, we need to rely on the `this.layer.needsRedraw` value [here](https://github.com/aframevr/aframe/blob/master/src/components/layer.js#L218) to trigger a redraw after the layer is added to the session layers array.

**Changes proposed:**
- Skipped all layer `tick()` function if `this.el.sceneEl.renderer.xr.isPresenting === false`
- Removed the force redraw logic in the 1st frame in `onEnterVR `.

**Testing**
Ran `npm start` and used port forwarding to test in a Quest 3 headset. The comic book example works smoothly.

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -211,7 +211,7 @@ module.exports.Component = registerComponent('layer', {
   },
 
   tick: function () {
-    if (!this.el.sceneEl.xrSession) { return; }
+    if (!this.el.sceneEl.xrSession || !this.el.sceneEl.renderer.xr.isPresenting) { return; }
     if (!this.layer && this.el.sceneEl.is('vr-mode')) { this.initLayer(); }
     this.updateTransform();
     if (this.data.src.complete && (this.pendingCubeMapUpdate || this.loadingScreen || this.visibilityChanged)) { this.loadCubeMapImages(); }
@@ -367,7 +367,6 @@ module.exports.Component = registerComponent('layer', {
       return;
     }
     xrSession.requestReferenceSpace('local-floor').then(this.onRequestedReferenceSpace);
-    this.needsRedraw = true;
     this.layerEnabled = true;
     if (this.quadPanelEl) {
       this.quadPanelEl.object3D.visible = false;
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -211,7 +211,7 @@ module.exports.Component = registerComponent('layer', {
   },
 
   tick: function () {
-    if (!this.el.sceneEl.xrSession || !this.el.sceneEl.renderer.xr.isPresenting) { return; }
+    if (!this.el.sceneEl.xrSession) { return; }
     if (!this.layer && this.el.sceneEl.is('vr-mode')) { this.initLayer(); }
     this.updateTransform();
     if (this.data.src.complete && (this.pendingCubeMapUpdate || this.loadingScreen || this.visibilityChanged)) { this.loadCubeMapImages(); }
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -299,8 +299,6 @@ class AScene extends AEntity {
           self.usedOfferSession |= useOfferSession;
           requestSession(xrMode, xrInit).then(
             function requestSuccess (xrSession) {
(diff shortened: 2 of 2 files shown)


mrxz:
Good catch, wondering if it might be better to ensure that we only set `sceneEl.xrSession` once the `XRSession` is fully setup. As the `layer` component isn't the only code assuming the presence of `xrSession` means it's ready.

Should be a matter of rearranging the [code in `a-scene.js`](https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L302-L314). Basically moving it into the `then` callback of `setSession`.

zjm-meta:
Yeah that makes perfect sense to me. I will update the commit to reflect this and do some tests. Thanks for the suggestion.

zjm-meta:
I did some test and I found out that we are generally not handling the undefined  `sceneEl.xrSession` gracefully in the `enterVR()` [functions](https://github.com/search?q=repo%3Aaframevr%2Faframe%20sceneEl.xrSession&type=code). For example, in the [layers.js](https://github.com/aframevr/aframe/blob/a578e3cda1df88a976c01900cc62b6dfe441ffa1/src/components/layer.js#L365-L368) we would just skip the initialization steps when xrSession is undefined instead of waiting for the promise. This breaks the experience if we move `self.xrSession = xrSession;`. Do you think we should change the way we implement `enterVR`?

dmarcos:
> I did some tests and I found out that we are generally not handling the undefined `sceneEl.xrSession` gracefully in the `enterVR()` [functions](https://github.com/search?q=repo%3Aaframevr%2Faframe%20sceneEl.xrSession&type=code). For example, in the [layers.js](https://github.com/aframevr/aframe/blob/a578e3cda1df88a976c01900cc62b6dfe441ffa1/src/components/layer.js#L365-L368) we would just skip the initialization steps when xrSession is undefined instead of waiting for the promise. This breaks the experience if we move `self.xrSession = xrSession;`. Do you think we should change the way we implement `enterVR`?

Thanks. Is the change we suggested not enough?

https://github.com/aframevr/aframe/pull/5489#issuecomment-1984221260

Feel free to suggest alternatives. We should definitively handle `sceneEl.xrSession` undefined gracefully.

mrxz:
The `enterVRSuccess(resolve);` should also be moved. Otherwise we'd indeed have `enterVR` events before `sceneEl.xrSession` is set. Basically `enterVR` should imply `sceneEl.xrSession` is available (and ready).

zjm-meta:
Thanks @mrxz ! Moving the position of `enterVRSuccess(resolve);` did the job. I have updated this PR to reflect this.

dmarcos:
@zjm-meta thanks so much for the effort. great work. congrats on your first contribution! 👏

PR #5491, opened 2024-03-07
Apply waiting and timeout on all assets, not just img, audio and video

### LINKED ISSUE #5130: Assets timeout produces an error with stacktrace
**Description:**

- A-Frame Version: 1.3.0 and master (2022-10-06)
- Platform / Device: All
- Reproducible Code Snippet or URL: `npm run start:https`
  It can be reproduced on https://192.168.1.15:9000/test/gltf-model/  (replace by your ip)
just change to timeout after 5ms : `<a-assets timeout="5">` in `examples/test/gltf-model/index.html`

It produces the following error:
```
core:a-assets:warn Asset loading timed out in   5 ms
component.js:736 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'constructor')
    at extendProperties (component.js:736:31)
    at NewComponent.callUpdateHandler (component.js:416:20)
    at NewComponent.updateProperties (component.js:300:12)
    at HTMLElement.value (a-entity.js:490:19)
    at HTMLElement.value (a-entity.js:456:14)
    at entityLoadCallback (a-entity.js:249:14)
    at emitLoaded (a-node.js:136:21)
components:gltf-model:warn Unexpected token '<', "<!DOCTYPE "... is not valid JSON 
```
I didn't investigate further on this issue.

The glb here is loading correctly after this error, the scene seems to work properly, but it may not always be the case as I read from someone on slack having this error.
The default 3s timeout is really too short in my opinion, I actually always increase it to 30s (timeout="30000") in all my projects to effectively disable that feature.

mrxz:
**Description:**
The asset loading system differentiated between ``, `<audio>`, `<video>` elements and the `<a-asset-item>` elements. Only the former would be awaited and subject to the timeout, the `<a-asset-item>` elements would use the normal 'wait till loaded' mechanism of `a-node.js`.

Besides the timeout not applying to `<a-asset-items>` it opened the door for race conditions where the timeout could cause a concurrent `super.load` resulting in two `loaded` events being emitted. This is the cause behind #5130  

This PR moves the `<a-asset-item>` handling into `a-assets` ensuring it waits on all assets in the same way. The timeout will now correctly cause the `loading` event being emitted without awaiting any (pending) assets.

Fixes #5130 

**Changes proposed:**
- Wait on `<a-asset-item>`s in `a-assets`
- Avoid race condition where `a-assets` would emit two `loaded` events

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -36,6 +36,7 @@ class AAssets extends ANode {
     var imgEl;
     var imgEls;
     var timeout;
+    var children;
 
     super.connectedCallback();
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

PR #5492, opened 2024-03-07
Remove attribute from DOM when removing component (through mixin removal/update)

mrxz:
**Description:**
Components on an entity are reflected in the DOM. In case an entities mixins change, some of its components might end up getting removed ([a-entity.js#L614-L615](https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L614-L615)). In this case the DOM attribute would remain. This PR ensures that, similar to `initComponent` the change is reflected in the DOM if not done already.

**Changes proposed:**
- Remove DOM attribute in case `removeComponent` is called without being caused by a `removeAttribute` call.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -379,6 +379,10 @@ class AEntity extends ANode {
     if (destroy) {
       component.destroy();
       delete this.components[name];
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

PR #5495, opened 2024-03-14
setup Screenshot  on getCanvas

arpu:
see https://github.com/aframevr/aframe/pull/5442#issuecomment-1997154460

@mrxz this should fix the Problem


dmarcos:
Thanks!

PR #5497, opened 2024-03-20
Update package.json shelljs dependabot Severity High 7.1 / 10

arpu:
dependabot

https://access.redhat.com/security/cve/cve-2022-0144

@dmarcos the shelljs is only used in the https://github.com/aframevr/aframe/blob/master/scripts/preghpages.js
is this still used?

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "markserv": "github:sukima/markserv#feature/fix-broken-websoketio-link",
     "mocha": "^10.0.0",
     "replace-in-file": "^2.5.3",
-    "shelljs": "^0.7.7",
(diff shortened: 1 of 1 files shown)


dmarcos:
Yeah it's used to deploy the examples. This is also a dev dependency. not bundled in the library

arpu:
dependabot do not like it :) i could not test it, but would be nice to include the update

dmarcos:
We gotta make sure there are no regressions. Also is a dev dependency that it's only used for A-Frame deployment. Pretty safe to ignore those warnings.

dmarcos:
I'll merge this and fingers crossed there's no regressions :) Thanks

Issue #4139, opened 2019-04-25
core:a-entity:warn Tried to remove `Object3D` of type:  mesh which was not defined.


asandersSAP:
**Description:**
On every entity loaded into the scene I get a console warning: 
`core:a-entity:warn Tried to remove `Object3D` of type:  mesh which was not defined.  aframe.min.js:23:633`

which is a mess with multiple entities. May you check? maybe it's just a small fix

Just tell me if you need more information
thanks!

- A-Frame Version:  0.9.1 
- Platform / Device: MacOS, every Browser

dmarcos:
If you provide a link with simple case to reproduce we can look into it. Glitch is a good option: https://glitch.com/~aframe

dmarcos:
Closing due to lack of link to reproduce

wyzwon:
@dmarcos
[Here's my go at what they meant.](https://issue-4139-demo.glitch.me/)

In this example the page uses an obj without a texture and sets a color at object creation. I was actually still getting a warning without the obj loaded as I had given the wrong file link and received a 404 error but the warning still appeared.

wyzwon:
Digging into the code a little, I have traced this back to when "update" is called in the 'obj-model' component. Update calls "resetMesh()" which calls "el.removeObject3D('mesh')" which evidently doesn't have a mesh. I don't know enough about this part of the code to know if update ever needs to reset a mesh or if it's safe to remove that line entirely.

PR #5499, opened 2024-03-22
Make sure the model was loaded before we attempt to remove the mesh

Elettrotecnica:
Fixes https://github.com/aframevr/aframe/issues/4139

I stumbled on this by having a look at a-painter to see how much it would take to have it running on latest aframe. The same warning would appear and seems the two lines at https://github.com/aframevr/a-painter/blob/0deba1f3314ee4cd053bf2a2ae2854994014dd65/index.html#L50-L51 are responsible.

This seems the same situation as per the example reported in https://github.com/aframevr/aframe/issues/4139#issuecomment-1371499138.

I basically moved the check for the model being present from the remove() to the resetMesh method.

All the best

### DIFF
--- a/src/components/obj-model.js
+++ b/src/components/obj-model.js
@@ -34,11 +34,11 @@ module.exports.Component = registerComponent('obj-model', {
   },
 
   remove: function () {
-    if (!this.model) { return; }
     this.resetMesh();
   },
 
   resetMesh: function () {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks so much

wyzwon:
I appreciate you taking the time to fix this!

Issue #5494, opened 2024-03-08
Utils device isIOS and isMobile checks fail on iPad Pro 12.9 Safari


IntraNoctem:
**Description:**

The utils device checks isIOS and isMobile fail on iPad Pro 12.9 when using Safari browser. In Chrome they succeed.

![iOS-isMobile](https://github.com/aframevr/aframe/assets/57447403/ea8250f3-a223-4af0-9e2d-9dec762ece13)

userAgent strings:
- 14.0: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1 Safari/605.1.15"
- 16.1: "Mozilla/5.0 (iPad; CPU OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/92.0.4515.90 Mobile/15E148 Safari/604.1"
- 17.1: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.1 Safari/605.1.15"

- A-Frame Version: 1.5.0
- Platform / Device:
  - iPad Pro 12.9 2021 (OS 14.0 and 17.1)
  - iPad Pro 12.9 2022 (iOS 16.1)

dmarcos:
Just curious. What you are trying to accomplish?

IntraNoctem:
There are some use cases in which it is necessary to know whether it is a mobile device, e.g. a help dialogue that describes either touch or mouse/keyboard gestures.

dmarcos:
Feel free to open an PR to fix it.

arpu:
some infos i found to this problem is https://github.com/pmndrs/detect-gpu/blob/master/src/internal/deviceInfo.ts#L15

dmarcos:
I'll take a PR if someone wants to fix it. Don't have access to the device to test. Thanks so much

PR #5502, opened 2024-03-24
Additional device check for Ipad Pro on Safari

Drkjr92:
**Description:**
Issue raised in [#5494](https://github.com/aframevr/aframe/issues/5494). 

The issue: iPadOS User Agent in Safari is same as on MacOS Catalina.

User case:
> There are some use cases in which it is necessary to know whether it is a mobile device, e.g. a help dialogue that describes either touch or mouse/keyboard gestures.

**Changes proposed:**
-  In `isTablet()` we want to extend the ipad check for those special cases where the user agent from the Ipad is labeled "MacOS"

Test also added and Results are shown below:

image

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -124,10 +124,10 @@ module.exports.isMobile = isMobile;
 function isTablet (mockUserAgent) {
   var userAgent = mockUserAgent || window.navigator.userAgent;
 
-  var isTablet = /ipad|Nexus (7|9)|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent);
+  var _isTablet = /ipad|Nexus (7|9)|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent);
 
   // Additional check for iPad or MacIntel with touch capabilities and not an MSStream device
-  return isTablet || isIpad();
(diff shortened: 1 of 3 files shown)


src/utils/device.js
```diff
@@ -123,10 +123,29 @@ module.exports.isMobile = isMobile;
  */
 function isTablet (mockUserAgent) {
   var userAgent = mockUserAgent || window.navigator.userAgent;
-  return /ipad|Nexus (7|9)|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent);
+
+  var _isTablet = /ipad|Nexus (7|9)|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent);
```
dmarcos, line 127: we don't use variable names with underscores. I also see the regular expression is checking for ipad. needed?

dmarcos: thanks

dmarcos: also wonder if this regex for tablet still provide good coverage. I see device versions hardcoded (e.g `Nexus 7|9`).

dmarcos:
Thanks so much! Congrats on your first contribution

Drkjr92:
Thank you!!

PR #5504, opened 2024-04-03
Fix toggleable dark/light mode button in hand-tracking example

vincentfretin:
**Description:**

In the button component, in the schema it's "toggable", this is what is used for the dark/light mode button.
The condition in onPressedStarted incorrectly use this.data.togabble instead of this.data.toggable, such the green color wasn't correctly toggled for the button.

**Changes proposed:**
- Fix the condition

### DIFF
--- a/examples/showcase/hand-tracking/button.js
+++ b/examples/showcase/hand-tracking/button.js
@@ -3,7 +3,7 @@ AFRAME.registerComponent('button', {
   schema: {
     label: {default: 'label'},
     width: {default: 0.11},
-    toggable: {default: false}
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks I think is misspelled in original code and fix :) Correct spell I think would be `toggleable`. Can we change it? Thanks

vincentfretin:
Indeed :) I'll look at modifying all occurrences later.

vincentfretin:
@dmarcos I modified the commit.

dmarcos:
Thanks so much!

PR #5508, opened 2024-04-05
use debug ^4.3.4 fixes #4612

### LINKED ISSUE #4612: Security vulnerability in 'debug' dependency
**Description:**
There is a security vulnerability in one of the dependencies of aframe. It looks like there is no update coming form the debug packages that is currently used. Maybe this package will help you out: https://github.com/visionmedia/debug

source:
https://snyk.io/test/npm/debug/2.6.8

- A-Frame Version: current
- Platform / Device: all

arpu:
using upstream debug ^4.3.4  

tested with http://localhost:8080/showcase/spatial-ui/
```
core:component:warn Unknown property `radius` for component `geometry`. +0ms
core:component:warn Unknown property `color` for component `material`. +0ms
```

old debug:

```
core:component:warn Unknown property `radius` for component `geometry`. 
core:component:warn Unknown property `color` for component `material`. 
```


dmarcos:
Yeah debug for was to remove the timestamps. Can we get rid of them? Thanks

arpu:
Yes, overwrite the debug.formatArgs, and add the colors and remove the Time works, Idea from   https://github.com/debug-js/debug/issues/582#issuecomment-1755718739

Original Code uses const/let https://github.com/debug-js/debug/blob/f66cb2d9f729e1a592e72d3698e3b75329d75a25/src/browser.js#L146 

@dmarcos should this be changed to vars?

src/utils/debug.js
```diff
@@ -10,18 +10,46 @@ var settings = {
...
+    (this.useColors ? '%c ' : ' ');
 
-  return d;
-};
-Object.assign(debug, debugLib);
+  if (!this.useColors) {
+    return;
+  }
+  this.color = getDebugNamespaceColor(this.namespace);
+  const c = 'color: ' + this.color;
```
dmarcos, line 31: no const or let. only vars

dmarcos:
Yes we don't use let / var. Code style has to be consistent with rest of code base.

dmarcos:
Thanks so much!

PR #5509, opened 2024-04-05
Added 'stencil' property to renderer system for enabling stencil buffer on WebGL context

mrxz:
**Description:**
Since Three.js version r163 the stencil buffer is disabled by default. Existing components that rely on it being present might now be broken, for example: https://github.com/mrxz/fern-aframe-components/issues/1. This PR introduces a `stencil` property on the renderer system, similar to the `alpha` property, which is passed to the `WebGLRenderer` constructor, ensuring the relevant buffer will be created on the WebGL context.

**Changes proposed:**
- Introduce `stencil` property to renderer system

### DIFF
--- a/docs/components/renderer.md
+++ b/docs/components/renderer.md
@@ -40,6 +40,7 @@ It also configures presentation attributes when entering WebVR/WebXR.
 | logarithmicDepthBuffer  | Whether to use a logarithmic depth buffer.                                      | auto          |
 | precision               | Fragment shader [precision][precision] : low, medium or high.                   | high          |
 | alpha                   | Whether the canvas should contain an alpha buffer.                              | true          |
+| stencil                 | Whether the canvas should contain a stencil buffer.                             | false         |
 | toneMapping             | Type of toneMapping to use, one of: 'no', 'ACESFilmic', 'linear', 'reinhard', 'cineon'  | 'no'          |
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks!

coderofsalvation:
the stencil-code of the https://xrfragment.org demoviewer is now in a catch22: if I enable `stencil:true`:

* it works in the desktop browser on my laptop (also in VR-mode via WebXR extension)
* randomly renders meshes on quest 2 in immersive-mode only

Any idea whether there some other factors at play here [besides bugs in my stencilcode]?

mrxz:
@coderofsalvation Three.js changed the default to no longer create a stencil buffer. This PR simply adds a way to re-enable it. So with `stencil: true` it should be how it was before.

The regression in your case is probably caused by a different change (possible in Three.js). Quickest way to find out is using `git bisect` to pinpoint which A-Frame commit introduced the problem for you.

coderofsalvation:
@mrxz thanks for reminding. So yes it `seemed` to work when I set `stencil: true` in desktop-mode, but not in immersive mode (because of  the latest included THREE-version I guess).
I've did a new build of the master-branch, and it seems it is no longer randomly rendering meshes [but there's still stencil-shenanigans in immersive mode on Quest only].
Thanks I will investigate further with `git bisect`.

PR #5512, opened 2024-04-08
Check exitVR property of keyboard-shortcuts to enable Escape shortcut

mrxz:
**Description:**
Noticed that the `keyboard-shortcuts` component only had an option to disable the `enterVR` shortcut, but it actually disabled both the enter and exit shortcuts. This doesn't match the documentation nor the user's expectation, so this PR splits it into a `enterVR` and `exitVR` property to configure the enter and exit shortcuts respectively.

**Changes proposed:**
- Introduce `exitVR` property to `keyboard-shortcuts` similar to the existing `enterVR` property

### DIFF
--- a/docs/components/keyboard-shortcuts.md
+++ b/docs/components/keyboard-shortcuts.md
@@ -14,7 +14,7 @@ keyboard-shortcuts component applies only to the [`<a-scene>` element][scene].
 ## Example
 
 ```html
-<a-scene keyboard-shortcuts="enterVR: false"></a-scene>
+<a-scene keyboard-shortcuts="enterVR: false; exitVR: false;"></a-scene>
 ```
 
 ## Properties
@@ -22,3 +22,4 @@ keyboard-shortcuts component applies only to the [`<a-scene>` element][scene].
 | Property    | Description                                           | Default Value |
 |-------------|-------------------------------------------------------|---------------|
 | enterVR     | Enables the shortcut to press 'F' to enter VR.        | true          |
+| exitVR      | Enables the shortcut to press 'Escape' to exit VR.    | true          |
--- a/src/components/scene/keyboard-shortcuts.js
+++ b/src/components/scene/keyboard-shortcuts.js
@@ -13,11 +13,6 @@ module.exports.Component = registerComponent('keyboard-shortcuts', {
     this.onKeyup = this.onKeyup.bind(this);
   },
 
-  update: function (oldData) {
-    var data = this.data;
-    this.enterVREnabled = data.enterVR;
-  },
-
   play: function () {
     window.addEventListener('keyup', this.onKeyup, false);
   },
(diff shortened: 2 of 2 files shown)


dmarcos:
Interesting. In what scenario did you notice the inconsistency? On desktop fullscreen, the esc key is the browser standard shortcut to exit. On mobile it doesn't apply. In headset each system has a different system to disengage immersive mode usually with controllers or hand tracking where keyboard is not avail.

mrxz:
I do occasionally use a Bluetooth keyboard with the Quest and also sometimes use PCVR to record WebXR sessions. In both cases the keyboard remains available to the user while in session. But that wasn't actually how I discovered it, that was simply through the immersive web emulator.

I customized the enter-vr experience and disabled the shortcut to prevent people from skipping it. At that point I noticed this also prevented me from hitting escape to exit the emulated WebXR session.

dmarcos:
Wonder if `esc` to exit VR should be a browser behavior. Similar to fullscreen. I think this PR is redundant for desktop / fullscreen?

dmarcos:
Are there scenarios where one wants one are not the other (exit / enter)?

mrxz:
> Wonder if esc to exit VR should be a browser behavior. Similar to fullscreen. I think this PR is redundant for desktop / fullscreen?

It's pretty much redundant for fullscreen on desktop (although the spec doesn't seem to prescribe Escape, just that ["User agents should provide a means of exiting fullscreen that always works"](https://fullscreen.spec.whatwg.org/#security-and-privacy-considerations))

> Are there scenarios where one wants one are not the other (exit / enter)?

It's not uncommon to have some steps before entering a session (e.g. selecting an avatar, typing a name, etc...), in which case just disabling the enter vr shortcut makes sense.

dmarcos:
> > Are there scenarios where one wants one are not the other (exit / enter)?
> 
> It's not uncommon to have some steps before entering a session (e.g. selecting an avatar, typing a name, etc...), in which case just disabling the enter vr shortcut makes sense.

For the sake of simplicity can we just keep it to one property? When concrete limitations arise we can add a second one. Not sure how much the shortcuts are actually used today.

mrxz:
> For the sake of simplicity can we just keep it to one property? When concrete limitations arise we can add a second one. Not sure how much the shortcuts are actually used today.

There are two distinct questions:

- Should either or both of the shortcuts be removed (as they might not be used)?
- Should it be possible to disable them individually or have one toggle?

While I can't comment on how often people use them, I do use exit-vr (esc) on a daily basis with the emulator. Wanting to disable the enter-vr shortcut lead to both being disabled, which contradicts both the name (`enterVR`) and the docs. From that POV, removing `exit-vr` harms DX and having only one property is a tad cumbersome.

vincentfretin:
I sometimes inadvertently press "f" because it's too close to "d" key.
On Chrome desktop (that have the VR button) it makes the experience full screen and you can't rotate, not great for a user that don't know what's going on.
On firefox (that have the fullscreen button) the "f" shortcut enter fullscreen and works great.
I usually disable the shortcut in my experiences and I have a custom VR button only showing if a headset is connected after some other ui screen to type the name and select avatar, so I didn't know the esc key wasn't working. Doing that, I know that I don't support VR from PC, it's ok for my use cases.
Having the option to disable the "f" key and still be able to use "esc" is good indeed.

dmarcos:
Thanks for this and for the patience

PR #5515, opened 2024-04-09
this.data may be undefined when the pressable component is removed

vincentfretin:
**Description:**

I just got a complete freeze in VR because of an error on line 
`if (distance < this.data.pressDistance) {` with this.data undefined.

I'm using the pressable component with a custom button component with an additional disabled property that is toggled on pressedstarted and remove the pressable component.
So the issue here is removing a component may still execute tick at least one time it seems. Should this happen? Is this a regression? I don't know.
This is with aframe master Apr 5, 2024 https://cdn.jsdelivr.net/gh/aframevr/aframe@98c706cc0e9d318742f03a20c39cb3b3cc36ce38/dist/aframe-master.min.js
I'm opening the PR mainly for discussion. I know there were recent changes in master related to removing component @mrxz

### DIFF
--- a/examples/showcase/hand-tracking/pressable.js
+++ b/examples/showcase/hand-tracking/pressable.js
@@ -11,6 +11,8 @@ AFRAME.registerComponent('pressable', {
   },
 
   tick: function () {
+    // this.data may be undefined when the pressable component is removed
+    if (!this.data) return;
     var handEls = this.handEls;
     var handEl;
     var distance;


vincentfretin:
It doesn't happen all the time, I had the issue one time over 30 tries I think. :) so it would be hard to test any previous build to see if I reproduce or not.

vincentfretin:
I'll push an example in a few days.

mrxz:
This is a bug, managed to reproduce it: https://glitch.com/edit/#!/aframe-5515-repro

Essentially ticks are called per component type. Removing components of the same type as is currently being processed is tricky. A-Frame would actually randomly skip `tick`/`tock` invocations (see #5400), but it now marks them for removal instead.

All of this works fine as long as either:
- Entire entities are removed
- The `tick`/`tock` of the removed components have already been processed

When the removed component(s) haven't had their `tick` called for that frame, this bug surfaces. Since the components have already been destroyed, their `data` is undefined. Should be fixable, though, just want to make sure it doesn't impact performance too much.

vincentfretin:
Thanks @mrxz for the quick investigation and fix!
I'm closing this, this wasn't intended to be merged.
#5516 is the real fix.

PR #5516, opened 2024-04-09
Check the isPlaying flag of the component when calling component behaviors

mrxz:
**Description:**
This PR fixes a regression introduced in #5478, which was noticed by @vincentfretin in #5515. In short, behaviours would be marked for removal when they belong to a component that is currently having their `tick`/`tock` called. This works fine in case an entity is removed or if the removed components have already been called, but fails if they are yet to be called.

This PR simply changes the check in `callComponentBehaviors` from `.el.isPlaying` to `.isPlaying` as components with behaviours have this flag. Generally it matches the `.el.isPlaying` state, unless the component hasn't (fully) initialized yet or is removed from an entity. Those are precisely the cases where the bug could surface.

The related unit tests required some changes in setup as they used Components that weren't completely initialized or inline defined behaviours.

Fixes/supersedes #5515

**Changes proposed:**
- Only call `tick` and `tock` on components that are currently playing

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -796,7 +796,7 @@ class AScene extends AEntity {
 
       behaviorSet.inUse = true;
       for (i = 0; i < behaviorSet.array.length; i++) {
-        if (!behaviorSet.array[i].el.isPlaying) { continue; }
+        if (!behaviorSet.array[i].isPlaying) { continue; }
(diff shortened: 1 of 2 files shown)


vincentfretin:
Thanks @mrxz you're the best!
I'm glad I found the issue myself and not during a demo with my customer :D

dmarcos:
Thanks so much!

PR #5521, opened 2024-05-02
Don't cache parsed value of selector(All) property types

mrxz:
**Description:**
This PR addresses a small regression introduced with #5474. The `selector` and `selectorAll` property types have a parse function that can be considered _impure_, i.e. the output depends on more than just the value being parsed. The current state of the DOM at the time of parsing factors in as well.

#5474 shifted the moment properties were parsed and cached in case of entities that hadn't loaded yet. This works fine as long as the selector can already match the relevant entities at that time. @arpu ran into a situation where a small HTML template was constructed in full _before_ being inserted into the scene. These could contain selectors that would refer to elements in the same snippet. For illustration, consider the following snippet being constructed dynamically post scene-load:
```HTML
<a-entity>
    <a-entity id="target"></a-entity>
    <a-entity component-with-selector="target: #target"></a-entity>
</a-entity>
```

To resolve this, the PR adds a `isCacheable` property to the property types which signals if the parsed value may be cached or not. This prevents the `selector/selectorAll` property types from being prematurely parsed and cached, better matching the old behaviour.

@arpu Could you verify that this solves the issue for you?

**Changes proposed:**
- Add `isCacheable` flag to property types `selector` and `selectorAll`
- Don't cache the parsed result in case a property isn't cacheable

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -426,7 +426,8 @@ Component.prototype = {
 
       // Parse the new value into attrValue (re-using objects where possible)
       var newAttrValue = key ? this.attrValue[key] : this.attrValue;
-      newAttrValue = parseProperty(newValue, propertySchema, newAttrValue);
+      // Some property types (like selectors) depend on external state (e.g. DOM) during parsing and can't be cached.
+      newAttrValue = propertySchema.isCacheable ? parseProperty(newValue, propertySchema, newAttrValue) : newValue;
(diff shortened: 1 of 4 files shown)


arpu:
i can confirm this fixed the problem to update vrland.io codebase to latest aframe master ( all tests a passed)

dmarcos:
Thank you!

Issue #5525, opened 2024-05-28
Geometry component warn


Roxcsas:
After updating to a-frame 1.6.0 i noticed all my a-entity that have geometry="..." and material="..." are causing this warning in the console even if they are working

> core:component:warn Unknown property `width` for component `geometry`
core:component:warn Unknown property `height` for component `geometry`
core:component:warn Unknown property `radius` for component `geometry`
core:component:warn Unknown property `color` for component `material`

dmarcos:
Thanks. Can you share an example with https://glitch.com/~aframe ?

Roxcsas:
> Thanks. Can you share an example with https://glitch.com/~aframe ?

here is an example
https://glitch.com/edit/#!/ambiguous-basalt-point

"side" is the only property of the ones i used which don't cause a warn

mrxz:
Surprised how this wasn't noticed before. The problem occurs with dynamic schemas, clobbering and style-string values, but that is the case when using the `geometry` or `material` components on `a-entity`. Using a mixin, primitive (e.g. `<a-box>`) or creating through code with objects doesn't trigger this.

In any case, I've created a fix for it: #5526

dmarcos:
Fixed in master. It will ship in next version but can use master builds in the meantime (link below):

https://cdn.jsdelivr.net/gh/aframevr/aframe@45652572bfd5b8939e30d5d504e62bae717795a5/dist/aframe-master.min.js

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5526

PR #5526, opened 2024-05-28
Fix incorrect unknown property warnings for components with dynamic schema

mrxz:
**Description:**
As reported in #5525 unknown property warnings can be logged, even when these properties exist for the component. This can occur when a component has a dynamic schema and the update clobbers using a style-string value.

This PR addresses the above issue by ensuring `recomputeData` behaves consistently when given either an object or a style-string value. Furthermore the silencing of unknown properties has been simplified and moved inside `updateSchemaIfNeeded`. In short unknown property warnings are always deferred with dynamic schemas (as a schema update might need to be processed), and the decision to silence them is handled by `updateSchemaIfNeeded` which knows whether or not the schema needed to updated.

Unit tests for this behaviour have been introduced as well, that should hopefully catch regressions in the future.

**Changes proposed:**
- Fix difference with regards to excess properties between object and style-string based updates
- Simplify deferred unknown property handling
- Introduce unit test verifying the unknown property warning message logging

### DIFF
--- a/src/core/component.js
+++ b/src/core/component.js
@@ -93,7 +93,6 @@ var Component = module.exports.Component = function (el, attrValue, id) {
 
   // Dynamic schema requires special handling of unknown properties to avoid false-positives.
   this.deferUnknownPropertyWarnings = !!this.updateSchema;
-  this.silenceUnknownPropertyWarnings = false;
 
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks!

PR #5529, opened 2024-05-29
Make isCompatibleTexture check stricter by requiring equal source

mrxz:
**Description:**
The compatibility check for textures allowed changes of the underlying source. Depending on the state of the texture in Three.js this might actually not be supported, causing incorrect reference counting, potentially leading to an exception upon disposal of the texture.

Instead of trying to inspect the internal state to make a decision, it's easier to make the check considerably stricter. This PR now requires the source to match the prior source. While this does lead to more texture instance being created when changing images, the corresponding `Source` (actual image data) and `WebGLTexture` resources remain shared whenever possible.

**Changes proposed:**
- Make `isCompatibleTexture` check stricter by requiring source to match previous source.


dmarcos:
Thanks!

PR #5542, opened 2024-06-15
Don't call raycaster setDirty on each mouse move unnecessarily

vincentfretin:
**Description:**

Before this PR, `setAttribute('raycaster', {origin, direction})` is called by the `cursor` component in `onMouseMove` triggering raycaster update method and `this.setDirty()` so actually calling `raycaster.refreshObjects()` on every tick with the default raycaster refresh interval that is 0, or every 100ms minimum if you set raycaster interval to 100 for example, even if nothing changed in the scene.

Some notes:
- cursor `this.el.setAttribute('raycaster', rayCasterConfig)` was introduced in 2017-07-18 22:28 in https://github.com/aframevr/aframe/commit/a4b91bcec6f2e222de8e98414456be6593416bab
- setDirty was introduced in 2017-09-19 20:26 in https://github.com/aframevr/aframe/commit/a463d6413f9ab20d126b0a70b2ae2be2645330a9 it was calling `refreshObjects` directly before that.

**Changes proposed:**
- In `raycaster` update, call `this.setDirty()` only if `objects` property changed

### DIFF
--- a/src/components/raycaster.js
+++ b/src/components/raycaster.js
@@ -131,7 +131,9 @@ module.exports.Component = registerComponent('raycaster', {
 
     if (oldData.enabled && !data.enabled) { this.clearAllIntersections(); }
 
-    this.setDirty();
+    if (data.objects !== oldData.objects) {
+      this.setDirty();
+    }
   },
 
   play: function () {


vincentfretin:
That issue is actually new in aframe 1.6.0. It wasn't an issue in 1.5.0.
So the `this.el.setAttribute('raycaster', rayCasterConfig)` in cursor component is now triggering the raycaster update function in 1.6.0. It wasn't the case in 1.5.0 @mrxz

vincentfretin:
You can test on the http://localhost:8080/showcase/ui/ example, put a console.log in raycaster refreshObjects function.

mrxz:
Prior to 1.6.0 the handling of instances of classes like `Vector3` when updating components had bugs (https://github.com/aframevr/aframe/issues/3681, https://github.com/aframevr/aframe/issues/5181). Passing in the same reference in subsequent `.setAttribute` calls would not trigger any update, as no changes would be detected, even if the values were actually changed.

Looking at the `cursor`, it does call `.setAttribute` with newly computed values, with the expectation that the `raycaster` gets updated. So `update` now being called with 1.6.0 is IMO the correct behaviour. That the `raycaster` always marks itself dirty regardless of what changed, seem to be a different bug that now surfaces, for which this PR seems like a good fix.

dmarcos:
Thanks so much!

PR #5543, opened 2024-06-15
update: magicwindow is false by default (look-controls.md)

coderofsalvation:
Somewhere AFRAME set this default to false, while the documentation (and my codebase) was still assuming true.


coderofsalvation:
nvm: error on my side

PR #5544, opened 2024-06-16
Add assetStringify for asset based properties

mrxz:
**Description:**
Asset based property types could either be a string, or directly refer to an `HTMLElement` (using an id selector, or passing instance to it). The stringification used for these property types was just the default one, which works for strings, but fails for `HTMLElement` instances.

This PR introduces an `assetStringify` that attempts to construct a suitable id selector. In case that isn't possible it will resort to using the `src` attribute instead.

See https://github.com/aframevr/aframe-inspector/issues/722 for some context where this caused issues when using `flushToDOM`.

**Changes proposed:**
- Introduce an `assetStringify` for asset based property types

### DIFF
--- a/src/core/propertyTypes.js
+++ b/src/core/propertyTypes.js
@@ -8,18 +8,18 @@ var nonCharRegex = /[,> .[\]:]/;
 var urlRegex = /url\((.+)\)/;
 
 // Built-in property types.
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

dmarcos:
Thanks so much!

PR #5545, opened 2024-06-26
Update coffee-spawner.js

addy1997:
`var wristRotation = evt.detail.wristRotation
`
declared but not used anywhere


dmarcos:
Thanks

PR #5550, opened 2024-06-28
event and function name corrected

addy1997:
pinchedmoved -> pinchmoved (as given in the aframe docs)
onPinchedMoved() -> onPinchMoved()


examples/showcase/hand-tracking/slider.js
```diff
@@ -46,10 +46,10 @@ AFRAME.registerComponent('slider', {
 
     this.el.appendChild(pickerEl);
 
-    pickerEl.addEventListener('pinchedmoved', this.onPinchedMoved);
+    pickerEl.addEventListener('pinchmoved', this.onPinchMoved);
```
vincentfretin, line 49: Is the example still working?
The event `pinchedmoved` event comes from `pinchable` component set by the `slider` component on `pickerEl` here. That's not related to `hand-tracking-controls` where we indeed have an event `pinchmoved`.

https://github.com/aframevr/aframe/blob/230d4f7724ce4b62560a21e1ab0ec022241ed3de/examples/showcase/hand-tracking/pinchable.js#L57

I'm not sure what's the difference between those events. The `pinchable` component may not be needed anymore?

addy1997: Oh! I thought it is the `pinchmoved` event coming from the `hand-tracking-controls`. Yes, the example should work fine. Neither do I know what is the difference.

PR #5551, opened 2024-06-29
Add AgX and neutral tone mappings to the list

vincentfretin:
**Description:**

In recent three.js version, we have now the new tone mapping THREE.AgXToneMapping and THREE.NeutralToneMapping

**Changes proposed:**
- Add 'AgX" and 'neutral' to the toneMapping list


dmarcos:
Thanks!

Issue #4635, opened 2020-06-29
Documentation: add a small example of event-driven animation


Cheaterman:
Hi, I hope you're doing good!

I'm admittedly an uber-noob, but I didn't realize how to fire the custom "animation__somename" animations in https://aframe.io/docs/1.0.0/components/animation.html .

I finally realized I needed to .emit() on the element thanks to https://stackoverflow.com/a/53412175 this kind soul.

Would it be possible to add something along the lines of that SO answer to the documentation?

Thanks a lot in advance, and thanks for the great framework!

Cheaterman:
While we're at it, this documentation says:

If the property is a vec3, that is also supported (e.g., property: someComponent.vec3Value; to: 5 5 5).

But to animate scale, I simply did "property: scale", I don't know if people are supposed to infer it somehow? But I feel like it might be a mistake. :-)

dmarcos:
PRs welcome

Cheaterman:
Hi Diego and thanks for your answer. Before I get started, could you point me to:

1°) The place in the repository where the relevant page of documentation lives ;
and
2°) Any documentation style guide or new contributor guide I should comply with for that PR ;

Thanks again.

dmarcos:
All docs live in the [docs directory](https://github.com/aframevr/aframe/tree/master/docs). For this probably maybe [animation component docs](https://github.com/aframevr/aframe/blob/master/docs/components/animation.md#animating-different-types-of-values) is the place. Maybe in your case I might be you didn't get to read the [events docs](https://aframe.io/docs/1.0.0/introduction/javascript-events-dom-apis.html#emitting-an-event-with-emit) before using animations?

Cheaterman:
Hi again Diego and thanks for your great answer!

First, touché, I indeed hadn't read the events docs! But secondly, maybe I'm just not familiar enough with JS, but to be honest with you I wouldn't necessarily have made the association between the .emit() used in the events docs, and the "startEvents:" used in the animations docs, if only for the fact that the mouseenter/mouseleave events aren't user-defined.

In any case I'll make the PR about the vec3value thing, but in light of my comment above, do you think it would be a good idea to add a second example with user-defined events?

Thanks again!

dmarcos:
[...]

PR #5552, opened 2024-06-30
Example code for Animating on custom events (#4635)

kewalkishang:
**Description:**
Need for example regarding the use of custom events to trigger animations in A-Frame. The current documentation does not explicitly show how to use the .emit() function to start animations based on custom events, which can be confusing for new users.

**Changes proposed:**
- Created a glitch example that demonstrates how to trigger animations using startEvents, pauseEvents, resumeEvents for custom events and then emitting them.
- Added a link to the glitch example in the animation doc.


dmarcos:
Thanks. The example uses an old version of A-Frame (1.6.0). Also I don't see any animations being triggered. I might be missing something

kewalkishang:
Hi @dmarcos, thanks for the review. 
Updated A-Frame to the latest version (Had made the example a while back).
The animation is on the cube. It changes its color from Red to blue in 2.5 seconds now(It was 5 seconds before). The idea was to give users ample time to pause the animation in between and try the pauseEvents and resumeEvents as well.
Do let me know how I can make the example even better.

Best, 
Kewal

dmarcos:
Thanks for the patience. I'm reluctant to add more glitches to the examples because those need maintenance too and easily got out of date. Could we handle this by adding a couple of lines and pointing to the custom events section of the docs? 

https://aframe.io/docs/1.6.0/introduction/javascript-events-dom-apis.html#emitting-an-event-with-emit

Thanks

kewalkishang:
Hey @dmarcos.
I see. Makes sense.
Added a simple custom event in the existing example and a link to the custom events section for reference.
[Changes](https://github.com/kewalkishang/aframe/blob/event-driven-animation-example/docs/components/animation.md#animating-on-events)
Do let me know if this is what you had in mind.

Thanks.

docs/components/animation.md
```diff
@@ -140,15 +140,16 @@ We can use the `startEvents` property to animate upon events:
   geometry="primitive: box"
   material="color: red"
   animation__mouseenter="property: components.material.material.color; type: color; to: blue; startEvents: mouseenter; dur: 500";
-  animation__mouseleave="property: components.material.material.color; type: color; to: red; startEvents: mouseleave; dur: 500";>
+  animation__mouseleave="property: components.material.material.color; type: color; to: red; startEvents: mouseleave; dur: 500";
+  animation__customevent="property: components.material.material.color; type: color; from: red; to: blue; startEvents: startanim001; dur: 500";>
```
dmarcos, line 144: can we change the event name? startanim001 not super friendly. What about 'triggeranimation`?

dmarcos:
Thanks so much for sticking with it!

PR #5555, opened 2024-07-11
Add workflows path to the scan

JarLob:
Doh, by filtering by `src` I actually excluded workflows from the scan. Fixing it.


dmarcos:
Thanks!

Issue #5505, opened 2024-04-03
Duplicate pinch events on the Quest Pro


zakaton:
Performing a single pinch emits multiple "pinchstarted" and "pinchended" events on the Quest Pro (I don't have any other headsets to test this on). [Here's a video of the issue comparing A-Frame 1.5.0 and 1.4.0](https://www.youtube.com/watch?v=3uQYY2OVRL0)

- A-Frame Version: 1.5.0
- Platform / Device: Quest Pro
- Reproducible Code Snippet or URL: [glitch code](https://glitch.com/edit/#!/a-frame-pinch-issue-quest-pro) and the [live version](https://a-frame-pinch-issue-quest-pro.glitch.me/)

```
<html>
  <head>
    <title>A-Frame pinch issue on Quest Pro</title>
    <script src="https://aframe.io/releases/1.5.0/aframe.min.js"></script>
  </head>

  <body>
    <a-scene>
      <a-entity id="rightHand" hand-tracking-controls="hand: right;"></a-entity>

      <a-camera>
        <a-entity position="0 0 -1">
          <a-text
          id="numberOfPinches"
          color="black"
          align="center"
          value="0 pinches"
        ></a-text>
        <a-plane color="white" position="0 0 -0.01" height="0.3" width="1.1"></a-plane>
        </a-entity>
        
        <a-sky hide-on-enter-ar color="grey"></a-sky>
      </a-camera>
    </a-scene>

    <script>
      const rightHand = document.getElementById("rightHand");
      let numberOfPinches = 0;
      const numberOfPinchesEntity = document.getElementById("numberOfPinches");
      rightHand.addEventListener("pinchstarted", (event) => {
        numberOfPinches++;
        console.log({ numberOfPinches });
        numberOfPinchesEntity.setAttribute(
          "value",
          `${numberOfPinches} pinches`
        );
      });
    </script>
  </body>
</html>

```

dmarcos:
Thanks. Super helpful if someone can also verify on Quest 2 / 3.

mrxz:
Very useful test, thanks! I can reproduce the behaviour on a Quest 3 and managed to bisect it to the following commit (8341329d7112116d127c62b8aa4ebefba0539124):
```
8341329d7112116d127c62b8aa4ebefba0539124 is the first bad commit
commit 8341329d7112116d127c62b8aa4ebefba0539124
Author: Diego Marcos Segura <diego.marcos@gmail.com>
[...]

PR #5565, opened 2024-08-15
POC: Reparent grabbed entity onto wrist object upon grab

mrxz:
**Description:**
Proof-of-concept demonstrating the simplified logic for grabbing when reparenting.
The grabbed element is reparented to the `wristObject` upon grab, and inversely restored to its original parent upon release.

Unrelated to this change:

- grabbing is a bit finicky (see https://github.com/aframevr/aframe/issues/5505)
- `hand-tracking-grab-controls` is broken due to a Three.js update:
```
2c70bc24596dd644084e66b0010dab5b1ff208ef is the first bad commit
commit 2c70bc24596dd644084e66b0010dab5b1ff208ef
Author: Diego Marcos <diego.marcos@gmail.com>
Date:   Thu Jul 25 18:25:35 2024 -0700

    Bump THREE to r167

 package.json | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
 ```

### DIFF
--- a/src/components/hand-tracking-grab-controls.js
+++ b/src/components/hand-tracking-grab-controls.js
@@ -24,16 +24,6 @@ registerComponent('hand-tracking-grab-controls', {
     el.setAttribute('obb-collider', {trackedObject3D: trackedObject3DVariable, size: 0.04});
 
     this.auxMatrix = new THREE.Matrix4();
-    this.auxQuaternion = new THREE.Quaternion();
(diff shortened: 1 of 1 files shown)


mrxz:
@dmarcos Rebased and updated this PR so that it no longer stores a temp variable on the grabbed entity. No longer marking as a draft as I think it's worth fixing `hand-tracking-grab-controls` which is currently broken on master.

dmarcos:
Thanks!

PR #5568, opened 2024-08-26
Fix error link

jzhangdev:
**Description:**

The WebXR Mesh Detection Module documentation uses an error link.


dmarcos:
Thanks so much!

PR #5574, opened 2024-09-07
Add aframe-injected attribute to the created a-loader-title div

vincentfretin:
Add aframe-injected attribute to the created a-loader-title div so it doesn't show up when copying a-scene to clipboard in aframe-inspector (fix aframevr/aframe-inspector#764)


dmarcos:
Thanks!

Issue #5546, opened 2024-06-26
Remove renderer physicallyCorrectLights property and update default lighting setup (r165)


vincentfretin:
**Description:**

- A-Frame Version: aframe master (post 1.6.0) with r165 and above
- Platform / Device: all

`WebGLRenderer.useLegacyLights` was removed in three r165, default scene now looks much darker.
We really should've defaulted to useLegacyLights=false (physicallyCorrectLights true in aframe renderer) when we switched the default for colorManagement to true in aframe 1.5.0, but that's done we can't go back in the past.

We should:

- Remove the renderer physicallyCorrectLights property that is now always true. No way to put that to false anymore.
- Update the default lighting setup to be roughly the same than when physicallyCorrectLights was false
- Document the breaking change for aframe 1.7.0, users will have to adjust their own lighting
- In parallel, we may change the lights intensity in the environment component for all presets as well and do a minor release 1.4.0

mrxz:
For context: https://discourse.threejs.org/t/updates-to-lighting-in-three-js-r155/53733#migrating-2:

> The intensities for ambient, hemisphere, directional lights and light maps **can be restored by multiplying PI** with the existing light intensity values.

Also, since things won't be the exact same as before, let's also update the default `decay` value for lights to 2 (matching Three.js defaults). That way the A-Frame defaults should be correct for physically accurate lighting out of the box:  
> It’s important to understand that using the new lighting mode is just one prerequisite for physically correct lighting. You also have to:
>    * apply a real-world scale to your scene (meaning 1 world unit = 1 meter).
>    * **not change the default decay values of 2 for all spot and point lights in your scene.**

dmarcos:
Anyone wants to open a PR?

vincentfretin:
Each point I mentioned can be a separate PR, but the two first points are probably good to do together. That may be a good first time contribution if anyone wants to contribute.

@mrxz maybe you know, is there maybe some formula we can apply to find the equivalent light color to have more or less the same rendering?

mrxz:
> @mrxz maybe you know, is there maybe some formula we can apply to find the equivalent light color to have more or less the same rendering?

The colour itself does not need to be changed, only the intensity by a factor of PI. Since the decay/falloff is different the results won't be identical, but there isn't really anything that can be done about that. Though I expect that the difference won't really be noticeable.

dmarcos:
[...]

PR #5577, opened 2024-09-17
Reintroduce light intensity scale factor removed in THREE r165 after WebGLRenderer.useLegacyLights deprecation (fix #5556, #5546)

### LINKED ISSUE #5556: Colors against dark in master vs 1.6.0. Probably related to THREE changes

dmarcos:
Is this good? @vincentfretin @mrxz

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -20,7 +20,7 @@ module.exports.Component = registerComponent('light', {
     color: {type: 'color', if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     envMap: {default: '', if: {type: ['probe']}},
     groundColor: {type: 'color', if: {type: ['hemisphere']}},
-    decay: {default: 1, if: {type: ['point', 'spot']}},
+    decay: {default: 2, if: {type: ['point', 'spot']}},
     distance: {default: 0.0, min: 0, if: {type: ['point', 'spot']}},
     intensity: {default: 1.0, min: 0, if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot', 'probe']}},
     penumbra: {default: 0, min: 0, max: 1, if: {type: ['spot']}},
@@ -281,7 +281,11 @@ module.exports.Component = registerComponent('light', {
     var distance = data.distance;
     var groundColor = new THREE.Color(data.groundColor);
     groundColor = groundColor.getHex();
-    var intensity = data.intensity;
+    // WebGLRenderer.useLegacyLights has been removed in THREE r165
+    // Lights intensity had an implicit scaleFactor of PI now removed.
+    // It's reintroduced here since it's easier to think about integers vs multiples of PI.
+    var intensityScaleFactor = Math.PI;
+    var intensity = data.intensity * intensityScaleFactor;
     var type = data.type;
     var target = data.target;
     var light = null;
--- a/docs/components/light.md
+++ b/docs/components/light.md
@@ -97,7 +97,7 @@ Directional lights are the most efficient type for adding realtime shadows to a 
 <a-light type="directional" light="castShadow:true;" position="1 1 1" intensity="0.5" shadow-camera-automatic="#objects"></a-light>
 ```
 
-The `shadow-camera-automatic` configuration maps to `light.shadowCameraAutomatic` which tells the light to automatically update the shadow camera to be the minimum size and position to encompass the target elements. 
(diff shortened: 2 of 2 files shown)


dmarcos:
Colors of the hello world look same as before after applying the intensity scale factor

src/components/light.js
```diff
@@ -20,7 +20,7 @@ module.exports.Component = registerComponent('light', {
     color: {type: 'color', if: {type: ['ambient', 'directional', 'hemisphere', 'point', 'spot']}},
     envMap: {default: '', if: {type: ['probe']}},
     groundColor: {type: 'color', if: {type: ['hemisphere']}},
-    decay: {default: 1, if: {type: ['point', 'spot']}},
+    decay: {default: 2, if: {type: ['point', 'spot']}},
```
vincentfretin, line 23: Why did you change decay?

src/components/light.js
```diff
@@ -281,7 +281,11 @@ module.exports.Component = registerComponent('light', {
     var distance = data.distance;
     var groundColor = new THREE.Color(data.groundColor);
     groundColor = groundColor.getHex();
-    var intensity = data.intensity;
+    // WebGLRenderer.useLegacyLights has been removed in THREE r165
+    // Lights intensity had an implicit scaleFactor of PI now removed.
+    // It's reintroduced here since it's easier to think about integers vs multiples of PI.
+    var intensityScaleFactor = Math.PI;
+    var intensity = data.intensity * intensityScaleFactor;
```
vincentfretin, line 288: It shouldn't change the intensity for scene that used `renderer="physicallyCorrectLights: true"` already.

vincentfretin (commented):
We could keep the physicallyCorrectLights property on renderer just to add a condition in the light component. But we should remove the lines https://github.com/aframevr/aframe/blob/2aca98cbbc5f1f03431ce07f107709baa1718df8/src/systems/renderer.js#L38-L40 and modify the tests.

#5546 for context.

src/components/light.js
dmarcos: This is @mrxz suggestion in https://github.com/aframevr/aframe/issues/5546#issuecomment-2191818981

src/components/light.js
dmarcos: Not sure I understand the comment. Without the factor scene is dark. What should we do?

dmarcos:
> We could keep the physicallyCorrectLights property on renderer just to add a condition in the light component. But we should remove the lines
> 
> https://github.com/aframevr/aframe/blob/2aca98cbbc5f1f03431ce07f107709baa1718df8/src/systems/renderer.js#L38-L40
> 
> and modify the tests.
> #5546 for context.

I think should be fine to remove the `physicallyCorrectLights` property in the renderer?

src/components/light.js
vincentfretin: We have aframe experiences with renderer="physicallyCorrectLights:true" and light intensity that was carefully set, those scenes are not dark, we don't want the intensity to change for those.

mrxz:
As @vincentfretin pointed out, there are already scenes that have `physicallyCorrectLights` enabled and should not have their intensities scaled. If we get rid of the property on `renderer`, then there won't be a way to distinguish the two.

Besides that, applying the scaling only in the `light` component will miss lights created in other ways. Components might directly create lights or a glTF file might include them. Even the `reflection` component manually sets light intensities, which would need to be updated as well ([reflection.js#L20](https://github.com/aframevr/aframe/blob/master/src/components/scene/reflection.js#L20)).

Instead of this PR, I think we should either:

1. Follow Three.js and deprecate the "legacy lights" entirely. The default light setup and things like the environment-component can be updated. Document what users need to do to upgrade their scenes.
2. Retain the old behaviour by re-introducing `useLegacyLights` in `super-three`. This ensures that everything behaves like before, instead of only the `light` component.

Personally not a fan of diverging from Three.js, but forcing upgrading users to make changes with no way to opt-out is also far from ideal.

dmarcos:
Leaning towards deprecation

dmarcos:
I multiplied the light defaults and values in examples by 3.14

https://github.com/aframevr/aframe/commit/8813d3d2818634ed4900409d99544b60587aa300

Also updated the environment component defaults and published 1.4.0

Any other outstanding 3rd party components that we should adjust?

vincentfretin:
@dmarcos can you please also remove the physicallyCorrectLights option from the renderer system?

dmarcos:
@vincentfretin done. thanks 🫡

PR #5579, opened 2024-09-20
Use generic controller instead of oculus specific for comic book sample

idrisshah:
Check for invalid texture in draw call

This is for Issue 5578.

**Changes proposed:**
-
-
-

### DIFF
--- a/examples/showcase/comicbook/index.html
+++ b/examples/showcase/comicbook/index.html
@@ -30,8 +30,8 @@
       </a-assets>
       <a-entity
         layer="src: #page1; height: 3; width: 1.9505"></a-entity>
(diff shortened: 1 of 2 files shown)


dmarcos:
laser-controls don't have utility here and are designed for laser pointer kind of interactions. If you want to support controllers other than quest you can just add other headset specific controllers to the entity. What's your case?

For the texture issue I recommend opening a separate PR. Thanks

dmarcos:
Left some comments on the issue https://github.com/aframevr/aframe/issues/5578

Let me know if you have any questions

dmarcos:
Closing due to inactivity

PR #5451, opened 2024-02-08
Remove WebVR support and polyfill

mrxz:
**Description:**
This PR removes WebVR support and WebVR polyfill (#5447), leaving only the WebXR code paths. This allows quite a lot of code to be removed and simplified. Especially the various `device-controls` components. While making the adjustments to the controllers and their unit tests it became painfully clear that there are many inconsistencies in the way controllers are handled. Many of them also exhibit changed behaviour compared to WebVR (e.g. naming of buttons changed), which showed as regressions in the unit tests.

Marked as a draft as this PR will require extensive reviewing/testing. Seeing the poor state of the controller components and their associated tests, I'll probably create separate PRs that focus on moving the unit tests over to WebXR, followed by simplifying/fixing the controller components, which will leave this PR purely with the WebVR removal.

**Changes proposed:**
- Remove WebVR polyfill
- Remove WebVR code paths/support
- Update all `controls` components to only support/use WebXR
- Changed unit test to test WebXR behaviour, while removing ones the only applied to WebVR

PR #5581, opened 2024-09-28
Gracefully handles when calling getGamepads() is not allowed...

DougReeder:
… from the context (such as an IFRAME), by the Document's Permission Policy, by catching the DOMException.

When AFrame is run in an IFRAME without the needed permission, the call to Navigator.getGamepads() throws a DOMException which prevents AFrame from loading properly.

### DIFF
--- a/src/systems/tracked-controls-webvr.js
+++ b/src/systems/tracked-controls-webvr.js
@@ -64,7 +64,11 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
         this.el.emit('controllersupdated', undefined, false);
       }
     } catch (e) {
-      console.warn('A-Frame requires additional permissions to list the gamepads. If this is running in an IFRAME, you need to add `gamepads` to the `allow` attribute. If this is running as the top-level page, the HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+      if (e.name === 'SecurityError') {
+        console.warn('A-Frame requires additional permissions to list the gamepads. If this is running in an IFRAME, you need to add `gamepads` to the `allow` attribute. If this is running as the top-level page, the HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+      } else {
+        console.error('Can\'t update controller list:', e);
+      }
     }
   }
 });
--- a/src/systems/tracked-controls-webvr.js
+++ b/src/systems/tracked-controls-webvr.js
@@ -64,7 +64,7 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
         this.el.emit('controllersupdated', undefined, false);
       }
     } catch (e) {
-      console.warn('can\'t update controller list:', e);
+      console.warn('A-Frame requires additional permissions to list the gamepads. If this is running in an IFRAME, you need to add `gamepads` to the `allow` attribute. If this is running as the top-level page, the HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
     }
   }
 });
--- a/src/systems/tracked-controls-webvr.js
+++ b/src/systems/tracked-controls-webvr.js
@@ -65,7 +65,11 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
       }
     } catch (e) {
       if (e.name === 'SecurityError') {
-        console.warn('A-Frame requires additional permissions to list the gamepads. If this is running in an IFRAME, you need to add `gamepads` to the `allow` attribute. If this is running as the top-level page, the HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+        if (window.self === window.top) {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+        } else {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The iframe `allow` attribute must not block this origin.', e);
+        }
       } else {
         console.error('Can\'t update controller list:', e);
       }
--- a/src/systems/tracked-controls-webvr.js
+++ b/src/systems/tracked-controls-webvr.js
@@ -41,26 +41,30 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
    * Update controller list.
    */
   updateControllerList: function () {
-    var controllers = this.controllers;
-    var gamepad;
-    var gamepads;
-    var i;
-    var prevCount;
+    try {
+      var controllers = this.controllers;
+      var gamepad;
+      var gamepads;
+      var i;
+      var prevCount;
 
-    gamepads = navigator.getGamepads && navigator.getGamepads();
-    if (!gamepads) { return; }
+      gamepads = navigator.getGamepads && navigator.getGamepads();
+      if (!gamepads) { return; }
 
-    prevCount = controllers.length;
-    controllers.length = 0;
-    for (i = 0; i < gamepads.length; ++i) {
-      gamepad = gamepads[i];
-      if (gamepad && gamepad.pose) {
-        controllers.push(gamepad);
+      prevCount = controllers.length;
+      controllers.length = 0;
+      for (i = 0; i < gamepads.length; ++i) {
+        gamepad = gamepads[i];
+        if (gamepad && gamepad.pose) {
+          controllers.push(gamepad);
+        }
       }
-    }
 
-    if (controllers.length !== prevCount) {
-      this.el.emit('controllersupdated', undefined, false);
+      if (controllers.length !== prevCount) {
+        this.el.emit('controllersupdated', undefined, false);
+      }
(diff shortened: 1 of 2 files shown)


dmarcos:
warn message is not super useful. Can we just detect we're in an iframe environment and show a more descriptive message?

DougReeder:
If the IFRAME is something like `<iframe src="https://other.com/game" allow="gamepad"></iframe>`, it will have permission to get gamepads.  There are other permissions an IFRAME might or might not have, such as accelerometer, fullscreen, gyroscope, and xr-spatial-tracking. [Permissions-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy)  These permissions can also be controlled for the top-level context (the page).   So it's not clear to me that whether or not we're in an IFRAME is useful info.

What error message would you consider useful?

dmarcos:
> What error message would you consider useful?

Not sure. Something that helps the user understand what's going on and how to fix it. e.g:

```
You are running A-Frame inside an iframe that requires additional permissions to access some APIs . In order to access the gamepads API you need to add `gamepads` to the allow attribute
```

dmarcos:
left a comment in PR

src/systems/tracked-controls-webvr.js
```diff
@@ -64,7 +64,11 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
         this.el.emit('controllersupdated', undefined, false);
       }
     } catch (e) {
-      console.warn('A-Frame requires additional permissions to list the gamepads. If this is running in an IFRAME, you need to add `gamepads` to the `allow` attribute. If this is running as the top-level page, the HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+      if (e.name === 'SecurityError') {
```
dmarcos, line 67: is this specific enough? We shouldn't check if we are in an iframe?

https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t

DougReeder: No other code inside the try-catch can throw a SecurityError.

src/systems/tracked-controls-webvr.js
```diff
@@ -41,26 +41,38 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
...
-      this.el.emit('controllersupdated', undefined, false);
+      if (controllers.length !== prevCount) {
+        this.el.emit('controllersupdated', undefined, false);
+      }
+    } catch (e) {
+      if (e.name === 'SecurityError') {
+        if (window.self === window.top) {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
+        } else {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The iframe `allow` attribute must not block this origin.', e);
```
dmarcos, line 71: Message needs rewording. A-Frame doesn't require additional permissions but the iframe permissions policy. Is not the solution to add gamepad to the allow attribute e.g `allow="gamepad"`? allow doesn't bock origins afaik?

src/systems/tracked-controls-webvr.js
```diff
@@ -41,26 +41,38 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
...
 
-    if (controllers.length !== prevCount) {
-      this.el.emit('controllersupdated', undefined, false);
+      if (controllers.length !== prevCount) {
+        this.el.emit('controllersupdated', undefined, false);
+      }
+    } catch (e) {
+      if (e.name === 'SecurityError') {
+        if (window.self === window.top) {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The HTTP `Permissions-Policy` header must not exclude this origin in the `gamepad` directive.', e);
```
dmarcos, line 69: Same for this message. How common is this issue? I think I've never seen a server excluding the gamepad API

src/systems/tracked-controls-webvr.js
```diff
@@ -41,26 +41,38 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
...
+        } else {
+          console.warn('A-Frame requires additional permissions to list the gamepads. The iframe `allow` attribute must not block this origin.', e);
+        }
+      } else {
+        console.error('Can\'t update controller list:', e);
```
dmarcos, line 74: why is this error and above just warnings?

src/systems/tracked-controls-webvr.js
DougReeder: allow="gamepad" is equivalent to allow="gamepad 'self'" which is fine if the A-Frame app is at the same origin as the top-level page.  If the A-Frame app is  on a different origin (for example, if the aframe app is on itch.io), you need allow="gamepad aframe-app-origin.com" or allow="gamepad *"

src/systems/tracked-controls-webvr.js
DougReeder: I haven't seen this in the wild, but it would exist if an aframe app was deployed from an server with a very restrictive policy (as set in the headers, which the a-frame app can't control).

src/systems/tracked-controls-webvr.js
DougReeder: If the A-frame app doesn't need the controllers, failing to get the controllers will not keep the app from working. If the a-frame app does need controllers, not getting the controllers is a problem that has to be fixed. Since this code can't determine which, a warning is reasonable.  (However, if the error isn't caught at all, A-Frame doesn't fully initialize.)

"Can\'t update controller list" would only occur if there was some unforeseen error and would likely be a programming error in A-Frame.

src/systems/tracked-controls-webvr.js
```diff
@@ -41,26 +41,38 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
...
-      this.el.emit('controllersupdated', undefined, false);
+      if (controllers.length !== prevCount) {
+        this.el.emit('controllersupdated', undefined, false);
+      }
+    } catch (e) {
+      if (e.name === 'SecurityError') {
+        if (window.self === window.top) {
+          console.warn('The HTTP `Permissions-Policy` header must not block this origin in the `gamepad` directive, to allow A-Frame to list the gamepads.', e);
+        } else {
+          console.warn('The iframe `allow` attribute must not block the origin of this A-Frame app, to allow A-Frame to list the gamepads.', e);
```
dmarcos, line 71: This message is still not clear. I wouldn't know what to do and I'm well versed in web dev. What about?

```
It looks like your A-Frame experience is running inside an iframe. In order to access the gamepads API you need to add allow="gamepad" to your iframe attributes
```

src/systems/tracked-controls-webvr.js
```diff
@@ -41,26 +41,38 @@ module.exports.System = registerSystem('tracked-controls-webvr', {
...
 
-    if (controllers.length !== prevCount) {
-      this.el.emit('controllersupdated', undefined, false);
+      if (controllers.length !== prevCount) {
+        this.el.emit('controllersupdated', undefined, false);
+      }
+    } catch (e) {
+      if (e.name === 'SecurityError') {
+        if (window.self === window.top) {
+          console.warn('The HTTP `Permissions-Policy` header must not block this origin in the `gamepad` directive, to allow A-Frame to list the gamepads.', e);
```
dmarcos, line 69: We can remove this. Seems pretty rare. Can add later if necessary

dmarcos:
What is the general context of this error? Are you still using the webvr code path?

DougReeder:
See [this example of the current code failing to load properly](https://fog-glowing-september.glitch.me/)

`examples/test/iframe/index.html` in the PR has been modified to show the PR code gracefully handling the lack of gamepad permission.

dmarcos:
Two thoughts here

1. Don't think we should check for gamePads if the scene is not in VR mode. I would add a guard here to check for `this.el.is('vr-mode')` here https://github.com/aframevr/aframe/blob/master/src/systems/tracked-controls-webvr.js#L50

It should get rid of the error message

3. For next version we should deprecate the WebVR path and all of this will be a non-issue. There's a WIP PR open https://github.com/aframevr/aframe/pull/5451

DougReeder:
Adding that check does allow A-Frame to start up, but as soon as the user enters vr-mode, the Uncaught SecurityError prevents tick() from running normally and rendering doesn't happen, unless we catch the SecurityError.

dmarcos:
I merged a PR to remove WebVR support. This should no longer be an issue. Can you give master builds a try and let me know? 

https://cdn.jsdelivr.net/gh/aframevr/aframe@ce9fd2321c97daf01bf9f098ad8b625c1603bc63/dist/aframe-master.min.js

Thanks!

DougReeder:
With that version of A-Frame, I do not observe the issue, either on startup or in the tick() handler.

dmarcos:
Thanks for the PR, the back and forth and verifying that master is now working. Now that we have deprecated WebVR we no longer need this PR

PR #5586, opened 2024-10-18
Update index.html

rajeshwariramya:
I removed the meta refresh tag, as it's not necessary for a basic HTML template. If you want to add any content to the body

**Description:**

**Changes proposed:**
-
-
-

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="refresh" content="0; url=examples/">
   </head>
-  <body></body>
+  <body>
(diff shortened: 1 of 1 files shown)


dmarcos:
@vincentfretin Do we still need the index file with webpack?

vincentfretin:
Indeed this file is not used anymore, it was only for budo to do the redirect to examples it seems. With webpack it serves examples directly as the root.
You can remove the file.

dmarcos:
@rajeshwariramya you want to remove the file? thanks

dmarcos:
I removed the file. Thanks

PR #5591, opened 2024-10-23
Camera docs: switch from var to const

nikolas:


dmarcos:
We use ES5 syntax accross A-Frame code base. thanks

PR #5592, opened 2024-10-23
Update github ci to ubuntu 24

nikolas:


dmarcos:
Why is this change necessary?

dmarcos:
In general we don't update dependencies versions unless there's a reason to do so

dmarcos:
Closing this for now. we can reopen if necessary. thanks

PR #5597, opened 2024-10-26
Update a-video.md with `muted` attribute on <video/> for autoplay

4www:
**Description:**

**Changes proposed:**
- update the docs with `muted` on the `<video/>` element to start autoplaying. Non a-frame, but helps figure out why a video does not autoplay video only `autoplay`


dmarcos:
This is more appropriate for the FAQ. We already have some relevant info:

https://aframe.io/docs/1.6.0/introduction/faq.html#why-does-my-video-not-play

Feel free to complete expand. Thanks

PR #5606, opened 2024-11-14
Cursor component: improve xrselect rayOrigin logic

mrxz:
**Description:**
Continuation of #5528 adding logic to convert the XRPose coming from WebXR into world space. Intended usage of `cursor="rayOrigin: xrselect"` is to place it on the `a-scene` (similar to `rayOrigin: mouse`), but can also be placed elsewhere as the logic no longer depends on the entity it is on.

**Changes proposed:**
- Add and remove WebXR event listeners when changing `rayOrigin` type to/from `xrselect` and when entering/leaving XR
- Convert the WebXR pose to world space accounting for any camera rig
- Handle the case no pose is available (before tracking is established or when temporarily lost)
- Fixed a bug where a `selectend` from a different input source would incorrectly reset the `activeXRInput` freezing the active input source.

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -255,6 +255,7 @@ module.exports.Component = registerComponent('cursor', {
     return function (evt) {
       var bounds = this.canvasBounds;
       var camera = this.el.sceneEl.camera;
+      var cameraElParent = camera.el.object3D.parent;
       var left;
       var point;
       var top;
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

dmarcos:
Tested this on Vision Pro and seems to work as expected. Thanks

PR #5607, opened 2024-11-14
Add additional isWebXRAvailable check to isAppleVisionPro to not have a false positive on iPhones / iPads (fix #5605)

### LINKED ISSUE #5605: isMobileVR() erroneously return true on iPad since AVP support
**Description:**

- A-Frame Version: master https://github.com/aframevr/aframe/commit/189993048b4f330872bac0c243a7a3de7415bb0d
  https://cdn.jsdelivr.net/gh/aframevr/aframe@3a633d512b90724d856040676e575b7e190110a8/dist/aframe-master.min.js 
- Platform / Device: iPad
- Reproducible Code Snippet or URL: https://aframe-is-mobile-vr.glitch.me

`AFRAME.utils.device.isMobileVR()` erroneously return true on iPad (probably also on iPhone) since the changes to support Apple Vision Pro in https://github.com/aframevr/aframe/commit/c590f85b30e0577442e4bf73c3f5fdedf1f21eab. It returned false on aframe 1.6.0

I'm not sure how to fix it on aframe side.

Currently in my solidjs project I'm using a headsetConnected variable/signal that I set like this:
```js
  const [headsetConnected, setHeadsetConnected] = createSignal(false);
  const [supportsVRSession, setSupportsVRSession] = createSignal(false);
  createEffect(() => {
    const errorHandler = function (err: Error) {
      console.error("WebXR session support error: " + err.message);
    };
    if (navigator.xr !== undefined && navigator.xr.isSessionSupported) {
      // Current WebXR spec uses a boolean-returning isSessionSupported promise
      navigator.xr
        .isSessionSupported("immersive-vr")
        .then(function (supported) {
          setSupportsVRSession(supported);
        })
        .catch(errorHandler);
    }
  });

  createEffect(() => {
    // We don't use AFRAME.utils.device.checkHeadsetConnected() because it set supportsVRSession internally after
    // the promise is resolved so the checkHeadsetConnected() call may be done before the promise is resolved.
    setHeadsetConnected(supportsVRSession() && AFRAME.utils.device.isMobileVR());
  });
```

where I add an additional check to see if the browser supports a vr session.
I don't use `AFRAME.utils.device.checkHeadsetConnected()` or `checkVRSupport()` because those are not reliable if you call it too early, the `navigator.xr.isSessionSupported` promise may not have been resolved yet, I actually had that issue.

Maybe we should expose `checkVRSupportAsync` that return a promise? Just a thought.

vincentfretin:
Now that WebXR is enabled by default on Apple Vision Pro, we can add a check for navigation.xr to discriminate AVP and iPhones / iPads.

This fixes #5605

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -127,9 +127,8 @@ function isAppleVisionPro () {
   // Safari for Apple Vision Pro presents itself as a desktop browser.
   var isMacintosh = navigator.userAgent.includes('Macintosh');
   // Discriminates between a "real" desktop browser and Safari for Vision Pro.
-  // Note: need to check for posible false positives on iPhones / iPads.
   var hasFiveTouchPoints = navigator.maxTouchPoints === 5;
-  return isMacintosh && hasFiveTouchPoints;
+  return isMacintosh && hasFiveTouchPoints && isWebXRAvailable;
 }
 module.exports.isAppleVisionPro = isAppleVisionPro;
 
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -128,6 +128,8 @@ function isAppleVisionPro () {
   var isMacintosh = navigator.userAgent.includes('Macintosh');
   // Discriminates between a "real" desktop browser and Safari for Vision Pro.
   var hasFiveTouchPoints = navigator.maxTouchPoints === 5;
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #5608, opened 2024-11-14
Fix render order of background fade in User Interface example

mrxz:
**Description:**
The background fade effect in the UI showcase was broken. This can be traced back to the change in sorting made in #5341. The example assumed the document order was used for rendering, which would place the `fadeBackground` element between the movie posters and the info panel.

However, objects are now sorted, putting all opaque objects in order followed by all transparent objects (unsorted). This placed the `fadeBackground` after both the movie posters and the info panel.

To resolve this issue, this PR changes the `fadeBackground` to not be transparent, but use subtractive blending. This allows it to be part of the opaque queue. The render order can then be configured as follows: 0 (background + movie posters), 1 (fadeBackground), 2 (info panel), transparent elements.

**Changes proposed:**
- Change `fadeBackground` to use subtractive blending instead of transparency and set desired renderOrder.

### DIFF
--- a/examples/showcase/ui/index.html
+++ b/examples/showcase/ui/index.html
@@ -73,7 +73,7 @@
         <a-entity
           id="fadeBackground"
           geometry="primitive: sphere; radius: 2.5"
-          material="color: black; side: back; shader: flat; transparent: true; opacity: 0.6" visible="false">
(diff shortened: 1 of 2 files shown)


dmarcos:
thanks!

PR #5609, opened 2024-11-14
Use simple pinch end threshold avoiding repeated pinch starts

mrxz:
**Description:**
A partial revert of 8341329d7112116d127c62b8aa4ebefba0539124 fixing #5505.

Having a fixed pinch end distance that is larger than pinch start should avoid any repeated pinch start events. The original change was made to reduce/avoid a pinch being too "sticky", so I've kept the new pinch end distance closer to the pinch start than before as a compromise.

Main goal is to get the pinching behaviour in a reliable and working state again. Other solutions can be explored either in community components or incrementally refined in the core.

**Changes proposed:**
- Replace dynamic pinch end distance threshold criteria with fixed threshold


dmarcos:
I think this needs some adjustment. Tried with this on Meta Quest 3

https://glitch.com/edit/#!/a-pickle-rick?path=index.html%3A10%3A44

Three script tags. One with this PR, master and 1.6.0. Master and 1.6.0 work as expected

mrxz:
> I can pinch and grab the model. With this PR the model spins without control.

That's unrelated to this PR. The model spinning already happened on master, but just got fixed in #5565. This PR didn't include that fix yet, but I've now rebased it. The reproduction in the linked issue (#5505) can also be used to test the change.

dmarcos:
Thanks!

PR #5611, opened 2024-11-18
Avoid attempting to load cubemap when envMap is not set for probe light (fix #5610)

### LINKED ISSUE #5610: error switching to AR mode on Galaxy Tab A9+
**Description:**
console shows this error on all scenes tested (including https://aframe.io/aframe/examples/boilerplate/ar-hello-world/) when switching to AR mode:

src-loader.js:23 Uncaught TypeError: n is not a function
    at src-loader.js:23:5
    at XMLHttpRequest.<anonymous> (src-loader.js:146:11)

I don't have another suitable android device to test with, and it may not impact functionality, but it is scary...

- A-Frame Version: 1.6.0
- Platform / Device: Galaxy Tab A9+
- Reproducible Code Snippet or URL: https://aframe.io/aframe/examples/boilerplate/ar-hello-world

mrxz:
**Description:**
In case no `envMap` was set for a probe light an attempt was still made to load it as a cube map. This resulted in a failed request and an error message in the console (see #5610)

This PR updates the `updateProbeLight` function to avoid this as well as clean up the caching logic. The cache would contain promises while the cube map was loading, but these never called the `resolve` method. This meant that if two probe lights would load the same `envMap` one of them would endlessly wait on a never-resolving Promise.

Fixes #5610 

**Changes proposed:**
- Don't continue loading cube map when `envMap` is not set
- Make the cube map loading promise resolve the resulting `LightProbe` instance

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -345,25 +345,30 @@ module.exports.Component = registerComponent('light', {
     if (!data.envMap) {
       // reset parameters if no map
       light.copy(new THREE.LightProbe());
+      return;
     }
 
+    // Populate the cache if not done for this envMap yet
+    if (probeCache[data.envMap] === undefined) {
+      probeCache[data.envMap] = new window.Promise(function (resolve) {
+        utils.srcLoader.validateCubemapSrc(data.envMap, function loadEnvMap (urls) {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you very much!

PR #5614, opened 2024-11-25
Make cursor component work when camera is not an entity like in the inspector

vincentfretin:
**Description:**

@mrxz #5606 broke the inspector that uses a threejs camera directly, not an entity.

**Changes proposed:**
- Check if camera.el is defined and only transform local to world coordinates if it is defined

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -263,7 +263,7 @@ module.exports.Component = registerComponent('cursor', {
     return function (evt) {
       var bounds = this.canvasBounds;
       var camera = this.el.sceneEl.camera;
-      var cameraElParent = camera.el ? camera.el.object3D.parent : null;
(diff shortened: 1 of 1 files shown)


mrxz:
Ah, I didn't account for `scene.camera` being set directly and not through the camera system, but for the inspector it does make sense.

Instead of the additional check, it might be better to move the assignment of `cameraElParent` into the `rayOrigin === 'xrselect'` clause. While the inspector is open there shouldn't be any `xrselect` events. And the xrselect raycasting only makes sense when the XRPose is translated into world space, so I'd rather see it bail/error out than to fallback to treating reference space as world space, which might not be correct.

vincentfretin:
Got it, I did the change you suggested.

dmarcos:
Thanks!

PR #5622, opened 2025-01-08
Cursor component: call addWebXREventListeners also when rayOrigin is entity

vincentfretin:
**Description:**

When I tried to update https://aframe-xr-starterkit.glitch.me to use master, the buttons on the wrist were not clickable anymore.
```html
  <script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@4d97225a04cfdc9ab3e44f84687f70d6816a366b/dist/aframe-master.min.js"></script><!-- Dec 31, 2024 not working -->
  <!--<script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@86d847fa80dcaa1c7306a93890000d5bb3183416/dist/aframe-master.min.js"></script>--><!-- Nov 15, 2024 not working, include https://github.com/aframevr/aframe/commit/655ab162d43f3d67327d4da69c3625aef324abb0 -->
  <!--<script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@8e52e89e813b440ee52e0a78aaddde33f98657e0/dist/aframe-master.min.js"></script>--><!-- Nov 15, 2024 working, don't include https://github.com/aframevr/aframe/commit/655ab162d43f3d67327d4da69c3625aef324abb0 -->
```

The culpurit was commit https://github.com/aframevr/aframe/commit/655ab162d43f3d67327d4da69c3625aef324abb0 that removed the webxr listeners for cursor with rayOrigin: entity (the default).

**Changes proposed:**
- Modify the check to use rayOrigin !== mouse, that what it was really before the commit

PR #5623, opened 2025-01-08
Cursor component: remove addWebXREventListeners/removeWebXREventListeners in update

vincentfretin:
**Description:**

Those calls were introduced in https://github.com/aframevr/aframe/commit/655ab162d43f3d67327d4da69c3625aef324abb0
If you set `cursor="rayOrigin: xrselect"`, it calls `addWebXREventListeners` in `update` and then in `onEnterVR`, so adding the listeners twice.
Later if you update a property that is not rayOrigin like fuse, it will call `addWebXREventListeners` and right away `removeWebXREventListeners`, so registering and unregistering listeners, it doesn't make sense.

**Changes proposed:**
- Remove `addWebXREventListeners`/`removeWebXREventListeners` from update, only keep `addWebXREventListeners` in `onEnterVR`, and the `removeWebXREventListeners` in `removeEventListeners` that is called in `pause`/`remove`.

I tested that on Android examples/showcase/ui/index.html with `<a-scene xr-mode-ui="XRMode: xr"` and `cursor="rayOrigin: xrselect; fuse: false"`. When I click on AR button, touching the image opens the modal correctly.
Am I missing something @mrxz?

### DIFF
--- a/src/components/cursor.js
+++ b/src/components/cursor.js
@@ -88,13 +88,6 @@ module.exports.Component = registerComponent('cursor', {
     if (this.data.rayOrigin === oldData.rayOrigin) { return; }
     if (this.data.rayOrigin === 'entity') { this.resetRaycaster(); }
     this.updateMouseEventListeners();
-    // Update the WebXR event listeners if needed
-    if (this.data.rayOrigin === 'xrselect') {
-      this.addWebXREventListeners();
-    }
-    if (oldData.rayOrigin === 'xrselect') {
-      this.removeWebXREventListeners();
-    }
   },
 
   tick: function () {


mrxz:
The [first guard](https://github.com/aframevr/aframe/pull/5623/files#diff-0e82e2c917e7127027721c5ab28da273907fdf1ddebc653d68cda0a9f97b17a7L88) should already prevent adding and removing the event listeners in the same `update`. Effectively there should only be two possible scenarios:

- No active XRsession when cursor is created, `update` does _not_ add WebXR event listeners, upon `onEnterVR` they are added
- Entering VR session before creating cursor so not added in `onEnterVR`, upon creation they are added in `update`

The general idea behind the logic is to handle the cases a cursor is created or has its `rayOrigin` changed _during_ an XR session. Instead of removing the logic, it should consider both `xrselect` and `entity` instead of just `xrselect`, similar to your fix in https://github.com/aframevr/aframe/pull/5622.

vincentfretin:
Okay, it makes sense now. Thanks for the explanation. I'm closing it and I'll change the condition as part of #5622

PR #5624, opened 2025-01-09
Be more specific about XRMode possible values for xr-mode-ui

vincentfretin:
I didn't see right away that you could also just set "ar" to have ar only, I had to look at the component code.


dmarcos:
Thanks.

PR #5625, opened 2025-01-10
Cursor component: remove raycaster-closest-entity-changed listener when removing the component

vincentfretin:
**Description:**

Remove the `raycaster-closest-entity-changed` listener (event introduced in https://github.com/aframevr/aframe/commit/981cb8105cf0d59a28b3a7ec3943cfad45dc2964) when removing the cursor component.
This fixes `getIntersection on undefined` error when we dynamically remove cursor and raycaster components on the scene when entering VR and there is a hand raycaster emitting this event.

**Changes proposed:**
- The `raycaster-closest-entity-changed` listener is added in `addEventListeners` but never removed, remove it in `removeEventListeners` like the other listeners.


dmarcos:
Thank you!

PR #5627, opened 2025-01-13
Update preghpages/ghpages commands to use recent versions

vincentfretin:
**Description:**
Update `preghpages`/`ghpages` commands to use recent versions without vulnerabilities.
Before:
```
39 vulnerabilities (14 moderate, 21 high, 4 critical)
```
After:
```
24 vulnerabilities (9 moderate, 13 high, 2 critical)
```

I tested `npm run preghpages` before and after, there is no change to the gh-pages folder produced.

Similar change also done in aframe-inspector repo: https://github.com/aframevr/aframe-inspector/pull/785

**Changes proposed:**
- Remove shx dependency that should have been removed when we moved from browserify to webpack
- Replace `ghpages -p folder` by `gh-pages -d folder` that does exactly the same thing. ghpages is a non maintained wrapper over gh-pages
- Update replace-in-file package used in the `scripts/preghpages.js` script


dmarcos:
Thanks!

PR #5628, opened 2025-01-13
Remove --experimental-require-module flag for the test:node command

vincentfretin:
Remove `--experimental-require-module` flag for the `test:node` command, not needed on latest node 22.x.


dmarcos:
Thanks!

PR #5632, opened 2025-01-17
Use import/export instead of require()/module.exports

mrxz:
**Description:**
As discussed in #5363 replacing CommonJS (`require()`/`module.exports`) with ES Modules (`import`/`export`) is the first step. This PR is a rebase of an earlier experiment I did, which replaces _most_ usages. 

A couple of notes:
- No build-step would still require additional work, as some dependencies aren't ESM
- ESLint requires the `ecmaVersion` to be 6 or higher, but this also allows all of the ES6 (2015) syntax and constructs. There doesn't seem to be a way to only allow import/export. Though it might be time to allow ES6 in the code base(?)
- Some import statements could be refined, e.g. `import * as X from 'x.js';` followed by `var y = X.y` can be simplified to `import { y } from 'x.js';`. This is done in some places, but not all. Generally the `utils` is being imported whole in a lot of places, where importing specific utilities would suffice. Generally preferable to only import what is needed to aid three-shaking (down the line).

I'll place some comments on this PR to highlight some significant changes. 

**Changes proposed:**
- Use `import`/`export` instead of `require()`/`module.exports`


.eslintrc.json
```diff
@@ -4,6 +4,9 @@
     "ecmaVersion": 12
   },
   "rules": {
+    /* To work towards no-build. */
+    "import/extensions": ["warn", "always", { "js": "always" }],
```
mrxz, line 8: Webpack doesn't need the `.js` suffix, but if the goal is to work towards a no-build setup it will be required. Thought it would be a good idea to start enforcing it regardless.

src/components/index.js
```diff
@@ -1,57 +1,57 @@
...
+import './line.js';
+import './link.js';
+import './logitech-mx-ink-controls.js';
+import './look-controls.js';
+import './magicleap-controls.js';
+import './material.js';
+import './meta-touch-controls.js';
+import './obb-collider.js';
+import './obj-model.js';
+import './oculus-go-controls.js';
```
mrxz, line 27: Diff looks a tad strange, as I moved `meta-touch-controls.js` up to maintain alphabetical order.

src/components/scene/inspector.js
```diff
@@ -1,8 +1,8 @@
 /* global AFRAME, INSPECTOR_VERSION */
-var AFRAME_INJECTED = require('../../constants').AFRAME_INJECTED;
-var pkg = require('../../../package');
-var registerComponent = require('../../core/component').registerComponent;
-var utils = require('../../utils/');
+import { AFRAME_INJECTED } from '../../constants/index.js';
+import { registerComponent } from '../../core/component.js';
+import pkg from '../../../package.json';
```
mrxz, line 4: Worth pointing out that importing JSON files would also not work in a no-build setup. Perhaps once JSON modules are readily available in browsers.

src/core/readyState.js
```diff
@@ -20,16 +20,19 @@ function waitForDocumentReadyState () {
...
   setTimeout(function () {
     document.dispatchEvent(new CustomEvent('aframeready'));
   });
 }
-module.exports.emitReady = emitReady;
+
+/** Resets the canInitializeElements flag, used for testing */
+export function reset () {
+  canInitializeElements = false;
+}
```
mrxz, line 38: Anything imported through `import` is not mutable. The `readyState.test.js` used to override this value to simulate various scenarios. As a workaround this method is introduced.

An alternative would be to create a holder and mutate its contents, e.g.:
```js
export var canInitializeElements = {
  value: false
}
```

But I'd rather add this method to accommodate tests than to change all consumers because of the tests.

src/core/scene/a-scene.js
```diff
@@ -1,24 +1,26 @@
...
+
 // Require after.
-var AEntity = require('../a-entity').AEntity;
-var ANode = require('../a-node').ANode;
-var initPostMessageAPI = require('./postMessage');
+import { AEntity } from '../a-entity.js';
+import { ANode } from '../a-node.js';
+import { initPostMessageAPI } from './postMessage.js';
+
+import '../../utils/ios-orientationchange-blank-bug.js';
```
mrxz, line 16: Conditional logic is moved into this class. So while it's always "loaded", it's only applied for iOS. This is simply due to the fact that `require()` is synchronous, but `import()` isn't. Which would then either require a top-level await, or an awkwardly placed `.then()`

src/index.js
```diff
@@ -21,41 +54,11 @@ if (!window.cordova && window.location.protocol === 'file:') {
 
 // CSS.
 if (utils.device.isBrowserEnvironment) {
+  window.logs = debug;
```
mrxz, line 57: This used to be part of `debug.js`, but would be executed only when `isBrowserEnvironment` which is defined in `device.js`, which in turn depends on `debug.js`. This circular dependency caused issues.

src/index.js
```diff
@@ -21,41 +54,11 @@ if (!window.cordova && window.location.protocol === 'file:') {
...
   require('./style/aframe.css');
   require('./style/rStats.css');
```
mrxz, line 59: These are the only real `require` statements left in the `src/` folder. Not sure what the best way to get rid of them is, but that's probably best suited for a follow-up task.

src/lib/three.js
```diff
@@ -1,11 +1,27 @@
-var THREE = require('./three.mjs').default;
```
mrxz, line 1: Since we're now using modules, there is no need for this file to act as an intermediate for `three.mjs`.

tests/__init.test.js
```diff
@@ -5,60 +5,67 @@
...
-  xrSession.supportedFrameRates = [90];
-  xrSession.requestReferenceSpace = function () { return Promise.resolve(); };
-  return Promise.resolve(xrSession);
-};
-
 /* WebXR Stub */
-navigator.xr = navigator.xr || {};
+if (!navigator.xr) {
+  navigator.xr = {};
+}
```
mrxz, line 11: Because the tests are now using ESM, this implicitly enabled strict-mode as well, which prevents a lot of overwriting of getter-only or read-only properties. In this case the `xr` property on `navigator` is getter-only when defined.

tests/__init.test.js
```diff
@@ -5,60 +5,67 @@
...
+  import('index.js'),
+  import('core/scene/a-scene.js')
+]).then(function (results) {
+  AFRAME = results[0].value.default;
+  AScene = results[1].value.AScene;
+
+  // Make sure WebGL context is not created since CI runs headless.
+  // Stubs below failed once in a while due to asynchronous test setup / teardown.
+  AScene.prototype.setupRenderer = function () {};
+});
```
mrxz, line 32: Asynchronous loading of AFRAME and a-scene, as otherwise the above WebXR stubbing would happen too late.

tests/components/scene/stats.test.js
```diff
@@ -11,14 +9,14 @@ suite('stats', function () {
...
-    utils.getUrlParameter = function () { return 'false'; };
+    window.history.replaceState(null, '', originalLocationUrl + '?stats=false');
   });
 
   teardown(function () {
     var el = this.sceneEl;
     el.parentNode.removeChild(el);
 
-    utils.getUrlParameter = originalGetUrlParameter;
+    window.history.replaceState(null, '', originalLocationUrl);
```
mrxz, line 19: Instead of trying to overwrite the `getUrlParameter` util (which is now read-only) the URL is actually rewritten using the history API.

tests/components/scene/xr-mode-ui.test.js
```diff
@@ -1,12 +1,13 @@
 /* global assert, process, setup, suite, test */
-var entityFactory = require('../../helpers').entityFactory;
+import { entityFactory } from '../../helpers.js';
 
-var UI_CLASSES = ['.a-orientation-modal', '.a-enter-vr-button'];
+var UI_CLASSES = ['.a-orientation-modal', '.a-enter-vr'];
```
mrxz, line 4: The `a-hidden` class is applied to the wrapper around the button. The test was incorrect before.

tests/core/controls.test.js
```diff
@@ -407,7 +407,7 @@ suite('rotation controls on camera with touch drag (integration unit test)', fun
 
     canvas = sceneEl.canvas;
     sceneEl.isMobile = true;
-    canvas.clientWidth = 1000;
+    assert.isAtLeast(canvas.clientWidth, 1000);
```
mrxz, line 410: Again an instance where one can't set values that are read-only. From my testing the line of code wasn't actually needed, but I've added an assertion to at least make sure the desired state is there.

webpack.config.js
```diff
@@ -4,7 +4,11 @@ var webpack = require('webpack');
 module.exports = {
   entry: './src/index.js',
   output: {
-    library: 'AFRAME',
+    library: {
+      name: 'AFRAME',
+      type: 'var',
+      export: 'default'
+    },
```
mrxz, line 11: This instructs WebPack to basically ensure that when loading the UMD bundle, the default export will be available under the name AFRAME. This matches the current behaviour of the bundle AFAICT.

src/core/system.js
```diff
@@ -69,7 +65,7 @@ System.prototype = {
    */
   updateProperties: function (rawData) {
     var oldData = this.data;
-    if (!Object.keys(schema).length) { return; }
+    if (!Object.keys(this.schema).length) { return; }
```
vincentfretin, line 68: Huh, you found an issue it seems, so before that was always false?
And now it's returning true for single property system if I understand it? That's okay?

src/lib/three.js
```diff
@@ -1,11 +1,27 @@
...
+import * as SUPER_THREE from 'three';
+import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
+import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
+import { KTX2Loader } from 'three/examples/jsm/loaders/KTX2Loader.js';
+import { OBB } from 'three/addons/math/OBB.js';
+import { OBJLoader } from 'three/examples/jsm/loaders/OBJLoader.js';
+import { MTLLoader } from 'three/examples/jsm/loaders/MTLLoader.js';
+import * as BufferGeometryUtils from 'three/examples/jsm/utils/BufferGeometryUtils.js';
+import { LightProbeGenerator } from 'three/examples/jsm/lights/LightProbeGenerator.js';
```
vincentfretin, line 9: We can replace all "three/examples/jsm/" by "three/addons/" like for the OBB import for consistency.

src/utils/src-loader.js
```diff
@@ -184,7 +184,7 @@ function validateAndGetQuerySelector (selector) {
   }
 }
 
-module.exports = {
+export default {
```
vincentfretin, line 187: We could also export each function individually.

tests/utils/isIOSOlderThan10.test.js
```diff
@@ -1,5 +1,5 @@
 /* global assert, suite, test */
-var isIOSOlderThan10 = require('utils/isIOSOlderThan10');
+var isIOSOlderThan10 = require('utils/isIOSOlderThan10.js');
```
vincentfretin, line 2: You missed this one? `src/utils/isIOSOlderThan10.js` is not converted either.

src/components/scene/stats.js
```diff
@@ -1,8 +1,8 @@
-var registerComponent = require('../../core/component').registerComponent;
-var RStats = require('../../../vendor/rStats');
-var utils = require('../../utils');
-require('../../../vendor/rStats.extras');
-require('../../lib/rStatsAframe');
+import { registerComponent } from '../../core/component.js';
+import RStats from '../../../vendor/rStats.js';
+import * as utils from '../../utils/index.js';
+import '../../../vendor/rStats.extras.js';
+import '../../lib/rStatsAframe.js';
```
vincentfretin, line 5: I think we can just remove the condition with module.exports in `src/lib/rStatsAframe.js`

src/utils/index.js
```diff
@@ -1,46 +1,47 @@
...
   warn('`utils.isIOS` has moved to `utils.device.isIOS`');
   return device.isIOS(arguments);
-};
-module.exports.isOculusGo = module.exports.device.isOculusGo = function () {
+}
+export function isOculusGo () {
   warn('`utils.isOculusGo` has been deprecated, use `utils.device.isMobileVR`');
-};
-module.exports.isMobile = function () {
+}
```
vincentfretin, line 40: So before isGearVR and isOculusGo were available on AFRAME.utils and AFRAME.utils.device I see.
Those functions are deprecated since a long time now.
Shouldn't we just remove those two functions now @dmarcos? We also removed gearvr-controls from the repo.

vincentfretin (commented):
Wow that's an insane amount of work you did here, thank you!

src/components/scene/inspector.js
vincentfretin: The new web standard will be:
```js
import pkg from '../../../package.json' with { type: "json" };
```
currently supported by node if you test it with `npm run test:node`.
For webpack, babel-generator gives a warning 
```
You are using import attributes, without specifying the desired output syntax.
Please specify the "importAttributesKeyword" generator option, whose value can be one of:
 - "with"        : `import { a } from "b" with { type: "json" };`
 - "assert"      : `import { a } from "b" assert { type: "json" };`
 - "with-legacy" : `import { a } from "b" with type: "json";`
```
Not sure where you configure it.

src/lib/three.js
```diff
@@ -1,11 +1,27 @@
...
+import { DeviceOrientationControls } from '../../vendor/DeviceOrientationControls.js'; // THREE.DeviceOrientationControls
+
+var THREE = global.THREE = SUPER_THREE;
+
 // TODO: Eventually include these only if they are needed by a component.
-global.THREE = THREE;
-require('../../vendor/DeviceOrientationControls');
+THREE.DRACOLoader = DRACOLoader;
```
vincentfretin, line 15: Before we had
```js
var THREE = { ...SUPER_THREE };
```
that was needed for node that you can test with `npm run test:node`
you get `TypeError: Cannot add property DRACOLoader, object is not extensible`
without it.

src/lib/three.js
```diff
@@ -1,11 +1,27 @@
...
+import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
+import { KTX2Loader } from 'three/examples/jsm/loaders/KTX2Loader.js';
+import { OBB } from 'three/addons/math/OBB.js';
+import { OBJLoader } from 'three/examples/jsm/loaders/OBJLoader.js';
+import { MTLLoader } from 'three/examples/jsm/loaders/MTLLoader.js';
+import * as BufferGeometryUtils from 'three/examples/jsm/utils/BufferGeometryUtils.js';
+import { LightProbeGenerator } from 'three/examples/jsm/lights/LightProbeGenerator.js';
+import { DeviceOrientationControls } from '../../vendor/DeviceOrientationControls.js'; // THREE.DeviceOrientationControls
+
+var THREE = global.THREE = SUPER_THREE;
```
vincentfretin, line 12: Also should be `window.THREE` in a no build context. `global` works only via webpack or node.

vincentfretin:
Using importmap works!

`examples/boilerplate/hello-world/importmap.html`
```html
<!DOCTYPE html>                                                                    
<html>                                                                             
  <head>                                                                           
    <meta charset="utf-8">                                                         
    <title>Hello, World! • A-Frame</title>                                         
    <meta name="description" content="Hello, World! • A-Frame">                    
    <script type="importmap">                                                      
      {                                                                            
        "imports": {                                                               
          "aframe": "../../../dist/aframe-master.module.min.js",                   
          "three": "https://cdn.jsdelivr.net/npm/super-three@0.172.0/build/three.module.js",
          "three/addons/": "https://cdn.jsdelivr.net/npm/super-three@0.172.0/examples/jsm/"
        }                                                                          
      }                                                                            
    </script>                                                                      
    <script type="module">                                                         
      import AFRAME from "aframe";                                                 
//      import * as THREE from "three";                                            
      import { ParametricGeometry } from "three/addons/geometries/ParametricGeometry.js";
    </script>                                                                      
  </head>                                                                          
  <body>                                                                           
    <a-scene background="color: #ECECEC">                                          
      <a-box position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9" shadow></a-box>
      <a-sphere position="0 1.25 -5" radius="1.25" color="#EF2D5E" shadow></a-sphere>
      <a-cylinder position="1 0.75 -3" radius="0.5" height="1.5" color="#FFC65D" shadow></a-cylinder>
      <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4" shadow></a-plane>
    </a-scene>                                                                     
  </body>                                                                          
</html> 
```

```patch
diff --git a/src/index.js b/src/index.js
index 07b97618..6ba0c75b 100644
--- a/src/index.js
+++ b/src/index.js
@@ -68,7 +68,7 @@ if (!window.AFRAME_ASYNC) {
   readyState.waitForDocumentReadyState();
 }
 
-export default {
+var AFRAME = {
   AComponent: Component,
   AEntity: AEntity,
   ANode: ANode,
@@ -95,3 +95,5 @@ export default {
   utils: utils,
   version: pkg.version
 };
+window.AFRAME = AFRAME;
+export default AFRAME;
diff --git a/src/lib/three.js b/src/lib/three.js
index a8e34a06..7fb4d65b 100644
--- a/src/lib/three.js
+++ b/src/lib/three.js
@@ -9,7 +9,7 @@ import * as BufferGeometryUtils from 'three/examples/jsm/utils/BufferGeometryUti
 import { LightProbeGenerator } from 'three/examples/jsm/lights/LightProbeGenerator.js';
 import { DeviceOrientationControls } from '../../vendor/DeviceOrientationControls.js'; // THREE.DeviceOrientationControls
 
-var THREE = global.THREE = SUPER_THREE;
+var THREE = window.THREE = {...SUPER_THREE};
 
 // TODO: Eventually include these only if they are needed by a component.
 THREE.DRACOLoader = DRACOLoader;
```

`webpack.prod-module.config.js` (remove libraryName, set libraryTarget 'module' with experiments.outputModule: true and externals three):
```js
var path = require('path');
var webpack = require('webpack');
var TerserPlugin = require('terser-webpack-plugin');

module.exports = {
  entry: './src/index.js',
  output: {
    libraryTarget: 'module',
    path: path.resolve(__dirname, 'dist'),
    publicPath: '/dist/',
    filename: 'aframe-master.module.min.js'
  },
  experiments: {
    outputModule: true
  },
  externalsType: 'module',
  externals: {
    three: 'three'
  },
  devtool: 'source-map',
  mode: 'production',
  devServer: {
    port: process.env.PORT || 9000,
    hot: false,
    liveReload: true,
    static: {
      directory: 'examples'
    }
  },
  plugins: [
    new webpack.DefinePlugin({
      INSPECTOR_VERSION: JSON.stringify(
        process.env.INSPECTOR_VERSION
      )
    }),
    new webpack.ProvidePlugin({
      Buffer: ['buffer', 'Buffer']
    })
  ],
  module: {
    rules: [
      {
        test: /\.js$/,
        use: {
          loader: 'babel-loader'
        }
      },
      {
        test: /\.css$/,
        use: ['style-loader', 'css-loader']
      }
    ]
  },
  optimization: {
    minimize: true,
    minimizer: [
      new TerserPlugin({
        terserOptions: {
          compress: {
            passes: 2
          },
          format: {
            comments: false
          }
        },
        extractComments: false
      })
    ]
  }
};
```

```sh
npx webpack --config webpack.prod-module.config.js
npm start
```

http://localhost:8080/boilerplate/hello-world/importmap.html

src/lib/three.js
vincentfretin: We also have the error with webpack `output.libraryTarget: 'module'` so we need it back.
Also we should use window instead of global for a future no build context. global works only via webpack or node.

```js
var THREE = window.THREE = {...SUPER_THREE};
```

webpack.config.js
vincentfretin: Yes but we should still set `window.AFRAME = AFRAME;` ourself for the module build used with importmap.

vincentfretin:
For the `npm run test:node` I think we should remove that node support. I fixed it last time because I could fix it, but that is really used nowhere, nobody complained that the test was broken when we switched to three ES modules at that time, that's just useful to import aframe in node and get the version (from the doc) :D

src/constants/index.js
```diff
@@ -1,7 +1,7 @@
...
-  DEFAULT_HANDEDNESS: 'right',
-  keyboardevent: require('./keyboardevent')
-};
+import { KEYCODE_TO_CODE } from './keyboardevent.js';
+
+export var AFRAME_CDN_ROOT = window.AFRAME_CDN_ROOT || 'https://cdn.aframe.io/';
+export var AFRAME_INJECTED = 'aframe-injected';
+export var DEFAULT_CAMERA_HEIGHT = 1.6;
+export var DEFAULT_HANDEDNESS = 'right';
+export var keyboardevent = KEYCODE_TO_CODE;
```
vincentfretin, line 7: Shouldn't it be 
```js
export var keyboardevent = {KEYCODE_TO_CODE}
```
here or
```js
import * as keyboardevent from './keyboardevent.js';
export var keyboardevent = keyboardevent;
```
?
We don't seem to import `keyboardevent` from `constants/index.js` anywhere, we always import it from `constants/keyboardevent.js`

We could also just remove the line.

src/components/scene/inspector.js
mrxz: For me webpack (or rather babel) complains that the `@babel/plugin-syntax-import-attributes` is missing, and after installing and adding it to the plugins of `babel-loader`, it works fine.

But I don't think there's any upside to pre-emptively using this syntax. In fact, I'd rather see us avoiding importing `package.json` altogether. Currently the entirety of `package.json` gets embedded into the A-Frame bundle, despite only needing the version property.

src/components/meta-touch-controls.js
```diff
@@ -474,10 +470,7 @@ var componentConfig = {
       button.material.color.set(color);
     }
   }
-};
-
-registerComponent('oculus-touch-controls', componentConfig);
-module.exports.Component = registerComponent('meta-touch-controls', componentConfig);
+});
```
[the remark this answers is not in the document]
mrxz: Overlooked this while resolving merge conflicts during the rebase as my local branch was from before the rename. Will restore it so that the component is registered under both names.

Though, thinking about it, I'm not convinced this is a good way to achieve backwards compatibility. Something trivial like the following will already start behaving differently:
```HTML
<a-entity id="leftHand" laser-controls="hand: left" oculus-touch-controls="buttonHighlightColor: red" raycaster="objects: [mixin='box']"></a-entity>
```
The `laser-controls` will add the new `meta-touch-controls` while the `oculus-touch-controls` attribute will cause a second instance to be added. Pulling the trigger will result in two events being emitted...

src/core/system.js
mrxz: Before it was actually always true, since `schema` was imported and in scope, see [system.js#L2](https://github.com/aframevr/aframe/blob/master/src/core/system.js#L2).

Now it's only true if the schema has no keys, meaning there should be nothing to do anyway. I believe this was the original intention behind the guard. For any system that is single-property or multi-property nothing changes, only those that don't have any properties, and thus don't expect any data, will now skip the data parsing.

tests/utils/isIOSOlderThan10.test.js
mrxz: Actually it can be removed. It's only referenced in this test and never exposed in the bundle. This check seems to originally be used to workaround a bug in the webvr-polyfill on Safari.

src/constants/index.js
mrxz: Ah, you're right, but I agree with just removing this line. We indeed only ever import directly from `constants/keyboardevent.js` it seems.

src/components/scene/inspector.js
vincentfretin: I agree with you, I wouldn't introduce the new syntax either. We should just inject the aframe version as part of the build like we do for INSPECTOR_VERSION really. Importing package.json was only to get the aframe version when importing from node AFAICT. I don't see a real use of importing aframe in node where there is no custom elements api.

src/components/meta-touch-controls.js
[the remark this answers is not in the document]
vincentfretin: True. I think we should just remove it.

src/core/system.js
vincentfretin: Right, I completely misread the condition. So okay all good. I find it hard to process in my mind what `!Object.keys(this.schema).length` means. `Object.keys(this.schema).length === 0` would be clearer to me.

tests/utils/isIOSOlderThan10.test.js
vincentfretin: Talking about webvr-polyfill we removed. I think we should also remove the cardboardModeEnabled flag in xr-mode-ui component and documentation.

src/components/meta-touch-controls.js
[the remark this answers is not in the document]
mrxz: @dmarcos What is your thoughts on this? IMHO we should either remove it, or implement proper backwards compatibility.

src/lib/three.js
mrxz: Now I recall having it that way before, but changing it while testing tree-shaking, as `{...SUPER_THREE}` prevents most tree-shaking. But indeed, it should be `{...SUPER_THREE}` for the time being.

Ultimately I think it would be better to have the A-Frame code import from `three` and its addons directly where applicable. Then construct this "SUPER_THREE + some addons" only for the final exports and populating the global scope. But that is beyond the scope of this PR.

src/components/scene/inspector.js
mrxz: It was used for logging the THREE and WebVR polyfill versions in the past as well, but the former now uses `THREE.REVISION` and the latter has since been removed. Injecting the version is a lot cleaner.

vincentfretin:
> For the `npm run test:node` I think we should remove that node support. I fixed it last time because I could fix it, but that is really used nowhere, nobody complained that the test was broken when we switched to three ES modules at that time, that's just useful to import aframe in node and get the version (from the doc) :D

I tried yesterday to fix it, that meant using the `with { type: "json" }` for importing package.json (and we need an additional babel plugin to support that syntax for the bundles), using `type: "module"` in `package.json`, renaming `webpack.config.js` to `webpack.config.cjs`, but in the end we are still blocked with the two conditional require for the css. In node, an ES module can't use require.

src/lib/three.js
vincentfretin: Yes I totally agree. For example importing `LightProbeGenerator` should really be done in `src/components/light.js`. We can do that in a separate PR.

vincentfretin:
Just reviewed again after you force pushed, your changes are all okay. :+1: The webpack module build still working properly.
Using globalThis instead of window or global is indeed correct, I completely forgot that one existed.

src/components/meta-touch-controls.js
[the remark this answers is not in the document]
dmarcos: I would maintain backwards compatibility.  `laser-controls` and `oculus-touch-controls` are not meant to be used together. We can perhaps add a clause somewhere to make `oculus-touch-controls` a no-op if `meta-touch-controls` defined and viceversa. That way they can't be added together.

```
<a-entity id="leftHand" laser-controls="hand: left" oculus-touch-controls="buttonHighlightColor: red"></a-entity>
```

dmarcos:
Thanks for this. Regular script tag still works right?

dmarcos:
It needs rebase as well.

src/components/meta-touch-controls.js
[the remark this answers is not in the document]
mrxz: If they aren't meant to be used together, we shouldn't use them together in the [docs here](https://aframe.io/docs/1.6.0/introduction/entity-component-system.html#composition) :sweat_smile: 

Either way, a quick search on GitHub reveals that they _are_ being used together quite a lot. Often times it's just superfluous, but even those usages will "break" when upgrading to 1.7.0. Not to mention users directly referencing `el.components['oculus-touch-controls']` or querying on `[oculus-touch-controls]`.

Since this isn't really in the scope of this PR, I'll restore the double registration for now.

mrxz:
> Thanks for this. Regular script tag still works right?

Yes, that's the point of this PR. Switching from CommonJS to ESM internally while still outputting an UMD bundle that works like before. 

With this done, it becomes possible to create an ESM bundle and using it in an importmap, as @vincentfretin demonstrates, but that isn't part of this PR.

vincentfretin:
Yes, that is good to be merged for me.
I'll create a PR that create the new ESM bundle with documentation and an example after this is merged.

src/utils/index.js
dmarcos: yeah. we can remove those functions

dmarcos:
Very nice work. @vincentfretin are you working on follow up examples / docs?

src/utils/index.js
vincentfretin: done in #5635

tests/core/controls.test.js
vincentfretin: On my Ubuntu machine, the window opens with width 886
`AssertionError: expected 886 to be at least 1000`
I changed it to 800 in #5641

dmarcos:
@vincentfretin would be nice to have docs/examples with the new ways to import / consume aframe. Do you have anything handy we can incorporate to docs / readme / examples / maybe a glitch

vincentfretin:
Yes, I said in the comment just above https://github.com/aframevr/aframe/pull/5632#issuecomment-2604194934 I'll do that. ;)
I did a glitch, see my comment https://github.com/aframevr/aframe/issues/5363#issuecomment-2600960334 but I'll add it in the examples in the repo. I did today #5640 to create the module bundle. I'll create tomorrow or Saturday another PR with the example and documentation.

dmarcos:
> Yes, I said in the comment just above [#5632 (comment)](https://github.com/aframevr/aframe/pull/5632#issuecomment-2604194934) I'll do that. ;) I did a glitch, see my comment [#5363 (comment)](https://github.com/aframevr/aframe/issues/5363#issuecomment-2600960334) but I'll add it in the examples in the repo. I did today #5640 to create the module bundle. I'll create tomorrow or Saturday another PR with the example and documentation.

Thanks so much. Just making sure I understood and we’re in same page. Really excited about getting best of both modules and old school builds.

vincentfretin:
Example and doc in #5645

PR #5633, opened 2025-01-20
Only apply position and orientation of spatial anchor space to anchored entity

### LINKED ISSUE #5630: Using spatial anchor causes scale to get overruled
**Description:**

I have a simple `a-box` with `scale` set to something that is not the default. If I place `anchored` on it and call `createAnchor` as per the documentation, my scale gets set back to a uniform `1`.

It appears that `tick` in the `anchored` component does:

```js
pose = frame.getPose(this.anchor.anchorSpace, refSpace);
object3D.matrix.elements = pose.transform.matrix;
object3D.matrix.decompose(object3D.position, object3D.rotation, object3D.scale);
```

This seems to overrule the entire matrix of the object, including the scale. I verified this by placing `const oldScale = object3D.scale.clone();` before the code and `object3D.scale.copy(oldScale);` after to restore it.

- A-Frame Version: 1.6.0
- Platform / Device: Meta Quest 3
- Reproducible Code Snippet or URL: /

mrxz:
**Description:**
As reported in #5630 the `anchored` component would set the scale to 1, despite only needing to position and orientate the entity based on the spatial anchor's pose. This PR replaces the matrix decomposition with a straightforward copy of the position and orientation, leaving the scale of the entity as-is.

Fixes #5630 

**Changes proposed:**
- Only apply position and orientation of spatial anchor space to anchored entity

### DIFF
--- a/src/components/anchored.js
+++ b/src/components/anchored.js
@@ -52,8 +52,9 @@ module.exports.Component = registerComponent('anchored', {
     refSpace = xrManager.getReferenceSpace();
 
     pose = frame.getPose(this.anchor.anchorSpace, refSpace);
-    object3D.matrix.elements = pose.transform.matrix;
-    object3D.matrix.decompose(object3D.position, object3D.rotation, object3D.scale);
+    // Apply position and orientation, leave scale as-is (see aframevr/aframe#5630)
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

PR #5635, opened 2025-01-23
Remove deprecated AFRAME.utils.isGearVR and AFRAME.utils.isOculusGo functions

vincentfretin:
**Description:**

Remove deprecated AFRAME.utils.isGearVR and AFRAME.utils.isOculusGo functions


dmarcos:
Thanks!

PR #5636, opened 2025-01-23
Fix wrong description for AFRAME.utils.device.checkHeadsetConnected()

vincentfretin:
**Description:**

Fix wrong description for AFRAME.utils.device.checkHeadsetConnected() function, this has nothing to do with orientation data.


dmarcos:
Thanks!

PR #5638, opened 2025-01-23
Remove cardboardModeEnabled property in xr-mode-ui

vincentfretin:
Remove `cardboardModeEnabled` property in `xr-mode-ui` now that we removed webvr-polyfill and WebVRManager.


dmarcos:
Thanks!

PR #5639, opened 2025-01-23
Remove references to Cardboard in documentation

vincentfretin:
Remove references to Cardboard in documentation

### DIFF
--- a/docs/components/laser-controls.md
+++ b/docs/components/laser-controls.md
@@ -9,10 +9,9 @@ source_code: src/components/laser-controls.js
 examples: []
 ---
 
-The laser-controls component provides tracked controls with a laser or ray
-cursor shooting out to be used for input and interactions. It only requires
-a way to orientate the laser and an input the user can trigger.
-This makes it usable on a wide range of devices. Using laser-controls we can
+The laser-controls component provides a tracked controller with a laser or ray
+cursor shooting out to be used for input and interactions. All headsets ship with some form of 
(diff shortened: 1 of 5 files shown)


docs/components/laser-controls.md
```diff
@@ -14,8 +14,8 @@ examples: []
 The laser-controls component provides tracked controls with a laser or ray
 cursor shooting out to be used for input and interactions. *DoF* stands for
 [degrees of freedom][dof]. Because they only require rotation and some form of
-input, laser-based interactions scale well across 0 DoF (gaze-based,
-Cardboard), and 6 DoF (Vive, Oculus Touch). If desired, we can get a consistent form of interaction that works
+input, laser-based interactions scale well across 0 DoF (gaze-based),
```
mrxz, line 17: There might be some confusion why gaze-based is referred to as 0 DoF. With eye tracking it's common to use 'gaze' to mean a 3 DoF input, especially now Cardboard isn't mentioned any more.

This section could perhaps be reworded, I don't think the term DoF needs to be introduced at all. Perhaps something along the lines of:
> The laser-controls component provides tracked controls with a laser or ray
cursor shooting out to be used for input and interactions. It only requires a way to orientate the laser and an input the user can trigger. This makes it usable on a wide range of devices. Using laser-controls we can get a consistent form of interaction that works across all VR platforms with a single line of HTML.

vincentfretin: Okay, I used your paragraph.

dmarcos:
Thanks!

PR #5640, opened 2025-01-23
Add webpack config to create a module bundle to be used with importmap

vincentfretin:
**Description:**

Add webpack config to create a module bundle to be used with importmap

**Changes proposed:**
- Move shared config in webpack.common.js
- Create new webpack.module.config.js and `npm run dist:module` command

After merge, please commit the new bundle:
```sh
npm run dist
git add dist/aframe-master.module.min.js dist/aframe-master.module.min.js.map
git commit -m"Add new module bundle"
```


dmarcos:
Thanks!

PR #5642, opened 2025-01-25
Fix wrong assert in a getDOMAttribute test, it now returns undefined instead of empty object

vincentfretin:
**Description:**

That was discussed in https://github.com/aframevr/aframe-inspector/issues/741#issuecomment-2284890160
since the big refactoring and optimization of getDOMAttribute/getAttribute, that particular getDOMAttribute test wrongly passes because `shallowDeepEqual(undefined, {})` is true.

**Changes proposed:**
- Fix the test to use assert.equal and change the expected value.

### DIFF
--- a/tests/core/a-entity.test.js
+++ b/tests/core/a-entity.test.js
@@ -727,9 +727,9 @@ suite('a-entity', function () {
       assert.notOk('height' in componentData);
     });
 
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

PR #5645, opened 2025-01-27
Add Teapot example that uses the new ES module bundle with importmap

vincentfretin:
**Description:**

Add Teapot example that uses the new ES module bundle with importmap.

**Changes proposed:**

- Add Teapot example (use importmap and import from three/addons)
- Add the new ES module bundle and importmap example to the faq
- Update super-three version in examples and docs in prerelease script
- Create the aframe-v{nextVersion}.module.min.js / aframe.module.min.js bundle in prerelease script

### DIFF
--- a/examples/boilerplate/teapot/index.html
+++ b/examples/boilerplate/teapot/index.html
@@ -1,0 +1,83 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Teapot • A-Frame</title>
+    <meta name="description" content="Teapot • A-Frame" />
+    <script type="importmap">
+      {
+        "imports": {
+          "aframe": "../../../dist/aframe-master.module.min.js",
+          "three": "https://cdn.jsdelivr.net/npm/super-three@0.172.0/build/three.module.js",
+          "three/addons/": "https://cdn.jsdelivr.net/npm/super-three@0.172.0/examples/jsm/",
+          "aframe-extras/controls": "https://cdn.jsdelivr.net/gh/c-frame/aframe-extras@7.5.x/dist/aframe-extras.controls.min.js"
+        }
+      }
+    </script>
+    <script type="module">
+      // We use an importmap to load three as a module to avoid the warning "Multiple instances of Three.js being imported."
+      import AFRAME from "aframe";
+      // AFRAME and THREE variables are available globally, the imported aframe-master.module.min.js bundle basically does:
+      // import * as THREE from "three"
+      // window.THREE = THREE
+      import "aframe-extras/controls"; // This uses the THREE global variable in the bundle.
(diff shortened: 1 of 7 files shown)


examples/boilerplate/teapot/index.html
```diff
@@ -0,0 +1,83 @@
...
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Teapot • A-Frame</title>
+    <meta name="description" content="Teapot • A-Frame" />
+    <script type="importmap">
+      {
+        "imports": {
+          "aframe": "../../../dist/aframe-master.module.min.js",
+          "three": "https://cdn.jsdelivr.net/npm/super-three@0.172.0/build/three.module.js",
```
dmarcos, line 11: Problem is that will have to update this in the future since it’s locked to r172. Is it possible to have an example without a hardcoded version?

mrxz: It could be an option to tag the super-three versions with the A-Frame version they are used in, so you'd be able to use `super-three@aframe-1.7.0`. For master we could consider pointing it to `node_modules` instead of using a CDN? That way when checking out older revisions, running `npm install` ensures that the right super-three version is referenced.

dmarcos: yeah. pointing to node_modules sounds fine to me. trying to avoid small things that will eventually break and hard to keep track of over time.

examples/index.html
```diff
@@ -157,6 +157,7 @@ <h2>Examples</h2>
       <li><a href="boilerplate/3d-model/">3D Model (glTF)</a></li>
       <li><a href="mixed-reality/anchor/">Anchor (Mixed Reality)</a></li>
       <li><a href="mixed-reality/real-world-meshing/">Real World Meshing (Mixed Reality)</a></li>
+      <li><a href="boilerplate/teapot/">Teapot (use importmap and import from three/addons)</a></li>
```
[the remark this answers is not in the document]
dmarcos: yeah. boilerplate/importmap or boilerplate/modules. not sure

examples/boilerplate/teapot/index.html
vincentfretin: Did you see my changes to the release script? There is no need to update manually, the script will update the super-three version in examples and docs similar to the aframe version.
But indeed for master that should reference the current version in node_modules. I just added an additional directory path to the webpack dev server config, and the preghpages script rewrite the url with the cdn.

examples/index.html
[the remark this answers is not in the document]
vincentfretin: Okay, I renamed to boilerplate/importmap.

vincentfretin:
I did the changes. There is no manual changes needed to update url or version for super-three in development with webpack or when the docs are released, all is done via the preghpages and prerelease scripts.

dmarcos:
Thank you!

PR #5646, opened 2025-01-27
Update attribute mapping tables in the docs for mesh primitives

mrxz:
**Description:**
The attribute mapping tables in the docs for the various mesh primitives was outdated. To remedy this, I've updated the script used to generate these to output pre-formatted tables.

**Changes proposed:**
- Update `scripts/docs-primitives` to output pre-formatted tables
- Update attribute tables for mesh primitive


dmarcos:
Thanks!

PR #5650, opened 2025-02-01
Rename sortObjects to sortTransparentObjects in example, foveationLevel can also be set dynamically

vincentfretin:
Rename sortObjects to sortTransparentObjects in example, foveationLevel can also be set dynamically


dmarcos:
Thanks!

PR #5651, opened 2025-02-02
Enable eslint no-unused-vars and array-callback-return rules

vincentfretin:
Enable eslint no-unused-vars and array-callback-return rules and fix all the errors in the repo.


vincentfretin:
I reviewed the other disabled rules, those are mostly styling rules, but those two no-unused-vars and array-callback-return is catching real issues, so better enabling them.

dmarcos:
Thanks!

PR #5653, opened 2025-02-03
Remove https://dpdb.webvr.rocks from CSP

vincentfretin:
**Description:**

**Changes proposed:**
- Remove https://dpdb.webvr.rocks from the CSP example, that link was the devices db for webvr-polyfill


dmarcos:
Thank you

PR #5654, opened 2025-02-03
Expose ES module, remove browserify mention and use import syntax in documentation

vincentfretin:
**Description:**

Expose ES module, remove browserify mention and use import syntax in documentation.

In a webpack project, before you may have used:
```js
import "aframe/src/index.js";
```
and in webpack you had to add the Buffer polyfill:
```js
   plugins: [
    new webpack.ProvidePlugin({
      Buffer: ["buffer", "Buffer"],
    }),
```
I think only me and @arpu are doing this :D

Now that the ES module is exposed in the npm package, you can just use:
```js
import AFRAME from "aframe";
```
and remove the polyfill from your webpack configuration.

It will also be compatible with other modules you may include like bloom.js that also use `import AFRAME from "aframe";`

**Changes proposed:**
- Add module key to package.json to expose aframe-master.module.min.js, bundlers like webpack give priority to the module key instead of main key. The main key is used with really old bundlers like browserify.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,14 @@
   "version": "1.6.0",
   "description": "A web framework for building virtual reality experiences.",
   "homepage": "https://aframe.io/",
-  "main": "dist/aframe-master.js",
+  "main": "./dist/aframe-master.js",
+  "module": "./dist/aframe-master.module.min.js",
+  "exports": {
+    ".": {
+      "import": "./dist/aframe-master.module.min.js",
+      "require": "./dist/aframe-master.js"
+    }
+  },
   "scripts": {
     "dev": "cross-env INSPECTOR_VERSION=dev webpack serve --port 8080",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -3,7 +3,8 @@
   "version": "1.6.0",
   "description": "A web framework for building virtual reality experiences.",
   "homepage": "https://aframe.io/",
-  "main": "dist/aframe-master.js",
+  "main": "./dist/aframe-master.js",
+  "module": "./dist/aframe-master.module.min.js",
```
mrxz, line 7: The `"module"` field never really became standardized, instead we should use `"exports"` (https://webpack.js.org/guides/package-exports/). Luckily most tooling can handle things gracefully, so we can keep `module` in there as well for older tools that might not support `exports` yet.

vincentfretin: good point, I added exports

dmarcos:
Thanks!

PR #5658, opened 2025-02-09
Add 3D gaussian splats support

dmarcos:
Last big item for 1.7.0 release. Still work in progress. Based on https://github.com/quadjr/aframe-gaussian-splatting implementation

I adapted style, made a few minor tweaks and added commented thoroughly. I still don't understand some of the details. Good opportunity to deepen understanding of 3D gaussian splats for those interested. Before merge we should thoroughly understand the code and optimize.

I recommend getting acquainted with the original paper: https://arxiv.org/pdf/2308.04079

In the code comments you'll see references to other relevant literature.

In VR on a Quest 3 when the sample splat is fully loaded (~1.1M splats rendered) I get around 14fps. 

Quick overview:

1. As output of the training process splats are represented as ellipsoids with a position, scale, color and shape that are later projected to 2D ellipses.
2. Each splat is rendered as a quad with a color that gets positioned, scaled and shaped based on each ellipse characteristics.
3. Splats are sorted in a worker. Splats are rendered as instances and we have to sort them out to render them in the correct order (back to front)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -84,6 +84,13 @@
       }
     },
     {
+      /* This module uses ES8 async / await due to WebXR Anchor Module integration */
+      "files": ["./src/components/splat.js"],
+      "parserOptions": {
+        "ecmaVersion": 8
+      }
+    },
+    {
       /* This code is external, and the ES5 restrictions do not apply to it. */
       "files": ["./src/lib/**/*.js"],
       "parserOptions": {
--- a/src/components/splat.js
+++ b/src/components/splat.js
@@ -1,0 +1,910 @@
+  /* global Worker, Blob, fetch */
+  import { registerComponent } from '../core/component.js';
+  import THREE from '../lib/three.js';
+
+  registerComponent('splat', {
+    schema: {
+      src: {type: 'asset'},
+      cutoutEntity: {type: 'selector'},
+      pixelRatio: {type: 'number', default: 1},
+      xrPixelRatio: {type: 'number', default: 0.5}
+    },
+
+    init: function () {
+      this.initWorker();
+    },
+
+    update: function () {
+      var data = this.data;
+      var sceneEl = this.el.sceneEl;
+      if (data.src) { this.loadSplat(); }
+      if (data.pixelRatio > 0) {
+        sceneEl.renderer.setPixelRatio(this.data.pixelRatio);
+      }
+      if (data.xrPixelRatio > 0) {
+        sceneEl.renderer.xr.setFramebufferScaleFactor(this.data.xrPixelRatio);
+      }
+    },
+
+    loadSplat: function () {
+      var sceneEl = this.el.sceneEl;
+      var src = this.data.src;
+
+      if (!src) { return; }
+
+      // Postpone if scene not loaded.
+      if (!sceneEl.hasLoaded) {
+        sceneEl.addEventListener('loaded', this.loadSplat.bind(this));
+        return;
+      }
+
+      if (this.data.cutoutEntity) {
+        this.worldToCutoutMatrix = new THREE.Matrix4();
+        this.cutout = this.data.cutoutEntity.object3D;
+      }
+
+      var isPly = src.endsWith('.ply');
+      this.loadedVerticesNumber = 0;
+      this.worker.postMessage({method: 'clear'});
+      if (isPly) {
+        this.loadPlyFile(src);
+      } else {
+        this.loadSplatFile(src);
+      }
+    },
+
+    loadPlyFile: async function loadPlyFile (src) {
+      var response = await fetch(src);
+      var reader = response.body.getReader();
+
+      var rowLength = 3 * 4 + 3 * 4 + 4 + 4;
+      var bytesDownloaded = 0;
+      var bytesFileTotal = response.headers.get('Content-Length');
+      bytesFileTotal = bytesFileTotal ? parseInt(bytesFileTotal) : undefined;
+      var chunks = [];
+      var start = Date.now();
+      var lastReportedProgress = 0;
+
+      while (true) {
+        try {
+          var dataReceived = await reader.read();
+          if (dataReceived.done) {
+            console.log('Completed download.');
+            break;
+          }
+          var newChunk = dataReceived.value;
+          bytesDownloaded += newChunk.length;
+          chunks.push(newChunk);
+
+          // Downloar progress stats.
+          if (bytesFileTotal) {
+            var mbps = (bytesDownloaded / 1024 / 1024) / ((Date.now() - start) / 1000);
+            var percent = bytesDownloaded / bytesFileTotal * 100;
+            if (percent - lastReportedProgress > 1) {
+              console.log('Download progress:', percent.toFixed(2) + '%', mbps.toFixed(2) + ' Mbps');
+              lastReportedProgress = percent;
+            }
+          } else {
+            console.log('Download progress:', bytesDownloaded, ', unknown total');
+          }
+        } catch (error) {
+          console.error(error);
+          break;
+        }
+      }
+
+      if (bytesDownloaded === 0) { return; }
+      var chunksBufferLength = chunks.reduce(function add (sum, chunk) { return sum + chunk.length; }, 0);
+      var concatenatedChunks = new Uint8Array(chunksBufferLength);
+
+      // Concatenate the chunks into a single Uint8Array
+      var offset = 0;
+      for (var chunk of chunks) {
+        concatenatedChunks.set(chunk, offset);
+        offset += chunk.length;
+      }
+
+      concatenatedChunks = new Uint8Array(this.processPlyBuffer(concatenatedChunks.buffer));
+      var numVertices = Math.floor(concatenatedChunks.byteLength / rowLength);
+      await this.initGL(numVertices);
+      this.pushDataBuffer(concatenatedChunks.buffer, numVertices);
+    },
+
+    loadSplatFile: async function loadSplatFile (src) {
+      var response = await fetch(src);
+      var reader = response.body.getReader();
+
+      var rowLength = 3 * 4 + 3 * 4 + 4 + 4;
+      var bytesDownloaded = 0;
+      var bytesProcessed = 0;
+      var bytesFileTotal = response.headers.get('Content-Length');
+      bytesFileTotal = bytesFileTotal ? parseInt(bytesFileTotal) : undefined;
+
+      if (bytesFileTotal) {
+        var numVertices = Math.floor(bytesFileTotal / rowLength);
+        await this.initGL(numVertices);
+      }
+
+      var chunks = [];
+      var start = Date.now();
+      var lastReportedProgress = 0;
+
+      while (true) {
+        try {
+          var dataReceived = await reader.read();
+          if (dataReceived.done) {
+            console.log('Completed download.');
+            break;
+          }
+          var newChunk = dataReceived.value;
+          bytesDownloaded += newChunk.length;
+          chunks.push(newChunk);
+
+          // Downloar progress stats.
+          if (bytesFileTotal) {
+            var mbps = (bytesDownloaded / 1024 / 1024) / ((Date.now() - start) / 1000);
+            var percent = bytesDownloaded / bytesFileTotal * 100;
+            if (percent - lastReportedProgress > 1) {
+              console.log('Download progress:', percent.toFixed(2) + '%', mbps.toFixed(2) + ' Mbps');
+              lastReportedProgress = percent;
+            }
+          } else {
+            console.log('Download progress:', bytesDownloaded, ', unknown total');
+          }
+
+          var bytesToProcess = bytesDownloaded - bytesProcessed;
+          if (bytesFileTotal && bytesToProcess > rowLength) {
+            var concatenatedChunksbuffer = new Uint8Array(bytesToProcess);
+            var offset = 0;
+            for (var chunk of chunks) {
+              concatenatedChunksbuffer.set(chunk, offset);
+              offset += chunk.length;
+            }
+
+            var chunkBytesProcessed = this.processSplatChunk(concatenatedChunksbuffer, bytesToProcess);
+            bytesProcessed += chunkBytesProcessed;
+
+            // Reset chunks array. Copy leftover bytes if chunks not perfect multiple of a splat data structure.
+            chunks.length = 0;
+            if (bytesToProcess > chunkBytesProcessed) {
+              var extraData = new Uint8Array(bytesToProcess - chunkBytesProcessed);
+              extraData.set(concatenatedChunksbuffer.subarray(bytesToProcess - extraData.length, bytesToProcess), 0);
+              chunks.push(extraData);
+            }
+          }
+        } catch (error) {
+          console.error(error);
+          break;
+        }
+      }
+    },
+
+    processSplatChunk: function (chunksBuffer, chunkSize) {
+      var rowLength = 3 * 4 + 3 * 4 + 4 + 4;
+      var vertexCount = Math.floor(chunkSize / rowLength);
+      var bytesProcessed = vertexCount * rowLength;
+      var chunkBuffer = new Uint8Array(bytesProcessed);
+      chunkBuffer.set(chunksBuffer.subarray(0, chunkBuffer.byteLength), 0);
+      this.pushDataBuffer(chunkBuffer.buffer, vertexCount);
+      return bytesProcessed;
+    },
+
+    initGL: async function initGL (numVertices) {
+      console.log('initGL', numVertices);
+      this.el.object3D.frustumCulled = false;
+      var renderer = this.el.sceneEl.renderer;
+      var gl = renderer.getContext();
+      var maxTextureSize = gl.getParameter(gl.MAX_TEXTURE_SIZE);
+      this.maxVertices = maxTextureSize * maxTextureSize;
+
+      // Clamp number of vertices to maximum texture size.
+      if (numVertices > this.maxVertices) {
+        console.log('Number of vertices limited to ', this.maxVertices, numVertices);
+        numVertices = this.maxVertices;
+      }
+
+      this.bufferTextureWidth = maxTextureSize;
+      this.bufferTextureHeight = Math.floor((numVertices - 1) / maxTextureSize) + 1;
+
+      this.centerAndScaleData = new Float32Array(this.bufferTextureWidth * this.bufferTextureHeight * 4);
+      this.covAndColorData = new Uint32Array(this.bufferTextureWidth * this.bufferTextureHeight * 4);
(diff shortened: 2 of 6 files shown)


src/components/splat.js
```diff
@@ -0,0 +1,910 @@
+  /* global Worker, Blob, fetch */
+  import { registerComponent } from '../core/component.js';
+  import THREE from '../lib/three.js';
+
+  registerComponent('splat', {
+    schema: {
+      src: {type: 'asset'},
+      cutoutEntity: {type: 'selector'},
+      pixelRatio: {type: 'number', default: 1},
+      xrPixelRatio: {type: 'number', default: 0.5}
```
mrxz, line 10: These two properties don't belong on the component. They should probably be made part of the `renderer` system. Though the pixelRatio is a bit tricky given that we default to `window.devicePixelRatio`, so maybe express it as a factor on top of the `devicePixelRatio`?

src/components/splat.js
```diff
@@ -0,0 +1,910 @@
...
+      var data = this.data;
+      var sceneEl = this.el.sceneEl;
+      if (data.src) { this.loadSplat(); }
+      if (data.pixelRatio > 0) {
+        sceneEl.renderer.setPixelRatio(this.data.pixelRatio);
+      }
+      if (data.xrPixelRatio > 0) {
+        sceneEl.renderer.xr.setFramebufferScaleFactor(this.data.xrPixelRatio);
+      }
+    },
```
mrxz, line 27: Update doesn't handle changes to `cutoutEntity`.

src/components/splat.js
```diff
@@ -0,0 +1,910 @@
...
+          var dataReceived = await reader.read();
+          if (dataReceived.done) {
+            console.log('Completed download.');
+            break;
+          }
+          var newChunk = dataReceived.value;
+          bytesDownloaded += newChunk.length;
+          chunks.push(newChunk);
+
+          // Downloar progress stats.
```
mrxz, line 79: Typo
```suggestion
          // Download progress stats.
```

src/components/splat.js
```diff
@@ -0,0 +1,910 @@
...
+      bytesFileTotal = bytesFileTotal ? parseInt(bytesFileTotal) : undefined;
+      var chunks = [];
+      var start = Date.now();
+      var lastReportedProgress = 0;
+
+      while (true) {
+        try {
+          var dataReceived = await reader.read();
+          if (dataReceived.done) {
+            console.log('Completed download.');
```
mrxz, line 72: I'd prefer no logging from components, but if we do it should be using `debug`.

src/components/splat.js
```diff
@@ -0,0 +1,910 @@
...
+      var bytesProcessed = vertexCount * rowLength;
+      var chunkBuffer = new Uint8Array(bytesProcessed);
+      chunkBuffer.set(chunksBuffer.subarray(0, chunkBuffer.byteLength), 0);
+      this.pushDataBuffer(chunkBuffer.buffer, vertexCount);
+      return bytesProcessed;
+    },
+
+    initGL: async function initGL (numVertices) {
+      console.log('initGL', numVertices);
+      this.el.object3D.frustumCulled = false;
```
mrxz, line 194: This doesn't seem necessary. The actual mesh has frustum culling disabled as well further down in this method.

src/components/splat.js
```diff
@@ -0,0 +1,910 @@
...
+      // Wait until texture is ready
+      while (true) {
+        var centerAndScaleTextureProperties = renderer.properties.get(this.centerAndScaleTexture);
+        var covAndColorTextureProperties = renderer.properties.get(this.covAndColorTexture);
+        if (centerAndScaleTextureProperties && centerAndScaleTextureProperties.__webglTexture &&
+          covAndColorTextureProperties && centerAndScaleTextureProperties.__webglTexture) {
+          break;
+        }
+        await new Promise(resolve => setTimeout(resolve, 10));
+      }
```
mrxz, line 282: Instead of this polling it should just tell Three.js to initialize the textures: https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer.initTexture

.eslintrc.json
```diff
@@ -83,6 +83,13 @@
         "ecmaVersion": 6
       }
     },
+    {
+      /* This module uses ES8 async / await due to WebXR Anchor Module integration */
```
mrxz, line 87: Comment incorrectly refers to WebXR Anchor Module. That being said, I don't believe `async/await` is a hard requirement for `splat.js` and it can be rewritten in the A-Frame code style for consistency, avoiding this exception altogether.

mrxz:
My main concern with a `splat` component in the core is that things evolve quickly, so we'd might end up with a component that is quickly outdated. At the very least let's explicitly name it something like `3d-gaussian-splat`, to not confuse it with 2DGS or other splatting techniques.

I haven't really been following Gaussian splatting developments all that closely, but quadjr's aframe component is pretty old. How does it compare to implementations in [PlayCanvas](https://github.com/playcanvas/engine/tree/main/src/scene/gsplat), [8thwall/Scaniverse](https://www.8thwall.com/docs/studio/guides/splats/) or [gsplat.js](https://github.com/huggingface/gsplat.js)? It seems that most have switched to compressed and quantized formats, although it seems no clear standard file format has emerged.

> Based on https://github.com/quadjr/aframe-gaussian-splatting implementation

Please include the original license at the start of this file.

> In VR on a Quest 3 when the sample splat is fully loaded (~1.1M splats rendered) I get ~14fps.

Into the Scaniverse (https://www.intothescaniverse.com/) mostly runs at 72fps on a Quest 3, so that's probably something we should aim for as well. Obviously depends on the amount of splats amongst other factors, but a small object-centered example would probably work better. Could even make it grabbable.

arpu:
from my perspective its to early to add this in the main branch, better to external git, ther eis a lot of discussion going on to bring splats to gltf/glb ( with mesh opt compression) 

https://github.com/KhronosGroup/glTF/issues/2454

https://github.com/CesiumGS/glTF/tree/proposal-KHR_gaussian_splatting/extensions/2.0/Khronos/KHR_gaussian_splatting

dmarcos:
> from my perspective its to early to add this in the main branch, better to external git, ther eis a lot of discussion going on to bring splats to gltf/glb ( with mesh opt compression)
> 
> [KhronosGroup/glTF#2454](https://github.com/KhronosGroup/glTF/issues/2454)
> 
> https://github.com/CesiumGS/glTF/tree/proposal-KHR_gaussian_splatting/extensions/2.0/Khronos/KHR_gaussian_splatting

A-Frame role has always been at the forefront: We built on custom elements v1 before being standard. We also shipped WebVR, WebXR + extensions (hand tracking, layers, immersive navigation…) and gltf before becoming standard.

Standards are better designed when based on real world usage and feedback. Should be practice consolidation not prescription. Standard for gs3d probably premature.

A-Frame mission is to make 3D content creation accesible to anyone. I see exciting opportunities at the intersection with AI. I would love for A-Frame to participate and be top of mind of researchers and early adopters.

koktavy:
It would be very exciting to see A-Frame adopt 3DGS, even as part of an extras/experimental package. And A-Frame implementation of @mkkellogg's Three.js 3DGS viewer might be a better path to keep up with active development efforts. 

https://github.com/mkkellogg/GaussianSplats3D/tree/main

src/components/splat.js
dmarcos: yeah. don't love the API yet either

src/components/splat.js
dmarcos: yeah. I'll remove. still WIP.

dmarcos:
> It would be very exciting to see A-Frame adopt 3DGS, even as part of an extras/experimental package. And A-Frame implementation of @mkkellogg's Three.js 3DGS viewer might be a better path to keep up with active development efforts.
> 
> https://github.com/mkkellogg/GaussianSplats3D/tree/main

Thanks! I'm excited too.

Looked at mkkellog implementation but it's pretty complicated / advanced already. I went with https://github.com/quadjr/aframe-gaussian-splatting that is a very minimal port of the original (first?) WebGL implementation https://github.com/antimatter15/splat

My goal is to first have a minimal and performant implementation that we can later use as a base for more complex stuff. I would love to later have an example of a nice and simple viewer with great controls on every platform.

mrxz:
> My goal is to first have a minimal and performant implementation that we can later use as a base for more complex stuff. I would love to later have an example of a nice and simple viewer with great controls on every platform.

A good chunk of the complexity of [mkkellogg/GaussianSplats3D](https://github.com/mkkellogg/GaussianSplats3D/) seems to be to make it performant and to support a wide array of formats/splat types. Both things that I'd argue are essential for a `splat` component in A-Frame to be useful. 

I don't really see the benefit in re-inventing the wheel in this case. Rendering of Gaussian splats is very expensive, so unless/until performance is on-par, wouldn't user just create their own components or use a community made one that does wrap any of the more mature implementations out there?

dmarcos:
I limited the number of splats rendered at any given time by adding the follow statement in [line 532](https://github.com/aframevr/aframe/pull/5658/files#diff-e4515112be40fec1b8989e3765af7a0b49c61b7f1f53b0d30092554a17df8356R531)

```
 if (validCount > 150000) { break; }
```

We can render at 90fps on Quest up to ~150k splats.

mrxz:
> Wheel far from invented. Stuff evolving quick.

On the training, compression, rendering and sorting side there's plenty to experiment with. But this is about making it a core component of A-Frame. From the perspective of the user the relevance is being able to add "splats" to their scene. People will expect to be able to use the output of tools like SuperSplat and Scaniverse and have it perform comparable as well.

In other words, we'd end up implementing loading of compressed .PLY files, .SPZ files, supporting the spherical harmonics for view dependent lighting and implementing the sorting in WASM using SharedArrayBuffer, progressive loading, perhaps even 2DGS support. All things that have been done numerous times.

For experimenting, the core isn't the right place IMO. And for providing a component for users, we'd end up implementing (and maintaining!) a lot of features. I agree with @koktavy that wrapping [mkkellogg/GaussianSplats3D](https://github.com/mkkellogg/GaussianSplats3D/) is a better path to stay up to date. We can always contribute upstream if we do find improvements.

> mkkellogg performance is ~24fps on Quest 3 on the [simple demo scene](https://projects.markkellogg.org/threejs/gaussian_splats_3d_demos/vr.html?mode=0). Playcanvas ~17fps (similar to ours) on the same truck splat on this PR's example.

Both of those render at full resolution. I'm not able to reproduce ~17fps using the component in this PR when setting the `xrPixelRatio` to `1.0`, so there's quite a bit of catching-up to do. Ideally there'd be an apples-to-apples comparison with various implementations all rendering the same scene under the same conditions.

> Scenes from [Into the Scaniverse](https://www.intothescaniverse.com/) are the only ones I've seen decently performing on Quest 3 (~72fps). Presumably cherry picked.

Not sure if they really cherry pick them, as they do allow users to create their own and post them on the map. From what I gather they do a lot of pre-processing. One thing I have noticed is that the background is actually static despite the Gaussian Splat appearance. This presumably helps quite a bit as it saves on the number of splats as well as reduces overdraw caused by large splats in the background. 

> We can render at 90fps on Quest3 up to ~150k splats.

Splats don't have a fixed cost associated with them. They vary in size and spatial distribution. Overdraw really kills performance AFAICT, so ~150k splats representing a "full scene" or ~150k focussed on a single object will likely perform wildly different.

mrxz:
> Any example of mkkellog working at least 72fps on Quest?

I'm not aware of any example, but it would be good to have one (or more) files that are known to be able to rendered at these framerates. Personally I'd probably rip them from 'Into the Scaniverse', but those obviously can't be shared. I did find this repo hosting a demo splat with the claim that they've managed to render it at 90fps in WebXR on a Quest 3: https://github.com/gmix-tech/small_splats

sarbo82:
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

vincentfretin:
> > Any example of mkkellog working at least 72fps on Quest?
> 
> I'm not aware of any example, but it would be good to have one (or more) files that are known to be able to rendered at these framerates. Personally I'd probably rip them from 'Into the Scaniverse', but those obviously can't be shared. I did find this repo hosting a demo splat with the claim that they've managed to render it at 90fps in WebXR on a Quest 3: https://github.com/gmix-tech/small_splats

The https://github.com/gmix-tech/small_splats/blob/main/ply/jade_drums.ply model (240k points) indeed runs at 90 fps on my Quest 3 with this splat component, and also with mkkellogg with `splatAlphaRemovalThreshold: 20`.
It's far more better quality on my desktop with mkkellogg compared to that splat component, on VR it seems similar.
For the truck (1.5 million), the splat component renders horribly seeing really points, running at 13 fps in VR. With mkkellogg it's far better quality, at 15-20fps with `truck/truck_high.ksplat` and `sphericalHarmonicsDegree: 2` and `antialias: false` on the renderer.
I tested the garden, it's 5 millions, 4fps, unusable in VR.

The https://superspl.at/view?id=7656a8ed interior model (1 million) from https://www.youtube.com/watch?v=LB1kop59J38 runs at 30 fps with mkkellogg, better than the 15fps directly on supersplat actually.

vincentfretin:
If you want to experiment with https://github.com/mkkellogg/GaussianSplats3D here is the component I did today.
I can't put that on glitch, gtlich doesn't support setting the headers as far as I know so you will get the error
`Uncaught (in promise) DataCloneError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': SharedArrayBuffer transfer requires self.crossOriginIsolated.`
So here is the code:

in webpack.config.js add:
```js
    headers: {
      'Cross-Origin-Embedder-Policy': 'require-corp',
      'Cross-Origin-Opener-Policy': 'same-origin',
    }
```

Run the repo with
```
npm run start:https
```

create `examples/showcase/gaussian-splat/index.html`
```html
<html>
  <head>
     <script type="importmap">
      {
        "imports": {
          "@mkkellogg/gaussian-splats-3d": "https://cdn.jsdelivr.net/npm/@mkkellogg/gaussian-splats-3d@0.4.7/build/gaussian-splats-3d.module.js",
          "aframe": "https://cdn.jsdelivr.net/gh/aframevr/aframe@cd543c6cbf52c81f14ac0d3a41cb9a883047c7ac/dist/aframe-master.module.min.js",
          "three": "https://cdn.jsdelivr.net/npm/super-three@0.173.5/build/three.module.js",
          "three/addons/": "https://cdn.jsdelivr.net/npm/super-three@0.173.5/examples/jsm/"
        }
      }
    </script>
    <script type="module">
/**
 * @license MIT
 */
import AFRAME from 'aframe';
import * as GaussianSplats3D from "@mkkellogg/gaussian-splats-3d";
AFRAME.registerComponent("kgaussiansplat", {
  schema: {
    src: { type: "string", default: "" },
    up: { type: "vec3", default: { x: 0, y: -1, z: 0 } },
    sphericalHarmonicsDegree: { type: "int", oneOf: [0, 1, 2], default: 0 },
    splatAlphaRemovalThreshold: { type: "int", min: 0, max: 255, default: 1 },
    sceneRevealMode: { type: "string", oneOf: ["Default", "Gradual", "Instant"], default: "Default" },
  },
  init() {
    this.defaultUp = new THREE.Vector3(0, 1, 0);
    this.viewer = new GaussianSplats3D.DropInViewer({
      dynamicScene: true, // if it is set to false, I don't see anything
      halfPrecisionCovariancesOnGPU: true,
      inMemoryCompressionLevel: 2, // for non ksplat files
      sceneRevealMode: GaussianSplats3D.SceneRevealMode[this.data.sceneRevealMode],
      sphericalHarmonicsDegree: this.data.sphericalHarmonicsDegree,
    });
    const showLoadingUI = false;
    this.viewer
      .addSplatScenes(
        [
          {
            path: this.data.src,
            splatAlphaRemovalThreshold: this.data.splatAlphaRemovalThreshold,
          },
        ],
        showLoadingUI
      )
      .then(() => {
        this.el.setObject3D("mesh", this.viewer);
      });
  },

  update(oldData) {
    if (
      (oldData.src !== undefined && oldData.src !== this.data.src) ||
      (oldData.splatAlphaRemovalThreshold !== undefined &&
        oldData.splatAlphaRemovalThreshold !== this.data.splatAlphaRemovalThreshold)
    ) {
      this.remove();
      this.init();
      return;
    }
    if (
      oldData.sphericalHarmonicsDegree !== undefined &&
      oldData.sphericalHarmonicsDegree !== this.data.sphericalHarmonicsDegree
    ) {
      this.viewer.setActiveSphericalHarmonicsDegrees(this.data.sphericalHarmonicsDegree);
    }

    // Calculate rotation from default up vector (0,1,0) to desired up vector
    if (!this.defaultUp.equals(this.data.up)) {
      if (this.data.up.x === 0 && this.data.up.y === -1 && this.data.up.z === 0) {
        this.el.object3D.rotation.set(0, Math.PI, Math.PI);
      } else {
        const targetUp = new THREE.Vector3().copy(this.data.up).normalize();
        const quaternion = new THREE.Quaternion();
        quaternion.setFromUnitVectors(targetUp, this.defaultUp);
        this.el.object3D.quaternion.copy(quaternion);
      }
    }
  },

  remove() {
    const mesh = this.el.getObject3D("mesh");
    if (mesh) {
      this.el.removeObject3D("mesh");
      mesh.dispose();
      // https://github.com/mkkellogg/GaussianSplats3D/blob/2dfc83e497bd76e558fe970c54464b17b5f5c689/src/DropInViewer.js#L118-L120
      // in v0.4.7 DropInViewer.dispose doesn't dispose those:
      mesh.callbackMesh.material.dispose();
      mesh.callbackMesh.geometry.dispose();
    }
    this.viewer = null;
  },
});
    </script>
  </head>
  <body>
    <a-scene inspector="url: https://cdn.jsdelivr.net/gh/c-frame/aframe-editor@1.7.x/dist/aframe-editor.min.js">
      <a-entity kgaussiansplat="src: jade_drums.ply; up: 0 -1 0; sphericalHarmonicsDegree: 0; splatAlphaRemovalThreshold: 20"></a-entity>
      <a-entity id="rig" position="0 0 2">
        <a-entity id="camera" position="0 1.6 0" camera look-controls wasd-controls></a-entity>
      </a-entity>
    </a-scene>
  </body>
</html>
```

Download https://github.com/gmix-tech/small_splats/blob/main/ply/jade_drums.ply or any ply, splat, ksplat, spz files and put it in the same directory and change the kgaussiansplat with the correct filename of course. You can enable the inspector to tweak the properties.

open https://localhost:8080/showcase/gaussian-splat/

vincentfretin:
With `sharedMemoryForWorkers: false` I can put it on Github Pages.
demo: https://vincentfretin.github.io/aframe-kgaussiansplat/
code: https://github.com/vincentfretin/aframe-kgaussiansplat

average-bit:
@dmarcos https://github.com/NYU-ICL/image-gs

PR #5661, opened 2025-02-12
Modernize the A-Frame+framework answer in FAQ

vincentfretin:
Most of the links were broken or point to very old version of the framework.
I tried to add useful general information about the different popular frameworks.
If someone use React, Vue or Svelte and have some links we can point to similar to what I did for SolidJS, then please share.


dmarcos:
Thanks

PR #5662, opened 2025-02-13
Fix some jsdoc errors found with eslint valid-jsdoc rule

vincentfretin:
Fix some jsdoc errors found with eslint valid-jsdoc rule and also uniformize style.

Tested with 
```
    "valid-jsdoc": [
      "error",
      {
        "requireReturn": false,
        "requireReturnType": true,
        "requireParamDescription": false,
        "requireReturnDescription": false,
        "requireParamType": true,
        "prefer": {
          "params": "param",
          "return": "returns"
        },
        "preferType": {
          "Any": "any",
          "array": "Array",
          "bool": "boolean",
          "Boolean": "boolean",
          "Number": "number",
          "Object": "object",
          "String": "string"
        }
      }
    ],
```

in rules in `.eslintrc.json`

I didn't enable the rule because there is still 220 problems of `Missing JSDoc for parameter 'x'` and `Missing JSDoc @returns for function.`

### DIFF
--- a/src/components/material.js
+++ b/src/components/material.js
@@ -163,9 +163,7 @@ export var Component = registerComponent('material', {
    * (Re)create new material. Has side-effects of setting `this.material` and updating
    * material registration in scene.
    *
-   * @param {object} data - Material component data.
-   * @param {object} type - Material type to create.
-   * @returns {object} Material.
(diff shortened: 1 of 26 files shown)


dmarcos:
Thanks!

PR #5663, opened 2025-02-14
Fix html errors in examples

vincentfretin:
**Description:**

Fix html errors in examples

**Changes proposed:**
- Fix missing start or end tags
- Remove end tag for self-closing tags


dmarcos:
Thanks!

PR #5665, opened 2025-02-14
Add prettier config for html files only

vincentfretin:
**Description:**

For those using vscode with prettier addon to format the html files, add a prettier config so the js embed in the html file match the repo style with single quote and no trailing comma.
The config only applies to html files in the examples directory.

If you want to reformat all files
```
npm install -D prettier
npx prettier --write 'examples/**/index.html'
```
but that does lots of changes.


dmarcos:
Thanks

PR #5666, opened 2025-02-18
Enable eslint indent rule

vincentfretin:
**Description:**

I added a prettier config in #5665 that applies 2 spaces indent only for html and js inside the html for new contributed examples.
For js files that 2 spaces indent wasn't enforced.

**Changes proposed:**
- Enable the eslint indent rule to use 2 spaces in js files, use proper options to minimize code changes
- Fix all indent errors

### DIFF
--- a/examples/js/spatial-ui/spatial-utils.js
+++ b/examples/js/spatial-ui/spatial-utils.js
@@ -269,9 +269,7 @@ SPATIAL.utils = {
       uc = i < s || i >= 3 * s ? ur : ul;
       vc = i < 2 * s ? vh : vl;
 
-      uvs.push(uc, vc,
-              uc + ul * cosa, vc + vl * sina,
-              uc + ul * cosb, vc + vl * sinb);
+      uvs.push(uc, vc, uc + ul * cosa, vc + vl * sina, uc + ul * cosb, vc + vl * sinb);
 
       phia = phib;
     }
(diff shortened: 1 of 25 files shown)


examples/js/spatial-ui/spatial-utils.js
```diff
@@ -269,9 +269,7 @@ SPATIAL.utils = {
       uc = i < s || i >= 3 * s ? ur : ul;
       vc = i < 2 * s ? vh : vl;
 
-      uvs.push(uc, vc,
-              uc + ul * cosa, vc + vl * sina,
-              uc + ul * cosb, vc + vl * sinb);
+      uvs.push(uc, vc, uc + ul * cosa, vc + vl * sina, uc + ul * cosb, vc + vl * sinb);
```
dmarcos, line 272: This rule makes the code less legible. It's good to be able to break a complicated list of parameters passed to a function in multiple lines

vincentfretin: We have the same line in one other place in this file, so I decided to do the same for this one, but I can do the other way around, put it in several lines at both places.
The indent rule with `"CallExpression": { "arguments": "first" }` allowing to align the other lines on the same position as the first argument just warned about 1 missing space indentation for this particular case.
I fixed it in the next commit, you're right that's more readable.

dmarcos:
Thanks!

PR #5668, opened 2025-02-19
Check for presence of UniformsLib/Utils instead of importing them from three/src

mrxz:
**Description:**
To support the `WebGPURenderer` of three.js, some modifications were made to `msdf.js` and `sdf.js` (#5655). By importing from `three/src/` directly a different copy is being included into the bundle. This extended to various Color/Matrix/Vector classes as well. Functionally this didn't cause any problems AFAICT, but it did bloat the bundle size (~19kiB to the minified build).

This PR replaces the imports with a check if they exist instead. This does mean that the fog uniforms are missing when using the `WebGPURenderer`, but these shaders currently don't work with `WebGPURenderer` anyway and even if they did, fog is handled differently when using it.

**Changes proposed:**
- Check if `UniformsUtils`/`UniformsLib` exist instead of importing them from `three/src/`


vincentfretin:
Thank you for checking that.

dmarcos:
Thanks!

PR #5669, opened 2025-02-19
Handle tock in AScene.render instead of onAfterRender

mrxz:
**Description:**
Currently the `tock` logic is handled in the `onAfterRender` callback on the scene. However, in case of post-processing or when multiple scene renders happen in one frame, this would fire tock at incorrect times or even multiple times.

This PR moves the `tock` handling logic into the `AScene.render` method, similar to how `tick` is handled. Care is taken to preserve the correct `camera` parameter. Although, I propose we deprecate this parameter, as I fail to see why its needed.

**Changes proposed:**
- Move `tock` handling into `AScene.render`


dmarcos:
Thanks!

PR #5670, opened 2025-02-19
Import directly from three and only use lib/three for AFRAME.THREE export

mrxz:
**Description:**
The `lib/three.js` file is used to augment the `THREE` namespace. This is needed to ensure backwards compatibility for the global `THREE` and `AFRAME.THREE`. But there is no reason for the core to use the global THREE and/or import from the `lib/three.js` file. Instead it can directly import from `three`, or `three/addons/` if needed.

This makes the code in core closer to vanilla three, improves type hints provided by the IDE and is one step closer to things like tree-shaking.

One challenge that did arise was the conditional initialization of the renderer to either `WebGLRenderer` or `WebGPURenderer`. By importing from `three` directly, Webpack was emitting a warning that the export `WebGPURenderer` was not found (or vice versa with `WEBGPU=true`). I did not find an easy way to resolve this, so opted for using the `webpack.ProvidePlugin` to provide the relevant renderer constructor to use. 
@vincentfretin let me know what you think.

**Changes proposed:**
- Import directly from `three` and `three/addons` in core.
- Always use `three` imports instead of relying on global `THREE` in core.


mrxz:
> I did not find an easy way to resolve this, so opted for using the `webpack.ProvidePlugin` to provide the relevant renderer constructor to use.
> @vincentfretin let me know what you think.

Actually, this approach didn't work, as it bakes the choice between WebGL/WebGPU into the output bundle. For the module without a bundled Three it should determine it at runtime. Otherwise it defeats the purpose.

Changed it closer to how it was before, but obfuscated it a bit to trick Webpack. Not ideal, but this was essentially what was already happening in the old approach.

vincentfretin:
Yes, I was going to say that it won't work for the webgpu example using `npm run start:https`. It worked with `npm run start:webgpu` but that command is mainly to test the other examples that are not based on importmap.
Great you found a solution that work in the end.

vincentfretin:
Can you please also do it in the test
```diff
git diff tests
diff --git a/tests/core/scene/a-scene.test.js b/tests/core/scene/a-scene.test.js
index e1803177..762ead61 100644
--- a/tests/core/scene/a-scene.test.js
+++ b/tests/core/scene/a-scene.test.js
@@ -76,9 +76,8 @@ suite('a-scene (without renderer) - WebXR', function () {
       // Mock renderer.
       assert.ok(sceneEl.renderer);
       // Mock renderer is not a real WebGLRenderer.
-      assert.notOk(
-          (THREE.WebGLRenderer && sceneEl.renderer instanceof THREE.WebGLRenderer) ||
-          (THREE.WebGPURenderer && sceneEl.renderer instanceof THREE.WebGPURenderer));
+      var rendererImpl = ['WebGLRenderer', 'WebGPURenderer'].find(function (x) { return THREE[x]; });
+      assert.notOk(sceneEl.renderer instanceof THREE[rendererImpl])
     });
   });
```

You can run the tests with
```
npm run test:chrome
```
and
```
WEBGPU=true npm run test:chrome
```
We should maybe add testing with WebGPURenderer in CI too.

mrxz:
Done

> We should maybe add testing with WebGPURenderer in CI too.

It wouldn't hurt to have some testing with WebGPURenderer in the CI. Though I don't think we need to run the full test suite. The biggest risk is probably changes that (implicitly) rely/import something `WebGLRenderer` specific, breaking `WebGPURenderer`.

dmarcos:
Thanks!

PR #5671, opened 2025-02-20
Fix local development instructions

vincentfretin:
Fix local development instructions, port is 8080 and add the start:https command

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,21 @@ cd aframe && npm install  # Install dependencies.
 npm start  # Start the local development server.
 ```
 
-And open in your browser **[http://localhost:9000](http://localhost:9000)**.
+And open in your browser [http://localhost:8080](http://localhost:8080).
+
+If you want to test the examples in VR, you need to run it with https:
+
+```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -182,7 +182,15 @@ cd aframe && npm install  # Install dependencies.
...
-And open in your browser **[http://localhost:9000](http://localhost:9000)**.
+And open in your browser [http://localhost:8080](http://localhost:8080).
+
+With https:
+
+```sh
+npm run start:https
+```
+
+And open in your browser [https://localhost:8080](https://localhost:8080).
```
mrxz, line 193: Perhaps a good idea to add that the certificate is self-signed and the browser will show a warning screen when (first) accessing. Can imagine people unfamiliar with the details will think something is wrong at first.

vincentfretin: Good idea, I rewrote that, better now?

mrxz: Yes, nice that it now also states when/why you'd want https.

dmarcos:
Thanks!

PR #5676, opened 2025-02-23
Fix broken links.

diarmidmackenzie:
**Description:**

The current links don't work

**Changes proposed:**

Change them to links that do work.
Looks like an error left over from #5213.


dmarcos:
Thanks!

PR #5682, opened 2025-03-04
Put back the process.browser === true check in isBrowserEnvironment

### LINKED ISSUE #5679: Css styles are not applied with next.js
Hello. The height of the canvas does not match the size of the window. stats styles are missing. The problem appeared only in version 1.7, in 1.6 everything works fine. The problem manifests itself on nextj. This example is from an empty project.

**page.tsx:**
`"use client"
import {useEffect, useState} from "react";

export default function Home() {

    const [isRenderer, setIsRenderer] = useState(false)

    useEffect(() => {
        if (window !== undefined) {
            import("aframe");
            setIsRenderer(true)
        }
    }, []);

    if (!isRenderer)
        return null

    return <a-scene stats={true}>
        <a-entity/>
        </a-scene>
}`

**package.json:**
`{
  "name": "testnextjs",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev --turbopack",
    "build": "next build",
    "start": "next start",
    "lint": "next lint"
  },
  "dependencies": {
    "aframe": "^1.7.0",
    "next": "15.2.0",
    "react": "^19.0.0",
    "react-dom": "^19.0.0"
  },
  "devDependencies": {
    "@eslint/eslintrc": "^3",
    "@types/node": "^20",
    "@types/react": "^19",
    "@types/react-dom": "^19",
    "eslint": "^9",
    "eslint-config-next": "15.2.0",
    "typescript": "^5"
  }
}`

Image

Image

vincentfretin:
**Description:**

NextJS client rendered don't show the styles in A-Frame 1.7.0 because of my commit https://github.com/aframevr/aframe/commit/84d4607ef2d84d9349badfccf211d0e54fcfe838 It worked on 1.6.0.
I didn't know that framework polyfilled process with process.browser when rendering client side.

This fixes #5679

**Changes proposed:**

- Put back the `process.browser === true` check in `isBrowserEnvironment` for Next.js with "use client" to have the styles

### DIFF
--- a/src/utils/device.js
+++ b/src/utils/device.js
@@ -171,8 +171,9 @@ export function isLandscape () {
  * Check if running in a browser or spoofed browser (bundler).
  * We need to check a node api that isn't mocked on either side.
  * `require` and `module.exports` are mocked in browser by bundlers.
+ * process is polyfilled with process.browser true by Next.js with "use client".
  */
-export var isBrowserEnvironment = typeof process === 'undefined';
+export var isBrowserEnvironment = typeof process === 'undefined' || process.browser === true;
 
 /**
  * Check if running in node on the server.


dmarcos:
Thanks!

Issue #5691, opened 2025-03-12
New module build with latest webpack not working with npm usage


vincentfretin:
I'm testing a custom build aframe-master.module.min.js directly in node_modules/aframe/dist in my project but that fails with
```
ERROR in ./node_modules/aframe/dist/aframe-master.module.min.js 1:119490-119519
Module not found: Error: Can't resolve './' in '/home/vincentfretin/myproject/node_modules/aframe/dist'
resolve './' in '/home/vincentfretin/myproject/node_modules/aframe/dist'
```
That 119490 column is
![Image](https://github.com/user-attachments/assets/7a9453a9-80d0-4a9b-a85a-6d8b20eafbb9)
using
```
new URL("./",import.meta.url)
```
with webpack 5.98.0
that line is just before including `node_modules/three/examples/jsm/utils/BufferGeometryUtils.js`
https://github.com/supermedium/three.js/blob/super-r173-5/examples/jsm/utils/BufferGeometryUtils.js#L14
but I don't think that's relevant.

Previous build was this:
![Image](https://github.com/user-attachments/assets/fa44498c-0341-4b82-b660-752442f35d2d)
using
```
document.baseURI||self.location.href
```
but I don't know from which webpack version it was built with because we don't commit the package-lock.json and I completely removed package-lock.json and node_modules a few days ago.

@mrxz what version of webpack do you currently have on your machine in the aframe repo?

vincentfretin:
I checked different versions with `grep import.meta.url dist/aframe-master.module.min.js`
The generation changed in 5.96.0 to use import.meta.url.
https://github.com/webpack/webpack/releases
so reverting to 5.95.0 Sep 25, 2024 would fix the issue.

mrxz:
The version I had in `package-lock.json` was 5.92.0. After updating to 5.98.0 I also see the same behaviour, so it's definitely a change (regression?) in Webpack.

A quick look suggests that it might be the same issue as this one: https://github.com/webpack/webpack/issues/19209. Good news is that a fix has been made, but AFAICT hasn't been released yet. So we'll have to wait and see.

Is there a specific reason we don't commit the `package-lock.json` file? @dmarcos

vincentfretin:
the v1.7.0 tag and also latest master build doesn't have import.meta.url
so the aframe bot still use an old webpack version. @dmarcos can you confirm which version is currently used by the bot looking at the version in node_modules/webpack/package.json?
[...]

PR #5692, opened 2025-03-12
Use webpack 5.95.0 to have a correct module build

vincentfretin:
**Description:**

webpack 5.96.0-5.98.0 versions break the `aframe-master.module.min.js` build because using `import.meta.url`, see #5691

**Changes proposed:**
- Specify an exact version of webpack in `package.json`, latest working version was 5.95.0


dmarcos:
Thanks

PR #5695, opened 2025-03-15
Fix the npm link to really search for aframe-component tag

vincentfretin:
Fix the npm search link and mention the A-Frame Wiki for a community-maintained list of components

Same change on the site in https://github.com/aframevr/aframe-site/pull/555


dmarcos:
Thanks!

PR #5696, opened 2025-03-25
Support bloom component with initial enabled:false

vincentfretin:
**Description:**

The bloom component wasn't working properly with initial enabled:false, it still enabled the bloom effect.

**Changes proposed:**

- Be sure to create the composer and bind only if the component is enabled.


dmarcos:
Thanks

PR #5698, opened 2025-03-31
Fix minor error in documentation for reflection component

jgbarah:
**Description:**

There is a </a-scene> out of place in the example.

**Changes proposed:**

Fix the </a-scene> which is out of place.


dmarcos:
Thanks!

Issue #5397, opened 2023-11-28
Redo stats component


dmarcos:
With something simpler more lightweight. Current is heavy and overkill.

dmarcos:
Alternatives could be https://github.com/mrdoob/stats.js or https://github.com/RenaudRohlinger/stats-gl

arpu:
https://github.com/TheoTheDev/three-perf

dmarcos:
This shouldn't be too hard if someone wants to help

arvind-iyer:
Is there a preference on which of these three to use? three-perf seems to be showing even more information than the existing stats UI, has anyone checked if it significantly faster?

mrxz:
With WebGPU on the horizon, it's worth pointing out that `stats-gl` already supports it (despite its name).

Personally I don't think the `stats` component needs to show as much information as it currently does. The key point is showing the fps/raf and CPU/GPU timings. The other metrics can be nice, but have the potential downside of confusing users. It's not always trivial to interpret them, which can lead to people focussing on the wrong metrics when facing performance issues.

dmarcos:
Yeah. stats should be much simpler. We could even have it to link to brief explanations on what of each numbers mean and low hanging fruit recommendations on how to improve them.

Issue #5677, opened 2025-02-24
Screenshots blank when taken in post-processing example


jywarren:
First, AMAZING work especially on postprocessing in 1.7.0 - huge fan and so grateful.

Screenshots come out blank when I use this code, and I was able to confirm on the postprocessing example: https://aframe.io/aframe/examples/showcase/post-processing/ that this happens there too.

```js
document.querySelector('a-scene').components.screenshot.capture('equirectangular');
```

(using code from https://aframe.io/docs/1.7.0/components/screenshot.html#methods)

The following image results (and same for perspective screenshots):

![Image](https://github.com/user-attachments/assets/f325a587-ac79-4419-8354-97023c06baa9)

Is there an order of render change or something to do with postprocessing? The other showcase examples (in the [1.7.0 blog post](https://aframe.io/blog/aframe-v1.7.0/)) do not have this issue; they render normally. 

Thank you!!

jywarren:
Hi, thank you, is there any way to redirect the screenshot function to the alternative render target?

jywarren:
Oof, sorry, I would offer to try to help but this code is a little out of my depth:

https://github.com/aframevr/aframe/blob/e89db7d5ece9be5f912d4445a435ab0cec060e49/src/components/scene/screenshot.js#L143-L159

Is the postprocessing code's render target accessible such that this screenshot code could reference it?

https://github.com/aframevr/aframe/blob/6a497bf4d274b713dd6d81b7e6fc5a0e87015dd5/examples/showcase/post-processing/bloom.js#L50-L54

I don't quite know enough about the architecture to know but perhaps screenshot component could have something like `setRenderTarget()` or a parameter like `document.querySelector('a-scene').components.screenshot.capture('equirectangular', myRenderTarget);` ?

mrxz:
The problem isn't so much that the `EffectComposer` renders into its own render target, but more that it renders the final output to the screen (or WebXR swapchain texture). It will ignore the currently set render target, which breaks the `screenshot` component.

It is possible to instruct the `EffectComposer` to not output to the "screen" and use the final texture, however it shouldn't be the responsibility of the `screenshot` component to know about the post-processing. So there isn't really an easy fix.

[...]

PR #5700, opened 2025-04-06
Use stats-gl for A-Frame stats component

arvind-iyer:
**Description:**
Instead of the current stats implementation, here I have used stats-gl for the performance statistics as discussed in the issue thread #5397 

**Changes proposed:**
- Modify components/scene/stats.js to use stats-gl 
- Add stats-gl to the package.json
- Remove previously used rStatsAframe.js file 

Updates to documentation have not been done here yet


mrxz:
Thanks, nice to have a simpler `stats` component. There are still a couple more files related to `rStats` that can be removed as well, namely: [`vendor/rStats.js`](https://github.com/aframevr/aframe/blob/master/vendor/rStats.js), [`vendor/rStats.extras.js`](https://github.com/aframevr/aframe/blob/master/vendor/rStats.extras.js) and [`src/style/rStats.css`](https://github.com/aframevr/aframe/blob/master/src/style/rStats.css). As well as the line in `index.js` including the .css file:
https://github.com/aframevr/aframe/blob/29f29fdfd53709a242b913186b092fb08a40971a/src/index.js#L59

Functionally there is an annoying edge-case. With post-processing we override the `render` method similar to what `stats-gl` does under the hood. But this means that the order in which the components initialize matter. I'm wondering if we should perhaps manually call `stats.begin()/.end()/.update()` from `scene.js`. This would avoid that issue as well as include the `.tick/.tock` execution time in the CPU metric.

dmarcos:
Can we remove the unnecessary files? Can we deal with post-processing separately ideally without coupling with a-scene?

mrxz:
No worries, it's more thinking out loud than anything concrete. The way post-processing is done leads to conflicts with components that patch/hook into rendering (`screenshot` (https://github.com/aframevr/aframe/issues/5677) and `stats`). Ideally there'd be a clean interface for these components provided by `a-scene` as it controls the render loop.

The fact that the measured CPU time does not include any system or component logic is a bigger issue IMO. Users would expect it to reflect those as well. But merging this PR as-is, is fine. The above mentioned issue is something that needs to be addressed anyway, at which point the `stats` can be updated as well.

dmarcos:
Needs the tests fixed

dmarcos:
Thanks so much!

vincentfretin:
@arvind-iyer removing the stats component doesn't properly remove the two graphs, it just freeze them. You can test in the console with `sceneEl.removeAttribute("stats");` Can you please look at it?

vincentfretin:
The number of draw calls was an interesting metric for me, but I can still access the stats with a console.table:
```js
console.table(AFRAME.scenes[0].renderer.info.memory);
console.table(AFRAME.scenes[0].renderer.info.render);
```

@diarmidmackenzie your [stats-panel](https://diarmidmackenzie.github.io/aframe-components/components/stats-panel/) component will probably be broken by this change, you may want to get back some of the previous code into your component.

vincentfretin:
> @arvind-iyer removing the stats component doesn't properly remove the two graphs, it just freeze them. You can test in the console with `AFRAME.scenes[0].removeAttribute("stats");` Can you please look at it?

I fixed it in #5733

PR #5708, opened 2025-04-24
In layer component enableCompositorLayer verify this.quadPanelEl exists before toggling the visibility

vincentfretin:
**Description:**

In layer component `enableCompositorLayer` verify `this.quadPanelEl` exists before toggling the visibility.

**Changes proposed:**
- Add a if condition, similar to other places in the code. Also we may have a similar condition here to toggle a sphereEl visibility when we'll add support for equirect layer.


dmarcos:
Thanks

PR #5709, opened 2025-04-27
Use XRMediaBinding.createQuadLayer to render flat video on a layer

vincentfretin:
Use `XRMediaBinding.createQuadLayer` to render flat video on a layer.
Previous code was rendering at 90fps a 30fps video, and there were weird issues on the forest when you looked at the sides, also in rare cases I had the video flipped on the y axis.
I did this work as part of adding support for equirect stereo videos. The previous code had really bad performance issues with a 360 mono video and too complicated to modify to support stereo video. I'll make a separate PR to add 360 and 180 mono-equirect, stereo-left-right-equirect, stereo-top-bottom-equirect.

See the threejs webxr vr layers example:
https://threejs.org/examples/#webxr_vr_layers
and code
https://github.com/mrdoob/three.js/blob/master/examples/webxr_vr_layers.html

Tested on Quest3 on the example https://localhost:8080/test/layer/


vincentfretin:
Support for equirect layer is in my branch https://github.com/vincentfretin/aframe/tree/layer-equirect based on that PR changes. 
I still need to work on an interactive example.

dmarcos:
Thanks

Issue #5705, opened 2025-04-22
"anchored" component documentation issue


zakaton:
in the [documentation](https://aframe.io/docs/1.7.0/components/anchored.html) for the **anchored** component, it says to use the function

```el.components.createAnchor(position, quaternion)```

when it should be

```el.components.anchored.createAnchor(position, quaternion)```

including the component name `anchored`

vaibhavyxx:
Hi, I submitted a PR with the fix as you suggested in [PR #5711](https://github.com/aframevr/aframe/pull/5711), hope this helps!

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5711

PR #5711, opened 2025-05-01
Fixed the anchored documentation error

vaibhavyxx:
**Description:**
As suggested by #5705, I replaced the function el.components.createAnchor(position, quaternion) with el.components.anchored.createAnchor(position, quaternion) instead.

**Changes proposed:**
-Made the change in the function as mentioned above
-
-


dmarcos:
Thanks so much! Congrats on your first contribution! 🥳 👏

Issue #5704, opened 2025-04-22
"real-world-mesh" component's "meshMixin" is a boolean when it should be a string


zakaton:
in the [real-world-mesh](https://github.com/aframevr/aframe/blob/104801058452519c086b89dd06a13b39d5d4f9f0/src/components/scene/real-world-meshing.js#L16) component, the schema should have **meshMixin** be an empty string _''_ instead of a boolean _true_

before:

```
schema: {
    filterLabels: {type: 'array'},
    meshesEnabled: {default: true},
    meshMixin: {default: true}, // should be a string like planeMixin
    planesEnabled: {default: true},
    planeMixin: {default: ''}
  },
```

after:
```
schema: {
    filterLabels: {type: 'array'},
    meshesEnabled: {default: true},
    meshMixin: {default: ''}, // changed to empty string
    planesEnabled: {default: true},
    planeMixin: {default: ''}
  },
```

dmarcos:
Thanks. A PR always super appreciated

vaibhavyxx:
Hi, are you suggesting it to be a string because planeMixin is a string as well? Just wanted to know your thoughts, as I don't mind submitting a PR for it.

dmarcos:
Fixed by https://github.com/aframevr/aframe/pull/5712

PR #5712, opened 2025-05-04
Converted "real-world-mesh" component's meshMixin boolean into a string

vaibhavyxx:
**Description:**
Changed the variable of meshMixin into a string from bool

**Changes proposed:**
- As suggested by #5704, I changed the real-world-mesh component's meshMixin into a string


dmarcos:
Thanks so much

PR #5713, opened 2025-05-05
Update installation.md Python 3 command to `python3`

daj:
**Description:**

When following the instructions on macOS 15.4, after installing the recommended version of Python from python.org, I found that the Python 3 executable is `python3`, not `python`.

Some quick internet searching suggests that `python3` is probably the most reliable command to run across multiple OSes.

I confirmed the command seems to work as expected:
```
% python3 -m http.server
Serving HTTP on :: port 8000 (http://[::]:8000/) ...
::1 - - [04/May/2025 23:10:46] "GET /my-test-vr.html HTTP/1.1" 200 -
```

**Changes proposed:**

Simple docs change from `python` to `python3`.

### DIFF
--- a/docs/introduction/installation.md
+++ b/docs/introduction/installation.md
@@ -55,10 +55,10 @@ editors. Both support remixing or forking:
 For the options below, we should develop projects using a local server so that
 files are properly served. Options of local servers include:
 
-- Running `npm i -g five-server@latest && five-server --port=8000` in a terminal
+- Run `npm i -g five-server@latest && five-server --port=8000` in a terminal
   in the same directory as your HTML file.
-- Running `python -m SimpleHTTPServer` (or `python3 -m http.server` for Python 3)
-  in a terminal in the same directory as your HTML file.
+- Run `python3 -m http.server` in a terminal in the same directory as your
+  HTML file.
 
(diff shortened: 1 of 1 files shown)


dmarcos:
@mrxz @vincentfretin Any thoughts?

vincentfretin:
It's the same on latest ubuntu, at least on 22.04, there is no python alias anymore (that was pointing to python2), you need to execute python3 command or python2 explicitly.
I personally would get rid of the `python -m SimpleHTTPServer` completely in the doc now and only document the `python3 -m http.server` command. Python 2 end of life was January 1st, 2020.

mrxz:
As far as I've encountered, it's common to have `python3` and `python2` for their respective versions, with `python` symlinking to the "default" one, though virtually always python 2 for backwards compatibility.

I agree with @vincentfretin that we should just list the python3 command and remove the old one.

dmarcos:
@daj can we switch to `python3` as described above?

daj:
@dmarcos I updated it, please let me know if I need to make any other changes.

vincentfretin (approved):

dmarcos:
Thanks @dan and congrats on your first contribution 👏

PR #5717, opened 2025-05-10
Add minFilter and magFilter to material component replacing npot

mrxz:
**Description:**
As discussed in #5706 the `npot` property was only intended for working with non-power-of-two textures in WebGL 1. It's being used in several places to indirectly set the `minFilter` to `THREE.LinearFiltering`. This PR introduces `minFilter` and `magFilter` properties allowing the filtering methods to be set directly and explicitly.

**Changes proposed:**
- Add `minFilter` and `magFilter` properties to `material` component to set texture filtering.
- Remove `npot` property from `material` component

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -64,10 +64,11 @@ depending on the material type applied.
 | alphaTest    | Alpha test threshold for transparency.                                                                                                            | 0             |
 | depthTest    | Whether depth testing is enabled when rendering the material.                                                                                     | true          |
 | flatShading  | Use `THREE.FlatShading` rather than `THREE.StandardShading`.                                                                                      | false         |
-| npot         | Use settings for non-power-of-two (NPOT) texture.                                                                                                 | false         |
 | offset       | Texture offset to be used.                                                                                                                        | {x: 0, y: 0}  |
 | opacity      | Extent of transparency. If the `transparent` property is not `true`, then the material will remain opaque and `opacity` will only affect color.   | 1.0           |
 | repeat       | Texture repeat to be used.                                                                                                                        | {x: 1, y: 1}  |
+| magFilter    | Which magnifying filter to use when sampling textures. Can be one of `linear` or `nearest`.                                                       | `linear`      |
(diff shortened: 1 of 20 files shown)


dmarcos:
PR ended pretty big

mrxz:
> PR ended pretty big

Not really? The majority is the auto-generated attribute tables for primitives in the docs. Kind of unavoidable when material properties change.

The actual changes are minimal: `npot` out and `mag/minFilter` in:
```
 docs/components/material.md                       |  3 ++-
 src/components/layer.js                           |  2 +-
 src/components/material.js                        |  6 +++++-
 src/extras/primitives/primitives/a-sky.js         |  2 +-
 src/extras/primitives/primitives/a-videosphere.js |  2 +-
 src/utils/material.js                             | 23 +++++++++++------------
 6 files changed, 21 insertions(+), 17 deletions(-)
```

vincentfretin:
I verified with some console.log that by default on a texture we have
```
wrapS = THREE.ClampToEdgeWrapping;
wrapT = THREE.ClampToEdgeWrapping;
magFilter = THREE.LinearFilter;
minFilter = THREE.LinearMipmapLinearFilter;
```
so here the default on the minFilter and magFilter properties in this PR matches that and just setting minFilter to linear indeed does the same thing than the previous code.

Also I confirm this is still needed to set minFilter here.
I don't see any difference between minFilter LinearMipmapLinearFilter and LinearFilter on a 4096x2048 image on Quest 3
https://github.com/aframevr/aframe/blob/master/examples/boilerplate/panorama/puydesancy.jpg
but I see a difference in sharpness on a 8192x4096 image with a bit of flickering at times
https://github.com/c-frame/aframe-stereo-component/blob/master/examples/basic_image/textures/L.jpg
Of course it's even better with a mono equirect layer and there is no flickering at all.

So the changes in the PR looks good to me. Thanks for the work @mrxz 
If some third party components used `npot: true`, it will need to be updated to use `minFilter: linear` instead.

dmarcos:
Thanks!

PR #5718, opened 2025-05-12
Add layer cubemap example

vincentfretin:
Add layer cubemap example that @dmarcos did https://x.com/dmarcos/status/1731813890307260419 to the repository so we can test it to make sure we don't do any regression.


dmarcos:
Thanks. Should this be a showcase? more compelling and complex than a test

ig-raiderler:
Great implementation of this feature. I particularly like the approach to error handling.

vincentfretin:
In my opinion the layer quad example should also be in the Examples section in this case.
I moved both examples layer quad and layer cubemap examples from Tests to the Examples section, that way those are more visible.

dmarcos:
I meant should not be in the test directory. Tests is simple code for routine manual testing of individual features. this one is more elaborate

dmarcos:
Thanks so much!

PR #5720, opened 2025-05-14
Bunny video shouldn't autoplay in Layer Quad example

vincentfretin:
**Description:**

Sometimes when loading the test/layer example, I would hear the video sound, that video shouldn't autoplay, it should play only after clicking twice on Next.

**Changes proposed:**
- Remove autoplay attribute


dmarcos:
Thanks

PR #5721, opened 2025-05-14
Fix condition that check that XRWebGLBinding is defined in layer component

vincentfretin:
**Description:**

Fix condition that check that XRWebGLBinding is defined, also be sure to play the video even if layers is not supported.
Indeed if XRWebGLBinding is really undefined it would give an error. @dmarcos if you can verify on your AVP with some console.log if XRWebGLBinding is really undefined and also XRMediaBinding, that would be great.

I want that layer component to work on both Quest and AVP with the fallback to plane or spheres if layers is not supported currently on AVP. For this I think layers should be in webxr optionalFeatures and not requiredFeatures but using optionalFeatures with 'layers' just give an invisible layer without error on Quest it seems, that's another issue to be reported. So the changes here are only half the changes for the layer component to work on AVP.

**Changes proposed:**
- change condition from `!XRWebGLBinding` to `typeof XRWebGLBinding === 'undefined'`
- move the src.play() line above that condition

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -361,7 +361,8 @@ export var Component = registerComponent('layer', {
   onEnterVR: function () {
     var sceneEl = this.el.sceneEl;
     var xrSession = sceneEl.xrSession;
-    if (!sceneEl.hasWebXR || !XRWebGLBinding || !xrSession) {
+    if (this.data.src.play) { this.data.src.play(); }
+    if (!sceneEl.hasWebXR || typeof XRWebGLBinding === 'undefined' || !xrSession) {
       warn('The layer component requires WebXR and the layers API enabled');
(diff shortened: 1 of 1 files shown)


mrxz:
> For this I think layers should be in webxr optionalFeatures and not requiredFeatures but using optionalFeatures with 'layers' just give an invisible layer without error on Quest it seems, that's another issue to be reported.

That's probably due to the following logic in A-Frame:
https://github.com/aframevr/aframe/blob/650338363b1ead391683d91a9499550c61dcc220/src/core/scene/a-scene.js#L288

dmarcos:
Thanks!

vincentfretin:
follow up on #5722

PR #5722, opened 2025-05-15
Make layers feature optional for type quad so it works on devices not supporting WebXR layers

vincentfretin:
**Description:**

Make layers feature optional for type quad so it works on devices not supporting WebXR layers. Please confirm @dmarcos that the Layer Quad example is working with the fallback to a plane on AVP.

**Changes proposed:**
- Use optionalFeatures layers for types that have a fallback, requiredFeatures for types that don't have a fallback
- Add this.layerEnabled in the condition to initLayer(), layerEnabled is false for devices not defining XRWebGLBinding or XRMediaBinding (when I tested the VR button on Chrome desktop, XRWebGLBinding was defined, XRMediaBinding was undefined)

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -285,7 +285,7 @@ export class AScene extends AEntity {
                 self.usedOfferSession = false;
               }
 
-              vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1;
+              vrManager.layersEnabled = xrSession.enabledFeatures ? xrSession.enabledFeatures.indexOf('layers') !== -1 : true;
               vrManager.setSession(xrSession).then(function () {
                 vrManager.setFoveation(rendererSystem.foveationLevel);
                 self.xrSession = xrSession;
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -16,6 +16,7 @@ export var Component = registerComponent('layer', {
   init: function () {
     this.quaternion = new THREE.Quaternion();
     this.position = new THREE.Vector3();
+    this.layerEnabled = false;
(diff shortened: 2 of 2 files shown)


src/core/scene/a-scene.js
```diff
@@ -285,7 +285,7 @@ export class AScene extends AEntity {
                 self.usedOfferSession = false;
               }
 
-              vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1;
+              vrManager.layersEnabled = xrInit.requiredFeatures.indexOf('layers') !== -1 || xrInit.optionalFeatures.indexOf('layers') !== -1;
```
mrxz, line 288: The fact that `layers` is part of the `optionalFeatures` does not mean that it will be available in the session. The check should use [`enabledFeatures`](https://www.w3.org/TR/webxr/#dom-xrsession-enabledfeatures) to determine this.

Though, sadly, some older WebXR implementations like on the Pico Neo3 (Link) and Pico 4 don't implement this, and `enabledFeatures` is undefined...

vincentfretin: Thanks. Indeed I see such usage in threejs repo https://github.com/search?q=repo%3Amrdoob%2Fthree.js%20enabledFeatures&type=code
I'll change with the line you suggested.
I hope AVP have xrSession.enabledFeatures properly defined then, otherwise the example won't fallback to the plane, please confirm @dmarcos

vincentfretin:
I got confirmation from another user that XRWebGLBinding and XRMediaBinding are undefined on AVP.
It should be good to merge.

dmarcos:
Thanks!

PR #5724, opened 2025-05-16
Reintroduce a needsRedraw flag that can be set to true to update the quad layer if you use a canvas as src

vincentfretin:
**Description:**

If you use a canvas as src, and later update the canvas, the layer is not updated. I propose to reintroduce a needsRedraw flag on the component that can be set with `this.el.components.layer.needsRedraw = true` from another component that updated the canvas.

**Changes proposed:**
- Reintroduce a needsRedraw flag that can be set to true to update the quad layer if you use a canvas as src

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -16,6 +16,9 @@ export var Component = registerComponent('layer', {
   init: function () {
     this.quaternion = new THREE.Quaternion();
     this.position = new THREE.Vector3();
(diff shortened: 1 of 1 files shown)


vincentfretin:
I will need to rebase this one after #5722 is merged, I'll have a conflict.

dmarcos:
Thank you!

PR #5726, opened 2025-05-24
Set texture.generateMipmaps according to specified minFilter

vincentfretin:
**Description:**

Set `texture.generateMipmaps` according to specified `minFilter`.
Since https://github.com/mrdoob/three.js/pull/29677 that's included since three r170 you need to set `texture.generateMipmaps` to `false` explicitly if using `minFilter` 'linear' or 'nearest' to not generate mipmaps.

**Changes proposed:**
- add `texture.generateMipmaps = minFilter !== THREE.NearestFilter && minFilter !== THREE.LinearFilter;`

### DIFF
--- a/src/utils/material.js
+++ b/src/utils/material.js
@@ -52,6 +52,7 @@ export function setTextureProperties (texture, data) {
     texture.wrapT = wrapT;
     texture.magFilter = magFilter;
     texture.minFilter = minFilter;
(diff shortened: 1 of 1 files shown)


mrxz:
Change looks good to me. Especially now that we explicitly expose `mag/minFilter` as properties, it's nice to avoid generating mipmaps when we know they aren't needed.

dmarcos:
Thank you

PR #5727, opened 2025-05-24
Add migration instruction for npot: true to minFilter: linear

vincentfretin:
Add a migration note in future changelog about removal of `npot: true` in #5717 that needs to be replaced by `minFilter: linear` in your own code.


dmarcos:
Thank you

PR #5731, opened 2025-05-25
Fix inline link check in docsLint.js script

mrxz:
**Description:**
The `docsLint.js` script checks relative links to see if the file they point to exists. For inline links this check wasn't performed correctly, causing issue to go unreported. This PR fixes the script.

Fixes for incorrect links are incorporated in #5730 

**Changes proposed:**
- Properly handle outcome of link checking for inline links to report any errors


dmarcos:
Thank you!

PR #5733, opened 2025-06-01
Remove stats div from the DOM when removing stats component

vincentfretin:
**Description:**

#5700 changes didn't properly remove the added div when the stats component is removed.

**Changes proposed:**
- Remove stats div from the DOM when removing stats component


dmarcos:
Thank you!

PR #5735, opened 2025-06-02
Use `loadCubeMapTexture` instead of `CubeTextureLoader` for light probes

### LINKED ISSUE #5734: Probe Lighting Errors On Load and Doesn't Contribute to Scene Lighting
## Description

The [Probe Lighting](https://aframe.io/docs/1.7.0/components/light.html#probe) type (e.g., `<a-light type="probe"></a-light>`) fails to properly load necessary cubemaps and (silently) errors out resulting in no lighting contribution to the seen. This issue affects A-Frame versions 1.6.0 and higher.

### Probe Lighting in A-Frame 1.5.0 (Expected)

Probe Lighting in A-Frame 1.5.0

### Probe Lighting in A-Frame 1.60-1.7.1 (Bug)
Probe Lighting in A-Frame 1.6.0-1.7.1

## Glitch Code Examples

> [!IMPORTANT]
> The Glitch is set to A-Frame version 1.7.1 to see the bug. Change the version to 1.5.0 to see Probe Lighting working as expected.

- [Glitch | Code](https://glitch.com/edit/#!/silicon-periwinkle-eggplant)

- [Glitch | Live Site](https://glitch.com/edit/#!/silicon-periwinkle-eggplant)

### Code Example Console Error

`GET https://silicon-periwinkle-eggplant.glitch.me/[object%20HTMLImageElement] 404 (Not Found)   [object%20HTMLImageElement]:1`

### Three.js Loading Manager Error

Three.js loading manager reports the following console error when setup to provide verbose file loading information. The error seems to occur when the `THREE.LightProbeGenerator()` attempts to load the requisite cubemaps.

`Error loading: [object HTMLImageElement].`

One might expect instead of returning `[object HTMLImageElement]` that it returns the image URL instead.

### Analysis

I believe this breaking change may have been introduced in A-Frame 1.6.0 with this commit:

[Move envMap loading into material utils and system](https://github.com/aframevr/aframe/commit/e0dddcd3d9a439baf75afe316ebdc0dcb2e520ea)

Prior to 1.6.0 the A-Frame cubemap loader could only return image URLS. From 1.6.0 forward it can return image URLs *and* HTMLImageElements.

I believe the core issue is that passing the object type `HTMLImageElement` is incompatible with the Three.js [LightProbeGenerator](https://threejs.org/docs/?q=lightpro#examples/en/lights/LightProbeGenerator) (it expects image URLs) and so the lightprobe silently errors out and never contributes to the scene lighting.

It's also possible that the HTMLImageElement is incompatible with the Three.js [FileLoader](https://threejs.org/docs/?q=loader#api/en/loaders/FileLoader) and/or [CubeTextureLoader](https://threejs.org/docs/?q=cube#api/en/loaders/CubeTextureLoader) so it's worth considering if the issue is occuring at that level.

mrxz:
Fixes #5734 

**Description:**
Probe lights used an instance of `THREE.CubeTextureLoader` to create a `CubeTexture` which was then fed into the `LightProbeGenerator`. With #5454 the possible values for `srcs` for an envMap were broadened, meaning they were no longer guaranteed to be URLs, but could also contain `HTMLImageElement`.

The `loadCubeMapTexture` method on the material system already handles all possible values. This PR replaces the `light` component to use `loadCubeMapTexture` instead of manually using a `CubeTextureLoader` instance.

**Changes proposed:**
- Use `loadCubeMapTexture` when loading a cubemap texture for light probes

### DIFF
--- a/src/components/light.js
+++ b/src/components/light.js
@@ -6,7 +6,6 @@ import * as mathUtils from '../utils/math.js';
 
 var degToRad = THREE.MathUtils.degToRad;
 var warn = debug('components:light:warn');
-var CubeLoader = new THREE.CubeTextureLoader();
 
 var probeCache = {};
 
@@ -349,12 +348,12 @@ export var Component = registerComponent('light', {
     }
 
     // Populate the cache if not done for this envMap yet
+    var sceneEl = this.el.sceneEl;
     if (probeCache[data.envMap] === undefined) {
-      probeCache[data.envMap] = new window.Promise(function (resolve) {
-        srcLoader.validateCubemapSrc(data.envMap, function loadEnvMap (urls) {
-          CubeLoader.load(urls, function (cube) {
-            var tempLightProbe = LightProbeGenerator.fromCubeTexture(cube);
-            probeCache[data.envMap] = tempLightProbe;
+      probeCache[data.envMap] = new Promise(function (resolve) {
+        srcLoader.validateCubemapSrc(data.envMap, function loadEnvMap (srcs) {
+          sceneEl.systems.material.loadCubeMapTexture(srcs, function (texture) {
+            var tempLightProbe = LightProbeGenerator.fromCubeTexture(texture);
             resolve(tempLightProbe);
           });
         });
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #5739, opened 2025-06-20
Fix race condition in gltf-model when changing src while loading

mrxz:
**Description:**
The `gltf-model` component would unconditionally change the model in the `gltfLoad` callback. However, it's possible for the `src` value to have changed in the meantime. In case the second model finishes loading before the first one, the component would incorrectly end up showing the first model (once it finishes loading as well).

This PR adds a check to ensure that the `src` hasn't changed (similar to what is done for textures).

**Changes proposed:**
- Check if `src` changed while loading model to avoid showing the wrong model
- Always remove previous model, even when setting to an empty value
- Add unit tests covering the `src` change and removal logic.

### DIFF
--- a/src/components/gltf-model.js
+++ b/src/components/gltf-model.js
@@ -36,12 +36,13 @@ export var Component = registerComponent('gltf-model', {
     var el = this.el;
     var src = this.data;
 
-    if (!src) { return; }
-
     this.remove();
+
+    if (!src) { return; }
 
     this.ready.then(function () {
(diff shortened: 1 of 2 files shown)


vincentfretin (approved):
This looks great! And the added tests are testing indeed those cases.

@kfarr I think I remember seeing that race condition happening on 3dstreet with the model preview on hover when switching fast enough between cards when we used the same entity, and we currently remove and create a new entity because of that. That was one of the reason anyway.

dmarcos:
Thank you!

PR #5743, opened 2025-07-02
Fix jsdoc for parse and styleParse in styleParser.js, second param obj is optional

vincentfretin:
**Description:**

I got a typescript error in my project copying that parse and styleParse functions and using only one param, properly marking the second param as optional fixed it.
Note that if you're using `AFRAME.utils.styleParser.parse(astring)` directly there is no error because only the first param is in the `@types/aframe` types defined in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3829c3cb5e731141ebf73816117e5a5a6866be17/types/aframe/index.d.ts#L340


dmarcos:
Thanks!

PR #5753, opened 2025-08-15
feat(meta-touch-controls): add support for Meta Quest 3S Xbox Edition

mimosapudical:
**Description:**

**Changes proposed:**
-This PR adds support for the meta-quest-touch-plus-v2 profile in the meta-touch-controls component.
-Currently, the Xbox Edition model files are not included, so the component will fall back to using the existing meta-quest-touch-plus models.
-The profile recognition logic is implemented and ready for use.

### DIFF
--- a/src/components/meta-touch-controls.js
+++ b/src/components/meta-touch-controls.js
@@ -101,6 +101,20 @@ var CONTROLLER_PROPERTIES = {
       modelPivotOffset: new THREE.Vector3(0, 0, 0),
       modelPivotRotation: new THREE.Euler(0, 0, 0)
     }
+  },
+  'meta-quest-touch-plus-v2': {
+    left: {
+      modelUrl: META_CONTROLLER_MODEL_BASE_URL + 'quest-touch-plus-v2-left.glb',
+      rayOrigin: { origin: {x: 0.0065, y: -0.0186, z: -0.05}, direction: {x: 0.1239, y: -0.5944, z: -0.7945} },
+      modelPivotOffset: new THREE.Vector3(0, 0, 0),
+      modelPivotRotation: new THREE.Euler(0, 0, 0)
+    },
+    right: {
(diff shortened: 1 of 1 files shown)


mrxz:
Thanks for the PR. Do you happen to have the Meta Quest 3S Xbox Edition controllers?

The profiles list for these should report `meta-quest-touch-plus` as their first fallback. So even without any changes, I'd expect them to show up in A-Frame just fine (without the Xbox branded models). While having the profile explicitly supported is nice, it should not point towards non-existing model files in the `modelUrl`.

The models can be found in the `webxr-input-profiles` repository: https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/assets/profiles/meta-quest-touch-plus-v2
You can create a PR on [aframevr/assets](https://github.com/aframevr/assets) to add them.

mimosapudical:
> Thanks for the PR. Do you happen to have the Meta Quest 3S Xbox Edition controllers?
> 
> The profiles list for these should report `meta-quest-touch-plus` as their first fallback. So even without any changes, I'd expect them to show up in A-Frame just fine (without the Xbox branded models). While having the profile explicitly supported is nice, it should not point towards non-existing model files in the `modelUrl`.
> 
> The models can be found in the `webxr-input-profiles` repository: https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/assets/profiles/meta-quest-touch-plus-v2 You can create a PR on [aframevr/assets](https://github.com/aframevr/assets) to add them.

Thanks for the feedback! I don’t have the Meta Quest 3S Xbox Edition controllers, so I couldn’t test with the actual hardware. I have added explicit support for the meta-quest-touch-plus-v2 profile in the component without including the model files. The controller should still fall back to meta-quest-touch-plus and work in A-Frame. I have also created a separate PR on aframevr/assets to add the corresponding models from the webxr-input-profiles repository.

PR #5756, opened 2025-08-28
Logitech MX Ink improvements

FilipeLopesPires:
**Description:** Improvements to A-Frame's integration of Logitech's MX Ink Stylus

**Changes proposed:**
- Allow stylus visualization when in AR mode (i.e. Meta Quest passthrough / mixed reality)
- Support brush stroke size variation based on controller button / tip pressure applied

### DIFF
--- a/examples/showcase/painter/components/brush.js
+++ b/examples/showcase/painter/components/brush.js
@@ -2,7 +2,8 @@
 AFRAME.registerComponent('brush', {
   schema: {
     color: {type: 'color', default: '#ef2d5e'},
-    size: {default: 0.01, min: 0.001, max: 0.3},
+    size: {default: 0.01, min: 0.001, max: 1.0},
+    pressureMultiplier: {default: 5.0, min: 0.01, max: 10.0},
     enabled: {default: true},
     hand: {default: 'left'}
   },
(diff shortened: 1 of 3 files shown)


FilipeLopesPires:
@dmarcos would you be the best PR reviewer for this?

examples/showcase/painter/components/brush.js
```diff
@@ -68,6 +73,16 @@ AFRAME.registerComponent('brush', {
     this.painting = false;
   },
 
+  onButtonChanged: function (evt) {
+    if (!this.data.enabled) { return; }
+    if (!this.painting) { return; }
+    if (evt.detail.state.value === 1) {
+      this.data.size = this.originalSize;
```
dmarcos, line 80: don't modify the `this.data.size` It will be overridden if the component property changes. Have a `this.sizeFactor` variable that is applied below
```this.stroke.setSize(this.data.size * this.sizeFactor)```

dmarcos:
@FilipeLopesPires thanks for the patience. It's close. I left a comment. Thanks so much!

examples/showcase/painter/components/brush.js
FilipeLopesPires: Nice catch. SizeFactor added, take a look when you get a chance. And thanks for the review!

dmarcos:
Thank you!

Issue #5570, opened 2024-09-05
Bad orientation of hand models in hand-controls component.  


AryehMischel:
**Hand models used for the hand-controls component do not line up properly with hands on quest 2 (and presumably all devices)**

- A-Frame Version: 1.42 ++
- Quest 2:
- I believe the issue is because of Fixes https://github.com/aframevr/aframe/issues/4848 when the component was switched from targetRaySpace to gripSpace and the models were never updated to reflect this change.
- I made a glitch showing this issue.
- https://glitch.com/edit/#!/aframe-hand-models-test  (for context the default hands are the ones currently used in A-Frame, and then the ones with the offset are my own models that I edited to try and correct the rotation...

dmarcos:
each controller has different orientation for the real hands and it's difficult to have the proper orientation for all. `hand-controls` is an old component I no longer recommend. Instead, I recommend switching to regular controller models like `oculus-touch-controls` or `hand-tracking-controls` if you want a hand representation. This way virtual and physical controllers will always match. Haptics and proprioception will be accurate.

AryehMischel:
I get that completely, I think there are two important points to make. 
**1).** If we are not recommending people use the hand-controls then it really should be removed from the [aframe school](https://aframe.io/aframe-school/#/12). As of right now, I believe hand-controls are the most popular way that people are implementing VR hands. I say this because there are youtube videos with thousands of views, and it's what people historically used with the super-hands component. I have not been able to test the hand-controls component with other devices, but I wonder if the current orientation of the model is accurate with any VR Controller. From a little experimenting, I found that the orientation of the hands are fine in Quest Two if we set the handModelOrientationX to 0. If we are saying that the hands cannot accommodate all controllers I think it makes sense to prioritize the most popular ones and simply remove this offset. 
**2).** The problem with using either oculus-touch-controls or hand-tracking as a replacement for hand-controls  is that far as I can see there is no out-of-the-box way to represent the user's controllers as animated hand models (which is a pretty standard thing in vr). Please correct me if I am wrong! I know it isn't rocket science for someone to take a hand model, create some poses, and animate between them as the user interacts with the controller. However, it does seem like a large barrier for most people who might be trying to create simple VR interactions. I hope this makes sense. I am sorry if this comes off as aggressive. I super appreciate this awesome ecosystem you and several others have built and continue to maintain/grow. 
![image](https://github.com/user-attachments/assets/1d578169-e7f8-4c73-9898-8ae344ba28f8)

[...]

Issue #5321, opened 2023-06-25
Hand-controls (and with laser-controls) misaligned with Meta Quest 2 controllers in 1.4.2


smeybi:
**Description:**
_hand-controls_
The hands are misaligned at the Meta Quest 2 controllers
`<a-entity id="rightHand" hand-controls="hand: right;"></a-entity>`

The hand controllers are 45 degrees too far on the x axis. They should be -45 degrees, .rotation.x=-0.785398

_laser-controls_
The lasers with hands are misaligned at the Meta Quest 2 controllers
`<a-entity laser-controls="hand: right"></a-entity>`
The laser controllers should be maybe -48.6727 degrees on the x axis, .rotation.x=-0.85

_raycaster_
The hand object blocks the raycaster. It is missing near: 0.1 or similar at default, I think it is default near: 0.
Workaround: `<a-entity laser-controls="hand: right" raycaster="near: 0.1"></a-entity>`

- A-Frame Version: 1.4.2 
- Platform / Device: Meta Quest 2

I do not know exactly the rotations, I hope it helps to fix!

vincentfretin:
PR #5766 fixes the issue.

Issue #5305, opened 2023-06-11
hand-controls misaligned with HTC Vive controllers in 1.4.2


wyzwon:
**Description:**
As of commit 5b29ac2, the hands no longer line up with the HTC Vive controllers.

This is the log for the specific commit in which the issue first appears:

> hand-controls: rotates models to match WebXR definition (fix #4848) (#5261)

It is a small commit that only changes handModelOrientation. It seems the Vive is not in agreement with the new WebXR standard. Is this an issue that exists at the level of SteamVR and their WebXR implementation? If so, maybe a workaround is possible in the mean time?

Note: Due to an issue with the build bot at the time, you won't see the issue in the cdn.jsdelivr.net link supplied for commit 5b29ac2 or any other commit before 7050d7b (the last April 19th commit). You'll have to remake the build yourself if you want to try those commits for any reason.

- A-Frame Version: 1.4.2 (5b29ac2 and up)
- Platform / Device: gen 1 HTC Vive / Windows 10 / Steam VR 1.26.3 (beta) / Chrome 114

mrxz:
I was able to reproduce the issue with Valve Index controllers. The problem is entirely in A-Frame due to inconsistencies in how it handles various controls. When using `hand-controls`, there is a chain of dependencies as follows:

- `hands-controls`
- `oculus-touch-controls`, `vive-controls`, etc... (annoyingly enough it's missing `valve-index-controls`)
- `tracked-controls`
- `tracked-controls-webxr` (or in case of WebVR `tracked-controls-webvr`)

The `hands-controls` assumes that the object is in _gripSpace_ in which case the rotation added in commit https://github.com/aframevr/aframe/commit/5b29ac2a24abec8209624c65e3fd335e9fcfecef is correct. However, the default space used by `tracked-controls` is _targetRaySpace_. Literally only `oculus-touch-controls` deviates from the default and configures `tracked-controls` to be _gripSpace_, all other controls components result in _targetRaySpace_, meaning the entire pose of the hand (position and rotation) is mostly bogus.

As a workaround you can manually override the space for `tracked-controls`:
```HTML
<a-entity hand-controls="hand: left" tracked-controls="space: gripSpace"></a-entity>
```

wyzwon:
Using **tracked-controls="space: gripSpace** seems to put their center of mass back in line with the physical controllers but the hands are still pitched 90 degrees forward relative to their old alignment.

mrxz:
[...]

PR #5766, opened 2025-12-07
feat(hand-controls): fix hand model rotation alignment

filandra:
**Description:**
Default rotation and position is aligned with Meta Quest Controllers.
The adjustment for Pico4 has been removed.
It set the handModelOrientationX to -45 degrees, which now is the default value. As a result this adjustment is not needed anymore.

### DIFF
--- a/src/components/hand-controls.js
+++ b/src/components/hand-controls.js
@@ -123,17 +123,9 @@ export var Component = registerComponent('hand-controls', {
     var handModelOrientationZ = hand === 'left' ? Math.PI / 2 : -Math.PI / 2;
     // The WebXR standard defines the grip space such that a cylinder held in a closed hand points
     // along the Z axis. The models currently have such a cylinder point along the X-Axis.
-    var handModelOrientationX = el.sceneEl.hasWebXR ? -Math.PI / 2 : 0;
+    var handModelOrientationX = el.sceneEl.hasWebXR ? -Math.PI / 4 : 0;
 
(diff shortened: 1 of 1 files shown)


filandra:
This change is addressing issues: #5570, #5305, #5321

dmarcos:
Thanks. If any rotation is needed the rotation should be handled internally or `hand-controls` is not working properly.

filandra:
@dmarcos sorry for the delay - the holidays were very busy hahah. What do you mean by internally? Should it be handled in some other module? I figured that I would place the rotation handling into the same place where it was being already modified previously.

vincentfretin:
@filandra This means removing offSetRotation offSetPosition properties you added, put back the offset directly in radians inline in the code, so Math.PI / 4 for everything.

filandra:
@vincentfretin Well the whole point of the PR is to add the ability to set custom rotation and position. If not here, where should this logic be added?

vincentfretin:
The rotation is wrong on all Quest headsets right? After you fix the rotation inline in the component, why would you need to customize it?

dmarcos:
Yeah we should just fix the rotation. No need of a public API

filandra:
@dmarcos @vincentfretin I believe now the implementation should be correct?

vincentfretin:
Yes, that's correct, thanks. I tested on the showcase/tracked-controls/index.html example, with lowPoly, highPoly and toon hands on my Quest 3, hands position are correct with that change. @dmarcos good to merge, you can also close the linked issues.

PR #5769, opened 2025-12-13
Fix crash setting transform on undefined when layer is pending recreation

vincentfretin:
**Description:**

Fix crash setting transform on undefined when layer is destroyed in updateSrc and layer is not yet recreated because image has not loaded yet (this.texture is undefined).
Crash reproduced on the Comic book example while in VR switch to the next image. I didn't have the issue on the other layer quad example.
In https://github.com/aframevr/aframe/commit/a59273715b16918f132d7fcaf6fa92c7036270b3 I had changed updateSrc to  destroy the layer on src change, not sure if I tested again that example after that change or the issue appears now because of my connection, anyway this PR fixes it. 

**Changes proposed:**
- if this.layer is undefined after initLayer, return


dmarcos:
thank you!

PR #5770, opened 2025-12-13
Remove useless lines in tracked-controls component

vincentfretin:
**Description:**

Remove useless lines in tracked-controls component, those should have been removed when tracked-controls-webxr was renamed to tracked-controls in https://github.com/aframevr/aframe/commit/627698d4bd00e055549c0b91573a66f721d8dfef


dmarcos:
Thanks!

PR #5774, opened 2025-12-13
Fix all controls to use id instead of idPrefix on tracked-controls

### LINKED ISSUE #5772: Unknown property `idPrefix` for component `tracked-controls`
Running the tests, I got the message
Unknown property `idPrefix` for component `tracked-controls`
logitech-mx-ink-controls and meta-oculus-controls are using id correctly and it's used as idPrefix in checkControllerPresentAndSetup, but all the other controls use idPrefix when setting the tracked-controls component, not sure when this is has been modified, I didn't found the commit where it changed. I'm guessing we just need to change idPrefix by id for all the other components listed here:

```
$ ag idPrefix src/
src/utils/tracked-controls.js
10: * @param {object} idPrefix - Prefix to match in gamepad id if any.
13:export function checkControllerPresentAndSetup (component, idPrefix, queryObject) {
19:  controller = isControllerPresent(component, idPrefix, queryObject);
57:export function findMatchingControllerWebXR (controllers, idPrefix, handedness, index, iterateProfiles, handTracking) {
72:          controllerMatch = profiles[j].startsWith(idPrefix);
76:        controllerMatch = profiles.length > 0 && profiles[0].startsWith(idPrefix);

src/components/hp-mixed-reality-controls.js
120:      idPrefix: GAMEPAD_ID,

src/components/generic-tracked-controller-controls.js
130:      idPrefix: GAMEPAD_ID_PREFIX,

src/components/vive-focus-controls.js
102:      idPrefix: GAMEPAD_ID_PREFIX

src/components/valve-index-controls.js
121:      idPrefix: GAMEPAD_ID_PREFIX,

src/components/pico-controls.js
113:      idPrefix: GAMEPAD_ID,

src/components/oculus-go-controls.js
106:      idPrefix: GAMEPAD_ID_PREFIX

src/components/windows-motion-controls.js
176:      idPrefix: GAMEPAD_ID_PREFIX,

src/components/vive-controls.js
124:      idPrefix: GAMEPAD_ID_PREFIX,

src/components/magicleap-controls.js
115:      idPrefix: GAMEPAD_ID_COMPOSITE,

src/components/tracked-controls.js
72:   * Handle update controller match criteria (such as `id`, `idPrefix`, `hand`, `controller`)
```

vincentfretin:
**Description:**

Fix all controls to use id instead of idPrefix when setting the tracked-controls component, only logitech-mx-ink-controls and meta-touch-controls were correct (fix #5772)

The tests pass with three 177

### DIFF
--- a/src/components/generic-tracked-controller-controls.js
+++ b/src/components/generic-tracked-controller-controls.js
@@ -127,7 +127,7 @@ export var Component = registerComponent('generic-tracked-controller-controls', 
     }
     el.setAttribute('tracked-controls', {
       hand: data.hand,
-      idPrefix: GAMEPAD_ID_PREFIX,
+      id: GAMEPAD_ID_PREFIX,
       iterateControllerProfiles: true
     });
     if (!this.data.defaultModel) { return; }
--- a/src/components/hp-mixed-reality-controls.js
+++ b/src/components/hp-mixed-reality-controls.js
@@ -117,7 +117,7 @@ export var Component = registerComponent('hp-mixed-reality-controls', {
 
     el.setAttribute('tracked-controls', {
       // TODO: verify expected behavior between reserved prefixes.
-      idPrefix: GAMEPAD_ID,
+      id: GAMEPAD_ID,
       hand: data.hand,
       controller: this.controllerIndex
     });
--- a/src/components/magicleap-controls.js
+++ b/src/components/magicleap-controls.js
@@ -112,7 +112,7 @@ export var Component = registerComponent('magicleap-controls', {
 
     el.setAttribute('tracked-controls', {
       // TODO: verify expected behavior between reserved prefixes.
-      idPrefix: GAMEPAD_ID_COMPOSITE,
+      id: GAMEPAD_ID_COMPOSITE,
       hand: data.hand,
       controller: this.controllerIndex
     });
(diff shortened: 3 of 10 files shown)


vincentfretin:
That would be good if someone having one of those headset could verify this is working properly.
Does the controllers worked at all on the aframe 1.7.1 release? And what about now with the PR?

dmarcos:
thanks!

mrxz:
> That would be good if someone having one of those headset could verify this is working properly.
> Does the controllers worked at all on the aframe 1.7.1 release? And what about now with the PR?

Haven't run any tests, but I do recall testing it when removing the `webvr` variant and not running into any issues. I believe it still works due to the following pattern in the components:

- `checkIfControllerPresent` filters correctly on the `GAMEPAD_ID`
- If a matching controller is found the `injectTrackedControls` method is called
- The invalid `idPrefix` property is set, meaning `id` ends up being the default (empty string)
- [`tracked-controls` uses  `controllerUtils.findMatchingControllerWebXR` ](https://github.com/aframevr/aframe/blob/master/src/components/tracked-controls.js#L75), which now matches any controller.

In short, there are two checks/filters that use a different ID value. The former still guarantees the right `xyz-controls` becomes active, and the latter matches any and all controller (of the expected handedness). So this would've only caused issues in case there's more than one controller with the same handedness, which is very uncommon.

Looking at the overall flow, it would be nicer if there's only one lookup/check, completely avoiding the possibility of the two mismatching. But at least this fixes the potential issue for now.

Issue #5757, opened 2025-09-01
Hand / Controller models disapear which switching modes.


KooIaIa:
This is happening in my site and another unrelated persons site so it appears to be a bug in latest A-frame.

Example: https://strata-design-demo.vercel.app/examples/1/view

Controls look like this when its working and you switch from hands to controllers:
<img width="800" height="400" alt="Working" src="https://github.com/user-attachments/assets/2bef24ec-0f1f-4eab-a1c1-7d1fd257ce09" />

The bug is that the hands or controllers disappear frequently when doing this switching:

<img width="1052" height="592" alt="Image" src="https://github.com/user-attachments/assets/d557a21b-6f6a-40c0-8ffd-bf319bafa7f9" />

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/31d742ff-3e27-4a26-b924-52f50f2506dd" />

Notice how the laser attached to the hand is still there and tracking is happening. It appears to just be the visual logic hiding the hands or controllers that is breaking.

The problem resolves itself on Quest when going in and out of the Oculus menu.

electrum-bowie:
Up 😆

vincentfretin:
Your site is down. Can you please create a simple test case with https://krabbel.dev so we see what components you're using and please tell us exactly what was the aframe version and components that was used when you saw that issue.
I'm also having sometimes that issue with [handy-controls](https://github.com/AdaRoseCannon/handy-work) and just hiding my hand behind my back and putting it back in front fixes the issue generally. I'm not sure if the issue is in the handy-controls component, in XRControllerModelFactory or threejs or a Meta browser issue with the disconnected/connected events, I don't currently have time to spend hours debugging that, that's really low on my priority list.
My guess is that's probably a race condition with a quick succession of controller disconnected, hand connected, hand disconnected controller connected that toggles the visibility of the models.

KooIaIa:
https://krabbel.dev/p/pretty-cougars-smash

I just did this minimal code on the default Krabbel sample.

      <a-entity id="rightController" oculus-touch-controls="hand: right"></a-entity>
      <a-entity id="leftController" oculus-touch-controls="hand: left"></a-entity>
      <a-entity id="rightHand" hand-tracking-controls="hand: right; modelStyle: mesh"></a-entity>
      <a-entity id="leftHand" hand-tracking-controls="hand: left; modelStyle: mesh"></a-entity>
 
I added cubes so you can see them when they turn invisible.

[...]

PR #5777, opened 2025-12-15
Fix controller model invisible sometimes when toggling between hands and controllers

vincentfretin:
Fix controller model invisible sometimes when toggling between hands and controllers but not the two at the exact same time.
See explanation in #5757


dmarcos:
thanks so much!

PR #5778, opened 2025-12-16
Set the image cache when img is loaded

vincentfretin:
Set the image cache when img is loaded to avoid a potential non resolving issue if the same img url is loaded via ImageLoader, as discussed in https://github.com/aframevr/aframe/pull/5776#discussion_r2620143211

Also use THREE.Cache methods instead of accessing directly THREE.Cache.files in tests.


dmarcos:
thanks!

PR #5779, opened 2025-12-16
Show loading screen while downloading images assets

vincentfretin:
**Description:**

Set timeout mechanism and loading screen earlier at interactive state so we see the loading screen while images are loading.
That was a regression since https://github.com/aframevr/aframe/commit/0d9721884feeb59825dc8c0bebfe73efea34e670 where we changed everything to wait on complete state that includes DOM ready and resources like images loaded.

**Changes proposed:**
- Implements connectedCallback for a-assets to wait for interactive and similar in a-scene to set the renderer and loading screen earlier.

PR done with the help of Claude Code with Opus 4.5

### DIFF
--- a/src/core/a-assets.js
+++ b/src/core/a-assets.js
@@ -17,6 +17,24 @@ class AAssets extends ANode {
     this.timeout = null;
   }
 
+  /**
+   * Override connectedCallback to initialize at 'interactive' instead of 'complete'.
+   * This allows the timeout mechanism to work before loading the images.
+   * If we wait for 'complete', all resources (including images) are already loaded.
+   */
+  connectedCallback () {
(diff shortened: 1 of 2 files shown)


vincentfretin:
You can test it on the 360 image example examples/boilerplate/panorama/index.html with Chrome network tab set to Slow 4G for example.

vincentfretin:
Note that if the timeout is hit, the loading screen will still show until all images are loaded (triggering the readyState complete), then the loading screen hide and we can still have some a-asset-item elements downloading.
It doesn't make sense to hide the loading screen before readyState complete anyway, the page will be just white, because all entities and components are loaded only in readyState complete.

PR #5782, opened 2025-12-17
Fix cursor component triggering clicks on wrong controller's hovered entity in VR

vincentfretin:
**Description:**

You can reproduce the issue on https://aframe.io/aframe/examples/showcase/ui/ that is using the laser-controls component.

When using two controllers in VR with laser-controls, the cursor component listens to WebXR session events (`selectstart`, `selectend`) which fire globally for any input source. Without filtering by handedness:

1. Left controller hovers over an image
2. Right controller hovers over the floor
3. User presses trigger on right controller
4. Click event fires on the image (wrong!) because left cursor's `intersectedEl` was used

**Changes proposed:**
- Added `hand` schema property to cursor component
- Filter `selectstart` events in `onCursorDown` by comparing `evt.inputSource.handedness` with configured `hand`
- Filter `selectend` events in `onCursorUp` similarly
- laser-controls now passes its `hand` property to the cursor component

PR done with the help of Claude Code with Opus 4.5


dmarcos:
Thank you

PR #5798, opened 2026-03-29
Remove pcfsoft shadow type, deprecated in r182, pcf is now soft as well

vincentfretin:
**Description:**

Remove pcfsoft shadow type, deprecated in r182, pcf is now soft as well.
You should wait for three to be updated to r182+ before merging this.

**Changes proposed:**
- Remove pcfsoft from the available shadow types
- Change pcfsoft to pcf in the docs and examples
- Add migration guide in changelog


mrxz:
Perhaps we can add a generic fallback in `shadow#init` for when the given shadow type isn't known. That should more gracefully handle any mistyped/unknown types and has the added benefit of not breaking when users upgrade, while a warning in the console can still clue them in to change/fix it.

vincentfretin:
Right, the warning on threejs side `THREE.WebGLShadowMap: PCFSoftShadowMap has been deprecated. Using PCFShadowMap instead.`.
I propose `shadow type "pcfsoft" is not supported, falling back to "pcf". To remove this warning set <a-scene shadow="type: pcf">.`
I did the change.

PR #5799, opened 2026-03-30
Remove renderer colorManagement: true in examples, that's the default

vincentfretin:
Remove renderer colorManagement: true in examples, that's the default.
This is to avoid unnecessary copy and paste from users looking at the examples code.


dmarcos:
Thanks!

PR #5800, opened 2026-03-30
Replace Clock instance by Timer instance (three r183+)

vincentfretin:
Replace Clock instance by Timer instance. Clock is deprecated in three r183.
To be merged after the repo is updated to three r183+

I didn't call `self.timer.connect(document);` 
The `connect(document)` call on the timer leverages the Page Visibility API similar to Clock to prevent large time jumps when the browser tab is inactive.
I don't think we want it?

FYI networked-aframe accessed that clock instance so I did some changes in https://github.com/networked-aframe/networked-aframe/pull/521 That's the only occurrence of it I saw in community components.

### DIFF
--- a/src/core/scene/a-scene.js
+++ b/src/core/scene/a-scene.js
@@ -42,7 +42,7 @@ export class AScene extends AEntity {
     var self;
     super();
     self = this;
-    self.clock = new THREE.Clock();
+    self.timer = new THREE.Timer();
     self.isIOS = isIOS;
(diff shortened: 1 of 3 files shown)


vincentfretin:
Actually I think it's good to have the `self.timer.connect(document)`, it fixes the issue with orbit-controls="autoRotate: true" that speed up the rotation when you come back to the tab. I amended the commit.

PR #5804, opened 2026-04-04
Fix tracked-controls.controller undefined on controllerconnected after reconnect

vincentfretin:
When a controller disconnects and reconnects, `el.components['tracked-controls'].controller` is `undefined` at the time the `controllerconnected` event fires, producing an error can't access gamepad on undefined on the haptics component at https://github.com/supermedium/superframe/blob/7439b2f77ade516077e8edfc41991ee0fc9d07a9/components/haptics/index.js#L31 so losing the vibrations when you put both controllers on the ground and picking them up again.

**Root cause:** The specific controller component (e.g. meta-touch-controls) registers its `controllersupdated` listener before tracked-controls' listener (because tracked-controls is created later during `injectTrackedControls`). On reconnect:

1. The specific controller's `controllersupdated` handler fires first
2. It calls `checkControllerPresentAndSetup` → `injectTrackedControls` → `setAttribute('tracked-controls', ...)` with unchanged data
3. A-Frame skips `update()`/`updateController()` because the data hasn't changed
4. `controllerconnected` is emitted with `tracked-controls.controller` still `undefined`
5. tracked-controls' own `controllersupdated` listener fires after, but too late

**Fix:** In `checkControllerPresentAndSetup`, explicitly call `trackedControls.updateController()` before emitting `controllerconnected`. The check is skipped on first connect (tracked-controls doesn't exist yet) to avoid a redundant call.

### DIFF
--- a/src/utils/tracked-controls.js
+++ b/src/utils/tracked-controls.js
@@ -27,7 +27,15 @@ export function checkControllerPresentAndSetup (component, idPrefix, queryObject
   // Update controller presence.
   if (isPresent) {
     component.addEventListeners();
+    // On reconnect, injectTrackedControls calls setAttribute('tracked-controls', ...)
+    // with unchanged data so update/updateController is skipped.
+    // tracked-controls' own controllersupdated listener would call updateController,
+    // but it was registered after the specific controller component's listener
+    // (e.g. meta-touch-controls) so it fires too late.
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

dmarcos:
Thank you!

PR #5810, opened 2026-04-11
Expose premultipliedAlpha on material component

vincentfretin:
## Summary

Since three.js r178 ([mrdoob/three.js#31246](https://github.com/mrdoob/three.js/pull/31246)), `THREE.MultiplyBlending` emits a repeated console warning unless the material has `premultipliedAlpha = true`:

```
THREE.WebGLState: MultiplyBlending requires material.premultipliedAlpha = true
```

A-Frame's `material` component exposes `blending: multiply` but does not expose `premultipliedAlpha`, so users had no way to silence the warning from markup alone.

This PR:

- Adds `premultipliedAlpha` to the `material` schema (default `false`), giving users direct control from markup.
- Forces `premultipliedAlpha = true` whenever `blending` is `multiply`, regardless of the user-supplied value, so the three.js warning cannot accidentally be triggered through the component's public API.

## Changes

- `src/components/material.js` — add `premultipliedAlpha` schema property; force it on for `blending: multiply`, otherwise pass through the user value.
- `docs/components/material.md` — document the new property.
- `tests/components/material.test.js` — four tests in the `blending` suite covering:
  - forced on when blending is multiply
  - forced on even when user sets it to false with multiply
  - defaults to false for non-multiply blending
  - respects user value for non-multiply blending

## Test plan

- [x] `npm test` — all tests pass.
- [x] Manually verified the warning no longer appears on an entity using `material="blending: multiply; transparent: true"`.

### DIFF
--- a/docs/components/material.md
+++ b/docs/components/material.md
@@ -64,6 +64,7 @@ depending on the material type applied.
 | flatShading  | Use `THREE.FlatShading` rather than `THREE.StandardShading`.                                                                                      | false         |
 | offset       | Texture offset to be used.                                                                                                                        | {x: 0, y: 0}  |
 | opacity      | Extent of transparency. If the `transparent` property is not `true`, then the material will remain opaque and `opacity` will only affect color.   | 1.0           |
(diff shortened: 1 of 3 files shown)


dmarcos:
Thank you!

PR #5811, opened 2026-04-11
Fix font url to Roboto-msdf.json in msdf example

vincentfretin:
The rawgit url gave a 404, I replaced it by a jsdelivr url. Set a camera position farther away to see all the msdf and sdf texts.


dmarcos:
Thank you

Issue #5028, opened 2022-03-21
Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script


samiemad:
**Description:**
When loading the library, the browser throws an error:
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source 
of script in the following Content Security Policy directive: "script-src 'self' [...]"
```

Looking into the stack trace, I see these lines from https://github.com/aframevr/aframe/blob/master/dist/aframe-master.js
```js
//getters for the private vars
;['width', 'height', 
  'descender', 'ascender',
  'xHeight', 'baseline',
  'capHeight',
  'lineHeight' ].forEach(addGetter)

function addGetter(name) {
  Object.defineProperty(TextLayout.prototype, name, {
    get: wrapper(name),
    configurable: true
  })
}

//create lookups for private vars
function wrapper(name) {
  return (new Function([
    'return function '+name+'() {',
    '  return this._'+name,
    '}'
  ].join('\n')))()
}
```

Looks like aframe requires unsafe-eval to be allowed which is quite dangerous. Is there a workaround to use aframe without enabling `unsafe-eval`??

Can this be replaced by something safer? like:
```js
    get: function () {
      return this['_'+name]
    },
```

- A-Frame Version: 1.3.0
- Platform / Device: PC / Chrome
- Reproducible Code Snippet or URL:

Thanks!

dmarcos:
Can you provide a link to an example to reproduce?

An example loading the library just fine: https://glitch.com/~aframe

dmarcos:
Do you need that meta tag? FWIW the `unsafe-eval` code you pointed comes from a dependency bet seems safe. Just adding getters automatically to the object.

samiemad:
Yes, I agree it seems safe. But this setting is global. So if we allow it, it will be allowed for the whole web app and makes the whole app security lower. After all it's called `unsafe` for a reason.

Anyway, if you help me track that dependency I would be happy to try to fix the issue and create a PR.

dmarcos:
Code is in https://github.com/Jam3/layout-bmfont-text/blob/master/index.js#L221

used by https://github.com/Jam3/three-bmfont-text

dmarcos:
The quickest option is probably make and serve your own A-Frame build with the code corrected / removed.

samiemad:
Thanks. I'll have a look and see what I can do.

[...]

PR #5821, opened 2026-05-30
Update text component to fix unsafe-eval error with a strict CSP

vincentfretin:
Update three-bmfont-text version to include https://github.com/dmarcos/three-bmfont-text/pull/5 to avoid using "new Function()" that triggers unsafe-eval error with a strict CSP (Content Security Policy), see #5028 for context.
Add a unsafe-eval error non regression test that runs in CI with Chrome and Firefox.


vincentfretin:
Test is failing as expected until the other PR is merged and this PR is rebased
```
SUMMARY:
✔ 1163 tests completed
ℹ 17 tests skipped
✖ 1 test failed

FAILED TESTS:
  text component CSP
    ✖ loading the A-Frame bundle does not require unsafe-eval
      Chrome 148.0.0.0 (Linux x86_64)
    Error: Loading the A-Frame bundle triggered a CSP unsafe-eval violation: script-src eval. The text component dependency three-bmfont-text must not use eval() / new Function().
        at check (build/commons.js:250863:21)

```

mrxz:
Small nitpick: the test explicitly mentions the `text` component (in its name and failure messages), but the tested condition that there are no CSP violation when loading the A-Frame bundle is generic. Ultimately it doesn't matter where/how an `eval`/`new Function` gets pulled in, this test will catch that regression.

vincentfretin:
You're absolutely right.
I moved tests/components/text-csp.test.js to tests/csp/no-unsafe-eval.test.js and the error now says:
Error: Loading the A-Frame bundle triggered a CSP unsafe-eval violation: script-src eval. The bundle (or one of its dependencies) must not use eval() / new Function()

vincentfretin:
Locally the test is green, but in CI it's failing.
I reproduce it locally with
`TEST_ENV=ci npm run test:chrome`

The in-memory `commons.js` is istanbul-instrumented under `TEST_ENV=ci`, and istanbul injects a `new Function("return this")()` (its global-object lookup) into every instrumented src/ module.

I'll write the test in another way.

vincentfretin:
The test pass on Chrome and Firefox. I updated the PR title and description.

dmarcos:
Thank you!

Issue #5396, opened 2023-11-27
Consider an A-Frame API to expose THREE.BatchedMesh 


dmarcos:
Need to still study how it actually works. Anyone already familiar with it more than welcome to propose an API. Thanks!

Context:

https://github.com/mrdoob/three.js/issues/22376

vincentfretin:
The API is quite similar to THREE.InstancedMesh, see https://github.com/diarmidmackenzie/instanced-mesh for inspiration.
I also created https://github.com/networked-aframe/networked-aframe/issues/485 in networked-aframe for investigation of using it for avatars.

vincentfretin:
I shared it last time on [discord](https://discord.com/channels/479784974917042186/479787777949433900/1442122700893065346), [iwsdk](https://iwsdk.dev/) has a [FlexBatchedMesh](https://github.com/facebook/immersive-web-sdk/blob/8d931089406ec7eb692bee0e7b759eece4d5c94a/packages/xr-input/src/visual/utils/flex-batched-mesh.ts) class that has a similar logic than the instanced-mesh aframe component, it's creating a `BatchedMesh` instance for each material group and call `setMatrixAt` with the calculated world matrices.
It's used on the controller models to reduce the draw calls just wrapping the model [here](https://github.com/facebook/immersive-web-sdk/blob/8d931089406ec7eb692bee0e7b759eece4d5c94a/packages/xr-input/src/visual/impl/animated-controller.ts#L99).
The `updateMatrixWorld` is patched to update the matrices in all the `BatchedMesh` instances and all the meshes have `matrixAutoUpdate = false`  set.

vincentfretin:
I have a private version of aframe-editor working with BatchedMesh in both view and editor mode, one BatchedMesh for each gltf-model sharing the same src. The main issue I had is that the cursor component doesn't emit the events because the intersected element is the BatchedMesh and doesn't change. This change could fix that
branch https://github.com/vincentfretin/aframe/tree/raycaster-batchId2  commit https://github.com/vincentfretin/aframe/commit/9da630215a05f8d81ac6404138804d2b648213bf so the emitted intersection contains batchId/instanceId

or a variant, raycaster could do a mapping batchId to entity, this is branch https://github.com/vincentfretin/aframe/tree/raycaster-batchId commit https://github.com/vincentfretin/aframe/commit/5c30a82892deb255a5fdb94ca04e33f9d048d734
[...]

PR #5824, opened 2026-06-09
Support per-instance cursor events on BatchedMesh / InstancedMesh

vincentfretin:
I open this for feedback. I just started using it in a project where I use a component to handle an hover/open/close animation on an instance of an InstancedMesh in a glb.

[Live example](https://raw.githack.com/vincentfretin/aframe/raycaster-batchId2-build/examples/test/raycaster-batched-mesh/index.html)

Multiple a-entities sharing a single BatchedMesh (draw-call reduction) or InstancedMesh (instancing) all map to the same host entity, so the raycaster and cursor collapse hover / click tracking across every logical instance: ray moves between two instances of the same shared mesh, nothing fires.

raycaster: also fire raycaster-closest-entity-changed when the closest hit's (entity, batchId, instanceId) tuple changes, even if the entity is unchanged. No per-instance mapping in core — the raycaster's role stops at surfacing batchId / instanceId on intersection details.

cursor: track (intersectedEl, intersection.batchId, intersection.instanceId) as the hover identity. Transitions within a shared mesh now fire mouseleave with the OLD tuple and mouseenter with the NEW tuple. The stored intersection is used for event payloads so leave events carry the outgoing instance's details, not the incoming one.

App-level mapping: the new examples/test/raycaster-batched-mesh example shows a `batch-proxy` component on each host entity that holds its own map from id to member entity and forwards mouseenter / mouseleave / click to the matching per-instance entity. Keeping the map in userspace lets apps encode whatever identity makes sense (batchId, instanceId, custom composite keys) without expanding the core component API.

Tests cover intersection details carrying batchId / instanceId, raycaster-closest-entity-changed firing on tuple-only change, and the cursor's tuple-aware mouseleave/mouseenter sequencing with the outgoing tuple in detail.intersection.

### DIFF
--- a/docs/components/raycaster.md
+++ b/docs/components/raycaster.md
@@ -204,6 +204,27 @@ A-Frame will call `.refreshObjects()` automatically when an entity is appended
 or detached from the scene, but it will not get called during normal DOM
 mutations (e.g., some entity changes its `class`).
 
+## Per-instance Intersections on BatchedMesh / InstancedMesh
+
+When multiple a-entities share a single `THREE.BatchedMesh` or
+`THREE.InstancedMesh` (for draw-call reduction), the raycaster keeps
+resolving every intersection to the single host entity that owns the
+shared mesh. To distinguish instances, check
+`intersection.batchId` / `intersection.instanceId` on the event detail.
+
+`raycaster-closest-entity-changed` fires whenever the closest
+`(entity, batchId, instanceId)` tuple changes, including transitions
+between two instances of the same shared mesh. The cursor component
+treats the tuple as its hover identity, so `mouseenter` and
+`mouseleave` fire on those transitions too — the event's
+`intersection` detail carries the per-instance ids.
+
+Apps keep their own map from `batchId` / `instanceId` to logical
+entities and can proxy the host's cursor events to the matching
+per-instance entity. See the
+[raycaster-batched-mesh example](https://github.com/aframevr/aframe/tree/master/examples/test/raycaster-batched-mesh)
+for a working setup.
+
 ## Customizing the Line
 
 If `showLine` is set to `true`, the raycaster will configure the line given the
--- a/examples/index.html
+++ b/examples/index.html
@@ -208,6 +208,7 @@
       <li><a href="test/raycaster/">Raycaster</a></li>
       <li><a href="test/raycaster/simple.html">Raycaster (Simple)</a></li>
       <li><a href="test/raycaster/sprite.html">Raycaster (Sprite)</a></li>
+      <li><a href="test/raycaster-batched-mesh/">Raycaster (BatchedMesh / InstancedMesh)</a></li>
       <li><a href="test/shaders/">Shaders</a></li>
       <li><a href="test/shadows/">Shadows</a></li>
       <li><a href="test/text/">Text</a></li>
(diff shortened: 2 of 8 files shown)


vincentfretin:
I fixed the tests I forgot to run yesterday when I fixed the uv issue and added a non regression test for the fresh uv coordinates on click. I created another branch with the dist so you can see the [example](https://raw.githack.com/vincentfretin/aframe/raycaster-batchId2-build/examples/test/raycaster-batched-mesh/index.html)

mrxz:
Conceptually all raycasting and cursor interactions happen on the _entity_ level, regardless of what the Three.js contents of that entity is. While exposing more details about the intersections so that the app-level logic can handle it is fine, the changes to the events don't seem correct to me. Though obviously both are needed for the example to work properly.

For one, `raycaster-closest-entity-changed` now fires even if the _entity_ hasn't changed, as a change in either `instanceId` and `batchId` is enough. Secondly I doubt that the tuple (entity, batchId, instanceId) is sufficient, as there can in theory be more than one `InstancedMesh` or `BatchedMesh` inside an entity. Say moving the cursor from instance 1 of one `InstancedMesh` to instance 1 of another `InstancedMesh` would then not registered.

My main concern is with users that simply load any arbitrary GLTF file. Depending on the file itself, an `InstancedMesh` might end up inside the entity. For no apparent reason to the user, this model behaves differently compared to others, with `mouseenter`/`mouseleave` events emitted without ever leaving the model, for example.

Perhaps the `raycaster` can be changed such that the entity level events remain unchanged, but in addition it emits "scenegraph level" events. Even without corresponding changes in `cursor` this might already be useful in case users are looking for raycasting against `InstancedMesh`/`BatchedMesh` or even just individual `THREE.Mesh` instances.

The `cursor` component could then be expanded to support two distinct modes, either the (current) entity based tracking or the "scenegraph" based on. (Obviously a better/clearer name than "scenegraph" would be needed). Though I'm not sure the complexity is really worth it. Depending on if/how #5396 gets implemented, it might be better to keep everything operate on the entity level, and have the `raycaster`/`cursor` know how to handle the A-Frame built-in way of doing `BatchedMesh`?

vincentfretin:
As always, you provide valid points! Thanks for taking the time to look into it.
Having several InstancedMesh inside a glb is indeed possible, I actually just got one. It would indeed trigger mouseenter/mouseleave events when I hover on it and unecessary triggers recalculating the hover box for that entity in my editor if I was still using mouseenter/mouseleave, that's indeed not what I want for that case. My current code is using raycaster-closest-entity-changed and checking if the entity actually changed to update the hover box, so the semantic of that event is lost here, I agree.
You're right about instanceId not being enough, it also needs the InstanceMesh uuid I guess, my current use case is to extract a named InstancedMesh from a loaded glb into its own entity to have clickable class on it, so I only tested raycast on a single InstancedMesh until now.
Having other events emitted could be a solution. Thanks for the feedback, I may revisit the PR later when I have more needs. Currently it does the job for my current raycast usage in my project where I don't fall in the edges cases discussed here.

vincentfretin:
My other use case is that I have a modified gltf-model component that postpone loading the model and if there are several times the same glb in the scene it loads the glb only once and creates several BatchedMesh (one for each material) from it and instances for the other occurrences. In that configuration I have an entity for each occurrences. That's where the example in the PR comes from initially, I have an entity for each occurences and using the position/rotation/scale to create the instance matrix.
If I included the BatchedMesh uuid in the tuple, I would have the issue that it triggers mouseleave/mouseenter events for the same entity because on different BatchedMesh. Because currently batchId is the same for all created BatchedMesh for a glb and I don't have the BatchedMesh uuid in the tuple, it doesn't have that issue as a side-effect.

PR #5826, opened 2026-06-10
Fix cdn build to use super-three 0.184.0

vincentfretin:
The a-frobot built the cdn build still with 0.181.0, we also got the error downloading super-three 0.184.0 in the tests https://github.com/aframevr/aframe/commit/ec6f6a8dd956a23dd930c073aab4617bebf4f3cb#commitcomment-188253469
Also the package-lock.json wasn't pushed with the correct super-three version so here it is, it should triggers a correct new build.

Separate note, we should switch to Github CI for the build #5813

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "load-bmfont": "^1.2.3",
         "stats-gl": "^3.6.0",
         "super-animejs": "^3.1.0",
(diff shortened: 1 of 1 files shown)


vincentfretin:
Ah the a-frobot actually doesn't check for package-lock.json to trigger a build. I added that to the new pending Github CI.
So we need another change to trigger the build.

PR #5828, opened 2026-06-13
Force premultipliedAlpha for subtractive blending too

vincentfretin:
three.js's WebGLState requires material.premultipliedAlpha = true for both SubtractiveBlending and MultiplyBlending. The previous fix (#5810) only handled multiply, so subtractive blending (e.g. in the showcase/ui example) still emitted "THREE.WebGLState: SubtractiveBlending requires material.premultipliedAlpha = true". Extend the force to subtractive.

I got the infinite error log on the https://localhost:8080/showcase/ui/ example. (`npm run start:https` in the repo)


dmarcos:
thank you!

PR #5829, opened 2026-06-13
Add stopOn and pauseOn event properties to sound component

vincentfretin:
Add `stopOn` and `pauseOn` properties to the sound component to stop or pause the sound on a user-defined event, mirroring the existing `on` property. Listeners are now only registered when the event name is non-empty, fixing the previous behavior where an empty-string listener was registered for `on`.


dmarcos:
thanks

PR #5831, opened 2026-06-13
Fix a-sky losing material.side 'back' because of previous object recycling

vincentfretin:
This fixes the issue I had with the environment component https://github.com/supermedium/aframe-environment-component/pull/89

Fix primitive default components clobbered by recycled component attrValue mergeComponentData merged the entity-defined value into a primitive's default/extra component data with `utils.extend` (`Object.assign`), which copies undefined-valued keys. A component's `attrValue` is a pooled object that keeps its keys (valued undefined) when recycled, so a reused `attrValue` like `{side: undefined, ..., src: 'sky.webp'}` would clobber the primitive defaults (e.g. a-sky's material side: back -> undefined -> material schema default 'front'). This made `<a-sky material="src: ...">` created via createElement + setAttribute intermittently lose side/shader/minFilter depending on object-pool state.

Only merge in defined properties so stale recycled keys no longer overwrite the extra/default data.

### DIFF
--- a/src/core/a-entity.js
+++ b/src/core/a-entity.js
@@ -830,7 +830,16 @@ function mergeComponentData (attrValue, extraData) {
 
   // Merge multi-property data.
   if (extraData.constructor === Object) {
-    return utils.extend(extraData, utils.styleParser.parse(attrValue || {}));
+    var parsedAttrValue = utils.styleParser.parse(attrValue || {});
+    // Only merge in defined properties. `attrValue` may be a component's pooled `attrValue`
+    // object that still carries stale keys from a previous use (recycled objects keep their
+    // keys with `undefined` values), and those must not clobber the extra/default data
+    // (e.g. an a-sky primitive's `material.side: back`).
+    for (var key in parsedAttrValue) {
+      if (parsedAttrValue[key] === undefined) { continue; }
+      extraData[key] = parsedAttrValue[key];
+    }
+    return extraData;
   }
 
   // Return data, precedence to the defined value.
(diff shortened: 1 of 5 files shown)


vincentfretin:
I finally understood what was in my project that triggered it.
I have
```html
        <a-entity
          geometry="primitive: sphere;
                    radius: 198;
                    segmentsWidth: 64;
                    segmentsHeight: 64;"
          material="color: #FFF; shader: flat; side: back; minFilter: linear"
          scale="-1 1 1"
          id="left-eye"
          visible="false"
        ></a-entity>

        <a-entity
          geometry="primitive: sphere;
                    radius: 198;
                    segmentsWidth: 64;
                    segmentsHeight: 64;"
          material="color: #FFF; shader: flat; side: back; minFilter: linear"
          scale="-1 1 1"
          id="right-eye"
          visible="false"
        ></a-entity>
```
then my close360 function is called that does 
```js
  const leftEyeSphere = document.querySelector("#left-eye");
  const rightEyeSphere = document.querySelector("#right-eye");
  if (leftEyeSphere && rightEyeSphere) {
    leftEyeSphere.object3D.visible = false;
    rightEyeSphere.object3D.visible = false;
    // dispose texture
    leftEyeSphere.removeAttribute("material");
    rightEyeSphere.removeAttribute("material");
  }
```

then my scene json loads that creates 
```
{
  "element": "a-sky",
  "components": {
    "geometry": "",
    "material": "src: sky.webp",
  }
}
```

so I reproduced it in the test.

dmarcos:
thank you!!!

Issue #5737, opened 2025-06-16
Export all glitches before 8th July


vincentfretin:
List of glitches used in documentation:
https://glitch.com/~aframe
https://glitch.com/edit/#!/aframe-school-js
https://glitch.com/edit/#!/aframe-registry
https://glitch.com/edit/#!/aframe-animated-lights
https://glitch.com/~aframe-aincraft
https://glitch.com/edit/#!/aframe-modify-model-material
https://glitch.com/~aframe-gallery
https://glitch.com/~aframe-vaporwave
https://glitch.com/~networked-aframe -> https://stackblitz.com/github/networked-aframe/naf-project
https://aframe-360-gallery.glitch.me
https://glitch.com/edit/#!/aframe-basic-interaction
https://glitch.com/edit/#!/wide-cream
https://glitch.com/~aframe-school-cursor-handler/
https://glitch.com/~aframe-event-set-component
https://glitch.com/~aframe-school-cursor-handler/
https://glitch.com/~aframe-refraction-system
https://glitch.com/edit/#!/aframe-animation-events
https://glitch.com/~aframe-shooting-stars
https://glitch.com/edit/#!/aframe-displacement-shader
https://glitch.com/edit/#!/aframe-simpler-shader
https://glitch.com/edit/#!/aframe-displacement-registershader
https://glitch.com/edit/#!/aframe-displacement-offset-registershader
https://glitch.com/edit/#!/aframe-simpler-shader
https://glitch.com/edit/#!/aframe-texture-shader
https://glitch.com/edit/#!/aframe-modify-model-material
https://glitch.com/edit/#!/ex-2-geometry
https://glitch.com/edit/#!/ex-2a-texture-on-shape
https://glitch.com/edit/#!/ex-2b-shape-as-entity
https://glitch.com/edit/#!/ex-2c-animating-shapes
https://glitch.com/edit/#!/ex-2d-animating-shapes-with-components
https://glitch.com/edit/#!/a-frame-mouse-click-example
https://glitch.com/~aframe-text-fonts-demo
https://glitch.com/edit/#!/aframe-adding-a-sky
https://glitch.com/edit/#!/aframe-changing-the-sky
https://glitch.com/edit/#!/aframe-360-video-example
https://glitch.com/~aframe-basic-guide-with-environment

All those can be exported in a single repo aframevr/glitch-examples for example and we can update the urls in the doc to use:
[...]

PR #5832, opened 2026-06-14
Update installation.md

Utopiah:
removed Glitch

**Description:**
sadly closed for a while now

**Changes proposed:**
- just left codepen

### DIFF
--- a/docs/introduction/installation.md
+++ b/docs/introduction/installation.md
@@ -16,33 +16,7 @@ is primarily HTML and JavaScript.
 
 The fastest way is to start playing from within the browser.
 
-### Remix on Glitch
-
-![Glitch](https://cloud.githubusercontent.com/assets/674727/24480466/54b17d22-1499-11e7-8a18-d4f76b49ad07.jpg)
-
-[glitch]: https://glitch.com/~aframe
-
-[Glitch][glitch] provides an online code editor with instant deployment and
-hosting of web sites. The editor supports both front-end and back-end code as
-well as multiple files and directories. Glitch lets us remix (i.e., copy)
-existing projects and make them our own and instantly host and deploy changes
-for everyone to see.
-
-[Hit **Remix** on this A-Frame project][glitch], mess with the HTML in
-`index.html`, and see your site published live on each change! The base A-Frame
-Glitch, for example, is published at `aframe.glitch.me`, but we will provide your
-own custom URL name.
-
-Below are a few other A-Frame Glitches for starters:
-
-- [aincraft](https://github.com/aframevr/aframe/tree/master/examples/docs/aincraft) - Minecraft demo.
-- [360-gallery](https://github.com/aframevr/aframe/tree/master/examples/docs/360-gallery/) - 360&deg; image gallery.
-- [community-components](https://github.com/aframevr/aframe/tree/master/examples/docs/community-components) - Showcase of various community components.
-- [aframe-vaporwave](https://glitch.com/~aframe-vaporwave) - Retro-futuristic scene.
-
-### Other Code Editors
-
-Below are a couple of A-Frame starter kits on other browser-based code
+Below are a couple of A-Frame starter kits on browser-based code
 editors. Both support remixing or forking:
 
 - [CodePen &mdash; A-Frame](https://codepen.io/mozvr/pen/BjygdO)


vincentfretin:
I wanted to replace it with another solution to quickly create the examples from github repo, see my comment https://github.com/aframevr/aframe/issues/5737#issuecomment-2983294166
stackblitz is not working, but codesandbox is working, you need to log in with an account to edit though. If we update the urls used in the examples in the gh-pages branch, that should be seamless.
If https://krabbel.dev could support that feature, that would be even better!

vincentfretin:
first step:
- #5838

vincentfretin:
@Utopiah if you're willing to, could you rewrite the section to use codesandbox urls, like this:
https://codesandbox.io/p/sandbox/github/aframevr/aframe/tree/gh-pages/examples/docs/community-components
It auto opens the index.html file.

Utopiah:
@mrxz is it something that'd be good for krabbel and thus you or might it bring too much early attention and thus no be good for resources?

Utopiah:
FWIW I do think replacing with an alternative is better... but until it's done I also think removing broken links for newcomers is more important. Expert will know that they didn't do anything wrong, beginners will assume they somehow messed up and won't have a first good contact with AFrame.

Utopiah:
@vincentfretin unlucky but I just try the CSB preview URL in the HMD and the "open code sandbox" is just above the VR button, so I didn't manage to test it in VR.

mrxz:
> @mrxz is it something that'd be good for krabbel and thus you or might it bring too much early attention and thus no be good for resources?

Went into some detail in a comment on the issue @vincentfretin linked to, see https://github.com/aframevr/aframe/issues/5737#issuecomment-4802876710

In any case, I'm totally fine with linking to Krabbel as a ways to get started and experiment with A-Frame. Though, unlike Glitch, the intention isn't for hosting projects (no custom domain, no backend, no CDN, etc...). For this section in particular that would mean rewording it quite a bit. Though to a large extend the same would apply for codesandbox.io. Probably best to list them both along side Codepen, as functionality wise they are closer together than Codepen and Glitch were.

vincentfretin:
> @vincentfretin unlucky but I just try the CSB preview URL in the HMD and the "open code sandbox" is just above the VR button, so I didn't manage to test it in VR.

Ah I didn't see that! They injects the button with a random id. I managed to put the button at the top with:
```html
    <style>
      iframe[id^="sb__open-sandbox"] {
        top: 16px !important;
        bottom: auto !important;
      }
    </style>
```

maybe we could add the rule in aframe style.css? Although I don't quite like to put css to fix external services css in core, that would be the easiest solution to fix any aframe project on codesandbox. We could also inject that style tag in all examples when we publish on the gh-pages branch, but that will apply only for projects created from the gh-pages branch.

vincentfretin:
- #5839

dmarcos:
> > @vincentfretin unlucky but I just try the CSB preview URL in the HMD and the "open code sandbox" is just above the VR button, so I didn't manage to test it in VR.
> 
> Ah I didn't see that! They injects the button with a random id. I managed to put the button at the top with:
> 
> ```
>     <style>
>       iframe[id^="sb__open-sandbox"] {
>         top: 16px !important;
>         bottom: auto !important;
>       }
>     </style>
> ```
> 
> maybe we could add the rule in aframe style.css? Although I don't quite like to put css to fix external services css in core, that would be the easiest solution to fix any aframe project on codesandbox. We could also inject that style tag in all examples when we publish on the gh-pages branch, but that will apply only for projects created from the gh-pages branch.

we can add it to a-frame styles I guess. just add a good comment

vincentfretin:
Yes, the css rule was added in #5841 and will be effective in the next release of aframe so any project on CodeSandbox will have the fix. For the examples from the repo, the css rule is also added by the new remix-dropdown.js that adds the REMIX button for supported platforms, currently Krabbel and CodeSandbox.

We need help to rewrite that section in installation.md with Krabbel and CodeSandbox links, and to fix the first point of #5737 to replace the hello world link everywhere, I suggest just to point to https://krabbel.dev instead, but make sure to explain that those platforms are dev platforms not hosting platforms.

There also other exported glitches we need to put back in the repo in the examples/docs directory and assets in https://github.com/aframevr/assets with the correct relative urls, eventually some small changes to work on latest aframe.

PR #5833, opened 2026-06-15
Fix obb-collider to work on text-geometry

vincentfretin:
**Description:**

Fix `obb-collider` to work on `text-geometry` with `obb-collider="centerModel: true"`

**Changes proposed:**
- Don't assume mesh comes from glf-model, use generic `el.getObject3D('mesh');`
- Call `if (this.data.centerModel) { this.centerModel(); }` before `this.updateCollider();` in init, same as in `onModelLoaded`, text-geometry doesn't have a model-loaded event so it didn't center the mesh


dmarcos:
thanks

PR #5834, opened 2026-06-21
[docs] Fix code snippet for AFRAME_ASYNC, need to call emitReady(), not ready()

vincentfretin:
And error slipped through #5481 `ready()` doesn't exist, it's `emitReady()`
The paragraph above the code snippet was correct, it's just the code snippet that was wrong.


dmarcos:
thank you

PR #5838, opened 2026-06-24
Point gh-pages examples at the released A-Frame CDN build

vincentfretin:
Rewrite the example aframe-master.js and aframe-master.module.min.js references to `https://aframe.io/releases/<version>/aframe(.module).min.js`, taking the version from package.json. Read the super-three version from the released tag's package.json instead of master so the examples reference the three version that shipped with this release.


dmarcos:
Thanks

PR #5839, opened 2026-06-28
Reposition CodeSandbox button in gh-pages examples

vincentfretin:
## Description

When generating the gh-pages examples, inject a small `<style>` block that moves the CodeSandbox "Open Sandbox" button to the top of the page so it no longer overlaps the VR/AR enter buttons.

The style is only injected into the example HTML pages where a CDN link was actually replaced (tracked via the `hasChanged` result from `replace-in-file`), so untouched pages are left alone.

### DIFF
--- a/scripts/preghpages.js
+++ b/scripts/preghpages.js
@@ -29,14 +29,35 @@ shell.cp('-r', [
   '*.md'
 ], 'gh-pages');
 
+// Track which HTML files had at least one link replaced, so the CodeSandbox
+// style below is only injected into the examples we actually modified.
+const modifiedFiles = new Set();
+
 function htmlReplace (before, after) {
-  replaceInFileSync({
+  const results = replaceInFileSync({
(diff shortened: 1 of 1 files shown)


vincentfretin:
Alternative is to include the css rule directly in aframe.css #5841 but it will be working only in the next aframe release.

dmarcos:
injecting perhaps si a bit too magical. we can add it to aframe styles. I can redeploy the examples

vincentfretin:
We actually need a new aframe build so a new official release 1.8.1 to make the aframe styles effective.
I was thinking about including a script tag remix-dropdown.js in all examples in a similar way of this PR that creates a dropdown to remix on codesandbox or krabbel, we could also inject the style via this script to fix the Open in Sandbox button, that way no immediate release needed.

Closing this one since #5841 is merged. I will create another PR with my proposal.

PR #5840, opened 2026-06-28
Rewrite example js/assets paths to the gh-pages CDN

vincentfretin:
## Description

Point the `../../js/` and `../../assets/` references in the gh-pages examples at the jsDelivr-hosted `aframevr/aframe@gh-pages` copies, e.g.:

`../../js/hide-on-play.js` → `https://cdn.jsdelivr.net/gh/aframevr/aframe@gh-pages/examples/js/hide-on-play.js`

This makes the released examples load their helper scripts and assets from the CDN rather than relative paths.


mrxz:
Looks good, though I'm slightly concerned with the fragility of the approach. For example if we end up with a deeper nested examples at some point, it might reference `../../../js/`, so one more indirection.

Maybe we could expand the regex to include the opening `"`? It won't solve the above, but should avoid generating an invalid URL at least.

mrxz:
Also, now that the number of replacements is increasing, it might make sense to ensure these are done in one go, instead of calling `replaceInFileSync` multiple times. ~~Not really familiar with the `replace-in-file` module, but based on the documentation this seems possible: https://github.com/adamreisnz/replace-in-file#multiple-replacements-with-different-options~~

Edit: never mind, seems to be doing a loop as well.

scripts/preghpages.js
```diff
@@ -40,3 +40,5 @@ function htmlReplace (before, after) {
 htmlReplace('../../../dist/aframe-master.module.min.js', `https://aframe.io/releases/${aframeVersion}/aframe.module.min.js`);
 htmlReplace('../../../dist/aframe-master.js', `https://aframe.io/releases/${aframeVersion}/aframe.min.js`);
 htmlReplace(/\.\.\/\.\.\/\.\.\/super-three-package/g, `https://cdn.jsdelivr.net/npm/super-three@${threeVersion}`);
+htmlReplace(/\.\.\/\.\.\/js\//g, 'https://cdn.jsdelivr.net/gh/aframevr/aframe@gh-pages/examples/js/');
+htmlReplace(/\.\.\/\.\.\/assets\//g, 'https://cdn.jsdelivr.net/gh/aframevr/aframe@gh-pages/examples/assets/');
```
mrxz:
```js
function htmlReplace (before, after) {
  return (input) => input.replace(before, after);
}

replaceInFileSync({
  files: 'gh-pages/**/*.html',
  processor: [
    htmlReplace('../../../dist/aframe-master.module.min.js', `https://aframe.io/releases/${aframeVersion}/aframe.module.min.js`),
    htmlReplace('../../../dist/aframe-master.js', `https://aframe.io/releases/${aframeVersion}/aframe.min.js`),
    htmlReplace(/\.\.\/\.\.\/\.\.\/super-three-package/g, `https://cdn.jsdelivr.net/npm/super-three@${threeVersion}`),
    htmlReplace(/\.\.\/\.\.\/js\//g, 'https://cdn.jsdelivr.net/gh/aframevr/aframe@gh-pages/examples/js/'),
    htmlReplace(/\.\.\/\.\.\/assets\//g, 'https://cdn.jsdelivr.net/gh/aframevr/aframe@gh-pages/examples/assets/')
  ]
});
```

This way we can avoid opening and writing files multiple times.

vincentfretin: Yes, I even removed the replace-in-file dependency, actually removing 10 packages, and I wrote a simple 9 lines replaceInFileSync function. We have fs.globSync since node 22, no need for a glob dependency.

vincentfretin:
I did a copy of the examples directory before and after the refactoring and compared, it's the same. Good to merge.

dmarcos:
thanks

PR #5841, opened 2026-06-28
Move the CodeSandbox "Open Sandbox" button to the top left

vincentfretin:
Move the CodeSandbox "Open Sandbox" button to the top left so it does not overlap the VR/AR buttons.

Alternative to #5839


dmarcos:
thanks

PR #5847, opened 2026-07-04
Add renderer backend and reversedDepthBuffer properties, pass multiview option to WebGPURenderer

### LINKED ISSUE #5749: Add support for multiview with WebGPURenderer
You can use the new WebGPURenderer with aframe 1.7.1 with importmap, if you alias three to three.webgpu.js, aframe will use WebGPURenderer instead of WebGLRenderer, see source of the example https://aframe.io/aframe/examples/showcase/webgpu/ and using the latest super-three 0.177.0. But to use multiview with WebGPURenderer you will need a change here https://github.com/aframevr/aframe/blob/9ff61ffaf98c42b255fb17674eddb0d85c3239ba/src/core/scene/a-scene.js#L529-L547 the option is just called multiview and not multiviewStereo in WebGPURenderer it seems https://github.com/mrdoob/three.js/blob/b8f5c787fe6afbcbc442a6990f36a6c42d7405ca/src/renderers/webgpu/WebGPURenderer.js#L40C26-L40C35

Update the example https://aframe.io/aframe/examples/performance/multiview-extension/?multiview=on or duplicate it to test it with the new WebGPURenderer.
I was answering this on discord, so I'm creating the issue if anyone want to test it and do the PR.

vincentfretin:
Fixes #5749

## New renderer properties

- `multiviewStereo` is now also passed as the `multiview` option used by `WebGPURenderer` (`WebGLRenderer` keeps reading `multiviewStereo`, each renderer ignores the other's key).
- New `reversedDepthBuffer` property, passed as-is to both `WebGLRenderer` and `WebGPURenderer`.
- New `backend` property (`webgpu` or `webgl`): `backend: webgl` maps to the `forceWebGL` option of `WebGPURenderer`. When unset, nothing is passed so the WebGPU build keeps its current behavior (try WebGPU, fall back to WebGL2). Ignored by `WebGLRenderer`.

## Wait for async renderer init before starting the render loop

`WebGPURenderer` initializes its backend asynchronously via `init()`. Previously `renderStarted` was set and `renderstart` emitted while the backend was still initializing (three.js `setAnimationLoop` recovers by awaiting init internally, but anything reacting to `renderstart` calling e.g. `renderer.render()` or `compileAsync()` would throw `.render() called before the backend is initialized`). The render loop kickoff now awaits `renderer.init()` when it exists; `WebGLRenderer` has no `init` method and keeps the fully synchronous path.

## Tests

The renderer attribute parsing is extracted from `setupRenderer()` into `getRendererConfig()` so it is unit-testable without creating a WebGL context (`setupRenderer` is stubbed on CI). Added tests for `multiviewStereo`/`multiview`, `reversedDepthBuffer`, `backend` and the async init behavior. There were no existing tests covering `multiviewStereo`.

Also adds a code note that `setPoseTarget` (A-Frame specific super-three patch to `WebXRManager`) will need an equivalent in the `XRManager` used by `WebGPURenderer` when WebXR support is added.

Verified at runtime with both builds (`npm run dev` and `npm run start:webgpu`) driving headless Chrome: options reach the renderer (`backend.isWebGLBackend`, `reversedDepthBuffer`, `xr._useMultiviewIfPossible`) and `renderer.initialized` is now `true` when `renderstart` fires. Actual multiview rendering in VR still needs a headset test with `examples/performance/multiview-extension`.

### DIFF
--- a/docs/components/renderer.md
+++ b/docs/components/renderer.md
@@ -28,7 +28,7 @@ It also configures presentation attributes when entering WebVR/WebXR.
 | Property                | Description                                                                     | Default Value |
 |-------------------------|---------------------------------------------------------------------------------|---------------|
 | antialias               | Whether to perform antialiasing. If `auto`, antialiasing is disabled on mobile. | auto          |
-| backend                 | Backend used by `THREE.WebGPURenderer`: webgpu or webgl. Ignored with the default build using `THREE.WebGLRenderer`. | webgpu        |
+| backend                 | Backend used by `THREE.WebGPURenderer`: auto (WebGPU if available, WebGL 2 otherwise) or webgl. Ignored with the default build using `THREE.WebGLRenderer`. | auto          |
 | colorManagement         | Whether to use a color-managed linear workflow.                                 | true          |
 | highRefreshRate         | Increases frame rate from the default (for browsers that support control of frame rate). | false         |
 | foveationLevel          | Amount of foveation used in VR to improve perf, from 0 (min) to 1 (max).        | 1             |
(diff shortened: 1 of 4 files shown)


mrxz:
> New `backend` property (`webgpu` or `webgl`): `backend: webgl` maps to the `forceWebGL` option of `WebGPURenderer`. When unset, nothing is passed so the WebGPU build keeps its current behavior (try WebGPU, fall back to WebGL2). Ignored by `WebGLRenderer`.

I think it would be better to use the values `webgl` and `auto` for now (defaulting to the latter). It's quite likely that some users might end up with setups that _require_ WebGPU and won't work with the WebGL backend of the `WebGPURenderer`. Down the line we could then introduce a `webgpu` value for this property that asserts WebGPU support.

For reference, quite a few WebGPU examples in Three.js _require_ WebGPU support and show a simple error message _"Your browser does not support [WebGPU](https://gpuweb.github.io/gpuweb/) yet"_ when it's not supported.

vincentfretin:
yes, I changed webpgu to auto

PR #5848, opened 2026-07-04
Adapt reflection and screenshot components to WebGPURenderer

vincentfretin:
Makes the `reflection` and `screenshot` components work with `WebGPURenderer` (three.webgpu.js build, see #5749 for context).

## Changes

- **Cube render target**: `WebGLCubeRenderTarget` is not exported by the three.webgpu.js build, which uses `CubeRenderTarget` instead. Both components now select the class at runtime with the same trick a-scene uses to select the renderer implementation, so webpack can't statically resolve the export. This also removes the `WebGLCubeRenderTarget` webpack warning when building with `WEBGPU=true`. `THREE.CubeCamera` works unchanged with either target.
- **Equirectangular projection**: `WebGPURenderer` does not support `RawShaderMaterial`/`ShaderMaterial` (not in its node library). When `THREE.TSL` is available, the cubemap-to-equirectangular projection is built as a `MeshBasicNodeMaterial` with TSL nodes implementing the same longitude/latitude math as the GLSL shaders; the cube texture is swapped per capture through the texture node's `.value`.
- **Pixel readback**: `WebGPURenderer` only supports asynchronous read back. `renderCapture` uses `readRenderTargetPixelsAsync` when `readRenderTargetPixels` is not available. `getCanvas()` still returns the canvas synchronously with `WebGLRenderer` (no breaking change) and returns a `Promise` resolving to the canvas with `WebGPURenderer` (documented in `screenshot.md`). `capture()` works on both.
- **RGBAFormat**: the equirect cube render target used `RGBFormat`, which is not color-renderable with the WebGL 2 backend of `WebGPURenderer` (incomplete framebuffer, black capture). Switched to `RGBAFormat`; the projection outputs alpha 1 and the readback path already assumed 4 bytes per pixel.

## Verification

Tested with a scene with distinct sky/box/ground colors on both builds (`npm run dev` and `npm run start:webgpu`) driving headless Chrome, probing pixels of the resulting canvas:

- Perspective and equirectangular captures produce identical colors/orientation on `WebGLRenderer` and on `WebGPURenderer` (WebGL 2 fallback backend — no WebGPU adapters on Linux Chrome).
- The `reflection` component sets a live cube texture as `scene.environment` on both renderers without errors.
- Scene component tests and lint pass.

Note: the actual WebGPU backend readback couldn't be tested on Linux; worth a quick check of equirect orientation in a desktop Chrome with WebGPU support.

`ar-hit-test.js` uses the same `readRenderTargetPixels` pattern but is WebXR-only, so it's left untouched until WebXR support lands for `WebGPURenderer`.

### DIFF
--- a/src/components/scene/screenshot.js
+++ b/src/components/scene/screenshot.js
@@ -6,6 +6,10 @@ import * as THREE from 'three';
 // Written so that Webpack can't statically determine the export used;
 // only one of the two exists depending on the three.js build.
 var cubeRenderTargetImpl = ['WebGLCubeRenderTarget', 'CubeRenderTarget'].find(function (x) { return THREE[x]; });
+// TSL and MeshBasicNodeMaterial only exist in the three.js build with
+// WebGPURenderer, same Webpack trick as above.
+var TSL = ['TSL'].map(function (x) { return THREE[x]; })[0];
+var MeshBasicNodeMaterial = ['MeshBasicNodeMaterial'].map(function (x) { return THREE[x]; })[0];
 
 var VERTEX_SHADER = [
   'attribute vec3 position;',
@@ -70,7 +74,7 @@ export var Component = registerComponent('screenshot', {
       this.cubeMapSize = device ? device.limits.maxTextureDimension2D : 2048;
     }
     // WebGPURenderer does not support RawShaderMaterial, use a node material with TSL.
-    if (THREE.TSL) {
+    if (TSL) {
       this.material = this.createNodeMaterial();
     } else {
       this.material = new THREE.RawShaderMaterial({
@@ -98,8 +102,7 @@ export var Component = registerComponent('screenshot', {
    * WebGPURenderer which does not support RawShaderMaterial.
    */
   createNodeMaterial: function () {
-    var TSL = THREE.TSL;
-    var material = new THREE.MeshBasicNodeMaterial({side: THREE.DoubleSide});
+    var material = new MeshBasicNodeMaterial({side: THREE.DoubleSide});
     var uv = TSL.uv();
     var longitude = TSL.float(1).sub(uv.x).mul(2 * Math.PI).sub(Math.PI).add(Math.PI / 2);
     var latitude = uv.y.mul(Math.PI);
--- a/src/components/scene/screenshot.js
+++ b/src/components/scene/screenshot.js
@@ -60,9 +60,15 @@ export var Component = registerComponent('screenshot', {
   setup: function () {
     var el = this.el;
     if (this.canvas) { return; }
-    var gl = el.renderer.getContext();
-    if (!gl) { return; }
-    this.cubeMapSize = gl.getParameter ? gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE) : 2048;
+    var ctx = el.renderer.getContext();
+    if (!ctx) { return; }
+    if (ctx.getParameter) {
+      this.cubeMapSize = ctx.getParameter(ctx.MAX_CUBE_MAP_TEXTURE_SIZE);
+    } else {
+      // ctx is a GPUCanvasContext; cube map faces are limited by maxTextureDimension2D.
+      var device = el.renderer.backend.device;
+      this.cubeMapSize = device ? device.limits.maxTextureDimension2D : 2048;
+    }
     // WebGPURenderer does not support RawShaderMaterial, use a node material with TSL.
     if (THREE.TSL) {
       this.material = this.createNodeMaterial();
@@ -73,6 +79,7 @@ export var Component = registerComponent('screenshot', {
         fragmentShader: FRAGMENT_SHADER,
         side: THREE.DoubleSide
       });
+      this.cubeTextureUniform = this.material.uniforms.map;
     }
     this.quad = new THREE.Mesh(
       new THREE.PlaneGeometry(1, 1),
@@ -101,8 +108,8 @@ export var Component = registerComponent('screenshot', {
       TSL.cos(latitude),
       TSL.cos(longitude).mul(TSL.sin(latitude)).negate()
     );
-    this.cubeTextureNode = TSL.cubeTexture(new THREE.CubeTexture(), dir);
-    material.colorNode = TSL.vec4(this.cubeTextureNode.rgb, 1);
+    this.cubeTextureUniform = TSL.cubeTexture(new THREE.CubeTexture(), dir);
+    material.colorNode = TSL.vec4(this.cubeTextureUniform.rgb, 1);
     return material;
   },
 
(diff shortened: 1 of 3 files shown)


vincentfretin:
Testing help wanted: on this machine I could only verify `WebGPURenderer` running on its WebGL 2 fallback backend (Chrome on Linux has no WebGPU adapters). If someone has a machine with WebGPU support (`chrome://gpu` shows WebGPU: Hardware accelerated), could you test this PR with the actual WebGPU backend?

```sh
gh pr checkout 5848
# or without the GitHub CLI:
# git fetch https://github.com/vincentfretin/aframe.git webgpu-reflection-screenshot && git checkout FETCH_HEAD
npm ci
npm run dist
npm run start:webgpu
```

Then open `https://localhost:8080/examples/` (accept the self-signed certificate), pick any example and check in the devtools console that the WebGPU backend is really active:

```js
AFRAME.scenes[0].renderer.backend.isWebGPUBackend // should be true
```

Things to verify:

- `<ctrl> + <alt> + s`: perspective screenshot downloads, correct content, not upside down.
- `<ctrl> + <alt> + <shift> + s`: equirectangular screenshot, correct content and orientation. This exercises the new TSL projection material and the async pixels read back, and is the part I couldn't fully verify: the readback orientation could differ between the WebGPU backend and the WebGL 2 fallback backend.
- Reflection: run `document.querySelector('a-scene').setAttribute('reflection', '')` in the console, check there is no error and `AFRAME.scenes[0].object3D.environment` is set.

src/components/scene/screenshot.js
```diff
@@ -57,13 +62,18 @@ export var Component = registerComponent('screenshot', {
     if (this.canvas) { return; }
     var gl = el.renderer.getContext();
```
mrxz: Nitpick: the name `gl` is no longer accurate as it might be a `GPUCanvasContext`. Maybe `ctx` would be an apt alternative covering both?

src/components/scene/screenshot.js
```diff
@@ -57,13 +62,18 @@ export var Component = registerComponent('screenshot', {
...
     if (!gl) { return; }
-    this.cubeMapSize = gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE);
-    this.material = new THREE.RawShaderMaterial({
-      uniforms: {map: {type: 't', value: null}},
-      vertexShader: VERTEX_SHADER,
-      fragmentShader: FRAGMENT_SHADER,
-      side: THREE.DoubleSide
-    });
+    this.cubeMapSize = gl.getParameter ? gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE) : 2048;
```
mrxz: Wasn't immediately clear why `2048` would be a safe, though it appears that `maxTextureDimension2D` (which applies for cube map faces AFAICT) is at least 8192 and 4096 in the recent compatibility mode. Since the code later on clamps it to 2048 either way, it's fine using that value.

Ideally we'd query `maxTextureDimension2D`, or at the very least add a comment describing why this is safe to do.

src/components/scene/screenshot.js
```diff
@@ -155,7 +185,11 @@ export var Component = registerComponent('screenshot', {
       el.camera.getWorldQuaternion(cubeCamera.quaternion);
       // Render scene with cube camera.
       cubeCamera.update(el.renderer, el.object3D);
-      this.quad.material.uniforms.map.value = cubeCamera.renderTarget.texture;
+      if (this.cubeTextureNode) {
+        this.cubeTextureNode.value = cubeCamera.renderTarget.texture;
+      } else {
+        this.quad.material.uniforms.map.value = cubeCamera.renderTarget.texture;
+      }
```
mrxz: Since both the uniform and `TextureNode` have the texture in a `.value` property, we can store `this.cubeTextureUniform` for both cases (TSL and ShaderMaterial). In fact `TSL.uniformTexture` is the same as `TSL.texture` so should be fine to change `cubeTextureNode` to `cubeTextureUniform` semantically.

mrxz:
> Testing help wanted: on this machine I could only verify `WebGPURenderer` running on its WebGL 2 fallback backend (Chrome on Linux has no WebGPU adapters). If someone has a machine with WebGPU support (`chrome://gpu` shows WebGPU: Hardware accelerated), could you test this PR with the actual WebGPU backend?

It should be possible to enable WebGPU on Linux on most systems, see: https://github.com/gpuweb/gpuweb/wiki/Implementation-Status#chromium-chrome-edge-etc

In any case, a quick test shows that both the `perspective` and `equirectangular` results are upside down when using WebGPU. Also, the screenshot shortcuts no longer seem to work, but this appears to be a regression introduced with https://github.com/aframevr/aframe/pull/5442. Essentially the `onKeyDown` is no longer binding this by the time `play` is called, since the entire `init` is renamed to `setup` and deferred. Programmatically triggering the captures does still work, which I've used.

Edit: opened a PR with a fix: https://github.com/aframevr/aframe/pull/5849

> * Reflection: run `document.querySelector('a-scene').setAttribute('reflection', '')` in the console, check there is no error and `AFRAME.scenes[0].object3D.environment` is set.

The `environment` property appears to be set correctly. At first blush the result looks correct visually as well, but haven't tested/compared extensively.

vincentfretin:
Oh I completely missed that it was working now on Linux behind additional cli flags, thank you!
I got webpgu working on Ubuntu 22.04, AMD Ryzen 5 5600X, nvidia driver 610.
```
cp /usr/share/applications/google-chrome.desktop ~/.local/share/applications/
edit ~/.local/share/applications/google-chrome.desktop and change Exec to:
Exec=/usr/bin/google-chrome-stable --enable-unsafe-webgpu --ozone-platform=x11 --use-angle=vulkan --enable-features=Vulkan,VulkanFromANGLE %U
update-desktop-database ~/.local/share/applications/
```

src/components/scene/screenshot.js
vincentfretin: Renamed to `ctx` in 2a9ab7de.

src/components/scene/screenshot.js
vincentfretin: Done in 2a9ab7de: it now queries `maxTextureDimension2D` from the device limits (with a 2048 fallback if the device is somehow not available yet).

src/components/scene/screenshot.js
vincentfretin: Nice, done in 2a9ab7de: `this.cubeTextureUniform` is now set in both branches (the `map` uniform for ShaderMaterial and the `cubeTexture` node for TSL) and `setCapture` has a single `.value` assignment.

vincentfretin:
@mrxz Thanks a lot for testing and for the Linux flags pointer! I got the actual WebGPU backend working locally (NVIDIA, Vulkan) with `--enable-unsafe-webgpu --ozone-platform=x11 --use-angle=vulkan --enable-features=Vulkan,VulkanFromANGLE` and could reproduce and fix the upside down captures in 2a9ab7de.

The cause: pixels read back from WebGL are bottom-up but top-down with the WebGPU backend, and the equirectangular projection quad renders vertically inverted so it needs the opposite flip of the perspective projection. The vertical flip is now based on `renderer.coordinateSystem`. Note it depends on the backend, not the renderer class: `WebGPURenderer` on its WebGL 2 fallback backend reads back bottom-up like `WebGLRenderer`, which is why my initial fallback-only testing looked correct.

Verified on the three configurations by probing pixels of the resulting canvas on a scene with distinct sky/box/ground colors:

| Configuration | perspective | equirectangular | reflection |
|---|---|---|---|
| WebGPU backend (NVIDIA/Vulkan) | ✅ | ✅ | ✅ |
| WebGL 2 fallback backend | ✅ | ✅ | ✅ |
| WebGLRenderer | ✅ (sync) | ✅ (sync) | ✅ |

Review comments also addressed in the same commit. And thanks for tracking down the keyboard shortcut regression separately in #5849.

PR #5849, opened 2026-07-05
Fix screenshot component keyboard shortcuts

mrxz:
**Description:**
The keyboard shortcuts for making screenshots weren't properly working since #5442. The deferred setup takes place _after_ `play` is called which causes the `onKeyDown` event handler to be used before it's bound to the current this.

**Changes proposed:**
- Re-introduce a `init` method for binding `onKeyDown`.


dmarcos:
Thank you

